PR #171, opened 2023-09-10
Added WebUrlDocumentLoader support

jmgang:
- Added a loader for loading simple web url documents using jsoup, with an option to retain or remove html tags. 
- Fixed a misspelling of jsoup version on the pom file.


langchain4j:
@jmgang Hi, thank you for your contribution! We have discussed this feature before and I think I did a mistake. There is already a way to load web pages:
`Document document = UrlDocumentLoader.load("http://example.com", HTML)`
I feel that WebUrlDocumentLoader will be redundant. What do you think?
And sorry for this, my bad...

jmgang:
> @jmgang Hi, thank you for your contribution! We have discussed this feature before and I think I did a mistake. There is already a way to load web pages: `Document document = UrlDocumentLoader.load("http://example.com", HTML)` I feel that WebUrlDocumentLoader will be redundant. What do you think? And sorry for this, my bad...

@langchain4j I agree and no worries. I should have taken a look at UrlDocumentLoader to realize that it's doing the same thing as WebUrlDocumentLoader does. Closing this PR. Thanks for pointing that out.

Issue #97, opened 2023-08-15
langchain4j integration with redis


radhakrishna67:
I am waiting for a redis integration with langchain4j.

langchain4j:
@radhakrishna67 redis as an embedding store? we will add this to our backlog.

radhakrishna67:
Yes, Vectore Store

radhakrishna67:
is there any update when this is going to be implemented?

Martin7-1:
> is there any update when this is going to be implemented?

Hi! I'm trying to contribute redis integration. PR is expected to be mentioned within a week. Draft PR can see [PR](https://github.com/Martin7-1/langchain4j/pull/3)

langchain4j:
Was added long ago

PR #174, opened 2023-09-12
redis integration as embedding store

Martin7-1:
Hi! I make some progress on integrating with `Redis` to support #97. Most of the work is done, just remaining some points to check:

1. `Embedding` results in `findRelevant` method depends on Redis encoding. The whole workflow is that we pass a `byte` array as `Embedding` and Redis encodes it as a string, but it seems that we can't decode it (currently I use `ISO_8859_1` charset to decode it because it seems the most correct in my test, you can check out `RedisEmbeddingStoreImpl#toEmbeddingMatch` method). it seems that we can't get correct embedding from search result. (But I am trying to figure out a way to solve the problem)
2. I use Redissearch `KNN` algorithm with cosine similarity, but it seems that the score is not as expected (even though I use the same `referenceEmbedding` to search, it does not return any result whose score is 1). I don't know whether it's a problem or not, because when I use the command directly to search, it seems that it return the same result.
3. it seems that `Redis` do not support search based on both `maxResults` and `minScore`, at present I ignore `minScore` param. I am trying to find a way to support search on them at the same time

currently it's store in `HASH` like LangChain do, but `Redis` also support `JSON` to do vector search. Using `JSON` may solve problem 1. (I'm not sure)

Look Forward to your Code Review and Opinion!


langchain4j-redis/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.honton.chas</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+                <configuration>
+                    <!-- jedis use MIT license but is not acceptable(I don't know why) -->
```
langchain4j, line 79: try changing
`<name>(The MIT License|MIT License)</name>`
to
`<name>(The MIT License|MIT License|MIT)</name>`
in langchain4j-parent/pom.xml

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/DataType.java
```diff
@@ -0,0 +1,23 @@
+package dev.langchain4j.store.embedding.redis;
+
+/**
+ * Redis vector field data type
+ */
+public enum DataType {
```
langchain4j: can it be package-private?

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/DataType.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * only support FLOAT32
```
langchain4j: if we supports only FLOAT32, why do we need DataType class?

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/MetricType.java
```diff
@@ -0,0 +1,21 @@
+package dev.langchain4j.store.embedding.redis;
+
+/**
+ * Redis vector field distance Metric
+ */
+public enum MetricType {
```
langchain4j, line 6: can it be package-private?

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/MetricType.java
```diff
@@ -0,0 +1,21 @@
...
+
+    COSINE("COSINE"),
+    IP("IP"),
+    L2("L2");
+
+    private final String name;
```
langchain4j: why do you need name field when you can always get name of enum using `.name()` method?

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStoreImpl.java
```diff
@@ -0,0 +1,174 @@
...
+ */
+public class RedisEmbeddingStoreImpl implements EmbeddingStore<TextSegment> {
+
+    private static final Logger log = LoggerFactory.getLogger(RedisEmbeddingStoreImpl.class);
+    private static final Gson GSON = new Gson();
+
+    private final JedisPooled client;
+    private final RedisSchema schema;
+
+    public RedisEmbeddingStoreImpl(String url, Integer dimension) {
```
langchain4j: no authentication?

langchain4j-redis/src/test/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStoreImplTest.java
```diff
@@ -0,0 +1,84 @@
...
+    void testScore() {
+        String id = store.add(Embedding.from(Arrays.asList(0.50f, 0.85f, 0.760f, 0.24f)),
+                TextSegment.from("test string", Metadata.metadata("field", "value")));
+        System.out.println("id=" + id);
+
+        // use the same embedding to search
+        List<EmbeddingMatch<TextSegment>> res = store.findRelevant(Embedding.from(Arrays.asList(0.50f, 0.85f, 0.760f, 0.24f)), 1);
+        res.forEach(System.out::println);
+
+        // the result embeddingMatch score is 5.96046447754E-8, but expected is 1 because they are same vectors.
```
langchain4j, line 83: 5.96046447754E-8 is 0,00000006 which is very close to 0. It is a cosine distance (in the range [0..2] where 0 is identical and 2 is opposite), not score(in the range [0..1] where 0 is opposite and 1 is identical). In the [documentation](https://redis.io/docs/interact/search-and-query/search/vectors/) they subtract distance from 1 to get the cosine similarity (in range [-1..1] where -1 is opposite and 1 is identical): `1 - distance`. We need to further convert similarity to score: `score = (2 - distance) / 2`

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStoreImpl.java
```diff
@@ -0,0 +1,174 @@
...
+        url = ensureNotNull(url, "url");
+        dimension = ensureNotNull(dimension, "dimension");
+
+        client = new JedisPooled(url);
+        schema = RedisSchema.builder().dimension(dimension).build();
+
+        createIndex(schema.getIndexName());
```
langchain4j: can you move `isIndexExist()` check out of `createIndex()` method?
```java
if (!isIndexExist(..)) {
  createIndex(..);
}
```

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStoreImpl.java
```diff
@@ -0,0 +1,174 @@
...
+        int size = ids.size();
+        for (int i = 0; i < size; i++) {
+            String id = ids.get(i);
+            Embedding embedding = embeddings.get(i);
+            TextSegment textSegment = embedded == null ? null : embedded.get(i);
+            Map<byte[], byte[]> vectorField = new HashMap<>();
+            vectorField.put(schema.getVectorFieldName().getBytes(), ToByteArray(embedding.vector()));
+            if (textSegment != null) {
+                vectorField.put(schema.getScalarFieldName().getBytes(), textSegment.text().getBytes());
+                vectorField.put(schema.getMetadataFieldName().getBytes(), GSON.toJson(textSegment.metadata().asMap()).getBytes());
```
langchain4j: we plan to add filtering by metadata soon. Is it possible to do filtering when all metadata is stored in a single field as a byte array?

langchain4j:
Hi @Martin7-1 thanks for your effort!

> 1. `Embedding` results in `findRelevant` method depends on Redis encoding. The whole workflow is that we pass a `byte` array as `Embedding` and Redis encodes it as a string, but it seems that we can't decode it (currently I use `ISO_8859_1` charset to decode it because it seems the most correct in my test, you can check out `RedisEmbeddingStoreImpl#toEmbeddingMatch` method). it seems that we can't get correct embedding from search result. (But I am trying to figure out a way to solve the problem)

Worth to try JSON index, see below

> 2. I use Redissearch `KNN` algorithm with cosine similarity, but it seems that the score is not as expected (even though I use the same `referenceEmbedding` to search, it does not return any result whose score is 1). I don't know whether it's a problem or not, because when I use the command directly to search, it seems that it return the same result.

I have left a comment in the review regarding this

> 3. it seems that `Redis` do not support search based on both `maxResults` and `minScore`, at present I ignore `minScore` param. I am trying to find a way to support search on them at the same time

You can do minScore filtering manually after the search result from redis is returned, you can use `CosineSimilarity.between()` and `RelevanceScore.fromCosineSimilarity()`

> currently it's store in `HASH` like LangChain do, but `Redis` also support `JSON` to do vector search. Using `JSON` may solve problem 1. (I'm not sure)

Worth a try, they use JSON index in the example: https://redis.io/docs/interact/search-and-query/search/vectors/

langchain4j-redis/pom.xml
Martin7-1: It works!

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/DataType.java
Martin7-1: Actually I want to support user customizing  `RedisSchema`. But at this stage I think this function is not very important, so I think it can be retained package-public. It doesn't matter

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/MetricType.java
Martin7-1: see above in `DataType`

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/MetricType.java
Martin7-1: done, remove it already.

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/DataType.java
Martin7-1: already removed

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStoreImpl.java
Martin7-1: I forgot it, already added

langchain4j-redis/src/test/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStoreImplTest.java
Martin7-1: > 5.96046447754E-8 is 0,00000006 which is very close to 0. It is a cosine distance (in the range [0..2] where 0 is identical and 2 is opposite), not score(in the range [0..1] where 0 is opposite and 1 is identical). In the [documentation](https://redis.io/docs/interact/search-and-query/search/vectors/) they subtract distance from 1 to get the cosine similarity (in range [-1..1] where -1 is opposite and 1 is identical): `1 - distance`. We need to further convert similarity to score: `score = (2 - distance) / 2`

I see. Thanks for your explanation! I changed the way to calculate right score by `score = (2 - distance) / 2`

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStoreImpl.java
Martin7-1: done

langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStoreImpl.java
Martin7-1: metadata filtering is not supported using byte arrays, now I use JSON to store data

Martin7-1:
Thanks for your code review! I have made the following changes below:

1. try JSON Index, It works fine.
2. fix score calculation
3. filter by `minScore` after the search result from redis is returned

langchain4j:
@Martin7-1 awesome! Is it ready for review or do you plan to do more changes?

Martin7-1:
> @Martin7-1 awesome! Is it ready for review or do you plan to do more changes?

@langchain4j Hi! It seems that we will no longer use dynamic loading to load the embedding store. I think that I need a little time to remove it(maybe in Tuesday or Wednesday). You can review the other parts first and then I will change them together.

langchain4j:
@Martin7-1 I am a bit in a hurry, so I will take care of that, no worries. Thank you!

Martin7-1:
> @Martin7-1 I am a bit in a hurry, so I will take care of that, no worries. Thank you!

OK. Thanks!

PR #189, opened 2023-09-22
Better handling of whole number types in prompts

Heezer:
I've realized that the whole number types (int, long etc.) will be output with fraction digit 0 like "1.0" which leads sometimes to misinterpretation on AI engines side


ghost (approved):

langchain4j:
@Heezer thanks!

PR #203, opened 2023-09-30
"Getting started" fixed to use OpenAI dep instead of generic one

Heezer:


ghost (approved):
@Heezer thanks for fixing!

PR #211, opened 2023-10-08
feat: upgrade Lombok to 1.8.30 for JDK21 compatibility

SimonVerhoeven:
Using 1.8.28 results in 

` java: java.lang.NoSuchFieldError: Class com.sun.tools.javac.tree.JCTree$JCImport does not have member field 'com.sun.tools.javac.tree.JCTree qualid'`


ghost (approved):
@SimonVerhoeven thank you for fixing this!

Issue #217, opened 2023-10-10
[FEATURE] BOM: add Spring Boot modules


me0106:
Now, if you want to reference multiple langchain4j modules, you always need to define multiple 'artifacts'. 

like this: 
```xml
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>dev.langchain4j</groupId>
                <artifactId>langchain4j-core</artifactId>
                <version>0.23.0</version>
            </dependency>
            <dependency>
                <groupId>dev.langchain4j</groupId>
                <artifactId>langchain4j-open-ai</artifactId>
                <version>0.23.0</version>
            </dependency>
            <dependency>
                <groupId>dev.langchain4j</groupId>
                <artifactId>langchain4j-azure-open-ai</artifactId>
                <version>0.23.0</version>
            </dependency>
            <dependency>
                <groupId>dev.langchain4j</groupId>
                <artifactId>langchain4j-elasticsearch</artifactId>
                <version>0.23.0</version>
            </dependency>
        </dependencies>
    </dependencyManagement>
```
It would be great if a single BOM could be provided. Just like `jackson-bom`
```xml
    <dependencyManagement>
        <dependencies>
              <dependency>
                  <groupId>com.fasterxml.jackson</groupId>
                  <artifactId>jackson-bom</artifactId>
                  <version>${jackson-bom.version}</version>
                  <type>pom</type>
                  <scope>import</scope>
              </dependency>
        </dependencies>
    </dependencyManagement>

Martin7-1:
Sounds great. I think it's a good way to manage artifacts.

langchain4j:
@me0106 could you please clarify? Do you expect that declaring a single dependency will automatically import all available langchain4j modules? Or that it will only import right versions and you will still need to explicitly define which modules you want? Thanks

me0106:
@langchain4j 
This PR is actually what I need,
https://github.com/langchain4j/langchain4j/pull/227
I usually define all the dependencies required for the entire project in a parent to avoid various version definitions in the submodules, including variable references in the format of ${xxx.version}. If there is no unified BOM provided, I have to define all versions of langchain4j in the parent pom,
```xml
--------//parent pom
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>dev.langchain4j</groupId>
                <artifactId>langchain4j-core</artifactId>
                <version>0.23.0</version>
            </dependency>
            <dependency>
[...]

PR #227, opened 2023-10-13
add BOM to manage artifacts

Martin7-1:
see #217, it's a good way to manage artifacts.

Now users can use code below to import dependency:

```xml
    <dependencyManagement>
        <dependencies>
              <dependency>
                  <artifactId>langchain4j-bom</artifactId>
                  <version>${langchain4j.version}</version>
                  <type>pom</type>
                  <scope>import</scope>
              </dependency>
        </dependencies>
    </dependencyManagement>
```

### DIFF
--- a/langchain4j-bom/pom.xml
+++ b/langchain4j-bom/pom.xml
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>0.23.0</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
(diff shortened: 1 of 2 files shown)


Heezer (approved):
Looks good from side! 👍

langchain4j:
@Martin7-1 correct me if I am wrong, but I think this is not what was meant in https://github.com/langchain4j/langchain4j/issues/217. As far as I understood, the user wants a simple way to import all modules at once. The provided change helps with importing only versions of modules (not the modules themselves), but the versions are always the same, so there should be no problem with keeping them in sync even without BOM. Let me double check with the user...

Martin7-1:
@langchain4j Actually there doesn't seem to be a scenario where all modules need to be imported. Our modules contains different model providers and different embedding stores, and users just need to choose one of them. I think in #217, the user wants a way to manage related `langchain4j` modules, when updated, the user doesn't need to fix the conflict between modules(e.g. embedding store implementations depend on `langchain4j-core` module, change `EmbeddingStore` interface may need to change more than one module version).

You can wait until you have double checked with the user before reviewing the PR.

langchain4j:
@Martin7-1 I am not sure where from conflict between module versions can potentially come from. Every release we update the version for every dependency. So langchain4j-core:0.23.0 is compatible with langchain4j-vespa:0:23:0 and so on. This means that ideally users should update the version only in one place (if they use variables like ${langchain4j.version}), or for every imported module (but in any case to the same version).

Martin7-1:
@langchain4j I'm going to close this PR because the user didn't response, will reopen it if the user clarify his/her use case

PR #230, opened 2023-10-16
Add a MongoDB Embedded Store to the project

bvanloocke:
I wrote a MongoDBEmbeddingStore for personal use. I'm happy to contribute it to the main project as other developers may like to use MongoDB as well.

Unfortunately it works not on standard mongodb database. It's a feature of the  MongoDB Atlas, the managed, cloud-based database service provided by MongoDB. Luckily for us it has a free tier. 
That makes it difficult for local testing.


BartDuwez (approved):
We need this too!

PR #235, opened 2023-10-18
Make some aspects of the project swappable by external integrations

geoand:
# Purpose

This purpose of this pull request is to make it possible for advanced integrators to swap out some of the dependencies with others.

For a first iteration, I did not want to require current consumers of this library to have to make any changes - for them this pull requests changes nothing whatsoever.
This means that I did not do any sort of of changes to the structure or name of the Maven modules. I believe that if we want to do that, if can be done at a latter point.

# Changes

## Gson

All this change does is move Gson out of the model classes. This is needed by integrations that plan to exclude Gson from the classpath, so that the models can continue to be loaded without incurring ClassLoading issues.

## Template SPI

I introduced `PromptTemplateFactory` which can be used for integrations to supply their own version of the template engine.
Implementations are loaded using Java's ServiceLoader mechanism and if none are found, the default `MustachePromptTemplateFactory` is used.

I have leveraged this SPI my Quarkus integration [here](https://github.com/geoand/quarkus-langchain4j/blob/487711af4c7b465dbc40b90314c67de712ab8a15/runtime/src/main/java/io/quarkiverse/langchain4j/QuarkusPromptTemplateFactory.java).

## Make OpenAiStreamingResponseBuilder thread-safe

This class needs to be thread safe because it is called when a streaming result comes back
and there is no guarantee that this thread will be the same as the one that initiated the request

## Improve Javadoc of `@V`

This simply generalizes the condition under which `@V` is not necessary

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolExecutionRequest.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolExecutionRequest.java
@@ -1,19 +1,9 @@
 package dev.langchain4j.agent.tool;
 
-import com.google.gson.Gson;
-import com.google.gson.reflect.TypeToken;
-
-import java.lang.reflect.Type;
-import java.util.Map;
+import static dev.langchain4j.internal.Utils.quoted;
 import java.util.Objects;
 
-import static dev.langchain4j.internal.Utils.quoted;
-
 public class ToolExecutionRequest {
-
-    private static final Gson GSON = new Gson();
-    private static final Type MAP_TYPE = new TypeToken<Map<String, Object>>() {
-    }.getType();
 
     private final String name;
     private final String arguments;
@@ -29,15 +19,6 @@ public class ToolExecutionRequest {
 
     public String arguments() {
         return arguments;
-    }
-
-    public Map<String, Object> argumentsAsMap() {
-        return GSON.fromJson(arguments, MAP_TYPE);
-    }
-
-    public <T> T argument(String name) {
-        Map<String, Object> arguments = argumentsAsMap();
-        return (T) arguments.get(name);
     }
 
     @Override
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolExecutionRequestUtil.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolExecutionRequestUtil.java
@@ -1,0 +1,21 @@
+package dev.langchain4j.agent.tool;
+
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Type;
+import java.util.Map;
+
+public class ToolExecutionRequestUtil {
+    public static final Gson GSON = new Gson();
+    public static final Type MAP_TYPE = new TypeToken<Map<String, Object>>() {
+    }.getType();
+
+    public static <T> T argument(ToolExecutionRequest toolExecutionRequest, String name) {
+        Map<String, Object> arguments = argumentsAsMap(toolExecutionRequest.arguments()); // TODO cache
+        return (T) arguments.get(name);
+    }
+
+    public static Map<String, Object> argumentsAsMap(String arguments) {
+        return GSON.fromJson(arguments, MAP_TYPE);
+    }
+}
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/input/MustachePromptTemplateFactory.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/input/MustachePromptTemplateFactory.java
@@ -1,0 +1,35 @@
+package dev.langchain4j.model.input;
+
+import com.github.mustachejava.DefaultMustacheFactory;
+import com.github.mustachejava.Mustache;
+import com.github.mustachejava.MustacheFactory;
+import dev.langchain4j.spi.prompt.PromptTemplateFactory;
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.util.Map;
+
+class MustachePromptTemplateFactory implements PromptTemplateFactory {
+
+    private static final MustacheFactory MUSTACHE_FACTORY = new DefaultMustacheFactory();
+
+    @Override
+    public MustacheTemplate create(PromptTemplateFactory.Input input) {
+        StringReader stringReader = new StringReader(input.getTemplate());
+        return new MustacheTemplate(MUSTACHE_FACTORY.compile(stringReader, input.getName()));
+    }
+
+    static class MustacheTemplate implements Template {
+
+        private final Mustache mustache;
+
+        MustacheTemplate(Mustache mustache) {
+            this.mustache = mustache;
+        }
+
+        public String render(Map<String, Object> vars) {
+            StringWriter writer = new StringWriter();
+            mustache.execute(writer, vars);
+            return writer.toString();
+        }
+    }
+}
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/input/PromptTemplate.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/input/PromptTemplate.java
@@ -1,21 +1,15 @@
 package dev.langchain4j.model.input;
 
-import com.github.mustachejava.DefaultMustacheFactory;
-import com.github.mustachejava.Mustache;
-import com.github.mustachejava.MustacheFactory;
-
-import java.io.StringReader;
-import java.io.StringWriter;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+import static java.util.Collections.singletonMap;
+import dev.langchain4j.spi.ServiceHelper;
+import dev.langchain4j.spi.prompt.PromptTemplateFactory;
 import java.time.Clock;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.util.HashMap;
 import java.util.Map;
-
-import static dev.langchain4j.internal.ValidationUtils.ensureNotBlank;
-import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
-import static java.util.Collections.singletonMap;
 
 /**
  * Represents a template of a prompt that can be reused multiple times.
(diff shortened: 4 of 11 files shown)


geoand:
By the way, I also want to do something similar to AiServices (and potentially others) in a follow up

langchain4j:
Hi @geoand thanks a lot, I will have time tomorrow to look into it

geoand:
Wonderful, thanks!

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolExecutionRequest.java
```diff
@@ -11,10 +11,6 @@
 
 public class ToolExecutionRequest {
 
-    private static final Gson GSON = new Gson();
```
langchain4j, line 14: I guess gson imports should be removed as well?

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolExecutionRequestUtil.java
```diff
@@ -0,0 +1,21 @@
+package dev.langchain4j.agent.tool;
+
+import com.google.gson.Gson;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Type;
+import java.util.Map;
+
+public class ToolExecutionRequestUtil {
```
langchain4j, line 8: How is it supposed to work in Quarkus? Can this class be somehow swapped during build time with another impl? I had an impression this should be done with SPI as well, or?

langchain4j-core/src/main/java/dev/langchain4j/model/input/PromptTemplate.java
```diff
@@ -27,18 +21,35 @@
  */
 public class PromptTemplate {
 
-    private static final MustacheFactory MUSTACHE_FACTORY = new DefaultMustacheFactory();
+    private static final PromptTemplateFactory FACTORY = factory();
+
+    private static PromptTemplateFactory factory() {
+        for (PromptTemplateFactory factory : ServiceHelper.loadFactories(PromptTemplateFactory.class)) {
+            return factory;
```
langchain4j, line 28: I assume this is done in a for loop to avoid an explicit check if the colleciton is empty?

langchain4j-core/src/main/java/dev/langchain4j/model/input/structured/DefaultStructuredPromptFactory.java
```diff
@@ -0,0 +1,51 @@
...
+            throw new IllegalArgumentException(
+                    String.format(
+                            "%s should be annotated with @StructuredPrompt to be used as a structured prompt",
+                            structuredPromptClassName
+                    )
+            );
+        }
+    }
+
+    private static Map<String, Object> extractVariables(Object structuredPrompt) {
```
langchain4j, line 46: It feels like only this method should have been extracted as an interface, to allow plugging other serialization libs instead of gson. Or do you expect that users would want to change the remaining functionality of this class?

langchain4j-core/src/main/java/dev/langchain4j/spi/ServiceHelper.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) 2011-2019 Contributors to the Eclipse Foundation
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+ * which is available at https://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
```
langchain4j: Is this comment required?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingResponseBuilder.java
```diff
@@ -12,19 +12,24 @@
 import dev.langchain4j.model.output.TokenUsage;
 
 import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static dev.langchain4j.model.openai.InternalOpenAiHelper.finishReasonFrom;
 
+/**
+ * This class needs to be thread safe because it is called when a streaming result comes back
+ * and there is no guarantee that this thread will be the same as the one that initiated the request
```
langchain4j: Handling thread will be almost definitely **not** the thread that initiated the request, but there should be no problem in this case, or?

langchain4j/src/main/java/dev/langchain4j/service/V.java
```diff
@@ -27,7 +27,8 @@
  * String chat(@V String name, @V int age);
  * </pre>
  * <p>
- * When using Spring Boot, defining the value of this annotation is not required.
+ * When the application code has been built with the Java compiler's {@code -parameters} flag enabled
```
langchain4j: Please read the comment above, it it already mentioned there. Spring Boot case was mentioned here additionally because of it's popularity (I doubt many users of Spring Boot know that `-parameters` is enabled implicitly).

langchain4j-core/src/main/java/dev/langchain4j/model/input/MustachePromptTemplateFactory.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.StringReader;
+import java.io.StringWriter;
+import java.util.Map;
+
+class MustachePromptTemplateFactory implements PromptTemplateFactory {
+
+    private static final MustacheFactory MUSTACHE_FACTORY = new DefaultMustacheFactory();
+
+    @Override
+    public MustacheTemplate create(PromptTemplateFactory.Input input) {
```
langchain4j, line 16: Could you please explain the purpose of interface `Input`? Why not just passing `String template` as an argument to the factory?

ghost (commented):
@geoand Thanks a lot! I have left some minor comments, would appreciate if you could address them.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolExecutionRequestUtil.java
geoand: Indeed! The reason I have not done anything about it is because I have not yet reached the point where I am ready to deal witth Tools.

langchain4j-core/src/main/java/dev/langchain4j/model/input/PromptTemplate.java
geoand: It's really a matter of style, we can go whichever way you prefer :)

langchain4j-core/src/main/java/dev/langchain4j/model/input/structured/DefaultStructuredPromptFactory.java
geoand: The functionality of this class should remain the same for other implementations, the difference is how the metadata is obtained.

In my Quarkus implementation for example, there is no need to do look up any annotations or do any validation because that functionality is provided at build time.

langchain4j-core/src/main/java/dev/langchain4j/spi/ServiceHelper.java
geoand: Good catch!

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingResponseBuilder.java
geoand: Exactly, because this class is called by a thread that handles SSE events, it needs to be thread safe.

langchain4j/src/main/java/dev/langchain4j/service/V.java
geoand: Cool, I'll remove this commit

langchain4j-core/src/main/java/dev/langchain4j/model/input/MustachePromptTemplateFactory.java
geoand: Sure!

The reason I have this is because there are actually two things required, the `template` and the `name`. 
Experience has shown us, that when there are multiple parameters, they have a way to multiple in time when more integrations come along. If that happens, having such an interface allows for seamless introduction of the new parameter, without having to either break existing SPI impls or add odd default methods

geoand:
@langchain4j I have force pushed changes to PR according to your review.

Thanks a lot for your time@

langchain4j-core/src/main/java/dev/langchain4j/model/input/PromptTemplate.java
langchain4j: @geoand no, it's fine, I just wanted to make sure I understand it right

langchain4j-core/src/main/java/dev/langchain4j/model/input/structured/DefaultStructuredPromptFactory.java
langchain4j: ah right, makes sense, thanks!

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingResponseBuilder.java
langchain4j: So far I've seen that okhttp handles SSE events using a single thread, and it makes total sense as the order of events is important. So it makes sense that they are consumed sequentially using the same thread. Of course other HTTP frameworks can do this differently, but I struggle to see a reason why would they handle a single stream of events using multiple threads, because it would require additional cumbersome synchronization to ensure the order of events preserved. I am not against this change (better safe than sorry), but I am trying to understand the reason.

geoand: Right, that makes perfect sense!

This comes into play in Quarkus and even Spring Boot if one handles the incoming request in a blocking thread pool, but makes HTTP requests in an async manner (think WebClient of Spring WebFlux) and then waits for the results on the blocking thread pool.

For a user, all that is hidden and there is no synchronization needed.

langchain4j: Not sure I understood your response. Do you mean that, when using reactive http clients, this class can receive SSE events from different threads?

geoand: It can certainly happen yes, because if the client uses an event loop for all IO, then it can very easily happen that the responses come back on a random threads

langchain4j: Ok, thanks for explanation!

ghost (approved):

geoand:
Thanks a lot for the review!

BTW, I also have a PR open for openai4j and I plan to open another here next week (hopefully)

langchain4j:
@geoand Thanks for your work!
Good that you mentioned openai4j, I somehow missed the notification there... I will try to check it asap

PR #248, opened 2023-10-25
Make AiServices swappable by external implementations

geoand:
# Purpose

The idea here is that advanced integrations would be able to supply their own version of `AiServices`. The use case I have [implemented](https://github.com/geoand/quarkus-langchain4j/blob/master/core/runtime/src/main/java/io/quarkiverse/langchain4j/QuarkusAiServicesFactory.java) using these SPIs allows Quarkus to do a lot of the work at build time instead of runtime, while also using our preferred set of dependencies.

# Changes

## Json

Introduce and SPI that allows integrations to swap out a custom Json implementation.

## ChatMessage Json SPI

Same idea as above but for the `ChatMessage` specifically

## InMemoryEmbeddingStore Json SPI

Same idea as above but for `InMemoryEmbeddingStore` specifically

## AiServices

This change builds on the previous two and is the actual meat of the change. It makes the least possible changes to the project that will allow for integrations to provide their own implementation of AiServices.
Existing usages of AiServices via langchain4j see no change whatsoever. 

### Note

The SPI is deliberately under specified as I prefer at this very early stage to place an additional burden on integrators (me in this case :)) where various parts need to be reimplemented, rather than trying to make the SPI to specific which could cause problems when and if new integrations come in the future. 
I believe that if interest for such an integration manifests, we can all work together to define the exact methods and types that will be needed by everyone thus allowing for maximum code reuse.


geoand:
@langchain4j I have rebased this PR onto the latest main in order to fix a merge conflict

geoand:
@langchain4j is there anything I can do to make this PR easier to review or explain my motivation better?

Thanks!

ghost (approved):
@geoand Thank you and sorry for the delay with review!

geoand:
No problem at all, thanks a ton for the review!

geoand:
@langchain4j do you have any idea when you will be doing a new release?

I completely understand if you don't know at this point, just wondering 😃

langchain4j:
@geoand did a release today 😃

geoand:
That's awesome, thanks so much!

PR #253, opened 2023-10-26
feat: in case a model like Vertex does not have a toolspecification do not use the generate method expecting one

### LINKED ISSUE #252: [FEATURE] in case a model like Vertex does not support a toolspecification, use the normal generate method
Currently `AiServicesline 406: Response<AiMessage> response = context.chatModel.generate(messages, context.toolSpecifications)` leads to issues for models without a toolspecification, whilst the normal generate method could be used.

SimonVerhoeven:
resolves #252 252


ghost (approved):
@SimonVerhoeven thanks!

PR #268, opened 2023-11-07
typo fix in VertexAI support

### LINKED ISSUE #259: [BUG] Typo in classname VextexAiLanguageModel.java
**Describe the bug**
This is not a bug per se but to make things consistent, could we change the classname of the following file: https://github.com/langchain4j/langchain4j/blob/main/langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VextexAiLanguageModel.java to VertexAiLanguageModel instead of VextexAiLanguageModel.

- LangChain4j version: e.g. 0.23.0

**Additional context**
The change if done will affect the Test case also, so that filename/classname will require a change too.

Heezer:


langchain4j:
🤦🤦🤦 
Thank you @Heezer !

ghost (approved):

PR #274, opened 2023-11-09
Make sure to add the SystemMessage as the first element of the messages.

jiangsier-xyz:
Make sure to add the SystemMessage as the first element of the messages list. (It should not be at last, see ensureCapacity(...))
And, when it comes to rolling messages, LinkedList offers superior performance.


langchain4j:
@jiangsier-xyz current behavior is intended: if a new system message is added, it should not be associated with exiting messages that were exchanged in the presence of another system message, thus it is added to the end. This is done to avoid confusion of LLM. Example:
```
1st system message: always answer in German
.... communication is german ...
2nd system message: always answer in English
```

If you put second system message instead of the first one and german communication will follow, this will be very confusing for the LLM

jiangsier-xyz:
> @jiangsier-xyz current behavior is intended: if a new system message is added, it should not be associated with exiting messages that were exchanged in the presence of another system message, thus it is added to the end. This is done to avoid confusion of LLM. Example:
> 
> ```
> 1st system message: always answer in German
> .... communication is german ...
> 2nd system message: always answer in English
> ```
> 
> If you put second system message instead of the first one and german communication will follow, this will be very confusing for the LLM

Got it. According to some model specifications (like Qwen), the system message must be the first element in the messages list. I looked into the OpenAI document, and although it is a typical usage, but not mandatory. So, for now, let's continue with the current behavior.

BTW, it might be useful to add an optional property "name" to both the SystemMessage and AIMessage. This property is supported by the [OpenAI API](https://platform.openai.com/docs/api-reference/chat/create).

PR #275, opened 2023-11-09
fix: add properties for s3 loader integration tests to fix failing tests when credentials are absent

jmgang:


ghost (approved):
@jmgang thank you!

Issue #206, opened 2023-10-03
[FEATURE] Use MemoryId in @Tool ?


AbdullahGheith:
How am i supposed to do this? this doesnt work

	@Tool("Saves phonenumber")
	void savePhonenumber(@MemoryId String sessionId, String phonenumber) {
		System.out.println(sessionId + " @ Saving phone number to db: "+phonenumber);
	}
	
My user is chatting through this:
	String chat(@MemoryId String memoryId, @UserMessage String userMessage);

chat.chat(userId, prompt.toUserMessage().text());

and i want to retreive that userid, how do i do it?

langchain4j:
@AbdullahGheith Good question, we do not support this yet, but probably should. Will take a look at this feature...

AbdullahGheith:
If multiple people are chatting with the bot, how am I supposed to differentiate between the users? Any suggestion until it gets fixed?

langchain4j:
In theory you could try to inform LLM about userId (e.g. in the system prompt) and make one of Tool parameters to be userId (annotated with `@P("userId")`), but there is no guarantee LLM will not hallucinate it. So use it on your own risk

lukasstanek:
We create an AiServices / Tools instance for each session and hold it inside of a hashmap to work around this currently. But I agree this feature would be very helpful!

We also tried the approach with adding the sessionid to the prompt and it generally worked, but we had the feeling that we don't want to burden the ai with unnecessary tasks.

benedictstrube:
Also highly interested in that feature! Could be implemented the same we way we pass a `@MemoryId` to the prompting interface methods, wdyt?

benedictstrube:
@langchain4j You can close this, I think :)

PR #276, opened 2023-11-09
Feat: Adding Memory Id to Tool Method Call

benedictstrube:
Implemented a new annotation `@ToolMemoryId` (naming debatable, but wanted to make it different to `@MemoryId`). If a tool method parameter is annotated as such, the `ToolExecutor` will pass the memory id annotated with `@MemoryId` from the prompt method interface to the tool method. That enables tools to securely access the identifier for individual users without relying on the used LLM service to pass the correct user id back as `@P` annotated tool parameter.


benedictstrube:
Implements #206

langchain4j:
@benedictstrube thank you for your contribution! I want to include your change in the new release today, so I had to apply your changes manually (since there were conflicts) to not delay the release. So I am closing this PR

PR #278, opened 2023-11-11
add bedrock bom

Martin7-1:
Hi! It seems like new `langchain4j-bedrock` module is not included in `langchain4j-bom`, so I add it.


ghost (approved):
@Martin7-1 thank you!

PR #280, opened 2023-11-13
Fix a token usage statistical issue in DefaultAiServices

jiangsier-xyz:
The statistical logic on token usage is consistent with AiServiceStreamingResponseHandler.

And more:
When it comes to rolling messages in MessageWindowChatMemory/TokenWindowChatMemory, LinkedList offers superior performance. ArrayList moves all the elements when the first element is deleted.


ghost (approved):
@jiangsier-xyz thank you for fixing this bug!

PR #282, opened 2023-11-13
Fixes #241: Added support for Neo4j Vector Index

### LINKED ISSUE #241: Adding Neo4j/Cypher support for vector search and knowledge graph integration

Knowledge Graphs are widely seen as an important part for GenAI applications.
So integrating them into orchestration frameworks is a great way of giving users this capability.

Neo4j has added integrations to langchain and llamaindex for vector search, GraphQAChain for query generation, RAG and RAG with sources, where people can use their existing graph as part of their applications or generate new graphs from unstructured data.

We would like to propose and also happy to implement Neo4j support in langchain4j similiar to how we did it in LangChain.

https://python.langchain.com/docs/integrations/providers/neo4j

* Vector Store
* GraphQAChain
* RetrieverChain
* GraphTransformer 

Not implementing the support, which would leave users without a valuable option. :)

**Additional context**

* [Neo4j GenAI](https://neo4j.com/generativeai)
* [Neo4j Vector Store](https://neo4j.com/docs/cypher-manual/current/indexes-for-vector-search/)
* [Docker+Neo4j GenAI Stack](https://github.com/docker/genai-stack)
* [LangChain Docs](https://python.langchain.com/docs/integrations/providers/neo4j)
* [LlamaIndex Docs](https://gpt-index.readthedocs.io/en/latest/examples/index_structs/knowledge_graph/Neo4jKGIndexDemo.html)

Happy to answer any questions in comments.

vga91:
Fixes #241: Added support for Neo4j Vector Index

This commit brings support for Neo4j graph database in general, and uses the vector index functionality, generally available since version 5.13.

Mostly aligned with the existing WeaviateEmbeddingStoreImpl implementation and tests.
The tests have some additional Neo4j node assertion to check that the nodes involved are correctly created.

The module creates indexes, i.e. `"CALL db.index.vector.createNodeIndex(<indexName>, <label>, <embeddingProperty>, <dimension>, <distanceType>)"`, if needed, for the vector search .

The required configurations are:
- the Neo4j index dimension parameter
- the Neo4j Java Driver connection instance
- as an alternative to the Neo4j Java Driver, we can create a `Neo4jEmbeddingStore.builder().withBasicAuth(<url>, <username>, <password>)`, which will create a Driver connection instance under the hood

It is possible to customize, via the builder:
- the index name (with default `langchain-embedding-index`)
- the Neo4j node label  (with default `Document`)
- the Neo4j property key which save the embeddings  (with default `embeddingProp`)
- the Neo4j index distanceType parameter
- the metadata prefix (with default `metadata.`)
- the text property key (with default `text`), which store the text field of the `TextSegment.java`

Created an example PR as well, on `langchain4j-examples` repo: https://github.com/langchain4j/langchain4j-examples/pull/23


langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Map;
+
+public class Neo4jEmbeddingUtils {
+    
+    /* not-configurable strings, used in `UNWIND $rows ...` statement */
+    public static final String EMBEDDINGS_ROW_KEY = "embeddingRow";
+    public static final String ID_ROW_KEY = "id";
+    
+    /* default configs */
+    public static final String DEFAULT_EMBEDDING_PROP = "embeddingProp";
```
jexp: change to `embedding`

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,82 @@
...
+
+        Map<String, Object> row = new HashMap<>();
+        row.put(ID_ROW_KEY, id);
+
+        Map<String, Object> properties = new HashMap<>();
+        if (embedded != null) {
+            TextSegment segment = embedded.get(idx);
+            properties.put(store.getText(), segment.text());
+            Map<String, String> metadata = segment.metadata().asMap();
+            metadata.forEach((k, v) -> properties.put(store.getMetadataPrefix() + k, Values.value(v)));
```
jexp, line 74: can we have the default option to be the metadata-keys without a prefix, and only if a prefix is configured it is used?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        row.put(EMBEDDINGS_ROW_KEY, Values.value(embedding.vector()));
```
jexp, line 77: should we use a float[] here, need to check with the drivers team.

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
vga91: Changed

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
vga91: Now the default prefix is `""`

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+    private final String embeddingProperty;
+    private final String label;
+    private final String text;
+
+    /**
+     * Creates an instance of Neo4jEmbeddingStore defining a {@link Driver} 
+     * starting from uri, user and password
+     */
+    public static class Neo4jEmbeddingStoreBuilder {
+        public Neo4jEmbeddingStoreBuilder withBasicAuth(String uri, String user, String password) {
```
jexp, line 68: Should also allow for database name

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+     * @param label: the optional label name (default: "Document")
+     * @param property: the optional property name (default: "embedding")
+     * @param distanceType: the optional distanceType (default: "cosine")
+     * @param metadataPrefix: the optional metadata prefix (default: "")
+     * @param text: the optional text property name (default: "text")
+     * @param indexName: the optional index name (default: "langchain-embedding-index")
+     */
+    @Builder
+    public Neo4jEmbeddingStore(
+            SessionConfig config,
```
jexp, line 91: Good question if we want to leak the session config from neo4j into the client code. I would probably start with database name and then add a builder methods for other options.

Should the builder constructor be public?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+            String label,
+            String property,
+            Neo4jDistanceType distanceType,
+            String metadataPrefix,
+            String text,
+            String indexName) {
+        
+        /* required configs */
+        ensureNotNull(driver, "driver");
+        ensureNotNull(dimension, "dimension");
```
jexp: should also check that dimension is > 0 and  <= 4096 (currently Neo4j only supports sizes to 2048) but will increase that.

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+
+        this.driver = driver;
+        this.dimension = dimension;
+        
+        /* optional configs */
+        this.config = getOrDefault(config, SessionConfig.forDatabase("neo4j"));
```
jexp: make database configurable

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+    public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedded) {
+        List<String> ids = embeddings.stream()
+                .map(ignored -> randomUUID())
+                .toList();
+        addAllInternal(ids, embeddings, embedded);
+        return ids;
+    }
+
+    @Override
+    public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbedding, int maxResults, double minScore) {
```
jexp, line 184: what is the default for minScore?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+    private void bulk(List<String> ids, List<Embedding> embeddings, List<TextSegment> embedded) {
+
+        List<Map<String, Object>> rows = IntStream.range(0, ids.size())
+                .mapToObj(idx -> toRecord(this, idx, ids, embeddings, embedded))
+                .toList();
+
+        try (var session = session()) {
+            String statement = """
+                            UNWIND $rows AS row
+                            MERGE (u:%1$s {%2$s: row.%2$s})
```
jexp: add backticks around label and property name, and row values.
and sanitize label and property names to avoid injection attacks.

jexp: if we merge on the id - property we should make sure that a constraint exists for that id property and create if it doesn't.

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+                            SET u += row.%3$s
+                            WITH row, u
+                            CALL db.create.setNodeVectorProperty(u, $embeddingProperty, row.%4$s)
+                            RETURN *""".formatted(
```
jexp: we should return count(*) to avoid huge results coming back (think of all the input data and all the nodes with their properties coming back).

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+                    ID_ROW_KEY,
+                    PROPS,
+                    EMBEDDINGS_ROW_KEY);
+
+            Map<String, Object> params = Map.of(
+                    "rows", rows,
+                    "embeddingProperty", this.embeddingProperty
+            );
+            
+            session.run(statement, params).consume();
```
jexp: we also need to batch this from the outside, so that we limit each tx to 10k rows or so. and use 
tx.executeWrite() for each batch.

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+    private void createIndexIfNotExist() {
+        if (!indexExists()) {
+            createIndex();
+        }
+    }
+
+    private boolean indexExists() {
+        try (var session = session()) {
+            Map<String, Object> params = Map.of("name", this.indexName);
+            return session.run("SHOW INDEX WHERE type = 'VECTOR' AND name = $name", params)
```
jexp: also verify that it has the same label + embedding property

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+                "embeddingProperty", this.embeddingProperty,
+                "dimension", this.dimension,
+                "distanceType", this.distanceType.getValue());
+
+        // create vector index
+        try (var session = session()) {
+            session.run("CALL db.index.vector.createNodeIndex($indexName, $label, $embeddingProperty, $dimension, $distanceType)",
+                    params);
+
+            session.run("CALL db.awaitIndexes()").consume();
```
jexp: the default is 300s (5min) here, not sure if we want to pass in an explicit value that is large enough for the index population to finish in case there is already data in the db.

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+import org.neo4j.driver.SessionConfig;
+
+import static dev.langchain4j.store.embedding.neo4j.Neo4jEmbeddingUtils.*;
+
+/**
+ * Represents a Vector index as an embedding store.
+ * Annotated with `@Getter` to be used in {@link Neo4jEmbeddingUtils}
+ */
+@Getter
+public class Neo4jEmbeddingStore implements EmbeddingStore<TextSegment> {
```
jexp, line 37: is there any API in langchain4j to allow existing data in the database to be embedded and indexed?
what does the whole roundtrip look like in an example?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+        try (var session = session()) {
+            String statement = """
+                            UNWIND $rows AS row
+                            MERGE (u:%1$s {%2$s: row.%2$s})
+                            SET u += row.%3$s
+                            WITH row, u
+                            CALL db.create.setNodeVectorProperty(u, $embeddingProperty, row.%4$s)
+                            RETURN *""".formatted(
+                    this.label,
+                    ID_ROW_KEY,
```
jexp: should be configurable

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+        /* required configs */
+        ensureNotNull(driver, "driver");
+        ensureNotNull(dimension, "dimension");
+
+        this.driver = driver;
+        this.dimension = dimension;
+        
+        /* optional configs */
+        this.config = getOrDefault(config, SessionConfig.forDatabase("neo4j"));
+        this.label = getOrDefault(label, DEFAULT_LABEL);
```
jexp, line 111: id should also be configurable

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+        this.embeddingProperty = getOrDefault(property, DEFAULT_EMBEDDING_PROP);
```
jexp: I'm missing the builder methods for all these other config options? are they somewhere else?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public class Neo4jEmbeddingUtils {
+    
+    /* not-configurable strings, used in `UNWIND $rows ...` statement */
+    public static final String EMBEDDINGS_ROW_KEY = "embeddingRow";
+    public static final String ID_ROW_KEY = "id";
```
jexp: should be also configurable

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+    
+    /* default configs */
+    public static final String DEFAULT_EMBEDDING_PROP = "embedding";
+    public static final String PROPS = "props";
+    public static final String DEFAULT_IDX_NAME = "langchain-embedding-index";
```
jexp: probably use a similar namings like we do in the langchain integration?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+    public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbedding, int maxResults, double minScore) {
+        var embeddingValue = Values.value(referenceEmbedding.vector());
+        
+        try (var session = session()) {
+            return session
+                    .run("""
+						CALL db.index.vector.queryNodes($indexName, $maxResults, $embeddingValue)
+						YIELD node, score
+						WHERE score >= $minScore
+						RETURN node, score
```
jexp: can we make configurable which properties we want to return or the full node?
because if we return all the huge properties like the embedding + big texts that can become a lot

we should also add the option like in langchain integration to provide a retrievalQuery in the config that is applied after the node/score is retrieved from the index so that the user can optionally find more contextually relevant data in the neighbourhood.

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+    public static final String DEFAULT_LABEL = "Document";
+    public static final String DEFAULT_TEXT_PROP = "text";
+    
+
+    public static EmbeddingMatch<TextSegment> toEmbeddingMatch(Neo4jEmbeddingStore store, Record neo4jRecord) {
+        var node = neo4jRecord.get("node").asNode();
```
jexp: we should make configurable what we actually want to return from the database and then also from the find method, otherwise it might be a lot of data that is returned.

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,83 @@
...
+
+        var metaData = new HashMap<String, String>();
+        node.keys().forEach(key -> {
+            Set<String> notMetaKeys = Set.of(ID_ROW_KEY, store.getEmbeddingProperty(), store.getText());
+            if (!notMetaKeys.contains(key)) {
+                metaData.put(key.replace(store.getMetadataPrefix(), ""), node.get(key).asString());
```
jexp: asString or asObject?

langchain4j-neo4j/src/test/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStoreTest.java
```diff
@@ -0,0 +1,371 @@
...
+    void initEmptyNeo4jEmbeddingStore() {
+        embeddingStore = Neo4jEmbeddingStore.builder()
+                .withBasicAuth(neo4jContainer.getBoltUrl(), USERNAME, ADMIN_PASSWORD)
+                .dimension(384)
+                .build();
+    }
+
+    @AfterEach
+    void afterEach() {
+        session.run("MATCH (n) DETACH DELETE n");
```
jexp, line 83: also drop the index?

langchain4j-neo4j/src/test/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStoreTest.java
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    @Test
+    void should_add_embedding() {
+        Embedding embedding = embeddingModel.embed(randomUUID()).content();
```
jexp, line 80: extract the id and compare with the id that is returned

langchain4j-neo4j/src/test/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStoreTest.java
```diff
@@ -0,0 +1,371 @@
...
+
+        checkEntitiesCreated(relevant.size(), 
+                iterator -> checkDefaultProps(embedding, match, iterator.next()));
+    }
+
+    @Test
+    void should_add_embedding_with_id() {
+
+        String id = randomUUID();
+        Embedding embedding = embeddingModel.embed(randomUUID()).content();
```
jexp: different id ??

jexp: oh this is meant to be a text?

langchain4j-neo4j/src/test/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStoreTest.java
jexp: perhaps use something else as text ? using randomUUID() in two places is confusing :)

jexp (changes requested):
Looks good, thanks so much @vga91 
Please have a look at my comments.

The configuration of what is returned / retrievalQuery and some other extra things can wait for a later PR.

Id configuration (probably the id-constraint), backticks etc. would be good to have in the first PR for security/performance reasons.

I didn't see any docs or an end-to-end example. Not sure how that works in langchain4j.

langchain4j-neo4j/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+    <artifactId>langchain4j-neo4j</artifactId>
+    
+    <packaging>jar</packaging>
+    <name>LangChain4j integration with Neo4j Vector Index</name>
+
+    <description>Requires Java 17</description>
+
+    <properties>
+        <testcontainers.version>1.19.0</testcontainers.version>
+        <maven.compiler.source>17</maven.compiler.source>
```
langchain4j, line 23: would be extremely nice to support java 8, seems that it is supported by neo4j: https://neo4j.com/docs/operations-manual/current/installation/requirements/

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
langchain4j: Not yet, most of the existing integrations are with "native" vector stores, so there was no need to embed and index existing data.

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,244 @@
...
+            Neo4jDistanceType distanceType,
+            String metadataPrefix,
+            String text,
+            String indexName) {
+        
+        /* required configs */
+        ensureNotNull(driver, "driver");
+        ensureNotNull(dimension, "dimension");
+
+        this.driver = driver;
```
langchain4j: minor: it is supposed to be used like this: `this.driver = ensureNotNull(driver, "driver");`

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
langchain4j: the default is 0, meaning that it will not filter by score

langchain4j:
Hi @vga91, thanks a gain for your contribution! There are comments from @jexp, could you please have a look?

langchain4j-neo4j/pom.xml
jexp: Not for the current versions that have vector support, those are on 17.
Also this uses the client drivers, so it's independent of the server installation.

jexp: Which is at 17 too
https://neo4j.com/docs/java-manual/current/get-started/#java-driver-get-started-installation

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: Added databaseName config as well

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: Added databaseName config as well

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: I added `sanitize()` to labels and properties.
I didn't add backticks here, since together with `sanitize`, it could create double-backticks

vga91: Added a [`CREATE CONSTRAINT` statement](https://github.com/langchain4j/langchain4j/pull/282/commits/36cd2f46c1280e53068bd2b9cd26941e17dfb0ac#diff-89b713731c6e14e29b6208dc8c14a92aff1a1baad94b44b8a9b388db8be97eb9R249)

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: Changed to 60

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: Added

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: The getter and builder method are auto-generate via[Lombok annotations.](https://github.com/langchain4j/langchain4j/pull/282/commits/36cd2f46c1280e53068bd2b9cd26941e17dfb0ac#diff-89b713731c6e14e29b6208dc8c14a92aff1a1baad94b44b8a9b388db8be97eb9R86)

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: Added a retrievalQuery config, with default:
```
RETURN properties(node) AS metadata, node.id AS id, node.text AS text, node.embedding AS embedding, score
```

langchain4j-neo4j/src/test/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStoreTest.java
vga91: Added

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: Changed

vga91:
Sorry for the delay. 
I just pushed the following changes:
- databaseName config
- `"id"` property config
- added auto-constraint for the property `"id"`
- dimension: check that is > 0 and <= 4096 
- added sanitize to label and property name
- added batch every 10,000 rows
- added check to `SHOW INDEX WHERE type = 'VECTOR' ..`
- changed `CALL db.awaitIndexes()` to `CALL db.awaitIndexes(60)`
- added `retrievalQuery` to customize what is returned in `CALL db.index.vector.queryNodes`

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,310 @@
...
+            return;
+        }
+        ensureTrue(ids.size() == embeddings.size(), "ids size is not equal to embeddings size");
+        ensureTrue(embedded == null || embeddings.size() == embedded.size(), "embeddings size is not equal to embedded size");
+
+        bulk(ids, embeddings, embedded);
+    }
+
+    private void bulk(List<String> ids, List<Embedding> embeddings, List<TextSegment> embedded) {
+        Collection<List<Map<String, Object>>> rowsBatched = getRowsBatched(this, ids, embeddings, embedded);
```
jexp: Would it be better to return a stream here instead of a collection? So that we stream one chunk at a time? And don't allocate all the memory upfront?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,113 @@
...
+            Map<String, String> metadata = segment.metadata().asMap();
+            metadata.forEach((k, v) -> properties.put(store.getMetadataPrefix() + k, Values.value(v)));
+        }
+
+        row.put(EMBEDDINGS_ROW_KEY, Values.value(embedding.vector()));
+        row.put(PROPS, properties);
+        return row;
+    }
+
+    public static Collection<List<Map<String, Object>>> getRowsBatched(Neo4jEmbeddingStore store, List<String> ids, List<Embedding> embeddings, List<TextSegment> embedded) {
```
jexp: perhaps change to return a stream of chunks, where don't use a grouping collector but a IntStream of startIndex*batchSize + batchSize to create a chunk at a time.

jexp (approved):
Looks good to me.

Only 2 small things

- can you change the chunking to streaming
- add some unit tests for Neo4jEmbeddingUtils (for the chunking esp. off by sizes) and the sanitation

jexp (approved):
Looks good, thanks so much @vga91 
Good to go for my @langchain4j

jexp:
@langchain4j not sure how to resolve all the conversations, there is no "resolve all" and the individual ones don't show "Resolve" Menu/Button for me.

jexp (approved):
Good to go for me @langchain4j (but the "review required" doesn't go away)

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jDistanceType.java
```diff
@@ -0,0 +1,16 @@
...
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
+public enum Neo4jDistanceType {
+    /* cosine similarity */
+    COSINE("cosine"),
+    
+    /* euclidean distance */
+    EUCLIDEAN("euclidean");
```
langchain4j: I would remove euclidean as it is not quite useful for embeddings and we assume the use of cosine everywhere

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,311 @@
...
+    public Neo4jEmbeddingStore(
+            SessionConfig config,
+            Driver driver,
+            int dimension,
+            String label,
+            String embeddingProperty,
+            String idProperty,
+            Neo4jDistanceType distanceType,
+            String metadataPrefix,
+            String text,
```
langchain4j: this probably should be `textProperty`?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,311 @@
...
+        ensureTrue(embedded == null || embeddings.size() == embedded.size(), "embeddings size is not equal to embedded size");
+
+        bulk(ids, embeddings, embedded);
+    }
+
+    private void bulk(List<String> ids, List<Embedding> embeddings, List<TextSegment> embedded) {
+        Stream<List<Map<String, Object>>> rowsBatched = getRowsBatched(this, ids, embeddings, embedded);
+
+        rowsBatched.forEach(rows -> {
+            try (var session = session()) {
```
langchain4j: should session be created for each row separately?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,311 @@
...
+                        this.sanitizedIdProperty,
+                        PROPS,
+                        EMBEDDINGS_ROW_KEY);
+                
+                Map<String, Object> params = Map.of(
+                        "rows", rows,
+                        "embeddingProperty", this.embeddingProperty
+                );
+
+                session.executeWrite(tx -> tx.run(statement, params).consume());
```
langchain4j: was it intended to write each embedding in a separate transaction?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -0,0 +1,311 @@
...
+                "embeddingProperty", this.embeddingProperty,
+                "dimension", this.dimension,
+                "distanceType", this.distanceType.getValue());
+
+        // create vector index
+        try (var session = session()) {
+            session.run("CALL db.index.vector.createNodeIndex($indexName, $label, $embeddingProperty, $dimension, $distanceType)",
+                    params);
+
+            session.run("CALL db.awaitIndexes(60)").consume();
```
langchain4j: will 60 be enough for all use cases or should this be configurable?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+import static org.neo4j.cypherdsl.support.schema_name.SchemaNames.sanitize;
+
+public class Neo4jEmbeddingUtils {
```
langchain4j: please make this class package-private

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,120 @@
...
+    public static final String DEFAULT_EMBEDDING_PROP = "embedding";
+    public static final String PROPS = "props";
+    public static final String DEFAULT_IDX_NAME = "vector";
+    public static final String DEFAULT_LABEL = "Document";
+    public static final String DEFAULT_TEXT_PROP = "text";
+
+    public static EmbeddingMatch<TextSegment> toEmbeddingMatch(Neo4jEmbeddingStore store, Record neo4jRecord) {
+        Map<String, String> metaData = new HashMap<>();
+        neo4jRecord.get("metadata").asMap().forEach((key, value) -> {
+            Set<String> notMetaKeys = Arrays.asList(store.getIdProperty(), store.getEmbeddingProperty(), store.getText())
```
langchain4j: It is quite wasteful to create a `List` and a `Set` for each metadata entry of each embedding match. Can you create this `Set` in ctor of `Neo4jEmbeddingStore` during init?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,120 @@
...
+        });
+
+        Metadata metadata = new Metadata(metaData);
+
+        Value text = neo4jRecord.get(store.getText());
+        TextSegment textSegment = text.isNull()
+                ? null
+                : TextSegment.from(text.asString(), metadata);
+        List<Number> embeddingList = neo4jRecord.get(store.getEmbeddingProperty())
+                .asList(Value::asNumber);
```
langchain4j: WIll `.asList(Values.ofFloat())` work? Then you will be able to create `Embedding.from(List<Float>)` directly without a need for `toFloatArray()`

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -0,0 +1,120 @@
...
+                );
+    }
+
+    public static String sanitizeOrThrows(String value, String config) {
+        return sanitize(value)
+                .orElseThrow(() -> {
+                    String invalidSanitizeValue = String.format("The value %s, to assign to configuration %s, cannot be safely quoted",
+                            value,
+                            config);
+                    throw new RuntimeException(invalidSanitizeValue);
```
langchain4j: should be `return new ...` instead of `throw new ...`

ghost (changes requested):
@vga91 thanks a lot and sorry for the delayed review. Could you please check my comments?

langchain4j:
@vga91 also the build fails, might be due to lombok

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jDistanceType.java
vga91: Do you mean, removing the Neo4jDistanceType.java class at all and expliciting "cosine" everywhere, 
or just removing `EUCLIDEAN("euclidean");` from here?

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: Yes actually it's better, I'll change it

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: Yes the intention is to execute a transaction for each batch

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
vga91: changed

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingStore.java
vga91: 60 is a quite high value, but indeed a configuration value is better.
Added an `awaitIndexTimeout` config

vga91:
@langchain4j 

I've made the various changes, except the EUCLIDEAN stuff I asked about.

I'm not sure why the build fails, executing a `mvn --batch-mode install -DskipITs` locally, with temurin `21.0.1-tem` (which should be the same version) it works, and the lombok version should be compatible with JDK 21..

It's probably due to [the version](https://github.com/langchain4j/langchain4j/pull/282/files#diff-a88727153e9e65be269b619a6a6351f881a574c0aec035bcbcb90d5c97fb2e14R10), which was 0.23 and now I changed it to 0.24?

Anyway from what I see, the CI doesn't start unless all the conversations are resolved, so I guess we could check it out after doing that

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jDistanceType.java
langchain4j: removing Neo4jDistanceType completely

vga91: removed

langchain4j:
@vga91 I think you need to add an excludion for neo4j module in `.github/workflows/main.yaml` for java 8 and 11:
```
          - java_version: '8'
            included_modules: '-pl !langchain4j-graal,!langchain4j-neo4j,!langchain4j-opensearch'
          - java_version: '11'
            included_modules: '-pl !langchain4j-graal,!langchain4j-neo4j'
```

ghost (approved):
@vga91 good job, thank you a lot!

langchain4j:
@vga91 there seems to be [test failures](https://github.com/langchain4j/langchain4j/actions/runs/7259989386/job/19778223144) in neo4j module, could you please check?

vga91:
@langchain4j I created a PR that should solve the problem: https://github.com/langchain4j/langchain4j/pull/368

langchain4j:
@vga91 thank you!

langchain4j:
Hi @vga91, it seems that there are still [problems](https://github.com/langchain4j/langchain4j/actions/runs/7337911358/job/19979565733?pr=381) with Neo4j tests, could you please check? Thanks!

langchain4j:
@vga91 same problem here: https://github.com/langchain4j/langchain4j/actions/runs/7396118145/job/20120618008?pr=396
I would appreciate a lot if you could fix this, this blocks our CI. Thank you!

vga91:
@langchain4j I created another PR for it https://github.com/langchain4j/langchain4j/pull/441

PR #289, opened 2023-11-17
Adding support for VertexAi matching engine.

pascalconfluent:


langchain4j-vertex-ai/pom.xml
```diff
@@ -17,12 +17,30 @@
...
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>32.1.3-jre</version>
+        </dependency>
+
+        <!-- https://mvnrepository.com/artifact/org.apache.commons/commons-lang3 -->
+        <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-lang3</artifactId>
```
langchain4j: please avoid adding extra dependencies when not absolutely necessary

langchain4j-vertex-ai/pom.xml
```diff
@@ -17,12 +17,30 @@
     <name>LangChain4j integration with Google Vertex AI</name>
 
     <dependencies>
+        <!-- https://mvnrepository.com/artifact/com.google.guava/guava -->
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
```
langchain4j: please avoid adding extra dependencies when not absolutely necessary

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
```diff
@@ -0,0 +1,295 @@
...
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+import static dev.langchain4j.internal.ValidationUtils.ensureTrue;
+import static java.util.Collections.singletonList;
+
+/**
+ * A matching engine that uses Vertex AI to store and match embeddings.
+ */
+@SuperBuilder
+@Slf4j
+public class VertexAiMatchingEngine implements EmbeddingStore<TextSegment> {
```
langchain4j: Seems that they are renaming it to "Vector Search" now, let's maybe name it VertexAiEmbeddingStore to be in line with other integrations?

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    protected EmbeddingModel embeddingModel() {
+        return embeddingModel;
+    }
+
+    @Override
+    protected double getCosineSimilarity(Embedding embedding, Embedding referenceEmbedding) {
+        return CosineSimilarity.between(embedding, referenceEmbedding);
```
langchain4j: It would be better to keep the behavior of all embedding stores the same. Score should be a number from 0 to 1.

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
```diff
@@ -0,0 +1,61 @@
...
+                .bucketName("[BUCKET_NAME]")
+                .indexId("[INDEX_ID]")
+                .deployedIndexId("[INDEX_DEPLOYED_ID]")
+                .indexEndpointId("[INDEX_ENDPOINT_ID]")
+                .location("us-central1")
+                .project("[PROJECT]")
+                .build();
+
+
+        final List<String> indices = matchingEngine.allIndices();
```
langchain4j: You should probably move this logic to `clearStore`

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
```diff
@@ -0,0 +1,61 @@
...
+
+@Disabled("To run this test, you must have provide your own endpoint, project and location")
+public class VertexAiMatchingEngineIT extends EmbeddingStoreIT {
+
+    private VertexAiMatchingEngine matchingEngine;
+    private EmbeddingModel embeddingModel;
+
+    @BeforeEach
+    void beforeEach() {
+        embeddingModel = VertexAiEmbeddingModel.builder()
```
langchain4j: Why not simply using `AllMiniLmL6V2QuantizedEmbeddingModel`?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
```diff
@@ -0,0 +1,290 @@
+package dev.langchain4j.model.vertexai;
```
langchain4j: please move this class to dev.langchain4j.store.embedding.vertexai

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
```diff
@@ -0,0 +1,290 @@
...
+
+import static dev.langchain4j.internal.Utils.*;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+import static dev.langchain4j.internal.ValidationUtils.ensureTrue;
+import static java.util.Collections.singletonList;
+
+/**
+ * A matching engine that uses Vertex AI to store and match embeddings.
+ */
+@SuperBuilder
```
langchain4j: should it be @Builder?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAiEmbeddingIndex.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+    
+    /**
+     * Adds an embedding to the index.
+     *
+     * @param id        the id of the embedding
+     * @param embedding the embedding
+     */
+    public void addEmbedding(String id, Embedding embedding) {
+        ensureNotBlank(id, "id cannot be blank");
```
langchain4j: no need to write "cannot be blank", "cannot be null", etc

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAiEmbeddingIndex.java
```diff
@@ -0,0 +1,68 @@
...
+
+        records.add(new VertexAiEmbeddingIndexRecord(id, embedding.vectorAsList()));
+
+        if (embedding.dimension() > dimensions) {
+            dimensions = embedding.dimension();
+        }
+    }
+
+    @Override
+    public String toString() {
```
langchain4j: please rename into toJson

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAiEmbeddingIndex.java
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Represents an embedding index.
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+public class VertexAiEmbeddingIndex {
+    private final List<VertexAiEmbeddingIndexRecord> records;
+    private int dimensions;
```
langchain4j: I am not sure I understand the need for this property, could you please explain?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAIDocument.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+/**
+ * This class is used to serialize the document to JSON.
+ */
+@Getter
+@NoArgsConstructor
+@AllArgsConstructor
+public class VertexAIDocument {
```
langchain4j: please rename into VertexAiDocument

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
```diff
@@ -0,0 +1,61 @@
...
+                .publisher("google")
+                .modelName("textembedding-gecko@001")
+                .maxRetries(3)
+                .build();
+
+        matchingEngine = VertexAiMatchingEngine
+                .builder()
+                .endpoint("us-central1-aiplatform.googleapis.com:443")
+                .bucketName("[BUCKET_NAME]")
+                .indexId("[INDEX_ID]")
```
langchain4j: what is the difference between indexId, deployedIndexId and indexEndpointId?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
```diff
@@ -0,0 +1,111 @@
...
+            final MatchServiceSettings.Builder settings = MatchServiceSettings
+                    .newBuilder()
+                    .setEndpoint(endpoint);
+            if (credentialsProvider != null) {
+                settings.setCredentialsProvider(credentialsProvider);
+            }
+
+            return settings.build();
+        } catch (Exception e) {
+            log.error("Error initializing Vertex AI MatchServiceSettings", e);
```
langchain4j: no need to log AND rethrow an exception, please check in all places

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
```diff
@@ -0,0 +1,290 @@
...
+     * @return the index endpoint
+     */
+    private IndexEndpointService initIndexEndpoint() {
+        ensureNotNull(endpoint, "endpoint is null");
+        ensureNotNull(location, "location is null");
+        ensureNotNull(project, "project is null");
+
+        return IndexEndpointService
+                .builder()
+                .endpoint(endpoint)
```
langchain4j: is is supposed to be used like this: `.endpoint(ensureNotNull(endpoint, "endpoint"))...`

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
```diff
@@ -0,0 +1,111 @@
...
+                .addQueries(query)
+                .setReturnFullDatapoint(returnFullDatapoint)
+                .build();
+
+        try (MatchServiceClient client = MatchServiceClient.create(getSettings())) {
+            final FindNeighborsResponse response = client.findNeighbors(request);
+            return response.getNearestNeighborsList()
+                    .stream()
+                    .flatMap(neighbor -> neighbor.getNeighborsList().stream())
+                    .filter(neighbor -> neighbor.getDistance() >= minScore)
```
langchain4j: This is not correct, you should convert distance into score first.
Distance (0..2): the smaller - the better match
Score (0..1): the bigger - the better match

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAIDocument.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final Gson GSON = new Gson();
+
+    private String memberOfIndex;
```
langchain4j: is it needed for something?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
```diff
@@ -0,0 +1,290 @@
...
+
+            // Upload the document to GCS
+            getGcpBlobService().upload("documents/" + id, new VertexAIDocument(indexFileId, textSegment));
+        }
+
+        final String filename = "indexes/" + indexFileId + ".json";
+        log.info("Uploading {} index to GCS.", filename);
+
+        // Upload the index to GCS
+        getGcpBlobService().upload(filename, index.toString());
```
langchain4j: is it really required to upload the index (embeddings) both into a bucket AND index endpoint?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
```diff
@@ -0,0 +1,290 @@
...
+    private final String indexEndpointId;
+    private final String deployedIndexId;
+    private final String indexId;
+    private final CredentialsProvider credentialsProvider;
+    @Getter
+    @Builder.Default
+    private final boolean avoidDups = true;
+    @Getter
+    @Builder.Default
+    private final boolean returnFullDatapoint = true;
```
langchain4j: should we maybe set it to false by default? it is rarely that embeddings itself are needed, but it adds to the latency

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
```diff
@@ -0,0 +1,111 @@
...
+                    .map(neighbor -> {
+                        final String id = neighbor.getDatapoint().getDatapointId();
+                        final String path = "documents/" + id;
+                        final String content = gcpBlobService.download(path);
+                        final VertexAIDocument document = VertexAIDocument.fromJson(content);
+                        final IndexDatapoint resultDatapoint = neighbor.getDatapoint();
+                        return new EmbeddingMatch<>(neighbor.getDistance(),
+                                id,
+                                Embedding.from(resultDatapoint.getFeatureVectorList()),
```
langchain4j: since you made returnFullDatapoint configurable, the check is needed here that the embedding is returned

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
```diff
@@ -0,0 +1,111 @@
...
+                    .stream()
+                    .flatMap(neighbor -> neighbor.getNeighborsList().stream())
+                    .filter(neighbor -> neighbor.getDistance() >= minScore)
+                    .map(neighbor -> {
+                        final String id = neighbor.getDatapoint().getDatapointId();
+                        final String path = "documents/" + id;
+                        final String content = gcpBlobService.download(path);
+                        final VertexAIDocument document = VertexAIDocument.fromJson(content);
+                        final IndexDatapoint resultDatapoint = neighbor.getDatapoint();
+                        return new EmbeddingMatch<>(neighbor.getDistance(),
```
langchain4j: it should not be distance, but score

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/IndexEndpointService.java
```diff
@@ -0,0 +1,152 @@
...
+    private final String publicEndpoint = initPublicEndpoint();
+
+    /**
+     * Deletes the embedding index.
+     *
+     * @param embeddingIndices the embedding indices
+     */
+    public void deleteIndices(List<String> embeddingIndices) {
+        if (StringUtils.isEmpty(indexId)) {
+            log.warn("Index is not specified, skipping delete.");
```
langchain4j: this should fail in my opinion, you are using this in tests only

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAIDocument.java
```diff
@@ -0,0 +1,32 @@
...
+    private TextSegment segment;
```
langchain4j: I would probably instead define `String text` and `Map<String, String> metadata` as fields in this class instead of `TextSegment` and map it manually because the structure of `TextSegment` might change at some point

ghost (commented):
Hi @pascalconfluent thanks a lot for your contribution!
I have left some comments, please check them, thanks!

langchain4j-vertex-ai/pom.xml
pascalconfluent: Fixing it, replacing with isNullOrBlank from Utils class

langchain4j-vertex-ai/pom.xml
pascalconfluent: Removed this dependency

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
pascalconfluent: Yes, I was about to change the name :)

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
pascalconfluent: In VertexAI is depends on the algorithm you are using when creating the index. The one use default in the unit tests is not recommended by Google. But happy to revert back and put a note in the unit tests.

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
pascalconfluent: Done!

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
pascalconfluent: Sure

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
pascalconfluent: Done!

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
pascalconfluent: Done!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
pascalconfluent: Yeap!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAiEmbeddingIndex.java
pascalconfluent: K

langchain4j:
BTW is this a recommended way to store and search embeddings using Vertex AI? Somehow feels overly complicated and in some places inefficient.

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
langchain4j: which algorithms are supported and wich is the default one? can't find it in the code...

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAiEmbeddingIndex.java
pascalconfluent: Done!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAiEmbeddingIndex.java
pascalconfluent: VertexAiEmbeddingIndex?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAIDocument.java
pascalconfluent: Done

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
pascalconfluent: That is how VertexAI Vector Search works. A Index has to be deployed and has an endpoint.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
pascalconfluent: Done.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
pascalconfluent: Coding style, not worries if you like it better this way

pascalconfluent: Done.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
pascalconfluent: For the documentation yes and looked at the LangChain python version of and they also do the same.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAIDocument.java
pascalconfluent: Yeap. This is what is saved and retrieve to return the segment with the result for a find

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
pascalconfluent: VertexAI doesn't return score but a distance.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
pascalconfluent: It return an empty array. I think we should also return an empty embedding.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/IndexEndpointService.java
pascalconfluent: Before creating an index we need to have file in the GCP other the index cannot be created. Will had the index creation later.

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
pascalconfluent: It is defined during index creation. Here is different algorithm that can be configured:

**SQUARED_L2_DISTANCE**	        

>  Euclidean (L2) Distance

**L1_DISTANCE**

> Manhattan (L1) Distance

**DOT_PRODUCT_DISTANCE**	

> Default value. Defined as a negative of the dot product.

**COSINE_DISTANCE**	

> Cosine Distance. We strongly suggest using DOT_PRODUCT_DISTANCE + UNIT_L2_NORM instead of the COSINE distance. Our algorithms have been more optimized for the DOT_PRODUCT distance, and when combined with UNIT_L2_NORM, it offers the same ranking and mathematical equivalence as the COSINE distance.

pascalconfluent: I am using the default one for testing!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAiEmbeddingIndex.java
pascalconfluent: Not sure what you are referring to.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAIDocument.java
pascalconfluent: Done!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
pascalconfluent: True!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
pascalconfluent: By default it uses dot product which the value increase in relation with the similarity (like the cosine). 
Right now we only support dot product (recommended by Google). I am looking at supporting all of them.
Here is the doc:
https://developers.google.com/machine-learning/clustering/similarity/measuring-similarity

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
langchain4j: Hmm... is index created automatically by your implementation or user manually creates it? I see that you create `VertexAiEmbeddingIndex` each time `addAll()` is called, but I can't find the code responsible for defining algorithm in your implementation...

But my initial point was that instead of using `CosineSimilarity.between(embedding, referenceEmbedding)`, please use `RelevanceScore.fromCosineSimilarity(CosineSimilarity.between(embedding, referenceEmbedding))`. This way it will be in line with all other integrations. And it should work the same for both cosine similarity and dot product + L2 normalization.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAiEmbeddingIndex.java
langchain4j: I am referring to `private int dimensions;`. You write into this variable, but seems that never read it.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
langchain4j: Ah, so `neighbor.getDistance()` returns dot product (similarity measure)? Very confusing naming...
If `neighbor.getDistance()` will return dot product (assuming L2 norm) or cosine similarity (-1..1), we should convert it into relevance score (0..1) using `RelevanceScore.fromCosineSimilarity()`. The `score` we have in `EmbeddingMatch` has a range from 0 (not relevant) to 1 (highly relevant).

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/VertexAIDocument.java
langchain4j: I can't find where this field is accessed, could you please point me in the right direction?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
langchain4j: An empty `Embedding` makes no sense, if we don't have an embedding, it is better to return null

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
langchain4j: Please see comments above

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/IndexEndpointService.java
langchain4j: Sorry, I am not sure I understand. `VertexAiVectorSearchIT.matchingEngine.indexId` should already be present when `VertexAiVectorSearchIT.clearStore()` is called and subsequently `VertexAiVectorSearch.deleteIndex()` calls `IndexEndpointService.deleteIndices()`. BTW I guess `deleteIndices` is not quite right name, since you are not deleting an index/indices, rather datapoints (using google naming)

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithoutMetadataIT.java
```diff
@@ -235,11 +235,15 @@ void should_return_correct_score() {
 
         EmbeddingMatch<TextSegment> match = relevant.get(0);
         assertThat(match.score()).isCloseTo(
-                RelevanceScore.fromCosineSimilarity(CosineSimilarity.between(embedding, referenceEmbedding)),
+                getCosineSimilarity(embedding, referenceEmbedding),
                 withPercentage(1)
         );
     }
 
+    protected double getCosineSimilarity(Embedding embedding, Embedding referenceEmbedding) {
```
langchain4j: please rename `getCosineSimilarity()` into `relevanceScore()`

langchain4j-vertex-ai/src/main/java/dev/langchain4j/store/embedding/vertexai/VertexAiVectorSearch.java
```diff
@@ -0,0 +1,281 @@
...
+    @Getter
+    @Builder.Default
+    private final boolean avoidDups = true;
+    /**
+     * If true, the full datapoint will be returned in the response.
+     * Default is false for performance reasons.
+     */
+    @Getter
+    @Builder.Default
+    private final boolean returnFullDatapoint = false;
```
langchain4j: I would also rename `returnFullDatapoint` into `retrieveEmbeddingsOnSearch`, we have this naming in `MilvusEmbeddingStore` already.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/store/embedding/vertexai/VertexAiVectorSearch.java
```diff
@@ -0,0 +1,281 @@
...
+     *
+     * @param index the index
+     */
+    public void deleteIndex(final String index) {
+        // Delete index and all documents
+        final String path = "indexes/" + index + ".json";
+        final String json = getGcpBlobService().download(path);
+        if (json != null) {
+            final VertexAiEmbeddingIndex embeddingIndex = VertexAiEmbeddingIndex.fromJson(json);
+            final List<String> indicesToDelete = embeddingIndex
```
langchain4j: maybe indicesToDelete -> embeddingIdsToDelete?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngine.java
langchain4j: [LangChain implementation](https://github.com/langchain-ai/langchain/blob/ee94ef55ee6ab064da08340817955f821dfa6261/libs/langchain/langchain/vectorstores/matching_engine.py#L159) seem to upload embeddings to GCS bucket first and then update the index, providing a reference to the bucket (where to read embeddings from). But in your implementation you provide embeddings directly in the index, so it feels that uploading them to GCS is unnecessary. Could you please remove this line and test if it still works? Thanks!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/store/embedding/vertexai/internal/IndexEndpointService.java
```diff
@@ -0,0 +1,153 @@
...
+        }
+    }
+
+    /**
+     * Upsert the embedding index.
+     *
+     * @param embeddingIndex the embedding index
+     */
+    public void upsertEmbedding(VertexAiEmbeddingIndex embeddingIndex) {
+        if (isNullOrBlank(indexId)) {
```
langchain4j, line 65: Why is this check required? Creating `IndexEndpointService` without `indexId` should fail

ghost (changes requested):
Hi @pascalconfluent, could you please check my comments? Thanks!

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiMatchingEngineIT.java
pascalconfluent: Each List<Embedding> has an index file attached to it. This index file is used by the matching engine to index it. If you create a batch mode index, then matching engine scan all the files to add the embeddings to the index. In short, an index is created in 2 phases, first create an index in matching engine (that is where you define the algorithm, ....), then you need to create files that represent a list of embeddings. I implemented the 2 phase, will probably do the first one but this one can be done manually using the UI. 

For the relevance, yes I can do that and assume that the index is created with the right algorithm.

By the way I looked the implementation from the python's langchain and they are doing the same.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/internal/MatchingService.java
pascalconfluent: That's debatable. Same goes with strings and other empty/null entities. 
Anyway, happy to return null instead of empty if no result.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/store/embedding/vertexai/VertexAiVectorSearch.java
pascalconfluent: Done!

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithoutMetadataIT.java
pascalconfluent: Done!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/store/embedding/vertexai/VertexAiVectorSearch.java
pascalconfluent: Done!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/store/embedding/vertexai/internal/IndexEndpointService.java
pascalconfluent: Yes, just did that for convenience but happy to make it failed.

geoand:
I assume this has gone pretty stale at this point :)

langchain4j:
AFAIK @glaforge wanted to work on adding Vertex AI Search soon, so maybe this PR can be revived?

glaforge:
Would be interesting to know if @pascalconfluent wants to update the PR?
We'll have to add support for metadata filtering as well, I guess.
But that would be great to see Vertex AI Search support (used to be called Matching Engine)

pascalconfluent:
Yes. I want to finish this PR.

langchain4j:
@pascalconfluent closing this PR due to inactivity. Please feel free to reopen if/when needed.

PR #298, opened 2023-11-19
OpenAI DALL·E support

Heezer:
Motivating generated picture 😄 ❤️ 
image


langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/ResponseLoggingInterceptor.java
```diff
@@ -0,0 +1,46 @@
...
+    return response;
+  }
+
+  static void log(Response response) {
+    try {
+      log.debug(
+        "Response:\n- status code: {}\n- headers: {}\n- body: {}",
+        response.code(),
+        RequestLoggingInterceptor.inOneLine(response.headers()),
+        getBody(response)
```
kumlien, line 29: Do you want to log the body?

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/RequestLoggingInterceptor.java
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  private static void log(Request request) {
+    try {
+      log.debug(
+        "Request:\n- method: {}\n- url: {}\n- headers: {}\n- body: {}",
+        request.method(),
+        request.url(),
+        inOneLine(request.headers()),
+        getBody(request)
```
kumlien, line 36: Do you want to log the body?

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public void setUrl(String url) {
+        this.url = url;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
```
langchain4j, line 67: I would replace this builder with a public ctor and `Image.from(String url)` static factory, builder is not very useful for a single field

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -0,0 +1,35 @@
+package dev.langchain4j.data.image;
+
+public class Image {
+    private String url;
+
+    public String getUrl() {
```
langchain4j: minor: please avoid using "get" and "set" in getters and setters, we are following a modern "record" style

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ImageProcessingModel.java
```diff
@@ -0,0 +1,8 @@
+package dev.langchain4j.model.chat;
+
+import dev.langchain4j.data.image.Image;
+import dev.langchain4j.model.output.Response;
+
+public interface ImageProcessingModel {
```
langchain4j: What about renaming it into `ImageGeneratingModel`, as it basically generates an image from a prompt at this point? Processing sounds like it processes images, which is not true, at least at the moment.
DALLE2 and probably other providers also allow to edit and create variations of existing image, so maybe these methods will be added in the future, then we can name it simply `ImageModel`, (as opposed to `LanguageModel`, `EmbeddingModel`, etc). What do you think?

I would also move it into a separate "image" package, as it is not related to a chat.

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/AuthorizationHeaderInjector.java
```diff
@@ -0,0 +1,20 @@
...
+import org.jetbrains.annotations.NotNull;
+
+@RequiredArgsConstructor
+class AuthorizationHeaderInjector implements Interceptor {
+
+  private final String apiKey;
+
+  @NotNull
+  public Response intercept(Interceptor.Chain chain) throws IOException {
+    Request request = chain.request().newBuilder().addHeader("Authorization", "Bearer " + this.apiKey).build();
```
langchain4j: All this logic is already implemented in openai4j library, I would suggest moving missing relevant parts to image generation there

langchain4j-open-ai-dall-e/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>0.24.0</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-open-ai-dall-e</artifactId>
```
langchain4j: I am not sure we should introduce a new module for a specific OpenAI model, we have all their models already in langchain4j-open-ai

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleClient.java
```diff
@@ -0,0 +1,83 @@
...
+    Duration connectTimeout,
+    Duration readTimeout,
+    Duration writeTimeout,
+    Proxy proxy,
+    boolean logRequests,
+    boolean logResponses
+  ) {
+    OkHttpClient.Builder cBuilder =
+      (new OkHttpClient.Builder()).callTimeout(getOrDefault(callTimeout, ofSeconds(60)))
+        .connectTimeout(getOrDefault(connectTimeout, ofSeconds(60)))
```
langchain4j: a lot of logic from this class already exists in openai4j

pom.xml
```diff
@@ -28,14 +28,17 @@
...
+
         <!-- embedding stores -->
         <module>langchain4j-cassandra</module>
         <module>langchain4j-chroma</module>
         <module>langchain4j-elasticsearch</module>
         <module>langchain4j-milvus</module>
         <module>langchain4j-opensearch</module>
-        <module>langchain4j-pgvector</module>
         <module>langchain4j-pinecone</module>
+        <module>langchain4j-pgvector</module>
```
langchain4j: g comes before i, or? :)

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleImageProcessingModel.java
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Instantiates OpenAI DALL·E image processing model.
+   * Find the parameters description <a href="https://platform.openai.com/docs/api-reference/images/create">here</a>.
+   * Only one image can be generated pro request currently.
+   *
+   * @param model dall-e-3 is default one
+   * @param downloadFolder specifies the local folder where the generated image will be downloaded to (in case provided).
+   */
+  @Builder
+  public OpenAiDalleImageProcessingModel(
```
langchain4j: Maybe OpenAiImageModel? Dalle is a name of the current models, and it is set as a parameter. I can imagine they adding a new generation of models that will be called differently but will have the same API

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleRequest.java
```diff
@@ -0,0 +1,19 @@
+package dev.langchain4j.image.openai.dalle;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Builder
+@Data
+public class OpenAiDalleRequest {
```
langchain4j: Maybe `OpenAiImageGenerationRequest` ?

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiImageDownloader.java
```diff
@@ -0,0 +1,37 @@
...
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static String extractFileName(String url) {
+    Pattern pattern = Pattern.compile(".*/([^/?]+)\\?.*");
+    Matcher matcher = pattern.matcher(url);
+
+    return matcher.find() ? matcher.group(1) : "image.png";
```
langchain4j: for "image.png" case: should we maybe add a timestamp to it?

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleImageProcessingModel.java
```diff
@@ -0,0 +1,105 @@
...
+      .size(size)
+      .model(model)
+      .quality(quality)
+      .style(style)
+      .build();
+
+    OpenAiDalleResponse response = withRetry(() -> client.generate(request), maxRetries);
+    String url = response.getData().get(0).getUrl();
+    if (downloadFolder != null) {
+      url = OpenAiImageDownloader.downloadFile(url, downloadFolder);
```
langchain4j: what about doing downloading also withRetry()?

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -0,0 +1,35 @@
+package dev.langchain4j.data.image;
+
+public class Image {
+    private String url;
```
langchain4j: Not sure if it is a good idea, but what about using `URI` instead of a `String` here?

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleImageProcessingModel.java
```diff
@@ -0,0 +1,105 @@
...
+  public static final String DALL_E_QUALITY_STANDARD = "standard";
+  public static final String DALL_E_QUALITY_HD = "hd";
+  public static final String DALL_E_STYLE_VIVID = "vivid";
+  public static final String DALL_E_STYLE_NATURAL = "natural";
+
+  private final String model;
+  private final String size;
+  private final String quality;
+  private final String style;
+  private final String downloadFolder;
```
langchain4j: nitpick: what about `String downloadFolder` -> `Path downloadDirectory` ?

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleResponse.java
```diff
@@ -0,0 +1,17 @@
+package dev.langchain4j.image.openai.dalle;
+
+import java.util.List;
+import lombok.Data;
+
+@Data
+class OpenAiDalleResponse {
```
langchain4j: Maybe `OpenAiImageGenerationResponse`?

ghost (commented):
@Heezer thanks a lot! Please check my comments.
I trust that it works, but I would add at least one integration test to make sure it keeps working in the future.
Thanks!

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer: I've just implemented the initial simplified version here with `url` field only, the next would be `b64_json` and `revised_prompt`, see https://platform.openai.com/docs/api-reference/images and it still makes sense from my side to keep this class generic, therefore I've put it into core. Tell me if you still see the changes needed here.

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer: done

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ImageProcessingModel.java
Heezer: Yes, my intention about the name was to be more generic in case of https://platform.openai.com/docs/api-reference/images/createEdit, indeed this is supported by DALL·E 2 only. I'm fine with `ImageModel` and to be honest I cannot remember why I've put it into `chat` 😄, `image` package makes more sense 😄

Heezer: done

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/AuthorizationHeaderInjector.java
Heezer: yes, auth and logging interceptors are there, but I was not sure who is in charge for `openai4j` and whether I needs to touch it 😄 ok, now I know this lib is yours as well and if you suggest to move it...

langchain4j-open-ai-dall-e/pom.xml
Heezer: hm... why not? going to pack it into existing `langchain4j-open-ai` then

pom.xml
Heezer: who cares? 😂 fixed

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/AuthorizationHeaderInjector.java
Heezer: I've finally managed to refactor the main logic to openai4 lib, so please review it: https://github.com/ai-for-java/openai4j/pull/5

Heezer:
> @Heezer thanks a lot! Please check my comments. I trust that it works, but I would add at least one integration test to make sure it keeps working in the future. Thanks!

integra test is within openai4j project

langchain4j-open-ai-dall-e/pom.xml
Heezer: done

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleClient.java
Heezer: refactored

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleImageProcessingModel.java
Heezer: done

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleRequest.java
Heezer: done in openai4j

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiImageDownloader.java
Heezer: done in openai4j

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleImageProcessingModel.java
Heezer: this part resides in openai4j now

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer: done in openai4j

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleImageProcessingModel.java
Heezer: done in openai4j

langchain4j-open-ai-dall-e/src/main/java/dev/langchain4j/image/openai/dalle/OpenAiDalleResponse.java
Heezer: done in openai4j

Heezer:
> @Heezer thanks a lot! Please check my comments. I trust that it works, but I would add at least one integration test to make sure it keeps working in the future. Thanks!

inte test added

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -0,0 +1,69 @@
+package dev.langchain4j.data.image;
+
+import java.net.URI;
+import java.util.Objects;
+
+public final class Image {
+
+    private URI url;
+    private String revisedPrompt;
```
langchain4j, line 12: I am not sure we should leak this field, at least right now (we can always add it later), as it is OpenAI-specific and it is not clear if this approach will be adopted by other model provider(s)

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -0,0 +1,69 @@
+package dev.langchain4j.data.image;
+
+import java.net.URI;
+import java.util.Objects;
+
+public final class Image {
+
+    private URI url;
```
langchain4j, line 10: I think we should add a base64 string here as an alternative to URL

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(url, revisedPrompt);
+    }
+
+    @Override
+    public String toString() {
+        return "Image{" + "url='" + url + '\'' + ", revisedPrompt='" + revisedPrompt + '\'' + '}';
```
langchain4j: minor: there is `Utils.quoted()`, see `ToolParameters.toString()` as an example

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
```diff
@@ -0,0 +1,9 @@
+package dev.langchain4j.model.image;
+
+import dev.langchain4j.data.image.Image;
+import dev.langchain4j.model.output.Response;
+import java.util.List;
+
+public interface ImageModel {
+  Response<List<Image>> generate(String prompt);
```
langchain4j: I propose to simplify a bit and return a single image in the API to be consistent with other our model classes that return a single answer for a single prompt. Dalle-3 anyway supports only n=1 and user can always call the model multiple times if he wants multiple images, but I doubt it will be used often in programmatic way as images cost quite a lot. We can always add another method that returns a list in the future (if it will be indeed needed).

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
```diff
@@ -0,0 +1,149 @@
...
+            .persistTo(persistTo);
+
+        if (withPersisting != null && withPersisting) {
+            cBuilder.withPersisting();
+        }
+
+        this.client = cBuilder.build();
+
+        this.maxRetries = getOrDefault(maxRetries, 3);
+        this.model = model;
```
langchain4j, line 89: I would add a default model to be consistent with other model classes and be able to create a model just by providing an api key.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
```diff
@@ -0,0 +1,149 @@
...
+            .builder()
+            .prompt(prompt)
+            .size(size)
+            .quality(quality)
+            .style(style)
+            .user(user)
+            .responseFormat(responseFormat);
+
+        if (DALL_E_2.equals(model)) {
+            requestBuilder.model(dev.ai4j.openai4j.image.ImageModel.DALL_E_2);
```
langchain4j, line 145: minor: would be nice to have overloaded `model(String)` method in openai4j (as well as change `Builder.model` from `ImageModel` to `String`) to not limit only to these models in enum and not to force to make a conversion from string to enum here

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        if (n != null) {
+            requestBuilder.n(n);
```
langchain4j: minor: this is why I use object types instead of primitives in configs, it is easier to pass them through (no need to check if it is null or not) and also there is an understanding if user defined a config or not (primitive will always have a value).

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
```diff
@@ -0,0 +1,149 @@
...
+
+        return Response.from(
+            response
+                .data()
+                .stream()
+                .map(d -> Image.builder().url(d.url()).revisedPrompt(d.revisedPrompt()).build())
+                .collect(Collectors.toList())
+        );
+    }
+}
```
langchain4j, line 150: Could you please add a static `withApiKey(String apiKey)` factory method for parity with other models?

langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiImagesGenerationIT.java
```diff
@@ -0,0 +1,105 @@
...
+import dev.langchain4j.data.image.Image;
+import dev.langchain4j.model.output.Response;
+import java.io.File;
+import java.net.URI;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class OpenAiImagesGenerationIT {
```
langchain4j: minor: `OpenAiImageModelIT`

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
```diff
@@ -0,0 +1,149 @@
...
+
+    public static class OpenAiImageModelBuilder {
+
+        public OpenAiImageModelBuilder withPersisting() {
+            withPersisting = true;
+            return this;
+        }
+
+        public OpenAiImageModelBuilder logRequests() {
+            logRequests = true;
```
langchain4j: This is a really nice change but I would propose to make it in a separate PR and update all model classes at once to keep the consistency

ghost (changes requested):
@Heezer great job, thank you! Please check comments

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
langchain4j: @jdubois please check this as well

jdubois: Yes, I already have this, thanks!

jdubois: I haven't pushed yet because of the base64 part, it will be there soon

langchain4j: @jdubois this is to your comment
> This is fixing my 2 biggest issues with this API (not limiting to just 1 image, and being able to get the image content and not just the URL)


I am not convinced we should return a List, I would start from a single Image and later on we can add another mothod to return more (if required)

jdubois: We could do both....
Also we need to add a new parameter here to support both base64 and URL, I'll give it a try in my PR

langchain4j: both works, although a bit premature. In any case my main point here is to avoid forcing user to go `.get(0)` all the time

jdubois: Argh, I couldn't make it work and just read the doc: ` Dall-e-3 models only support a value of 1.`!! So that would only be for DALL E 2, so not really worth it.
I'll just return one value, then.

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer: on the other side other models can simply ignore this field, I opt for this, since it's pity to remove good work done 😄

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer: yes, as discussed in #359, I'm fine with that

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
Heezer: guys, I don't really won't to throw my support of multiple images, instead I would have two methods:
```
public interface ImageModel {
  Response<Image> generate(String prompt);
  default Response<List<Image>> generateMultiple(String prompt) {
    throw new IllegalArgumentException("Operation is not supported");
  }
}
```
like it's non blocking, what do you think?

langchain4j: Why do you want to have multiple images at once? It is currently supported only with outdated dalle2 model...

Heezer: If I speak as user, then sure I want to have image variants to my prompt to choose from and technically - the response from DALL·E contains always a list, I suppose DALL·E 3 will support it, this is just matter of time

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
Heezer: the default model is defied in openai4j already, find following code in examples:
````
OpenAiImageModel model = OpenAiImageModel.builder().apiKey(System.getenv("OPENAI_API_KEY")).build();
````

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
Heezer: intention here was to use the default value defined in openai4j, so if primitive type is used here, then I have to define a default value here again, right?

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
langchain4j: Ok, how about `Response<List<Image>> generate(String prompt, int n)`?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
langchain4j: No, no need to define default value for n, it is an optional parameter. So if you have `Integer n` here, and `Integer n` in openai4j, then you can simply do `.n(n)` and not worry about if it is null or not. If it was not befined by user and is null, then it will have no effect and OpenAI's default value will be used. If user defined it, then it will be used.

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
Heezer: sure, I like this!

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer: done, inte test extended

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer: done

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
Heezer: done

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
Heezer: this part is obsolete now

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
Heezer: done

langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiImagesGenerationIT.java
Heezer: done

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiImageModel.java
Heezer: done

ghost (approved):
@Heezer thank you so much, great job!

Issue #299, opened 2023-11-19
[BUG] NP in OpenAITokenizer


kumlien:
Please provide as much details as possible, this will help us to deliver a fix as soon as possible.
Thank you!

**Describe the bug**
The OpenAiTokenizer has  a method for estimating the number of tokens for the Tools at hand. One step in this process is to estimate the number of tokens for the parameters of a Tool, using the properties of the parameters. This fails when a Tool doesn't have any parameters, the Tokenizer still tries to access the properties of the parameters (which now are null). The access of the properties needw to be done in a null-safe way. 

**Log and Stack trace**
`Caused by: java.lang.NullPointerException: Cannot invoke "dev.langchain4j.agent.tool.ToolParameters.properties()" because the return value of "dev.langchain4j.agent.tool.ToolSpecification.parameters()" is null
	at dev.langchain4j.model.openai.OpenAiTokenizer.estimateTokenCountInToolSpecifications(OpenAiTokenizer.java:93) ~[langchain4j-open-ai-0.24.0.jar:na]
	at dev.langchain4j.model.openai.OpenAiStreamingChatModel.generate(OpenAiStreamingChatModel.java:119) ~[langchain4j-open-ai-0.24.0.jar:na]
	at dev.langchain4j.model.openai.OpenAiStreamingChatModel.generate(OpenAiStreamingChatModel.java:91) ~[langchain4j-open-ai-0.24.0.jar:na]
	at dev.langchain4j.service.AiServiceTokenStream$AiServiceOnStart.start(AiServiceTokenStream.java:78) ~[langchain4j-0.24.0.jar:na]`

**To Reproduce**
Set up a OpenAiStreamingChatModel using the default Tokenizer (OpenAiTokenizer). Add a Tool without parameters like:
`@Tool("This tool is used to get the current date and time in ISO-8601 format")
    public String getCurrentDateAndTime() {
        Logger.info("Returning current local date with param '{}'");
        return LocalDateTime.now().format(DateTimeFormatter.ISO_DATE_TIME);
    }`

**Expected behavior**
Expected behavior is not to throw a NP

**Please complete the following information:**
- LangChain4j version: 0.24.0
- Java version: 21
- Spring Boot version (if applicable): 3.1.5

**Additional context**

PR #303, opened 2023-11-20
#299 Add null check for ToolsParameters in tokenCountEstimation

kumlien:
Add null check when estimating the number of tokens in the ToolParameters. 
The parameters are always null for a Tool with no parameters.


ghost (approved):
@kumlien thank you!

PR #310, opened 2023-11-25
Use Testcontainers annotation

eddumelendez:
`@Container` is used along with `@Testcontainers` for initialization.
The current test has `PostgreSQLContainer` annotated with `@Container`
and it is manually initialized.


ghost (approved):
@eddumelendez thank you!

PR #312, opened 2023-11-25
Use ElasticsearchContainer in Elasticsearch IT

eddumelendez:
Current test execution depends on a running elasticserach instance,
which depends on a manual execution of docker command. This commit
introduces testcontainers as in other modules to run such instance
along with tests, improving DevEx.


ghost (approved):
@eddumelendez great job, thank you!

PR #314, opened 2023-11-25
Enable Milvus IT by using Testcontainers

eddumelendez:
Currently, Milvus IT is disables because it requires a manual step
to run the compose file. Adding Testcontainers, the infrastructure
is in place to run the tests.


ghost (approved):
@eddumelendez learning new things from you today, thank you!

PR #315, opened 2023-11-26
Use Testcontainers in Ollama IT

eddumelendez:
Currently, integration tests in Ollama module are disabled because
it needs a Ollama instance running in order to execute them.
Testcontainers provides this infrastructure not only by running
the ollama container but also by automating the pull model step.

This commit, use the Singleton Container approach to reuse the single
instance across multiple IT. Also, pull step is only executed when the
image is `ollama/ollama`.

The behavior on this is:
1st execution:
1. Pull `ollama/ollama` image
2. Start the container based on `ollama/ollama` image
3. Download the `orca-mini` model
4. Create an image based on the current state (with the model in it)
5. Declare the container ready to use
6. Run test

Next executions:
1. Look for the local image created in the 1st execution
2. Start the container based on the local image
3. Declare the container ready to use
4. Run test

1st execution is expected to take longer because of the model (3GB).
Next execution are way more faster.


ghost (approved):
@eddumelendez that is incredible, thank you so much! We should definitely do the same for LocalAI module

eddumelendez:
@langchain4j I tried to do the same for LocalAI but IIRC I was not able to load the model for the tests. If you can share the steps to run the instance with the model, I can try to reproduce it using Testcontainers. Feel free to create an issue with the steps and assign it to me.

bidek:
@eddumelendez Is it possible to use Testcontainers with GPU support? In Docker, I can pass the `--gpus=all` option to the run command, for example, `docker run -d --gpus=all -v ollama:/root/.ollama -p 11434:11434 --name ollama ollama/ollama`.

I am asking because I added image input support for Ollama #430 but can't figure out how to set up GPU with integration tests. The current setup waits 1 minute for a response, and after that, it times out. For now, I have changed the timeout to 3 minutes. My PC is able to return a response in about 2 minutes and 20 seconds.

Waiting nearly 3 minutes for one test isn't a good development experience. I know they are integration tests, so by definition, they can be slower than unit tests, but maybe we can do something about it. Additionally, I'd like to note that the proposed GPU support should only be available for local development, as the current CI pipeline doesn't allow GPU support.

PR #328, opened 2023-12-01
Migration of the AzureOpenAiChatModel to use the Azure OpenAI SDK

### LINKED ISSUE #325: [FEATURE] Use Azure OpenAI SDK to connect to Azure OpenAI
Azure OpenAI support was added in https://github.com/langchain4j/langchain4j/pull/107 by @kuraleta and uses the https://github.com/ai-for-java/ai4j API to connect.

There is an official Azure OpenAI SDK at https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/openai/azure-ai-openai which is official and should be better maintained, so my proposal is to migrate Azure OpenAI support to this SDK.

Then, I believe this was used because you have both an Azure OpenAI and an OpenAI support, and they are both using the same https://github.com/ai-for-java/ai4j API. I'm not sure why you separated them, so once this first migration is done, I'd love to have a look at unifying your Azure OpenAI and your OpenAI support.

jdubois:
This PR fixes https://github.com/langchain4j/langchain4j/issues/325

- [x] Migrate AzureOpenAiChatModel
- [x] Migrate AzureOpenAiEmbeddingModel
- [x] Migrate AzureOpenAiLanguageModel
- [x] Migrate AzureOpenAiStreamingChatModel
- [x] Migrate AzureOpenAiStreamingLanguageModel
- [x] Add a full suite of tests


clun:
wow if even rockstars start contributing here......

jdubois:
Ahah I've had a good look at your work here @clun and I'll try to do the same for Azure Cognitive Search...

geoand:
Really nice!

The only thing I can think of with this one is that it breaks integrations using `dev.ai4j.openai4j.spi.OpenAiClientBuilderFactory` because with this that client is swapped out and never comes into play. 
From an integrator's perspective, the ideal thing here would be to have an Azure specific SPI that would allow us to use the previous client. 
If this is not an option, we'll have to pretty much duplicate the functionality of this module in the `quarkus-langchain4j` extension.

jdubois:
Thank you @geoand but why do you need this client? For Azure OpenAI, it's better to use the official client (and I'm not saying the client is better at all, just that it's the official one)

geoand:
There are a few reasons:

* It reuses the network stack we have in Quarkus - this important for us for keeping runtime footprint (across various metrics) to a minimum
* It makes compilation to native-image without any necessary configuration 
* It provides us out of the box observability features

jdubois:
Thanks for insights, I had no idea this client was using some Quarkus-specific APIs. As the Azure SDK usually works very well with Spring, I'm guessing there will be the opposite argument for Spring users.
I believe it would be better to use the official Azure SDK here, just because it's official (again, not because it's better - and to be honest I don't believe it is!), and then have framework-specific optimisations in each framework (like `quarkus-langchain4j`). Then I'm concerned (for Quarkus and for Azure), if you need to re-implement the whole SDK just for this. Let me ping some friends inside Microsoft who work on both Quarkus and the Azure SDK, maybe they will have some ideas (hey @agoncal !!).

geoand:
👍🏼 

Just to clarify, it would work in Quarkus too, it just won't have all the bells and whistles I mentioned above.

langchain4j:
@jdubois pardon me for the oversight, @geoand brought good points.

Existing client does not use/depend on any Quarkus-specific APIs, but it has an [SPI](https://github.com/ai-for-java/openai4j/blob/b5fac4c891df38c46f30b31860f9d0f9b0fe76b4/src/main/java/dev/ai4j/openai4j/OpenAiClient.java#L43) that allows Quarkus to use their network stack and have all the benefits mentioned above. It does not mean that users of Spring will see any downsides.

I would propose to put this PR on hold for the moment and allow some time to explore options. Thanks for understanding!

jdubois:
No it's indeed the same issue with Spring:

- With "ai4j" you're using Retrofit + OKHttp, and I'm guessing this works fine with Mutiny, so it's all good for Quarkus.
- With the Azure SDK, you're using Netty + Reactor, which means it's very good for Spring.

Both will work with the other framework, but they'll add more dependencies (and I'm guessing a second event loop), so it'll be less efficient for the "secondary" framework than for the "primary" framework.

Now I don't want to enter the framework wars, here the issue for me is whether you should be using the official Azure SDK, from the Azure team, to connect to Azure OpenAI. Or use the client from https://github.com/ai-for-java which is a totally unknown organization (their org page is empty).

langchain4j:
@jdubois fair point. [openai4j](https://github.com/ai-for-java/openai4j) is ours, so I think I can adapt it to use Netty+Reactor by default, then both Spring and Quarkus users will benefit. What is the problem with using non-official library?

geoand:
> so I think I can adapt it to use Netty+Reactor by default

I think this is a good way to go

jdubois:
@langchain4j it's weird, so you're 1 person and coded everything by yourself? You should clarify it somewhere, because I didn't understand at all that LangChain4J and ai-for-java belonged to the same person - you're not using the same GitHub ID for both, and there's no name anywhere, so there's no way to know.

Concerning using the official library: 

1. It's first an issue of trust of your software supply chain, like here I have no idea who codes what and how things are maintained, when as an Azure customer people are looking for something they can trust. They are sending their secret keys and credentials through those librairies. They are running this code in production, and don't want any license issue.
2. Microsoft guarantees the maintenance and an SLA, and works on non-public stuff in advance, so by using the official library you should have more features and less bugs. I'm not saying it's better coded (It's not!!! As most of it is automatically generated), then that's also why I'm interested in LangChain4J (you provide a better quality, neutral abstraction layer).
3. If you're a user of Azure OpenAI, chances are that you're also using other parts of Microsoft cloud. Maybe some Cognitive Search (for the RAG pattern), Cosmos DB or Blob storage (for storing your files). All of those work using the same Azure SDK.
4. It's a mix of the 3 points above: if you want authentication with Active Directory, then you'll need to go use the Azure SDK. And many people will request something like this, they're not going to have their Azure OpenAI service open on the Internet or easily accessible, that's one of the main points of using Azure.

Anyway, I get your point in having something specific for Quarkus, so I have 2 other proposals for you:

- We keep both implementations here, in different directories. We should share the integration tests, and then it'll be easy for people to switch.
- I put this "langchain-azure-open-ai-official" library in a different repo, hopefully under the Azure organization. This would require some paperwork, and I'm guessing some collaboration on your side.

geoand:
Thanks for elaborating @jdubois.

Just to be clear, this repo does not have anything specific to Quarkus. All it has is an SPI that can be used by code to swap out various parts - this is what the Quarkus extension uses.

langchain4j:
@jdubois good point, I would indeed prefer using oficial SDK if I were Azure user. I think we can indeed use official Azure SDK for `langchain4j-azure-open-ai` module and keep using `openai4j` for `langchain4j-open-ai`. `langchain4j-azure-open-ai` in current state (using `openai4j`) will be moved as-is to `langchain4j-quarkus` and everyone will be happy.

jdubois:
Oh yes totally agree @langchain4j !

jdubois:
I'm putting this PR as "ready for review", as I finished the following tasks:

- Migrate all the original `langchain4j-azure-open-ai` module to use the official Azure AI SDK
- Add integration tests, and test everything (using my subscription, we'll need to check if that can be automated)
- The original `langchain4j-azure-open-ai` module is migrated to a new `langchain4j-quarkus` module

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -1,36 +1,46 @@
...
  * Represents an OpenAI language model, hosted on Azure, that has a chat completion interface, such as gpt-3.5-turbo.
  * <p>
- * Mandatory parameters for initialization are: baseUrl, apiVersion and apiKey.
+ * Mandatory parameters for initialization are: endpoint, serviceVersion, apiKey, deploymentName and modelName.
  * <p>
  * There are two primary authentication methods to access Azure OpenAI:
  * <p>
  * 1. API Key Authentication: For this type of authentication, HTTP requests must include the
- * API Key in the "api-key" HTTP header.
+ * API Key in the "Authorization" HTTP header as follows: `Authorization: Bearer OPENAI_API_KEY`
```
langchain4j: [This page](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference) says there are 2 auth options:
1. `API Key authentication` uses API Key sent as `api-key` header (is supported by langchain4j/openai4j so far)
2. `Microsoft Entra authentication` uses Microsoft Entra token sent into `Authorization Bearer ...` header (see option 2 in javadoc below, currently not supported).

As far as I understand, Entra token and API key are different things. With your changes, this class still takes `apiKey`, and then sends it as an `api-key` header.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -62,31 +76,42 @@ public AzureOpenAiChatModel(String baseUrl,
...
+        clientOptions.setConnectTimeout(timeout);
+        clientOptions.setResponseTimeout(timeout);
+        clientOptions.setReadTimeout(timeout);
+        clientOptions.setWriteTimeout(timeout);
+        clientOptions.setProxyOptions(proxyOptions);
+
+        HttpClient httpClient = new NettyAsyncHttpClientProvider().createInstance(clientOptions);
+
+        HttpLogOptions httpLogOptions = new HttpLogOptions();
+        if (logRequests != null) {
```
langchain4j: can be not null, but false

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -62,31 +76,42 @@ public AzureOpenAiChatModel(String baseUrl,
...
+            httpLogOptions.setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS);
```
langchain4j: Is this applicable to both requests and responses? Javadoc does not say this and I don't have a key to check :(
If yes, what about renaming `logRequests` into `logRequestsAndResponses`?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -42,18 +52,22 @@
...
     private final Double topP;
     private final Integer maxTokens;
     private final Double presencePenalty;
     private final Double frequencyPenalty;
     private final Integer maxRetries;
-    private final Tokenizer tokenizer;
 
-    public AzureOpenAiChatModel(String baseUrl,
-                                String apiVersion,
+    public AzureOpenAiChatModel(String endpoint,
```
langchain4j, line 68: optional: should we maybe add another ctor that takes `OpenAIClient` as an argument to give user more flexibility?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -42,18 +52,22 @@
...
+                                String serviceVersion,
                                 String apiKey,
+                                String deploymentName,
+                                String modelName,
```
langchain4j: Having both `deploymentName` and `modelName` is confusing. `modelName` is now used only to create a tokenizer, so maybe we should just remove it. If one wants to configure a tokenizer, one can provide an instance of a `Tokenizer` himself.
And if tokenizer is not provided, I think it is safe to default to `new OpenAiTokenizer(GPT_3_5_TURBO)`.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -108,29 +133,27 @@ private Response<AiMessage> generate(List<ChatMessage> messages,
...
         }
         if (toolThatMustBeExecuted != null) {
-            requestBuilder.functionCall(toolThatMustBeExecuted.name());
+            options.setFunctionCall(new FunctionCallConfig(toolThatMustBeExecuted.name()));
         }
 
-        ChatCompletionRequest request = requestBuilder.build();
-
-        ChatCompletionResponse response = withRetry(() -> client.chatCompletion(request).execute(), maxRetries);
+        ChatCompletions chatCompletions = withRetry(() -> client.getChatCompletions(deploymentName, options), maxRetries);
```
langchain4j: `OpenAIClientBuilder` has `retryOptions`, we should probably use them instead of `withRetry()`

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -233,8 +279,8 @@ public Builder maxRetries(Integer maxRetries) {
             return this;
         }
 
-        public Builder proxy(Proxy proxy) {
-            this.proxy = proxy;
+        public Builder ProxyOptions(ProxyOptions proxyOptions) {
```
langchain4j: ProxyOptions -> proxyOptions

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModel.java
langchain4j, line 1: All of the comments in `AzureOpenAiChatModel` also apply here and other model classes

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
```diff
@@ -111,55 +132,51 @@ private void generate(List<ChatMessage> messages,
...
-                })
-                .onComplete(() -> {
-                    Response<AiMessage> response = responseBuilder.build();
-                    handler.onComplete(response);
-                })
-                .onError(handler::onError)
-                .execute();
+        AzureOpenAiStreamingResponseBuilder responseBuilder = new AzureOpenAiStreamingResponseBuilder(inputTokenCount);
+
+        client.getChatCompletionsStream(deploymentName, options)
```
langchain4j: What will happen if an error happens during streaming? User of this class will expect to get an exception inside `StreamingResponseHandler.onError()`.
optional: would be great to have a test for this case

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingLanguageModel.java
```diff
@@ -41,67 +54,105 @@
...
-                            response.content().text(),
-                            response.tokenUsage(),
-                            response.finishReason()
-                    ));
-                })
-                .onError(handler::onError)
-                .execute();
+        AzureOpenAiStreamingResponseBuilder responseBuilder = new AzureOpenAiStreamingResponseBuilder(inputTokenCount);
+
+        client.getCompletionsStream(deploymentName, options)
```
langchain4j: same here about error handling

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -0,0 +1,176 @@
...
+import dev.langchain4j.model.output.TokenUsage;
+
+import java.util.*;
+
+import static com.azure.ai.openai.models.ChatRole.*;
+import static dev.langchain4j.data.message.AiMessage.aiMessage;
+import static dev.langchain4j.model.output.FinishReason.*;
+import static java.util.stream.Collectors.toList;
+
+public class InternalAzureOpenAiHelper {
```
langchain4j: minor: can be package-private

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAIChatModelIT.java
```diff
@@ -0,0 +1,180 @@
...
+        assertThat(response.content().toolExecutionRequest().name()).isEqualTo(toolName);
+
+        // We should get a response telling how to call the "getCurrentWeather" function, with the correct parameters in JSON format.
+        System.out.println(response);
+
+        // We can now call the function with the correct parameters.
+        ToolExecutionRequest toolExecutionRequest = response.content().toolExecutionRequest();
+        WeatherLocation weatherLocation = BinaryData.fromString(toolExecutionRequest.arguments()).toObject(WeatherLocation.class);
+        int currentWeather = 0;
+        if (Objects.equals(toolExecutionRequest.name(), toolName)) {
```
langchain4j: this will always return true as you already assert this on line 99

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAIChatModelIT.java
```diff
@@ -0,0 +1,180 @@
...
+        assertThat(weather).isEqualTo("The weather in Paris, France is 35 degrees celsius.");
+
+        // Now that we know the function's result, we can call the model again with the result as input.
+        ToolExecutionResultMessage toolExecutionResultMessage = toolExecutionResultMessage(toolName, weather);
+        SystemMessage systemMessage = SystemMessage.systemMessage("If the weather is above 30 degrees celsius, recommend the user wears a t-shirt and shorts.");
+
+        List<ChatMessage> chatMessages = new ArrayList<>();
+        chatMessages.add(userMessage);
+        chatMessages.add(systemMessage);
+        chatMessages.add(toolExecutionResultMessage);
```
langchain4j, line 130: minor: although this is probably working (at least for old functions, will not work for new tools), AFAIK one is supposed to put AiMessage (the one returned in the first LLM call) right before the tool message

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -156,29 +181,28 @@ public static class Builder {
...
-        private Proxy proxy;
+        private ProxyOptions proxyOptions;
         private Boolean logRequests;
-        private Boolean logResponses;
 
         /**
          * Sets the Azure OpenAI base URL. This is a mandatory parameter.
          *
-         * @param baseUrl The Azure OpenAI base URL in the format: https://{resource}.openai.azure.com/openai/deployments/{deployment}
+         * @param endpoint The Azure OpenAI base URL in the format: https://{resource}.openai.azure.com/
```
langchain4j: minor: base URL -> endpoint

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -156,29 +181,28 @@ public static class Builder {
         private Double frequencyPenalty;
         private Duration timeout;
         private Integer maxRetries;
-        private Proxy proxy;
+        private ProxyOptions proxyOptions;
         private Boolean logRequests;
-        private Boolean logResponses;
 
         /**
          * Sets the Azure OpenAI base URL. This is a mandatory parameter.
```
langchain4j: minor: base URL -> endpoint

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -156,29 +181,28 @@ public static class Builder {
...
          */
-        public Builder baseUrl(String baseUrl) {
-            this.baseUrl = baseUrl;
+        public Builder endpoint(String endpoint) {
+            this.endpoint = endpoint;
             return this;
         }
 
         /**
          * Sets the Azure OpenAI API version. This is a mandatory parameter.
```
langchain4j: minor: API  -> service

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -193,6 +217,28 @@ public Builder apiKey(String apiKey) {
...
+            return this;
+        }
+
+        /**
+         * Sets the model name in Azure OpenAI. This is a mandatory parameter.
+         *
+         * @param modelName The model name.
+         * @return builder
+         */
+        public Builder modelName(String modelName) {
```
langchain4j: can be removed (see comment above)

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModel.java
```diff
@@ -100,17 +128,20 @@ private Response<List<Embedding>> embedTexts(List<String> texts) {
...
+            Embeddings response =  withRetry(() -> client.getEmbeddings(deploymentName, options), maxRetries);
 
-            embeddings.addAll(response.data().stream()
-                    .map(openAiEmbedding -> Embedding.from(openAiEmbedding.embedding()))
-                    .collect(toList()));
+            for (EmbeddingItem embeddingItem : response.getData()) {
+                List<Double> openAiVector = embeddingItem.getEmbedding();
+                float[] langChainVector = new float[openAiVector.size()];
+                for (int index = 0; index < openAiVector.size(); index++) {
+                    langChainVector[index] = openAiVector.get(index).floatValue();
```
langchain4j: optional: would be nice to move this logic into `public static Embedding from(List<Double> vector) {...}` inside `Embedding`

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingResponseBuilder.java
```diff
@@ -0,0 +1,130 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+import static dev.langchain4j.model.openai.InternalOpenAiHelper.finishReasonFrom;
+
+/**
+ * This class needs to be thread safe because it is called when a streaming result comes back
+ * and there is no guarantee that this thread will be the same as the one that initiated the request,
+ * in fact it almost certainly won't be.
+ */
+public class AzureOpenAiStreamingResponseBuilder {
```
langchain4j: minor: can be package-private

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModelIT.java
```diff
@@ -0,0 +1,35 @@
...
+
+        assertThat(response.content().vector()).hasSize(1536);
+
+        TokenUsage tokenUsage = response.tokenUsage();
+        assertThat(tokenUsage.inputTokenCount()).isEqualTo(2);
+        assertThat(tokenUsage.outputTokenCount()).isNull();
+        assertThat(tokenUsage.totalTokenCount()).isEqualTo(2);
+
+        assertThat(response.finishReason()).isNull();
+    }
```
langchain4j, line 44: Could you please add this test and make sure it is green? Thanks!
```
    @Test
    void should_embed_in_batches() {

        int batchSize = 16;
        int numberOfSegments = batchSize + 1;

        List<TextSegment> segments = new ArrayList<>();
        for (int i = 0; i < numberOfSegments; i++) {
            segments.add(TextSegment.from("text " + i));
        }

        Response<List<Embedding>> response = model.embedAll(segments);
        System.out.println(response);

        assertThat(response.content()).hasSize(numberOfSegments);
        assertThat(response.content().get(0).dimension()).isEqualTo(1536);

        TokenUsage tokenUsage = response.tokenUsage();
        assertThat(tokenUsage.inputTokenCount()).isEqualTo(numberOfSegments * 2);
        assertThat(tokenUsage.outputTokenCount()).isNull();
        assertThat(tokenUsage.totalTokenCount()).isEqualTo(numberOfSegments * 2);

        assertThat(response.finishReason()).isNull();
    }
```

ghost (commented):
@jdubois great job, thanks a lot!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: Oh you're right, let me fix this

jdubois: BTW I'm only checking the first authentication option, with "api-key", as I'm first trying to have feature parity with the existing code.
Then, as we're using the Azure SDK, I'm pretty sure that supporting the other authentication type will be easy.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: Oh indeed, that doesn't make any sense.
I'm refactoring this to use the primitive boolean type, so this won't happen at all.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: It's only logging requests, then we are already logging the responses in a System.out... I'm refactoring all this to have this correctly setup with Logback, including the responses in the tests.

Here's an example of what is being logged:
```
{"az.sdk.message":"HTTP request","method":"POST","url":"https://judubois-openai.openai.azure.com//openai/deployments/gpt-35-turbo-0613/chat/completions?api-version=2023-09-01-preview","tryCount":"1","Date":"Thu, 07 Dec 2023 09:33:20 GMT","Content-Length":"123","api-key":"REDACTED","Content-Type":"application/json","x-ms-client-request-id":"c8aa92ec-7531-4864-bfd2-6db889f7503b","accept":"application/json","User-Agent":"azsdk-java-azure-ai-openai/1.0.0-beta.5 (21.0.1; Mac OS X; 14.1.2)","contentLength":123,"body":"{\"messages\":[{\"role\":\"user\",\"content\":\"hello, how are you?\"}],\"max_tokens\":3,\"temperature\":0.7,\"model\":\"gpt-35-turbo-0613\"}"}

Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java]: I've set up Logback for having the logs correctly, and refactored the existing System.out.println to us the logger instead.
We're having indeed requests & responses, so I'm renaming the method `logRequestsAndResponses ` as proposed.

File: langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java (hunk 4 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java:68]: I've refactored the code to make it possible to set the `OpenAIClient`. It makes the code more complex, but that's indeed better for the end user.

File: langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java (hunk 5 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java]: Indeed, the `Tokenizer` is enough for configuring this. I have removed the `modelName` parameter.

File: langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java (hunk 6 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java]: Refactored the code to use the RetryOptions (using exponential backoff), and delegated the creation of the OpenAI client to a utility method.

File: langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java (hunk 7 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java]: Done

File: langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModel.java
Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModel.java:1]: Yes!

File: langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java (hunk 8 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java]: Fixed it using a try...catch block.
Unfortunately I can't do an integration test for this...

File: langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingLanguageModel.java (hunk 9 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingLanguageModel.java]: Same, done!

File: langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java (hunk 10 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java]: Indeed! Done.

File: langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAIChatModelIT.java (hunk 11 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAIChatModelIT.java]: Good catch, removed!

File: langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAIChatModelIT.java (hunk 12 above)
Comment (jdubois) [reply langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAIChatModelIT.java:130]: I'm putting:

```Java
        List<ChatMessage> chatMessages = new ArrayList<>();
        chatMessages.add(systemMessage);
        chatMessages.add(userMessage);
        chatMessages.add(toolExecutionResultMessage);
```

Is that what you mean?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: Done

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: Fixed

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: Fixed

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModel.java
jdubois: Done

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingResponseBuilder.java
jdubois: Done

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModelIT.java
jdubois: Added the test, but there's an issue, as input token count and total token count is (numberOfSegments * 3) and not (numberOfSegments * 2). Could you check?

jdubois:
Thank you so much for the detailed review @langchain4j !!! I fixed most of the comments, except 2 discussions that are still opened, could you check them?

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAIChatModelIT.java
langchain4j: No, I meant that there should be an `aiMessage` between `userMessage` and `toolExecutionResultMessage`, like so:
```
chatMessages.add(systemMessage);
chatMessages.add(userMessage);
chatMessages.add(response.content());
chatMessages.add(toolExecutionResultMessage);
```
Otherwise model will not know (explicitly) that it called a tool and with what parameters...
But this is not critical for this test, I am just overly obsessed with having production-like scenarios in tests 😆

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModelIT.java
langchain4j: My bad, it should indeed be x3, thanks!

jdubois:
Argh I'm seeing the license compliance with Logback. That's really bad, I didn't know about this! Let me check.

jdubois:
So I believe the Logback license issue is a false positive, as it is dual licensed ( https://logback.qos.ch/license.html ) and as many projects (including Spring Boot) use it by default.
Let me check if I can fix this, otherwise we'll migrate to Log4J2, as this is only for testing this isn't really important.

langchain4j:
I was not sure about using logback due to unclear license and I am no legal expert to risk it, so I just use tinylog which is apache 2:
```
            <dependency>
                <groupId>org.tinylog</groupId>
                <artifactId>tinylog-impl</artifactId>
                <scope>test</scope>
            </dependency>

            <dependency>
                <groupId>org.tinylog</groupId>
                <artifactId>slf4j-tinylog</artifactId>
                <scope>test</scope>
            </dependency>
```

But log4j also sounds good

langchain4j:
To fix it you need to add an exclusion to the license plugin, see example in `langchain4j-weaviate` pom, but this is probably a bad idea. I would better keep compliance on and use another logger

jdubois:
I migrated to Log4J2, as it's the other supported framework by Spring Boot. Anyway, it shouldn't be very important.

ghost (approved):
@jdubois awesome, thank you so much for your work and patience!

jdubois:
Wonderful, thank you so much @langchain4j !!

PR #332, opened 2023-12-08
Use Testcontainers for Weaviate IT

eddumelendez:
Run `semitechnologies/weaviate` image using `GenericContainer` and
allow rapid feeback when using Weaviate. Relax `apiKey` constraint
to allow empty value for testing purposes.


ghost (commented):
@eddumelendez thank you, this is a great idea, although I would keep the old test in place to make sure that integration with their cloud service is still working, and add a new one for the docker instance, or vice versa

ghost (approved):

PR #333, opened 2023-12-08
enable langchain4j-graal build in workflow

shalk:
langchain4j-graal  build can be enabled in workflow. 
The compile is fast. only take fews second.


ghost (approved):
@shalk thank you! It was disabled because of [the problem](https://github.com/langchain4j/langchain4j/actions/runs/6920294295/job/18824625702?pr=297) I was not able to resolve at that time.

PR #339, opened 2023-12-11
spring-boot-starter add azure openai support

QuantumBear:
Add azure support for spring-boot-starter


ghost (approved):
@QuantumBear thank you!

PR #358, opened 2023-12-15
Retriever Interface Add a default method with memoryId

leojames:
Simultaneously, **it is also necessary to modify the Retriever interface by changing it to**:

```java
default List<T> findRelevant(Object memoryId, String text);
```

This modification is needed because, in many instances, I need to identify which user executed this interface method. By passing the `memoryId`, I can transmit this information. Through this information, it becomes possible to discern the queries of different users since there might be scenarios with concurrent calls from multiple users to this method. Furthermore, each `memoryId` may correspond to multiple query conditions.


.gitignore
```diff
@@ -33,7 +34,4 @@ build/
 .vscode/
 
 ### Mac OS ###
-.DS_Store
-
-### .env files contain local environment variables ###
-.env
```
langchain4j: probably removed by accident?

langchain4j-cassandra/src/main/java/dev/langchain4j/store/embedding/cassandra/AstraDbEmbeddingStore.java
```diff
@@ -3,8 +3,13 @@
 import com.datastax.astra.sdk.AstraClient;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.dtsx.astra.sdk.cassio.MetadataVectorCassandraTable;
+import dev.langchain4j.data.embedding.Embedding;
```
langchain4j: changes in this files seem to be by accident

langchain4j-chroma/src/main/java/dev/langchain4j/store/embedding/chroma/ChromaEmbeddingStore.java
```diff
@@ -171,6 +171,13 @@ public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbeddi
                 .collect(toList());
     }
 
+    @Override
+    public List<EmbeddingMatch<TextSegment>> findRelevant(Object memoryId, String originText, Embedding referenceEmbedding, int maxResults, double minScore) {
+        return findRelevant(referenceEmbedding,maxResults,minScore);
```
langchain4j: This logic is repeating, can you put it as a default method in `EmbeddingStore` interface?

langchain4j-core/src/main/java/dev/langchain4j/data/message/ChatMessageType.java
```diff
@@ -7,7 +7,8 @@ public enum ChatMessageType {
     SYSTEM,
     USER,
     AI,
-    TOOL_EXECUTION_RESULT;
+    TOOL_EXECUTION_RESULT,
+    Relevant;
```
langchain4j: What's up with the case? Also, it is not clear why is there a need for a new type of message.

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -79,4 +79,31 @@ default List<EmbeddingMatch<Embedded>> findRelevant(Embedding referenceEmbedding
...
+     * Finds the most relevant (closest in space) embeddings to the provided reference embedding.
+     * By default, minScore is set to 0, which means that the results may include embeddings with low relevance.
+     * @param originText the origin text of user input
+     * @param referenceEmbedding The embedding used as a reference. Returned embeddings should be relevant (closest) to this one.
+     * @param maxResults         The maximum number of embeddings to be returned.
+     * @return A list of embedding matches.
+     * Each embedding match includes a relevance score (derivative of cosine distance),
+     * ranging from 0 (not relevant) to 1 (highly relevant).
+     */
+    default List<EmbeddingMatch<Embedded>> findRelevant(Object memoryId,String originText,Embedding referenceEmbedding, int maxResults) {
```
langchain4j: why do you need to pass `originText` into `EmbeddingStore`?

langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchEmbeddingStore.java
```diff
@@ -259,6 +259,13 @@ public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbeddi
         }
     }
 
+    @Override
+    public List<EmbeddingMatch<TextSegment>> findRelevant(Object memoryId, String originText, Embedding referenceEmbedding, int maxResults, double minScore) {
+        return findRelevant(referenceEmbedding, maxResults, minScore);
```
langchain4j: again, this can be a default method in `EmbeddingStore`

langchain4j/src/main/java/dev/langchain4j/chain/ConversationalRetrievalChain.java
```diff
@@ -52,7 +52,7 @@ public String execute(String question) {
 
         question = ensureNotBlank(question, "question");
 
-        List<TextSegment> relevantSegments = retriever.findRelevant(question);
+        List<TextSegment> relevantSegments = retriever.findRelevant("default",question);
```
langchain4j: why is this hardcoded?

langchain4j-core/src/main/java/dev/langchain4j/data/message/RelevantAiMessage.java
```diff
@@ -0,0 +1,67 @@
+package dev.langchain4j.data.message;
+
+import java.util.Objects;
+
+import static dev.langchain4j.data.message.ChatMessageType.Relevant;
+import static dev.langchain4j.internal.Utils.quoted;
+
+public class RelevantAiMessage extends ChatMessage {
```
langchain4j: What is the difference between `RelevantMessage` and `RelevantAiMessage`?

ghost (changes requested):
@leojames thanks for your contribution! I am not sure I understand the point of this PR, would be nice if you could provide more info. I see that you've introduced a new type of `ChatMessage` (actually 2) and changed interfaces of `Retriever` and `EmbeddingStore`, but I do not see where these new things are used. Please explain the need for new message type and new methods on interfaces, as it does not seem to be related to the description of this PR. Thank you.

langchain4j-cassandra/src/main/java/dev/langchain4j/store/embedding/cassandra/AstraDbEmbeddingStore.java
leojames: useless

langchain4j-chroma/src/main/java/dev/langchain4j/store/embedding/chroma/ChromaEmbeddingStore.java
leojames: yes  i would add a default interface in EmbeddingStore

langchain4j-core/src/main/java/dev/langchain4j/data/message/ChatMessageType.java
leojames: Sometimes ,such as , the system recomond some infos ,but the next infos is unrelated with last infomation ,so i would delete last information , otherwise ,the size of LLM is limited. when i use last 20 messages ,i found that the size not enough .so i identify the relavant messge

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
leojames: it is useless ,if i have memoryId,i will remove

langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchEmbeddingStore.java
leojames: yes make it default

langchain4j/src/main/java/dev/langchain4j/chain/ConversationalRetrievalChain.java
leojames: The reason I hardcoded this part is that we need a default message queue.

langchain4j-core/src/main/java/dev/langchain4j/data/message/RelevantAiMessage.java
leojames: RelevantAiMessage is useless  i will remove

leojames:
This time, I only modified the Retriever interface by adding a default implementation. A section of the EmbeddingStore interface also saw the addition of a default implementation. The purpose of this is to pass the "MemoryId" field, allowing users to customize the Chain and utilize "MemoryId" to distinguish between different users' requests. This is crucial because it's not just the EmbeddingText field that needs to be queried, and different users may have additional query conditions.

langchain4j-core/src/main/java/dev/langchain4j/retriever/Retriever.java
```diff
@@ -5,4 +5,8 @@
 public interface Retriever<T> {
 
     List<T> findRelevant(String text);
+
+    default List<T> findRelevant(Object memoryId,String text){
+       return findRelevant(text);
```
langchain4j: Please throw an exception here: `throw new RuntimeException("Not implemented");`

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -79,4 +79,33 @@ default List<EmbeddingMatch<Embedded>> findRelevant(Embedding referenceEmbedding
...
+     * By default, minScore is set to 0, which means that the results may include embeddings with low relevance.
+     * @param memoryId           The memoryId used Distinguishing query requests from different users.
+     * @param referenceEmbedding The embedding used as a reference. Returned embeddings should be relevant (closest) to this one.
+     * @param maxResults         The maximum number of embeddings to be returned.
+     * @return A list of embedding matches.
+     * Each embedding match includes a relevance score (derivative of cosine distance),
+     * ranging from 0 (not relevant) to 1 (highly relevant).
+     */
+    default List<EmbeddingMatch<Embedded>> findRelevant(Object memoryId,Embedding referenceEmbedding, int maxResults) {
+        return findRelevant(memoryId,referenceEmbedding, maxResults, 0);
```
langchain4j: Please throw an exception here: `throw new RuntimeException("Not implemented");`

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -79,4 +79,33 @@ default List<EmbeddingMatch<Embedded>> findRelevant(Embedding referenceEmbedding
...
+     * @param referenceEmbedding The embedding used as a reference. Returned embeddings should be relevant (closest) to this one.
+     * @param maxResults         The maximum number of embeddings to be returned.
+     * @param minScore           The minimum relevance score, ranging from 0 to 1 (inclusive).
+     *                           Only embeddings with a score of this value or higher will be returned.
+     * @return A list of embedding matches.
+     * Each embedding match includes a relevance score (derivative of cosine distance),
+     * ranging from 0 (not relevant) to 1 (highly relevant).
+     */
+     default List<EmbeddingMatch<Embedded>> findRelevant(Object memoryId,Embedding referenceEmbedding, int maxResults, double minScore){
+         return findRelevant(referenceEmbedding, maxResults,minScore);
```
langchain4j: Please throw an exception here: `throw new RuntimeException("Not implemented");`

ghost (changes requested):

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
leojames: OK

ghost (approved):
@leojames thank you!

PR #359, opened 2023-12-15
Add image generation support with Azure OpenAI

jdubois:
This uses the same LangChain classes as https://github.com/langchain4j/langchain4j/pull/298 by @Heezer 

- langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
- langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java

I copy/pasted those 2 files here as #298 isn't merged yet, and my goal is that we use the same API in the end.


Heezer:
Thanks! There are some refactorings of #298 on the way, the PR is close to be approved. I'll keep you posted!

jdubois:
Awesome! I got access to Dall-e 3 this morning, so I'll do integration tests.

jdubois:
And integration tests are passing!

Here's the first ever image created with the integration test `AzureOpenAiImageModelIT` and it's beautiful:

generated_00

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer, line 1: please update, this file has been updated in #298

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
Heezer, line 1: please update, this file has been updated in #298

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -187,6 +188,12 @@ public static AiMessage aiMessageFrom(com.azure.ai.openai.models.ChatResponseMes
         }
     }
 
+    public static Image imageFrom(com.azure.ai.openai.models.ImageGenerationData imageGenerationData) {
+        return Image.builder()
+                .url(imageGenerationData.getUrl())
```
Heezer: please support `revised_prompt` too, it is inside of `Image` now

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,254 @@
...
+                .setN(1)
+                .setQuality(quality)
+                .setSize(size)
+                .setUser(user)
+                .setStyle(style)
+                .setResponseFormat(responseFormat);
+
+        ImageGenerations imageGenerations = client.getImageGenerations(deploymentName, options);
+
+        return Response.from(imageFrom(imageGenerations.getData().get(0)));
```
Heezer: please support multiple images too

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
```
Heezer, line 117: please support multiple images too - parameter `n`

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
Heezer: if `response_format` is `b64_json` the image is within `base64Data` field instead of `url`, it shall be supported here too

Heezer (changes requested):
Thanks a lot! See my comments.

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -0,0 +1,36 @@
+package dev.langchain4j.data.image;
+
+public class Image {
+
+    private String url;
```
glaforge: What if an image is represented as a Base64 encoded string?
For example, in Gemini, an image is either passed as a Base64 encoded string or as a URI to a google cloud storage bucket.
I fear this Image class is not generic enough to work with other models that support images.

Heezer: My intention was to keep `Image` ready-to-use for the client as much as possible. This means if Base64 is configured, then the image(s) download happens under the hood. In the end the client gets the URL to local file here. Check #298, there is an inte test too.

jdubois:
Thanks @Heezer !! This is fixing my 2 biggest issues with this API (not limiting to just 1 image, and being able to get the image content and not just the URL)

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiImageModelIT.java
```diff
@@ -20,11 +23,15 @@ void should_generate_image() {
...
+                .setN(1)
                 .build();
 
-        Response<Image> response = model.generate("A coffee mug in Paris, France");
+        Response<List<Image>> response = model.generate("A coffee mug in Paris, France");
 
         logger.info(response.toString());
-        assertThat(response.content().url()).isNotBlank();
+
+        Image image = response.content().get(0);
```
Heezer: nice! you can assert the list size additionally:
`assertThat(response.content()).hasSize(1);`

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
langchain4j: I agree with @glaforge, we should add base64 content field to `Image` to make it more versatile. We should also not force user to use our built-in persisting logic, it should be an optional "nice to have".

jdubois: I'm currently coding this part, and indeed I agree too with @glaforge - it feels strange to write the image locally, for me it should be passed in the base64 field instead.

glaforge: Maybe also adding the mime type, to accomodate PNG, JPEG, etc.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
jdubois: this supports base64 and url now

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
jdubois: The way I coded it now:
- only 1 image as Dall-e 3 only supports 1
- base64 data is stored in the variable, and not written as a temp file on disk
- I haven't put the mime type as only PNG is supported at the moment

This needs to be discussed and aligned with the work on OpenAI

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,257 @@
...
+                                ProxyOptions proxyOptions,
+                                boolean logRequestsAndResponses) {
+
+        this(deploymentName, quality, size, user, style, responseFormat);
+        this.client = setupOpenAIClient(endpoint, serviceVersion, apiKey, timeout, maxRetries, proxyOptions, logRequestsAndResponses);
+    }
+
+    private AzureOpenAiImageModel(String deploymentName, String quality, String size, String user, String style, String responseFormat) {
+        this.deploymentName = getOrDefault(deploymentName, "dall-e-3");
+        this.quality = getOrDefault(ImageGenerationQuality.fromString(quality), ImageGenerationQuality.STANDARD);
```
langchain4j: I would keep all the optional parameters (all except deployment name) without defaults on our side, let it be decided by MS/OpenAI

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,257 @@
...
+import dev.langchain4j.model.output.Response;
+
+import java.nio.file.Path;
+import java.time.Duration;
+
+import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.model.azure.InternalAzureOpenAiHelper.imageFrom;
+import static dev.langchain4j.model.azure.InternalAzureOpenAiHelper.setupOpenAIClient;
+
+public class AzureOpenAiImageModel implements ImageModel {
```
langchain4j, line 38: Could you please add a link to the documentation in the javadoc?
Did not find one from MS, this is what OpenAI has: https://platform.openai.com/docs/api-reference/images/create

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -130,7 +129,6 @@ void should_call_function_with_argument() {
         List<ChatMessage> chatMessages = new ArrayList<>();
         chatMessages.add(systemMessage);
         chatMessages.add(userMessage);
-        chatMessages.add(aiMessage);
```
langchain4j: was it removed on purpose?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,257 @@
...
+        private String quality;
+        private String size;
+        private String user;
+        private String style;
+        private String responseFormat;
+        private Duration timeout;
+        private Integer maxRetries;
+        private ProxyOptions proxyOptions;
+        private boolean logRequestsAndResponses;
+        private boolean withPersisting;
```
langchain4j: not used

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,257 @@
...
+
+        private Path persistTo;
```
langchain4j: not used

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,257 @@
...
+            return this;
+        }
+
+        /**
+         * Sets the quality of the image. This is an optional parameter.
+         *
+         * @param quality The quality of the image.
+         * @return builder
+         */
+        public Builder quality(String quality) {
```
langchain4j, line 184: I would add an overloaded method that accepts an enum

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,257 @@
...
+            return this;
+        }
+
+        /**
+         * Sets the size of the image. This is an optional parameter.
+         *
+         * @param size The size of the image.
+         * @return builder
+         */
+        public Builder size(String size) {
```
langchain4j, line 206: I would add an overloaded method that accepts an enum

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,257 @@
...
+            return this;
+        }
+
+        /**
+         * Sets the style of the image. This is an optional parameter.
+         *
+         * @param style The style of the image.
+         * @return builder
+         */
+        public Builder style(String style) {
```
langchain4j, line 239: I would add an overloaded method that accepts an enum

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
```diff
@@ -0,0 +1,257 @@
...
+            return this;
+        }
+
+        /**
+         * Sets the response format of the image. This is an optional parameter.
+         *
+         * @param responseFormat The response format of the image.
+         * @return builder
+         */
+        public Builder responseFormat(String responseFormat) {
```
langchain4j, line 261: I would add an overloaded method that accepts an enum

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(url, base64Data, revisedPrompt);
+    }
+
+    @Override
+    public String toString() {
+        return "Image{" + "url='" + url + '\'' +
```
langchain4j: minor: there is `Utils.quoted()`

ghost (commented):
@jdubois great job, thanks you! Just a few minor things, please check comments

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiImageModel.java
jdubois: Just to be clear: I'm letting all those parameters be nullable, so if they are not set you will get the default value from MS/OpenAI

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
Heezer: I'm fine with adding base64 field to `Image`, sure.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
jdubois: Argh I shouldn't have mixed this into this PR!
So yes: I got an error here, and I realised that `toolExecutionResultMessage` gets the result of the `aiMessage`, so I don't think we need to add the `aiMessage` here. I'm not an expert of this part, WDYT?

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
jdubois: I called it `base64Data` as it's the name used in Azure OpenAI and that allows to have a `base64MimeType` in the future to do like @glaforge suggested.

jdubois: I think we're good for the moment, so marking this as resolved

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
langchain4j: I have previously added `aiMessage` here on purpose. `toolExecutionResultMessage` is the response to `aiMessage`. In the real-world scenario `aiMessage` should be present, otherwise the model can get confused. Moreover, in the newer OpenAI "tools" api (which deprecates "functions" api, search for "tool" and "function" [here](https://platform.openai.com/docs/api-reference/chat/create)) each `toolExecutionResultMessage` should correspond to a tool call present in `aiMessage`. If you don't supply `aiMessage`, you will get an error like this:
```
{
    "error": {
        "message": "Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.",
        "param": "messages.[0].role",
        "code": null
    }
}
```

langchain4j:
@jdubois looks good! There is now a conflict due to changes by @Heezer merged to main. Once this is fixed, we can merge 🚀

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
jdubois: Thank you, I really messed up on this one! Let me check and try to understand the issue.

jdubois: So, the issue is that AiMessage can't have a text that is "null" now (the integration test will fail because the field is empty), so I'm forcing this to "" in https://github.com/jdubois/langchain4j/commit/1b97059c359dddb932f15ced0a76e9384b899ea0 - @langchain4j I believe this is correct, WDYT?

jdubois (commented):
All comments from the review should be OK

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
langchain4j: @jdubois sorry, I am not sure I understand the problem. `AiMessage.text == null` is ok for the case where `AiMessage.toolExecutionRequests != null`. AiMessage can contain either `text` or `toolExecutionRequests`, but not both. There should be no empty `AiMessage.text`. I just did a [change](https://github.com/langchain4j/langchain4j/pull/370/files) to reflect these constraints, please check.

langchain4j: My bad, the constructors in `AiMessage` were really confusing, hopefully now it is better. I was not paying much attention to ctors as I usually use static factory methods, there it was quite clear which inputs are required to create `AiMessage`

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -129,8 +128,8 @@ void should_call_function_with_argument() {
 
         List<ChatMessage> chatMessages = new ArrayList<>();
         chatMessages.add(systemMessage);
-        chatMessages.add(userMessage);
         chatMessages.add(aiMessage);
+        chatMessages.add(userMessage);
```
langchain4j: now the ordering is wrong... it should be `system -> user -> ai -> tool`

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
jdubois: @langchain4j so I migrated to it in https://github.com/langchain4j/langchain4j/pull/359/commits/39207e89d0768bf59b82d8b064be600fe7a5b91f and that compiles, but the integration tests won't work.

I'm getting:
```json
com.azure.core.exception.HttpResponseException: Status code 400, "{
  "error": {
    "message": "'content' is a required property - 'messages.2'",
    "param": null,
    "code": null
  }
}
```

This is because the text is `null` and not empty. I agree this looks like an Azure OpenAI SDK issue, then the only way to solve this currently is to pass an empty String.

langchain4j: Ah I see... Can we adapt mapping logic from langchain4j to azure messages and insert empty string if aiMessage.text is null?

jdubois: So I created a specific AiMessage implementation for Azure OpenAI in https://github.com/langchain4j/langchain4j/pull/359/commits/5ffc013d0e1b707097b2f8401a35bc1c2cff4ce5 - I'm not a huge fan of this code, but it does the job and doesn't pollute the core LangChain4J implementation, so I guess it's the best option.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiImageModelIT.java
```diff
@@ -0,0 +1,75 @@
...
+        Response<Image> response = model.generate("A coffee mug in Paris, France");
+
+        logger.info(response.toString());
+
+        Image image = response.content();
+        assertNotNull(image);
+        assertNotNull(image.url());
+        logger.info("The remote image is here: {}", image.url());
+
+        assertNull(image.base64());
```
Heezer: Feel free to rename `base64` (I've named it like this 😞) into `base64Data` in your branch, which is better name

jdubois: Oh, thanks! OK I'll do this.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
langchain4j: Yeah me too... this can be simplified to this change in `InternalAzureOpenAiHelper.toOpenAiMessage()`:
`new ChatRequestAssistantMessage(getOrDefault(message.text(), ""));` and you don't need a whole new class for this

jdubois: Thanks, done!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureAiMessage.java
```diff
@@ -0,0 +1,27 @@
+package dev.langchain4j.model.azure;
+
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.data.message.AiMessage;
+
+import java.util.List;
+
+import static java.util.Arrays.asList;
+
+public class AzureAiMessage extends AiMessage {
```
langchain4j: This class is not used

ghost (dismissed):
@jdubois just need to remove unused `AzureAiMessage` and we're good

ghost (approved):
@jdubois thank you, great job!

jdubois:
Awesome!! Thanks @langchain4j @Heezer

PR #361, opened 2023-12-15
Extracted OpenAI Spring Boot starter into a separate module

ghost:


langchain4j:
@jdubois @Heezer do you mind taking a quick look at this PR? I want to make sure I did not miss anything before I start extracting starters for other model providers. Thanks!

jdubois:
@langchain4j the first thing I would do is upgrade to the latest Spring Boot 3.2.0 release, as the 2.7.x branch is end of life now. I can have a look if you want!

langchain4j:
@jdubois that would require minimum java 17... Please let's not start this discussion right now 😅

langchain4j:
I've just tested this dependency on Spring 3.2.0 project, works good. It does not pull any 2.X dependencies, so no worries here.

jdubois:
Ah ah OK, but we need to check if this will work fine with Spring Boot 3.2

jdubois:
Oh same comment at the same time :-)

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/AutoConfig.java
```diff
@@ -0,0 +1,136 @@
...
+@EnableConfigurationProperties(Properties.class)
+public class AutoConfig {
+
+    private static final String MISSING_CONFIG_ERROR = "You are requesting an instance of an %s, " +
+            "but no '%s.%s' configuration properties are provided.";
+    private static final String MISSING_API_KEY_ERROR = "Please provide an OpenAI API key through " +
+            "the '%s.%s.api-key' property.";
+
+    @Bean
+    @Lazy
```
jdubois: I would not make those beans Lazy by default

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-open-ai</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter</artifactId>
+        </dependency>
+
```
jdubois: You should add `spring-boot-autoconfigure-processor` here, have a look at https://docs.spring.io/spring-boot/docs/3.2.0/reference/htmlsingle/#features.developing-auto-configuration.custom-starter.autoconfigure-module

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/AutoConfig.java
```diff
@@ -0,0 +1,136 @@
...
+    @ConditionalOnMissingBean
```
jdubois: A `@ConditionalOnProperty` would be easier to configure when you'll have several beans (I'm guessing you'll have them all in the same package)

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/AutoConfig.java
```diff
@@ -0,0 +1,136 @@
...
+                .frequencyPenalty(chatModelProperties.getFrequencyPenalty())
+                .timeout(chatModelProperties.getTimeout())
+                .maxRetries(chatModelProperties.getMaxRetries())
+                .logRequests(chatModelProperties.getLogRequests())
+                .logResponses(chatModelProperties.getLogResponses())
+                .build();
+    }
+
+    @Bean
+    @Lazy
```
jdubois: Same as above, I don't think Lazy is good here

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/AutoConfig.java
```diff
@@ -0,0 +1,136 @@
...
+                .temperature(languageModelProperties.getTemperature())
+                .timeout(languageModelProperties.getTimeout())
+                .maxRetries(languageModelProperties.getMaxRetries())
+                .logRequests(languageModelProperties.getLogRequests())
+                .logResponses(languageModelProperties.getLogResponses())
+                .build();
+    }
+
+    @Bean
+    @Lazy
```
jdubois: Same as above, I don't think Lazy is good here

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/AutoConfig.java
```diff
@@ -0,0 +1,136 @@
...
+                .modelName(embeddingModelProperties.getModelName())
+                .timeout(embeddingModelProperties.getTimeout())
+                .maxRetries(embeddingModelProperties.getMaxRetries())
+                .logRequests(embeddingModelProperties.getLogRequests())
+                .logResponses(embeddingModelProperties.getLogResponses())
+                .build();
+    }
+
+    @Bean
+    @Lazy
```
jdubois: Same as above, I don't think Lazy is good here

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/Properties.java
```diff
@@ -0,0 +1,26 @@
...
+
+    @NestedConfigurationProperty
+    LanguageModelProperties languageModel;
+
+    @NestedConfigurationProperty
+    EmbeddingModelProperties embeddingModel;
+
+    @NestedConfigurationProperty
+    ModerationModelProperties moderationModel;
+}
```
jdubois: Don't forget to add the new "Image" support that was merged today

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/test/java/dev/langchain4j/openai/spring/AutoConfigIT.java
```diff
@@ -0,0 +1,176 @@
...
+import org.springframework.boot.autoconfigure.AutoConfigurations;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import static dev.langchain4j.model.openai.OpenAiModelName.GPT_3_5_TURBO;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+@EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = "sk.*")
```
jdubois: What is this "sk" prefix? If this comes from Semantic Kernel, I would wait a bit before copying them as their Spring Boot starter isn't finished yet

jdubois (commented):
I'm lacking time to due a full review, sorry if this isn't complete yet

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/test/java/dev/langchain4j/openai/spring/AutoConfigIT.java
langchain4j: No, this is an OpenAI's API key prefix, I guess it stands for secret key

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/Properties.java
langchain4j: Good catch, thanks!

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/AutoConfig.java
jdubois: Sorry the ConditionnalBean is what they are using in Spring AI (see https://github.com/spring-projects/spring-ai/blob/1a8fcac206070990f312986f34649bda9c6ac003/spring-ai-spring-boot-autoconfigure/src/main/java/org/springframework/ai/autoconfigure/azure/openai/AzureOpenAiAutoConfiguration.java#L43C3-L43C3 ), the trick is that they split their code in one starter per technology, and that's more work but that's probably the best idea

langchain4j: I am splitting into one autoconfigure/starter per provider. I see that they mix all the autoconfigurations in one module, then have starters per each providers that references autoconfigure module. Not sure what's the point?

jdubois: We're looking at the same code with @SandraAhlgrimm :-)
Not sure why they did it, then yes it should be separated in one starter per provider.

langchain4j: Hi @SandraAhlgrimm :)
I am changing `@ConditionalOnMissingBean` now to `@ConditionalOnProperty(PREFIX + ".chat-model.api-key")`

jdubois: No I would keep `@ConditionalOnMissingBean` now, so that people can create their own if they want to. Then if you split starters by providers, it's as good as a missing property.

langchain4j: `@ConditionalOnProperty(PREFIX + ".chat-model.api-key")` simplifies a bit, I don't need to check if user provided properties or not. User still can provide his own `@Bean`

langchain4j: Not sure I fully understand your point. Let's say user added `langchain4j-open-ai-spring-boot-starter` to their dependencies. If he/she does not provide api key, bean is not created automatically. If they provide API key, it will be created automatically. If they wish to configure their own, they remove properties and provide their own `@Bean`. Seems good, no?

jdubois: Yes that works too. That's what's happening also quite frequently with some Spring Boot starters when you need to add a key like `.enabled=true` (for example `management.endpoint.info.enabled=true`).
But with `@ConditionalOnMissingBean` it works by magic (no key), as adding the starter is enough to tell you want that feature. And then if you want to create your own bean (because there's a feature you need that's not available in the starter), you can then re-use the keys, including the API key one.

langchain4j: And if user defines both properties and his own `@Bean`, it should fail to avoid ambiguity because there are now 2 beans. This is not possible with `@ConditionalOnMissingBean`

langchain4j: Every starter `@Bean` that we will provide will have at least one mandatory property (e.g. API key for managed services and base URL for local deployments), so user won't be able to just add a starter dependency and get a bean automagically, they will have to define at least one property. So I guess we can use the presence of this property to judge if the bean should be created automatically or not.

I could use `@ConditionalOnMissingBean`, but then it sohuld be either `@Lazy` or `@ConditionalOnProperty`. `langchain4j-open-ai-spring-boot-starter` provides 5 beans, if they will be just `@ConditionalOnMissingBean`, Spring Boot will attempt to create all of them and will fail because they have mandatory properties (API key).

spring-boot-starters/langchain4j-open-ai-spring-boot-starter/pom.xml
langchain4j: Adding this dependency resulted in generating `spring-autoconfigure-metadata.properties` file with this content:
`dev.langchain4j.openai.spring.AutoConfig=`, not sure how does it help? This info is already present in `org.springframework.boot.autoconfigure.AutoConfiguration.imports` and `spring.factories` files...

langchain4j:
@jdubois I have pushed updates that address your comments, could you please check? Thanks!

One more thing: I anticipate that it will be common to use multiple models (with different configs) for different purposes in the same app, so langchain4j should support that. But I am not sure how to better implement this in auto configuration. Any ideas?

maxandersen:
is examples and starters not better put in the examples repo or separate module ? otherwise depmanagement in the pom and bom's now drag in a tail of dependencies that is not actually needed for langchain4j.

langchain4j:
@maxandersen I am not sure I understand the problem. Spring Boot starters will reside in their own maven modules/artifacts, so there should be no problem. BOM pulls only versions of langchain4j modules, so it makes sense to include all related dependencies there (that's the point of the BOM, right?). Another pom that is referencing spring boot starters is `langchain4j-aggregator`, which noone should use. Could you please elaborate? Thanks!

maxandersen:
Reasons:

- any dependency in your bom/pom risk being analyzed by tools doing cve anytics and Langchain4j will be unnecessary marked as having issues when it doesn't. Same reason we don't push for putting quarkus extension/samples into this repo. 

- if it is separate you could use spring 3.x and have java 17 examples instead of limit to java 8/11

langchain4j:
@maxandersen good point, I have [moved](https://github.com/langchain4j/langchain4j-spring/pull/1) it to https://github.com/langchain4j/langchain4j-spring

PR #368, opened 2023-12-19
Fix Neo4jEmbeddingStoreTest error

vga91:
See https://github.com/langchain4j/langchain4j/actions/runs/7259989386/job/19778223144

Changed to `MATCH (n:%s) RETURN n ORDER BY n.text` to make tests more deterministic,
since rarely the sorting of "MATCH (n:%s) RETURN n" is not based on the order of arrival of the `embeddingModel.embed(..)`.


ghost (approved):
@vga91 thank you!

PR #390, opened 2023-12-21
ToolSpecification and ToolSpecifications docs and tests.

crutcher:
Add ToolSpecificationTest.
Add ToolSpecificationsTest.

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
@@ -8,7 +8,6 @@ import static dev.langchain4j.internal.Utils.quoted;
 import static java.util.Arrays.asList;
 
 public class ToolSpecification {
-
     private final String name;
     private final String description;
     private final ToolParameters parameters;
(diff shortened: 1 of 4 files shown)


crutcher:
By inspection, this error *Should Not Happen*;

- that arg has a `@P()` annotation, and should get a "description" property.
- this is what happens on local unittests (in IntelliJ)

I'm concerned there's some form of annotation stripping. It could also be some form of masking?

```
Error:  Failures: 
Error:    ToolSpecificationsTest.test_toJsonSchemaProperties:190 
Expecting actual:
  [JsonSchemaProperty { key = "type", value = integer }]
to contain exactly (and in same order):
  [JsonSchemaProperty { key = "type", value = string },
    JsonSchemaProperty { key = "description", value = foo }]
but some elements were not found:
  [JsonSchemaProperty { key = "type", value = string },
    JsonSchemaProperty { key = "description", value = foo }]
and others were not expected:
  [JsonSchemaProperty { key = "type", value = integer }]
```

crutcher:
found the issue, which was implementation-dependent ordering of the getDeclaredMethods() array; worked around that.

Class.getDeclaredMethods()
> The elements in the returned array are not sorted and are not in any particular order.

ghost (approved):

PR #393, opened 2023-12-21
InMemoryChatMemoryStore, tests

crutcher:
Add InMemoryChatMemoryStoreTest with full coverage. Make InMemoryChatMemoryStore synchronized on its messages for thread protection.


langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/InMemoryChatMemoryStore.java
```diff
@@ -13,20 +13,22 @@
  * This storage mechanism is transient and does not persist data across application restarts.
  */
 public class InMemoryChatMemoryStore implements ChatMemoryStore {
-
     private final Map<Object, List<ChatMessage>> messagesByMemoryId = new ConcurrentHashMap<>();
 
     @Override
+    synchronized
```
oujingzhou: messagesByMemoryId is ConcurrentHashMap, and it is thread-safe, so there is no need to make method synchronized

crutcher: done

ghost (approved):
@crutcher thank you!

PR #402, opened 2023-12-21
Support Google Vertex AI Gemini

ghost:
This PR adds a basic support of Gemini (text).
Next time images and tools will be added.


langchain4j:
@glaforge could you please take a look? Thanks!

langchain4j-vertex-ai-gemini/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-vertexai</artifactId>
+            <version>0.1.0</version>
```
glaforge: In our documentation, FYI, we usually suggest to use the google cloud libraries BOM, instead of specifying the version of the artifact:

```xml
<dependencyManagement>
    <dependencies>
        <dependency>
            <artifactId>libraries-bom</artifactId>
            <groupId>com.google.cloud</groupId>
            <scope>import</scope>
            <type>pom</type>
            <version>26.29.0</version>
        </dependency>
    </dependencies>
</dependencyManagement>

<dependencies>
    <dependency>
        <groupId>com.google.cloud</groupId>
        <artifactId>google-cloud-vertexai</artifactId>
    </dependency>
    ...
</dependencies>
```

I guess it's maybe better in case we're importing various other google cloud related libraries, so that there's no version conflict across libraries and their transitive dependencies. But in this case here for this module, I guess this is fine to use the version explicitly.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/StreamingChatResponseBuilder.java
```diff
@@ -0,0 +1,51 @@
...
+            return;
+        }
+
+        List<Candidate> candidates = partialResponse.getCandidatesList();
+        if (candidates.isEmpty() || candidates.get(0) == null) {
+            return;
+        }
+
+        String content = ResponseHandler.getText(partialResponse);
+        if (content != null) {
```
glaforge: `ResponseHandler#getText()` either throws an exception or returns a non empty string.
It's always safe to check against `null`, but maybe we could check the length is not empty?
On the other hand, appending an empty string won't change anything anyway :-)

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
```diff
@@ -0,0 +1,82 @@
...
+    public VertexAiGeminiChatModel(String project,
+                                   String location,
+                                   String modelName,
+                                   Float temperature,
+                                   Integer maxOutputTokens,
+                                   Integer topK,
+                                   Float topP,
+                                   Integer maxRetries) {
+
+        GenerationConfig.Builder generationConfigBuilder = GenerationConfig.newBuilder();
```
glaforge, line 52: Should we check the ranges of the values below?
Or we just expect the underlying sdk to take care of checking them?
ie. topK is 1..40, topP is 0..1, etc.
https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
```diff
@@ -0,0 +1,82 @@
...
+        if (topP != null) {
+            generationConfigBuilder.setTopP(topP);
+        }
+        this.generationConfig = generationConfigBuilder.build();
+
+        try (VertexAI vertexAI = new VertexAI(
+                ensureNotBlank(project, "project"),
+                ensureNotBlank(location, "location"))
+        ) {
+            this.model = new GenerativeModel(ensureNotBlank(modelName, "modelName"), vertexAI);
```
glaforge: Another thing we can configure (later on, in a future version) is the ability to filter/moderate content. There's the notion of `SafetySettings` in Gemini, which looks like so:

```java
      GenerativeModel model = new GenerativeModel(modelName, generationConfig, vertexAI);

      List<SafetySetting> safetySettings = Arrays.asList(
          SafetySetting.newBuilder()
              .setCategory(HarmCategory.HARM_CATEGORY_HATE_SPEECH)
              .setThreshold(SafetySetting.HarmBlockThreshold.BLOCK_LOW_AND_ABOVE)
              .build(),
          SafetySetting.newBuilder()
              .setCategory(HarmCategory.HARM_CATEGORY_DANGEROUS_CONTENT)
              .setThreshold(SafetySetting.HarmBlockThreshold.BLOCK_MEDIUM_AND_ABOVE)
              .build()
      );
```

It's a bit verbose though, but I'm wondering if there's something more generic that could be used across language models?
I think the OpenAI model has a simple moderation flag, for example? Here it's more fine-grained, so not sure how to make something really generic.

glaforge (approved):
This looks excellent!
I've tested it with some simple examples of chat conversations, and even data extraction with `AiServices`, and everything works great so far! Congrats!
I added a few comments, but it's more open questions or suggestions, nothing that necessarily need a reply or changes to the code.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/RoleMapper.java
```diff
@@ -0,0 +1,17 @@
+package dev.langchain4j.model.vertexai;
+
+import dev.langchain4j.data.message.ChatMessageType;
+
+class RoleMapper {
+
+    static String map(ChatMessageType type) {
+        switch (type) {
+            case SYSTEM:
```
oujingzhou, line 9: According to Gemini API: https://cloud.google.com/vertex-ai/docs/generative-ai/model-reference/gemini
USER: Specifies content that's sent by you.
MODEL: Specifies the model's response.

Maybe SYSTEM should map to "model"?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -0,0 +1,74 @@
...
+    void should_generate_response() {
+
+        // given
+        ChatLanguageModel model = VertexAiGeminiChatModel.builder()
+                .project("langchain4j")
+                .location("us-central1")
+                .modelName("gemini-pro")
+                .build();
+
+        UserMessage userMessage = UserMessage.from("What is the capital of Germany?");
```
King-Chau, line 28: Can you add tests for SystemMessage and AiMessage?

oujingzhou, line 35: Can you add tests for SystemMessage request?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/RoleMapper.java
langchain4j: We have agreed with @glaforge to deny system messages alltogether for the moment as they are not supported by Gemini (yet?)

langchain4j-vertex-ai-gemini/pom.xml
langchain4j: good point, will change

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/StreamingChatResponseBuilder.java
langchain4j: good catch!

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
langchain4j: Yes, we expect underlying SDK to do validations.
Otherwise we will be buried by all sorts of validations and will need to keep track of all the updates to them...

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/RoleMapper.java
glaforge: As @langchain4j says, we think it's better to disallow system messages as they are not supported by Gemini.
Furthermore, if you pass a "model" message as a first message, then follow up with "user"'s messages, it'll complain anyway that it's not the right user/model alternation of messages.
So it has to be user/model/user/model/etc.
If you want to mimic system messages, the first user message should start with the system context: "You're an expert chatbot in blahblah/n/nHow do I do this?"

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
langchain4j: OpenAI has a separate [moderation endpoint](https://platform.openai.com/docs/guides/moderation/quickstart), did not see an option to set moderation policy right in the completion endpoint. But it is a good idea, I will add an option to ctor to configure this

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
langchain4j: we will deny system message for now, will add a test for this

langchain4j:
@glaforge thanks a lot for the review! I have addressed your comments. I have also added ctors that accept `GenerativeModel` and `GenerationConfig` for more flexibility. User can also setup `SafetySetting` when creating `GenerativeModel`.

glaforge (approved):
Fantastic!

PR #408, opened 2023-12-22
Replace Mustache templates with simple implementation to reduce transitive dependencies

ghost:
One can provide their own prompt template implementation via SPI if needed.


Willam2004:
Mustache templates  support simple condition template render, but  simple implementation is not support. e.g
:  {{#schema}}  output text {{/schema}}

langchain4j:
Hi @Willam2004, this is a trade-off we had to make.
We assume that most users do not need advanced prompt templating features like loops and conditions, thus we removed mustache as a dependency from `langchain4j-core` which will benefit most of the users.

Users that need advanced prompt templating features now have such options:
1. Build `String`s outside of templating engine (e.g. using simple java `for`/`while` loops or `if`/`else` statements)
2. Ignore `PromptTemplate` in LangChain4j and use templating engine of their choice and provide `String`s to LangChain4j APIs
3. Implement `dev.langchain4j.spi.prompt.PromptTemplateFactory` interface with the templating engine of their choice and continue using `PromptTemplate`

We could probably provide 3 as a separate module (e.g. `langchain4j-mustache`), but this is not a priority at the moment.
Thank you for your understanding.

PR #410, opened 2023-12-22
Maintain a conversation and data extraction test methods

glaforge:
I propose a couple extra test methods that check
- that conversations are maintained (ie. that it indeed works as a chat which maintains the context of the conversation)
- that data extraction via `AiServices` works as expected


langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -111,4 +111,82 @@ void should_allow_custom_generativeModel_and_generationConfig() throws IOExcepti
...
+
+    static class Person {
+        String name;
+        int age;
+    }
+
+    interface PersonExtractor {
+        @dev.langchain4j.service.UserMessage(value = {
+            "Extract the name and age of the person described below.",
+            "Return a JSON document with a \"name\" and an \"age\" property, following this structure: ",
```
langchain4j, line 156: The idea of `AiServices` is that user does not need to specify this detailed information on how and what to return (formatting).
User just defines method return type and `AiServices` takes care of the rest (under the hood, it automatically creates and provides json schema to the model).

So I would suggest to simplify this to:
```
interface PersonExtractor {
        @dev.langchain4j.service.UserMessage("Extract information about a person from {{it}}")
        Person extractPerson(String text);
    }
```

You will also need to add this to `pom.xml` to make it compile:
```
        <dependency>
            <artifactId>langchain4j</artifactId>
            <scope>test</scope>
        </dependency>
```

ghost (commented):
@glaforge thanks a lot! Please see my comment.

langchain4j:
@glaforge do you plan to finish this PR or should we close it?

glaforge:
No, let's close it.
Maybe I'll contribute this back to the examples repository instead, at a later time.

PR #428, opened 2023-12-29
Implemented support for image inputs

ghost:
Support for image inputs was added in LangChain4j chat API as well as OpenAI & Gemini integrations


langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -111,4 +121,148 @@ void should_allow_custom_generativeModel_and_generationConfig() throws IOExcepti
...
         assertThat(response.content().text()).contains("Berlin");
     }
+
+    @Test
+    void should_accept_text_and_image_from_public_url() {
+
+        // given
+        Image image = Image.builder()
+                .url(CAT_IMAGE_URL)
+                .mimeType("image/png")
```
langchain4j: Hi @glaforge, would be awesome if VertexAI SDK could detect mime type automatically (like OpenAI does) without forcing end user to specify it manually

langchain4j:
@glaforge could you please review Gemini part? Thanks!

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
```diff
@@ -3,33 +3,44 @@
 import com.google.cloud.vertexai.VertexAI;
 import com.google.cloud.vertexai.api.GenerationConfig;
 import com.google.cloud.vertexai.generativeai.preview.GenerativeModel;
-import dev.langchain4j.data.message.AiMessage;
-import dev.langchain4j.data.message.SystemMessage;
-import dev.langchain4j.data.message.UserMessage;
+import com.sun.org.apache.xerces.internal.impl.dv.util.Base64;
```
glaforge: You're not using the one in `java.util.Base64` instead?
(it was introduced in Java 8)

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/PartsMapper.java
```diff
@@ -0,0 +1,57 @@
+package dev.langchain4j.model.vertexai;
+
+import com.google.cloud.vertexai.api.Part;
+import com.sun.org.apache.xerces.internal.impl.dv.util.Base64;
```
glaforge: What about `java.util.Base64`?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -3,31 +3,41 @@
 import com.google.cloud.vertexai.VertexAI;
 import com.google.cloud.vertexai.api.GenerationConfig;
 import com.google.cloud.vertexai.generativeai.preview.GenerativeModel;
-import dev.langchain4j.data.message.AiMessage;
-import dev.langchain4j.data.message.SystemMessage;
-import dev.langchain4j.data.message.UserMessage;
+import com.sun.org.apache.xerces.internal.impl.dv.util.Base64;
```
glaforge: `java.util.Base64`?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
glaforge: Yeah, I know. Maybe we could derive it from the URL if we can, so the user doesn't have to?
In the image builder, you'd compute mimeType by looking at the file extension if it's available?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -111,4 +121,148 @@ void should_allow_custom_generativeModel_and_generationConfig() throws IOExcepti
...
+    @Test
+    void should_accept_text_and_image_from_public_url() {
+
+        // given
+        Image image = Image.builder()
+                .url(CAT_IMAGE_URL)
+                .mimeType("image/png")
+                .build();
+        UserMessage userMessage = UserMessage.from(
+                TextContent.from("What do you see? Reply in one word."),
```
glaforge: I've read that Gemini sometimes gives better results when the images are given first, before the text.
For a simple test, I guess it doesn't matter, since it works just fine, but I was told to change a test case in the Gemini Java samples to put images first, text second.
This comment is more of an information, than a request to change, so no need to switch the two. Really just FYI :-)

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
```diff
@@ -185,4 +196,160 @@ public void onError(Throwable error) {
...
+
+        // when
+        TestStreamingResponseHandler<AiMessage> handler = new TestStreamingResponseHandler<>();
+        visionModel.generate(singletonList(userMessage), handler);
+        Response<AiMessage> response = handler.get();
+
+        // then
+        assertThat(response.content().text()).containsIgnoringCase("cat");
+
+        assertThat(response.tokenUsage().inputTokenCount()).isEqualTo(268);
```
glaforge: I'm wondering if there's an update to the model, if the input token count could change? (thus making the assertion break)
I know images, regardless of their size, counts for a fixed amount of token. So the variability may come more from the short text itself.
Hopefully it'll just be a matter of updating the number, should it ever change some day.

glaforge (approved):
It looks great to me!
It's interesting to see that it required quite a bit of changes here and there so it works across providers.
I'm wondering why you're not using the `java.util.Base64` class instead of the internal sun one?
Otherwise, looks great and easy to use when users mix text and images!

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
langchain4j: Yeah this is something we could do, but would be nice to keep it as simple as possible.

I've just experimented a bit and it seems that second part of mime type does not matter, it should start with `image/` hovewer. Even if the image is png, I can send `image/jpeg` and it works fine. I am not sure if this is a feature or a bug that we can abuse, but we could also just set the default mime type to `image/png` or whatever and it should work.

Another option would be to open a connection and get mime type using `url.openConnection().getContentType()`, but that might not always work...

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
langchain4j: Oh, good catch, did not notice that I imported the wrong one.
Thanks!

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
langchain4j: You're right, I will remove token assertions from here.

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
langchain4j: I have added simple mime type detection logic (by file extension, for popular image formats).
Please note that `PartMaker.fromMimeTypeAndData()` javadoc says that only `image/png` and `image/jpeg` are accepted, but it seems to accept other image types as well. So I have added mapping for many formats, not just `png` and `jpeg`, to make it easier for the user.

liushuaikobe:
@langchain4j  May I know when could I access these features in a released version?

langchain4j:
@liushuaikobe most probably end of this week or during the next week. In the meantime, you can build a snapshot

f361600134:
Awesome, this update is so timely, I just needed it!

PR #431, opened 2023-12-31
Adding metadata to information when calling chain.execute(string)

mateusscheper:
The goal of this change is to allow the user to add extra information when appending the retrieved docs informations on chain.execute(string). 
Example: 

```
chain = ConversationalRetrievalChain.builder()               
                .metadata(List.of("file_name"))
                [...]
                .build();
String response = chain.execute("My question");
```
When building the prompt, it will include the information like: 
```
information -> ...file_name: my_file.pdf
Content: my first retrieved content....

...file_name: my_file.pdf
Content: my second retrieved content...
```

Since it's my first contribution here, please let me know if something's wrong. Thanks!


langchain4j/src/main/java/dev/langchain4j/chain/ConversationalRetrievalChain.java
```diff
@@ -36,15 +36,19 @@ public class ConversationalRetrievalChain implements Chain<String, String> {
...
 
+    private final List<String> metadata;
+
     @Builder
     public ConversationalRetrievalChain(ChatLanguageModel chatLanguageModel,
                                         ChatMemory chatMemory,
                                         PromptTemplate promptTemplate,
-                                        Retriever<TextSegment> retriever) {
+                                        Retriever<TextSegment> retriever,
+                                        List<String> metadata) {
```
langchain4j: I would name it a bit more specific, something like `metadataKeysToInclude` or `includedMetadataKeys`

langchain4j/src/main/java/dev/langchain4j/chain/ConversationalRetrievalChain.java
```diff
@@ -69,10 +73,33 @@ public String execute(String question) {
...
+        }
+
+        if (!formattedText.toString().isEmpty()) {
+            formattedText.append("Content: ");
+        }
+
+        formattedText.append(textSegment.text());
+
+        return formattedText.toString();
+    }
```
langchain4j, line 104: could you please add a few unit tests to make sure this togic works as expected? e.g. without specified metadata keys to include, with 1,2 keys, with 0, 1, 2 segments, etc

ghost (commented):
@mateusscheper thanks a lot for your contribution! Please check my comments.

ghost (approved):
@mateusscheper looks great, thank you!

PR #441, opened 2024-01-03
Fix another Neo4jEmbeddingStoreTest error

vga91:
See here: https://github.com/langchain4j/langchain4j/actions/runs/7396118145/job/20120618008?pr=396

The fix made [here](https://github.com/langchain4j/langchain4j/pull/368) it is not enough,
because the test property is not always populated, see [here](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java#L71).

Therefore, I put an if-else on the iterator where needed


ghost (approved):
@vga91 thank you!

PR #443, opened 2024-01-03
Fix function without params for Azure Open AI

### LINKED ISSUE #442: [BUG] Azure Open AI produces incorrect json for tools without parameters
Please provide as much details as possible, this will help us to deliver a fix as soon as possible.
Thank you!

**Describe the bug**
The Azure OpenAI lib doesn't produce correct json for functions/tools without parameters. Current implementation (in InternalAzureOpenAiHelper) sets the parameter field to `{}` but it should be `{
"type": "object", 
"properties": {}
}`

**Log and Stack trace**
`Caused by: com.azure.core.exception.HttpResponseException: Status code 400, 
"{
  "error": {
    "message": "'{}' is not of type 'object' - 'functions.2.parameters'",
    "param": null,
    "code": null
  }
}`

**To Reproduce**
Use a tool without parameters

**Expected behavior**
No exception should be thrown.

- LangChain4j version: e.g. 0.25.0
- Java version: e.g. 21
- Spring Boot version (if applicable): e.g. 3.2.1

**Additional context**
functions soon to be replace by tools in the api but the schema for the parameters property seem to be the same so I guess it still makes sense to fix it.

kumlien:
Tools without parameters were not correctly serialized to json for Azure Open AI.
Change the json for the `parameters` property from `{}` to 
`{"type": "object", "properties": {}}` according to docs [here](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/function-calling?tabs=python#using-function-in-the-chat-completions-api-deprecated). 

fixes #442

### DIFF
--- a/langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
+++ b/langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
@@ -123,7 +123,7 @@ public class AzureOpenAiChatModelIT {
         Response<AiMessage> response = model.generate(Collections.singletonList(userMessage), toolSpecification);
 
         AiMessage aiMessage = response.content();
-        assertThat(aiMessage.text()).isBlank();
+        assertThat(aiMessage.text()).isNull();
 
         assertThat(aiMessage.toolExecutionRequests()).hasSize(1);
         ToolExecutionRequest toolExecutionRequest = aiMessage.toolExecutionRequests().get(0);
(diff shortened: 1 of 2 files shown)


langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -161,6 +161,42 @@ void should_call_function_with_argument(String deploymentName, String gptVersion
...
+
+        ToolSpecification noArgToolSpec = ToolSpecification.builder()
+                .name(toolName)
+                .description("Get the current date and time")
+                .build();
+
+        Response<AiMessage> response = model.generate(Collections.singletonList(userMessage), noArgToolSpec);
+
+        AiMessage aiMessage = response.content();
+        assertThat(aiMessage.text()).isBlank();
```
langchain4j: `text` should be `null` when there are `toolExecutionRequests`, could you please change the assertion to `.isNull()`? Would be cool if you could do the same for `should_call_function_with_argument` as well, thanks!

ghost (commented):
@kumlien thanks a lot! There is just one tiny thing left and we can merge.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
kumlien: `isNull` it is!

ghost (approved):
@kumlien thank you a lot!

PR #454, opened 2024-01-05
Tests and Docs for default EmbeddingModel methods.

crutcher:


langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
```diff
@@ -13,18 +13,34 @@
...
+     * @return the embedding.
+     */
     default Response<Embedding> embed(TextSegment textSegment) {
         Response<List<Embedding>> response = embedAll(singletonList(textSegment));
-        return Response.from(
-                response.content().get(0),
-                response.tokenUsage(),
-                response.finishReason()
-        );
+        assert response.content().size() == 1 :
```
langchain4j: Aren't assertions disabled by default? maybe better to use `ValidationUtils.ensureTrue(...)` or `ValidationUtils.ensureBetween(response.content().size(), 1, 1, "content")`

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
```diff
@@ -13,18 +13,34 @@
...
+     *
+     * @param text the text to embed.
+     * @return the embedding.
+     */
     default Response<Embedding> embed(String text) {
         return embed(TextSegment.from(text));
     }
 
+    /**
+     * Embed a text segment.
```
langchain4j: Probably worth mentioning that `TextSegment.metadata` is not embedded, just the `TextSegment.text`.

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
```diff
@@ -13,18 +13,34 @@
...
-                response.tokenUsage(),
-                response.finishReason()
-        );
+        assert response.content().size() == 1 :
+                String.format("Expected a single embedding, but got %d", response.content().size());
+        return Response.from(response.content().get(0), response.tokenUsage(), response.finishReason());
     }
 
+    /**
+     * Embeds a list of text segments.
```
langchain4j: Probably worth mentioning that `TextSegment.metadata` is not embedded, just the `TextSegment.text`.

ghost (changes requested):
@crutcher thanks a lot! Please check a few minor comments

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
crutcher: Done

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
crutcher: Done

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
crutcher: done

PR #455, opened 2024-01-05
Unify the 3 TokenCountEstimators into 2.

crutcher:


langchain4j-core/src/main/java/dev/langchain4j/model/chat/ChatTokenCountEstimator.java
```diff
@@ -14,7 +13,7 @@
  * Represents an interface for estimating the count of tokens in various text types such as a text, message, prompt, text segment, etc.
  * This can be useful when it's necessary to know in advance the cost of processing a specified text by the LLM.
  */
-public interface TokenCountEstimator {
+public interface ChatTokenCountEstimator extends TokenCountEstimator {
```
langchain4j, line 16: Is this renaming required? `ChatTokenCountEstimator` is in the `chat` package, so there is a distinction

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -198,7 +198,7 @@ private Response<AiMessage> generate(List<ChatMessage> messages,
     }
 
     @Override
-    public int estimateTokenCount(List<ChatMessage> messages) {
+    public int estimateChatMessagesTokenCount(List<ChatMessage> messages) {
```
langchain4j, line 201: I would avoid making breaking changes when not necessary

ghost (commented):
@crutcher thank you! Could you pelase provide a motivation behing this PR?

I was planning to revisit the design of these estimators and `Tokenizer` soon, and anticipate some breaking changes. Since this PR is not urgent, I would propose to put it on hold for a bit, to avoid making 2 breaking changes in a row. Hope for your understanding!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
crutcher: if we unify, we need to make sure that the interfaces are type erasure compatible.

we can't have multiple methods with the same erasure, so these became:
`estimateTokenCount(List<T> ...)` and caused conflicts.

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ChatTokenCountEstimator.java
crutcher: I see that argument; but it does really impact IDE code completion

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
langchain4j: Sure, but we can keep interfaces separately and there will be no problem with type erasure. As well as no breaking changes for users

PR #456, opened 2024-01-05
Support the Imagen image generation model from Google Cloud Vertex AI

glaforge:


langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImagenImageModel.java
```diff
@@ -0,0 +1,403 @@
...
+import static dev.langchain4j.internal.RetryUtils.withRetry;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotBlank;
+import static java.util.Collections.singletonList;
+
+/**
+ * Image model for the Google Cloud Vertex AI Imagen image generation models.
+ * Supports both versions of Imagen v1 and v2, respectively identified by the model names
+ * <code>imagegeneration@002</code> and <code>imagegeneration@005</code>.
+ */
+public class VertexAiImagenImageModel implements ImageModel {
```
langchain4j: It feels that "Imagen" should be removed from the class name, to be able to use it with other image models which might come later

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImagenImageModel.java
```diff
@@ -0,0 +1,403 @@
...
+     * Constructor of the Imagen image generation model.
+     *
+     * @param endpoint         the base URL of the API (eg. <code>https://us-central1-aiplatform.googleapis.com/v1/</code>)
+     * @param project          the Google Cloud Project ID
+     * @param location         the cloud region (eg. <code>us-central1</code>)
+     * @param publisher        the publisher of the model (<code>google</code> for Imagen)
+     * @param modelName        the name of the image model (<code>imagegeneration@002</code> or <code>imagegeneration@005</code>)
+     * @param seed             a fixed random seed number between 0 and 2^32-1
+     * @param language         the spoken language used for the prompt
+     * @param image            an image to edit
```
langchain4j: Maybe we need a separate method in `ImageModel` for image edition, like `Response<Image> edit(Image image, Image mask, String prompt);`?

Setting image and mask in the constructor feels hacky

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImagenImageModel.java
```diff
@@ -0,0 +1,403 @@
...
+            return this;
+        }
+
+        public Builder modelName(String modelName) {
+            this.modelName = modelName;
+            return this;
+        }
+
+        public Builder seed(Long seed) {
+            this.seed = seed & 0xffffffffL;
```
langchain4j: Do I understand it correctly that this masking ensures that provided `seed` is in a valid range? If yes, I would propose to do `ensureBetween(seed, MIN, MAX, "seed")` in the constructor instead, to fail fast if user provides wrong value instead of silently modifying it.

ghost (commented):
@glaforge great job, thank you! Please check my cooments

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImagenImageModel.java
glaforge: I'll call it `VertexAiImageModel` as suggested.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImagenImageModel.java
glaforge: The seed in an unsigned int, from 0 to 2^32, so I used a Long to make it fit, and the mask to keep it within bounds as you noticed. I'll go with the `ensureBetween()` method, it's a better idea than doing that transformation silently!

glaforge: I'll add a `Long` variant for that method, in addition to the existing `Integer` one (and `Double`).
(other option could be to just have a `Long` version)

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImagenImageModel.java
glaforge: I didn't want to change `ImageModel` since it's also used by OpenAI and AzureOpenAI, but I also felt it was not ideal to put those parameters in the constructor, instead of per-generation.
I'll add a new default method in the `ImageModel`.

glaforge: Actually, maybe several methods...

It's possible to generate an image from:
* a single text prompt (current implementation)
* an image with a text prompt (it updates the image but without a mask, as if the mask was the full picture)
* an image with a mask and a text prompt (to update only a portion of an image)

But then, those 3 methods also work for generating `n` images. So that would be 2 existing methods + 2 * 2 new methods.
What do you think?

```java
    default Response<Image> generate(Image image, String prompt) {
        throw new IllegalArgumentException("Operation is not supported");
    }

    default Response<List<Image>> generate(Image image, String prompt, int n) {
        throw new IllegalArgumentException("Operation is not supported");
    }

    default Response<Image> generate(Image image, Image mask, String prompt) {
        throw new IllegalArgumentException("Operation is not supported");
    }

    default Response<Image> generate(Image image, Image mask, String prompt, int n) {
        throw new IllegalArgumentException("Operation is not supported");
    }
```

glaforge: Also I'm wondering if we should call them `generate()` or if we should use a different method name? (`edit`, `update`, `modify`...)

glaforge: It's actually a little trickier.
Image editing can have a scale guidance, ie. how much the image should be edited within the mask.
An upscale operation is also an image editing operation, and requires an image size.
Not all image model will support all these things, and other image models may also support other things that Imagen doesn't support.
I'm not yet sure how to best approach this.

glaforge: Thinking of creating an `EditOperation` and an `UpscaleOperation` classes, but what are the commonalities between image models?

langchain4j: @glaforge I would keep it simple as of now and just add 1 or 2 methods:
`Response<Image> edit(Image image, String prompt);`
`Response<Image> edit(Image image, Image mask, String prompt);`
This should work for both VertexAI and OpenAI.
We can always add more methods if required (with `n`).

Scale guidance and other parameters for image edition can be set in the ctor for now (as we do for temperature, maxTokens, etc).
But the most variable part will be the image itself, right? This can be an argument to the `edit` method to make it flexible.

glaforge: Thanks for the guidance!

I was less worried about the `n` case, as in my mind, the `1` case is just a special case of `n` (at least that's the case with Imagen) 😄  So I was just thinking of redirecting `1` to `n` in my implementation. But I can indeed get rid of the `n` case.

For `edit`, I'll go as suggested and keep the other parameters in the constructor.
But somehow, I still feel a bit weird that we can't override temperature and max tokens on a per call basis and have to go through the constructor for this. But it's not really costly anyway to create a new instance, so I guess that's fine.

There's also the case of the _upscale_ operation. There's an image size parameter that is passed by constructor, and a mode variable, but that's only set transparently when we specify an image size. I guess I can do it as an edit operation, and have it in the constructor, just like scale guidance and others.

There's one last thing as well, which is the negative prompt. It seems lots of image models allow this (at least in their UIs, like StableDiffusion, Leonardo.AI, and indeed Imagen). Currently it's also in the constructor, but I feel like somehow this is quite integral to what an image model should support, so maybe later on we might have to support both a prompt String and a negative prompt String.

langchain4j: Regarding `n`, in my understanding in 95% cases `n=1`, so from the API perspective, it will be annoying for the user to always provide `n=1` and then do `.get(0)` all the time. If my current understanding is wrong and users will ask for a method with `n`, we can always add it, right?

Regarding other parameters, I agree that current API forces user to setup all parameters during model construction and does not allow to override them during usage. The idea was that these models will be mostly used in the automated way, so once the developer found fitting parameters for a specific task, he creates a model with concrete parameters and then the only variable thing during usage is the user input (text, images, etc).
So far I did not hear complaints in this regards. But I agree, we should think about making this more flexible, but also without overloading API with dozens of methods.

glaforge: I think I'm good now.
Can you please have another look and tell me what you think about it?

glaforge: I had forgotten to put the @Disabled annotation back and also point to your GCP project instead of mine. Now it's correct.

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
```diff
@@ -65,6 +69,19 @@ public String toString() {
                 " }";
     }
 
+    public static Image fromPath(Path path) {
```
langchain4j: I have doubts about this method: usually one needs to provide mime type when providing an `Image` as a content (instead of url) as an input to the model.

This method suggests that `Image` can be created from a `Path`, but this will not work (as of now) for 2 reasons:
```
ChatLanguageModel visionModel = VertexAiGeminiChatModel.builder()
            .project(System.getenv("GCP_PROJECT"))
            .modelName("gemini-pro-vision")
            .build();

UserMessage userMessage = UserMessage.from(
                ImageContent.from(Image.fromPath(Paths.get("C:/.../image.png"))),
                TextContent.from("What do you see? Reply in one word.")
        );
    
    visionModel.generate(userMessage);

```

1. when `url` is present, it takes precedence, so you will get `ClassCastException`
2. there is no mime type provided, so you will get `NullPointerException`

I would propose to remove this method for the moment, and I will create an issue to refactor `Image` to make sure it can be created only in correct state that can be used in all places (for Gemini, Imagen, OpenAI, etc).

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
```diff
@@ -29,4 +29,30 @@ public interface ImageModel {
...
     }
+
+    /**
+     * Given an existing image, edit this image following the given prompt.
+     *
+     * @param prompt The prompt to edit the image.
+     * @param image The image to be edited.
+     * @return The generated image Response.
+     */
+    default Response<Image> edit(String prompt, Image image) {
```
langchain4j: Could you please switch prompt and image places? Now it reads like the prompt is edited.

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
```diff
@@ -29,4 +29,30 @@ public interface ImageModel {
...
+        throw new IllegalArgumentException("Operation is not supported");
+    }
+
+    /**
+     * Given an existing image, edit this image following the given prompt,
+     * and apply the changes only to the part of the image delimited by the white area
+     * in the given black and white mask image.
```
langchain4j: "white area" and "black and white" parts are applicable to Imagen, but not to Dalle (they use transparency instead of black/white).

Maybe just "Given an existing image, edit this image following the given prompt and apply the changes only to the part of the image specified by the given mask."?

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
```diff
@@ -29,4 +29,30 @@ public interface ImageModel {
...
+     *
+     * @param prompt The prompt to edit the image.
+     * @param image The image to be edited.
+     * @param mask The black & white image mask to apply the edition of the image
+     *             only to the part delimited by the white area.
+     * @return The generated image Response.
+     */
+    default Response<Image> edit(String prompt, Image image, Image mask) {
```
langchain4j: Please move `prompt` to the end

ghost (commented):
@glaforge great job! Please check a few comments

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
glaforge: Funny! At first, I had the prompt at the end, and then thought that it would look more like the generate methods if prompt were first. But I agree, it sounds more logical to move it at the end!

langchain4j-core/src/main/java/dev/langchain4j/model/image/ImageModel.java
glaforge: Well do!

langchain4j-core/src/main/java/dev/langchain4j/data/image/Image.java
glaforge: OK. I hadn't looked at `ImageContent` and how it was working.
My `fromPath()` method was only used in my test, so I'll just move it in my test class as a local utility method for my test.

I wonder 2 things though:
- shouldn't `Image` also have a mime type field as well?
- could `ImageContent` also have `from()` methods that take `Path`, so that it's easier to load an image from the local file system?

langchain4j: `Image` has `mimeType` field (in `main` branch).
Yes, there should be all kinds of methods, will be addressed in https://github.com/langchain4j/langchain4j/issues/496

glaforge: Oops, sorry, yes, at some point there was no `mimeType` when I had started working on that PR, but it's here now!

Now, I'm good with the changes.
Please, when you have a moment, double check the latest changes 🙏

ghost (dismissed):
@glaforge looks great! Thank you!

ghost (approved):

PR #463, opened 2024-01-08
Document VertexAI authentication

ghost:


glaforge:
Yes, good idea to add some details about authentication.

As you need a GCP project to use Gemini and PaLM, when running locally you can just configure the `gcloud` SDK with your project that has those LLM APIs enabled. The Vertex AI libraries will automatically figure out how to authenticate to the APIs, without the need to export an JSON service account key, as long as you're logged in with the default application credentials, or with your own user account.

It's better to avoid generating such a key, as we run the risk of committing that key by mistake 😄 

For deployment, when running on GCP, same thing, authentication will be handled transparently.
I think it's only when deploying elsewhere that you need to have the JSON service account key.

langchain4j:
@glaforge do you mean it is better to swith options a) and b) ? Or remove option a (service account/json file) completely?

glaforge:
Switching a) and b) would be good.
I think I wouldn't remove the json file mention because not all developers necessarily use `gcloud`, so they would need that approach if they don't use it.

langchain4j:
@glaforge just did the change you suggested, could you please double-check?

glaforge:
Yup, I think it's better that way 👍

PR #465, opened 2024-01-08
Test coverage for EmbeddingStore defaults.

crutcher:
- Gave `findRelevant(id, emb, maxResults)` a default implemented in terms of `findRelevant(id, emb, maxResults, minScore)`; symmetric with the no-id methods.


ghost (approved):
@crutcher awesome, thanks!

PR #474, opened 2024-01-09
Add jacoco coverage targets to langchain4j-core, attached to the `verify` stage.

crutcher:


ghost (approved):
@crutcher awesome, thank you!

PR #484, opened 2024-01-09
Flesh out RetryUtils with a policy mechanism.

crutcher:
- Backed `RetryUtils.withRetry()` with a `RetryPolicy`.
- added support for exponential backoff and request jitter.

default retry times are now:
* 500 * (1 + uniform(0.20))
* 750 * (1 + uniform(0.20))
* 1250 * (1 + uniform(0.20)
* (500 * pow(1.5, attempt - 1)) * (1 + uniform(0.20))


ghost (approved):
@crutcher thank you!

isergeymd:
@crutcher Hello, I see that you are swallowing an InterruptedException in the while loop. Was this intentional? I am trying to cancel the thread execution, but it keeps retrying because of this.

I think it should looks smth like this :
```
while (true){
//
if (Thread.currentThread().isInterrupted()) {
            throw new RuntimeException("Execution cancelled by user");
        }
}
```

isergeymd:
raised bug here https://github.com/langchain4j/langchain4j/issues/4479

PR #485, opened 2024-01-09
ServiceHelper.{loadService, loadFactoryService}()

crutcher:


ghost (approved):

geoand:
@langchain4j unfortunately this change completely breaks the Quarkus integration.

The reason is that because of this, classes that try load stuff from a factory end up trying to call the static initializers of the default classes and fail utterly if their dependencies are not on the classpath.

As this defeats the purpose of the use of the factories, I would kindly request this be reverted. 

Thanks

langchain4j:
@geoand damn, sorry for the oversight. Sure, I will revert this.

geoand:
No problem at all.

Honestly even if I had seen, it's unlikely I would have spotted the problem myself :)

langchain4j:
@geoand can I still use new new way of loading for all model builders?

geoand:
Anything that used method references in a static field to the default implementation of a service, that then in turn has a static field that loads gson (or okhttp, etc.) will fail.

As long as that exact condition isn't met, Class Loading will work just fine :)

geoand:
Thanks for the revert @langchain4j!

Is there any chance we can have a `0.26.2` release containing this?

🙏🏼

langchain4j:
@geoand can this wait till thursday? If not, I can push to do it tomorrow.

geoand:
Of course,  no rush

PR #498, opened 2024-01-10
added convenience methods to Embedding:

ghost:
- `magnitude()`
- `cosineSimilarity(Embedding)`
- `relevanceScore(Embedding)`
- `from(List<Float>)` -> `from(List<? extends Number>)`


crutcher (dismissed):

langchain4j-core/src/main/java/dev/langchain4j/data/embedding/Embedding.java
```diff
@@ -37,6 +41,42 @@ public int dimension() {
         return vector.length;
     }
 
+    public float magnitude() {
```
crutcher, line 82:
```suggestion
    /**
     * Return the L2 Norm of the feature vector.
     *
     * <code>{@pre Sqrt(Sum(f[i] ** 2))}</code>
     *
     * @return float the norm.
     */
    public float magnitude() {
```

PR #510, opened 2024-01-15
Fix all jococo javadoc warnings in langchain4j-core.

crutcher:
A number of private constructors were added to namespace classes.

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/P.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/P.java
@@ -14,7 +14,8 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 public @interface P {
 
     /**
-     * Description of a parameter
+     * Description of a parameterjj
+     * @return the description of a parameter
      */
     String value();
 }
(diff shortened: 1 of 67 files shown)


langchain4j-core/src/main/java/dev/langchain4j/agent/tool/P.java
```diff
@@ -14,7 +14,8 @@
 public @interface P {
 
     /**
-     * Description of a parameter
+     * Description of a parameterjj
```
langchain4j:
```suggestion
     * Description of a parameter
```

langchain4j-core/src/main/java/dev/langchain4j/model/output/OutputParser.java
```diff
@@ -1,8 +1,21 @@
...
 
+    /**
+     * Parse the given text.
+     * @param text the text to parse.
+     * @return the parsed output.
+     */
     T parse(String text);
 
+    /**
+     * Description of the text format..
```
langchain4j:
```suggestion
     * Description of the text format.
```

ghost (dismissed):

PR #512, opened 2024-01-15
Add basic ErrorProne config to langchain4j-core.

crutcher:
See [ErrorProne](https://errorprone.info/index)

- Runs during `mvn verify`.
- fix for two extant issues.
- Stricter tests could be enabled later.

### DIFF
--- a/langchain4j-core/pom.xml
+++ b/langchain4j-core/pom.xml
@@ -79,6 +79,20 @@
                     <compilerArgs>
                         <arg>-XDcompilePolicy=simple</arg>
                         <arg>-Xplugin:ErrorProne</arg>
+                        <!--
+                         * See https://errorprone.info/docs/installation
+                         + See also .mvn/jvm.config
+                         +-->
+                        <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED</arg>
+                        <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.file=ALL-UNNAMED</arg>
+                        <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED</arg>
+                        <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.model=ALL-UNNAMED</arg>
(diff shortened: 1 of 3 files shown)


crutcher:
This has some issues with the java8 tests:
```
[INFO] Compiling 86 source files with javac [debug target 8] to target/classes
An exception has occurred in the compiler (1.8.0_392). Please file a bug against the Java compiler via the Java bug reporting page (http://bugreport.java.com/) after checking the Bug Database (http://bugs.java.com/) for duplicates. Include your program and the following diagnostic in your report. Thank you.
java.lang.UnsupportedClassVersionError: com/google/errorprone/ErrorProneJavacPlugin has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

crutcher:
When running in the CI environment; I have to enable `<fork>true</fork>`; and that seems to have some performance impact for CI which causes the tests to get cancelled (despite producing valid errorprone warnings).

And the version of errorprone isn't compatible with the java8 environment.

I could do some more work to make this work.

langchain4j:
@crutcher thanks a lot! Please do not invest much effort in fighting with java 8, I anticipate a move to 11 quite soon

langchain4j:
@crutcher I like the overall idea to have some strict checks, I am just not sure why exactly ErrorProne? It seems that there are lots of other more popular tools like Checkstyle, PMD, FindBugs, SonarQube, etc. Is there a reason why ErrorProne was chosen?
Thanks!

langchain4j-core/pom.xml
```diff
@@ -68,6 +68,47 @@
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.11.0</version>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
```
kpavlov, line 77: Language level has been upgraded to 17. It make sense to upgrade

kpavlov (commented):
Thank you, @crutcher 
Would it be possible to refresh your PR for JDK 17?

PR #514, opened 2024-01-15
Add support for Mistral AI open models

czelabueno:
Mistral AI continues its mission to deliver the best open models to the developer community. They are [growing quickly](https://mistral.ai/news/mixtral-of-experts/) and it's very helpful add support to Mistral AI models.


langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/ChatCompletionModel.java
```diff
@@ -0,0 +1,42 @@
...
+ * <ul>
+ *   <li>{@link #MISTRAL_TINY} - powered by Mistral-7B-v0.2</li>
+ *   <li>{@link #MISTRAL_SMALL} - powered by Mixtral-8X7B-v0.1</li>
+ *   <li>{@link #MISTRAL_MEDIUM} - currently relies on an internal prototype model</li>
+ * </ul>
+ * </p>
+ * 
+ * @see <a href="https://docs.mistral.ai/platform/endpoints/">Mistral AI Endpoints</a>
+ */
+enum ChatCompletionModel {
```
langchain4j:
```suggestion
public enum ChatCompletionModel {
```

langchain4j-mistral-ai/pom.xml
```diff
@@ -0,0 +1,113 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>0.26.0-SNAPSHOT</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-mistral-ai</artifactId>
```
langchain4j, line 12: Please add this artifact to `langchain4j-bom/pom.xml`

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,121 @@
...
+import java.io.IOException;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static dev.langchain4j.internal.Utils.isNullOrBlank;
+import static dev.langchain4j.model.output.FinishReason.*;
+import static java.util.stream.Collectors.toList;
+
+public class DefaultMistralAiHelper{
```
langchain4j:
```suggestion
class DefaultMistralAiHelper {
```
Please make sure all internal implementation classes that user should not see are package-private

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/EmbeddingModel.java
```diff
@@ -0,0 +1,27 @@
+package dev.langchain4j.model.mistralai;
+
+/**
+ * The EmbeddingModel enum represents the available embedding models in the Mistral AI module.
+ */
+enum EmbeddingModel {
```
langchain4j:
```suggestion
public enum EmbeddingModel {
```

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiClient.java
```diff
@@ -0,0 +1,199 @@
...
+                } else {
+                    try {
+                        ChatCompletionResponse chatCompletionResponse = GSON.fromJson(data, ChatCompletionResponse.class);
+                        ChatCompletionChoice choice = chatCompletionResponse.getChoices().get(0);
+                        String chunk = choice.getDelta().getContent();
+                        contentBuilder.append(chunk);
+                        handler.onNext(chunk);
+
+                        //Retrieving token usage of the last choice
+                        if(choice.getFinishReason() != null){
```
langchain4j: May I suggest a bit simpler solution?

```
EventSourceListener eventSourceListener = new EventSourceListener() {

            final StringBuilder contentBuilder = new StringBuilder();
            TokenUsage tokenUsage;
            FinishReason finishReason;

            @Override
            public void onOpen(EventSource eventSource, okhttp3.Response response) {
                LOGGER.debug("onOpen()");
                logResponse(response);
            }

            @Override
            public void onEvent(EventSource eventSource, String id, String type, String data) {

                LOGGER.debug("onEvent() {}", data);
                if ("[DONE]".equals(data)) {
                    Response<AiMessage> response = Response.from(
                            AiMessage.from(contentBuilder.toString()),
                            tokenUsage,
                            finishReason
                    );
                    handler.onComplete(response);
                } else {
                    try {
                        MistralChatCompletionResponse chatCompletionResponse = GSON.fromJson(data, MistralChatCompletionResponse.class);
                        MistralChatCompletionChoice choice = chatCompletionResponse.getChoices().get(0);
                        String chunk = choice.getDelta().getContent();
                        contentBuilder.append(chunk);
                        handler.onNext(chunk);

                        MistralUsageInfo usage = choice.getUsage();
                        if (usage != null) {
                            this.tokenUsage = tokenUsageFrom(usage);
                        }

                        String finishReason = choice.getFinishReason();
                        if (finishReason != null) {
                            this.finishReason = finishReasonFrom(finishReason);
                        }
                    } catch (Exception e) {
                        handler.onError(e);
                    }

                }
            }

            @Override
            public void onFailure(EventSource eventSource, Throwable t, okhttp3.Response response) {
                LOGGER.debug("onFailure()", t);
                logResponse(response);

                if (t != null) {
                    handler.onError(t);
                } else {
                    handler.onError(new RuntimeException(String.format("status code: %s; body: %s", response.code(), response.body())));
                }
            }

            @Override
            public void onClosed(EventSource eventSource) {
                LOGGER.debug("onClosed()");
            }
        };
```

This way token usage does not depend on presence of finish reason (Mistral can change something in the future...)

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiEmbeddingModel.java
```diff
@@ -0,0 +1,87 @@
...
+        EmbeddingRequest request = EmbeddingRequest.builder()
+                .model(modelName)
+                .input(textSegments.stream().map(TextSegment::text).collect(toList()))
+                .encodingFormat(MISTRALAI_API_CREATE_EMBEDDINGS_ENCODING_FORMAT)
+                .build();
+
+        EmbeddingResponse response = withRetry(() -> client.embedding(request), maxRetries);
+
+        List<Embedding> embeddings = response.getData().stream()
+                .map(mistralAiEmbedding -> Embedding.from(mistralAiEmbedding.getEmbedding()))
```
langchain4j, line 87: Do they always return embeddings in the same order as input strings? Double-checking because I have noticed that they provide `index` in `EmbeddingObject` for some reason...

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiModels.java
```diff
@@ -0,0 +1,88 @@
...
+                this.getModels().content().stream().map(ModelCard::getId).collect(toList())
+        );
+    }
+
+    /**
+     * Retrieves the list of all available models.
+     *
+     * @return the response containing the list of models
+     */
+    public Response<List<ModelCard>> getModels(){
```
langchain4j: keep only this and rename?

crutcher:
Can I suggest that the classes which re-use class names in the core libraries change? I'm aware that yes, other model bindings have reused names (such as `EmbeddingModel`); but doing so has a  big impact on code completion in IDEs, and it's a pattern I'd like to suggest we begin reversing.

I'd like to suggest, where a class name matches a `langchain4j-core` or `langchain4j` class, if the class is the not the same kind of thing, we change it entirely; and if it is the same kind of thing, we smurf the type.

For instance, `EmbeddingModel` is an interface in `langchain4j-core`, and an enum in this PR. For this one, I'd suggest `MistralEmbeddingModelType`.

But there are other collisions.

The rule of thumb I'd generally suggest is "what happens when I type <tab> while coding?"; does the IDE offer an unambiguous completion? Does it import the right library?

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiClient.java
```diff
@@ -0,0 +1,199 @@
...
+                throw toException(retrofitResponse);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public void streamingChatCompletion(ChatCompletionRequest request, StreamingResponseHandler<AiMessage> handler) {
+        EventSourceListener eventSourceListener = new EventSourceListener() {
+            StringBuilder contentBuilder = new StringBuilder();
```
langchain4j:
```suggestion
            StringBuffer contentBuilder = new StringBuffer();
```

langchain4j: to be on the safe side

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/ModelPermission.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+class ModelPermission{
```
langchain4j:
```suggestion
public class ModelPermission{
```

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiStreamingChatModelIT.java
```diff
@@ -0,0 +1,415 @@
...
+    void should_stream_answer_and_return_token_usage_and_finish_reason_stop() throws ExecutionException, InterruptedException, TimeoutException {
+
+        CompletableFuture<String> futureAnswer = new CompletableFuture<>();
+        CompletableFuture<Response<AiMessage>> futureResponse = new CompletableFuture<>();
+
+        // given
+        UserMessage userMessage = userMessage("What is the capital of Peru?");
+
+        // when
+        model.generate(userMessage.text(), new StreamingResponseHandler<AiMessage>() {
```
langchain4j: please use `TestStreamingResponseHandler`, see  `OpenAiStreamingChatModelIT` as an example

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiStreamingChatModelIT.java
```diff
@@ -0,0 +1,415 @@
...
+            }
+
+            @Override
+            public void onError(Throwable error) {
+                futureAnswer.completeExceptionally(error);
+                futureResponse.completeExceptionally(error);
+            }
+        });
+
+        String chunk = futureAnswer.get(10, SECONDS);
```
langchain4j: minor: `chunk` seems to be a bit misleading name. This is a complete `answer`, not just one chunk

czelabueno:
@crutcher ok I agree. This PR was updated with the convention names suggested.
cc: @langchain4j

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,121 @@
...
+    public static FinishReason finishReasonFrom(String mistralAiFinishReason) {
+        if (mistralAiFinishReason == null) {
+            return null;
+        }
+        switch (mistralAiFinishReason) {
+            case "stop":
+                return STOP;
+            case "length":
+                return LENGTH;
+            case "model_length":
```
langchain4j, line 100: What is the difference between `length` and `model_length`? Should they both map to `LENGTH`?

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,121 @@
...
+            default:
+                return null;
+        }
+    }
+
+    public static void logResponse(Response response){
```
langchain4j: It would be nice to log requests as well.

(Optional) It would be even better to log raw HTTP requests as is done in OpenAI integration for debugging potential mapping issues.

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,121 @@
...
+        try {
+            LOGGER.debug("Response code: {}", response.code());
```
langchain4j: Could you please add `logRequests` and `logResponses` parameters in builders of all mistral models? Similar to OpenAI model classes

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,121 @@
...
+            LOGGER.warn("Error while logging response", e);
+        }
+    }
+
+    private static String getResponseBody(Response response) throws IOException {
+       return isEventStream(response) ? "" : response.peekBody(Long.MAX_VALUE).string();
+    }
+
+    private static String getResponseHeaders(Response response){
+       return (String) StreamSupport.stream(response.headers().spliterator(),false).map(header -> {
```
langchain4j:
```suggestion
       return StreamSupport.stream(response.headers().spliterator(),false).map(header -> {
```

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,121 @@
...
+           String headerKey = header.component1();
+           String headerValue = header.component2();
+           if (headerKey.equals("Authorization")) {
+               headerValue = "Bearer " + headerValue.substring(0, 5) + "..." + headerValue.substring(headerValue.length() - 5);
+           } else if (headerKey.equals("api-key")) {
+               headerValue = headerValue.substring(0, 2) + "..." + headerValue.substring(headerValue.length() - 2);
```
langchain4j: Does mistral support `api-key` headers? Seems to be a copy-paste from openai4j module which supports Azure OpenAI (they use `api-key` instead of `Authorization`)

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,121 @@
...
+
+    private static String getResponseBody(Response response) throws IOException {
+       return isEventStream(response) ? "" : response.peekBody(Long.MAX_VALUE).string();
+    }
+
+    private static String getResponseHeaders(Response response){
+       return (String) StreamSupport.stream(response.headers().spliterator(),false).map(header -> {
+           String headerKey = header.component1();
+           String headerValue = header.component2();
+           if (headerKey.equals("Authorization")) {
```
langchain4j, line 110: Can response headers contain API key? This would be very dangerous...

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiModels.java
```diff
@@ -0,0 +1,88 @@
...
+                this.getModels().content().stream().map(MistralModelCard::getId).collect(toList())
+        );
+    }
+
+    /**
+     * Retrieves the list of all available models.
+     *
+     * @return the response containing the list of models
+     */
+    public Response<List<MistralModelCard>> getModels(){
```
langchain4j:
```suggestion
    public Response<List<MistralModelCard>> availableModels(){
```

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiModels.java
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Retrieves the details of a specific model.
+     *
+     * @param modelId the ID of the model
+     * @return the response containing the model details
+     */
+    public Response<MistralModelCard> getModelDetails(String modelId){
+        return Response.from(
+                this.getModels().content().stream().filter(modelCard -> modelCard.getId().equals(modelId)).findFirst().orElse(null)
```
langchain4j: I would propose to remove `getModelDetails()` and `get()` methods and keep only one which returns the list of models with details. User can then get ID and details of desired model himself

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralChatCompletionModel.java
```diff
@@ -0,0 +1,42 @@
...
+ * <ul>
+ *   <li>{@link #MISTRAL_TINY} - powered by Mistral-7B-v0.2</li>
+ *   <li>{@link #MISTRAL_SMALL} - powered by Mixtral-8X7B-v0.1</li>
+ *   <li>{@link #MISTRAL_MEDIUM} - currently relies on an internal prototype model</li>
+ * </ul>
+ * </p>
+ * 
+ * @see <a href="https://docs.mistral.ai/platform/endpoints/">Mistral AI Endpoints</a>
+ */
+enum MistralChatCompletionModel {
```
langchain4j: I would propose to rename it into `MistralAiChatModelName` to emphasize that it can be used with `MistralAiChatModel` model

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralEmbeddingModelType.java
```diff
@@ -0,0 +1,27 @@
+package dev.langchain4j.model.mistralai;
+
+/**
+ * The MistralEmbeddingModelType enum represents the available embedding models in the Mistral AI module.
+ */
+enum MistralEmbeddingModelType {
```
langchain4j: Maybe `MistralAiEmbeddingModelName`?

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralModelCard.java
```diff
@@ -0,0 +1,24 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.List;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class MistralModelCard {
```
langchain4j:
```suggestion
public class MistralAiModelCard {
```

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralModelPermission.java
```diff
@@ -0,0 +1,28 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+class MistralModelPermission {
```
langchain4j:
```suggestion
public class MistralAiModelPermission {
```

Please make sure that all user-facing classes are `public` and have `MistralAi` (instead of `Mistral`) prefix

ghost (changes requested):
@czelabueno thank you so much again for your contribution! 🤗 

Please check my comments, there is nothing critical, but would be nice to address those issues.
Thank you!

langchain4j-mistral-ai/pom.xml
czelabueno: alright!

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiClient.java
czelabueno: yeah ok I agree

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiEmbeddingModel.java
czelabueno: They always return embeddings in the same order as input, it's tested in the IT tests `MistralAiEmbeddingModelIT.should_embed_and_return_token_usage_with_multiple_inputs` and also in their official [python client](https://github.com/mistralai/client-python/blob/main/src/mistralai/client.py#L210C12-L210C12) lib use the same logic. That's why I didn't see it necessary to re-order it.

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiModels.java
czelabueno: Ok I wanted to provide some functions to the user but none of the client libs have it so I better remove it . You're right.

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiStreamingChatModelIT.java
czelabueno: Some test `OpenAiStreamingChatModelIT` methods still use `StreamingResponseHandler` but I assume you're migrating them all to `TestStreamingResponseHandler`, right?

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiStreamingChatModelIT.java
czelabueno: by using `TestStreamingResponseHandler` my futureAnswer is no longer needed.

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
czelabueno: It is in his [API specification](https://docs.mistral.ai/redocusaurus/plugin-redoc-0.yaml), according to what they tell me on Discord it is reserved for a future implementation, for now it is `null`

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralChatCompletionModel.java
czelabueno: alright, make sense. Good!

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralEmbeddingModelType.java
czelabueno: yeah!

czelabueno:
Hey @langchain4j I fixed the issues suggested in this commit [2c1a22f](https://github.com/langchain4j/langchain4j/pull/514/commits/2c1a22f9d033fe6ecddb4e303a5b731578bd42fd). Pls, check it again.

Thanks!

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
czelabueno: Ok, I made the necessary change to allow developer users to decide to log the request/response according to the log level they configure in their logger client libraries.

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
czelabueno: Done!

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
czelabueno: This method was an incomplete attempt to make it reusable for the request. The response does not contain API Key. I already removed it.

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiStreamingChatModelIT.java
langchain4j: yes, that's the plan. as well as making all tests model implementation agnostic, so that the same test logic applies to both blocking and streaming models.

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,120 @@
...
+
+    private static String maskAuthorizationHeaderValue(String authorizationHeaderValue) {
+        try {
+            Matcher matcher = MISTRAI_API_KEY_BEARER_PATTERN.matcher(authorizationHeaderValue);
+            StringBuffer sb = new StringBuffer();
+
+            while (matcher.find()) {
+                String bearer = matcher.group(1);
+                String token = matcher.group(2);
+                matcher.appendReplacement(sb, bearer + " " + token.substring(0, 7) + "..." + token.substring(token.length() - 7));
```
langchain4j:
```suggestion
                matcher.appendReplacement(sb, bearer + " " + token.substring(0, 2) + "..." + token.substring(token.length() - 2));
```
Just to be on the safe side. 4 symbols is enough to identify one of your keys...

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiChatModel.java
```diff
@@ -0,0 +1,121 @@
...
+                              Boolean logRequests,
+                              Boolean logResponses,
+                              Integer maxRetries) {
+
+        this.client = MistralAiClient.builder()
+                .baseUrl(formattedURLForRetrofit(getOrDefault(baseUrl, MISTRALAI_API_URL)))
+                .apiKey(ensureNotBlankApiKey(apiKey))
+                .timeout(getOrDefault(timeout, Duration.ofSeconds(60)))
+                .logRequests(getOrDefault(logRequests, false))
+                .logResponses(getOrDefault(logResponses, false))
```
langchain4j, line 72: for some reason **streaming** responses are logged now even if one does not provide a value for `logResponses` or set it explicitly to `false`

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/DefaultMistralAiHelper.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static dev.langchain4j.internal.Utils.isNullOrBlank;
+import static dev.langchain4j.model.output.FinishReason.*;
+import static java.util.stream.Collectors.toList;
+
+class DefaultMistralAiHelper{
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultMistralAiHelper.class);
```
langchain4j: minor: can be removed

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiClient.java
```diff
@@ -0,0 +1,206 @@
...
+
+        mistralAiApi = retrofit.create(MistralAiApi.class);
+    }
+
+    public MistralChatCompletionResponse chatCompletion(MistralChatCompletionRequest request) {
+        try {
+            retrofit2.Response<MistralChatCompletionResponse> retrofitResponse
+                    = mistralAiApi.chatCompletion(request).execute();
+            if (retrofitResponse.isSuccessful()) {
+                LOGGER.debug("ChatCompletionResponseBody: {}", retrofitResponse.body());
```
langchain4j: I guess theese can be removed now once you have a proper logging interceptors

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiClient.java
```diff
@@ -0,0 +1,206 @@
...
+
+    public void streamingChatCompletion(MistralChatCompletionRequest request, StreamingResponseHandler<AiMessage> handler) {
+        EventSourceListener eventSourceListener = new EventSourceListener() {
+            final StringBuffer contentBuilder = new StringBuffer();
+            TokenUsage tokenUsage;
+            FinishReason finishReason;
+
+            @Override
+            public void onOpen(EventSource eventSource, okhttp3.Response response) {
+                LOGGER.debug("onOpen()");
```
langchain4j: theese (here and below) should be logged only if `logResponses == true`

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiClient.java
```diff
@@ -0,0 +1,206 @@
...
+                        mistralAiApi.streamingChatCompletion(request).request(),
+                        eventSourceListener);
+    }
+
+    public MistralEmbeddingResponse embedding(MistralEmbeddingRequest request) {
+        try {
+            retrofit2.Response<MistralEmbeddingResponse> retrofitResponse
+                    = mistralAiApi.embedding(request).execute();
+            if (retrofitResponse.isSuccessful()) {
+                LOGGER.debug("EmbeddingResponseBody: {}", retrofitResponse.body());
```
langchain4j: can be removed

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiClient.java
```diff
@@ -0,0 +1,206 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    public MistralModelResponse listModels() {
+        try {
+            retrofit2.Response<MistralModelResponse> retrofitResponse
+                    = mistralAiApi.models().execute();
+            if (retrofitResponse.isSuccessful()) {
+                LOGGER.debug("ModelResponseBody: {}", retrofitResponse.body());
```
langchain4j: can be removed

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiRequestLoggingInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private void log(Request request) {
+        String message = "Request:\n- method: {}\n- url: {}\n- headers: {}\n- body: {}";
+        try {
+            if (LOGGER.isInfoEnabled()) {
+                LOGGER.info(message, new Object[]{request.method(), request.url(), getHeaders(request.headers()), getBody(request)});
+            } else if (LOGGER.isWarnEnabled()) {
+                LOGGER.warn(message, new Object[]{request.method(), request.url(), getHeaders(request.headers()), getBody(request)});
+            } else if (LOGGER.isErrorEnabled()) {
+                LOGGER.error(message, new Object[]{request.method(), request.url(), getHeaders(request.headers()), getBody(request)});
```
langchain4j: I am not sure I understand the purpose of this, could you please explain? Why would anyone want to log http requests and responses at error/warn levels?

I would suggest to log only at debug level and user will be able to fine-tune this via his logger configuration

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiResponseLoggingInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+        Request request = chain.request();
+        Response response = chain.proceed(request);
+        this.log(response);
+        return response;
+    }
+
+    private void log(Response response) {
+        String message = "Response:\n- status code: {}\n- headers: {}\n- body: {}";
+        try {
+            if (LOGGER.isInfoEnabled()) {
```
langchain4j: same here

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralChatCompletionModelName.java
```diff
@@ -0,0 +1,42 @@
...
+ * <ul>
+ *   <li>{@link #MISTRAL_TINY} - powered by Mistral-7B-v0.2</li>
+ *   <li>{@link #MISTRAL_SMALL} - powered by Mixtral-8X7B-v0.1</li>
+ *   <li>{@link #MISTRAL_MEDIUM} - currently relies on an internal prototype model</li>
+ * </ul>
+ * </p>
+ * 
+ * @see <a href="https://docs.mistral.ai/platform/endpoints/">Mistral AI Endpoints</a>
+ */
+public enum MistralChatCompletionModelName {
```
langchain4j:
```suggestion
public enum MistralAiChatModelName {
```

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralEmbeddingModelName.java
```diff
@@ -0,0 +1,27 @@
+package dev.langchain4j.model.mistralai;
+
+/**
+ * The MistralEmbeddingModelName enum represents the available embedding models in the Mistral AI module.
+ */
+public enum MistralEmbeddingModelName {
```
langchain4j:
```suggestion
public enum MistralAiEmbeddingModelName {
```

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiModelsIT.java
```diff
@@ -0,0 +1,27 @@
...
+
+    //https://docs.mistral.ai/models/
+    @Test
+    void should_return_all_model_cards(){
+        // when
+        Response<List<MistralAiModelCard>> response = models.availableModels();
+
+        // then
+        assertThat(response.content()).isNotEmpty();
+        assertThat(response.content().size()).isEqualTo(4);
```
langchain4j: this will fail at some point, please just assert that it is greater than 0

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiClient.java
```diff
@@ -0,0 +1,206 @@
...
+                    handler.onComplete(response);
+                } else {
+                    try {
+                        MistralChatCompletionResponse chatCompletionResponse = GSON.fromJson(data, MistralChatCompletionResponse.class);
+                        MistralChatCompletionChoice choice = chatCompletionResponse.getChoices().get(0);
+                        String chunk = choice.getDelta().getContent();
+                        contentBuilder.append(chunk);
+                        handler.onNext(chunk);
+
+                        MistralUsageInfo usageInfo = choice.getUsage();
```
langchain4j:
```suggestion
                        MistralUsageInfo usageInfo = chatCompletionResponse.getUsage();
```
I made a mistake in this code. Right now `MistralAiStreamingChatModelIT` fails because of this. Changing `choice` to `chatCompletionResponse` should fix it.

ghost (commented):
Hi @czelabueno, thanks a lot and sorry for the delayed review!
There are just a few things left (please check comments) and we can merge.
BTW, I plan to release early next week (in case you want this to be included in the release).

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiChatModel.java
czelabueno: If you tested it from IT classes, remember that I added `TestStreamingResponseHandler` as you asked and this has explicit printouts. Maybe you're seeing that.

```
public class TestStreamingResponseHandler<T> implements StreamingResponseHandler<T> {

    private final CompletableFuture<Response<T>> futureResponse = new CompletableFuture<>();

    @Override
    public void onNext(String token) {
        System.out.println("onNext: '" + token + "'");
    }

    @Override
    public void onComplete(Response<T> response) {
        System.out.println("onComplete: '" + response + "'");
        futureResponse.complete(response);
    }
```
    
Some test cases of the **Streaming** model create instances with logger and without logger. I confirmed that only instances created with logRequest and logResponse set to `true` are logged.

czelabueno:
Hi @langchain4j In this commit [2078d32](https://github.com/langchain4j/langchain4j/pull/514/commits/2078d32eef3da43622e6eb96f33b1669f0972444) you can find all the comments resolved and congratulations for all the effort you make. It would be great if Mistral AI was included in the next release as well.
PD: just in case, most java build is failing by an issue in the open-search module.

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiChatModel.java
langchain4j: Damn, that was indeed stupid from my side :D Thanks!

langchain4j: Actually... it was printing from both `MistralAiClient` and `TestStreamingResponseHandler`, but you have already fixed that, thank you!

ghost (approved):
@czelabueno great job, thank you!

czelabueno:
Wonderful, thank you so much @langchain4j

PR #523, opened 2024-01-17
Use the Jackson BOM to solve Jackson dependency errors

jdubois:
Using Azure OpenAI, on some Jackson mappings I'm getting the following error:
```
Caused by: java.lang.NoClassDefFoundError: com/fasterxml/jackson/annotation/JsonTypeInfo$Value
```
This is because not all our Jackson dependencies are using the same version. The intent of this commit is to fix this, by using the Jackson BOM.


langchain4j-parent/pom.xml
```diff
@@ -266,6 +265,13 @@
...
 
+            <dependency>
+                <groupId>com.fasterxml.jackson</groupId>
+                <artifactId>jackson-bom</artifactId>
+                <version>${jackson.version}</version>
+                <type>pom</type>
+            </dependency>
+
             <dependency>
                 <groupId>com.fasterxml.jackson.core</groupId>
```
langchain4j: Now once you've added `jackson-bom`, does it make sense to list all jackson dependencies here?

shalk: also. if add bom in  `<dependencyManagement>`, i think  need add `<scope> import </scope> `

jdubois: You are both right, I used the import scope and simplified the code

shalk:
LGTM

ghost (approved):
@jdubois thank you!

PR #524, opened 2024-01-17
Testing and efficiency work on HierarchicalDocumentSplitter.

crutcher:
- reworked SegmentBuilder as a String concat mechanism; since we were calling "build()" on it so often.


langchain4j:
HI @crutcher, thanks a lot! Could you please add more context to this?
What was the exact problem with `build()`?
Do you have some performance numbers before/after?
Thanks!

crutcher:
`segmentBuilder.build()` gets called several times for each append or overlap case; so keeping the result close-to the final result seems prudent.

.trim() is a no-op if there's no whitespace; so `segmentBuilder.build()` will mostly do nothing now.

```
    @Override
    public List<TextSegment> split(Document document) {
...
        for (String part : parts) {
            if (segmentBuilder.hasSpaceFor(part)) {
                segmentBuilder.append(part);
            } else {
                if (segmentBuilder.isNotEmpty() && !segmentBuilder.build().equals(overlap)) {
                    String segmentText = segmentBuilder.build();
                    segments.add(createSegment(segmentText, document, index.getAndIncrement()));
                    segmentBuilder.reset();
...
                    segmentBuilder.append(overlap);
                }
                if (segmentBuilder.hasSpaceFor(part)) {
                    segmentBuilder.append(part);
                } else {
...
                    segmentBuilder.append(part);
                    for (TextSegment segment : subSplitter.split(Document.from(segmentBuilder.build()))) {
                        segments.add(createSegment(segment.text(), document, index.getAndIncrement()));
                    }
                    segmentBuilder.reset();
...
                    segmentBuilder.append(overlap);
                }
            }
        }
        
        if (segmentBuilder.isNotEmpty() && !segmentBuilder.build().equals(overlap)) {
            segments.add(createSegment(segmentBuilder.build(), document, index.getAndIncrement()));
        }
```

crutcher:
Mostly this was pulling on coverage and documentation threads; but I saw the work duplication and decided to fix it while i was in there. I could try and benchmark the change; but I believe a lot of tests sped up as a result

langchain4j:
Ok, got it, thanks!

PR #527, opened 2024-01-19
Allow reading prompt (both system & user) message from a resource

gkfirst8:
This is to implement the suggestion done in https://github.com/langchain4j/langchain4j/discussions/508 :

It would be valuable to our team if we could specify a resource to read as a prompt-message. We use this technique for more things like SQL-queries. Makes reading and crafting these texts easier.

I would be happy to add this to the source via a PR but I thought let's test the waters first to see if more people would like this and how high the chance would be of the PR getting pulled.

It would work like this: where the annotation is read (eg. DefaultAiServices.prepareUserMessage) we could switch on available parameters to the annotation and read the resource specified (based on the methods-class using clazz.getResourceAsStream) in stead of using the provided text. (I do realize we could want to cache the resource reading when that becomes a bottleneck later)

What would you think? What would the names of the parameters to @SystemMessage & @Usermessage need to be? Something like "resource" or "resourceName"?

This is the code I could use to text resource:
```
public static String getString(Class<?> clazz, String name) {
    return getString(clazz.getResourceAsStream(name));
}

public static String getString(InputStream inputStream) {
    try (Scanner scanner = new Scanner(inputStream); //
            Scanner s = scanner.useDelimiter("\\A")) {
        return s.hasNext() ? s.next() : "";
    }
}
```

Any suggestions in this area would be welcome too.


gkfirst8:
I'm aware that I change the AI slightly: I made the value to the @UserMessage optional. But because the getPromptText method checks for empty, you will quickly be warned during testing that something is missing. (You do lose the chance of compile-time-checks but when one provides a useless prompt (like an empty String or the like) the compiler would not help anyway.)
Hope this will not be a show-stopper.

gkfirst8:
@langchain4j not sure about the procedure. Can you help me make this procedure a success, please? Maybe I forgot something?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -269,6 +279,35 @@ private static ChatMessage prepareUserMessage(Method method, Object[] args) {
         throw illegalConfiguration("For methods with multiple parameters, each parameter must be annotated with @V, @UserMessage, @UserName or @MemoryId");
     }
 
+    private static String getPromptText(Method method, String type, String resource, String[] value, String delimiter) {
+        String systemMessageTemplate;
```
langchain4j:
```suggestion
        String messageTemplate;
```
This method is appicable for user message as well

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -269,6 +279,35 @@ private static ChatMessage prepareUserMessage(Method method, Object[] args) {
...
+        } else {
+            systemMessageTemplate = String.join(delimiter, value);
+        }
+        if (systemMessageTemplate.trim().isEmpty()) {
+            throw illegalConfiguration("@%sMessage's template cannot be empty", type);
+        }
+        return systemMessageTemplate;
+    }
+
+    public static String getResourceText(Class<?> clazz, String name) {
```
langchain4j:
```suggestion
    private static String getResourceText(Class<?> clazz, String name) {
```

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -269,6 +279,35 @@ private static ChatMessage prepareUserMessage(Method method, Object[] args) {
...
+        return getText(clazz.getResourceAsStream(name));
+    }
+
+    public static String getText(InputStream inputStream) {
```
langchain4j:
```suggestion
    private static String getText(InputStream inputStream) {
```

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -269,6 +279,35 @@ private static ChatMessage prepareUserMessage(Method method, Object[] args) {
...
+        if (inputStream == null) {
+            return null;
+        }
+        try (Scanner scanner = new Scanner(inputStream); //
+                Scanner s = scanner.useDelimiter("\\A")) {
+            return s.hasNext() ? s.next() : "";
```
langchain4j, line 297: should it return `null` instead of an empty string?

langchain4j/src/main/java/dev/langchain4j/service/SystemMessage.java
```diff
@@ -14,7 +14,17 @@
...
+    String[] value() default "";
 
     String delimiter() default "\n";
+
+	/**
+	 * The resource to read the prompt template from. If the resource is not found,
+	 * an IllegalConfigurationException is thrown. If no resource is specified we
+	 * will fall-back to the value of {@link #value()}.
+	 *
+	 * The resource will be read from the class package of the class with the
```
langchain4j:
```suggestion
	 * The resource will be read by calling {@code getResourceAsStream(resource)} on the class```

File: langchain4j/src/main/java/dev/langchain4j/service/UserMessage.java [hunk 6]
```diff
@@ -18,4 +18,14 @@
     String[] value() default "";
 
     String delimiter() default "\n";
+
+	/**
+	 * The resource to read the prompt template from. If the resource is not found,
+	 * an IllegalConfigurationException is thrown. If no resource is specified we
+	 * will fall-back to the value of {@link #value()}.
+	 *
+	 * The resource will be read from the class package of the class with the
```
langchain4j:
```suggestion
	 * The resource will be read by calling {@code getResourceAsStream(resource)} on the class```
```

langchain4j/src/main/java/dev/langchain4j/service/SystemMessage.java
```diff
@@ -14,7 +14,17 @@
...
+	 * annotation in it.
```
langchain4j:
```suggestion
	 * containing the method annotated with {@code @SystemMessage}.
```

langchain4j/src/main/java/dev/langchain4j/service/UserMessage.java
```diff
@@ -18,4 +18,14 @@
...
+	 * annotation in it.
```
langchain4j:
```suggestion
	 * containing the method annotated with {@code @UserMessage}.
```

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
```diff
@@ -287,10 +288,16 @@ interface Chef {
         @UserMessage("Create recipe using only {{it}}")
         Recipe createRecipeFrom(String... ingredients);
 
+        @UserMessage(resource = "chefs-prompt-based-on-ingredients.txt")
```
langchain4j: I am wondering: would it be more readable with `fromResource`: `@UserMessage(fromResource = "chefs-prompt-based-on-ingredients.txt")`? WDYT?

ghost (commented):
@gkfirst8 Thanks a lot, great job! I have left just a few minor comments/suggestions.

langchain4j:
> @langchain4j not sure about the procedure. Can you help me make this procedure a success, please? Maybe I forgot something?

Not sure I understand what you mean. Could you please elaborate?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
gkfirst8: I like both. Maybe yours is more in line with the style of this repo?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
gkfirst8: When I read DefaultAiServices.getPromptText I would say that "" makes more sense: it will respond in the same way as when the "String[] value" is 'empty': with a message "template cannot be empty", indicating the resource was found but it was empty.

gkfirst8:
@langchain4j I was just a bit impatient, sorry.

Thank you for your nice comments. I've taken over almost everything. One more question for you remains.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
gkfirst8: I selected your suggestion.

gkfirst8:
@langchain4j what is still needed to get this PR approved?

ghost (approved):
@gkfirst8 thanks a lot! Sorry for the delay, I was out for some time.

7010G:
After testing, it was found that the resource name needs to be preceded by /, otherwise the resource cannot be accessed normally.

langchain4j:
@7010G thanks a lot for reporting! Fixed: https://github.com/langchain4j/langchain4j/pull/781

PR #528, opened 2024-01-19
enable Reproducible Builds

hboutemy:
see https://maven.apache.org/guides/mini/guide-reproducible-builds.html


ghost (approved):
@hboutemy thank you!

PR #529, opened 2024-01-19
Azure OpenAI: migrate to the new API, from "function" to "tools"

### LINKED ISSUE #521: [FEATURE] Azure OpenAI: use tools instead of deprecated functions
image
image
https://learn.microsoft.com/en-au/azure/ai-services/openai/how-to/function-calling?tabs=python#using-function-in-the-chat-completions-api-deprecated

jdubois:
As discussed in #521 there's a new API to call functions, which is documented at https://platform.openai.com/docs/api-reference/chat/create

I'm not sure we should merge it now:

- Those APIs are not compatible, so you use one or the other
- Contrary to the opened ticket, I can currently use the older "function" API call without any issue, not even a warning
- For GPT 4, you need to use the model version "1106-preview", as with the current default one ("0613"), it will fail as this new API is not available.

Fix #521


langchain4j-parent/pom.xml
```diff
@@ -266,6 +265,13 @@
                 <version>${elastic.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>com.fasterxml.jackson</groupId>
+                <artifactId>jackson-bom</artifactId>
+                <version>${jackson.version}</version>
+                <type>pom</type>
```
Martin7-1, line 301: Should add `<scope>import</scope>`

jdubois: Oh sorry that comes from #523 which should be merged before

langchain4j:
@jdubois That's strange 🤔 Vanilla OpenAI supports both old functions and new tools at the same time, I have just checked on these models:
```
[1] gpt-3.5-turbo
[2] gpt-3.5-turbo-0613
[3] gpt-3.5-turbo-1106
[4] gpt-3.5-turbo-16k
[5] gpt-3.5-turbo-16k-0613
[6] gpt-4
[7] gpt-4-0613
[8] gpt-4-1106-preview
```

jdubois:
@langchain4j when I tested, the old API ("functions") worked everywhere, and the newer one ("tools") only worked on the newer model. Then I don't think the API version is related to the model version, so you could have an older model with a newer API.
BTW did you get your API key to do the integration tests yourself? I'm the (internal) thread trying to solve it, I'm just not sure it's all good you on your side.

langchain4j:
@jdubois I did not get a key yet. I have asked via multiple channels, but no resposne so far :(
Would appreciate a lot if you could facilityte this somehow!

langchain4j:
@jdubois so what do we do with functions/tools then? As far as I understand, we cannot merge this PR now because this will break for user using older models, right? Are there any plans to support both at the same time to ease the transition?

jdubois:
@langchain4j for merging this PR: let me test again, typically with `gpt-4-0613`. If it works with older models, then it will be all good.
For the key, let me ping you privately, you should have had something.

jdubois:
@langchain4j I just did some tests with Azure OpenAI, and:
- Indeed the "old" API works on GPT 3.4 and GPT 4
- I'm now getting some Null responses when testing this PR, and this definitely wasn't the case when I coded it last week

I believe we can stay with the old API: it's "deprecated" but still working everywhere, when the newer one doesn't seem to be very stable at the moment.
We should rather wait a bit before merging this. At least let's wait until you get your API key and can do automated integration testing (which should be OK today!!!).

langchain4j:
@jdubois ok then this will not go into the release next week. I will take a look into this after the release

jdubois:
Argh sorry @langchain4j I tried to rebase with my IDE and I don't know what went wrong, but that's all broken now :-(

jdubois:
Fixed and synced with the main branch :-)

langchain4j:
@jdubois I have noticed that this is not a draft any more. Is it ready for review/merge?

jdubois:
Sorry @langchain4j I'm converting this to "draft" again.
The streaming part doesn't work properly, I need to rework this part completely.

langchain4j:
@jdubois just a gentle reminder :) Are we ready to merge this?

jdubois:
@langchain4j sorry, I need more time to test this.

langchain4j:
@jdubois no rush!

jdubois:
@langchain4j I did more tests so I'm putting this as "ready to review", the only thing that I don't understand here is that when I execute a tool the finish reason is `STOP` and not `TOOL_EXECUTION`.
This would be the logical finish reason, and that's how it's working in the `langchain4j-openai` module, so it seems specific to Azure OpenAI.

langchain4j:
@jdubois there are conflicts now, could you please take a look?

Any idea when funcitons will not be supported (how much time do we have)? Was not able to find it in the official docu...

jdubois:
@langchain4j I fixed the conflicts, and all my tests with GPT 3.5 and GPT 4 are green, so I believe we should merge it now.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingResponseBuilder.java
```diff
@@ -61,14 +65,22 @@ public void append(ChatCompletions completions) {
             return;
         }
 
-        FunctionCall functionCall = delta.getFunctionCall();
-        if (functionCall != null) {
-            if (functionCall.getName() != null) {
-                toolNameBuilder.append(functionCall.getName());
-            }
+        if (delta.getToolCalls() != null && !delta.getToolCalls().isEmpty()) {
+            ChatCompletionsToolCall toolCall = delta.getToolCalls().get(0);
```
langchain4j: It takes only the first tool call, is this intentional? Parallel tool calls should already be supported, according to the [docs](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/function-calling?tabs=python#supported-models)

jdubois: It's coming from the OpenAI implementation, but indeed that's wrong.
I'm on it, it'll take some time as I'm adding some integration tests.

jdubois: I can now have multiple tools calls, but then the answers are streamed synchronously. I'm adding a full test suite, so it's easier to see.

jdubois: @langchain4j here is the test suite, with 3 calls!

langchain4j: @jdubois awesome!

Could you please copy&adapt this tests into `AzureOpenAiChatModelIT` for sync model?

There are probably still 2 problems in `InternalAzureOpenAiHelper.toOpenAiMessage()`:
1. `chatRequestAssistantMessage.setFunctionCall(functionCallFrom(message));`: shouldn't `setToolCalls()` be used instead?
2. `return new ChatRequestFunctionMessage(nameFrom(message), toolExecutionResultMessage.text());`: shouldn't `ChatRequestToolMessage` be used instead?

jdubois: Indeed, I've found several issues there. I unfortunately don't have enough time right now, this is going to take me at least 2 weeks to fix.

langchain4j: @jdubois no worries! It does not seem to be an urgent change, right?

jdubois: Should be good now!

jdubois:
Hi @langchain4j can you do another review? This looks OK to me now, and this is now becoming important

langchain4j:
@jdubois ok, will review it asap

jdubois:
@langchain4j can you add @alzimmermsft as a reviewer to this PR? He's working on the Azure OpenAI Java SDK, and he just agreed to do a review, so that should be a great help!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingResponseBuilder.java
```diff
@@ -61,16 +70,32 @@ public void append(ChatCompletions completions) {
...
+        if (delta.getToolCalls() != null && !delta.getToolCalls().isEmpty()) {
+            for (ChatCompletionsToolCall toolCall : delta.getToolCalls()) {
+                ToolExecutionRequestBuilder toolExecutionRequestBuilder;
+                if (toolCall.getId() != null) {
+                    toolExecutionsIndex = toolCall.getId();
+                    toolExecutionRequestBuilder = new ToolExecutionRequestBuilder();
+                    toolExecutionRequestBuilder.idBuilder.append(toolExecutionsIndex);
+                    toolExecutionRequestBuilderHashMap.put(toolExecutionsIndex, toolExecutionRequestBuilder);
+                } else {
+                    toolExecutionRequestBuilder = toolExecutionRequestBuilderHashMap.get(toolExecutionsIndex);
```
alzimmermsft, line 82: Just so I understand the expected behavior here more, as this is the first time I'm looking at this package, the expected behavior for a `ChatCompletionsToolCall` if it doesn't have an ID set is to use the last seen `ToolExecutionRequestBuilder`?

I have no strong opinion or knowledge on whether that is the correct behavior but wanted to ask more in case that isn't the right behavior.

langchain4j:
@jdubois of course!

langchain4j:
@jdubois but I can't add @alzimmermsft, there is no such user in the reviewers search :/

jdubois:
Oh my fault, I forgot: to be a reviewer, you need to be part of the project.
@langchain4j you can create specific teams for this, I can have a look at what we did on @jhipster if you want.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingResponseBuilder.java
```diff
@@ -134,19 +174,28 @@ private TokenUsage tokenUsage(String content, Tokenizer tokenizer) {
...
         }
 
         return new TokenUsage(inputTokenCount, outputTokenCount);
     }
+
+    private static class ToolExecutionRequestBuilder {
+
+        private final StringBuffer idBuilder = new StringBuffer();
+        private final StringBuffer nameBuilder = new StringBuffer();
+        private final StringBuffer argumentsBuilder = new StringBuffer();
```
alzimmermsft, line 198: Can ignore if wanted, is the expectation for these to be used in a threaded scenario? If not, `StringBuilder` is preferable as it doesn't have synchronization as it's not thread safe.

langchain4j-azure-open-ai/pom.xml
```diff
@@ -26,6 +26,11 @@
             <artifactId>azure-ai-openai</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-xml</artifactId>
```
langchain4j: nit: this seems to be not required

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingResponseBuilder.java
```diff
@@ -61,16 +70,32 @@ public void append(ChatCompletions completions) {
...
+                    if (functionCall.getFunction().getName() != null) {
+                        toolExecutionRequestBuilder.nameBuilder.append(functionCall.getFunction().getName());
+                    }
+                    if (functionCall.getFunction().getArguments() != null) {
+                        toolExecutionRequestBuilder.argumentsBuilder.append(functionCall.getFunction().getArguments());
+                    }
+                }
             }
         }
+        logger.debug("toolExecutionRequestBuilderHashMap: " + toolExecutionRequestBuilderHashMap);
```
langchain4j: nit: is this required?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -178,37 +178,51 @@ private static String nameFrom(ChatMessage message) {
...
         FunctionDefinition functionDefinition = new FunctionDefinition(toolSpecification.name());
         functionDefinition.setDescription(toolSpecification.description());
         functionDefinition.setParameters(toOpenAiParameters(toolSpecification.parameters()));
-        return functionDefinition;
+        return new ChatCompletionsFunctionToolDefinition(functionDefinition);
+    }
+
+    public static BinaryData toToolChoice(ToolSpecification toolThatMustBeExecuted) {
+        FunctionCall functionCall;
+        if (ChatCompletionsToolSelectionPreset.values().stream()
```
langchain4j: I am not sure I understand how this works.
Here `toolThatMustBeExecuted.name()` (e.g. "calculator", "weather", etc.) is compared with `ChatCompletionsToolSelectionPreset.values()`  (e.g. "auto" or "none").
So, this condition will most likely never be true.

Is this intended logic?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -178,37 +178,51 @@ private static String nameFrom(ChatMessage message) {
...
+                .anyMatch(preset -> preset.toString().equals(toolThatMustBeExecuted.name()))) {
+
+            functionCall = new FunctionCall(ChatCompletionsToolSelectionPreset.fromString(toolThatMustBeExecuted.name()).toString(), "");
+        } else {
+            functionCall = new FunctionCall(toolThatMustBeExecuted.name(), "");
```
langchain4j: Is it correct to set empty (`""`) arguments?

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModelIT.java
```diff
@@ -234,17 +362,62 @@ public void onError(Throwable error) {
 
         AiMessage aiMessage = response.content();
         assertThat(aiMessage.text()).isNull();
+        List<ChatMessage> messages = new ArrayList<>();
+        messages.add(userMessage);
+        messages.add(aiMessage);
+        assertThat(aiMessage.toolExecutionRequests()).hasSize(3);
```
langchain4j, line 368: Could you please add a similar test to the `AzureOpenAiChatModelIT`, to make sure that multiple tools at once can be called with sync model as well?

ghost (commented):
@jdubois great job!

langchain4j-azure-open-ai/pom.xml
jdubois: Indeed, it's removed!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingResponseBuilder.java
jdubois: I removed it

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
jdubois: Indeed, I removed it

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
jdubois: I'm now using `toOpenAiParameters(toolThatMustBeExecuted.parameters()).toString()` to get the arguments. So it's the same as for `setTools`, just we put the parameters into the binary data, unwrap them (to get the JSON), and then wrap them again in the final `setToolChoice`.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModelIT.java
jdubois: Done!

jdubois:
@langchain4j thanks for the review, this should be all good!

ghost (approved):
@jdubois thank you!

PR #530, opened 2024-01-19
Azure AI Search support as an embedding store

### LINKED ISSUE #422: [FEATURE] Add support for Azure AI Search as EmbeddingStore
Currently multiple embedding stores already exist in langchain4j, however, Azure AI Search is not yet supported as EmbeddingStore. This store is often used when implementing RAG on the Azure platform with Azure OpenAI.

An implementation of https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java#L12 using https://learn.microsoft.com/en-us/java/api/com.azure.search.documents?view=azure-java-stable 

Getting the relevant documents 'manually' similar to how it's done in https://github.com/langchain4j/langchain4j-examples/blob/d47dbb92e4ab583401731992c5dca91c28c643cf/other-examples/src/main/java/ChatWithDocumentsExamples.java#L68 and then sending these as info to the LLM.

jdubois:
Fix #422

This code is inspired by @sinedied 's LangChain JS implementation from https://github.com/langchain-ai/langchainjs/blob/c9c879eb107822ddd9699464013985ef2e9b056d/libs/langchain-community/src/vectorstores/azure_aisearch.ts


langchain4j:
@jdubois awesome stuff! Would be cool to have Azure hybrid search support as well (probably as a `ContentRetriever` from https://github.com/langchain4j/langchain4j/pull/538)

jdubois:
Thanks @langchain4j ! Yes we should have it all, the hard parts are already done.
I’ll have a closer look at #538

langchain4j:
@jdubois this PR is marked as draft, but seems that it is ready for review already?

jdubois:
@langchain4j It's looking ready, but the integration tests are not satisfying and I don't understand why. Maybe you have some insights? I'm also going to check with the Azure AI Search team (this is what I planned to do today).
For example, running `EmbeddingStoreWithoutMetadataIT.should_return_correct_score()`, which works for all other embeddings stores, I have:

```
java.lang.AssertionError: 
Expecting actual:
  0.8461208
to be close to:
  0.9090677239710618
by less than 1% but difference was 6.924338232589762%.
(a difference of exactly 1% being considered valid)
```

This is with `AllMiniLmL6V2QuantizedEmbeddingModel` and the result is better if I switch to OpenAI with `text-embedding-ada-002` (the difference drops to 3%), but this is still failing.

I'm trying to understand if I corrupted some data or misconfigured something.

langchain4j:
@jdubois this is probably because you need to convert the score. We use "relevance score" which I am already not sure was a good idea to introduce...
It is basically a derivative of cosine similarity, but it compresses it into 0..1 range (instead of -1..1) for ease of use.

You can check `RelevanceScore.fromCosineSimilarity()` and javadoc of `EmbeddingStore.findRelevant(Embedding, int, double)`.

So you will need to convert it both before search (you need to convert minScore into cosineSimilarity and pass it to Azure search if it supports this kind of filtering) and after search (you need to convert cosine similarity from Azure into relevance score using `RelevanceScore.fromCosineSimilarity()`).

jdubois:
Thanks @langchain4j ! I'm also learning that Azure AI Search is using its own specific way to calculate the score: https://learn.microsoft.com/en-us/azure/search/vector-search-ranking#scores-in-a-vector-search-results

So I need to mix both to find the right number, probably!

jdubois:
And it's all good!! That was hard! @langchain4j  do you want me to put it "ready for review" now, or do we wait until I do the hybrid and semantic search as well?

langchain4j:
@jdubois up to you! FYI I plan to release on monday/tuesday

jdubois:
@langchain4j I'm putting it "ready for review", anyway the other hybrid & semantic search should only be new methods, so no big impact for the rest.

langchain4j:
@jdubois awesome, will check it tomorrow! BTW hybrid search probably would be better to do as a custom implementation of `ContentRetriever` or even `RetrievalAugmentor` from https://github.com/langchain4j/langchain4j/pull/538

langchain4j-azure-ai-search/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.honton.chas</groupId>
+                <artifactId>license-maven-plugin</artifactId>
+                <configuration>
+                    <skipCompliance>true</skipCompliance>
```
langchain4j: could you please add a comment with the reason why it was disabled?

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+                .indexName(INDEX_NAME)
+                .buildClient();
+
+        createOrUpdateIndex(embeddingModel);
+    }
+
+     void createOrUpdateIndex(EmbeddingModel embeddingModel) {
+
+        // Embed a test query to get the embedding dimensions
+        int embeddingDimensions = embeddingModel.embed("test").content().vector().length;
```
langchain4j: I would suggest accepting an `int dimensions` parameter into ctor/builder instead.
One should be able to create `EmbeddingStore` without `EmbeddingModel`.

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+                                        .setContentFields(new SemanticField(DEFAULT_FIELD_CONTENT))
+                                        .setKeywordsFields(new SemanticField(DEFAULT_FIELD_CONTENT)))));
+
+
+        SearchIndex index = new SearchIndex(INDEX_NAME)
+                .setFields(fields)
+                .setVectorSearch(vectorSearch)
+                .setSemanticSearch(semanticSearch);
+
+        searchIndexClient.createOrUpdateIndex(index);
```
langchain4j, line 162: this config has a lot of parameters to tweak. should we maybe accept a `SearchIndex` as an optional param in ctor/builder? in case user wants more flexibility

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+    }
+
+    void deleteIndex() {
```
langchain4j: maybe public? so that user can do that in case he wants

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+    /**
+     * Finds the most relevant (closest in space) embeddings to the provided reference embedding.
+     *
+     * @param referenceEmbedding The embedding used as a reference. Returned embeddings should be relevant (closest) to this one.
+     * @param maxResults         The maximum number of embeddings to be returned.
+     * @param minScore           The minimum relevance score, ranging from 0 to 1 (inclusive).
+     *                           Only embeddings with a score of this value or higher will be returned.
+     * @return A list of embedding matches.
+     * Each embedding match includes a relevance score (derivative of cosine distance),
+     * ranging from 0 (not relevant) to 1 (highly relevant).
```
langchain4j: If there are no changes in the doc (compared to `EmbeddingStore` javadoc), please remove this duplicated javadoc.
You can use
```
/**
 * {@inheritDoc}
 */
```
if you want to add something extra.

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+     */
+    @Override
+    public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbedding, int maxResults, double minScore) {
+        List<Float> vector = floatsArrayToList(referenceEmbedding.vector());
```
langchain4j:
```suggestion
        List<Float> vector = referenceEmbedding.vectorAsList();
```

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+                searchClient.search(null,
+                        new SearchOptions()
+                                .setVectorSearchOptions(new VectorSearchOptions().setQueries(vectorizedQuery)),
+                        Context.NONE);
+
+        List<EmbeddingMatch<TextSegment>> result = new ArrayList<>();
+        for (SearchResult searchResult : searchResults) {
+            Double score = fromAzureScoreToRelevanceScore(searchResult.getScore());
+            if (score < minScore) {
+                break;
```
langchain4j: should this be `continue`? or Azure always returns highest score first?

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+            }
+            SearchDocument searchDocument = searchResult.getDocument(SearchDocument.class);
+            String embeddingId = (String) searchDocument.get(DEFAULT_FIELD_ID);
+            List<Double> embeddingList = (List<Double>) searchDocument.get(DEFAULT_FIELD_CONTENT_VECTOR);
+            float[] embeddingArray = doublesListToFloatArray(embeddingList);
+            Embedding embedding = Embedding.from(embeddingArray);
+            String embbededContent = (String) searchDocument.get(DEFAULT_FIELD_CONTENT);
```
langchain4j:
```suggestion
            String embeddedContent = (String) searchDocument.get(DEFAULT_FIELD_CONTENT);
```

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+                LinkedHashMap metadata = (LinkedHashMap) searchDocument.get(DEFAULT_FIELD_METADATA);
+                List attributes = (List) metadata.get(DEFAULT_FIELD_METADATA_ATTRS);
+                Map<String, String> attributesMap = new HashMap<>();
+                for (Object attribute : attributes) {
+                    LinkedHashMap innerAttribute = (LinkedHashMap) attribute;
+                    String key = (String) innerAttribute.get("key");
+                    String value = (String) innerAttribute.get("value");
+                    attributesMap.put(key, value);
+                }
+                Metadata langChainMetada = Metadata.from(attributesMap);
```
langchain4j:
```suggestion
                Metadata langChainMetadata = Metadata.from(attributesMap);
```

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+        }
+        ensureTrue(ids.size() == embeddings.size(), "ids size is not equal to embeddings size");
+        ensureTrue(embedded == null || embeddings.size() == embedded.size(),
+                "embeddings size is not equal to embedded size");
+
+        List<Document> searchDocuments = new ArrayList<>();
+        for (int i = 0; i < ids.size(); ++i) {
+            Document document = new Document();
+            document.setId(ids.get(i));
+            document.setContent_vector(floatsArrayToList(embeddings.get(i).vector()));
```
langchain4j:
```suggestion
            document.setContent_vector(embeddings.get(i).vectorAsList());
```

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+                for (Map.Entry<String, String> entry : embedded.get(i).metadata().asMap().entrySet()) {
+                    Document.Metadata.Attribute attribute = new Document.Metadata.Attribute();
+                    attribute.setKey(entry.getKey());
+                    attribute.setValue(entry.getValue());
+                    attributes.add(attribute);
+                }
+                metadata.setAttributes(attributes);
+                document.setMetadata(metadata);
+            } else {
+                document.setContent("");
```
langchain4j: it does not permit `null`?

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+            }
+            searchDocuments.add(document);
+        }
+        List<IndexingResult> indexingResults = searchClient.uploadDocuments(searchDocuments).getResults();
+        for (IndexingResult indexingResult : indexingResults) {
+            if (!indexingResult.isSucceeded()) {
+                log.error("Failed to add embedding: {}", indexingResult.getErrorMessage());
+            } else {
+                log.info("Added embedding: {}", indexingResult.getKey());
```
langchain4j:
```suggestion
                log.debug("Added embedding: {}", indexingResult.getKey());
```

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+                document.setMetadata(metadata);
+            } else {
+                document.setContent("");
+            }
+            searchDocuments.add(document);
+        }
+        List<IndexingResult> indexingResults = searchClient.uploadDocuments(searchDocuments).getResults();
+        for (IndexingResult indexingResult : indexingResults) {
+            if (!indexingResult.isSucceeded()) {
+                log.error("Failed to add embedding: {}", indexingResult.getErrorMessage());
```
langchain4j: should we maybe throw an exception here instead of logging?

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+        for (int i = 0; i < doubles.size(); ++i) {
+            array[i] = doubles.get(i).floatValue();
+        }
+        return array;
+    }
+
+    /**
+     * Calculates LangChain4J's RelevanceScore from Azure AI Search's score.
+     *
+     * Score in Azure AI Search is the cosine distance, and not the cosine similarity.
```
langchain4j: seems that javadoc is not true: if score is cosine distance, then there is no need to do `double cosineDistance = (1 - score) / score;`

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/Document.java
```diff
@@ -0,0 +1,92 @@
+package dev.langchain4j.store.embedding.azure.search;
+
+import java.util.Collection;
+
+public class Document {
```
langchain4j:
```suggestion
class Document {
```

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/Document.java
```diff
@@ -0,0 +1,92 @@
...
+
+    public void setContent(String content) {
+        this.content = content;
+    }
+
+    public Collection<Float> getContent_vector() {
+        return content_vector;
+    }
+
+    public void setContent_vector(Collection<Float> content_vector) {
```
langchain4j:
```suggestion
    public void setContentVector(Collection<Float> contentVector) {
```

langchain4j-azure-ai-search/src/test/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStoreIT.java
```diff
@@ -0,0 +1,121 @@
...
+        assertThat(relevant.get(1).embedding()).isNotNull();
+        assertThat(relevant.get(1).embedded().text()).isIn(content1, content3, content5);
+        log.info("#2 relevant item: {}", relevant.get(1).embedded().text());
+        assertThat(relevant.get(2).embedding()).isNotNull();
+        assertThat(relevant.get(2).embedded().text()).isIn(content1, content3, content5);
+        log.info("#3 relevant item: {}", relevant.get(2).embedded().text());
+    }
+
+    @Test
+    void should_return_correct_score() {
```
langchain4j: this test can probably be removed as it is no different from original method except it adds logging. you can add logging to the base method if you want

langchain4j-azure-open-ai/pom.xml
```diff
@@ -18,10 +18,16 @@
 
     <dependencies>
 
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
```
langchain4j, line 23: this should not be required: `langchain4j-open-ai` imports `langchain4j-core` transitively

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+
+        List<Document> searchDocuments = new ArrayList<>();
+        for (int i = 0; i < ids.size(); ++i) {
+            Document document = new Document();
+            document.setId(ids.get(i));
+            document.setContent_vector(floatsArrayToList(embeddings.get(i).vector()));
+            if (embedded != null) {
+                document.setContent(embedded.get(i).text());
+                Document.Metadata metadata = new Document.Metadata();
+                metadata.setSource("langchain4j");
```
langchain4j: could you please explain what is the purpose of this field?

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -0,0 +1,426 @@
...
+        for (IndexingResult indexingResult : indexingResults) {
+            if (!indexingResult.isSucceeded()) {
+                log.error("Failed to add embedding: {}", indexingResult.getErrorMessage());
+            } else {
+                log.info("Added embedding: {}", indexingResult.getKey());
+            }
+        }
+    }
+
+    private List<Float> floatsArrayToList(float[] floats) {
```
langchain4j: can be removed

ghost (commented):
@jdubois great job, thank you!

langchain4j-azure-ai-search/pom.xml
jdubois: This was just a bad copy/paste, I removed it

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
jdubois: Indeed, that's much better!! Fixed.

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
jdubois: You are right, it's done!

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
jdubois: I would guess so, but I couldn't find it in the documentation, so migrating to "continue" to be safe

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
jdubois: I removed it

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
jdubois: yes

langchain4j-azure-open-ai/pom.xml
jdubois: I'm guessing you mean the version of the dependency... Removed!

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
jdubois: The "metadata source" is configured in the index in other languages (like JavaScript, in the commit linked at the top of this issue), but it doesn't seemed to be used. As I guessed it was the source of the data, I put "langchain4j" but indeed that's useless.
I've removed that line, but kept the field in the index, to have the same index configuration as in the other languages.

jdubois:
@langchain4j thanks for the review! It should be all good!

ghost (approved):
@jdubois awesome job, thank you!

PR #531, opened 2024-01-19
Allowing model builders to be extended and to be provided by `ServiceLoader`s

edeandrea:
This is a small prototype based on discussions originating from https://github.com/ai-for-java/openai4j/pull/13

The approach I took here is to allow for decorating the models/builders with additional functionality without having to extend model classes or builders. I did it for a single model in this prototype - the `OpenAiChatModel`, but this pattern could be applied to all of the other models across Langchain4J.

That doesn't mean you couldn't extend the model classes if you wanted to use inheritance. I just try to avoid it and use composition instead.

I also added a test which shows how it would be used. Downstream libraries (like Spring Boot or Quarkus) could use this mechanism to extend/enhance with their own capabilities which aren't necessarily part of the model.

Let me know what you think @geoand / @langchain4j !

Happy to continue conversation and see where we can bring this!


edeandrea:
I've actually thought of a way I can clean this up a bit more. I'll do that on Monday.

edeandrea:
@geoand / @langchain4j I've done some cleanup today and simplified the usage a bit. The generics in the new classes `ModelBuilder` and `ModelCustomizer` are a bit messy, but doing that allows a nice clean build dsl.

Take a look at the `ModelBuilderCustomizerTest` class I added which shows how you could use it to customize the same kind of model, whereas the new tests I added in `OpenAiChatModelIT` show how you could use it to return a completely different model that does something different.

geoand:
I like the idea!

edeandrea:
> I like the idea!

As you can see from the changes I made in the  `OpenAiChatModel` class, they are pretty minimal changes which can be applied to all of the model classes across Langchain4J. It also keeps langchain agnostic to what the end user wants to do.

The other thing I thought about was to add a `ServiceLoader` lookup in the model's `builder()` method, like is done in the openai4j project (https://github.com/ai-for-java/openai4j/blob/main/src%2Fmain%2Fjava%2Fdev%2Fai4j%2Fopenai4j%2FOpenAiClient.java#L49). This would allow implementers to provide their own builders completely.

WDYT @geoand / @langchain4j ?

geoand:
> The other thing I thought about was to add a ServiceLoader lookup in the model's builder() method, like is done in the ai-for-java project. This would allow implementers to provide their own builders completely.

Right, that's another good option and IMHO probably more applicable to the use cases we have in mind, as I don't expect normal users would have to decorate the models.

edeandrea:
I can add that into this prototype today and then we can decide if we want to expand the pattern throughout langchain4j

edeandrea:
@geoand / @langchain4j I added the factory/service loader into the prototype. It didn't add a whole lot - just creation of a factory class and overriding the `builder()` method that lombok would normally generate.

This pattern could certainly be applied across the board to all the other models in langchain4j. It would be cross-cutting a small change in lots of places.

Let me know your thoughts. If everyone is on board I can go ahead and continue.

geoand:
+1 from me

edeandrea:
> +1 from me

I see there is a test suite, but that suite doesn't seem to get executed when building (`main.yaml` has `-DskipITs`). Since this would be a cross-cutting set of changes, is there a way to run the test suite to verify nothing broke?

I ran them locally in just this one module where my prototype is using my own openai key.

edeandrea:
@geoand / @langchain4j I've implemented this across the board, except for in the `langchain4j-bedrock` module. It uses this [experimental Lombok SuperBuilder](https://projectlombok.org/features/experimental/SuperBuilder) thing which doesn't allow for much customization. It generates its builder classes as `final` and generates `private` constructors.

The `final` on the class you can get around by re-declaring the class as non-final, but you can not re-declare a constructor. That blows things up.

I didn't have much of an appetite to re-write everything in this module to not use lombok, so I've left things alone. I _could_ run delombok on all these classes and use the result of that, but it isn't very pretty and I don't know anything about this module.

langchain4j:
Hi @edeandrea, thanks a lot! Sorry for the silence, will take a look at it today

langchain4j:
I am probably missing something, but I am not sure I understand the need for `curtomizeWith()` API. If the user want's to decorate the model with his/her own behaviour (e.g. disabling requests), he can do this by using a regular decorator/proxy pattern instead of implementing custom interfaces with complex generics.

edeandrea:
> I am probably missing something, but I am not sure I understand the need for `curtomizeWith()` API. If the user want's to decorate the model with his/her own behaviour (e.g. disabling requests), he can do this by using a regular decorator/proxy pattern instead of implementing custom interfaces with complex generics.

If they wanted to re-use the existing model class (say `OpenAiChatModel` for example) the only current way would be for them to create their own builder implementation and instantiate it. They wouldn't be able to use `OpenAiChatModel.builder()`, unless they used the service loader mechanism, which they may not want to do in certain use cases.

Using the `customizeWith` allows users flexibility to return the same `OpenAiChatModel` instance without necessarily having to override the class. It will give the user full access to the built model class using the same builder dsl that they used to build the model in the first place.

Unless maybe I'm not understanding what you mean when you say

> he can do this by using a regular decorator/proxy pattern

Are you saying something like this (as a trivial example):

```java
public class MyModel implements ChatLanguageModel {
  private final OpenAiChatModel model;
  private final boolean someParam;

  public MyModel(OpenAiChatModel model, boolean someParam) {
    this.model = model;
    this.someParam = someParam;
  }

  // Other methods implemented from ChatLanguageModel
}

OpenAiChatModel chatModel = OpenAiChatModel.builder()
  .logRequests(true)
  .logResponses(true)
  // other things
  .build();

MyModel model = new MyModel(chatModel);

// Do stuff with the model
```

langchain4j:
```
Are you saying something like this (as a trivial example):

public class MyModel implements ChatLanguageModel {
  private final OpenAiChatModel model;
  private final boolean someParam;

  public MyModel(OpenAiChatModel model, boolean someParam) {
    this.model = model;
    this.someParam = someParam;
  }

  // Other methods implemented from ChatLanguageModel
}

OpenAiChatModel chatModel = OpenAiChatModel.builder()
  .logRequests(true)
  .logResponses(true)
  // other things
  .build();

MyModel model = new MyModel(chatModel);

// Do stuff with the model
```

Yes this is what I mean.
This seems to be much more usual/common/intuitive approach than using a custom customizer in the builder, at least for me.

> Using the customizeWith allows users flexibility to return the same OpenAiChatModel instance without necessarily having to override the class.

The example in `ModelBuilderCustomizerTest` allows customization only inside the `buildInternal`. This example is too synthetic, I am not sure I see how users will want to use this.
```
        ...

        @Override
        protected ChatModel buildInternal(ChatModel customizedModel) {
            return ChatModel.builder().text(this.altText).build();
        }
```

The example in `OpenAiChatModelIT` allows for more customization, but it does not return the same type actually, it returns `ExtendedChatLanguageModel`. In this case wrapping `OpenAiChatModel` into a `ExtendedChatLanguageModel` would be much simpler IMHO.

edeandrea:
I think I see your point. I'll go ahead and remove the customizer stuff.

Do you think its worthwhile to have a base interface that all builders inherit from?

langchain4j:
> Do you think its worthwhile to have a base interface that all builders inherit from?

I am not sure TBH. What do you think?
If there is no common logic, does it make sense to introduce this dependency between all of them?

edeandrea:
I've gone ahead and and removed the common interface and also re-named this PR to more accurately describe the change its proposing.

Let me know if you'd like me to re-sync from `main` as well.

ghost (approved):
@edeandrea awesome, thank you!

geoand:
@edeandrea now that this is in, do you plan to utilize it in the Quarkus extension (which has been updated to the latest LangChain4j version)?

geoand:
I'm trying to get this to work for Quarkus Langchain4j OpenAI and I am failing hard.

@edeandrea do you have an example of how to use this SPI in Quarkus LangChain4j?

edeandrea:
As part of our conversation on https://github.com/quarkiverse/quarkus-langchain4j/issues/755 I had done something in https://github.com/edeandrea/quarkus-langchain4j/commit/e13e9732ac64eda8bc800938ce7c1941dbabccc9

It seems though in LangChain4j not all the model builders have `public` constructors. Maybe some have been added since I tackled this and not followed similar constructs?

geoand:
I don't know (or care) about the details, all I want to do is add some extra fields to the Quarkus version of the ChatLanguageModel :)

PR #533, opened 2024-01-22
Public Ollama Client + 2 additional methods

bidek:
Public Ollama Client

- list model method
- get model details method

### Motivation
In my research project, I'm using Langchain4j, as anyone should :)
From my research, it seems that this client code is in sync with the Ollama API, and it is the easiest and most maintainable code. So, in my project, I use Langchain4j, and it's backed by the Ollama provider. In my use case, I need to be able to list models and, in the future, even create one. Is it possible to make the OllamaClient code public?

I wrote some thoughts about OpenAPI in Ollama in this issue. https://github.com/jmorganca/ollama/issues/716#issuecomment-1904415711
 So, if Ollama developers consider adding an OpenAPI endpoint, I will be the first to make the OllamaClient package-private again.


langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaApi.java
```diff
@@ -30,4 +27,12 @@ interface OllamaApi {
     @Headers({"Content-Type: application/json"})
     @Streaming
     Call<ResponseBody> streamingChat(@Body ChatRequest chatRequest);
+
+    @GET("/api/tags")
+    @Headers({"Content-Type: application/json"})
+    Call<ModelListResponse> modelList();
```
langchain4j: @bidek could you please add a few integration tests for these 2 new endpoints?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
```diff
@@ -172,6 +172,32 @@ public EmbeddingResponse embed(EmbeddingRequest request) {
...
+                return retrofitResponse.body();
+            } else {
+                throw toException(retrofitResponse);
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public ModelDetailsResponse modelDetails(ModelDetailsRequest modelDetailsRequest) {
```
langchain4j: I am wondering: should we expose these 2 new endpoints the same way as in https://github.com/langchain4j/langchain4j/pull/514/files#diff-cf3a590009831299f0335fe9bd4358e38e5c862a531bf6d046a310daa0843266R64 ?

ghost (commented):
@bidek thanks a lot!

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaApi.java
bidek: Added tests for the new endpoints

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
bidek: Assuming that `MistralAiModels` represents the API management component, I exposed the new endpoints similarly. My only concern is with the use of the `Response` class, given its model-centric documentation. Despite this doubt, I appreciate the concept of having a dedicated class for the management API.

bidek: And, I forgot to mention, because the client code is in a package-scoped context, requests and responses are also package-scoped. In the linked code from Mistral, requests and responses are package-scoped, but their components, like `MistralAiModelCard`, are public. I believe some clarification is needed here. Should certain parts of the client code be public? In the Mistral example, only a part of the response is used in `MistralAiModels`. However, in `OllamaModels`, all information from `ShowModelInformationResponse` is necessary. To address this, I introduced `OllamaModelInfo`, which is essentially another name for `ShowModelInformationResponse` with public access. I'm not a huge fan of this solution, so I would greatly appreciate some guidance or clarification.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModels.java
```diff
@@ -0,0 +1,53 @@
...
+    public Response<OllamaModelInfo> fetchModelInfo(String modelName) {
+        ShowModelInformationResponse response = withRetry(() -> client.showInformation(
+                ShowModelInformationRequest.builder()
+                        .name(modelName)
+                        .build()
+        ), maxRetries);
+        return Response.from(mapToModelInformation(response));
+    }
+
+    private OllamaModelInfo mapToModelInformation(ShowModelInformationResponse showInformationResponse) {
```
langchain4j: I would just rename `ShowModelInformationResponse` into `OllamaModelInfo` (or even `OllamaModelCard` to be consistent with mistral) and make it `public` instead of duplicating it

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModels.java
```diff
@@ -0,0 +1,53 @@
...
+                .build();
+        this.maxRetries = getOrDefault(maxRetries, 3);
+    }
+
+    public Response<List<OllamaModel>> availableModels() {
+        ModelsListResponse response = withRetry(client::listModels, maxRetries);
+        return Response.from(response.getModels());
+    }
+
+    public Response<OllamaModelInfo> fetchModelInfo(OllamaModel ollamaModel) {
```
langchain4j:
```suggestion
    public Response<OllamaModelCard> modelCard(OllamaModel ollamaModel) {
```

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModels.java
```diff
@@ -0,0 +1,53 @@
...
+        return fetchModelInfo(ollamaModel.getName());
+    }
+
+    public Response<OllamaModelInfo> fetchModelInfo(String modelName) {
```
langchain4j:
```suggestion
    public Response<OllamaModelCard> modelCard(String modelName) {
```

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModels.java
bidek: Done as suggested

ghost (approved):
@bidek thank you!

Issue #518, opened 2024-01-17
[BUG] MilvusEmbeddingStore findRelevant error


igerdy:
MilvusEmbeddingStore 
Current logic: when the collection does not exist
Step 1: Create a Collection
Step 2: Create index

Problem 1: When the collection is created successfully but the index creation fails, the task fails. There is a collection without index in milvus. This collection can still be saved, but cannot be queried.
Have you considered improving the logic here?

Problem 2: When sharding is required, there is no relevant functional system. Have you considered improving the logic here?

langchain4j:
Hi @igerdy thank you for reporting!
Regarding problem 1, good point, this should be improved.
Regarding problem 2, could you please provide more details? I am not sure I udnerstand it. Thank you.

igerdy:
Regarding question 2, the milvus SDK supports passing in the partition name when inserting, and the search also supports searching from multiple partitions. We hope that this feature can be introduced and can be changed to meet data isolation. Of course, I am testing this part of the function locally and hope to share it with everyone.

igerdy:
Regarding question 2: Milvus's sdk supports inserting the name of the incoming partition, and the search supports searching from multiple partitions. I hope to introduce this feature for better data isolation.
refer to:
io.milvus.param.dml.InsertParam#partitionName
io.milvus.param.dml.SearchParam#partitionNames
I am developing and testing this part of the function locally.
Of course, after the tests pass, I also hope to submit these codes to make this open source library better.

langchain4j:
@igerdy sounds awesome! Looking forward for your PR, thank you!

igerdy:
Regarding problem 2, a PR has been submitted.
But is this a temporary solution?
In theory, core cannot be compatible with the features of all stores.
Consider whether different stores create their own interfaces, but all extend the core interface.

langchain4j:
Hi @igerdy thanks for the PR!

> Consider whether different stores create their own interfaces, but all extend the core interface.

Indeed, I would support this option. This new partition-aware method should be only in `MilvusEmbeddingStore`.

PR #534, opened 2024-01-23
milvus supports partition operations

igerdy:
adds: 
 - The embedding store support inserting into specified partition.
  - The embedding store support searching from multiple partitions. 
 This feature is currently only supported by the milvus store. Milvus partitions belong to collections.

#518


langchain4j-core/src/main/java/dev/langchain4j/retriever/Retriever.java
```diff
@@ -30,4 +30,18 @@ default List<T> findRelevant(
...
+     * Find relevant items for the given text and partitions.
+     * Currently only milvus is supported.
+     *
+     * <p>Default implementation throws an exception.
+     *
+     * @param text          the text to search for.
+     * @param partitions    milvus collection partition list.
+     * @return  the list of relevant items.
+     */
+    default List<T> findRelevant(String text, List<String> partitions) {
```
langchain4j: I am not sure `Retriever` should be aware of partitions. `Retriever` is a much more generic concept that `EmbeddingStore`. `Retriever` can retrieve not only from `EmbeddingStore`, but from any data source. So the concept of partitions does not belong on this level.
`Retriever` is also supposed to be called in an automatic manned from `AiServices` or `ConversationalRetrievalChain`, both of which should be agnostic of partitions.

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -113,4 +113,40 @@ default List<EmbeddingMatch<Embedded>> findRelevant(
...
+     * <p>Default implementation throws an exception.
+     *
+     * @param referenceEmbedding The embedding used as a reference. Returned embeddings should be relevant (closest) to this one.
+     * @param maxResults         The maximum number of embeddings to be returned.
+     * @param partitionNames     The range of partitions used for querying, Can be empty.
+     * @return A list of embedding matches.
+     * Each embedding match includes a relevance score (derivative of cosine distance),
+     * ranging from 0 (not relevant) to 1 (highly relevant).
+     */
+    default List<EmbeddingMatch<Embedded>> findRelevant(Embedding referenceEmbedding, int maxResults, List<String> partitionNames) {
```
langchain4j: I would suggest placing this method on the `MilvusEmbeddingStore` as this is not necessarily applicable to all other implementations.

igerdy: I agree with your suggestion. 
My local version is under the milvus package: extending EmbeddingStore to MilvusEmbeddingStore, and then implementing MilvusEmbeddingStore.
Is this solution feasible? I'm doing local testing and will submit a PR after passing it.

langchain4j-core/src/main/java/dev/langchain4j/retriever/Retriever.java
igerdy: I agree with your point of view.
I need to ask which of the following two options is more suitable? Are there any other more reasonable options?
In the Milvus package:
### options 1：

> step 1：

```
pom.xml add
        <dependency>
            <artifactId>langchain4j</artifactId>
        </dependency>
```
> step 2：
```
public interface MilvusRetrieverExtend<T> extends Retriever<T> {
    List<T> findRelevant(String text, List<String> partitions);
}
```
> step 3：
```
public class MilvusRetriever extends EmbeddingStoreRetriever implements MilvusRetrieverExtend<TextSegment> {
    ......
    public MilvusRetriever(EmbeddingStore<TextSegment> embeddingStore, EmbeddingModel embeddingModel, int maxResults, Double minScore) {
        super(embeddingStore, embeddingModel, maxResults, minScore);
    }

    @Override
    public List<TextSegment> findRelevant(String text, List<String> partitions) {
           ......
    }
    ......
}
```

### options 2：
> step 1：
```
public interface MilvusEmbeddingStoreRetrieverExtend<T> extends Retriever<T> {

    default List<T> findRelevant(String text) {
        return findRelevant(text, Collections.emptyList());
    }

    List<T> findRelevant(String text, List<String> partitions);

}
```
> step 2：
```
public class MilvusEmbeddingStoreRetriever implements MilvusEmbeddingStoreRetrieverExtend<TextSegment> {
    ......
    public MilvusEmbeddingStoreRetriever(EmbeddingStore<TextSegment> embeddingStore,
                                   EmbeddingModel embeddingModel,
                                   int maxResults,
                                   Double minScore) {
        this.embeddingStore = embeddingStore;
        this.embeddingModel = embeddingModel;
        this.maxResults = maxResults;
        this.minScore = minScore;
    }

    @Override
    public List<TextSegment> findRelevant(String text, List<String> partitions) {
           ......
    }
    ......
}

langchain4j: I need to understand your use case better. How/where/when do you plan to decide which partition should be user for search? Will this be done statically or dynamically?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
langchain4j: @igerdy yes, this makes sense!

igerdy: okay, I'll submit this part of the code first.

langchain4j-core/src/main/java/dev/langchain4j/retriever/Retriever.java
igerdy: I only use milvus, and I think most people only use a certain store.
I hope milvus can be extended based on langchain, Standardize everyone’s usage habits.
For upper-layer applications, I hope to provide enough tool classes, but not pay too much attention to the specific implementation.
I plan to make it possible for each query to use a different partition range, or not specify partitions, by statically constructing and creating the collection join at query time.

PR #537, opened 2024-01-23
 Add Qwen multi-modal models implementation

jiangsier-xyz:
1. Add Qwen multi-modal models implementation, support for image and text information.

Example of key input fragment:
```json
"messages": [
    {
        "role": "user",
        "content": [
            {"image": "https://dashscope.oss-cn-beijing.aliyuncs.com/images/dog_and_girl.jpeg"},
            {"text": "What animal is in the picture?"}
        ]
    }
]
```

Example of key output fragment:
```json
"output": {
    "choices": [
        {
            "finish_reason": null,
            "message": {
                "role": "assistant",
                "content": [
                    {
                        "text": "The picture shows a dog sitting on the beach with its owner."
                    }
                ]
            }
        }
    ]
}
```

Note: The dashscope sdk supports local file URL starting with file://...

2. Add new optional parameters: baseUrl, maxTokens


langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/ImageFileTypeDetector.java
```diff
@@ -0,0 +1,29 @@
...
+ * (which might be sun.nio.fs.DefaultFileTypeDetector) performs poorly.
+ * Here, we are using java.net.URLConnection.guessContentTypeFromStream() to identify the mimeType,
+ * which yields very accurate results for image-type files.
+ */
+public class ImageFileTypeDetector extends FileTypeDetector {
+    @Override
+    public String probeContentType(Path path) {
+        try (InputStream in = new BufferedInputStream(
+                Files.newInputStream(path, StandardOpenOption.READ))) {
+            return URLConnection.guessContentTypeFromStream(in);
```
langchain4j: Nice! I should probably do the same in vertex module

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
```diff
@@ -36,10 +41,112 @@ static List<Message> toQwenMessages(Iterable<ChatMessage> messages) {
...
+                .content(toSingleText(message))
                 .build();
     }
 
+    static String toSingleText(ChatMessage message) {
+        switch (message.type()) {
+            case USER:
+                UserMessage userMessage = (UserMessage) message;
+                return userMessage.hasSingleText() ?
+                        ((TextContent) userMessage.contents().get(0)).text() : "";
```
langchain4j: should it concatenate all text contents if there are more than 1? (instead of returning empty string)
or throw an exception if it is not allowed

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
```diff
@@ -65,13 +172,35 @@ static String answerFrom(GenerationResult result) {
...
                 .map(GenerationResult::getUsage)
                 .map(usage -> new TokenUsage(usage.getInputTokens(), usage.getOutputTokens()))
                 .orElse(new TokenUsage(null, null));
     }
 
+    static TokenUsage tokenUsageFrom(MultiModalConversationResult result) {
+        return Optional.of(result)
+                .map(MultiModalConversationResult::getUsage)
+                .map(usage -> new TokenUsage(usage.getInputTokens(), usage.getOutputTokens()))
+                .orElse(new TokenUsage(null, null));
```
langchain4j: please return `null` instead of `new TokenUsage(null, null)`

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
```diff
@@ -65,13 +172,35 @@ static String answerFrom(GenerationResult result) {
...
+                .map(content -> content.get("text"))
+                .map(String.class::cast)
+                .orElse("");
+    }
+
     static TokenUsage tokenUsageFrom(GenerationResult result) {
         return Optional.of(result)
                 .map(GenerationResult::getUsage)
                 .map(usage -> new TokenUsage(usage.getInputTokens(), usage.getOutputTokens()))
                 .orElse(new TokenUsage(null, null));
```
langchain4j: please return `null` instead of new `TokenUsage(null, null)`

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
```diff
@@ -65,13 +172,35 @@ static String answerFrom(GenerationResult result) {
...
+                .map(MultiModalConversationOutput::getChoices)
+                .filter(choices -> !choices.isEmpty())
+                .map(choices -> choices.get(0))
+                .map(MultiModalConversationOutput.Choice::getMessage)
+                .map(MultiModalMessage::getContent)
+                .filter(contents -> !contents.isEmpty())
+                .map(contents -> contents.get(0))
+                .map(content -> content.get("text"))
+                .map(String.class::cast)
+                .orElse("");
```
langchain4j: should throw exception in this case?

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
```diff
@@ -90,4 +219,23 @@ static FinishReason finishReasonFrom(GenerationResult result) {
...
     }
+
+    static FinishReason finishReasonFrom(MultiModalConversationResult result) {
+        String finishReason = Optional.of(result)
+                .map(MultiModalConversationResult::getOutput)
+                .map(MultiModalConversationOutput::getChoices)
+                .filter(choices -> !choices.isEmpty())
+                .map(choices -> choices.get(0))
+                .map(MultiModalConversationOutput.Choice::getFinishReason)
+                .orElse("");
```
langchain4j, line 236: same here?

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenMultiModalChatModel.java
```diff
@@ -0,0 +1,103 @@
...
+import dev.langchain4j.model.dashscope.spi.QwenMultiModalChatModelBuilderFactory;
+import dev.langchain4j.model.output.Response;
+import dev.langchain4j.spi.ServiceHelper;
+import lombok.Builder;
+
+import java.util.List;
+
+import static dev.langchain4j.model.dashscope.QwenHelper.*;
+
+public class QwenMultiModalChatModel implements ChatLanguageModel {
```
langchain4j: Hmm... In OpenAI and Vertex integrations we use the same model class for text and multimodal. Do you think it is necessary to have this specific model class?

langchain4j-dashscope/src/main/resources/META-INF/services/java.nio.file.spi.FileTypeDetector
```diff
@@ -0,0 +1 @@
+dev.langchain4j.model.dashscope.ImageFileTypeDetector
```
langchain4j: Why is this required?

langchain4j-dashscope/src/test/java/dev/langchain4j/model/dashscope/QwenMultiModalChatModelIT.java
```diff
@@ -0,0 +1,51 @@
...
+import static dev.langchain4j.model.dashscope.QwenTestHelper.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class QwenMultiModalChatModelIT {
+    @ParameterizedTest
+    @MethodSource("dev.langchain4j.model.dashscope.QwenTestHelper#multiModalChatModelNameProvider")
+    public void should_send_image_url_and_receive_response(String modelName) {
+        String apiKey = apiKey();
+        if (Utils.isNullOrBlank(apiKey)) {
+            return;
```
langchain4j: Could you please use `@EnabledIfEnvironmentVariable(named = "QWEN_API_KEY", matches = ".+")` above the test class instead? With current approach the tests will be green even if they did not really run

langchain4j-dashscope/src/test/java/dev/langchain4j/model/dashscope/QwenMultiModalStreamingChatModelIT.java
```diff
@@ -0,0 +1,96 @@
...
+            return;
+        }
+
+        StreamingChatLanguageModel model = QwenMultiModalStreamingChatModel.builder()
+                .apiKey(apiKey)
+                .modelName(modelName)
+                .build();;
+
+        CompletableFuture<Response<AiMessage>> future = new CompletableFuture<>();
+        model.generate(multiModalChatMessagesWithImageUrl(), new StreamingResponseHandler<AiMessage>() {
```
langchain4j: Please use `TestStreamingResponseHandler` instead of creating anonymous handlers.
You'll need to add:
```
        <dependency>
            <classifier>tests</classifier>
            <type>test-jar</type>
            <scope>test</scope>
        </dependency>
```

ghost (dismissed):
@jiangsier-xyz awesome, thank you!

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
jiangsier-xyz: I once thought that the Qwen models would return a response to further confirm the user's intent, when it receives an empty string. However, it turns out this is not the case after actual testing. Furthermore, I've noticed that langchain4j's TextContent also does not allow empty strings. So an exception should indeed be thrown here.

Concatenating all the text (using '\n') also make sense. BTW, I recommend that UserMessage retains the text() method, which also concatenates TextContent and returns it, so that all ChatMessages can have a consistent interface and avoid having many switch-cases in the application which depends on langchain4j.

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
jiangsier-xyz: 👌

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
jiangsier-xyz: 👌

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
jiangsier-xyz: 👌

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
jiangsier-xyz: The finish-reason can be null. This empty string will be handled in the switch-cases below (ultimately returning null). The purpose of using an empty string is to avoid throwing a NullPointerException within the switch-cases. The reason for using switch-cases instead of if-else is due to the consideration that there may be more finish reasons beyond stop and length, and switch-cases would be neater in such cases.

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenMultiModalChatModel.java
jiangsier-xyz: I was considering two aspects:
1. DashScope provides two different top-level APIs for traditional language models and multimodal language models (Generation vs. MultiModalConversation).
2. The two sets of models receive different parameters (which means they may require different "Builders").

Regarding point 1, I think it can be managed. Ugly for implementers, simple for users.

Regarding point 2, by comparing QwenParam and MultiModalConversationParam, the parameters have a superset and subset relationship. In this case, I believe it is possible to abstract into "same model". If in the future the model supports parameters that diverge (for example, MultiModalConversationParam receives some tuning parameters related to images and videos), this would actually necessitate designing different builders, and refactoring at that point might significantly impact the users.

However, my personal judgment is that traditional language models that do not support multimodality (and the corresponding APIs in DashScope) will not evolve, gradually disappear, and will eventually be completely replaced by multimodal language models. So, OK, let's keep a single model implementation.

langchain4j-dashscope/src/test/java/dev/langchain4j/model/dashscope/QwenMultiModalChatModelIT.java
jiangsier-xyz: 👌

langchain4j-dashscope/src/test/java/dev/langchain4j/model/dashscope/QwenMultiModalStreamingChatModelIT.java
jiangsier-xyz: 👌

langchain4j-dashscope/src/main/resources/META-INF/services/java.nio.file.spi.FileTypeDetector
jiangsier-xyz: The dashscope sdk directly uses java.nio.file.Files.probeContentType() to obtain the mimeType of a file. It relies on SPI to inject a specific FileTypeDetector implementation, otherwise it defaults to using sun.nio.fs.DefaultFileTypeDetector, which performs poorly.

The purpose of this file is to inject my own implementation of FileTypeDetector through SPI. It will be called by java.nio.file.Files.probeContentType.

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenMultiModalChatModel.java
langchain4j: I completely agree with your judgement! Feel free to do either way, there seems to be no "right" or "wrong", especially at this point of time.

langchain4j-dashscope/src/main/java/dev/langchain4j/model/dashscope/QwenHelper.java
langchain4j: Yeah i think i prematurely annotated text() as @Deprecated, probably we should still keep it, at least while multimodal models are still rare. The intent was to force the user to use message-type-specific accessors, because fields that each message type contains diverge more and more and I am not sure it is a good idea to keep accessing all specific fields via the same getter. Moreover, user message becomes multimodal and aimessage will probably become soon too.

langchain4j-dashscope/src/main/resources/META-INF/services/java.nio.file.spi.FileTypeDetector
langchain4j: I see, thank you for explanation

ghost (approved):
@jiangsier-xyz great job, thank you!

PR #545, opened 2024-01-24
Fix illegal group reference when the value contains $$

ilopezluna:
Following [this answer](https://stackoverflow.com/a/20002683/1490806) `Matcher.quoteReplacement` has to be used to avoid `java.lang.IllegalArgumentException: Illegal group reference` whenever the content contains `$$`


ghost (approved):
@ilopezluna thank you a lot!

PR #549, opened 2024-01-25
Adding disabled implementation of all model interfaces

edeandrea:
Adding a default set of model implementations that essentially "disable" the model. All the operations throw a `ModelDisabledException`.

These can be used in tests as well as by other frameworks/libraries which extend Langchain4J where maybe users want to "not" make live calls when deploying into a certain environment for whatever reason.

For example, I plan to take this in the quarkus-langchain4j extension and based on user runtime configuration apply these models rather than other custom ones that Quarkus uses.


langchain4j:
Hi @edeandrea thanks a lot for your contribution! I will check this next week after the release.

edeandrea:
Once this is merged & released then quarkiverse/quarkus-langchain4j#130 will consume it.

edeandrea:
Hello. Just wanted to give a gentle ping on this to see about getting it merged.

ghost (approved):
@edeandrea thanks!

edeandrea:
Thank you!

PR #556, opened 2024-01-26
Update Vertex AI embedding to allow a batch size of 250

glaforge:


ghost (approved):

glaforge:
I just discovered a regression caused by my PR.
I tried to embed long snippets, and I got an exception telling me there's a limit of 20k tokens per call.
I wasn't aware of such limit per call, I'm sorry.

I'm curious what would be the best approach to solve this, and would be interested in your advice.
* First of all, I was thinking that we could make batch size a parameter to pass to the builder. We could change the default value from 250 to 50 for example, but allow people to chose a different value. I believe it's good to have that ability, but it doesn't prevent the problem to appear if you have too long snippets that total up 20k tokens.
* The better / complementary approach would be to track the token count, and stop adding once we're too close to the limit. That way, we'd have different batch sizes depending on the total number of tokens of the current batch.

Thoughts?

langchain4j:
@glaforge is there a tokenizer available for Vertex AI models from java?

glaforge:
There is a countToken [endpoint](https://cloud.google.com/vertex-ai/docs/generative-ai/get-token-count#textembedding-gecko) that can be called. It can be called in batch too, but I don't (yet) know the limits.
But imagine calling it, that'd double the REST calls?

In the current implementation, we accumulate the tokens used after each batch.
Maybe we should look at that number, and when we're 10% close to the limit, we don't add to that batch?

PR #567, opened 2024-01-30
extend MilvusEmbeddingStore

igerdy:
adds:
 - supports inserting by partition
 - supports inserting by search

### DIFF
--- a/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStore.java
+++ b/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStore.java
@@ -25,14 +25,17 @@ import io.milvus.param.MetricType;
 import io.milvus.param.dml.InsertParam;
 import io.milvus.param.dml.SearchParam;
 import io.milvus.response.SearchResultsWrapper;
+import org.apache.commons.lang3.StringUtils;
+
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 /**
  * Represents an <a href="https://milvus.io/">Milvus</a> index as an embedding store.
  * Does not support storing {@link dev.langchain4j.data.document.Metadata} yet.
  */
-public class MilvusEmbeddingStore implements EmbeddingStore<TextSegment> {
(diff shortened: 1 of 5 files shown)


langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStore.java
```diff
@@ -25,21 +25,25 @@
...
-public class MilvusEmbeddingStore implements EmbeddingStore<TextSegment> {
+public class MilvusEmbeddingStore implements MilvusEmbeddingStoreExtend<TextSegment> {
 
   static final String ID_FIELD_NAME = "id";
   static final String TEXT_FIELD_NAME = "text";
   static final String VECTOR_FIELD_NAME = "vector";
 
   private final MilvusServiceClient milvusClient;
   private final String collectionName;
+  private final String partitionName;
```
langchain4j: I am not sure we should include `partitionName` as a parameter of `MilvusEmbeddingStore`.
As far as I understand, `partitionName` is accepted by the constructor with a single purpose: to create a partition. `partitionName` is not used later on in the lifecycle of `MilvusEmbeddingStore`. I suppose it will be more intuitive to provide a method `public void createPartition(StringPartitionName)` instead.

langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStoreExtend.java
```diff
@@ -0,0 +1,46 @@
...
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.store.embedding.EmbeddingMatch;
+import dev.langchain4j.store.embedding.EmbeddingStore;
+
+import java.util.List;
+
+/**
+ * Milvus Embedding Store extend
+ */
+public interface MilvusEmbeddingStoreExtend<Embedded> extends EmbeddingStore<Embedded> {
```
langchain4j: Is this interface really required? Can't you just define these methods in `MilvusEmbeddingStore`?

langchain4j (commented):
@igerdy thanks a lot, great job! I have just 2 comments, please check.

PR #569, opened 2024-01-30
set default lifecycle plugins versions

hboutemy:
avoid using versions from Maven super-pom, that vary based on mvn version used

notice that GH workflows don't use `./mvnw` that is configured with Maven 3.9.6 (that provide quite recent plugins) but use default Ubuntu one that is much older (3.6.3?), with much older plugins


ghost (approved):
@hboutemy thanks a lot!

PR #572, opened 2024-01-30
Beautifying Maven output

agoncal:
Looking at the Maven output I thought it could benefit from a little renaming. I just changed the `<name>` in the `pom.xml`, nothing more. The output is like this at the moment:

Screenshot 2024-01-30 at 16 26 53

It could look like this if this PR is merged:

Screenshot 2024-01-30 at 16 42 38

Just a personal taste. Let me know if you like it or not (or want to change it). If not, just discard it, it's fine ;o)


ghost (approved):
@agoncal thank you! It is indeed more readable this way :)

PR #573, opened 2024-01-30
Update language models and tests for Azure OpenAI

jdubois:
This PR update the language models for Azure OpenAI, and improves some tests:

- It updates the current models available on Azure OpenAI, removing those which are not available like `gpt-4-0314`, `gpt-3.5-turbo-0301`
- It creates some upcoming models as described in https://openai.com/blog/new-embedding-models-and-api-updates
- It passes those updated model names to the integration tests and fixes some issues there: I have no idea why we were testing for "length" and not for "stop", so now we do both correctly.
- Also the Dall E 3 test was "disabled", so I enabled it again.


langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiModelName.java
```diff
@@ -4,30 +4,31 @@ public class AzureOpenAiModelName {
 
     // Use with AzureOpenAiChatModel and AzureOpenAiStreamingChatModel
     public static final String GPT_3_5_TURBO = "gpt-3.5-turbo"; // alias for the latest model
-    public static final String GPT_3_5_TURBO_0301 = "gpt-3.5-turbo-0301"; // 4k context
     public static final String GPT_3_5_TURBO_0613 = "gpt-3.5-turbo-0613"; // 4k context, functions
+    public static final String GPT_3_5_TURBO_0125 = "gpt-3.5-turbo-0125"; // 4k context, functions
```
langchain4j, line 8: You might want to split this class into multiple enums: one per model type, like OpenAiChatModelName, OpenAiEmbeddingModelName, etc. These enums can be then accepted by model builder for convenience (in addition to strings).

ghost (approved):
@jdubois awesome, thank you!

PR #513, opened 2024-01-15
Add support for nested custom types in structured output

ldmajewski:
The current implementation allows only simple types of structured output. I've added support for nested custom types.

PR #575, opened 2024-01-31
nested POJO extract

tenpigs267:
This PR enhance ServiceOutPutParser allowing outputFormatInstructions to document nested objects in jsonStructure.

Integration tests have been modified to add a nested address object. Maybe a dedicated test would be better?


mike-adonis (commented):
Looks good.

mike-adonis (approved):

tenpigs267:
I didn't seen there was already a PR (#513) for nested objects.
I think this version is simpler.

ghost (approved):
@tenpigs267 great job, thank you!

langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
```diff
@@ -123,7 +123,9 @@ private static String jsonStructure(Class<?> structured) {
         jsonSchema.append("{\n");
         for (Field field : structured.getDeclaredFields()) {
             String name = field.getName();
-            if (name.equals("__$hits$__")) {
+            if (name.equals("__$hits$__")
+                    || java.lang.reflect.Modifier.isStatic(field.getModifiers())
+                    || java.lang.reflect.Modifier.isFinal(field.getModifiers())) {
```
geoand, line 128: Unfortunately this line is severely breaking. 

Consider `ServiceOutputParserTest.Person`: By simply making is class immutable (and adding a constructor so JSON deserialization works), `outputFormatInstructions` will not return `{\n}`.
Furthermore, the same problem occurs for Java Records.

I am not sure what the motivation for this line was, but I propose at the very least, it should be removed.

langchain4j: @geoand thanks for the heads up! Does it break Quarkus extension?

langchain4j: @tenpigs267 BTW it also does not work if you have a recursion. So if I define it like this:

```java
class Person {

            private String firstName;
            private String lastName;
            private LocalDate birthDate;
            private List<Person> parents;
```

It will go into endless recursion.

geoand: > @geoand thanks for the heads up! Does it break Quarkus extension?

It breaks some of our tests that use Java Records

tenpigs267: thanks @langchain4j for #620 and sry for that.
For the recursive thing I'am going to add a PR with the code going through all involved classes and generating a list a class (and test it with openAI)

PR #579, opened 2024-01-31
Add `langchain4j-embeddings-all-minilm-l6-v2-q` to the BOM

boris-petrov:
Which seems needed as otherwise the version is not right:

https://mvnrepository.com/artifact/dev.langchain4j/langchain4j-bom/0.26.1

You can see that the `langchain4j-embeddings-all-minilm-l6-v2-q` artifact is referenced by version 0.25.0 instead of 0.26.1.

I'm not sure this is the right fix here?

cc @langchain4j, @Martin7-1

### DIFF
--- a/langchain4j-bom/pom.xml
+++ b/langchain4j-bom/pom.xml
@@ -104,7 +104,7 @@
                 <artifactId>langchain4j-mistral-ai</artifactId>
                 <version>${project.version}</version>
             </dependency>
-          
+
             <dependency>
                 <groupId>dev.langchain4j</groupId>
                 <artifactId>langchain4j-cohere</artifactId>
--- a/langchain4j-bom/pom.xml
+++ b/langchain4j-bom/pom.xml
@@ -203,6 +203,14 @@
                 <version>${project.version}</version>
             </dependency>
 
+            <!-- embeddings -->
+
+            <dependency>
+                <groupId>dev.langchain4j</groupId>
+                <artifactId>langchain4j-embeddings-all-minilm-l6-v2-q</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+
             <!-- code execution engines -->
 
             <dependency>


Martin7-1:
@boris-petrov Hi, thanks for your PR. Maybe we should change the version of `langchain4j-embeddings-all-minilm-l6-v2-q` from 0.25.0 to `${project.version}` in `langchain4j-parent`, because `langchain4j-parent` do not import `langchain4j-bom`

boris-petrov:
@Martin7-1 thanks for the help! I've done this. Should I remove the commit that adds `langchain4j-embeddings-all-minilm-l6-v2-q` to the BOM in that case?

langchain4j:
Hi @boris-petrov, the reason behind using an older version was to simplify the release process as `langchain4j-embeddings-all-minilm-l6-v2-q` lives in a separate repo. Having `${project.version}` would require first releasing `langchain4j-core`, then `langchain4j-embeddings-all-minilm-l6-v2-q`, then wait for unknown amount of time until `langchain4j-embeddings-all-minilm-l6-v2-q` becomes available on maven central and only then continue with releasing the rest of the modules in the main repo.

Maybe there is a way to use older version for testing and have latest version in the bom?
Thanks!

boris-petrov:
@langchain4j sorry, I can't really help with that. Perhaps of @Martin7-1 has any ideas?

Martin7-1:
> @langchain4j sorry, I can't really help with that. Perhaps of @Martin7-1 has any ideas?

Having the latest version in the bom will not affect the test version because bom is used for users to manage artifacts version but not used in our repo.

So the easiest way is just add `langchain4j-embeddings-all-minilm-l6-v2-q` to the bom, so that users can use it like other artifacts in langchain4j. This way will not affect tests to use older version of it.

boris-petrov:
Done. @Martin7-1 please approve if this is all that's needed.

@langchain4j after that this is good to go I guess?

boris-petrov:
Ping @Martin7-1, @langchain4j. :)

ghost (approved):
Thanks a lot and sorry for delay!

PR #581, opened 2024-01-31
langchain4j/model.output coverage tests.

crutcher:
Noticed weirdness in the errors for DateOutputParser and EnumOutputParser.

Expanded EnumOutputParser to be case-insensitive.
Restricted DateOutputParser to the expected format.


ghost (approved):
@crutcher good catch, thank you a lot! BTW `AiServicesIT.test_extract_date` and `test_extract_time` are failing, I will fix that myself to speedup and include this in the new release (today).

PR #591, opened 2024-02-02
Add IT for Azure Blob Storage with Testcontainers

eddumelendez:
Use Azurite image to start Azure Blob Storage service with Testcontainers.


document-loaders/langchain4j-document-loader-azure-storage-blob/src/test/java/dev/langchain4j/data/document/loader/azure/storage/blob/LocalAzureBlobStorageDocumentLoaderIT.java
```diff
@@ -0,0 +1,88 @@
...
+
+        assertThat(documents.get(1).text()).isEqualTo(TEST_CONTENT);
+        assertThat(documents.get(1).metadata().asMap()).hasSize(4);
+        assertThat(documents.get(1).metadata("source")).endsWith("/test-file.txt");
+    }
+
+    @AfterAll
+    public static void afterAll() {
+        blobServiceClient.getBlobContainerClient(TEST_CONTAINER).delete();
+    }
```
eddumelendez, line 87: I changed it from `AfterEach` to `AfterAll` from the original IT because the container will be deleted and the second test will fail because there is no container.

ghost (approved):
@eddumelendez thanks a lot!

PR #598, opened 2024-02-03
Fix JSON schema generation and prompt requirement

sarxos:
Hello @langchain4j Team :)

At the beginning, I would like to thank you for this awesome project! It rocks!

---

This change introduces an additional requirement in the prompt to reply with a plain text valid JSON and also fixes the schema generation by removing unwanted comma at the end of the last field descriptor in the JSON structure generator.

# Background

I was playing with langchain4j (specifically the quarkus-langchain4j extension) using Azure as the backend, and gpt-4-turbo (specifically 1106 preview) and experienced the following problems.
1. GPT replied with JSON wrapped in markdown causing JSON parser failure,
2. GPT replied with invalid JSON because it had an additional comma at the end of the last field descriptor.

Just FYI, it worked well with pure gpt-4, __not__ turbo.

# Analysis

Problem 1 is a result of gpt-4-turbo being simply less smart than a regular gpt-4. It requires more detailed prompts to do the job well.

Problem 2 is partially a side effect of 1, but is primarily caused by the bug in the JSON structure schema generator which includes an unwanted comma at the end of the JSON structure schema. Pure GPT-4 is smart enough to understand that this is wrong and fix it on its own without additional instructions, but the less smart gpt-4-turbo takes it as it is and includes this unwanted comma in the JSON output. This makes it invalid.

# Stack Trace

```plain
Caused by: com.fasterxml.jackson.core.JsonParseException: Unexpected character ('`' (code 96)): expected a valid value (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (String)"```json
{
"translatedTextLines": [
"Lithuania, my homeland! You are like health;",
"How much you must be valued, will only discover",
]
}
```"; line: 1, column: 2]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:2477)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:750)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportUnexpectedChar(ParserMinimalBase.java:674)
```

Even when we ask it not to produce markdown but plain text, it will again fail because of the comma:

```plain
Caused by: com.fasterxml.jackson.core.JsonParseException: Unexpected character (']' (code 93)): expected a valid value (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (String)"{
"translatedTextLines": [
"Lithuania, my homeland! You are like health;",
"How much you must be valued, will only discover",
]
}"; line: 5, column: 2]
	at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:2477)
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:750)
```

# Solution

This PR is to fix both problem 1 and problem 2:
* It changes the prompt instruction to produce a valid and plain text JSON.
* It also fixes the structure generator so that the comma of death is not included at the end of the last field descriptor.

Previous prompt (example):

```plain
...
You must answer strictly in the following JSON format: {
"records": (type: array of string),
"name": (type: string),
"age": (type: integer),
}
```

New prompt, after the fix:

```plain
...
You must answer in plain text, with valid JSON, in strictly the following format:  {
"records": (type: array of string),
"name": (type: string),
"age": (type: integer)
}
```

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
@@ -18,8 +18,14 @@ import java.util.*;
 import static dev.langchain4j.exception.IllegalConfigurationException.illegalConfiguration;
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.joining;
 
 public class ServiceOutputParser {
+
+    /**
+     * A coverage instrumentation field name.
+     */
+    private static final String COVERAGE_FIELD_NAME = "__$hits$__";
 
     private static final Map<Class<?>, OutputParser<?>> OUTPUT_PARSERS = new HashMap<>();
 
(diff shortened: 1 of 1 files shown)


morki:
I have problem with new line character added before generated JSON even with this change with model `Mistral-7B-OpenOrca-GGUF`. Can we also trim this whispace or set `JsonReader.setLenient(true) `?

langchain4j:
@sarxos thanks a lot for your contribution and sorry for the delayed review!
Your solution makes total sense, but there is now a conflict, could you pelase check? 🙏

langchain4j:
@morki sure, could you please open a PR?

langchain4j:
Has been fied already in another PR, sorry.

Issue #597, opened 2024-02-03
[FEATURE] Add responseFormat to AzureOpenAiChatModel and AzureOpenAiStreamingChatModel


langchain4j:
Currently there is no way to enable [json mode](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/json-mode?tabs=python) in Azure OpenAI, as well as many other parameters are missing.
`AzureOpenAiChatModel` and `AzureOpenAiStreamingChatModel` should support the same parameters `OpenAiChatModel` supports, or even more (all supported parameters).

langchain4j:
@jdubois FYI https://github.com/quarkiverse/quarkus-langchain4j/pull/266

jdubois:
Thanks @langchain4j ! Can you check #573 as I would like to have this merged first? I'll start working on this.

jdubois:
@langchain4j PR is ready at #605

jdubois:
Sorry I forgot to automatically close this issue with #605 - @langchain4j can you close this?

PR #605, opened 2024-02-05
Add support for responseFormat and other parameters

jdubois:
As requested in #597

- Fix support for responseFormat to have the JSON format available in chat (non-streaming and streaming)
- Add a test to secure it
- Add support for other parameters that were not available earlier (all parameters should now be supported)


ghost (approved):
@jdubois awesome, thanks!

PR #610, opened 2024-02-07
EmbeddingStore (Metadata) Filter API

ghost:
## New EmbeddingStore (metadata) `Filter` API
Many embedding stores, such as [Pinecone](https://docs.pinecone.io/docs/metadata-filtering) and [Milvus](https://milvus.io/docs/boolean.md) support strict filtering (think of an SQL "WHERE" clause) during similarity search.
So, if one has an embedding store with movies, for example, one could search not only for the most semantically similar movies to the given user query but also apply strict filtering by metadata fields like year, genre, rating, etc. In this case, the similarity search will be performed only on those movies that match the filter expression.

Since LangChain4j supports (and abstracts away) many embedding stores, there needs to be an embedding-store-agnostic way for users to define the filter expression.

This PR introduces a `Filter` interface, which can represent both simple (e.g., `type = "documentation"`) and composite (e.g., `type in ("documentation", "tutorial") AND year > 2020`) filter expressions in an embedding-store-agnostic manner.

`Filter` currently supports the following operations:

- Comparison:
  - `IsEqualTo`
  - `IsNotEqualTo`
  - `IsGreaterThan`
  - `IsGreaterThanOrEqualTo`
  - `IsLessThan`
  - `IsLessThanOrEqualTo`
  - `IsIn`
  - `IsNotIn`

- Logical:
  - `And`
  - `Not`
  - `Or`

These operations are supported by most embedding stores and serve as a good starting point. However, the list of operations will expand over time to include other operations (e.g., `Contains`) supported by embedding stores.

Currently, the DSL looks like this:
```java
Filter onlyDocs = metadataKey("type").isEqualTo("documentation");

Filter docsAndTutorialsAfter2020 = metadataKey("type").isIn("documentation", "tutorial").and(metadataKey("year").isGreaterThan(2020));
// or
Filter docsAndTutorialsAfter2020 = and(
    metadataKey("type").isIn("documentation", "tutorial"),
    metadataKey("year").isGreaterThan(2020)
);
```

## Filter expression as a `String`
Filter expression can also be specified as a `String`. This might be necessary, for example, if the filter expression is generated dynamically by the application or by the LLM (as in [self querying](https://python.langchain.com/docs/modules/data_connection/retrievers/self_query/)).

This PR introduces a `FilterParser` interface with a simple `Filter parse(String)` API, allowing for future support of multiple syntaxes (if this will be required).

For the out-of-the-box filter syntax, ANSI SQL's `WHERE` clause is proposed as a suitable candidate for several reasons:
- SQL is well-known among Java developers
- There is extensive tooling available for SQL (e.g., parsers)
- LLMs are pretty good at generating valid SQL, as there are tons of SQL queries on the internet, which are included in the LLM training datasets. There are also specialized LLMs that are trained for text-to-SQL task, such as [SQLCoder](https://huggingface.co/defog).

The downside is that SQL's `WHERE` clause might not support all operations and data types that could be supported in the future by various embedding stores. In such case, we could extend it to a superset of ANSI SQL `WHERE` syntax and/or provide an option to express filters in the native syntax of the store.

An out-of-the-box implementation of the SQL `FilterParser` is provided as a `SqlFilterParser` in a separate module `langchain4j-embedding-store-filter-parser-sql`, using [JSqlParser](https://github.com/JSQLParser/JSqlParser) under the hood.

`SqlFilterParser` can parse SQL "SELECT" (or just "WHERE" clause) statement into a `Filter` object:
- `SELECT * FROM fake_table WHERE userId = '123-456'`  ->  `metadataKey("userId").isEqualTo("123-456")`
- `userId = '123-456'`  ->  `metadataKey("userId").isEqualTo("123-456")`

It can also resolve `CURDATE()` and `CURRENT_DATE`/`CURRENT_TIME`/`CURRENT_TIMESTAMP`:
`SELECT * FROM fake_table WHERE year = EXTRACT(YEAR FROM CURRENT_DATE`  ->  `metadataKey("year").isEqualTo(LocalDate.now().getYear())`

## Changes in `Metadata` API
Until now, `Metadata` supported only `String` values. This PR expands the list of supported value types to `Integer`, `Long`, `Float` and `Double`. In the future, more types may be added (if needed).
The method `String get(String key)` will be deprecated later in favor of:
- `String getString(String key)`
- `Integer getInteger(String key)`
- `Long getLong(String key)`
- etc

New overloaded `put(key, value)` methods are introduced to support more value types:
- `put(String key, int value)`
- `put(String key, long value)`
- etc

## Changes in `EmbeddingStore` API
New method `search` is added that will become the main entry point for search in the future. All `findRelevant` methods will be deprecated later.
New `search` method accepts `EmbeddingSearchRequest` and returns `EmbeddingSearchResult`.
`EmbeddingSearchRequest` contains all search criteria (e.g. `maxResults`, `minScore`), including new `Filter`.
`EmbeddingSearchResult` contains a list of `EmbeddingMatch`.
```java
EmbeddingSearchResult search(EmbeddingSearchRequest request);
```

## Changes in `EmbeddingStoreContentRetriever` API
`EmbeddingStoreContentRetriever` can now be configured with a static `filter` as well as dynamic `dynamicMaxResults`, `dynamicMinScore` and `dynamicFilter` in the builder:
```java
ContentRetriever contentRetriever = EmbeddingStoreContentRetriever.builder()
                ...
                .maxResults(3)
                // or
                .dynamicMaxResults(query -> 3) // You can define maxResults dynamically. The value could, for example, depend on the query or the user associated with the query.
                ...
                .minScore(0.3)
                // or
                .dynamicMinScore(query -> 0.3)
                ...
                .filter(metadataKey("userId").isEqualTo("123-456")) // Assuming your TextSegments contain Metadata with key "userId"
                // or
                .dynamicFilter(query -> metadataKey("userId").isEqualTo(query.metadata().chatMemoryId().toString()))
                ...
                .build();
```
So now you can define `maxResults`, `minScore` and `filter` both statically and dynamically (they can depend on the query, user, etc.). These values will be propagated to the underlying `EmbeddingStore`.

## ["Self-querying"](https://python.langchain.com/docs/modules/data_connection/retrievers/self_query/)
This PR also introduces `LanguageModelSqlFilterBuilder` in `langchain4j-embedding-store-filter-parser-sql` module which can be used with `EmbeddingStoreContentRetriever`'s `dynamicFilter` to automatically build a `Filter` object from the `Query` using language model and `SqlFilterParser`.

For example:
```java
TextSegment groundhogDay = TextSegment.from("Groundhog Day", new Metadata().put("genre", "comedy").put("year", 1993));
TextSegment forrestGump = TextSegment.from("Forrest Gump", new Metadata().put("genre", "drama").put("year", 1994));
TextSegment dieHard = TextSegment.from("Die Hard", new Metadata().put("genre", "action").put("year", 1998));

// describe metadata keys as if they were columns in the SQL table
TableDefinition tableDefinition = TableDefinition.builder()
                .name("movies")
                .addColumn("genre", "VARCHAR", "one of [comedy, drama, action]")
                .addColumn("year", "INT")
                .build();

LanguageModelSqlFilterBuilder sqlFilterBuilder = new LanguageModelSqlFilterBuilder(model, tableDefinition);

ContentRetriever contentRetriever = EmbeddingStoreContentRetriever.builder()
                .dynamicFilter(sqlFilterBuilder::build)
                .build();

String answer = assistant.answer("Recommend me a good drama from 90s"); // Forrest Gump
```

## Which embedding store integrations will support `Filter`?
In the long run, all (provided the embedding store itself supports it).
In the first iteration, I aim to add support to just a few:
- `InMemoryEmbeddingStore`
- Elasticsearch
- Milvus

## Summary by CodeRabbit

- **New Features**
  - Introduced filters for checking key's value existence in a collection for improved data handling.
- **Enhancements**
  - Updated `InMemoryEmbeddingStoreTest` to extend a different class for improved testing coverage and added a new test method.
- **Refactor**
  - Made minor formatting adjustments in the assertion block for better readability.
- **Documentation**
  - Updated class hierarchy information for clarity.


andyflury:
This looks great @langchain4j Thanks for the quick turnaround! I just tested in with my application with a simple `equals` comparison and it worked like a charm!

Regarding your questions. I cannot comment on all of them, but here are a few thoughts
> * Which data types should `Metadata` support? Some embedding stores do not differentiate between number types (e.g., `Integer`, `Long`, `Float`, `Double`). Some store metadata as JSON, so type information is lost when persisting `Metadata` and then retrieving it back. Currently, I see two options: 1. Store an additional "type" field for each `Metadata` key-value pair and be strict during comparisons (e.g., fail if `Integer` is compared with `Long`). 2. Implement auto-casting (similar to [Java's](https://docs.oracle.com/javase/specs/jls/se7/html/jls-5.html#jls-5.1.2)) during comparisons. So, if the metadata value was originally a `Long`, but after persistence, it is retrieved as an `Integer`, ensure it automatically casts to a `Long` filter value and also automatically casts when the user invokes `getLong(key)`.

I would say types like:
* Strings
* Wrapped Primitives, (Integer, Long, Double, Boolean, etc.)
* Dates (LocalDate, LocalTime, etc). 
* Enums
* BigDecimal & BigInteger

I would strict comparisons. 

> * Should we use `and(x, y)` or `x.and(y)`syntax for combining operations? Or both? The first is more explicit and clear in terms of precedence, while the second is more convenient for auto-completion.

Personally I would prefer `x.and(y)` but both are fine

> * JOOQ/QueryDSL-like (`eq`, `lt`, `gte`, etc) or AssertJ-like (`isEqualTo`, `isLessThan`, `isGreaterThanOrEqualTo`) naming?

Personally I would prefer `eq`, `lt`, `gte`, etc.

> * Should `MetadataFilter` extend `Predicate<Metadata>`?

that sounds reasonable to me!

langchain4j:
@andyflury thanks for the feedback!

Another random thought: instead of having DSL just for metadata filtering, we could have a DSL for the whole search/retrieval, something like this:

```java
SearchRequest.forQueryEmbedding(queryEmbedding)
                .where(metadata -> metadata.field("author").eq("Stephen King").and(metadata.field("year").gt(2020)))
                .minScore(0.77)
                .maxResults(5)
                .build();
```

andyflury:
@langchain4j great idea, that would make things even more convenient. I'm all for it!

geoand:
This feature is awesome!

@jmartisk @karesti  you will probably interested in this  as well

andyflury:
@langchain4j would it potentially make sense to give the `EmbeddingStore` an optional filter constructor argument? That would be helpful for the new Advanced RAG when using a `QueryRouter` with `EmbeddingStores` and all of these stores are based on the same vector database index. 

Let's say we have two types of information in the same vector database index:

- information about cars (sourceType = 'CAR')
- information about bikes (sourceType = 'BIKE')

Then you could construct one `EmbeddingStore` with filter sourceType = 'CAR' and one `EmbeddingStore` with filter sourceType = 'BIKE' plus then have the `QueryRouter` choose the appropriate `EmbeddingStore` for each query.

langchain4j:
@andyflury that totally makes sense! However, it would probably be easier to define it on the `EmbeddingStoreContentRetriever` level, otherwise we will need to add this to constructor of every EmbeddingStore implementation (15+).

Also I plan to add a self-query retriever similar to the one in python version, where LLM will be able to construct metadata filter dynamically, potentially including multiple fields. This will help to narrow down the search.

Another thing that will probably be useful is to automatically construct filter by user (If @MemoryId) is present.
Then same store/index can be used for multiple users and their data will be separated.

andyflury:
@langchain4j Yes, you are right doing it inside `EmbeddingStoreContentRetriever` is probably better!

A [self-query retriever](https://python.langchain.com/docs/modules/data_connection/retrievers/self_query)  sound extremely interesting. 

This could potentially solve another challenge we are facing currently. We have started indexing all out customer support tickets. This works well when asking questions like "what is the best solution to solve problem xyz". But it does not work well for questions like "please summarize all tickets from the last month for customer xyz". I assume that with the self-query retriever, the system would automatically be able to set filters for `customerName` and `ticketDate`. 

so far I was using Tools for this, but having a self-query retriever (in combination with the `QueryRouter` would definitely be the better approach).

Would you also consider having something like an `SQLEmbeddingStore` combined with a self-query retriever, which would automatically set appropriate `WHERE` clauses?

langchain4j:
@andyflury 

> Would you also consider having something like an SQLEmbeddingStore combined with a self-query retriever, which would automatically set appropriate WHERE clauses?

Do you mean a `ContentRetriever` backed by an SQL DB?

andyflury:
> Do you mean a `ContentRetriever` backed by an SQL DB?

Yes, I think that could be useful. 

I'm still trying to figure out where to use RAG (i.e. `ContentRetriever`) and where to use Tools. After our discussion I'm now tempted to use `ContentRetrievers? for any sort of data retrieval (vector store, SQL databases, etc.) and use Tools only for cases where some sort of action is performed by an external system (e.g. create an Email, raise a ticket, etc.). 

Would you agree to this?

langchain4j:
@andyflury yes, exactly. To summarise:
- RAG is useful when you want to _automatically_ provide LLM the external _content_ which it does not have access to (by nature).
- Tools are useful to provide LLM access to external _actions_ which it does not have access to (by nature). LLM is the one who decides to call the tool or not.

However, `ContentRetriever` can be wrapped into a `@Tool` and be called by LLM _on demand_.

langchain4j:
@andyflury naming question: am I the only one who thinks that "self-querying retriever" is a confusing naming?

andyflury:
@langchain4j  what about `LLMContentRetriever` or `LanguageModelContentRetriever`

langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
```diff
@@ -97,17 +96,27 @@ private List<String> add(List<Entry<Embedded>> newEntries) {
...
 
         Comparator<EmbeddingMatch<Embedded>> comparator = comparingDouble(EmbeddingMatch::score);
         PriorityQueue<EmbeddingMatch<Embedded>> matches = new PriorityQueue<>(comparator);
 
+        Filter filter = embeddingSearchRequest.metadataFilter();
+
         for (Entry<Embedded> entry : entries) {
-            double cosineSimilarity = CosineSimilarity.between(entry.embedding, referenceEmbedding);
+
+            if (filter != null && entry.embedded instanceof TextSegment) {
```
glaforge, line 108: We're filtering only on `TextSegment`?
Maybe we should have a kind of interface for all kinds of embedded stuff that classes like `TextSegment`, `Image`, etc. could implement?

Say something like:
```
interface HasMetadata {
    Metadata metadata();
}
```

We would retrofit `TextSegment` and `Image` and similar embedded classes to implement that interface?

That way, it'd make the filtering capability more widely applicable to more multimodality use cases?

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/main/java/dev/langchain4j/store/embedding/filter/builder/sql/ColumnDefinition.java
```diff
@@ -0,0 +1,39 @@
...
+
+import static dev.langchain4j.internal.ValidationUtils.ensureNotBlank;
+
+@Experimental
+@ToString
+@EqualsAndHashCode
+public class ColumnDefinition {
+
+    private final String name;
+    private final String type;
```
glaforge, line 15: For the type, you use a string so that it's kinda future-proof or database-agnostic?
(ie. if some data store has a proprietary and/or non-standard-SQL type)

At first, I was thinking of using a class like `java.sql.Types`. Not necessarily depending on that class, but copying all the known existing SQL types. But I guess it's not future-proof.

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/main/java/dev/langchain4j/store/embedding/filter/builder/sql/LanguageModelSqlFilterBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+ * <br>
+ * It is useful for improving retrieval from an {@link EmbeddingStore} by narrowing down the search space.
+ * <br>
+ * For instance, if you have internal company documentation for multiple products in the same {@link EmbeddingStore}
+ * and want to search the documentation of a specific product without forcing the user to specify the
+ * {@link Filter} manually, you could use {@code LanguageModelSqlFilterBuilder} to automatically create the filter
+ * using a language model.
+ * <br>
+ * <br>
+ * First, describe the {@link Metadata} of your {@link TextSegment}
```
glaforge, line 41: > the {@link Metadata} of your {@link TextSegment}

As per my other remark maybe expanding to _something that has metadata_ (ie. something that's embeddable and that would implement a marker interface that implements a `Metadata metadata()` method.

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/main/java/dev/langchain4j/store/embedding/filter/builder/sql/LanguageModelSqlFilterBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+ * This means that only those {@link TextSegment}s with a {@link Metadata} entry {@code product = "iPhone"}
+ * will be considered for the search.
+ * <br>
+ * <br>
+ * It is recommended to use a capable language model, such as gpt-3.5-turbo,
+ * or the smaller one but fine-tuned for the text-to-SQL task, such as <a href="https://huggingface.co/defog">SQLCoder</a>.
+ * SQLCoder is also available via <a href="https://ollama.com/library/sqlcoder">Ollama</a>.
+ * <br>
+ * The default {@link PromptTemplate} in this class is suited for SQLCoder, but should work fine with
+ * capable language models like gpt-3.5-turbo and better.
```
glaforge, line 76: I'm curious what other models you tried?
Did you try PaLM or Gemini?

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/main/java/dev/langchain4j/store/embedding/filter/builder/sql/LanguageModelSqlFilterBuilder.java
```diff
@@ -0,0 +1,232 @@
...
+ * <br>
+ * You can override the default {@link PromptTemplate} using builder.
```
glaforge, line 78: Handy!

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/main/java/dev/langchain4j/store/embedding/filter/parser/sql/SqlFilterParser.java
```diff
@@ -0,0 +1,354 @@
...
+        } else if (expression instanceof InExpression) {
+            return map(((InExpression) expression));
+        } else if (expression instanceof Between) {
+            return map(((Between) expression));
+        } else {
+            throw illegalArgument("Unsupported expression: '%s'%s", expression, createGithubIssueLink(expression));
+        }
+    }
+
+    private static String createGithubIssueLink(Expression unsupportedExpression) {
```
glaforge, line 138: Smart 👍

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/test/java/dev/langchain4j/store/embedding/filter/parser/sql/SqlFilterParserTest.java
```diff
@@ -0,0 +1,1693 @@
...
+                        "age IN (18, 42)",
+                        key("age").in(18L, 42L)
+                ))
+                .add(of(
+                        "weight IN (67.8, 78.9)",
+                        key("weight").in(67.8d, 78.9d)
+                ))
+
+
+                // nin
```
glaforge, line 151: At first I wondered, _"what is this `nin` operator"_ but immediately understood it was `not in`.
I guess you want to keep it short, but `notIn` might be more readable?

glaforge (commented):
A few remarks or questions, but it's a big and important PR, neat! 👍

langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
langchain4j: Good point, I was thinking about supporting custom POJOs in addition to `TextSegment` in the `EmbeddingStore`, hence the generic. But I would avoid forcing user to use specific interfaces. Right now, filtering is supported by `Metadata` entries of `TextSegment`, but with custom POJOs there would probably be no need in dynamic `Metadata`, so any POJO field can be filtered by.

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/main/java/dev/langchain4j/store/embedding/filter/builder/sql/ColumnDefinition.java
langchain4j: Hmm. I did not think about using enum for this, but yes, since this class is used only to provide LLM with the table structure, we do not have to have strict types and user should have a full freedom here.

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/test/java/dev/langchain4j/store/embedding/filter/parser/sql/SqlFilterParserTest.java
langchain4j: Good point, I guess I better use more descriptive names!

langchain4j: New naming:

```
isEqualTo
isNotEqualTo
isGreaterThan
isGreaterThanOrEqualTo
isLessThan
isLessThanOrEqualTo
isBetween
isIn
isNotIn
```

Took it from AssertJ. WDYT?

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/EmbeddingStoreContentRetriever.java
```diff
@@ -6,68 +6,149 @@
...
+ * - {@code dynamicMinScore}: It is a {@link Function} that accepts a {@link Query} and returns a {@code minScore} value.
+ * It can be used to dynamically define {@code minScore} value, depending on factors such as the query,
+ * the user (using Metadata#chatMemoryId()} from {@link Query#metadata()}), etc.
+ * <br>
+ * - {@code filter}: The {@link Filter} that will be applied to a {@link dev.langchain4j.data.document.Metadata} in the
+ * {@link Content#textSegment()}.
+ * <br>
+ * - {@code dynamicFilter}: It is a {@link Function} that accepts a {@link Query} and returns a {@code minScore} value.
+ * It can be used to dynamically define {@code filter} value, depending on factors such as the query,
+ * the user (using Metadata#chatMemoryId()} from {@link Query#metadata()}), etc.
```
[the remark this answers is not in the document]
langchain4j: Nice catch!

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/main/java/dev/langchain4j/store/embedding/filter/builder/sql/LanguageModelSqlFilterBuilder.java
langchain4j: Replied in DM

langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
glaforge: Oh interesting, even POJOs.
I guess right now we can keep this `instanceof TextSegment` condition, as that's the sole thing that has metadata. But we'll have to revisit this line when we extend the support to other types.

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/main/java/dev/langchain4j/store/embedding/filter/builder/sql/ColumnDefinition.java
glaforge: Yeah, let's keep it that way, as it's not restricting possible types that could be used. That's fine.

embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql/src/test/java/dev/langchain4j/store/embedding/filter/parser/sql/SqlFilterParserTest.java
glaforge: It's more verbose of course, but I think this is more readable 👍

ansonma:
@langchain4j , Thanks a lot for this new feature, it is exactly what we need. May I know if I can use the metadata filter function for AWS OpenSearch embedding store? If not now, may I know when it will be available? Thanks.

langchain4j:
@ansonma I do not plan to add this feature to all embedding store integrations as I do not have enough time. I hope that someone from community could contribute this, considering that foundation is there, examples (elasticsearch, milvus) are there, it should be pretty easy.

YahiaDev:
Hello,
 thank you for the new feature. Can i use the  metadata filter with PgVectorEmbeddingStore ? 
 If not it will be available for futur release ? Did you have a work around to filter metadata with  PgVectorEmbeddingStore ? Thanks

langchain4j:
Hi, `Filter` support for `PgVectorEmbeddingStore` will be released next week.

ansonma:
Hello, my company is investigating Langchain4J as our long term AI framework for our Java stack apps, can someone help to add the OpenSearch metadata filtering function? This is the only piece missing for our requirements for now. Thanks in advance.

langchain4j:
@andyflury would you like to contribute this? It should be pretty straightforward. There is [an implementation for Elasticsearch](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchMetadataFilterMapper.java) already, I assume the one for OpenSearch will be similar/same? Thanks a lot in advance!

PR #613, opened 2024-02-07
[streaming] add a .shutdown() method to the StreamingChatLanguageModel interface

Shastick:
Adds a `.shutdown()` method to the `StreamingChatLanguageModel` interface: this allows callers to request implementations to release or close any resources that they acquired.

The need for this is best illustrated with the current [OpenAI streaming example](https://github.com/langchain4j/langchain4j-examples/blob/337186583f4dc5e4e122b0cdf0a42ddb586c7fe0/tutorials/src/main/java/_04_Streaming.java#L5): 

When the example is run, it will output `Done streaming` but the program won't terminate before another minute or so.

I assume that this is due to the underlying `OkHttpClient` that has an Executor that keeps running.

### Approach

I chose to _not_ add a default `shutdown()` implementation to the interface to force any implementation to ask "are there resources we should free when this needs to shut down).

For a few implementation it is unclear if resources would need to be closes, and/or some changes in external dependencies might be required.

### Testing

Tested against the [StreamingExample](https://github.com/langchain4j/langchain4j-examples/blob/b64efcb6aafdc3b763295c740ae37546a0b475aa/other-examples/src/main/java/StreamingExamples.java#L16) from the example repository: adding `.shutdown()` at the end of the main method causes the program to properly exit as soon as the method is complete.

### DIFF
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
@@ -596,4 +596,9 @@ public class AzureOpenAiStreamingChatModel implements StreamingChatLanguageModel
             }
         }
     }
+
+    @Override
+    public void shutdown() {
+        // no-op (this implementation's client has no shutdown method).
+    }
 }
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/chat/StreamingChatLanguageModel.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/chat/StreamingChatLanguageModel.java
@@ -60,4 +60,10 @@ public interface StreamingChatLanguageModel {
     default void generate(List<ChatMessage> messages, ToolSpecification toolSpecification, StreamingResponseHandler<AiMessage> handler) {
         throw new IllegalArgumentException("Tools are currently not supported by this model");
     }
+
+    /**
+     * Releases resources that were acquired or created by the implementation,
+     * such as request executors or connections to remote services.
+     */
+    void shutdown();
 }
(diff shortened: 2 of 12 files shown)


langchain4j:
@Shastick thanks! As far as I understand, this solution would require user to explicitly call `shutdown()`, which is definitely better than now, but I am wondering if we could make it even simpler. What about using daemon threads?

Shastick:
> What about using daemon threads?

That would be ideal, and doable easily for all situations where we are responsible for every component: however most if not all instantiations of OkHttp happen outside the scope of this codebase. In the case of the `OpenAiStreamingChatModel`, it is `dev.ai4j`'s that would need to be updated.

I guess we could suggest fixes to these dependencies, but I'd still argue we want a `.shutdown()` method (or any other mechanism for releasing resources, really): 

This allows library users to manage the lifecycle of their objects: a daemon thread would solve the problem of the program not terminating upon completion, but won't resolve having dangling, unreleased resources if the models are used in an application or service with a longer life.

As an alternative:

We could also consider using an `AutoCloseable` interface to make it more obvious and allow using the model in a `try-with-resource` block? `com.google.cloud.vertexai.VertexAI` (used in our `VertexAiGeminiStreamingChatModel` via the [`GenerativeModel` class](https://github.com/langchain4j/langchain4j/blob/8a40c0c5d3a623799fcb1c37f69e7f33843d5d71/langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModel.java#L28) does this. (Although in the end, it boils down to providing a `close()` method)

langchain4j:
@Shastick good point, unreleased resouces is a real problem here. Although I am not sure why the user should care about that. I mean, about closing it manually. When the streaming is completed, resource should be released automatically. As far as I understand, the issue is in okhttp library. I have found a couple of issues, here is one of them: https://github.com/square/okhttp/issues/4029
I am planning to migrate from okhttp/ retrofit to java 11 HTTP client soon, so this problem will probably be resolved.

BTW dev.ai4j is our client, so we can change it as required.

Shastick:
> @Shastick good point, unreleased resouces is a real problem here. Although I am not sure why the user should care about that. I mean, about closing it manually. When the streaming is completed, resource should be released automatically. As far as I understand, the issue is in okhttp library. I have found a couple of issues, here is one of them: [square/okhttp#4029](https://github.com/square/okhttp/issues/4029) I am planning to migrate from okhttp/ retrofit to java 11 HTTP client soon, so this problem will probably be resolved.

I agree that in its current form it would be pretty clunky, and that we should/could hide that away wherever possible, eg by closing resources when a request is done.

If this can be solved by migrating the client I guess it's not worth introducing a change to the interface – I'll close this PR.

(Side question: will support for java 8 be dropped if we start relying on the Java 11 client? Or is there a workaround?)

langchain4j:
@Shastick Yes I think we have to move forward from java 8

PR #616, opened 2024-02-08
Implement missing parameters in Azure OpenAI

jdubois:
Following #605 implement the missing parameters in the rest of the Azure OpenAI APIs, and clean up the existing one (so parameters are in the same order everywhere).


langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingLanguageModel.java
```diff
@@ -59,22 +59,36 @@ public class AzureOpenAiStreamingLanguageModel implements StreamingLanguageModel
...
-    private final Integer maxTokens;
+    private final Map<String, Integer> logitBias;
+    private final String user;
+    private final Integer n;
+    private final Integer logprobs;
+    private final Boolean echo;
+    private final List<String> stop;
     private final Double presencePenalty;
     private final Double frequencyPenalty;
+    private final Integer bestOf;
```
langchain4j: According to [docu](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference), seems that `bestOf` cannot be used with streaming

jdubois: Indeed! This is fixed

ghost (approved):
@jdubois awesome, thank you!

PR #617, opened 2024-02-08
use maven.compiler.release instead of source+target

hboutemy:
with such a setting, you can safely build only once the whole project with JDK 17 or even 21 without fearing any wrong API being injected in .class files


langchain4j-parent/pom.xml
```diff
@@ -479,6 +479,15 @@
     </distributionManagement>
 
     <profiles>
+        <profile>
+            <id>java11+</id>
+            <activation>
+                <jdk>[11,)</jdk>
+            </activation>
+            <properties>
+                <maven.compiler.release>8</maven.compiler.release>
```
langchain4j, line 488: I am not sure I understand the purpose of this, is it supposed to be 8 or 11?

ghost (commented):
@hboutemy thanks a lot!

langchain4j-parent/pom.xml
hboutemy: activation looks at the effective JDK you are building with
compiler.release defines the target release you are compiling to

then this profile reads: if you are compiling with a modern JDK (>11, that supports JEP-247 `javac --release`), use this release flag (that by default targets Java 8: this PR does not change anything at that level)

langchain4j: @hboutemy I see, thanks for explanation!

ghost (approved):
@hboutemy thanks a lot!

PR #622, opened 2024-02-10
refactor: AssertJ best practices

timtebeek:
Hi! Noticed _almost all_ tests used AssertJ, but in some cases JUnit was still used. In addition to that some tests don't use the most expressive assertions. Figured clean that up such that you get better assertions if any tests were to fail. Compare for instance
```diff
-        assertThat(document.metadata().asMap().size()).isEqualTo(4);
+        assertThat(document.metadata().asMap()).hasSize(4);
```
The first one will print expected 5 to be equal to 4, whereas the second one shows the contents of the map involved.

Being consistent with your test library also stops bad patterns from repeating accidentally through copy-and-paste. If you want to enforce these best practices through an automated pull request check that's also an option. Let me know if you'd want that as well. Hope that helps!


ghost (dismissed):
@timtebeek thanks a lot!
Yes, I would love to have consistency checks on pull requests!

timtebeek:
> @timtebeek thanks a lot! Yes, I would love to have consistency checks on pull requests!

Great to hear; I'll open up a separate PR for that, and fix the silly import issue this has developed in a method that was missed. Sorry about that!

ghost (approved):
@timtebeek thanks a lot again!

timtebeek:
You're welcome! And if you want these changes applied on PRs going forward the follow up PR landed in
- https://github.com/langchain4j/langchain4j/pull/673

PR #625, opened 2024-02-11
recursion error on nested POJO extractor

tenpigs267:
Following #575, here is a PR which add a set of visited classes sot that each class structure is defined only once (only when it is the root element class that is re-used, it is then defined a second time (see outputFormatInstructions_PersonWithParents unit test for a comprehensible example)
Approch I suggested on #575 seemed less natural (navigate through classes first and then generate classes structure)


langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
```diff
@@ -248,7 +248,7 @@ void should_extract_custom_POJO() {
                         "\"firstName\": (type: string),\n" +
                         "\"lastName\": (type: string),\n" +
                         "\"birthDate\": (type: date string (2023-12-31)),\n" +
-                        "\"address\": (type: {\n" +
+                        "\"address\": (type: dev.langchain4j.service.AiServicesIT$Address: {\n" +
```
langchain4j, line 247: would be nicer to have just `Address` instead of `dev.langchain4j.service.AiServicesIT$Address`

langchain4j/src/test/java/dev/langchain4j/service/ServiceOutputParserTest.java
```diff
@@ -203,4 +203,27 @@ void outputFormatInstructions_PersonWithFinalFields() {
                         "\"birthDate\": (type: date string (2023-12-31)),\n" +
                         "}");
     }
+
+    static class PersonWithParents {
+        private String firstName;
+        private String lastName;
+        private List<PersonWithParents> parents;
```
langchain4j, line 210: Could you please add separate tests for `Parent[]` and `Parent mother; Parent father` cases?
Thanks!

ghost (commented):
@tenpigs267 thanks a lot!

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
tenpigs267: I had three thought on that, the first I think is that it's the classname format that was already in use, the second is that package name may contain domain term that can be useful for llm understanding (or not), the third is the most important is for classname uniqueness.
So I stayed with the format already in use but if you think it's better, let's go with classname without packages names.

langchain4j: Good point, thanks for explanation!

langchain4j/src/test/java/dev/langchain4j/service/ServiceOutputParserTest.java
```diff
@@ -203,4 +203,27 @@ void outputFormatInstructions_PersonWithFinalFields() {
...
+
+    @Test
+    void outputFormatInstructions_PersonWithParents() {
+        String formatInstructions = ServiceOutputParser.outputFormatInstructions(PersonWithParents.class);
+
+        assertThat(formatInstructions).isEqualTo(
+                "\nYou must answer strictly in the following JSON format: {\n" +
+                        "\"firstName\": (type: string),\n" +
+                        "\"lastName\": (type: string),\n" +
+                        "\"parents\": (type: array of dev.langchain4j.service.ServiceOutputParserTest$PersonWithParents: {\n" +
```
langchain4j: Instead of printing PersonWithParents fields twice, should it just put "dev.langchain4j.service.ServiceOutputParserTest$PersonWithParents" before the first description and then just reference itself? This will potentially save a lot of tokens.

Something like this (pseudo-JSON):
```
PersonWithParents:
 - firstName
 - lastName
 - parents: array of PersonWithParents
```

langchain4j/src/test/java/dev/langchain4j/service/ServiceOutputParserTest.java
tenpigs267: outputFormatInstructions_PersonWithMotherAndFather have been added

langchain4j/src/test/java/dev/langchain4j/service/ServiceOutputParserTest.java
tenpigs267: I think it's ok.

langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
```diff
@@ -115,34 +115,36 @@ public static String outputFormatInstructions(Class<?> returnType) {
             return "\nYou must put every item on a separate line.";
         }
 
-        return "\nYou must answer strictly in the following JSON format: " + jsonStructure(returnType);
+        return "\nYou must answer strictly in the following JSON format: " + jsonStructure(returnType, new HashSet<>());
     }
 
-    private static String jsonStructure(Class<?> structured) {
+    private static String jsonStructure(Class<?> structured, Set<String> visited) {
```
langchain4j: minor: would be probably safer and more clear to use `Set<Class<?>> visited` instead of `Set<String> visited`

langchain4j/src/test/java/dev/langchain4j/service/ServiceOutputParserTest.java
langchain4j: Thanks! Could you please also add one test for `Parent[]` case since it has a different handling logic?

ghost (commented):
@tenpigs267 thanks a lot! Just 2 minor comments left and we can merge!

tenpigs267:
It should be ok @langchain4j

ghost (approved):
@tenpigs267 great job, thank you a lot!

langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
```diff
@@ -115,34 +115,36 @@ public static String outputFormatInstructions(Class<?> returnType) {
...
+        return "\nYou must answer strictly in the following JSON format: " + jsonStructure(returnType, new HashSet<>());
     }
 
-    private static String jsonStructure(Class<?> structured) {
+    private static String jsonStructure(Class<?> structured, Set<Class<?>> visited) {
         StringBuilder jsonSchema = new StringBuilder();
-        jsonSchema.append("{\n");
+        String simpleTypeName = simpleTypeName(structured);
+        visited.add(structured);
+        jsonSchema.append(simpleTypeName + ": {\n");
```
geoand, line 125: I'm curious: What's the reasoning for the class name before json structure?

tenpigs267: Hi, you can see it in ServiceOutputParserTest.outputFormatInstructions_PersonWithParentArray the PersonWithParentArray description is not repeated if used multiple time in your data structures.

geoand: Understood, but my worry is that this it produces something that is not valid json, for example:

```
You must answer strictly in the following JSON format: org.acme.ClaimSummary: {\n\"title\": (type: string),\n\"summary\": (type: string),\n\"date\": (type: string),\n\"location\": (type: string),\n\"policyNumber\": (type: string),\n}
```

tenpigs267: the format 
"propertyLabel": (type: classname), ...
seems to be well understand by LLMs, a maybe more reliable way could be to use json-schema but it is a lot more verbose, we should consider usage with token based API pricing. As far as I can say it seems to be a good compromise

geoand: Do you have any links explaining that using the name works? Asking because I've never seen this used before and it strikes me as odd that it works (and I did see it does, although my colleague @cescoffier ran into issues with it)

cescoffier: It didn't work with GPT 3.5 and Mistral for me. I had to switch back to String, sanitize and do the mapping myself.

tenpigs267: oh maybe you are just talking about the first org.acme.ClaimSummary?
_You must answer strictly in the following JSON format: **org.acme.ClaimSummary:** {\n\"title\": (...._
If this is what you are talking about, this was a suggestion from @langchain4j also to save token. It seemed to work well so I had no further objection

geoand: > oh maybe you are just talking about the first org.acme.ClaimSummary?

Yes exactly. This wasn't part of the prompt in previous versions and now its addition is breaking thing unfortunately

tenpigs267: mmmm no I have no reference to provide, if it break things we should revert this part before this commit : https://github.com/langchain4j/langchain4j/pull/625/commits/bbf8f67722bfeb33fe699875eaf11f4811affc85
Just waiting @langchain4j if he has some reference to provide and I will make a new PR

geoand: @langchain4j I really think the class name should be removed from the JSON as @cescoffier is having issues with it with various LLMs

langchain4j: Well I am not a big fan of having a class name in front of schema and I think I raised this concern before, but somehow it worked in ITs, so we went on. Maybe it makes sense to add it only when there is a cycle (self-reference).

But in any case, do you guys enable [json mode](https://docs.langchain4j.dev/tutorials/ai-services#json-mode) when doing POJO extraction?

cescoffier: @geoand it would make sense to enable the json mode if the result is not a String?

cescoffier: arf.... actually the json mode can be tricky. As it's not because you use the OpenAI client that you are actually using OpenAI..., I'm not sure how to configure it properly.

tenpigs267: the fix will be in #844

geoand: Hm... So Json mode is per model? That could be a problem because you could potentially have multiple AiServices using the same model

geoand: > the fix will be in #844

Thanks!

langchain4j: When return type is POJO, we could politely ask `ChatLanguageModel` to use json mode, if it supports it.
To avoid forcing user to set json mode on the model explicitly.

langchain4j: That would require overloading `generate()` methods to accept same parameters that are configured in the constructor of `ChatLanguageModel`. Long asked feature...

langchain4j: @geoand availability of "JSON mode" feature depends more on the inference runtime than on the model itself. For example, Ollama supports it for all models.

geoand: Sure, my point is that enabling this makes sense per AiService method

langchain4j: Sorry, don't get it. How to you want to "enable" it? In my understanding if return type is POJO and LLM provider supports JSON mode, library should automatically enable it, there is no harm in it.

geoand: What happens if you enable it but you don't need a json response, you only need simple string response for example?

langchain4j: I mean JSON mode will be enabled only for a single request to the LLM and only if the return type is POJO

geoand: Right, we are saying the same thing I think: that the library can / should enable json mode when it knows that json needs to be returned.

PR #626, opened 2024-02-12
Make the Vertex AI embedding model adapt automatically the number of batch calls to make

glaforge:
With this change, the batching of embedding calls will be done automatically, adjusting the number of text segments to calculate embeddings for, according to the max number of tokens per batch call (up to 20k) and the max number text segments per batch call (up to 250).
There are 2 new methods in the builder to specify the max number of tokens and max number of text segments per call, as sometimes, the calculation of tokens seems to not be correctly returned by the underlying model (with textembedding-gecko@001 it's fine, but textembedding-gecko@003 seems to calculate 5-10% less tokens on a per segment basis, than it returns for the overall call)


langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
```diff
@@ -113,6 +144,83 @@ public Response<List<Embedding>> embedAll(List<TextSegment> segments) {
...
     }
 
+    /**
+     * Calculates the number of tokens for each segment in the input list.
+     *
+     * @param segments a list of TextSegments
+     *
+     * @return a list of tokens counts for each segment
+     */
+    public List<Integer> getTokensCounts(List<TextSegment> segments) {
```
langchain4j: minor: is this method `public` on purpose?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
```diff
@@ -82,9 +111,11 @@ public Response<List<Embedding>> embedAll(List<TextSegment> segments) {
             List<Embedding> embeddings = new ArrayList<>();
             int inputTokenCount = 0;
 
-            for (int i = 0; i < segments.size(); i += BATCH_SIZE) {
+            List<Integer> tokensCounts = getTokensCounts(segments);
```
langchain4j: This step seems to add around 30% to the overall latency :( Would be nice if google could publish their tokenizer algo so that we could count tokens without the need to do network calls.

For now, I am wondering if we could approximate the calculation of tokens somehow? For the general use case. User could still configure it to use the proper estimation, but we could use a faster estimate instead of a slow but precise calculation by default.
2 options that come to mind:
- count number of words (by roughly splitting by space). it seems that 1 english word is rougly 1.2-1.3 tokens, so we can play safe and use 1.
-  count number of chars. it seems that 1000 chars is rougly 200-240 tokens, so we could play safe and use 180 or so.
WDYT?

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModelIT.java
```diff
@@ -52,4 +53,111 @@ void testEmbeddingModel() {
...
+            .location(System.getenv("GCP_LOCATION"))
+            .publisher("google")
+            .modelName("textembedding-gecko@001")
+            .maxRetries(3)
+            .build();
+
+        List<Integer> tokenCounts = model.getTokensCounts(createRandomSegments(5000, 1000));
+
+        assertThat(tokenCounts).hasSize(5000);
+        assertThat(tokenCounts.stream().mapToInt(i -> i).sum()).isGreaterThanOrEqualTo(1_000_000);
```
langchain4j, line 73: minor: please comment where `1_000_000` comes from

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
```diff
@@ -29,7 +34,15 @@
  * Represents a Google Vertex AI embedding model, such as textembedding-gecko.
  * See details <a href="https://cloud.google.com/vertex-ai/docs/generative-ai/embeddings/get-text-embeddings">here</a>.
  * <br>
- * Please follow these steps before using this model:
+ * <br>
+ * This embedding model transparently handles call batching, however the underlying API has imposes
+ * a maximum of 250 embeddings per call, with a max of 20,000 tokens per call.
+ * You can tweak those two parameters with the <code>maxBatchSize()</code> and
```
langchain4j: Maybe `maxBatchSize` -> `maxSegmentsPerBatch`?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
```diff
@@ -46,32 +59,48 @@
...
-                ensureNotBlank(project, "project"),
-                ensureNotBlank(location, "location"),
-                ensureNotBlank(publisher, "publisher"),
-                ensureNotBlank(modelName, "modelName")
-        );
+
         this.maxRetries = getOrDefault(maxRetries, 3);
+
+        this.maxBatchSize = getOrDefault(maxBatchSize, EMBEDDING_MAX_INPUTS_PER_REQUEST);
+        this.maxTokensPerBatch = getOrDefault(maxTokensPerBatch, MAX_TOTAL_TOKENS_COUNT_PER_EMBEDDING_CALL);
```
langchain4j: minor: please use DEFAULT_{PARAM_NAME} naming for default constants?
e.g. `DEFAULT_MAX_TOKENS_PER_BATCH`

ghost (commented):
@glaforge thanks a lot!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: Yes, it's on purpose, as it's a handy method for people who want to control the size of their prompts.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: Before sending this PR, I actually went through different prototypes.

Since I couldn't find a built-in function that would calculate token counts (without calling the server), I went with the approach of estimating the token counts, but even with a conservative approach, on some dummy text, I hit the limit. So I started to loop till I hit a token limit exception, with some splitting/dichotomy approach: let's try to see if this can handle everything, ah no, then split in two, and do two calls, ah one call fails, let's split again.
But in the end, all these tricks seemed a bit over-engineered, and a bit ugly to my taste.

So instead, I went the route of calling the API that counts token. It's not ideal, and adds overhead, but the code ended being a bit more elegant, and hopefully easier to maintain in the long run. Also, indeed making those calls add 30%, but although I haven't measured the overhead with my other approach (dichotomy + error handling), I had some overhead as well as I asked, say, to embed 250 segments, boom it fails, let's try again with 125 and 125, boom one fail, let's try 64 and 64, etc.

A strange oddity though is that counting and summing up individual text segments tokens doesn't necessarily gives the same number as when I send the all the segments together, which is a bit unintuitive, and I haven't found why, but kept the choice to specify limits (max number of segments, max number of tokens)

Now, that said, I have some hope to convince the people working on the models to share their embedding algorithm somehow, and if they ever do, we'll be able to replace the network API call very easily with the library code that would calculate the number of tokens.

So I think I'll stick to my current approach in this PR, unless you strongly disagree.

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModelIT.java
glaforge: OK, I can add a comment.
I think it's just that 5000 segments of 1000 characters is about 5 million characters.
And with an estimate of 5 letters per token, we have at least 1 million tokens in total.
It's a bit arbitrary.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: I like the suggestion.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: OK.
Good convention.

I'll update the PR (maybe tomorrow) with the comments addressed. 
Thank you for the review.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
langchain4j: Ok, makes sense! thanks for explanation

langchain4j-vertex-ai/src/test/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModelIT.java
glaforge: Comment added

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: Updated

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: Done, both constants with defaults updated with the naming convention

glaforge:
PR updated with your review comments.
Thanks for the advice!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
```diff
@@ -46,32 +59,48 @@
...
-        this.endpointName = EndpointName.ofProjectLocationPublisherModelName(
-                ensureNotBlank(project, "project"),
-                ensureNotBlank(location, "location"),
-                ensureNotBlank(publisher, "publisher"),
-                ensureNotBlank(modelName, "modelName")
-        );
+
         this.maxRetries = getOrDefault(maxRetries, 3);
+
+        this.maxSegmentsPerBatch = getOrDefault(maxSegmentsPerBatch, DEFAULT_MAX_SEGMENTS_PER_BATCH);
```
langchain4j: minor: should we validate that maxSegmentsPerBatch and maxTokensPerBatch are in the valid range? At least that they are positive? Upper boundaries might be changed by Google at any time as far as  I understand?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
langchain4j: minor: maybe `calculateTokenCount`? as opposed to `estimateTokenCount` which we have in other places?

"get" sounds like a getter and It can be unobvious for the user that "get" method does a network call.

ghost (dismissed):
@glaforge thanks a lot!

langchain4j:
@glaforge just in case, this can be merged. The 2 minor thingies can be addressed afterwards.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: Good question.
I can definitely check they are positive, that's a good idea. I'll add that.
But then, yes, I hesitated to check the boundaries, but since they can change (like when we went from 5 to 250) so I think we should not check ourselves and let the underlying API complain should another change happen.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: Good catch again, it's not a getter.
And I hadn't noticed the `TokenCountEstimator` interface.
I'll implement that interface.

langchain4j: Not sure it should be `TokenCountEstimator ` as estimator is named like this to emphasize estimation, not the exact count.
At least this is the case for OpenAI as we do not know exactly how they count tokens in some cases, like with tools, thus "estimate".
Also the interface probably dos not fit your use case?

glaforge:
I messed up a bit with Git, but I've renamed the `getTokensCounts()` method, but as you said, it's not an estimate, so I went with `calculateTokensCounts()`. 
And I've added positivity checks for the 2 max params.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: Just renamed the method. You're right, that interface is not ideal for the use case.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: Added positivity check

ghost (approved):
@glaforge thank you a lot!

PR #635, opened 2024-02-13
Adds a test to check that the API version is set to the latest if incorrect

agoncal:
Adds a test to check that if the Azure API version is incorrect, then the latest is used.

If you look at `InternalAzureOpenAiHelper.getOpenAIServiceVersion()` method, you see that it looks is the `serviceVersion` is in the allowed list. If not, then the latest version is used.


ghost (approved):
@agoncal thanks!

PR #636, opened 2024-02-13
Clarify the mandatory and optional Azure connection parameter

agoncal:
Based on https://github.com/langchain4j/langchain4j/pull/635 we can see that the `serviceVersion` is not mandatory. If not set, then the latest version of the API is used. So only `apikey` and `endpoint` are really mandatory. Then, is `serviceVersion` and `deploymentName` are not set, we set default values.


SandraAhlgrimm:
Hi @agoncal, does it make sense to link to the file where the default value for the deployment name is set? I would want to display the default value but as its likely to be updated and I don’t want outdated docs, linking could be helpful. Wdyt?

agoncal:
@SandraAhlgrimm actually, the default changes depending on the model. I also thought of externalising the values in an enum or a class similar to `AzureOpenAiModelName`. WDYT ?

Screenshot 2024-02-13 at 15 36 20

langchain4j:
@agoncal thank you!

ghost (approved):
cc @jdubois

PR #641, opened 2024-02-14
[FEATURE] Google web search integration

czelabueno:
Integrating [Google Search](https://developers.google.com/custom-search) as a `WebSearchEngine` and as a `Tool` for function calling

## Summary by CodeRabbit

- **New Features**
	- Introduced a comprehensive web search functionality using the Google Custom Search Engine, including the ability to retrieve and represent web search results with details such as title, snippet, and link.
	- Added a new `WebSearchTool` interface and its implementation for running web searches and fetching results.
	- Enhanced content retrieval capabilities with the addition of `WebSearchContentRetriever` for fetching web content based on queries.

- **Refactor**
	- Improved the structure and organization of web search-related classes and interfaces for better modularity and readability.

- **Chores**
	- Implemented logging interceptors for monitoring HTTP requests and responses with the Google Custom Search API, aiding in debugging and performance tracking.

PR #538, opened 2024-01-23
Foundation for advanced RAG

ghost:
So far, LangChain4j had only a simple (a.k.a., naive) RAG implementation: a single `Retriever` was invoked on each interaction with the LLM, and all retrieved `TextSegments` were appended to the end of the `UserMessage`. This approach was very limiting.

This PR introduces support for much more advanced RAG use cases. The design and mental model are inspired by [this article](https://blog.langchain.dev/deconstructing-rag/) and [this paper](https://arxiv.org/abs/2312.10997), making it advisable to read the article.

This PR introduces a `RetrievalAugmentor` interface responsible for augmenting a `UserMessage` with relevant content before sending it to the LLM. The `RetrievalAugmentor` can be used with both `AiServices` and `ConversationalRetrievalChain`, as well as stand-alone.

A default implementation of `RetrievalAugmentor` (`DefaultRetrievalAugmentor`) is provided with the library and is suggested as a good starting point. However, users are not limited to it and can have more freedom with their own custom implementations.

`DefaultRetrievalAugmentor` decomposes the entire RAG flow into more granular steps and base components:
- `QueryTransformer`
- `QueryRouter`
- `ContentRetriever` (the old `Retriever` is now deprecated)
- `ContentAggregator`
- `ContentInjector`

This modular design aims to separate concerns and simplify development, testing, and evaluation. Most (if not all) currently known and proven RAG techniques can be represented as one or multiple base components listed above.

Here is how the decomposed RAG flow can be visualized:
advanced-rag

This mental and software model aims to simplify the thinking, reasoning, and implementation of advanced RAG flows.

Each base component listed above has a sensible and simple default implementation configured in `DefaultRetrievalAugmentor` by default but can be overridden by more sophisticated implementations (provided by the library out-of-the-box) as well as custom ones. The list of implementations is expected to grow over time as we discover new techniques and implement existing proven ones.

This PR also introduces out-of-the-box support for the following proven RAG techniques:
- Query expansion
- Query compression
- Query routing using LLM
- [Reciprocal Rank Fusion](https://learn.microsoft.com/en-us/azure/search/hybrid-search-ranking)
- Re-ranking ([Cohere Rerank](https://docs.cohere.com/docs/reranking) integration is coming in a [separate PR](https://github.com/langchain4j/langchain4j/pull/539)).

PR #642, opened 2024-02-14
[FEATURE] WebSearchEngine as ContentRetriver and tool

czelabueno:
## Web Search Engine v1

`WebSearchEngine` interface is responsible to perform searches on the Web in response to a user query.

The most popular search engines can be implemented on top of this:
- [Google Search](https://developers.google.com/custom-search/docs/overview) (Integration is coming is a [separate PR](https://github.com/langchain4j/langchain4j/pull/641))
- [Tavily](https://github.com/langchain4j/langchain4j/pull/676)
- [SearchApi](https://github.com/langchain4j/langchain4j/issues/1132)
- SerpApi Search
- Bing Search
- etc

`WebSearchRequest` and `WebSearchResults` (response) follow [opensearch](https://github.com/dewitt/opensearch) foundation standard implemented by most web search engine libs like Google, Bing, Yahoo, etc and powered by W3C.

## Web Search Engine as ContentRetriever (for RAG)

This PR also introduces out-of-the-box `WebSearchContentRetriever` that is an implementation of a `ContentRetriever` as part of the advanced RAG flow. The default behavior is currently supported on this PR (v1):

1. Default: Developer wants snippet web pages back (v1)
2. Optional: Developer wants to retrieve only the relevant segments from scrapped web pages => developer configures `DocumentSplitter` and `EmbeddingModel` and `minScore` to return results (will come in v2).

## Web Search Engine as Tool (function calling)
Any implementation of a `WebSearchEngine` can be wrapped by `WebSearchTool`, so any LLM that supports function calling can use it as a tool.

Here is how the web search can be visualized:
web-search-engine

#### Others
- In the `langchain4j-examples` project, examples will be added of how to use it with GoogleSearch as a Tool and as a content retriever in a RAG flow in the next days.
- Documentation will be updated in the next days as well.


czelabueno:
Hi @langchain4j 
Could you check if this PR fits the advanced RAG architecture that you thought? This PR adds the basis for retrieve from new data sources (e.g., web).
I wait for your feedback to work on adjust it

langchain4j:
@czelabueno thank you so much, you did a great job here! This fits the design of advanced RAG very well.

As usual, I have just a few small nitpicks 😅 

Naming: I would not use the name `Tool` to refer to the web search engine, as `Tool` already has a different meaning: in the context of LLMs, `Tool` is something that LLM can call on demand, but in this case it is not necessarily the case, as web search engine can be also used as a `ContentRetriever`.

I would represent Google/Bing/SerpAPI/etc as a `WebSearchEngine` with a simple interface:
```
public interface WebSearchEngine {
    
    WebSearchResults search(WebSearchRequest reqeust);
}
```
`WebSearchRequest` might contain not only the query string, but also some metadata like user language, user location, etc. Search engines take a lot of such parameters into account when searching and it greatly improves results.
Also, optionally number of results, etc.

`WebSearchResults` can contain a list of URLs, optionally some metadata related to each result (like score, short description, whatever), as well as some metadata related to the whole search result.

The term proposed in the [advanced RAG PR](https://github.com/langchain4j/langchain4j/pull/538) is a "content retriever", it retrieves any content (any modality) from any source, given a user query. So `WebSearchEngine` and `SqlDataBase` (or whatever) can be "wrapped" into a `WebSearchContentRetriever,` etc.

Now, `WebSearchEngine` can also be exposed as a tool to the LLM, we can wrap it into a tool the same way `GraalVmJavaScriptExecutionEngine` is wrapped into `GraalVmJavaScriptExecutionTool`.

`WebSearchContentRetriever` looks good, I would just make `EmbeddingModel` optional, in case developer does not want to do RAG over the pages (if developer wants to get complete pages). And replace `EmbeddingStore` with a `DocumentSplitter` so that developer has some control over how contents of found pages are split before doing RAG over them. We can use `InMemoryEmbeddingStore` under the hood for storing and searhcing on the fly.

So there can be 2 "setups" for `WebSearchContentRetriever`:
1. Developer wants complete pages back  => developer does not configure `DocumentSplitter` and `EmbeddingModel`
2. Developer wants to retrieve only the relevant segments from found web pages => developer configures `DocumentSplitter` and `EmbeddingModel`, we split found pages into segments, embed them using `EmbeddingModel`, put that into an `InMemoryEmbeddingStore` and retrieve `maxResults` results and finally return those.

Regarding the `ScoringModel`, I think it is not required in `WebSearchContentRetriever`.
RAG flow can be configured to retrieve from both private (`EmbeddingStoreContentRetriever`) and public (`WebSearchContentRetriever`) sources, afterwards `ReRankingContentAggregator` can be used as a final step.

Hope this makes sense.
BTW, SerpAPI integration is also in the process (@kuraleta is on it), so we should sync in this regard.
Thanks a lot!

langchain4j:
@czelabueno BTW there is also [Tavily](https://tavily.com/) which @kuraleta also integrates now as a `TavilyContentRetriever` (because it does not really fit the `WebSearchEngine`)

czelabueno:
@langchain4j alright, this makes more sense to me. 
>  in the context of LLMs, Tool is something that LLM can call on demand.

That was my doubt because I was not sure how to consider it since only OpenAI uses the functions and Tools, but I understand that probably other LLMs like OpenAI API can use it in the future, LangChain category them as Tool but if we are in another approach it will be better avoid it :)

I will work on fit it in the next days I hope can serves as base for @kuraleta as well.

czelabueno:
Hi @langchain4j this PR has been updated according last insights.

Also, don't forget check this other #641 since it's related.

Thanks!

cc: @kuraleta

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+import lombok.Builder;
+
+import java.util.List;
+
+import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.ValidationUtils.*;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * A {@link ContentRetriever} backed by a {@link WebSearchEngine}.
```
langchain4j:
```suggestion
 * A {@link ContentRetriever} that retrieves relevant {@link Content} from the web using a {@link WebSearchEngine}.
```

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+ * <br>
+ * By default, this retriever fetches the  all web pages {@link Content} for a given {@link Query}.
```
langchain4j:
```suggestion
 * By default, this retriever returns complete web pages: one {@link Content} for each web page that a {@link WebSearchEngine} has returned for a given {@link Query}. For some use cases, this might be suboptimal, as complete web pages can contain too much irrelevant content. For example, for the query "Who painted the Mona Lisa?", Google returns a Wikipedia page that contains everything about the Mona Lisa. For such use cases, this retriever can be configured with a {@link DocumentSplitter} and an {@link EmbeddingModel}, which will be used to split web pages into smaller segments and find the most relevant ones, returning only those.
```

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+    public WebSearchContentRetriever(WebSearchEngine webSearchEngine, DocumentSplitter documentSplitter) {
+        this(webSearchEngine, null, documentSplitter,null);
+    }
+
+    /**
+     * Constructs a new WebSearchContentRetriever with the specified web search engine and embedding model.
+     * @param webSearchEngine The web search engine to use for retrieving web search results.
+     * @param embeddingModel The embedding model to use for generating embeddings and finding the most 3 relevant web search results.
+     */
+    public WebSearchContentRetriever(WebSearchEngine webSearchEngine, EmbeddingModel embeddingModel) {
```
langchain4j: I am almost sure web search engines already do a very good job at properly ranking web pages according to the user query, better than any `EmbeddingModel` can do.

The idea behind using `DocumentSplitter` and `EmbeddingModel` was to use them together, not separately (as can be done using this ctor and the one above).
The idea is to, instead of returning complete pages back (that might contain a lot of irrelevant information: for example, if user asks how long to bake a chicken, web page can contain the whole recipe, so it is more optimal to find only the piece that describes the cooking time), split pages into segments using `DocumentSplitter`, embed them using an `EmbeddingModel` and then find N (configurable) most relevant segments to the given `Query`.

So I guess there should be only 2 ctors:
`WebSearchContentRetriever(WebSearchEngine)` and `WebSearchContentRetriever(WebSearchEngine, DocumentSplitter, EmbeddingModel, int maxResults)` and in both cases all ctor arguments should be mandatory.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * A {@link ContentRetriever} backed by a {@link WebSearchEngine}.
+ * <br>
+ * By default, this retriever fetches the  all web pages {@link Content} for a given {@link Query}.
+ * <br>
+ * <br>
+ * Configurable parameters (optional):
+ * <ul>
+ *     <li>{@link DocumentSplitter} - To split the web pages into smaller text segments.</li>
+ *     <li>{@link EmbeddingModel} - To generate embeddings to ingest them into an internal {@link EmbeddingStore} and find the most
```
langchain4j:
```suggestion
 *     <li>{@link EmbeddingModel} - To embed segments and find the most
```
User does not have to know that we use `EmbeddingStore` internally, technically search can be done without an `EmbeddingStore`.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+ *     relevant web search results based on the {@link Query}</li>
```
langchain4j:
```suggestion
 *     relevant ones to the given {@link Query}</li>
```

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+ *     <li>{@link #maxResults} - The maximum number of results to retrieve by the internal {@link EmbeddingStore}.
```
langchain4j:
```suggestion
 *     <li>{@link #maxTextSegments} - The maximum number of text segments to return.
```
`maxResults` might be misunderstood as the number of results web search engine is returning (which is configured elsewhere in the `WebSearchEngine`)

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+        this.embeddingStore = new InMemoryEmbeddingStore<>();
+        this.maxResults = ensureGreaterThanZero(getOrDefault(maxResults, DEFAULT_MAX_RESULTS), "maxResults");
+    }
+
+    @Override
+    public List<Content> retrieve(Query query) {
+        WebSearchResults webSearchResults = webSearchEngine.search(query.text());
+        List<TextSegment> textSegments = webSearchResults.toTextSegments();
+        if (documentSplitter != null) {
+            textSegments = documentSplitter.splitAll(webSearchResults.toDocuments());
```
langchain4j: On a second thought, it actually seems to be a very good idea to give an option to just split pages without further search/filter, so that in the next step, re-ranking (and filtering) can be done with a much better model!

I guess this can be a third constructor: `(WebSearchEngine, DocumentSplitter)`

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+ * </ul>
+ */
+public class WebSearchContentRetriever implements ContentRetriever{
+
+    private static final int DEFAULT_MAX_RESULTS = 3;
+    
+    private final WebSearchEngine webSearchEngine;
+    private final EmbeddingModel embeddingModel;
+    private final DocumentSplitter documentSplitter;
+    private final EmbeddingStore<TextSegment> embeddingStore;
```
langchain4j: We should create a new `InMemoryEmbedding` store for each `retrieve()` call, otherwise later searches will return some content from previous searches. So it should be a local field in the method, not a class field.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,139 @@
...
+    @Override
+    public List<Content> retrieve(Query query) {
+        WebSearchResults webSearchResults = webSearchEngine.search(query.text());
+        List<TextSegment> textSegments = webSearchResults.toTextSegments();
+        if (documentSplitter != null) {
+            textSegments = documentSplitter.splitAll(webSearchResults.toDocuments());
+        }
+        if (embeddingModel != null) {
+            Embedding embedding = embeddingModel.embed(query.text()).content();
+            embeddingStore.add(embedding, TextSegment.from(query.text()));
```
langchain4j: this is not necessary, as it will always return it on the later search step as a first result...

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchEngine.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public interface WebSearchEngine {
+
+    /**
+     * Runs a search query on the web search engine and returns a pretty-string representation of the search results.
+     *
+     * @param query the search query
+     * @return a pretty-string representation of the search results
+     */
+    default String runSearch(String query) {
```
langchain4j: is this method required? How/where will it be used?

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchEngine.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Performs a search query on the web search engine and returns the search results.
+     *
+     * @param query the search query
+     * @return the search results
+     * @throws IllegalArgumentException if no web search results are found for the query
+     */
+    default WebSearchResults search(String query) {
+        WebSearchResults results = search(WebSearchRequest.from(query));
+        ensureNotNull(results, "No web search results found for query: " + query);
```
langchain4j: Is it possible to get no results at all?

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
```diff
@@ -0,0 +1,251 @@
...
+    private final String content;
+    private final Map<String, String> resultMetadata;
+
+    /**
+     * Constructs a WebSearchOrganicResult object with the given content.
+     *
+     * @param content The content of the search result.
+     */
+    public WebSearchOrganicResult(String content) {
+        this("noTitle", "noLink", content, null);
```
langchain4j: Why "noTitle" and "noLink" instead of null?
Also, how come there can be no link?

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchInformationResult.java
```diff
@@ -0,0 +1,140 @@
...
+        return new WebSearchInformationResult(totalResults, pageNumber, metadata);
+    }
+
+    /**
+     * Creates a new WebSearchInformationResult with the specified total results.
+     *
+     * @param totalResults The total number of results.
+     * @return The new WebSearchInformationResult.
+     */
+    public static WebSearchInformationResult informationResult(Long totalResults) {
```
langchain4j: I guess you took this approach from one of existing domain classes?

I feel like it was a bad idea to introduce so many ways to create a domain object:
- constructors
- static `ClassName.from(...)` factories
- static `className(...)` factories

I am considering deprecating the last one and not introduce new `className(...)` factories, WDYT?

ghost (commented):
@czelabueno thank you a lot, you did a great job, especially with opensearch standard, I didn't know about it!

I did not manage to finish this review yet, but these are comments I have for now. Will continue tomorrow.

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchEngine.java
czelabueno: This method is used by objects with`@Tool` due `DefaultToolExecutor` is unable to render a custom POJO as response so far (just `String`)

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchEngine.java
czelabueno: No, maybe the value of the name is confusing but I wanted to validate that if there is no result something went wrong and an exception could be thrown to the user

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchInformationResult.java
czelabueno: I totally agree. I no longer find a real use for it (only for unit tests perhaps, but that's it). I'll remove its.

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
czelabueno: Normally, any web page found by a web search must have a source link and a title at a minimum to display it as result, from a web searcher side these are more important even than content but for our purpose it is not worth validating it. I'm going to leave them as null to avoid dirty the response.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
czelabueno: yeah exactly and with a much better `embeddingStore` to search/filter using a better technique where user be able to set a `minScore` to retrieve more accurate `textSegments` in a next step.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
czelabueno: It has been changed now

czelabueno:
@langchain4j  I'm resuming this feature because I have more ideas to implement more RAG approaches where it'll be very nice have a web searches.
This PR has been changed fixing your last comments and my own ideas.

I was reading **Corrective RAG (CRAG)** approach published in this [paper](https://arxiv.org/pdf/2401.15884.pdf) to improve the robustness of generation of final responses. 
This approach calculate the confidence and give a final judgment based on score of each `document` evaluated. If each document evaluated is **ambiguous** or **incorrect** to user query OR If any document is irrelevant, so will supplement retrieval with **web search** results. 

For the LLMs with outdated information or limited knowledge could require this approach to avoid directly the hallucinations. In this approach is very important to seek complementary external knowledge if the retrieved results are all assumed irrelevant.

This feature will lead to starting a module called `langchain4j-corrective-rag` in the future.

Take a look and let me know what you think 😊

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchEngine.java
langchain4j: minor: I would move this check to `WebSearchRetults` constructor, so that it even canot be created empty

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchEngine.java
langchain4j: I would avoid having this logic here, it is not a responsibility of `WebSearchEngine` to format results.
If this is needed by a `WebSearchTool`, let's better move it there.

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchInformationResult.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+/**
+ * Represents general information about the web search performed.
+ * This includes the total number of results, the page number, and metadata.
+ * <p>
+ * The total number of results is the total number of web pages that are founded by the search engine in response to a search query.
```
langchain4j:
```suggestion
 * The total number of results is the total number of web pages that are found by the search engine in response to a search query.
```

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchResults.java
```diff
@@ -0,0 +1,166 @@
...
+ * {@link WebSearchResults} follow opensearch foundation standard implemented by most web search engine libs like Google, Bing, Yahoo, etc.
+ * <a href="https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#examples-of-opensearch-responses">OpenSearch#response</a>
+ * </p>
+ * <p>
+ * The organic search results are the web pages that are returned by the search engine in response to a search query.
+ * These results are typically ranked by relevance to the search query.
+ */
+public class WebSearchResults {
+
+    private final Map<String, Object> searchMetadata;
```
langchain4j, line 27: What is the difference between `searchMetadata` here and `searchInformation.metadata`? Can fields from `searchInformation` be merged into `searchMetadata`? Somehow seems duplicated/unclear

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
```diff
@@ -0,0 +1,204 @@
...
+    private final String link;
+    private final String content;
+    private final Map<String, String> resultMetadata;
+
+    /**
+     * Constructs a WebSearchOrganicResult object with the given content.
+     *
+     * @param content The content of the search result.
+     */
+    public WebSearchOrganicResult(String content) {
```
langchain4j: Please remove this ctor if result cannot exist without a link

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
langchain4j: Naming nitpick: I would rather rename `String link` into `URI url`, to be more consistent and strict.

If url and title are mandatory, please reflect that in the constructors:
```java
public WebSearchOrganicResult(URI url, String title) {
    this.url = ensureNotNull(url, "url");
    this.title = ensureNotBlank(title, "title");
    this.content = null;
    this.metadata = null;
}

public WebSearchOrganicResult(URI url, String title, String content) {
    this.url = ensureNotNull(url, "url");
    this.title = ensureNotBlank(title, "title");
    this.content = ensureNotBlank(content, "content");
    this.metadata = null;
}
```

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
```diff
@@ -0,0 +1,204 @@
...
+        this(null, null, content, null);
+    }
+
+    /**
+     * Constructs a WebSearchOrganicResult object with the given content and link.
+     *
+     * @param content The content of the search result.
+     * @param link    The link associated with the search result.
+     */
+    public WebSearchOrganicResult(String content, String link) {
```
langchain4j: Please order parameters consistently. In this ctor order is content, then link.
In other ctors it is link, then content.
I would propose to put parameters in the order of importance and optionality, so in this case:
link -> title -> content -> metadata ?

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
```diff
@@ -0,0 +1,204 @@
...
+    public String content() {
+        return content;
+    }
+
+    /**
+     * Returns the result metadata associated with the search result.
+     *
+     * @return The result metadata associated with the search result.
+     */
+    public Map<String, String> resultMetadata() {
```
langchain4j:
```suggestion
    public Map<String, String> metadata() {
```
nitpick: "result" is redundant here, it is already in "result" class.
Also not consistent with the rest of the fields: content, not resultContent, etc

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchPagination.java
```diff
@@ -0,0 +1,129 @@
...
+
+import static dev.langchain4j.internal.ValidationUtils.ensureGreaterThanZero;
+
+
+/**
+ * Represents pagination information for web search results. Normally it's used as part of a {@link WebSearchResults} to navigate through the search results.
+ * Typically, this includes the current page number, the URL of the next page, the URL of the previous page,
+ * and a map of page numbers to their respective URLs.
+ */
+public class WebSearchPagination {
```
langchain4j: Is this really required? For which use cases will it be useful?
If not, I would rather remove it (less code = less bugs = less work etc)

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
```diff
@@ -0,0 +1,204 @@
...
+ * This includes the title, link, content, and metadata of the web page.
+ * <p>
+ * These results are typically ranked by relevance to the search query.
+ * <p>
+ */
+public class WebSearchOrganicResult {
+
+    private final String title;
+    private final String link;
+    private final String content;
```
langchain4j, line 28: nitpick: "content" sounds like it is a content of a web page, which is not the case, right?
Should it be named "snippet" or something similar?

langchain4j-core/src/test/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetrieverTest.java
```diff
@@ -0,0 +1,497 @@
...
+import java.util.Map;
+import java.util.stream.Stream;
+
+import static java.util.Arrays.asList;
+import static java.util.stream.Collectors.toMap;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.*;
+import static org.mockito.Mockito.*;
+
+class WebSearchContentRetrieverTest {
```
langchain4j, line 24: Would be really great to have an integration test with a real search engine to make sure everything works together as expected

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,114 @@
...
+    public WebSearchContentRetriever(WebSearchEngine webSearchEngine) {
+        this(webSearchEngine, null);
+    }
+
+    /**
+     * Constructs a new WebSearchContentRetriever with the specified web search engine and document splitter.
+     * @param webSearchEngine The web search engine to use for retrieving web search results.
+     * @param documentSplitter The document splitter to use for splitting web search results into smaller text segments.
+     */
+    public WebSearchContentRetriever(WebSearchEngine webSearchEngine, DocumentSplitter documentSplitter) {
```
langchain4j: On a third thought 😅 
I wrongly assumed that web search results will contain the whole text of the pages, but it seems to be not the case, right?
They will contain only small snippets? In such a case I think we should make it possible to (optionally, if user configures) fetch/scrap? contents of web pages, ideally in parallel from all web pages/results, something similar to what we have in `DefaultRetrievalAugmentor` with `Executor`. In such case splitter/embedding model make sense, otherwise - no.
If this is too big of a change for this PR and you want to make it as a separate PR or not at all, then I propose to remove splitting/embedding logic from this class for the moment as this does not make any sense to split/embed small snippets.
Sorry about that!

ghost (commented):
@czelabueno thanks a lot, great job! I have left a few nitpicks, we are almost there 😅

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchEngine.java
czelabueno: makes sense, in fact I had thought about it that way from the beginning and that the `@WebSearchTool` is an abstract class that abstracts, formats and does everything necessary so that any web search that extends this class works automatically as an `@Tool` as well, however `@AiService` does not recognize `@Tools` from a parent abstract class.

My idea was to do something like this (sorry by the pseudocode):

```
public abstract class WebSearchTool {
  protected abstract WebSearchEngine webSearchEngine();

  @Tool("...")
  public String runSearch(String searchTerm){
    WebSearchResults results = webSearchEngie().search(WebSearchRequest.from(searchTerm));
    return results.prettyPrinting();
  }
}
```
Any implementation of a new web search could be a tool doing something as simple as this **E.g. Google Search Implementation**

```
public class GoogleWebSearchTool extends WebSearchTool {

   private final WebSearchEngine googleSearchEngine;

  //ctr
  GoogleWebSearchTool (....){
     this.googleSearchEngine = GoogleSearchEngine.builder()...
  }
  
  @Override
  protected WebSearchEngine webSearchEngie() {
  return googleSearchEngine;
  }
```

I will change my approach to an interface and test if it works as the first option with the current implementation of `@AiServices`, otherwise I will have to modify the `@AiServices` so that it accepts `@tool` from the parent methods of the tool objects. Any another suggestion?

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchResults.java
czelabueno: Sorry if it is not very clear in the documentation. `searchMetadata` is additional information about the execution context of the web search at the time of performing the search. `searchInformation.metadata` is every information associated with the search query, each web search implementation could set more values that `totalResults` and `pageNumber` (fields by default) there if they think needed.

E.g.
`searchMetadata`,  SerpAPI returns the status (success, failed, etc), the final url formed and the execution time, Google API returns something similar, another web search can return other fields that is why the `Map<String, Object>`.
`searchInformation.metadata` returns the total results of the query, total number of pages of the query. Other searcher implementations may have other relevant values that could store it in that (if there images, promotions, etc).

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
czelabueno: yes, initially it was named snipped and I think that because of the confusion that you thought it returned the full content of the page, you asked me to change it to content . Sorry I should have explained it better to you from the beginning. I'll return it as it was.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
czelabueno: Yes it contain only snippets. The traditional web searchers (google, bing, yahoo, etc) already have the intelligence to return the most relevant snippet of the page associated with the user query.
However, they have the problem of being very short and giving the impression that information is missing. I guess that was detected by Tavily AI, which complements a traditional web search with AI and scraped the whole page and rewrites more information in the `content` field.

Following you, we can do the same by using the URI and fetch/scrap the whole page, (perhaps using the `UrlDocumentLoader` or any URL document, suggest me one) and adding the splitter/embedding and the minScore to build a content enriched. I'd use `content` field to do that.

For now, this first version will support only the snippet and in another PR I will implement the content. that's make sense for you?

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
langchain4j: No worries! Sorry for misunderstanding! Most important is that we are on the same page now :)

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
langchain4j: Yeah, totally! Let's roll this first version with just snippets and second version with optional scrapping, splitting, retrieving most relevant parts.

I am not sure it will be 100% reliable to use only `UrlDocumentLoader` as many web sites have scrapping blocks, so we will probably need to find a better solution or just ignore pages which failed to be loaded with `UrlDocumentLoader`...

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchEngine.java
langchain4j: Hmm... Why not to just make it like this?

```java
public class WebSearchTool {

  private final WebSearchEngine webSearchEngine;

  public WebSearchTool(WebSearchEngine webSearchEngine) {
      this.webSearchEngine = ensureNotNull(webSearchEngine, "webSearchEngine");
  }

  @Tool("...")
  public String runSearch(String searchTerm){
      WebSearchResults results = webSearchEngie().search(WebSearchRequest.from(searchTerm));
      return format(results);
  }

  private String format(WebSearchResults results) {
      ...
  }
}
```

This way `WebSearchEngine` can be used with any `WebSearchEngine` and is responsible for formatting results.

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchResults.java
langchain4j: Perharps I still don't fully grasp it... But I still can't understand why can't we merge `searchMetadata`, `searchInformation.metadata` and `searchInformation.xyz` fields together in a single map for simplicity? To me it sounds the same "metadata"...
Could you please explain?

czelabueno: both are metadata at different levels. e.g.:
`search_metadata` = Search engine execution info. More related to each search engine.
`search_information.metadata` = Info related to the search user query, which aren't part of a web page result.
```
{
  "search_metadata": { // info about search execution
    "status": "Success",
    "processed_at": "2022-07-18 23:42:02 UTC",
    "total_time_taken": "1.35s"
  },
  "search_information": { // info about user query results
    "total_results": 343004,
    "page_number": 1,
    "metadata": {
      "contain_images": true,
      "images": [
        "url 1",
        "...."
      ]
    }
  },
  "results": [ // list of organic web pages.
    {
      "title": "page 1",
      "url": "url 1",
      "snippet": "web page snippet"
    }
  ]
}
```
We can simplify it into a single metadata, yes, but I would leave it to various web search implementations to be able to map the data they need. SerpAPI for example response lots of data...

czelabueno:
Hi @langchain4j 
This is a summary about of web search `v1` implementation:

- Component diagram updated in the PR description.
- `WebSeachTool` and `WebSearchContentRetriever` expect be constructed with any `WebSearchEngine` implementation.
- I had a fourth thought 😅, I added `content` field from now, so web result structure looks like: `title`, `url`, `snippet` OR `content`. Some web searches like Tavily Search must implement `WebSearchEngine` to be used as `WebSeachTool` and `WebSearchContentRetriever` out-of-the-box and will be fully compatible. The search engines powered by AI already supports scrapped url to fill content out-of-the-box, in this case any web search implementer would fill snippet or content depending of the search engine capabilities. 
- `WebSearchContentRetriever` v1 return whole text (snippet or content) as a `TextSegment`. Optional scrapping, splitting, retrieving most relevant parts will be part of v2.
- It has been added unit tests and integration tests with a real search engine. Any real search engine must pass parent IT as engine, tool and contentRetriever as my Google Search implementation (part of other PR).
- Graphic below show how works and how will be the JSON response for any search engine for `v1`:
 
WebSearchEngine-decision-graph

### Don't forget:
- This PR move InMemoryEmbeddingStore from langChain4j project to langChain4j-core project according to refactoring plan.
- `Google Search` implementation coming in a separated [PR#641](https://github.com/langchain4j/langchain4j/pull/641) 
- Examples and documentation will be added after merge.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/WebSearchContentRetriever.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.List;
+
+import static dev.langchain4j.internal.ValidationUtils.*;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * A {@link ContentRetriever} that retrieves relevant {@link Content} from the web using a {@link WebSearchEngine}.
+ * <br>
+ * By default, this retriever returns snippet of web pages: one {@link Content} for each web page that a {@link WebSearchEngine} has returned for a given {@link Query}.
```
langchain4j:
```suggestion
 * It returns one {@link Content} for each result that a {@link WebSearchEngine} has returned for a given {@link Query}. Depending on the {@link WebSearchEngine} implementation, the {@link Content#textSegment()} can contain either a snippet of a web page or a complete content of a web page.
```

langchain4j-core/src/main/java/dev/langchain4j/spi/store/embedding/inmemory/InMemoryEmbeddingStoreJsonCodecFactory.java
langchain4j, line 1: cc @geoand

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchInformationResult.java
```diff
@@ -0,0 +1,117 @@
...
+    public int hashCode() {
+        return Objects.hash(totalResults, pageNumber, metadata);
+    }
+
+    @Override
+    public String toString() {
+        return "WebSearchInformationResult{" +
+                "totalResults=" + totalResults +
+                ", pageNumber=" + pageNumber +
+                ", searchInformation=" + metadata +
```
langchain4j:
```suggestion
                ", metadata=" + metadata +
```

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
```diff
@@ -0,0 +1,221 @@
...
+                '}';
+    }
+
+    /**
+     * Converts this WebSearchOrganicResult to a TextSegment.
+     *
+     * @return The TextSegment representation of this WebSearchOrganicResult.
+     */
+    public TextSegment toTextSegment() {
+        return TextSegment.from(content != null ? content : snippet,
```
langchain4j: just to double-check: can both be not null at the same time?

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
```diff
@@ -0,0 +1,221 @@
...
+     * @return The Document representation of this WebSearchOrganicResult.
+     */
+    public Document toDocument() {
+        return Document.from(content != null ? content : snippet,
+                copyToMetadata());
+    }
+
+    private Metadata copyToMetadata() {
+        Metadata docMetadata = new Metadata();
+        docMetadata.add("title", title);
```
langchain4j: I am wondering... should title be prepended to the content/snippet instead?

`TextSegment` text:
```
{{title}}
{{content or snippet}}
```

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchRequest.java
```diff
@@ -0,0 +1,312 @@
...
+
+    private WebSearchRequest(Builder builder){
+        this.searchTerms = ensureNotBlank(builder.searchTerms,"searchTerms");
+        this.maxResults = builder.maxResults;
+        this.language = builder.language;
+        this.geoLocation = builder.geoLocation;
+        this.startPage = getOrDefault(builder.startPage,1);
+        this.startIndex = builder.startIndex;
+        this.safeSearch = getOrDefault(builder.safeSearch,true);
+        this.additionalParams = getOrDefault(builder.additionalParams, new HashMap<>());
```
langchain4j:
```suggestion
        this.additionalParams = getOrDefault(builder.additionalParams, () -> new HashMap<>());
```
minor: to avoid creating a map each time a search is done

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchTool.java
```diff
@@ -0,0 +1,49 @@
...
+
+    /**
+     * Runs a search query on the web search engine and returns a pretty-string representation of the search results.
+     *
+     * @param searchTerm the search user query
+     * @return           a pretty-string representation of the search results
+     */
+    @Tool({"MUST be used to search the web using a web search engine for organic web results.",
+            "An organic web result is a title, a link, a snippet or content and metadata of a web page"})
+    public String runSearch(@P("The search terms is equal to the user message content that looking for") String searchTerm) {
```
langchain4j:
```suggestion
    public String searchWeb(@P("Web search query") String searchTerm) {
```

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchTool.java
```diff
@@ -0,0 +1,49 @@
...
+        this.searchEngine = ensureNotNull(searchEngine, "searchEngine");
+    }
+
+    /**
+     * Runs a search query on the web search engine and returns a pretty-string representation of the search results.
+     *
+     * @param searchTerm the search user query
+     * @return           a pretty-string representation of the search results
+     */
+    @Tool({"MUST be used to search the web using a web search engine for organic web results.",
```
langchain4j:
```suggestion
    @Tool("This tool can be used to perform web searches using search engines such as Google, particularly when seeking information about recent events.")
```

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchTool.java
```diff
@@ -0,0 +1,49 @@
...
+            "An organic web result is a title, a link, a snippet or content and metadata of a web page"})
```
langchain4j:
```suggestion
```

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchTool.java
```diff
@@ -0,0 +1,49 @@
...
+    public String runSearch(@P("The search terms is equal to the user message content that looking for") String searchTerm) {
+        WebSearchResults results = searchEngine.search(searchTerm);
+        return format(results);
+    }
+
+    private String format(WebSearchResults results){
+        return results.results()
+                .stream()
+                .map(organicResult -> "Title: " + organicResult.title() + "\n"
+                        + "URL Source: " + organicResult.url().toString() + "\n"
```
langchain4j:
```suggestion
                        + "Source: " + organicResult.url().toString() + "\n"
```

langchain4j-core/src/test/java/dev/langchain4j/web/search/WebSearchIT.java
```diff
@@ -0,0 +1,103 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * A minimum set of tests that each implementation of {@link WebSearchEngine} must pass.
+ */
+public abstract class WebSearchIT {
```
langchain4j:
```suggestion
public abstract class WebSearchEngineIT {
```

langchain4j-core/src/test/java/dev/langchain4j/web/search/WebSearchIT.java
```diff
@@ -0,0 +1,103 @@
...
+                .as("At least one result should be contains 'weather' and 'New York' ignoring case")
+                .anySatisfy(result -> assertThat(result.snippet())
+                        .containsIgnoringCase("weather")
+                        .containsIgnoringCase("New York"));
+    }
+
+    @Test
+    void should_return_web_results_with_geolocation() {
+        // given
+        String searchTerm = "Who is the current president of France?";
```
langchain4j: minor: maybe we should use some location-specific query? Like put location to "france" and ask "Who is the current president?" without explicitly mentioning France in the query.

langchain4j-core/src/test/java/dev/langchain4j/web/search/WebSearchIT.java
```diff
@@ -0,0 +1,103 @@
...
+        // then
+        assertThat(webSearchOrganicResults).isNotNull();
+        assertThat(webSearchOrganicResults)
+                .as("At least one result should be contains 'Emmanuel Macro' ignoring case")
+                .anySatisfy(result -> assertThat(result.title())
+                        .containsIgnoringCase("Emmanuel Macro"));
+    }
+
+    @Test
+    void should_return_array_of_textSegments(){
```
langchain4j: minor: this should be a unit test imho

langchain4j-core/src/test/java/dev/langchain4j/web/search/WebSearchIT.java
```diff
@@ -0,0 +1,103 @@
...
+        // then
+        assertThat(webSearchResults.toTextSegments().size()).isGreaterThan(0);
+        assertThat(webSearchResults.toTextSegments())
+                .as("At least one result should be contains 'argentina' ignoring case")
+                .anySatisfy(textSegment -> assertThat(textSegment.text())
+                        .containsIgnoringCase("argentina"));
+    }
+
+    @Test
+    void should_return_array_of_documents(){
```
langchain4j: minor: this should be a unit test imho

langchain4j-core/src/test/java/dev/langchain4j/web/search/WebSearchToolIT.java
```diff
@@ -0,0 +1,25 @@
+package dev.langchain4j.web.search;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public abstract class WebSearchToolIT extends WebSearchIT{
+
+    @Test
+    void should_return_pretty_result_as_a_tool(){
```
langchain4j, line 44: I would make it a real IT with `ChatLanguageModel`, to make sure the tool is usable with the LLM

ghost (dismissed):
@czelabueno fantastic job! Just a few minor comments and we can merge!

langchain4j-core/src/main/java/dev/langchain4j/spi/store/embedding/inmemory/InMemoryEmbeddingStoreJsonCodecFactory.java
geoand: Thanks for the heads up!

langchain4j:
HI @czelabueno, how is it going? Please let me know if you need any help with this PR, I think it is almost done and it would be awesome to include this PR in the next releaxe (I hope early next week).

langchain4j-core/src/test/java/dev/langchain4j/web/search/WebSearchIT.java
czelabueno: yeah you're right, in fact `WebSearchResultsTest` add same unit test.

```java
    @Test
    void should_return_array_of_textSegments_with_snippet(){
```

langchain4j-core/src/test/java/dev/langchain4j/web/search/WebSearchIT.java
czelabueno: tha samething here:

```java
  @Test
  void should_return_array_of_documents_with_content(){
```

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
czelabueno: Yes, that is possible. If both be not null, my logic is that `content` prepend to `snippet` bcz it's supposed that `content` represent  web page scrapped by an AI (e.g Tavily Search have a closed AI to do that) or enriched by an external AI (a new LM integration) using embeding model, store, splitter, etc (v2).

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
czelabueno: that is a good question. So far I was considering `{{title}} and {{url}}` as a `TextSegment` or `Document` metadata. Only `{{content or snippet}}` set `TextSegment` or `Document` text.
However, `{{title}} and {{url}}` are mandatories in the constructor and bcz `TextSegment` and `Document` are used for rag purposes, make sense avoid that `text` is empty or null, so we can modify it like this:

`TextSegment.text` and `Document.text`

```
{{title}}   // Not nullable
{{content or snippet}} // Nullable
```

`TextSegment.metadata` and `Document.metadata`

```
{ "key": ''url" }, { "value": "myurl" } // Not nullable
{ "key": ''metadata1" }, { "value": "metadata1" } // Nullable
```

if you're agree, I can apply these changes asap :)

langchain4j-core/src/test/java/dev/langchain4j/web/search/WebSearchToolIT.java
czelabueno: yeah I agree, I've add a real `ChatLanguageModel` and I force that LLM recognize it as a tool and convert from a  `ToolSpecification` to a `ToolExecutionRequest`.

It'd looks better if we can use `AiServices` to execute the tool but due this IT is in the core, I believe will generate us cycle calling issue.

czelabueno:
Hi @langchain4j sorry for delay. I've fixed almost all comments. There is only one open question about the `TextSegment` and everything would be completed. Let me know your thoughts asap to close and merge

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
langchain4j: Yeah, I guess it is better to include `title` in the text of `TextSegment`/`Document`, as it will provide an additional context about the `snippet`/`content` to the scoring and/or language model.

And the metadata will contain url and custom metadata. I guess it is what you've suggested above, right?

langchain4j:
@czelabueno thanks a lot for the quick answer! I have replied to the only left question, I guess we are on the same page there. LMK when you are done with this, I will review it and merge asap.

langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchOrganicResult.java
czelabueno: yeah, that's how it is.

czelabueno:
@langchain4j I'm done and PR is sync with main to easier the merge. Just FYI once this PR is merged, I'll continue the following which is related to this PR:

- Google Search implementation as first web search client (already exists [PR#641](https://github.com/langchain4j/langchain4j/pull/641) but it's pending to update with last changes).
- Add a new rag example using external source like a web searches.
- Add more documentation about web search and examples.

PD: I saw the updates of `CONTRIBUTING.md` guidelines. I created another PRs before that, I'm not sure if you want me to create new issues or just it's valid for new ones? If it's required update issues LMK to regularize it :)

langchain4j:
@czelabueno Awesome, thank you! I will review this PR today.
No, you don't need to create new issues, this requirement was added mainly to:
1. prioritize the work and reviews
2. let other ppl know the issue is being planned/worked on and who is assigned

langchain4j:
@czelabueno I fixed a few tests and moved `InMemoryEmbeddingStore` back to `langchain4j` module, as this change is not required now. We can always move it in the future, if required.

The only problem left is that `WebSearchContentRetrieverIT`, `WebSearchEngineIT` and `WebSearchToolIT` are abstract and not running. But this will be "fixed" in the follow-up PR, right?

czelabueno:
@langchain4j yeah, they do not run on their own because they require a real web search engine. Once they have IT inheritors they will run together. In my [PR#641](https://github.com/langchain4j/langchain4j/pull/641) where I implement Google Search it will execute it all together.

Here the same thing will happen as with `EmbeddingStoreIT` or `EmbeddingStoreWithoutMetadataIT` who were waiting for real implementations of an Embedding Store.

langchain4j:
@czelabueno ah, got it! Then I will merge this now

ghost (approved):
@czelabueno great job, well done!

Issue #645, opened 2024-02-16
[BUG] QianFanChatModel Not Support SystemMessage 


demonSoul:
Please provide as much details as possible, this will help us to deliver a fix as soon as possible.
Thank you!

**Describe the bug**
After calling the toOpenAiMessages method, as long as there is a system message, the throw illegalArgument ("Unknown message type:"+message. type()) will be executed;
![image](https://github.com/langchain4j/langchain4j/assets/32609011/03babad0-b086-449b-9044-c86591220a26)

**Log and Stack trace**
Unknown message type System

**To Reproduce**

**Expected behavior**

**Please complete the following information:**
- LangChain4j version: e.g. 0.27.1
- Java version: 8

**Additional context**

Kugaaa:
The surface cause is `dev.langchain4j.model.qianfan.InternalQianfanHelper#toQianfanMessage` not support system role, throws `IllegalArgumentException`
I checked the official doc, the models in QianFan supported all not support system role, so I think this is the fundamental reason
https://cloud.baidu.com/doc/WENXINWORKSHOP/s/clntwmv7t

demonSoul:
> The surface cause is `dev.langchain4j.model.qianfan.InternalQianfanHelper#toQianfanMessage` not support system role, throws `IllegalArgumentException` I checked the official doc, the models in QianFan supported all not support system role, so I think this is the fundamental reason https://cloud.baidu.com/doc/WENXINWORKSHOP/s/clntwmv7t


`ChatCompletionRequest.Builder builder = ChatCompletionRequest.builder()
                    .messages(toOpenAiMessages(messages))
                    .temperature(temperature)
                    .topP(topP)
                    .penaltyScore(penaltyScore)
                    .system(getSystemMessage(messages))
                    .responseFormat(responseFormat)
                    ;`

**System (getSystemMessage (messages))**, but system messages are processed in the following steps,

Throwing an exception in the previous step actually blocked the subsequent system message processing

I have checked the interface of the Qianfan platform, and although it does not support system roles, there are system parameters in the interface that can be used for model character settings

![image](https://github.com/langchain4j/langchain4j/assets/32609011/10251d9e-3196-4e45-b04c-3800a343a55b)

langchain4j:
@lane12138 would you be able to take a look if system message can be supported as suggested by @demonSoul ?
Thanks!

langchain4j:
Fixed by https://github.com/langchain4j/langchain4j/pull/648

PR #648, opened 2024-02-20
fix  [BUG] QianFanChatModel Not Support SystemMessage

lane12138:
fix  [BUG] QianFanChatModel Not Support SystemMessage #645


ghost (approved):
@lane12138 thank you!

PR #653, opened 2024-02-21
Fix ChatMessage.text() method that is deprecated

jdubois:
The dev.langchain4j.data.message.ChatMessage has been updated, and this PR should use the new API with Azure OpenAI

### DIFF
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
@@ -143,6 +143,9 @@ class InternalAzureOpenAiHelper {
                                 return new ChatMessageTextContentItem(text);
                             } else if (content instanceof ImageContent) {
                                 ImageContent imageContent = (ImageContent) content;
+                                if (imageContent.image().url() == null) {
+                                    throw new IllegalArgumentException("Image URL is not present. Base64 encoded images are not supported at the moment.");
+                                }
                                 ChatMessageImageUrl imageUrl = new ChatMessageImageUrl(imageContent.image().url().toString());
                                 return new ChatMessageImageContentItem(imageUrl);
                             } else {
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
@@ -119,18 +119,18 @@ class InternalAzureOpenAiHelper {
     }
 
     public static com.azure.ai.openai.models.ChatRequestMessage toOpenAiMessage(ChatMessage message) {
-        if (message.type().messageClass().equals(AiMessage.class)) {
+        if (message instanceof AiMessage) {
             AiMessage aiMessage = (AiMessage) message;
             ChatRequestAssistantMessage chatRequestAssistantMessage = new ChatRequestAssistantMessage(getOrDefault(aiMessage.text(), ""));
             chatRequestAssistantMessage.setFunctionCall(functionCallFrom(message));
             return chatRequestAssistantMessage;
-        } else if (message.type().messageClass().equals(ToolExecutionResultMessage.class)) {
+        } else if (message instanceof ToolExecutionResultMessage) {
             ToolExecutionResultMessage toolExecutionResultMessage = (ToolExecutionResultMessage) message;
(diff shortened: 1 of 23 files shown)


langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -119,19 +119,32 @@ public static List<com.azure.ai.openai.models.ChatRequestMessage> toOpenAiMessag
     }
 
     public static com.azure.ai.openai.models.ChatRequestMessage toOpenAiMessage(ChatMessage message) {
-        if (message instanceof AiMessage) {
-            ChatRequestAssistantMessage chatRequestAssistantMessage = new ChatRequestAssistantMessage(getOrDefault(message.text(), ""));
+        if (message.type().messageClass().equals(AiMessage.class)) {
```
langchain4j: This can be simplified to `message instanceOf AiMessage`, the same for the rest of types below

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -119,19 +119,32 @@ public static List<com.azure.ai.openai.models.ChatRequestMessage> toOpenAiMessag
...
+            SystemMessage systemMessage = (SystemMessage) message;
+            return new ChatRequestSystemMessage(systemMessage.text());
+        } else if (message.type().messageClass().equals(UserMessage.class)) {
+            UserMessage userMessage = (UserMessage) message;
+            ChatRequestUserMessage chatRequestUserMessage;
+            if (userMessage.hasSingleText()) {
+                chatRequestUserMessage = new ChatRequestUserMessage(((TextContent) userMessage.contents().get(0)).text());
+            } else {
+                chatRequestUserMessage = new ChatRequestUserMessage(userMessage.contents().stream()
+                        .map(content -> ((TextContent) content).text())
```
langchain4j: `content` can be also `ImageContent`

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
jdubois: From a coding perspective, I liked my approach more, but then indeed 1) it's shorter and clearer and 2) I just learned it was more efficient to use instances !

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
jdubois: Thanks, fixed!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -120,18 +120,41 @@ public static List<com.azure.ai.openai.models.ChatRequestMessage> toOpenAiMessag
...
+                chatRequestUserMessage = new ChatRequestUserMessage(((TextContent) userMessage.contents().get(0)).text());
+            } else {
+                chatRequestUserMessage = new ChatRequestUserMessage(userMessage.contents().stream()
+                        .map(content -> {
+                            if (content instanceof TextContent) {
+                                String text = ((TextContent) content).text();
+                                return new ChatMessageTextContentItem(text);
+                            } else if (content instanceof ImageContent) {
+                                ImageContent imageContent = (ImageContent) content;
+                                ChatMessageImageUrl imageUrl = new ChatMessageImageUrl(imageContent.image().url().toString());
```
langchain4j, line 149: `imageContent.image().url()` can return `null` in case image is provided as `base64Data`, see `ImageContent` constructors for all possible combinations

jdubois: Indeed! But as we can't send a base64 string (the `ChatMessageImageContentItem` can only support URLs), what do we do in case of a base64 data, do we drop it?

langchain4j: Maybe throw an exception in such case? Dropping it silently will be very confusing for the user

jdubois:
Sorry, my Git history was completely broken by IntelliJ.... Closing this and opening a new PR, with this fix (and the Base64 issue solved)

PR #656, opened 2024-02-22
Remove direct dependency to the langchain4j-open-ai

jdubois:
There is currently a direct dependency on the `langchain4j-open-ai` module in order to use the `OpenAiTokenizer` class.

This code moves that dependency to the `test` scope, so that end-users don't get the full `langchain4j-open-ai` module (and its dependencies) in their classpath at runtime.


ghost (approved):
@jdubois nice, thank you!

PR #672, opened 2024-02-24
Dirty POC: building state machine "agent" with AI Services

ghost:
The idea is to define a set of states and rules for transitioning between them.
For each state, user can define a separate system message (prompt) and set of tools.
LLM can trigger transition from one state to another if it thinks it is necessary.
Transition can be direct or conditional on some predicate.

For example, LLM should not have access to the "cancelBooking" tool before it collected all the information from the customer and validated that the booking exists.

You can run `AiServicesWithStatesTest` and follow the script commented at the end of the file.
Notice how with state the system message and available set of tools is changing (logged in the console).

Credits: [From Clicks to Conversations: Designing LLM-powered Apps](https://youtu.be/WbhaZveZ2TM?t=1036) by @malywut

image

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/JsonSchemaProperty.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/JsonSchemaProperty.java
@@ -1,7 +1,9 @@
 package dev.langchain4j.agent.tool;
 
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Objects;
+import java.util.Set;
 
 import static dev.langchain4j.internal.Utils.quoted;
 import static java.util.Collections.singletonMap;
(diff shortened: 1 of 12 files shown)


mariofusco:
I have a similar need for a PoC that I'm developing these days, see https://github.com/mariofusco/quarkus-drools-llm/tree/main?tab=readme-ov-file#the-airline-chatbot-example

There I implemented the state machine with a rule engine and more in general it would be great to have a pluggable mechanism to allow using different options. If it's ok for you I could give a try to what you're proposing here (and maybe contribute with some changes with another pull request) and validate it with my use case.

langchain4j:
@mariofusco sounds great, I would definitiely love to see how this looks like with drools! Having a pluggable mechanism is a good idea. But first we need to see whether giving `AiService` N states is easier/clearer/more_flexible than having N `AiService`s (one per state) and allow the user to route and manage that himself using whatever he wants. WDYT?

mariofusco:
For now I'm not sure to be honest. The N `AiService`s (1 per state) approach actually works and indeed is the one I used for my PoC, but simply because it is the only option I had. Then I googled to check if langchain4j already had something to cover this requirement out-of-the-box and bumped into this pull request. For now I want to give it a try and see how it comes along and then we can further continue this discussion. Usually I don't know what I'm doing or about to do before I start writing down some code :)

malywut:
I am following this :)
Any feedback on the idea is much appreciated.

PR #673, opened 2024-02-24
Automatically review PRs with OpenRewrite recipe runs

timtebeek:
As discussed on https://github.com/langchain4j/langchain4j/pull/622#pullrequestreview-1895918197, this PR adds a workflow to automatically review pull requests. The rules are configurable and can be extended; I'd thought to start small, and then expand in separate pull requests where I both make the changes to the project, and add the recipe to the list of suggestions for future pull requests.

As an example of what this looks like in practice, have a look at https://github.com/openrewrite/rewrite-migrate-java/pull/421
Notice how we have a bot `github-actions` that places code suggestion comments that can be quickly applied by the user. This should make for a reasonably friction-free enforcement of code standards going forward, using mostly standard components + https://github.com/googleapis/code-suggester

image

I'll self review with a few small callouts below, just to make review easier.

Assumes that we first merge my other pull request, such that we don't get any suggestions for items fixed there.
- #622

## Summary by CodeRabbit

- **New Features**
    - Enhanced GitHub Actions workflows for pull request processing:
        - Automatically suggests improvements using the Google Code Suggester tool.
        - Securely processes untrusted code by running OpenRewrite recipes and uploading patches.
- **Refactor**
    - Improved test assertions in `VertexAiEmbeddingModelIT.java` and `AiServicesBuilderTest.java` for better code quality.


.github/workflows/receive-pr.yml
```diff
@@ -0,0 +1,54 @@
+# Description: This workflow runs OpenRewrite recipes against opened pull request and upload the patch.
+# Since this pull request receives untrusted code, we should **NOT** have any secrets in the environment.
+# https://securitylab.github.com/research/github-actions-preventing-pwn-requests/
+---
+name: receive-pr
+
+on:
+  pull_request:
+    types: [opened, synchronize]
```
timtebeek, line 9: Runs are quick; around ~1 minute locally here, so I figured no harm in doing this for every sync of a PR. Can be changed to just `opened` if you want to limit the runs.

.github/workflows/comment-pr.yml
```diff
@@ -0,0 +1,56 @@
...
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{github.event.workflow_run.head_branch}}
+          repository: ${{github.event.workflow_run.head_repository.full_name}}
+
+      # Download the patch
+      - uses: actions/download-artifact@v4
+        with:
+          name: patch
+          github-token: ${{ secrets.GH_PAT_ACTIONS_READ }}
```
timtebeek: This secret does not yet exist, and I of course don't have access to create it for this organization. It should be a GitHub personal access token with `actions:read` permissions on the target repo. This is used to download the patch and the PR number from the triggering workflow. Fairly innocuous, but unfortunately a manual step once.

.github/workflows/comment-pr.yml
```diff
@@ -0,0 +1,56 @@
...
+          PR_NUMBER=$(cat pr_number.txt)
+          echo "PR_NUMBER=$PR_NUMBER" >> $GITHUB_ENV
+          rm pr_number.txt
+
+      # Post suggestions as a comment on the PR
+      - uses: googleapis/code-suggester@v4
+        with:
+          command: review
+          pull_number: ${{ env.PR_NUMBER }}
+          git_dir: '.'
```
timtebeek, line 56: This is the part of the workflows which actually posts the comments; I've seen it work quite well in practice, but in very rare cases a suggestion can be off a tiny bit. If that ever becomes an issue then [reviewdog](https://github.com/reviewdog/reviewdog) is an alternative.

pom.xml
```diff
@@ -124,5 +124,33 @@
         </plugins>
     </reporting>
 
+    <profiles>
+        <profile>
+            <id>openrewrite</id>
```
timtebeek, line 175: Technically we don't absolutely need a profile for the workflows to work; but by having a profile here we more easily can bump the versions of the plugin and dependency, and folks can run the same fixes locally with just:
```bash
mvn --activate-profiles openrewrite org.openrewrite.maven:rewrite-maven-plugin:run
```

pom.xml
```diff
@@ -124,5 +124,33 @@
...
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.openrewrite.maven</groupId>
+                        <artifactId>rewrite-maven-plugin</artifactId>
+                        <version>5.23.1</version>
+                        <configuration>
+                            <activeRecipes>
+                                <recipe>org.openrewrite.java.OrderImports</recipe>
```
timtebeek: This one has not been applied yet; I do feel it would be beneficial to have a standard lightly enforced in the manner though, as IDEAs and settings will change between users, and automated enforcement then stops all discussions.

.github/workflows/comment-pr.yml
```diff
@@ -0,0 +1,56 @@
...
+      - completed
+
+jobs:
+  post-suggestions:
+    # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    env:
+      # https://docs.github.com/en/actions/reference/authentication-in-a-workflow#permissions-for-the-github_token
+      ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
timtebeek, line 20: This token you get for free with GitHub Actions, and by default allows write access to the PR. That's also why we should not execute any user contributed code in this PR, and why we have the split between receive and comment into two workflows, as recommended.

.github/workflows/comment-pr.yml
timtebeek: Looks like this also works with GITHUB_TOKEN, so this comment can be discarded after https://github.com/langchain4j/langchain4j/pull/673/commits/225b8603a820e4e6a7152df5695635c03f30c4f5

timtebeek:
Let me know if you'd like me to take out the OrderImports for now, or anything else I can do to help this PR along. The goal really is to minimize the noise on PRs while automating away common issues that could otherwise be (re)introduced.

pom.xml
MBoegers: I think having a dedicated profile is the right way.
As maven documentation states "Under other circumstances, a slightly different dependency set will be required, (...)" and that's the situation here.

Maybe chaining the plugin execution to a specific build step (verify) would be even more maven-ish.

timtebeek: Thanks! You mean to only add it to a build step in this particular profile? That could make sense indeed. I hadn't wanted to tie it into the lifecycle originally, as to not cause delays for folks, but within a profile that makes sense.

langchain4j:
@timtebeek thanks a lot! And sorry for the delayed response, I will try to check this ASAP.

timtebeek:
All good! For what it's worth @MBoegers tried the AI suggestions here and they didn't work, hence not applied. :)

langchain4j:
@timtebeek thank you and so sorry for the late reaction!

timtebeek:
> @timtebeek thank you and so sorry for the late reaction!

No problem at all! I know what it's like. :) Good to see this merged!

I've opened up a follow up pull request that clears out some new suggestions from the main branch, such that PRs only show suggestions on new/changed files.
- https://github.com/langchain4j/langchain4j/pull/1904

timtebeek:
You'll also want to enable "Always suggest updating pull request branches" on https://github.com/langchain4j/langchain4j/settings
image

That ensures that for any already open PRs the latest changes on main are reflected to post suggestions before a merge.

PR #678, opened 2024-02-26
Fix ChatMessage.text() method that is deprecated

jdubois:
Replace #653 where the Git history got messed up.


ghost (approved):
@jdubois thanks a lot!

PR #679, opened 2024-02-27
Add streaming API for Bedrock Anthropics

michalkozminski:
Currently, langchain4j doesn't have support for streaming of Bedrock Anthropics model, this PR tries to address it.

Adding support for Anthropics v2 and v2:1 streaming. 
New tests disabled due to need for AWS credentials but pass when enabled.


caosDvlp:
How about adding the option to have Tools with AWS Bedrock too?
In bedrock they are called "Agents", it would be nice to have also this option with Bedrock!

michalkozminski:
@caosDvlp I didn't explore it, it might be something to add as a next step.

caosDvlp:
Great @michalkozminski I would really appreciate it actually! :)
Thanks!

michalkozminski:
@langchain4j Would be amazing if we could get streaming support as it's currently not supported by bedrock integration. Could you please give a look on this PR?

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/BedrockStreamingChatModelIT.java
```diff
@@ -0,0 +1,57 @@
...
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class BedrockStreamingChatModelIT {
+    @Test
+    @Disabled("To run this test, you must have provide your own access key, secret, region")
+    void testBedrockAnthropicStreamingChatModel() throws ExecutionException, InterruptedException, TimeoutException {
+        BedrockAnthropicStreamingChatModel bedrockChatModel = BedrockAnthropicStreamingChatModel
+                .builder()
+                .temperature(0.50f)
```
langchain4j, line 23: minor: can we change this to `double` to not force the user adding "f"?

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/BedrockStreamingChatModelIT.java
```diff
@@ -0,0 +1,57 @@
...
+                .maxTokens(300)
+                .region(Region.US_EAST_1)
+                .maxRetries(1)
+                .build();
+
+        CompletableFuture<String> futureAnswer = new CompletableFuture<>();
```
langchain4j, line 29: minor: please use `TestStreamingResponseHandler`, see `MistralAiStreamingChatModelIT` as example.
you will need to import
```
        <dependency>
            <classifier>tests</classifier>
            <type>test-jar</type>
            <scope>test</scope>
        </dependency>
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockAnthropicStreamingChatModel.java
```diff
@@ -0,0 +1,33 @@
...
+import dev.langchain4j.model.bedrock.internal.AbstractBedrockStreamingChatModel;
+import lombok.Builder;
+import lombok.Getter;
+import lombok.experimental.SuperBuilder;
+
+@Getter
+@SuperBuilder
+public class BedrockAnthropicStreamingChatModel extends AbstractBedrockStreamingChatModel {
+    @Builder.Default
+    private final BedrockAnthropicChatModel.Types model = BedrockAnthropicChatModel.Types.AnthropicClaudeV2;
```
langchain4j, line 12: minor: would be nice to allow setting model id _also_ as a `String`

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/internal/AbstractBedrockStreamingChatModel.java
```diff
@@ -0,0 +1,87 @@
...
+                .contentType("application/json")
+                .accept("application/json")
+                .build();
+
+        AtomicReference<String> finalCompletion = new AtomicReference<>("");
+
+        InvokeModelWithResponseStreamResponseHandler.Visitor visitor = InvokeModelWithResponseStreamResponseHandler.Visitor.builder()
+                .onChunk(chunk -> {
+                    StreamingResponse sr = Json.fromJson(chunk.bytes().asUtf8String(), StreamingResponse.class);
+                    finalCompletion.set(finalCompletion.get() + sr.completion);
```
langchain4j, line 56: just curious: why not using `StringBuffer` instead?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/internal/AbstractBedrockStreamingChatModel.java
```diff
@@ -0,0 +1,87 @@
...
+                    handler.onNext(sr.completion);
+                })
+                .build();
+
+        InvokeModelWithResponseStreamResponseHandler h = InvokeModelWithResponseStreamResponseHandler.builder()
+                .onEventStream(stream -> stream.subscribe(event -> event.accept(visitor)))
+                .onComplete(() -> {
+                    handler.onComplete(Response.from(new AiMessage(finalCompletion.get())));
```
langchain4j, line 64: Could you please add `TokenUsage` and `FinishReason`?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/internal/AbstractSharedBedrockChatModel.java
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @param message chat message
+     * @return string
+     */
+    protected String chatMessageToString(ChatMessage message) {
+        switch (message.type()) {
+            case SYSTEM:
+                return message.text();
+            case USER:
+                return humanPrompt + " " + message.text();
```
langchain4j, line 62: I have missed this in the initial implementation somehow. Why is it needed to append `Human:`, etc?
I can imagine htis cna be useful for Llama2 and co, but not for all models.
I would at least make those prefixes empty by default and remove the space if prompt is empty.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/internal/AbstractSharedBedrockChatModel.java
```diff
@@ -0,0 +1,110 @@
...
+                .filter(message -> message.type() == ChatMessageType.SYSTEM)
+                .map(ChatMessage::text)
+                .collect(joining("\n"));
+
+        final String userMessages = messages.stream()
+                .filter(message -> message.type() != ChatMessageType.SYSTEM)
+                .map(this::chatMessageToString)
+                .collect(joining("\n"));
+
+        final String prompt = String.format("%s\n\n%s\n%s", context, userMessages, ASSISTANT_PROMPT);
```
langchain4j, line 83: Again, I know it was already in there. But `ChatLanguageModel` is not meant to be implemented like this. For that there is a `LanguageModel`. Doesn't aws support separate messages instead of a single prompt? At least for anthropic...

ghost (commented):
@michalkozminski thanks a lot! And pardon for the late review...

langchain4j:
@michalkozminski I have merged this PR to include it in today's release. Please address left comments in a separate PR. Thanks!

michalkozminski:
@langchain4j thanks will do

PR #686, opened 2024-03-01
Easy RAG

ghost:
[Easy RAG Example](https://github.com/langchain4j/langchain4j-examples/blob/main/rag-examples/src/main/java/_1_easy/Easy_RAG_Example.java)

Implementing RAG applications is hard. Especially for those who are just getting started exploring LLMs and RAG.

This PR introduces an "Easy RAG" feature that should help developers to get started with RAG as easy as possible.

With it, there is no need to learn about chunking/splitting/segmentation, embeddings, embedding models, vector databases, retrieval techniques and other RAG-related concepts.

This is similar to how one can simply upload one or multiple files into [OpenAI Assistants API](https://platform.openai.com/docs/assistants/overview) and the LLM will automagically know about their contents when answering questions.

Easy RAG is using local embedding model running in your CPU (GPU support can be added later).
Your files are ingested into an in-memory embedding store.

Please note that "Easy RAG" will not replace manual RAG setups and especially [advanced RAG techniques](https://github.com/langchain4j/langchain4j/pull/538), but will provide an easier way to get started with RAG.
The quality of an "Easy RAG" should be sufficient for demos, proof of concepts and for getting started.

To use "Easy RAG", simply import `langchain4j-easy-rag` dependency that includes everything needed to do RAG:
- Apache Tika document loader (to parse all document types automatically)
- Quantized [BAAI/bge-small-en-v1.5](https://huggingface.co/BAAI/bge-small-en-v1.5) in-process embedding model which has an impressive (for it's size) 51.68 [score](https://huggingface.co/spaces/mteb/leaderboard) for retrieval

Here is the proposed API:

```java
List<Document> documents = FileSystemDocumentLoader.loadDocuments(directoryPath); // one can also load documents recursively and filter with glob/regex

EmbeddingStore<TextSegment> embeddingStore = new InMemoryEmbeddingStore<>(); // we will use an in-memory embedding store for simplicity

EmbeddingStoreIngestor.ingest(documents, embeddingStore);

                .chatLanguageModel(model)
                .contentRetriever(EmbeddingStoreContentRetriever.from(embeddingStore))
                .build();

String answer = assistant.chat("Who is Charlie?"); // Charlie is a carrot...
```

`FileSystemDocumentLoader` in the above code loads documents using `DocumentParser` available in classpath via SPI, in this case an `ApacheTikaDocumentParser` imported with the `langchain4j-easy-rag` dependency.

The `EmbeddingStoreIngestor` in the above code:
- splits documents into smaller text segments using a `DocumentSplitter` loaded via SPI from the `langchain4j-easy-rag` dependency. Currently it uses `DocumentSplitters.recursive(300, 30, new HuggingFaceTokenizer())`
- embeds text segments using an `AllMiniLmL6V2QuantizedEmbeddingModel` loaded via SPI from the `langchain4j-easy-rag` dependency
- stores text segments and their embeddings into the specified embedding store

When using `InMemoryEmbeddingStore`, one can serialize/persist it into a JSON string on into a file.
This way one can skip loading documents and embedding them on each application run.

It is easy to customize the ingestion in the above code, just change
```java
EmbeddingStoreIngestor.ingest(documents, embeddingStore);
```
into
```java
EmbeddingStoreIngestor ingestor = EmbeddingStoreIngestor.builder()
                //.documentTransformer(...) // you can optionally transform (clean, enrich, etc) documents before splitting
                //.documentSplitter(...) // you can optionally specify another splitter
                //.textSegmentTransformer(...) // you can optionally transform (clean, enrich, etc) segments before embedding
                //.embeddingModel(...) // you can optionally specify another embedding model to use for embedding
                .build();

ingestor.ingest(documents)
```

Over time, we can add an auto-eval feature that will find the most suitable hyperparametes for a given documents (e.g. which embedding model to use, which splitting method, possibly advanced RAG techniques, etc.) so that "easy RAG" can be comparable to the "advanced RAG".

Related:
https://github.com/langchain4j/langchain4j-embeddings/pull/16


geoand:
cc @jmartisk  @cescoffier

cescoffier:
I like the idea. How do you compute the chunk size and the number of relevant documents to include? Should it be configurable with sensible defaults?

jmartisk:
This is really cool and easy. +1 to configurable document size and count

dliubars:
Size of the segment can be configurable, but it will have a sensible static default (probably something like 300 tokens and 30 tokens overlap).

Over time, we can implement an eval step to find the most optimal size for the given document(s). Also, it does hot have to be fixed, it can be variable in some range (min, max) to minimize information loss during embedding and maximize recall during retrieval.

I was working on the side to implement a hybrid splitting algorithm that takes into account both structural (headers, newlines, paragraphs, lists, etc) and semantic (using embeddings) signals to find the best splitting point. One can define min/max segment size and it will split given text into segments with size within that range while minimzing information loss by looking at dispersion of individual sentences in the potential segment (their deviation from the overall segment embedding (centroid)). Maybe this can be used here.

langchain4j:
BTW all credits for the idea goes to @dandreadis

maxandersen:
+1; maybe i'm missing something but isn't this not already feasible today if we just enabled sensible defaults on the existing api (or maybe this is exactly that) ?

Thus to be clear - 100% for enabling this kind of easy setup; especially if behind it is use of the existing apis so we know users can "escape" and adjust when/where needed.

document-parsers/langchain4j-document-parser-apache-tika/src/main/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParser.java
```diff
@@ -0,0 +1,73 @@
...
+        this.parseContext = getOrDefault(parseContext, new ParseContext());
+    }
+
+    // TODO allow automatically extract metadata (e.g. creator, last-author, created/modified timestamp, etc)
+
+    @Override
+    public Document parse(InputStream inputStream) {
+        try {
+            parser.parse(inputStream, contentHandler, metadata, parseContext);
+            String text = contentHandler.toString();
```
sberyozkin, line 65: Have a look at https://github.com/quarkiverse/quarkus-tika/blob/main/runtime/src/main/java/io/quarkus/tika/TikaParser.java at how the embedded content from the images etc can be collected as well. It is not essential for this PR but can be useful in the future

langchain4j: @sberyozkin thanks a lot, this will definitely be very useful once we add support for `Image`s and `Table`s in `Document`!

langchain4j:
@maxandersen Quarkus extension could probably enable sensible defaults (if not yet), but for the non-Quarkus folks there should be a way to do RAG hastle-free without a need to manually setup document loaders, splitters, embedding model, embedding store and orchestrating the ingestion, thus the proposed solution. Later it will be expanded to  support more advanced RAG strategies and do auto-evals. So imho this deserves it's own class/name/abstraction. Also, to stress with the name that this is an easy/demo/PoC feature.

document-parsers/langchain4j-document-parser-apache-tika/src/main/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParser.java
langchain4j: @sberyozkin sorry, just noticed that my comment was in a pending state 😬

geoand:
I like it a lot!

PR #692, opened 2024-03-04
Gemini function calling support

glaforge:
- Add Gemini function calling support, in both streaming and non-streaming models
- Upgrade the Gemini Java SDK

## Summary by CodeRabbit

- **New Features**
	- Introduced functionality for converting tool execution requests into function calls and vice versa, enhancing interaction capabilities.
	- Expanded message handling in chat models to support different message types and tool execution results.
	- Added support for generating chat messages based on tool specifications, allowing for more dynamic conversations.

- **Enhancements**
	- Improved type conversion and message construction processes for better efficiency and reliability.

- **Tests**
	- Added comprehensive test coverage for new functionalities, ensuring robustness and reliability of conversions and message handling.


langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/FunctionCallHelper.java
```diff
@@ -0,0 +1,130 @@
...
+                break;
+            case STRUCT_VALUE:
+                HashMap<String, Object> mapForStruct = new HashMap<>();
+                value.getStructValue().getFieldsMap().forEach((key, val) -> mapForStruct.put(key, unwrapProtoValue(val)));
+                unwrappedValue = mapForStruct;
+                break;
+            case LIST_VALUE:
+                unwrappedValue = value.getListValue().getValuesList().stream().map(FunctionCallHelper::unwrapProtoValue).collect(Collectors.toList());
+                break;
+            default: // NULL_VALUE, KIND_NOT_SET, and default
```
langchain4j, line 97: can the value in the `KIND_NOT_SET` case be not `null`?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/PartsMapper.java
```diff
@@ -36,14 +40,45 @@ class PartsMapper {
     }
 
     static List<Part> map(ChatMessage message) {
+        if (message instanceof AiMessage) {
+            AiMessage aiMessage = (AiMessage) message;
+
+            if (aiMessage.hasToolExecutionRequests()) {
+                return singletonList(Part.newBuilder()
+                    .setFunctionCall(
+                        //TODO: handling one function call, but can there be several?
```
langchain4j, line 49: Can we test this empirically? Some models (e.g. latest from OpenAI) support multiple function calls in parallel: `OpenAiChatModelIT.should_execute_multiple_tools_in_parallel_then_answer`

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/RoleMapper.java
```diff
@@ -6,6 +6,7 @@ class RoleMapper {
 
     static String map(ChatMessageType type) {
         switch (type) {
+            case TOOL_EXECUTION_RESULT:
```
langchain4j, line 9: There seems to be a [function](https://ai.google.dev/docs/function_calling#expandable-3) role?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/StreamingChatResponseBuilder.java
```diff
@@ -1,18 +1,22 @@
...
 
 import java.util.List;
+import java.util.Optional;
 
 class StreamingChatResponseBuilder {
 
     private final StringBuffer contentBuilder = new StringBuffer();
+
+    private boolean hasFunctionCall = false;
+    private FunctionCall functionCall;
```
langchain4j: Please make these 2 new fields `volatile` as there is no guarantee that each event will come on the same thread

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/StreamingChatResponseBuilder.java
```diff
@@ -1,18 +1,22 @@
...
 import dev.langchain4j.model.output.TokenUsage;
 
 import java.util.List;
+import java.util.Optional;
 
 class StreamingChatResponseBuilder {
 
     private final StringBuffer contentBuilder = new StringBuffer();
+
+    private boolean hasFunctionCall = false;
```
langchain4j: minor: do we really need this `boolean` field? Can we use `functionCall` null/not null?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
```diff
@@ -101,16 +99,51 @@ public VertexAiGeminiChatModel(GenerativeModel generativeModel,
...
     @Override
     public Response<AiMessage> generate(List<ChatMessage> messages) {
-
         List<Content> contents = ContentsMapper.map(messages);
-
         GenerateContentResponse response = withRetry(() -> generativeModel.generateContent(contents, generationConfig), maxRetries);
 
-        return Response.from(
+        Content content = ResponseHandler.getContent(response);
+        Optional<Part> containsFunctionCall = content.getPartsList().stream().filter(Part::hasFunctionCall).findFirst();
```
langchain4j: minor: in theory, there should be no funciton calls when using `generate(List<ChatMessage>)` method because there are no tool specs to start with. But user could use both `generate(List<ChatMessage>)` and `generate(List<ChatMessage>, List<ToolSpecification>)` and `generate(List<ChatMessage>, List<ToolSpecification>)` calls `generate(List<ChatMessage>)`. It is just a bit unobvious from the current implementation, maybe worth adding a comment that this code will be executed also when there are tools?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/FunctionCallHelper.java
```diff
@@ -0,0 +1,130 @@
...
+                unwrappedValue = null;
+                break;
+        }
+        return unwrappedValue;
+    }
+
+    static Tool convertToolSpecifications(List<ToolSpecification> toolSpecifications) {
```
langchain4j, line 104: Shouldn't it be a _list_ of `Tool`s from a list of `ToolSpecification`s?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
```diff
@@ -101,16 +99,51 @@ public VertexAiGeminiChatModel(GenerativeModel generativeModel,
...
+    public Response<AiMessage> generate(List<ChatMessage> messages, List<ToolSpecification> toolSpecifications) {
+        try {
+            Tool tool = FunctionCallHelper.convertToolSpecifications(toolSpecifications);
+            //TODO: hack for now, till the next release that will have a tool param
+            // in the underlying generateContent() method
+            this.generativeModel.setTools(Collections.singletonList(tool));
+            return generate(messages);
+        } finally {
+            //TODO: remove the finally block (see above todo)
+            this.generativeModel.setTools(Collections.emptyList());
```
langchain4j: Should we create a new instance of `generativeModel` in this method for the duration of this call instead of sharing the same instance and hope there will be no concurrent calls?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
```diff
@@ -101,16 +99,51 @@ public VertexAiGeminiChatModel(GenerativeModel generativeModel,
...
+        }
+    }
+
+    @Override
+    public Response<AiMessage> generate(List<ChatMessage> messages, ToolSpecification toolSpecification) {
+        if (toolSpecification == null) {
+            return generate(messages);
+        } else {
+            return generate(messages, Collections.singletonList(toolSpecification));
```
langchain4j, line 159: Does Gemini has a way to _force_ the model to call the provided tool?

The idea of this method (`generate(List<ChatMessage>, ToolSpecification)`) was to guarantee that provided `ToolSpecification` will be called.

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/FunctionCallHelperTest.java
```diff
@@ -0,0 +1,125 @@
...
+import static dev.langchain4j.model.vertexai.FunctionCallHelper.unwrapProtoValue;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class FunctionCallHelperTest {
+    @Test
+    void should_unwrap_proto_values() {
+        // check basic values
+        assertThat(unwrapProtoValue(Value.newBuilder().setStringValue("hello").build())).isEqualTo("hello");
+        assertThat(unwrapProtoValue(Value.newBuilder().setBoolValue(false).build())).isEqualTo(false);
+        assertThat(unwrapProtoValue(Value.newBuilder().setNumberValue(1.23).build())).isEqualTo(1.23);
```
langchain4j, line 31: minor: would be nice to test `int`, `long` and `float` as well

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -213,38 +221,85 @@ void should_accept_text_and_multiple_base64_images() {
...
-                .containsIgnoringCase("cat")
-                .containsIgnoringCase("dog")
-                .containsIgnoringCase("dice");
+            .containsIgnoringCase("cat")
+            .containsIgnoringCase("dog")
+            .containsIgnoringCase("dice");
+    }
+
+    @Test
+    void should_accept_tools_for_function_calling() {
```
langchain4j, line 260: minor: I plan to have a unified set of standartized tests for models the same way as `EmbeddingStoreIT` soon. Could you please copy/paste here tool-related tests from `OpenAiChatModelIT`?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -213,38 +221,85 @@ void should_accept_text_and_multiple_base64_images() {
...
+            "{\"location\":\"Paris\",\"forecast\":\"sunny\", \"temperature\": 20}");
+        allMessages.add(toolExecResMsg);
+
+        Response<AiMessage> weatherResponse = model.generate(allMessages);
+
+        // then
+        System.out.println("Answer: " + weatherResponse.content().text());
+        assertThat(weatherResponse.content().text()).containsIgnoringCase("sunny");
     }
 }
```
langchain4j, line 305: minor: would be awesome to extend tests in `AiServicesWithToolsIT` to be parameterized with the model (same way as in `AiServicesWithRagIT`) and add Gemini model to the OpenAI.

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
```diff
@@ -336,4 +341,55 @@ void should_accept_text_and_multiple_images_from_different_sources() {
                 .containsIgnoringCase("dog")
                 .containsIgnoringCase("dice");
     }
+
+    @Test
+    void should_accept_function_call() {
```
langchain4j, line 346: same comments as above

ghost (commented):
@glaforge thanks a lot, this looks great! I have left some comments, please check.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/FunctionCallHelper.java
glaforge: To be honest, I don't know when the `KIND_NOT_SET` case ever happens, so I don't know what kind of value it could contain, and I couldn't even find how to retrieve that value, so that's why I've just returned `null`.
Normally, I don't think this case happens in the Gemini API.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/PartsMapper.java
glaforge: Thanks for the pointer, I'll have a look!

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/RoleMapper.java
glaforge: This is a link to the Google AI / Labs documentation, but there's a similar [page](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#rest) that hints at a `function` role too. This actually new to me, but I haven't seen the service endpoint to ever use/return that role. I'm going to investigate further and ask around — but at least my impl works as it is currently without that role for now.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/StreamingChatResponseBuilder.java
glaforge: OK

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/StreamingChatResponseBuilder.java
glaforge: We can, I'll change that.

glaforge: Covered in 424b280991f53480981607c2695196ed50956738

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/RoleMapper.java
glaforge: Currently, the service rejects roles that are not `user` or `model`, so I'll stick with those two for now, and will update if ever a `function` role is indeed supported.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
glaforge: For now at least, unlike OpenAI, there's no notion of mandatory tool to call or not. So my understanding is that the model decides whether to use tools or not on its own without supervision (it's like OpenAI's `auto` mode, I guess.

However, in Gemini, it's also possible to set tools at model definition time, so we don't need to pass tools at each generation request. But soon in the underlying Java SDK, we'll be able to set tools on a per generation request basis, and they will override the tools defined when creating the model.

I guess I'll document the behavior in the JavaDocs.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
glaforge: Yes, that's another approach which is safer.
But I'm waiting for the new Gemini SDK to be released (hopefully at the end of the week), and I'll be able to remove this `try`/`finally` block, and thus have a cleaner solution.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
glaforge: Currently, there's no way to enforce a tool to be called, but maybe it'll come later.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/FunctionCallHelper.java
glaforge: That's a good question!
If I understand correctly, a `Tool` is a set of functions, and we can pass one or more tools.
But if there are several functions, should I create a tool per function?
What's the logic with the organisations of functions as set of tools?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/FunctionCallHelperTest.java
glaforge: Great remark!
I went with only `String`, `Boolean` and `Number` because the underlying gRPC API only supports `Number`s.
There's no distinct type for `int`, `long` and `float`.
A number is just a `double`.
I'm not sure how to proceed there?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/FunctionCallHelper.java
langchain4j: In my understanding 1 function = 1 tool. Tool can also be something else than function (at least in OpenAI nomenclature).

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/FunctionCallHelperTest.java
langchain4j: But aren't `Integer`, `Long`, `Float`, etc. `Number`s?

Would be nice to define a tool with all common types (including float, double, int, long, etc) and see if deserialization of Gemini tool invocation works well and the tool can be called without errors

langchain4j: Maybe not in this test, probably better in integration test

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
langchain4j: Makes sense!

FYI, I plan to do a release tomorrow because there are some long-awaited features and bugs pending to be released.
And probably another one on 15 or 22.
So it is up to you if you want to release it tomorrow and then refactor (only internal changes) or wait for a proper solution

glaforge: Ok, thanks for the heads up.
I'll try to work on it today / tonight, and I'll update you to see how far I am.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/FunctionCallHelper.java
glaforge: I had a look at the data models of OpenAI and Gemini.

* In OpenAI's [doc](https://platform.openai.com/docs/assistants/tools/defining-functions), there's indeed one top-level `tools` array, which can contain _tools_ which can be function declarations.
* But in Gemini's [doc](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/function-calling#function-calling-samples), the `tools` array can contain tools which can be a `function_declarations` that can in turn contain an array of function declarations.

Currently, indeed, I create one tool that contains all the function declarations.
It seems that's what most code samples I've seen are doing.
But I don't know if it makes a difference or not if I created one tool per function.
I think I'll stick to this approach for now, for lack of a better understanding of the internals.

glaforge: Got confirmation that there is currently only one type of tool which is a set of functions.
There may be other tool types later on, but for now, it makes sense to just return one `Tool` that contains an array of `function_declarations`.

ghost (approved):
@glaforge great job, thank you!

PR #722, opened 2024-03-09
Ability to specify the task type for calculating embeddings

glaforge:
It's possible to customize different kinds of tasks for embedding:

- `RETRIEVAL_QUERY`
- `RETRIEVAL_DOCUMENT`
- `SEMANTIC_SIMILARITY`
- `CLASSIFICATION, CLUSTERING`

## Summary by CodeRabbit

- **New Features**
	- Introduced `TaskType` enum for specifying task types in embedding instances, enhancing customization for retrieval, similarity, classification, and clustering tasks.
	- Added `taskType` parameter to the constructor for setting task type in embedding instances.

- **Tests**
	- Added tests for semantic similarity, text classification, and document retrieval embeddings using different task types.

### DIFF
--- a/langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
+++ b/langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
@@ -134,11 +134,13 @@ public class VertexAiEmbeddingModel implements EmbeddingModel {
                 List<Value> instances = new ArrayList<>();
                 for (TextSegment segment : batch) {
                     VertexAiEmbeddingInstance embeddingInstance = new VertexAiEmbeddingInstance(segment.text());
-                    // Title metadata is used for calculating embeddings for document retrieval
-                    embeddingInstance.setTitle(segment.metadata(titleMetadataKey));
                     // Specify the type of embedding task when specified
                     if (this.taskType != null) {
                         embeddingInstance.setTaskType(taskType);
+                        if (this.taskType.equals(TaskType.RETRIEVAL_DOCUMENT)) {
+                            // Title metadata is used for calculating embeddings for document retrieval
+                            embeddingInstance.setTitle(segment.metadata(titleMetadataKey));
+                        }
                     }
 
                     Value.Builder instanceBuilder = Value.newBuilder();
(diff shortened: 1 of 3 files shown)


langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
```diff
@@ -122,8 +130,16 @@ public Response<List<Embedding>> embedAll(List<TextSegment> segments) {
 
                 List<Value> instances = new ArrayList<>();
                 for (TextSegment segment : batch) {
+                    VertexAiEmbeddingInstance embeddingInstance = new VertexAiEmbeddingInstance(segment.text());
+                    // Title metadata is used for calculating embeddings for document retrieval
+                    embeddingInstance.setTitle(segment.metadata("title"));
```
langchain4j: [Docu](https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-text-embeddings) says title is valid only for `RETRIEVAL_DOCUMENT`.
Would be nice to document this in hte javadoc of this class.
I would also give user an option to define the key of metadata to use instead of hardcoding "title".

ghost (dismissed):
@glaforge thanks!

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
glaforge: How would you let the user customize the `title` property?

langchain4j: Perhaps something like `titleMetadataKey("my_title")` in the builder/ctor, but this is not urgent, can be added later. My main concern was in that if there is a "title" metadata entry and task type is not `RETRIEVAL_DOCUMENT`, then embedding call might fail?

glaforge: Indeed that's a bit weird, but that's what's happening: using `title` without `RETRIEVAL_DOCUMENT` would fail.
I'll add the suggested method to customize the metadata key.

glaforge: Pushed the extra builder method.

langchain4j: Maybe worth adding a condition here? Set `title` only for `RETRIEVAL_DOCUMENT` ?

glaforge: Good point: pushed a change to set `title` only for `RETRIEVAL_DOCUMENT`.
That way, it's still possible to use `title` when not using that embedding task type.

ghost (approved):
@glaforge thanks a lot!

PR #726, opened 2024-03-10
Allow HtmlTextExtractor to extract relative links from document

Wadimz:
By default `org.jsoup.parser` uses `""` base url while parsing HTML document. This results in empty links from HtmlTextExtractor if relative link is used in raw html, 
```    
public static Document parse(String html) {
        return Parser.parse(html, "");
}
```

ex. extracted text with links from  
`<a href="/menu1">here</a>` results in `here <>` 
but 
`<a href="http://example.org/menu1">here</a>` results in `here <http://example.org/menu1>` 

changes utilize document metadata to provide base url.

## Summary by CodeRabbit

- **New Features**
	- Enhanced text extraction from HTML content to include better link handling and metadata preservation.
- **Tests**
	- Added tests for the improved HTML text extraction process.


geoand (approved):

langchain4j:
@Wadimz thank you!

PR #737, opened 2024-03-12
extend MilvusEmbeddingStore:

igerdy:
extend MilvusEmbeddingStore:
 - supports inserting by partition
 - supports search by partition

### DIFF
--- a/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStore.java
+++ b/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStore.java
@@ -29,6 +29,7 @@ import io.milvus.param.MetricType;
 import io.milvus.param.dml.InsertParam;
 import io.milvus.param.dml.SearchParam;
 import io.milvus.response.SearchResultsWrapper;
+import org.apache.commons.lang3.StringUtils;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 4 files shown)


igerdy:
@langchain4j Please process this request

langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStore.java
```diff
@@ -116,21 +123,54 @@ public String add(Embedding embedding, TextSegment textSegment) {
...
   }
 
   public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedded) {
     List<String> ids = generateRandomIds(embeddings.size());
-    addAllInternal(ids, embeddings, embedded);
+    addAllInternal(ids, embeddings, embedded, null);
+    return ids;
+  }
+
+public String add(Embedding embedding, String partitionName) {
```
langchain4j, line 136: I am not sure duplicating each method in order to specify partition is the right way to go.
What about setting partition in the constructor/builder?

langchain4j (commented):
@igerdy thank you!

langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStore.java
igerdy: After obtaining a connection, different partitions may be operated during multiple operations. At the same time, it is also convenient to expand into a connection pool in the future to maintain connections.

langchain4j: @igerdy this approach is definitely not scalable. I propose to propagate partition via `EmbeddingStoreRequest` instead. We could add a `Map<String, Object>` field that will hold store-specific properties, then `MilvusEmbeddingStore` could read partition from it.

WDYT?

igerdy: Your idea is to add a class in the `core` module to cover the features of different `EmbeddingStore`?
Good suggestion, this will make the extensibility stronger.
But I think the possibility of multiple `EmbeddingStore` in a project is relatively low. Can it be extended by extending the interface and implementation in the specific EmbeddingStore module? What do you think?

Issue #698, opened 2024-03-05
[FEATURE] extend response for add some extension fields for llm invoke


Willam2004:
I'm always frustrated when i  invoke the llm service named dashscope. because it has returned GenerationResult which have a field names **requestId** for troubleshoot performance issues.  but now in order to return union result named **Response** that discard the **requestId**  info. 

```
  @Override
    public Response<String> generate(String prompt) {
        try {
            QwenParam param = QwenParam.builder()
                    .apiKey(apiKey)
                    .model(modelName)
                    .topP(topP)
                    .topK(topK)
                    .enableSearch(enableSearch)
                    .seed(seed)
                    .prompt(prompt)
                    .resultFormat(MESSAGE)
                    .build();

            GenerationResult generationResult = generation.call(param);

            return Response.from(answerFrom(generationResult),
                    tokenUsageFrom(generationResult), finishReasonFrom(generationResult));
        } catch (NoApiKeyException | InputRequiredException e) {
            throw new RuntimeException(e);
        }
    }

// Copyright (c) Alibaba, Inc. and its affiliates.
package com.alibaba.dashscope.aigc.generation;

import com.alibaba.dashscope.common.DashScopeResult;
import com.alibaba.dashscope.utils.JsonUtils;
import com.google.gson.JsonObject;
import lombok.Data;
import lombok.extern.slf4j.Slf4j;

@Slf4j
@Data
public final class GenerationResult {
  private String requestId;
  private GenerationUsage usage;
  private GenerationOutput output;

  private GenerationResult() {}

  public static GenerationResult fromDashScopeResult(DashScopeResult dashScopeResult) {
    GenerationResult result = new GenerationResult();
    result.setRequestId(dashScopeResult.getRequestId());
    if (dashScopeResult.getUsage() != null) {
      result.setUsage(
          JsonUtils.fromJsonObject(
              dashScopeResult.getUsage().getAsJsonObject(), GenerationUsage.class));
    }
    if (dashScopeResult.getOutput() != null) {
      result.setOutput(
          JsonUtils.fromJsonObject(
              (JsonObject) dashScopeResult.getOutput(), GenerationOutput.class));
    } else {
      log.error(String.format("Result no output: %s", dashScopeResult));
    }
    return result;
  }
}


 
```
i try use the aop  for logging the generation.call(param);, but the com.alibaba.dashscope.aigc.generation.Generation is final class !
so did extend response for add some extension fields for store the information , ex: requestId?

[...]

PR #742, opened 2024-03-13
Add a metadata field to the Response

jiangsier-xyz:
Add a metadata field to the Response that can be populated by model providers. (#698)

DashScope callings fill the requestId as the response metadata.

## Summary by CodeRabbit

- **New Features**
	- Enhanced response objects to include metadata, improving traceability and context of responses.
	- Introduced metadata handling in various models, enabling enriched data processing and response generation.
- **Refactor**
	- Updated methods across models to support metadata inclusion, ensuring consistency and reliability in data handling.
- **Tests**
	- Added comprehensive tests to validate the presence and correctness of metadata in responses, ensuring robustness and accuracy.

### DIFF
--- a/docs/docs/intro.md
+++ b/docs/docs/intro.md
@@ -17,8 +17,8 @@ Here's how:
    LangChain4j currently supports [15+ popular LLM providers](/integrations/language-models/)
    and [15+ embedding stores](/integrations/embedding-stores/).
 2. **Comprehensive Toolbox:**
-   During the past year, the community has been building numerous LLM-powered applications,
-   identifying common abstractions, patterns, and techniques. LangChain4j has refined these into practical code.
+   Over the past year, the community has been building numerous LLM-powered applications,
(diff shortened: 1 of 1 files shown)


jiangsier-xyz:
@langchain4j please take a look. Then I can continue to promote the implementation of features such as function-call on dashscope.

langchain4j:
@jiangsier-xyz sorry for the late reply. I foresee that the API will be changed in the near future, so `ChatLanguageModel` will not return `Result<T>` any more. We can still merge this PR if you think it is required (at least as a temporary solution).

jiangsier-xyz:
> @jiangsier-xyz sorry for the late reply. I foresee that the API will be changed in the near future, so `ChatLanguageModel` will not return `Result<T>` any more. We can still merge this PR if you think it is required (at least as a temporary solution).

It is an attempt to enhance the observability of the model. Acquiring the request_id field indeed aids consumers in obtaining technical feedback from the dashscope model provider. However, if the new design is slated for a merge in the upcoming few releases and pays attention to trace, there is no need to merge this PR.

PR #766, opened 2024-03-16
[DOCS] Update overview integration table with new integrations

czelabueno:
This PR remove `async` columns of the integration table and add `Anthropic` and `Zhipu AI` as new integrations.
Also, this PR add a new column named `Function calling` to display the capabilities already supported.

## Summary by CodeRabbit

- **Documentation**
	- Updated the supported LLM integrations documentation, including new providers Anthropic and Zhipu AI.
	- Renamed certain capabilities and added new ones like `Scoring` and `Function Calling` to reflect the latest offerings.


LizeRaes (approved):
Nice, thanks a lot!

ghost (approved):
@czelabueno thans a lot!

BTW is there a reson why `index` file has `mdx` extension, and not `md`?
It turned out that my IDEA can disply tables nicely only in md files.
I guess I can re-configure this, but if there is no need for `mdx`, I would rename it to `md` to avoid the same confusion for other contributors.

PR #779, opened 2024-03-20
Adds other embeddings dependencies to BOM

agoncal:
Only `langchain4j-embeddings-all-minilm-l6-v2-q` is defined in the BOM. Not sure you want to have all the other embeddings defined in the BOM. If yes, this PR adds the missing dependencies to the BOM. If not, you can discard this PR.


ghost (approved):
@agoncal good catch, thank you!

PR #788, opened 2024-03-21
[FEATURE]: add Hugging Face text embeddings inference (TEI) integration

CharleyXu:
## Context
The purpose is to be more flexible when using embedding and rerank models without taking up too much performance. Text Embeddings Inference (TEI) is a comprehensive toolkit designed for efficient deployment and serving of open source text embeddings models. It supports embeddings models, re-rankers and sequence classification models. This is now achieved by integrating TEI

## Change
add text embeddings inference(tei) integration

## Checklist
- I have added unit and integration tests for my change
- All unit and integration tests in the module I have added/changed are green
- I have added some documentation in test classes
- I have added tei module in the [BOM]


CharleyXu:
@langchain4j, Please take the time to code review, thanks！

langchain4j:
@CharleyXu thank you! ANd sorry for the late reaction!

How is it different from the existing HF module?

Martin7-1:
Looks like Text Embedding Inference is a way to deploy custom embedding / text classification models. see [text-embedding-inference](https://github.com/huggingface/text-embeddings-inference).

Existing HF module uses official inference API for chat and embedding. But TEI can deploy locally and only for embedding / text classificiation. That's the only difference.

BTW, TEI supports OpenAI API format, so I think it has no need to integration it in a new module.

dliubarskyi:
@Martin7-1 thank you for clarification!

Since OpenAI API is there, I would not create this new integration.

@CharleyXu thank you for your efforts though!

PR #800, opened 2024-03-21
[FEATURE]: Weaviate: Store Metadata + Support gRPC

ehsavoie:
When storing embeddings into Weaviate we are loosing the content metadata. The goal of this PR is to store those metadata as fields in weaviate so that we can get them back when we are finding the content.

## Context
When doing RAG with weaviate as vector database for embeddings we are loosing the metadata associated with the textSegment
## Change
I now store and get back the metadata in Weaviate.
Adding also support for gRPC as according to weaviate team it has better performance.

## Checklist


langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -65,7 +75,8 @@ public WeaviateEmbeddingStore(
             Integer port,
             String objectClass,
             Boolean avoidDups,
-            String consistencyLevel
+            String consistencyLevel,
+            List<Property> properties
```
langchain4j:
```suggestion
            Collection<String> metadataKeys
```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -30,32 +32,40 @@
 
 /**
  * Represents the <a href="https://weaviate.io/">Weaviate</a> vector database.
- * Current implementation assumes the cosine distance metric is used.
- * Does not support storing {@link dev.langchain4j.data.document.Metadata} yet.
+ * Current implementation assumes the cosine distance metric is used. Does not
```
dliubars: Please remove "Does not support metadata"

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -30,32 +32,40 @@
...
     private final WeaviateClient client;
     private final String objectClass;
     private final boolean avoidDups;
     private final String consistencyLevel;
+    private final List<Property> properties;
 
     /**
      * Creates a new WeaviateEmbeddingStore instance.
      *
-     * @param apiKey           Your Weaviate API key. Not required for local deployment.
```
dliubars, line 49: please revert previous formatting back

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -100,9 +112,12 @@ public String add(Embedding embedding) {
     /**
      * Adds a new embedding with provided ID to the store.
      *
-     * @param id        the ID of the embedding to add in UUID format, since it's Weaviate requirement.
```
dliubars: please revert previous formatting back

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -126,31 +141,38 @@ public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedde
     }
 
     /**
-     * {@inheritDoc}
```
dliubars: please revert previous formatting back

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -213,10 +235,33 @@ private List<String> addAll(List<String> ids, List<Embedding> embeddings, List<T
...
+                for (Property property : properties) {
+                    metadata.put(property.getName(), "");
+                }
+                props.put(METADATA, metadata);
+            }
+            props.put(METADATA, metadata);
+        } else {
+            props.put(METADATA_TEXT_SEGMENT, "");
+            for (Property property : properties) {
+                metadata.put(property.getName(), "");
```
dliubars: this logic seems strange, is it actually required?

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -10,6 +12,14 @@
...
+import dev.langchain4j.store.embedding.EmbeddingSearchRequest;
+import io.weaviate.client.v1.schema.model.Property;
+import java.util.Arrays;
+import java.util.List;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.data.Percentage.withPercentage;
+import org.junit.jupiter.api.Test;
 
 @Testcontainers
 class LocalWeaviateEmbeddingStoreIT extends EmbeddingStoreWithoutMetadataIT {
```
dliubars:
```suggestion
class LocalWeaviateEmbeddingStoreIT extends EmbeddingStoreIT {
```

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -43,4 +54,34 @@ protected EmbeddingModel embeddingModel() {
     protected void ensureStoreIsEmpty() {
         // TODO fix
     }
-}
\ No newline at end of file
+
+    @Test
+    void should_add_embedding_with_segment_and_metadata() {
```
dliubars: please remove this test, it is already present in `EmbeddingStoreIT`

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -228,6 +273,15 @@ private WeaviateObject buildObject(String id, Embedding embedding, String text)
 
     private static EmbeddingMatch<TextSegment> toEmbeddingMatch(Map<String, ?> item) {
         Map<String, ?> additional = (Map<String, ?>) item.get(ADDITIONALS);
+        final Metadata metadata = new Metadata();
+        if (item.get(METADATA) != null && item.get(METADATA) instanceof Map) {
+            Map<String, ?> resultingMetadata = (Map<String, ?>) item.get(METADATA);
+            for(Map.Entry<String, ?> entry : resultingMetadata.entrySet()) {
+                if(entry.getValue() != null && !"".equals(entry.getValue())) {
```
dliubars: how to know if there was no metadata entry or there was, but with an empty string value?

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
dliubars:
```suggestion
            Collection<String> metadataKeys
```
To keep it consistent with other integrations (e.g. redis).

Also, is it absolutely necessary? Is there a way to store/retrieve metadata without specifying explicitly each key?

dliubars (commented):
@ehsavoie thank you so much!

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
ehsavoie: No I should be using "null" instead

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
ehsavoie: Yes, i used "null" instead which seems to be the correct value for null

ehsavoie:
I moved from batch to data to create the objects. i don't understand why but the batch operation didn't create them while using the data API worked. I don't really understand why, all the more so as batch seemed to be the correct way to do it.

ehsavoie:
@dliubars I think I fixed all your comments.

cescoffier:
@dliubars can you have a look, please?

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreIT.java
langchain4j, line 1: `CloudWeaviateEmbeddingStoreIT` works on main branch but fails on this branch with this exception.
I suspect it is because I have created a new cluster. So all new users will face this issue.
```
java.lang.IllegalArgumentException: no graphql provider present, this is most likely because no schema is present. Import a schema first!

	at dev.langchain4j.store.embedding.weaviate.WeaviateEmbeddingStore.findRelevant(WeaviateEmbeddingStore.java:193)
	at dev.langchain4j.store.embedding.EmbeddingStore.findRelevant(EmbeddingStore.java:88)
	at dev.langchain4j.store.embedding.EmbeddingStoreWithoutMetadataIT.should_add_embedding_with_id(EmbeddingStoreWithoutMetadataIT.java:72)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

After running the same test from main branch, it starts to fail on this branch with another exception:
```
java.lang.IllegalArgumentException: Cannot query field "Testc9adc8fdf29040db9c644c68d2b62678" on type "GetObjectsObj".

	at dev.langchain4j.store.embedding.weaviate.WeaviateEmbeddingStore.findRelevant(WeaviateEmbeddingStore.java:199)
	at dev.langchain4j.store.embedding.EmbeddingStore.findRelevant(EmbeddingStore.java:88)
	at dev.langchain4j.store.embedding.EmbeddingStoreWithoutMetadataIT.should_add_embedding_with_id(EmbeddingStoreWithoutMetadataIT.java:72)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -1,30 +1,60 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import org.testcontainers.junit.jupiter.Container;
 import org.testcontainers.weaviate.WeaviateContainer;
 
-import static dev.langchain4j.internal.Utils.randomUUID;
-
 @Testcontainers
-class LocalWeaviateEmbeddingStoreIT extends EmbeddingStoreWithoutMetadataIT {
+class LocalWeaviateEmbeddingStoreIT extends EmbeddingStoreIT {
```
langchain4j, line 15: please update `EmbeddingStoreWithoutMetadataIT` to extend from `EmbeddingStoreIT` as well

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,45 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
      *                         provided text segment, which avoids duplicated entries in DB.
      *                         If false, then random ID will be generated.
      * @param consistencyLevel Consistency level: ONE, QUORUM (default) or ALL. Find more details <a href="https://weaviate.io/developers/weaviate/concepts/replication-architecture/consistency#tunable-write-consistency">here</a>.
+     * @param properties       List of properties of the object
+     * @param useGrpc          Use GRPC instead of HTTP
```
langchain4j: I can't find anything about GRPC in this PRs description, please add a note there

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,45 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
...
+     * @param grpcPort         The port, e.g. 50051. This parameter is optional.
      */
     @Builder
     public WeaviateEmbeddingStore(
             String apiKey,
             String scheme,
             String host,
             Integer port,
+            boolean useGrpc,
```
langchain4j:
```suggestion
            Boolean useGrpc,
```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,45 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
...
+            Collection<Property> properties
     ) {
         try {
 
             Config config = new Config(
                     ensureNotBlank(scheme, "scheme"),
                     concatenate(ensureNotBlank(host, "host"), port)
             );
-            this.client = WeaviateAuthClient.apiKey(config, getOrDefault(apiKey, ""));
+            if (useGrpc) {
```
langchain4j:
```suggestion
            if (getOrDefault(useGrpc, false)) {
```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,45 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
...
+                config.setGRPCSecured(false);
```
langchain4j: could you please explain why is this set to false?

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,45 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
...
+                config.setGRPCHost(host + ":" + getOrDefault(grpcPort, 50051));
+            }
+            if(getOrDefault(apiKey, "").isEmpty()) {
```
langchain4j:
```suggestion
            if (isNullOrBlank(apiKey)) {
```

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -1,30 +1,60 @@
...
+            .withEnv("AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED", "true")
             .withEnv("CLUSTER_HOSTNAME", "node1");
 
-    EmbeddingStore<TextSegment> embeddingStore = WeaviateEmbeddingStore.builder()
+    private final EmbeddingStore<TextSegment> embeddingStore = WeaviateEmbeddingStore.builder()
             .scheme("http")
             .host(weaviate.getHost())
-            .port(weaviate.getFirstMappedPort())
+            .port(weaviate.getMappedPort(8080))
+            .useGrpc(true)
```
langchain4j: could you please create 2 separate tests (1 for cloud and 1 for local) to test GRPC variant, instead of changing existing ones?

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,45 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
...
             String scheme,
             String host,
             Integer port,
+            boolean useGrpc,
+            Integer grpcPort,
             String objectClass,
             Boolean avoidDups,
-            String consistencyLevel
+            String consistencyLevel,
+            Collection<Property> properties
```
langchain4j:
```suggestion
            Collection<String> metadataKeys
```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -135,22 +155,29 @@ public List<EmbeddingMatch<TextSegment>> findRelevant(
...
+        fields.add(Field
+                .builder()
+                .name(ADDITIONALS)
+                .fields(
+                        Field.builder().name("id").build(),
+                        Field.builder().name("certainty").build(),
+                        Field.builder().name("vector").build()
+                )
+                .build());
+        if (properties != null && !properties.isEmpty()) {
```
langchain4j: minor: `properties` is never null

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -135,22 +155,29 @@ public List<EmbeddingMatch<TextSegment>> findRelevant(
...
+            List<Field> metadataFields = new ArrayList<>();
+            for (Property property : properties) {
+                metadataFields.add(Field.builder().name(property.getName()).build());
+            }
+            fields.add(Field.builder().name(METADATA).fields(metadataFields.toArray(new Field[0])).build());
```
langchain4j, line 176: Just in case: with this configuration (storing all metadata entries nested under `METADATA` field), will it be possible later to do [metadata filtering](https://github.com/langchain4j/langchain4j/pull/610) on them?

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -200,34 +227,62 @@ private List<String> addAll(List<String> ids, List<Embedding> embeddings, List<T
                     ? ids.get(i)
                     : avoidDups && embedded != null ? generateUUIDFrom(embedded.get(i).text()) : randomUUID();
             resIds.add(id);
-            objects.add(buildObject(id, embeddings.get(i), embedded != null ? embedded.get(i).text() : ""));
+            objects.add(buildObject(id, embeddings.get(i), embedded != null ? embedded.get(i) : null));
```
langchain4j, line 232: Existing users might have records with empty string in their weaviate databases, right?
Did you test that new code works with databases created with 0.29.1?

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -43,4 +73,4 @@ protected EmbeddingModel embeddingModel() {
     protected void ensureStoreIsEmpty() {
         // TODO fix
     }
-}
\ No newline at end of file
+}
```
langchain4j, line 71: `should_add_embedding_with_segment_with_metadata` fails with:

```
java.lang.AssertionError: 
Expecting actual:
  0.9465930461883545
to be close to:
  1.0

```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -200,34 +227,62 @@ private List<String> addAll(List<String> ids, List<Embedding> embeddings, List<T
...
-
+        Map<String, Object> metadata = new HashMap<>();
+        if (segment != null) {
+            props.put(METADATA_TEXT_SEGMENT, segment.text());
+            if (segment.metadata() != null && !segment.metadata().asMap().isEmpty()) {
+                for (Property property : properties) {
+                    if (segment.metadata().containsKey(property.getName())) {
+                        metadata.put(property.getName(), segment.metadata().get(property.getName()));
+                    } else {
+                        metadata.put(property.getName(), "null");
```
langchain4j: I doubt it is a good idea to store `"null"` string when there is no property. Is it possible to avoid adding this metadata entry or set it to `null`?

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -1,30 +1,60 @@
...
-    EmbeddingStore<TextSegment> embeddingStore = WeaviateEmbeddingStore.builder()
+    private final EmbeddingStore<TextSegment> embeddingStore = WeaviateEmbeddingStore.builder()
             .scheme("http")
             .host(weaviate.getHost())
-            .port(weaviate.getFirstMappedPort())
+            .port(weaviate.getMappedPort(8080))
+            .useGrpc(true)
+            .grpcPort(weaviate.getMappedPort(50051))
             .objectClass("Test" + randomUUID().replace("-", ""))
+            .properties(Arrays.asList(new Property[]{
```
langchain4j: It seems that dataType information is not used enywhere, can we get rid of it and simplify to this?

```java
.metadataKeys(List.of("string_empty", "string_space", ...))
```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
ehsavoie: the "null" value string is used in the json conversion, otherwise the field is not sent and then the object doesn't respect the schema when trying to obtain it.

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
ehsavoie: I'm not sure as I'm not really an expert on this but I don't see why not, you would put ["_metadata", "key"] as the path instead of key in the where filter (as far as I understand the examples on weaviate web site)

ehsavoie: Something like: 
WhereArgument whereUuid = WhereArgument.builder()
      .filter(WhereFilter.builder()
        .path(new String[]{"_metadata", "title"})
        .operator(Operator.Equal)
        .valueText("Man on the Moon")
        .build();

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreIT.java
ehsavoie: How may I reproduce your test ?

langchain4j: Just run `CloudWeaviateEmbeddingStoreIT` ? Or is it passing for you?

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
ehsavoie: Yes, it will work just the same as long as you don't want to get metadata that weren't stored as the fields don't exist

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreIT.java
ehsavoie: Passing now

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
ehsavoie: Well I used the sample code from Weaviate Client test code and i didn't have mean to set up a secured connection for GRPC. 
I've updated to make this configureable

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
ehsavoie: I couldn't figure out what is the GRPC configuration for the cloud testing. Thus I only added a local one

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,47 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
      *                         provided text segment, which avoids duplicated entries in DB.
      *                         If false, then random ID will be generated.
      * @param consistencyLevel Consistency level: ONE, QUORUM (default) or ALL. Find more details <a href="https://weaviate.io/developers/weaviate/concepts/replication-architecture/consistency#tunable-write-consistency">here</a>.
+     * @param metadataKeys     List of metadataKeys of the object
```
langchain4j:
```suggestion
     * @param metadataKeys     Metadata keys that should be persisted (optional)
```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,47 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
...
      */
     @Builder
     public WeaviateEmbeddingStore(
             String apiKey,
             String scheme,
             String host,
             Integer port,
+            Boolean useGrpc,
+            Boolean securedGrpc,
+            Integer grpcPort,
```
langchain4j, line 77: can we (re)use existing `port` parameter for grpc?

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,47 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
...
             Config config = new Config(
                     ensureNotBlank(scheme, "scheme"),
                     concatenate(ensureNotBlank(host, "host"), port)
             );
-            this.client = WeaviateAuthClient.apiKey(config, getOrDefault(apiKey, ""));
+            if (getOrDefault(useGrpc, Boolean.FALSE)) {
+                config.setGRPCSecured(getOrDefault(securedGrpc, Boolean.FALSE));
+                config.setGRPCHost(host + ":" + getOrDefault(grpcPort, 50051));
+            }
+            if(getOrDefault(apiKey, "").isEmpty()) {
```
langchain4j:
```suggestion
            if (isNullOrBlank(apiKey)) {
```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -56,30 +61,47 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
...
+                this.client = new WeaviateClient(config);
+            } else {
+                this.client = WeaviateAuthClient.apiKey(config, getOrDefault(apiKey, ""));
```
langchain4j:
```suggestion
                this.client = WeaviateAuthClient.apiKey(config, apiKey);
```

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -200,34 +229,62 @@ private List<String> addAll(List<String> ids, List<Embedding> embeddings, List<T
...
 
-    private WeaviateObject buildObject(String id, Embedding embedding, String text) {
+    private WeaviateObject buildObject(String id, Embedding embedding, TextSegment segment) {
         Map<String, Object> props = new HashMap<>();
-        props.put(METADATA_TEXT_SEGMENT, text);
-
+        Map<String, Object> metadata = new HashMap<>();
+        if (segment != null) {
+            props.put(METADATA_TEXT_SEGMENT, segment.text());
+            if (segment.metadata() != null && !segment.metadata().asMap().isEmpty()) {
```
langchain4j:
```suggestion
            if (!segment.metadata().asMap().isEmpty()) {
```
metadata is always non-null

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -22,26 +23,30 @@
...
  * Represents the <a href="https://weaviate.io/">Weaviate</a> vector database.
  * Current implementation assumes the cosine distance metric is used.
- * Does not support storing {@link dev.langchain4j.data.document.Metadata} yet.
  */
 public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
 
     private static final String METADATA_TEXT_SEGMENT = "text";
     private static final String ADDITIONALS = "_additional";
+    private static final String METADATA = "_metadata";
+    private static final String NULL_VALUE = "null";
```
langchain4j, line 43:
```suggestion
    private static final String NULL_VALUE = "<null>";
```
To make it clearer that this is a `null` value and not a "null" string

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -200,34 +229,62 @@ private List<String> addAll(List<String> ids, List<Embedding> embeddings, List<T
...
+            if (segment.metadata() != null && !segment.metadata().asMap().isEmpty()) {
+                for (String property : metadataKeys) {
+                    if (segment.metadata().containsKey(property)) {
+                        metadata.put(property, segment.metadata().get(property));
+                    } else {
+                        metadata.put(property, NULL_VALUE);
+                    }
+                }
+            } else {
+                for (String property : metadataKeys) {
```
langchain4j: there is code duplication in this method (e.g. setting `metadata.put(property, NULL_VALUE);` in the loop in 3 places), would be nice to redice it

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -1,30 +1,58 @@
...
-
 @Testcontainers
-class LocalWeaviateEmbeddingStoreIT extends EmbeddingStoreWithoutMetadataIT {
+class LocalWeaviateEmbeddingStoreIT extends EmbeddingStoreIT {
 
     @Container
     static WeaviateContainer weaviate = new WeaviateContainer("semitechnologies/weaviate:latest")
             .withEnv("QUERY_DEFAULTS_LIMIT", "25")
             .withEnv("DEFAULT_VECTORIZER_MODULE", "none")
+            .withEnv("AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED", "true")
```
langchain4j:
```suggestion
```
rolling back unnecessary change

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -1,30 +1,58 @@
...
             .withEnv("CLUSTER_HOSTNAME", "node1");
 
-    EmbeddingStore<TextSegment> embeddingStore = WeaviateEmbeddingStore.builder()
+    private final EmbeddingStore<TextSegment> embeddingStore = WeaviateEmbeddingStore.builder()
             .scheme("http")
             .host(weaviate.getHost())
-            .port(weaviate.getFirstMappedPort())
+            .port(weaviate.getMappedPort(8080))
```
langchain4j:
```suggestion
            .port(weaviate.getFirstMappedPort())
```
rolling back unnecessary change

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -1,30 +1,58 @@
...
+            .useGrpc(false)
```
langchain4j:
```suggestion
```
rolling back unnecessary change

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalWeaviateEmbeddingStoreIT.java
```diff
@@ -1,30 +1,58 @@
...
+            .grpcPort(weaviate.getMappedPort(50051))
```
langchain4j:
```suggestion
```
rolling back unnecessary change

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalGRPCWeaviateEmbeddingStoreIT.java
```diff
@@ -0,0 +1,74 @@
...
+import org.testcontainers.weaviate.WeaviateContainer;
+
+@Testcontainers
+class LocalGRPCWeaviateEmbeddingStoreIT extends EmbeddingStoreIT {
+
+    @Container
+    static WeaviateContainer weaviate = new WeaviateContainer("semitechnologies/weaviate:latest")
+            .withEnv("QUERY_DEFAULTS_LIMIT", "25")
+            .withEnv("DEFAULT_VECTORIZER_MODULE", "none")
+            .withEnv("AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED", "true")
```
langchain4j: is `AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED` required? seems to work fine without it

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalGRPCWeaviateEmbeddingStoreIT.java
```diff
@@ -0,0 +1,74 @@
...
+            .withEnv("CLUSTER_HOSTNAME", "node1");
+
+    private final EmbeddingStore<TextSegment> embeddingStore = WeaviateEmbeddingStore.builder()
+            .scheme("http")
+            .host(weaviate.getHost())
+            .port(weaviate.getMappedPort(8080))
+            .useGrpc(true)
```
langchain4j: It seems that this test still uses HTTP under the hood.
When looking into sources of `WeaviateClient`, it seems that only `batch()` method is aware of grpc, `graphQL()` (which we use) seems to only use HTTP.
Could you please double check if this is the case?

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalGRPCWeaviateEmbeddingStoreIT.java
```diff
@@ -0,0 +1,74 @@
...
+    static WeaviateContainer weaviate = new WeaviateContainer("semitechnologies/weaviate:latest")
+            .withEnv("QUERY_DEFAULTS_LIMIT", "25")
+            .withEnv("DEFAULT_VECTORIZER_MODULE", "none")
+            .withEnv("AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED", "true")
+            .withEnv("CLUSTER_HOSTNAME", "node1");
+
+    private final EmbeddingStore<TextSegment> embeddingStore = WeaviateEmbeddingStore.builder()
+            .scheme("http")
+            .host(weaviate.getHost())
+            .port(weaviate.getMappedPort(8080))
```
langchain4j, line 26: why is setting http port required for grpc test?

ghost (changes requested):
@ehsavoie thanks a lot! Please check my comments. Thank you

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalGRPCWeaviateEmbeddingStoreIT.java
ehsavoie: Yes you are correct, only batch uses gRPC reading the code from weaviate client

ehsavoie: Confirmed by Weaviate support: only batch inserts use gRPC

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalGRPCWeaviateEmbeddingStoreIT.java
ehsavoie: because weddon't do only inserts

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
ehsavoie: Given that we still need HTTP with gRPC I'll keep both.

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalGRPCWeaviateEmbeddingStoreIT.java
langchain4j: Do you know if there is a plan to extend grpc usage for querying as well?
Otherwise I would suggest renaming `useGrpc` into `useGrpcForInserts` or something like that.
Now it is confusing

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalGRPCWeaviateEmbeddingStoreIT.java
langchain4j: Got it. Could you please document in javadoc of `WeaviateEmbeddingStore` where http and where grcp is used?
It is confusing right now

langchain4j-weaviate/src/test/java/dev/langchain4j/store/embedding/weaviate/LocalGRPCWeaviateEmbeddingStoreIT.java
ehsavoie: No I don't know about that but it would mean renaming if they change that behaviour. I'd rather stick with weaviate own naming as weaviate doesn't specify that in its configuration

langchain4j: @ehsavoie if they do a bad job with their API, it does not mean we should do the same :) Right now it is very confusing.
If they will add an option to use gRPC for reads, we could either add `useGrpcForReads` or `useGrpc` which will apply to both. Do you see a problem with this approach? Thanks

ehsavoie: No I have no problem with that :)
Iwas just wondering if the host should be configurable as i noticed that on thir cloud it has a different host name than the HTTP one

ghost (approved):
@ehsavoie thanks a lot!

PR #803, opened 2024-03-22
fix: AiService is unable to correctly parse JSON responses

luguoj:
## Context
AiService is unable to correctly parse JSON responses.

## Change
Modify the implementation of the ServiceOutputParser.parse function. When parsing JSON responses, trim out the return type text information in the prefix

## Checklist
- [x] All unit and integration tests in the [core]


geoand:
I am fairly certain this has since already fixed in `main`

langchain4j:
@luguoj thank you!

But @geoand is right, this has been fixed already.

PR #817, opened 2024-03-23
Support relative paths without parent directory

### LINKED ISSUE #816: [BUG] FileSystemDocumentLoader cannot handle relative paths without parent directory
**Describe the bug**
When using `FileSystemDocumentLoader` with a relative path that consits only with a file name, a runtime exception is thrown. However, it should not be, because the file exists.

**Log and Stack trace**
```
java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.toAbsolutePath()" because the return value of "java.nio.file.Path.getParent()" is null
```

**To Reproduce**
A minimal reprodusible example:
```java
package org.example;

import dev.langchain4j.data.document.Document;
import dev.langchain4j.data.document.loader.FileSystemDocumentLoader;
import dev.langchain4j.data.document.parser.TextDocumentParser;

public class TestIssue {
        Document document = FileSystemDocumentLoader.loadDocument("text.txt", new TextDocumentParser());
        System.out.println(document.text());
    }
}
```

In this example I've also added a `text.txt` file in the working directory.

**Expected behavior**
The exception should not be thrown. The program should read the file and continue normal work.

- LangChain4j version: 0.28.0
- LLM(s) used: -
- Java version: 21
- Spring Boot version (if applicable): -

InAnYan:
## Context
Closes #816.

## Change
I've just reordered methods: `getParent` and `getAbsolutePath`. Quite a simple change.

Still, it may not work if the file is in the root (or may be work, haven't tested).

I had some problems with opening the project in InteliJ IDEA and I need some help, do I need to add some tests? I would like, but how to mock the filesystem?

## Checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/data/document/source/FileSystemSource.java
+++ b/langchain4j/src/main/java/dev/langchain4j/data/document/source/FileSystemSource.java
@@ -32,7 +32,7 @@ public class FileSystemSource implements DocumentSource {
     public Metadata metadata() {
         return new Metadata()
                 .add(FILE_NAME, path.getFileName().toString())
-                .add(ABSOLUTE_DIRECTORY_PATH, path.getParent().toAbsolutePath().toString());
+                .add(ABSOLUTE_DIRECTORY_PATH, path.toAbsolutePath().getParent().toString());
     }
 
     public static FileSystemSource from(Path filePath) {


ghost (approved):
@InAnYan thank you!

langchain4j:
@InAnYan maybe [this](https://www.baeldung.com/junit-5-temporary-directory) can help?

What problems did you have with IDE?

PR #819, opened 2024-03-24
Allow using AI Service with tools without memory

ghost:
## Context
Currenty AI Service can use tools only if there is a `ChatMemory` or `ChatMemoryProvider` configured.
This is probably an unnecessary and overly strict requirement. In practice, tools can be used without the memory that spans between calls to AI service, using only a "temporary memory" that spans single call to the AI service (that can include multiple interactions with LLM due to tool executions).

## Change
Allow using AI Services with tools without memory. If there is no memory configured, a "temporary memory" is used for each AI Service invocation to store all messages created/produced in scope of this AI Service invocation.

## Checklist


langchain4j:
cc @geoand

geoand:
👌

PR #821, opened 2024-03-25
Optimize imports using Intellij IDEA

jdubois:
Imports in the project are not cleaned up, and are not consistant: this is an update just using the "optimise imports" action from IntelliJ IDEA.

If that's fine with you @langchain4j , I can do another PR adding https://github.com/jhipster/prettier-java which is the official Prettier version for Java, maintained by the JHipster project.


jdubois:
The build is failing, but it seems to be because the `main` branch is currently broken. I'll wait until this is fixed and updated this branch.

langchain4j:
Pardon for the broken main, I have fixed it, but I can't resolve the conflict in this PR

PR #823, opened 2024-03-25
Optimize imports using Intellij IDEA

jdubois:
This is a new version of #821

Imports in the project are not cleaned up, and are not consistant: this is an update just using the "optimise imports" action from IntelliJ IDEA.

If that's fine with you @langchain4j , I can do another PR adding https://github.com/jhipster/prettier-java which is the official Prettier version for Java, maintained by the JHipster project.


jdubois:
Closing it as it's annoying to maintain, but super easy to recreate from scratch

langchain4j:
Yes, sorry, I will do it once I am done with more urgent stuff. I guess it should be also done in sync with prettier?

jdubois:
Prettier will do much more, here it's only the imports: that's simple and not risky

PR #744, opened 2024-03-13
Introduce SPI for creating Mistral AI Client

geoand:
This is very similar to what is done for OpenAI and it will directly enable [this](https://github.com/quarkiverse/quarkus-langchain4j/issues/371) 

cc @langchain4j 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new client for interacting with Mistral AI services, enabling chat completions, streaming chat completions, embeddings, and model listings.
- **Refactor**
	- Updated `MistralAiClient` to an abstract class with enhanced functionality and a new builder method for configuration.
- **Chores**
	- Changed access modifiers of several classes and methods from package-private to public, improving their usability across the codebase.

PR #834, opened 2024-03-26
Introduce SPI for creating Anthropic client

edeandrea:
## Context
This is very similar to what is done for Mistral (#744) and OpenAI. It will directly enable quarkiverse/quarkus-langchain4j#413.


ghost (approved):
@edeandrea thanks!

edeandrea:
My pleasure!

PR #840, opened 2024-03-27
feat: add ChatMemoryStore file implementation

vegetablest:
## Context
The current project only has an in-memory implementation for `ChatMemoryStore`.  To ensure message persistence, I added a file-based "FileChatMemoryStore" to store chat messages.

## Change
Added the `FileChatMemoryStore` class, which implements the `ChatMemoryStore` interface. This class allows storing chat messages in the local file system. Additionally, a test class, `FileChatMemoryStoreTest`, has been added to ensure the correctness of the `FileChatMemoryStore` implementation.

## Checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/FileChatMemoryStore.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/FileChatMemoryStore.java
@@ -1,0 +1,119 @@
+package dev.langchain4j.store.memory.chat;
+
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.ChatMessageDeserializer;
+import dev.langchain4j.data.message.ChatMessageSerializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Collections;
+import java.util.List;
+
+import static dev.langchain4j.internal.Exceptions.illegalArgument;
+import static dev.langchain4j.internal.Exceptions.runtime;
+
+/**
+ * Implementation of {@link ChatMemoryStore} that stores state of {@link dev.langchain4j.memory.ChatMemory} (chat messages) a local file.
+ * <p>
+ * This mechanism is more reliable than {@link InMemoryChatMemoryStore}.
+ */
+public class FileChatMemoryStore implements ChatMemoryStore {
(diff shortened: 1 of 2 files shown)


geoand:
CI is complaining that code coverage is not enough

vegetablest:
> CI is complaining that code coverage is not enough

Thanks for the reminder, I tried adding more unit tests.

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/FileChatMemoryStore.java
```diff
@@ -0,0 +1,133 @@
...
+        Path directoryPath = Paths.get(storageDirectory);
+        if (Files.exists(directoryPath)) {
+            if (!Files.isDirectory(directoryPath)) {
+                throw illegalArgument("Storage directory path exists but is a file: %s", storageDirectory);
+            }
+        } else {
+            try {
+                Files.createDirectories(directoryPath);
+            } catch (IOException e) {
+                throw runtime("Something is wrong, Failed to create folder %s", storageDirectory);
```
geoand, line 48: When an exception occurs, users will not know what happened the way this is done currently

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/FileChatMemoryStore.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+
+        private String storageDirectory = System.getProperty("java.io.tmpdir");
+        private String fileExtension = ".json";
```
geoand: Is there a good reason to make this confiigurable? Asking becayse the serializer and deserializer are not configurable so simply having a way to set the file extension can be deceptive

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/FileChatMemoryStore.java
vegetablest: > When an exception occurs, users will not know what happened the way this is done currently

I agree, I will commit new changes later when I have free time.

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/FileChatMemoryStore.java
vegetablest: > Is there a good reason to make this confiigurable? Asking becayse the serializer and deserializer are not configurable so simply having a way to set the file extension can be deceptive

Sorry, I'll modify it.

kpavlov (approved):
Thank you, @vegetablest 👍🏻

PR #842, opened 2024-03-27
(fix) addOptionalParameter for aiServices

zhangdashuai11:
The tool definition in aiServices needs to have a nullable parameter  i think


langchain4j:
@zhangdashuai11 I think there is a bug. Should it be `required` or `optional`?

```java
            P p = parameter.getAnnotation(P.class);
            if (Objects.nonNull(p) && p.required()) {
                builder.addOptionalParameter(parameter.getName(), toJsonSchemaProperties(parameter));
            }
```
You use `addOptionalParameter` when `required == true`, it should be vice-versa.

I am also wondering if instead of this we should use Optional<?> here? So, if the method parameter is Optional<?>, we use `addOptionalParameter` instead of `addParameter`. WDYT?

zhangdashuai11:
Yes, if the method parameter defined by the user is optional, we can set it as non-required. I didn't consider that the user might not annotate this. It seems like we don't have a strong dependency on the definition of this annotation.&nbsp;Let me make some modifications. Could you please take another look? Thank you for your help.






------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "langchain4j/langchain4j"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2024年4月11日(星期四) 晚上6:04
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [langchain4j/langchain4j] (fix) addOptionalParameter for aiServices (PR #842)





 
@zhangdashuai11 I think there is a bug. Should it be required or optional?
             P p = parameter.getAnnotation(P.class);             if (Objects.nonNull(p) &amp;&amp; p.required()) {                 builder.addOptionalParameter(parameter.getName(), toJsonSchemaProperties(parameter));             }  
You use addOptionalParameter when required == true, it should be vice-versa.
 
I am also wondering if instead of this we should use Optional here? So, if the method parameter is Optional, we use addOptionalParameter instead of addParameter. WDYT?

zhangdashuai11:
https://github.com/langchain4j/langchain4j/pull/922
this is the new pull request url&nbsp; thanks


------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "langchain4j/langchain4j"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2024年4月11日(星期四) 晚上6:04
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [langchain4j/langchain4j] (fix) addOptionalParameter for aiServices (PR #842)





 
@zhangdashuai11 I think there is a bug. Should it be required or optional?
             P p = parameter.getAnnotation(P.class);             if (Objects.nonNull(p) &amp;&amp; p.required()) {                 builder.addOptionalParameter(parameter.getName(), toJsonSchemaProperties(parameter));             }  
You use addOptionalParameter when required == true, it should be vice-versa.
 
I am also wondering if instead of this we should use Optional here? So, if the method parameter is Optional, we use addOptionalParameter instead of addParameter. WDYT?

PR #843, opened 2024-03-27
Fix #833: DocumentByParagraphSplitter regex JDK version adaptation

### LINKED ISSUE #833: [BUG] DocumentByParagraphSplitter does not split properly when \r\n line endings are used
It seems that `\r\n` is treated as a paragraph separator, which is wrong

Kugaaa:
## Context
See [#833](https://github.com/langchain4j/langchain4j/issues/833)

## Change
`DocumentByParagraphSplitter` split regular expression

## Checklist


ghost (approved):
@Kugaaa thanks a lot!

PR #844, opened 2024-03-27
removing classname before the json structure

tenpigs267:
## Context
This PR remove the classname that had been added in front of json structure in ServiceOutputParser as discusser here : https://github.com/langchain4j/langchain4j/pull/625#discussion_r1540700157

## Change
just removing the unwanted classname


ghost (approved):
@tenpigs267 thanks a lot!
I guess we should also make parsing JSON more robust, like extracting json from backticks (if present), removing garbage before/after curly braces, etc

tenpigs267:
> @tenpigs267 thanks a lot! I guess we should also make parsing JSON more robust, like extracting json from backticks (if present), removing garbage before/after curly braces, etc

yes removing backticks and unwanted comment around the doc could be a nice enhancement. I mainly use OpenAI and not sure if backticks is a common llm usage. That kind of code should be highly customisable for the library users so that it can adapt to a lot of situations... 

From the prompt side we can also increase reliability by adding a JSON output example (OpenAI works well without it but it is a common technique to improve reliability of the output)

PR #845, opened 2024-03-27
Refactor the code to avoid duplication between integrations

karesti:
Refactoring to allow reusing the code between integrations


ghost (approved):
@karesti thanks!

PR #853, opened 2024-03-28
ServiceOutputParser refactored for better flexibility

vaivanov95:
Context
Currently, when using DefaultAiServices API the output parser is pre-determined without an option for users to override them which means it's hard for users to make the required adjustments if for example a language model fails to produce output conforming to the instructions hard-coded in the ServiceOutputParser.

Change
Service output handlers are refactored to provide more control over output parsing without breaking the existing contract. A new class DefaultServiceOutputParser is added to replace the older, based on static methods ServiceOutputParser. ServiceOutputParser now uses pre-configured DefaultServiceOutputParser. ParsersBuilder introduced to give users an option to add a new parser or entirely redefine the parsers.

Output parsers are split into individual classes. Json also extracted into an output parser with an option for users to replace it with custom one by overriding the `OutputParserResolver.getDefaultParser()` method.


kwilliams-halosight:
I REALLY like this change. It cleans up the parser code nicely and provides several new and useful extension points.

I have a separate need that this change might help with:

I need to get the Usage information from the Response when using a Pojo. The problem is there's no way to get the response AND the pojo in the proxied AI Service object.

It looks to me like this code change could indirectly solve my problem.

I could just wrap the JSON parser with a simple facade that also extracts the usage.

Would it be better to use this (if accepted) or should I just add a method to fire events during parsing?

geoand:
I think this makes a lot of sense

langchain4j:
@kwilliams-halosight you can get both POJO and `TokenUsage` like this: https://github.com/langchain4j/langchain4j/blob/b998b052d9ddecf7652587919e1672a231122055/langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java#L650

langchain4j:
@vaivanov95 thanks a lot for your PR and so sorry for the late reaction! There are lots of conflicts now and this part of the code has been refactored already, so I assume this PR is not required any more.

PR #855, opened 2024-03-28
Integration with ClickHouse

Martin7-1:
## Context

[ClickHouse](https://clickhouse.com/) is the fastest and most resource efficient open-source database for real-time apps and analytics with full SQL support and a wide range of functions to assist users in writing analytical queries. Lately added data structures and distance search functions (like cosineDistance) as well as [approximate nearest neighbor search indexes](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/annindexes) enable ClickHouse to be used as a high performance and scalable vector database to store and search vectors with SQL.

## Change

Integration with Clickhouse as EmbeddingStore.

> Because of the [JSON](https://clickhouse.com/docs/en/sql-reference/data-types/json) type is experimental and is not fully supported by clickhouse-jdbc, currently the `metadata` is store by `Nullable(String)` type, which does not support Metadata Filtering.

## Checklist
- [x] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml) (only when a new module is added)

- [x] I have added a ClickhouseEmbeddingStoreIT that extends from either EmbeddingStoreIT or EmbeddingStoreWithFilteringIT


langchain4j-clickhouse/src/main/java/dev/langchain4j/store/embedding/clickhouse/ClickHouseEmbeddingStore.java
```diff
@@ -0,0 +1,310 @@
...
+                .map(ignored -> randomUUID())
+                .collect(toList());
+        addAllInternal(ids, embeddings, embedded);
+        return ids;
+    }
+
+    @Override
+    public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbedding, int maxResults, double minScore) {
+        try (Connection conn = createConnection();
+             Statement stmt = conn.createStatement();
```
tess1o: The statement has to be closed. Probably the best way is to wrap it with try with resources block

Martin7-1: It seems that it's already in try-with-resources block.

tess1o: It's not. Connection ("conn") will be closed, statement won't close automatically.  So you'll have a leak.

Martin7-1: But why? Statement is already in try-with-resources block. Please explain it in detail.

```java
try (Connection conn = createConnection();
       Statement stmt = conn.createStatement()) {
    // ...
}

Comment (tess1o) [reply langchain4j-clickhouse/src/main/java/dev/langchain4j/store/embedding/clickhouse/ClickHouseEmbeddingStore.java]: My bad, stupid syntax, didn't see the brackets correctly. Sorry.

**Comment by langchain4j** [comment]
Hi @Martin7-1, I really appreciate your contributions, and I apologize for the delayed reply. I'm definitely falling behind on PR reviews!

You've been contributing to LangChain4j since the early days and have provided many valuable PRs - thank you so much for that! I was wondering if you'd be interested in maintaining the https://github.com/langchain4j/langchain4j-community repository, where we’ll be adding most of the new integrations?

**Comment by Martin7-1** [comment]
> Hi @Martin7-1, I really appreciate your contributions, and I apologize for the delayed reply. I'm definitely falling behind on PR reviews!

@langchain4j Don't worry, this PR has a very low priority and [Clickhouse's support for ANN](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/annindexes) is still experimental. So I think you could review it after Clickhosue feature is stable.

> You've been contributing to LangChain4j since the early days and have provided many valuable PRs - thank you so much for that! I was wondering if you'd be interested in maintaining the https://github.com/langchain4j/langchain4j-community repository, where we’ll be adding most of the new integrations?

Of course! It's my pleasure, but I don't have any previous experience of maintaining a project of such large size, I would be very appreciated if I could get some guidance from you on what I should do :)

**Comment by langchain4j** [comment]
@Martin7-1 awesome! I was thinking about putting new integrations (and a few existing ones) that I do not have capacity to maintain into https://github.com/langchain4j/langchain4j-community repository where you (and maybe a few others) could maintain them (review PRs, fix bugs, release new versions). So you will get all the needed access and will be able to setup the repo/project structure, CI, etc. We can also include links to it from main repo and documentation so that people will be aware of it. How does it sound? BTW ClickHouse can be the first one in that repo.

**Comment by Martin7-1** [comment]
> @Martin7-1 awesome! I was thinking about putting new integrations (and a few existing ones) that I do not have capacity to maintain into https://github.com/langchain4j/langchain4j-community repository where you (and maybe a few others) could maintain them (review PRs, fix bugs, release new versions). So you will get all the needed access and will be able to setup the repo/project structure, CI, etc. We can also include links to it from main repo and documentation so that people will be aware of it. How does it sound? BTW ClickHouse can be the first one in that repo.

Sounds great! But one question is about documents. Looks like contributors need to contribute to `langchain4j-community` with code and `langchain4j` with document. Is there any solution?

**Comment by Martin7-1** [comment]
BTW, I'm working on using `ClickHouse` JSON type to store metadata and filtering. Since it's an experimental feature, maybe `clickhouse-java` library may not fully support it...

The reason I want to store metadata as JSON type is because `langchain` implement it in this way, but I think the alternative way is to create each metadata field as a column of `Clickhouse` table, although it may be a little complicated. (Because we need to define the data type of each column)

**Comment by langchain4j** [comment]
@Martin7-1 regarding documentation, I am sure we will figure it out with time.

@Martin7-1 regarding JSON for metadata: sounds reasonable, but I am sure you have better understanding, so please proceed as you see fit

**Comment by Martin7-1** [comment]
@langchain4j For ClickHouse, currently I implement it by treating each metadata key as different table column. Now it supports filter and removal. You could review it now :)

For `langchain4j-community`, I'm struggling with which of the following to use in root `pom.xml`:

* Implement a seperate parent `pom`, like `langchain4j-spring` and `langchain4j-embeddings` do. But I think it will cost more effort on manage dependencies in order to keep consistence with `langchain4j`
* Just set `langchain4j-parent` as parent pom, so that the dependencies can be managed together. `Spring Data` also do like this way.

WDYT?

Another point which needed disscuss is abount `bom`, should they put together or seperate? I thinik put together will be better, users can introduce them in a unified way in pom. [Spring Data Bom](https://github.com/spring-projects/spring-data-bom) also do so. WDYT?

**Comment by langchain4j** [comment]
@Martin7-1 

> For ClickHouse, currently I implement it by treating each metadata key as different table column. Now it supports filter and removal. You could review it now :)

Great! Perharps you want to have it as a first member of https://github.com/langchain4j/langchain4j-community ?

> Implement a seperate parent pom, like langchain4j-spring and langchain4j-embeddings do. But I think it will cost more effort on manage dependencies in order to keep consistence with langchain4j

What do you mean by more effort? It seems to work fine for langchain4j-spring and langchain4j-embeddings

> Another point which needed disscuss is abount bom, should they put together or seperate?

I would put it separate, as we have separate repositories and the lifecycle will be separate.

**Comment by Martin7-1** [comment]
> Great! Perharps you want to have it as a first member of https://github.com/langchain4j/langchain4j-community ?

Sure! I will create a PR to init project and a PR about ClickHouse asap.

> What do you mean by more effort? It seems to work fine for langchain4j-spring and langchain4j-embeddings

We may need to reintroduce some dependencies in `langchain4j-community` to keep the consistence (i.e. the version of dependencies) between `langchain4j` and `langchain4j-community`, such as test dependencies (`assertj`, `testcontainers` and so on) and common dependencies (slf4j-api, retrofit and so on). 

For example, users may introduce different versions of `retrofit` when they introduce `langchain4j-ollama` and a dependency which uses `retrofit` in `langchain4j-community` at the same time. We need extra effort to keep the consistence of dependencies's version if one day we bump it.

But if we just use `langchain4j-parent` as parent pom (or just introduce `langchain4j-bom` in `<dependencyManagement>` ?), all we need to do is to change the version of `langchain4j-parent`.

> I would put it separate, as we have separate repositories and the lifecycle will be separate.

OK. Sounds great!

**Comment by langchain4j** [comment]
@Martin7-1 

> But if we just use langchain4j-parent as parent pom (or just introduce langchain4j-bom in <dependencyManagement> ?), all we need to do is to change the version of langchain4j-parent.

If it works, why not? :)


I would also use different naming scheme for community integrations:
```xml
<dependency>
    <artifactId>langchain4j-community-{name}</artifactId>
    <version>0.35.0</version>
</dependency>
```

And packages:
```java
package dev.langchain4j.community.x.y.z;
```

Martin7-1:
Close and reopen in `langchain4j-community`

PR #861, opened 2024-03-29
[PoC] Implement a state machine in the AiServices

mariofusco:
For now this is only a PoC to start a discussion around the possibility of implementing a state machine inside the AiServces. This pull request is a complete rework of [this other one](https://github.com/langchain4j/langchain4j/pull/672) and tries to put my proposal at work using the same use case that I implemented [here](https://github.com/mariofusco/quarkus-drools-llm?tab=readme-ov-file#the-airline-chatbot-example).

In my opinion the original pull request was a bit inflexible, trying to model all the states and possible transactions among them inside the AiServices itself, and this also led to a more cumbersome API. The idea here is that how the transactions among the different states are modeled should be totally external to the AiServices. This increased flexibility also allows to model the state machine with other external tools, like a rule engine in my case. 

Note that this pull request (like the original one) still lacks the possibility to define a state machine for each user/conversation, but I believe this could be implemented with an id-based mechanism, very similar to the one already used for the chat memory. Moreover, since in my test example I'm using ollama that doesn't support tools, I didn't try to put also them under the control of the state machine, but once we agreed on the implementation this could also be added in a relatively easy way.

While working on this feature however I realized that probably also the concept of a proper state machine is quite over-engineered for the langchain4j needs, and what would be necessary is more simply a way to make system and user messages (and eventually tools) configurable in a programmatic way. I developed this idea in [another pull request](https://github.com/langchain4j/langchain4j/pull/862).

### DIFF
--- a/langchain4j/pom.xml
+++ b/langchain4j/pom.xml
@@ -54,6 +54,13 @@
 
         <dependency>
             <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-ollama</artifactId>
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
             <artifactId>langchain4j-open-ai</artifactId>
             <version>${project.version}</version>
             <scope>test</scope>
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceContext.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceContext.java
@@ -8,6 +8,7 @@ import dev.langchain4j.model.chat.ChatLanguageModel;
 import dev.langchain4j.model.chat.StreamingChatLanguageModel;
 import dev.langchain4j.model.moderation.ModerationModel;
 import dev.langchain4j.rag.RetrievalAugmentor;
+import dev.langchain4j.statemachine.StateManager;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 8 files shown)


langchain4j:
Hi @mariofusco, thanks a lot! I'll have a closer look after a long weekend, happy holidays!

mariofusco:
> Hi @mariofusco, thanks a lot! I'll have a closer look after a long weekend, happy holidays!

Consider this my Easter egg :) Enjoy your holidays.

langchain4j:
Hi @mariofusco, I like the idea of externalizing state transition logic, users should definitely be able to have total freedom there.

A bit of context: I was considering having 2 options for implementing state machines with AI Services:
1. To give the user total freedom. I think [this PR](https://github.com/langchain4j/langchain4j/pull/862) alligns pretty well with it and what users have requested.
2. To have an out-of-the-box implementation of a state machine that will fit 80% of use cases. It should be very easy to setup and customize. My [initial PR](https://github.com/langchain4j/langchain4j/pull/672) was addressing this option.

I will need more time to play with your proposal and other alternatives and see what will be a good fit for the option 2.
There are still many open questions like: who should drive/initiate state transitioning: LLM, application, or should it be configurable, etc, etc.

But for now I think it is safe to go ahead with option 1 as it will be helpful anyway for many other use cases not involving state machines.

mariofusco:
> But for now I think it is safe to go ahead with option 1 as it will be helpful anyway for many other use cases not involving state machines.

As I wrote, I completely agree that 1. is the best option: on the short term it will enable the implementation of a few use cases, including the use of a completely external state machine as in my use case, and it could also be seen as a first step for the implementation of an internal state machine if we will decide that it is necessary.

Since we are on the same page can we continue this discussion on the [other pull request](https://github.com/langchain4j/langchain4j/pull/862) and check together what's missing there? I guess that the first thing to be implemented is the support to configure messages for each user/conversation, correct? Can we use a mechanism similar to what has been already developed for the chat memory for this or do you have any better idea?

PR #867, opened 2024-04-01
fix ollama client response handle

NeatGuyCoding:
Thanks for @wangrushuang [wangrushuang](https://github.com.wangrushuang), she find this problem and I provide a way to fix that.


langchain4j:
@HashZhang please provide some context: what is the problem and how do you fix it?
Please fill in the PR template. Thank you.

NeatGuyCoding:
Sorry, the problem is that the original code uses a fixed length byte array buffer for the stream response. If one of the stream response is over 1024 then the last json may miss some subparts. Therefore we should use some separator to split the response and streaming deserialize it. Use BufferedReader would be a way to fix that.

langchain4j:
@HashZhang how did you test/reproduce the error? Thanks

ghost (approved):
@HashZhang thanks!

PR #870, opened 2024-04-01
refactor: redefine HuggingFaceModelName constant class

vegetablest:
## Context
As a constant class, the class is typically immutable and has a private constructor.


ghost (approved):
@vegetablest thanks!

PR #894, opened 2024-04-07
Removing incorrect end comma in jsonStructure for POJO extracting AI Services

LizeRaes:
## Context
I noted that for more complex POJO object, the JSON returned by the LLM often contained a last comma before closing bracket, making the JSON invalid and resulting in an error while parsing it. I traced it back to the proposed jsonStructure that is part of the SystemMessage, sending a wrong format to the LLM.

## Change
Removed the last comma that shouldn't be there in the jsonStructure + updated the unit tests

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
@@ -14,7 +14,6 @@ import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.util.*;
-import java.util.stream.Collectors;
 
 import static dev.langchain4j.exception.IllegalConfigurationException.illegalConfiguration;
 import static java.lang.String.format;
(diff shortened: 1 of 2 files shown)


langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
```diff
@@ -130,6 +130,7 @@ private static String jsonStructure(Class<?> structured, Set<Class<?>> visited)
             }
             jsonSchema.append(format("\"%s\": (%s),\n", name, descriptionFor(field, visited)));
         }
+        jsonSchema.delete(jsonSchema.lastIndexOf(","), jsonSchema.lastIndexOf(",")+1);
```
Shastick, line 133: Great one-liner fix. We could consider doing this cleanly with a .join() but that would require switching to streams so this is fine in my opinion.

Shastick (approved):

langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
mmanrai: This is causing following exception while trying to create json structure for classes/interfaces without declared fields: 

`Range [-1, 0) out of bounds for length 2. java.lang.RuntimeException: Range [-1, 0) out of bounds for length 2`

Shastick: Fixed with https://github.com/langchain4j/langchain4j/pull/975

PR #708, opened 2024-03-07
Tools: support schema for custom POJO parameters

gs-snagaraj:
Support POJOs as `@Tool` method parameters.
Repurpose `@Description` to be used to describe POJO fields.

Example:

```java
@Data
static class Query {

    @Description("List of fields to fetch records")
    List<String> select;

    @Description("List of conditions to filter on. Pass null if no condition")
    List<Condition> where;

    @Description("limit on number of records")
    Integer limit;

    @Description("offset for fetching records")
    Integer offset;
}

@Data
static class Condition {

    @Description("Field to filter on")
    String field;

    @Description("Operator to apply")
    Operator operator;

    @Description("Value to compare with")
    Object value;
}

enum Operator {

    EQUALS,
    NOT_EQUALS,
    IS_NULL,
    IS_NOT_NULL
}

@Tool
public String executeQuery(Query query) {
    ....
}
```

PR #897, opened 2024-04-08
Anthropic: Support tools

ghost:
## Context
Anthropic recently began supporting tools: https://docs.anthropic.com/claude/docs/tool-use

## Change
This PR adds support for tools in `AnthropicChatModel`.
`AnthropicStreamingChatModel` does not support tools yet, because Anthropic does not support tools when streaming.

## Checklist
- [X] Test in AiServicesWithToolsIT


docs/docs/tutorials/6-tools.md
```diff
@@ -121,29 +121,82 @@ Please note that tools/function calling is not the same as [JSON mode](/tutorial
...
+- `ToolSpecifications.toolSpecificationsFrom(Object)`
+- `ToolSpecifications.toolSpecificationFrom(Method)`
+
+```java
+class WeatherTools { 
+  
+    @Tool("Returns the weather forecast for a given city")
+    String getWeather(
+            @P("The city for which the weather forecast should be returned") String city,
+            TemperatureUnit temperatureUnit
```
daixi98, line 162: We can parse Enum, but nested object parameters are not supported yet. Is it correct?

daixi98 (approved):
LGTM

docs/docs/tutorials/6-tools.md
langchain4j: Yes, enums are supported: https://github.com/langchain4j/langchain4j/blob/1827302342eee393fddfce8a07f3cab2b7d5b9e2/langchain4j-core/src/test/java/dev/langchain4j/agent/tool/ToolSpecificationsTest.java#L60

Nested POJOs will also be supported soon: https://github.com/langchain4j/langchain4j/pull/708

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicMapper.java
```diff
@@ -44,52 +114,29 @@ static String toAnthropicSystemPrompt(List<ChatMessage> messages) {
...
                 .map(content -> content.text)
                 .collect(joining("\n"));
-        return AiMessage.from(text);
+
+        List<ToolExecutionRequest> toolExecutionRequests = contents.stream()
+                .filter(content -> "tool_use".equals(content.type))
+                .map(content -> ToolExecutionRequest.builder()
+                        .id(content.id)
+                        .name(content.name)
+                        .arguments(GSON.toJson(content.input))
```
geoand, line 129: This unfortunately breaks the cases where GSON is not used (like Quarkus which swaps the JSON handling for Jackson in the supported modules).

langchain4j: Oh right! Let me fix this

geoand: 🙏🏼

langchain4j: Using Gson [here](https://github.com/langchain4j/langchain4j/blob/d118c0c3b67e01bb2f159cf7a388533dac06ffad/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/DefaultAnthropicClient.java#L168) is ok for you?

geoand: That one never gets called in the Quarkus case (or at least I didn't see it), so it should be fine

Issue #586, opened 2024-02-01
[FEATURE] Ollama Authentication Header injection


sraeuchle:
Hey there, 

I have run a Ollama Instance behind a Authentication Proxy so i need to send a Authentication Header for every Ollama Request.
In your current OllamaClient implementation I can't handle that in any way. See: https://github.com/langchain4j/langchain4j/blob/057730bc7b2f50e0bca3614bbb28b477a24e2f61/langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java#L33

please add a OKHttpClient on your builder as optional so I am able to do an inject like so:
```
OkHttpClient client = new OkHttpClient.Builder().addInterceptor(new Interceptor() {
      @Override
      public Response intercept(Chain chain) throws IOException {
        Request newRequest  = chain.request().newBuilder()
            .addHeader("Authorization", "Bearer " + token)
            .build();
        return chain.proceed(newRequest);
      }
    }).build();

OllamaChatModel.builder()
    .client(client)
    ....
```

Thanks a lot!

mnnayeck:
total blocker :(

langchain4j:
Implemented in https://github.com/langchain4j/langchain4j-spring/pull/21

PR #911, opened 2024-04-09
Ollama add OkhttpClient inject

Martin7-1:
## Context

see #586 

## Change

Add `OkhttpClient` injection in `Ollama` models

## Checklist


ghost (commented):
Hi @Martin7-1 thanks a lot!

Please avoid adding `OkHttpClient` to `Ollama*Model` API as we plan to move away from okhttp pretty soon.

Instead, I would propose to allow the user to do the following (we already have this in `OpenAiChatModel`):
```java
OllamaChatModel.builder()
                ...
                .defaultHeaders(Map.of("Authorization", "Bearer ..."))
                .build();
```

WDYT?

Martin7-1:
> Hi @Martin7-1 thanks a lot!
> 
> Please avoid adding `OkHttpClient` to `Ollama*Model` API as we plan to move away from okhttp pretty soon.
> 
> Instead, I would propose to allow the user to do the following (we already have this in `OpenAiChatModel`):
> 
> ```java
> OllamaChatModel.builder()
>                 ...
>                 .defaultHeaders(Map.of("Authorization", "Bearer ..."))
>                 .build();
> ```
> 
> WDYT?

@langchain4j I think it's a good idea! Will change soon.

Martin7-1:
@langchain4j Hi! I've change it without `OkhttpClient` in constructor. Please have a look. BTW, there seems no official docker image with Auth, but I found some community docker image which implement it with Caddy, such as [ollama-auth](https://github.com/g1ibby/ollama-auth) and [ollama_proxy_server](https://github.com/ParisNeo/ollama_proxy_server). Maybe we need to choose community docker image to test custom headers. But it seems not a good way due to consistency...WDYT? Do you have any idea to test it with official docker image?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
```diff
@@ -200,6 +209,23 @@ public OllamaModelCard showInformation(ShowModelInformationRequest showInformati
...
     }
 
+    private Interceptor headerInterceptor(Map<String, String> customHeaders) {
+        return new Interceptor() {
+            @NotNull
+            @Override
+            public okhttp3.Response intercept(@NotNull Chain chain) throws IOException {
+                Request original = chain.request();
+                Request.Builder requestBuilder = original.newBuilder()
+                        .method(original.method(), original.body());
```
langchain4j: is there a reason to re-define method and body? [GenericHeaderInjector](https://github.com/ai-for-java/openai4j/blob/e4d41a0463c6ad8fdd88d709026dbab2a762fe67/src/main/java/dev/ai4j/openai4j/GenericHeaderInjector.java#L22) seems to work fine without it.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
```diff
@@ -200,6 +209,23 @@ public OllamaModelCard showInformation(ShowModelInformationRequest showInformati
         }
     }
 
+    private Interceptor headerInterceptor(Map<String, String> customHeaders) {
+        return new Interceptor() {
```
langchain4j: please make a copy of `customHeaders` in this method and use it instead of the original map

ghost (commented):
@Martin7-1 thank you!

langchain4j:
@Martin7-1 regarding testing, we can do manual testing here if there is no easy way to automate it.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
Martin7-1: done.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
Martin7-1: Actually no reason...already removed.

Martin7-1:
> @Martin7-1 regarding testing, we can do manual testing here if there is no easy way to automate it.

Ok. I will add some tests.

ghost (approved):
@Martin7-1 thank you a lot!

langchain4j:
@Martin7-1 Could you please add this new parameter to https://github.com/langchain4j/langchain4j-spring/blob/main/langchain4j-ollama-spring-boot-starter/src/main/java/dev/langchain4j/ollama/spring/AutoConfig.java?

Martin7-1:
> @Martin7-1 Could you please add this new parameter to https://github.com/langchain4j/langchain4j-spring/blob/main/langchain4j-ollama-spring-boot-starter/src/main/java/dev/langchain4j/ollama/spring/AutoConfig.java?

@langchain4j ok. I've done it in [langchain4j-spring#21](https://github.com/langchain4j/langchain4j-spring/pull/21).

PR #912, opened 2024-04-09
Update to Azure Identity 1.12.0

jdubois:


oscerd (approved):
I believe it would be better to use the azure-sdk-bom instead of managing singular versions

jdubois:
As long as we only have Azure Identity and Azure Blob Storage, I'd rather stick to the direct dependencies and not the BOM: the BOM is automatically released every month (see https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/boms/azure-sdk-bom/README.md ), so you might need to wait a few weeks to get the newest versions.

ghost (approved):
@jdubois thanks!

PR #922, opened 2024-04-11
Tools:  Allow using Optional<T> method parameters

zhangdashuai11:
when use tools  if user define method parameter as Optional addOptionalParameter


langchain4j:
@zhangdashuai11 please add tests and document this change. Please see https://github.com/langchain4j/langchain4j/blob/main/CONTRIBUTING.md

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
```diff
@@ -56,6 +57,10 @@ public static ToolSpecification toolSpecificationFrom(Method method) {
             if (parameter.isAnnotationPresent(ToolMemoryId.class)) {
                 continue;
             }
+            //if user define method parameter as Optional addOptionalParameter
+            if (parameter.getType().equals(Optional.class)) {
+                builder.addOptionalParameter(parameter.getName(), toJsonSchemaProperties(parameter));
+            }
             builder.addParameter(parameter.getName(), toJsonSchemaProperties(parameter));
```
langchain4j, line 64: This should probybly be in `else` block.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
```diff
@@ -56,6 +57,10 @@ public static ToolSpecification toolSpecificationFrom(Method method) {
             if (parameter.isAnnotationPresent(ToolMemoryId.class)) {
                 continue;
             }
+            //if user define method parameter as Optional addOptionalParameter
+            if (parameter.getType().equals(Optional.class)) {
+                builder.addOptionalParameter(parameter.getName(), toJsonSchemaProperties(parameter));
```
langchain4j, line 62: `toJsonSchemaProperties` will receive `Optional` type, but should probably receive an actual type

ghost (changes requested):

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
Kugaaa: `dev.langchain4j.agent.tool.DefaultToolExecutor#coerceArgument` also need packe `Optional`?

Also, I have a question: Is `Optional` suitable as a parameter?
[Why should Java 8's Optional not be used in arguments](https://stackoverflow.com/questions/31922866/why-should-java-8s-optional-not-be-used-in-arguments)

langchain4j: I have no strong opinion on this, as always it depends...

```java
@Tool
void method(@P(optional = true) String parameter) {...}
```

or 

```java
@Tool
void method(Optional<String> parameter) {...}
```

what would you choose?

Kugaaa: I vote for option one~

langchain4j: @geoand @jmartisk WDYT?

geoand: I don't have a super strong opinion either, but I propably slightly lean towards `1`

langchain4j: Ok, I guess it does not harm to have both options, but for now let's go with the first one then

langchain4j:
Implemented in https://github.com/langchain4j/langchain4j/pull/1370

PR #925, opened 2024-04-11
Fix #876

### LINKED ISSUE #876: [BUG] The method 'toEmbedTexts' in DefaultZhipuAiHelper may have a bug.
When processing textSegments in the 'toEmbedTexts' method of DefaultZhipuAiHelper, only the first item of the result after streaming conversion is taken. If the size of the passed List<TextSegment> is greater than 1, this handling will cause: 1. Data loss, 2. An inconsistency between the entries of 'embeddings' and 'segments' during the subsequent embeddingStore.addAll(embeddings, segments), resulting in detection exceptions. It is unclear whether the 'return embedText.get(0)' is simply a bug or intentionally designed this way.
bug

xermaor:
## Context

Fix: #876 

## Checklist
- [ ] All unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules are green


ghost (approved):
@1402564807 thank you!

Issue #903, opened 2024-04-09
[FEATURE] AI Service method parameter annotated with @UserMessage should support template placeholders


langchain4j:
Reported in https://github.com/langchain4j/langchain4j/issues/805.

Currently, `@UserMessage` annotation placed on the method of AI Service supports placeholders:
```java
interface Greeter {

        @UserMessage("Hi, my name is {{name}}")
        String greet(@V("name") String name);
}
```

Placeholders should be also supported when some String parameter is annotated with `@UserMessage`:
```java
interface Greeter {

        String greet(@UserMessage String userMessage, @V("name") String name);
}
...
String greeting = greeter.greet("Hi, my name is {{name}}", "Klaus");
```

edeandrea:
Should this have been closed when #930 was merged?

langchain4j:
@edeandrea yeah, thanks!

PR #930, opened 2024-04-12
AIService: support template placeholders for parameters annotated wit…

daixi98:
## Context

For Issue #903. The goal of this task is to support template placeholders for parameters annotated with `@UserMessage`.
## Change

When generating UserMessage template, we will first check if there is any `@UserMessage` annotation on top of the method. If not, we will then check if there is any parameter annotated with `@UserMessage`, and we will use the value of that parameter as the user message template.

## Checklist


ghost (approved):
@daixi98 thanks a lot! I was working on the same thing (forgot to assign myself). I am now doing refactoring in this class and adding more ITs.

Issue #896, opened 2024-04-08
[FEATURE] Support for custom types in openai tool/function interface


limoiie:
**Is your feature request related to a problem? Please describe.**

Currently, only Java's built-in types (boolean, byte, short, int, long, float, double, BigDecimal, BigInteger, String, Array, and Array-like) are supported as parameters for tools or functions. Still, custom types may be needed for convenience.

*Example*: Suppose I am developing database tools; each tool works with a specific column type. Without custom types, I would need to repeatedly specify database, table, and column names in each tool's interface. 
```java
class DatabaseToolkit {
    @Tool
    void listByDatetime(
        @P("Database name, it can be ....")
        String dbName,
        @P("Table name, it can be ....")
        String tableName,
        @P("Column name, it can be ....")
        String columnName,
        long start, long end) {
        // ...
    }

    @Tool
    void listByMatch(
        // bad: repeated definition and description
        @P("Database name, it can be ....")
        String dbName,
        @P("Table name, it can be ....")
        String tableName,
        @P("Column name, it can be ....")
        String columnName,
        String pattern) {
        // ...
    }
   
    // more repeated... 
}
```

By organizing these names in a custom type like Column, I can avoid repeating them in every tool or function interface. 

```java
class Column {
        @P("Database name, it can be ....")
        String dbName;
        @P("Table name, it can be ....")
        String tableName;
        @P("Column name, it can be ....")
        String columnName;
        // ...
}

class DatabaseToolkit {
    @Tool
    void listByDatetime(
        Column column,
        long start, long end) {
        // ...
    }

    @Tool
    void listByMatch(
        Column column,
        String pattern) {
        // ...
    }
   
    // more repeated... 
}
```

Additionally, I can manipulate multiple columns at once by using a list of Column instances. Achieving this functionality would be challenging without custom types.

**Describe the solution you'd like**

The current implementation describes the tool or function interface, including all its parameters, in [JsonSchema](https://json-schema.org/understanding-json-schema/), as required by [OpenAI doc](https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools). Python LangChain does so using [pydantic](https://github.com/pydantic/pydantic). By following this convention, we can support custom type to be tool/function parameters in two steps:
- Build a JsonSchema Generator to convert correctly annotated custom type into JsonSchema;
- Build a JsonSchema Validator to validate and coerce the argument generated by LanguageModel into the custom type.

[...]

PR #941, opened 2024-04-16
Feature #896: Support custom-type tool parameter

limoiie:
## Context

Langchain4j currently only supports Java built-in types as tool parameters.
While it's advisable to maintain a flat and simple tool interface,
there are situations where using custom-type tool parameters is preferred.

## Change

This PR delegates the generation of `ToolSpecification` and the deserialization of `ToolExecutionRequest`'s arguments to `ToolJsonSchemas` using `JsonSchemaService`. `JsonSchemaService` is a separate pluggable dependency via SPI. This PR created two implementations for `JsonSchemaService`:
- `DefaultJsonSchemaService` (in module `langchain4j-core`) is consistent with the old behavior of Langchain4j. It supports only Java Built-in types as tool parameters, and sanitizes the tool execution arguments before deserialization
- `VictoolsJsonSchemaService` (in a separated model `langchain4j-jsonschema-service-victools`) adopts [victools/jsonschema-generator](https://github.com/victools/jsonschema-generator)  for JSON Schema generation and arguments deserialization.

Besides SPI, the user can override the default by customizing an implementation of `JsonSchemaService` via `AiServices#jsonSchemaService`

The original `ToolSpecification` is kept for now to stay compatible with the current codebase. It might be replaced by `ToolJsonSchema` in the future if needed.

## Checklist
- [x] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml) (only when a new module is added)


langchain4j:
Issue: https://github.com/langchain4j/langchain4j/issues/896

sbeugen:
Hi @langchain4j, will this PR be merged any time soon? Without this feature it's not even possible to call a tool with a List type.

langchain4j:
Hi @sbeugen yes, I plan to tool into it this week

langchain4j:
@sbeugen

> Without this feature it's not even possible to call a tool with a List type.

List of primitives is currently supported as a tool parameter, here is an example: https://github.com/langchain4j/langchain4j/blob/ae5c29893f37c73bde5f1621a6c7a0a75bcac93d/langchain4j/src/test/java/dev/langchain4j/service/StreamingAiServicesWithToolsIT.java#L68

sbeugen:
@langchain4j 
You are right. With 0.31.0 it works. I still had an older version.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolJsonSchema.java
```diff
@@ -0,0 +1,52 @@
+package dev.langchain4j.agent.tool;
+
+import dev.langchain4j.jsonschema.JsonSchema;
+
+import java.util.Map;
+
+/**
+ * The JsonSchema that describes a {@link Tool}.
+ */
+public class ToolJsonSchema {
```
langchain4j, line 10: Is there a reason a new type is introduced instead of using `ToolSpecification`?

langchain4j-core/src/test/java/dev/langchain4j/agent/tool/ToolExecutionRequestUtilTest.java
```diff
@@ -6,19 +6,19 @@
...
-
-        assertThat(ToolExecutionRequestUtil.argumentsAsMap(request.arguments()))
-                .containsEntry("foo", "bar")
-                        .containsEntry("qux", 12.0);
-
-        assertThat((String) ToolExecutionRequestUtil.argument(request, "foo"))
-                .isEqualTo("bar");
-        assertThat((Number) ToolExecutionRequestUtil.argument(request, "qux"))
-                .isEqualTo(12.0);
+        // ToolExecutionRequest request = ToolExecutionRequest.builder()
```
langchain4j, line 9: Probably commented out by mistake?

ghost (commented):
Hi @limoiie, thank you a lot! And sorry for the delay with the review. This is quite a big change and hard to grasp at once.

I just wanted to double-check, before I dive deeper: did you have a chance to test if OpenAI and other LLM providers successfully accept the generated schema and can generate adequate json back?

Issue #904, opened 2024-04-09
[FEATURE] Support more advanced POJO/JSON extraction features


langchain4j:
See https://github.com/LizeRaes/feedback-analyzer/blob/main/src%2Fmain%2Fjava%2Fengineering%2Fepic%2FAtomicFeedback.java

For example, possible values of `Tag` in `List<Tag> tags` are not propagated in the prompt

remyohajinwa:
Hi I am taking a look at this

langchain4j:
Hi @LizeRaes, the link does not work any more :(
Do you still remember what was the problem with `List<Tag> tags`?
Thanks!

LizeRaes:
Here's the [working link](https://github.com/LizeRaes/feedback-analyzer/blob/main/src/main/java/engineering/epic/datastorageobjects/AtomicFeedback.java).
This was a List of Enum and was not supported back then, I think it's long okay right now :)

langchain4j:
@LizeRaes thanks!

Issue #916, opened 2024-04-10
Cover all extraction use cases with ITs


langchain4j:
- all listed [here](https://docs.langchain4j.dev/tutorials/ai-services/#output-parsing-aka-structured-outputs)
- as well as `Map`
- Complex POJOs (with lists, enum lists, etc)

Make sure it is covered for all extraction methods (JSON schema, JSON mode, tools, plain text, etc).

## Strings
- [x] `String`
- [x] `List<String>`
- [x] `Set<String>`

## POJOs
- [x] `POJO`
- [x] `List<POJO>`
- [x] `Set<POJO>`

## Enums
- [x] `Enum`
- [x] `List<Enum>`
- [x] `Set<Enum>`

## Primitives
- [x] `boolean`
- [x] `int`
- [x] `long`
- [x] `float`
- [x] `double`
- [ ] `List<primitive>`
- [ ] `Set<primitive>`

## Other
- `Map<K, V>`
- `Result<T>` where `T` is any type listed above
- All common Java data types
- `Optional<T>`

langchain4j:
Partially covered in `AiServicesWithJsonSchemaIT`, `AiServicesWithJsonSchemaWithDescriptionsIT` and `AiServicesIT`

Issue #909, opened 2024-04-09
[FEATURE] Consider supporting Iterable<String> / Stream<String> / etc as a return type of AI Services


langchain4j:
Currently streaming in AI Services can be only done using `TokenStream`.
Consider supporting other types like `Iterable<String>` / `Stream<String>` / etc.

langchain4j:
Related: https://github.com/langchain4j/langchain4j/issues/886

Issue #725, opened 2024-03-10
Allow @Description on Enum values


BlueJoneleth:
Hi, 

I mentioned it in Discord. It would be handy if you could annotate enumvalues with a description so you could tell the AI when to use it. 

What currently work is this:    
` @Description("
            "Extract the following values based on {{it}}" +
            "Possible values and when to use them: " +
            "-ENUMVALUE1: description1," +
            "-ENUMVALUE2: description2")
    private List<ENUMTYPE> enumList;`

What would be handy is this: 

` public enum ENUMTYPE{

    @Description("description1")
    ENUMVALUE1,
    @Description("description2")
    ENUMVALU2`

langchain4j:
@BlueJoneleth good idea, thanks! Are you going to open a PR? 🙏

langchain4j:
Also need to make sure that `List<Enum>` works as well:
```java
enum Tag {
    RoomEnvironment, Catering, Timing
}

class Pojo {
    List<Tag> tags;
}
```

PrimosK:
I am also interested in the following functionalities:

1. Annotating enums with @Description annotation
2. Properly supporting the return of a list of enums

**Side note:**

**a.)** While looking at the code I've noticed that `dev.langchain4j.model.output.EnumOutputParser#parse` is not called anywhere.

**b.)** [Sentiment_Extracting_AI_Service_Example](https://github.com/langchain4j/langchain4j-examples/blob/201e07d33c6361cc35d84a351fb96161609cd048/tutorials/src/main/java/_08_AIServiceExamples.java#L117) does not work well with the gpt-4o model. It returns enums enclosed in [ and ] which causes deserialization to fail. This issue occurred several times during my testing.

**Proposed Solution**

In addition to features (1. and 2.) my plan is to also fix both issues:

**Fix Issue (a)**: This seems like an easy fix.

**Fix Issue (b)**: This will require more detailed instructions, preferably with an example. Here is a suggested approach:

```
Please answer with valid JSON strictly according to this JSON Schema Definition:
|||
{
  "type": "string",
  "oneOf": [
    {
      "const": "SUNNY",
      "description": "Clear skies with no clouds or rain."
    },
    {
      "const": "CLOUDY",
      "description": "Mostly cloudy with minor chances of rain."
    },
    {
      "const": "RAIN",
      "description": "Precipitation with a majority of the time being rainy."
    }
  ]
}
|||

Example answer:
"A"
```

[...]

PR #975, opened 2024-04-18
#904 Support advanced POJO/JSON extraction features

remyohajinwa:
## Context

This was tested and seen that the POJO/JSON Extraction works perfectly. Sample test scenario was derived from the [feedback-analyzer](https://github.com/LizeRaes/feedback-analyzer/blob/main/src/main/java/engineering/epic/datastorageobjects/AtomicFeedback.java) repository.

I also took care of an edge case (where we have a class without fields) which was not handled. without this change, we would have exceptions as shown in the screenshot below:

Screenshot 2024-04-18 at 22 16 55

## Checklist
- [ x All unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules are green

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
@@ -130,7 +130,11 @@ public class ServiceOutputParser {
             }
             jsonSchema.append(format("\"%s\": (%s),\n", name, descriptionFor(field, visited)));
         }
-        jsonSchema.delete(jsonSchema.lastIndexOf(","), jsonSchema.lastIndexOf(",")+1);
+
+        int trailingCommaIndex = jsonSchema.lastIndexOf(",");
+        if (trailingCommaIndex > 0) {
+            jsonSchema.delete(trailingCommaIndex, trailingCommaIndex +1);
+        }
         jsonSchema.append("}");
         return jsonSchema.toString();
     }
--- a/langchain4j/src/test/java/dev/langchain4j/service/ServiceOutputParserTest.java
+++ b/langchain4j/src/test/java/dev/langchain4j/service/ServiceOutputParserTest.java
@@ -232,6 +232,10 @@ class ServiceOutputParserTest {
         private PersonWithParentArray[] parents;
     }
 
+    static class ClassWithNoFields {
+
+    }
+
     @Test
     void outputFormatInstructions_PersonWithParentArray() {
         String formatInstructions = ServiceOutputParser.outputFormatInstructions(PersonWithParentArray.class);
@@ -246,6 +250,15 @@ class ServiceOutputParserTest {
                             "\"parents\": (type: array of dev.langchain4j.service.ServiceOutputParserTest$PersonWithParentArray)\n" +
                             "})\n" +
                         "}");
+    }
+
+    @Test
+    void outputFormatInstructions_ClassWithNoFields() {
+        String formatInstructions = ServiceOutputParser.outputFormatInstructions(ClassWithNoFields.class);
+
+        assertThat(formatInstructions).isEqualTo("\n" +
+                "You must answer strictly in the following JSON format: {\n" +
+                "}");
     }
 
     static class PersonWithMotherAndFather {


langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
```diff
@@ -130,7 +130,11 @@ private static String jsonStructure(Class<?> structured, Set<Class<?>> visited)
             }
             jsonSchema.append(format("\"%s\": (%s),\n", name, descriptionFor(field, visited)));
         }
-        jsonSchema.delete(jsonSchema.lastIndexOf(","), jsonSchema.lastIndexOf(",")+1);
+
+        int trailingCommaIndex = jsonSchema.lastIndexOf(",");
+        if (trailingCommaIndex > 0) {
+            jsonSchema.delete(trailingCommaIndex, trailingCommaIndex +1);
+        }
         jsonSchema.append("}");
```
Shastick, line 138: Thanks for the fix!

I just ran into the same issue, and was considering fixing it with `String.join()` instead of the gymnastics around the corner cases.

Ie, the whole function would become something like:

```
private static String jsonStructure(Class<?> structured, Set<Class<?>> visited) {
        List<String> descriptions = new ArrayList<>();
        for (Field field : structured.getDeclaredFields()) {
            String name = field.getName();
            if (name.equals("__$hits$__") || java.lang.reflect.Modifier.isStatic(field.getModifiers())) {
                // Skip coverage instrumentation field.
                continue;
            }
            descriptions.add(format("\"%s\": (%s)", name, descriptionFor(field, visited)));
        }

        return "{\n" + String.join(",\n", descriptions) + "\n}";
    }
```

**note**: edited the code
**note2**: the suggested code would likely require an extra newline in your additional test case

LizeRaes:
Great, thank you for the fix!

Shastick (approved):

langchain4j:
@remyohajinwa what is the use case for the class without fields?

Shastick:
> @remyohajinwa what is the use case for the class without fields?

There likely is no real use case, but some quarkus initialisation code (when using the quarkus+langchain4j integration) seems to call the method with a class that has no fields, resulting in an app that cannot start.

The stack-trace in the description happens when you attempt to register an Ai Service (using `
@RegisterAiService` if I recall correctly), if you use the current release of the quarkus-langchain4j integration.

langchain4j:
@Shastick I am aware of Quarkus use case, I was just wondering if there is any other use case, because
```
assertThat(formatInstructions).isEqualTo("\n" +
                "You must answer strictly in the following JSON format: {\n" +
                "}");
```
does not make sense for me

Shastick:
> @Shastick I am aware of Quarkus use case, I was just wondering if there is any other use case, because
> 
> ```
> assertThat(formatInstructions).isEqualTo("\n" +
>                 "You must answer strictly in the following JSON format: {\n" +
>                 "}");
> ```
> 
> does not make sense for me

Indeed, this is not representative of anything meaningful. Should we add some logic to return something else when no fields are present?

On the other hand, if anyone does come up with a use-case where this might be suitable, we may still want to have something along these lines? (perhaps get rid of the newline)

LizeRaes:
I digged a bit deeper, it's not Quarkus-specific, it happens when having a POJO with a List of an Enum (and potentially variations on that theme).

This works:
```
static class Recipe {
            @Description("each step should be described in 6 to 8 words, steps should rhyme with each other")
            private List<String> steps;
}
```

This produces the forementioned error:
```
static class AtomicFeedback {
            @Description("""
                    all relevant topic tags, picked strictly from this list:
                        Examples,
                        RoomEnvironment,
                        Catering
                    """)
            public List<Category> tags;
            @Description("severity of the issue in percentage (0 = no problem, 100 = critical)")
            public String severity;
        }

        enum Tag {
            RoomEnvironment,
            Catering,
            Timing
        }
```

Another thing I noted, possibly related, is that the specific enum values are also not passed as options to the model, hence they stand specified in @Description. I think this would be covered by #909.

langchain4j:
@LizeRaes https://github.com/langchain4j/langchain4j/issues/909 is about streaming-specific return types.
I guess you meant this: https://github.com/langchain4j/langchain4j/issues/725
Also related: https://github.com/langchain4j/langchain4j/issues/916

LizeRaes:
@langchain4j indeed #725 was the one I had in mind. Would be an awesome addition for making AiServices even more powerful!

Issue #928, opened 2024-04-11
[BUG] Infinite retries when max retry is set to 0.


mutayTorre:
**Describe the bug**
I tried setting the max retry to zero to prevent automatic retries when there is an error. It seems setting it to 0, doesn't stop it from retrying, instead it retries till infinity.

**Log and Stack trace**
```
d.l.i.RetryUtils - Exception was thrown on attempt 18 of 0
dev.ai4j.openai4j.OpenAiHttpException: {
  "error": {
    "message": "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_ZuTo6iUVTy0Vuw6R1u09mPgK",
    "type": "invalid_request_error",
    "param": "messages.[5].role",
    "code": null
  }
}

        at dev.ai4j.openai4j.Utils.toException(Utils.java:8)
        at dev.ai4j.openai4j.SyncRequestExecutor.execute(SyncRequestExecutor.java:28)
        at dev.ai4j.openai4j.RequestExecutor.execute(RequestExecutor.java:59)
        at dev.langchain4j.model.openai.OpenAiChatModel.lambda$generate$1(OpenAiChatModel.java:148)
        at dev.langchain4j.internal.RetryUtils$RetryPolicy.withRetry(RetryUtils.java:192)
        at dev.langchain4j.internal.RetryUtils.withRetry(RetryUtils.java:229)
        at dev.langchain4j.model.openai.OpenAiChatModel.generate(OpenAiChatModel.java:148)
        at dev.langchain4j.model.openai.OpenAiChatModel.generate(OpenAiChatModel.java:113)
        at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:130)
        at agents.$Proxy44.chat(Unknown Source)
```

**To Reproduce**
Set the `maxRetries` to 0 when building an `OpenAiChatModel` and then try to send a message to the LLM that will trigger an error. 

**Expected behavior**
It shouldn't retry when the `maxRetries` is set to 0.

**Please complete the following information:**
- LangChain4j version: 0.28.0
- LLM(s) used: GPT_4_1106_PREVIEW
- Java version:  Java 11

**Additional context**

langchain4j:
Thanks a lot for reporting!

tess1o:
[...]

PR #981, opened 2024-04-19
Bugfix/fix 0 retry times

VaderKai:
## Context

#928 
## Change

Fix wireless retry when retry count is 0
Fix the total number of executions to 1+ retry times
if retry times is less than or equal to 0 is 0
## Checklist


jmartisk:
You're turning "attempts" into "retries" here I'd say. If it's called 'attempts', shouldn't 0 mean an error (`IllegalArgumentException`), and 1 to attempt exactly once?
That is, unless we rename the whole thing to "retries" in the code, but it's currently named "attempts".

VaderKai:
> 

so we need do a choice,the utils name is RetryUtils,we can rename the method name or forbidan 0 attempts params,
i think rename is more reasonably,but changes is big,do you thinking about it?

jmartisk:
I would personally keep it as attempts, so 0 means error, and 1 means try exactly once. But it depends what @langchain4j thinks

VaderKai:
> I would personally keep it as attempts, so 0 means error, and 1 means try exactly once. But it depends what @langchain4j thinks

I've changed the number of attempts to a minimum of 1, and 1 for illegal parameters (<=0).

ghost (approved):
@VaderKai thank you!

Issue #956, opened 2024-04-17
[FEATURE] AI Services: Option to dynamically select system message depending on the user message


langchain4j:
There should be a way to dynamically select a system message based on the given user message.
Something similar to https://js.langchain.com/docs/modules/chains/additional/multi_prompt_router

Consider using `TextClassifier` under the hood.

sjivan:
@langchain4j what's the status of this?

langchain4j:
@sjivan nothing so far

PR #993, opened 2024-04-21
[FEATURE] AI Services: Option to dynamically select system message de…

KaisNeffati:
…pending on the user message #956

## Context

https://github.com/langchain4j/langchain4j/issues/956

## Change
There are two annotations, `@RegisterSystemSpecs` and `@SystemSpec`, which facilitate dynamic selection of system messages to accompany user inputs when interfacing with a Large Language Model (LLM). This dynamic selection process involves an initial query to the AI model to determine the most relevant system specification. Should no relevant system be found, a default `SystemSpec` is utilized, ensuring that a response is always generated.

This approach mirrors the methodology used in Langchain, which also involves querying the LLM to select the most appropriate system message. The prompt used for this operation is consistent with the one employed by Langchain.

Using a vector search with a TextClassifier for this feature do not seems to fit to this use case because vector search always returns a result, which may not always align with the relevance required for effective system message selection. In contrast, a direct query to the AI allows for a determination of whether a suitable `SystemSpec` exists for the given user message.

Several tests have been conducted to ensure robustness

When the `@SystemMessage` annotation or a System Message Provider is used in conjunction with `@RegisterSystemSpecs` or a System Spec Provider, the latter are overridden and ignored
 
## Checklist

### DIFF
--- a/docs/docs/tutorials/5-ai-services.md
+++ b/docs/docs/tutorials/5-ai-services.md
@@ -129,6 +129,84 @@ Friend friend = AiServices.builder(Friend.class)
 ```
 As you can see, you can provide different system messages based on a chat memory ID (user or conversation).
 
+
+## @RegisterSystemSpecs
+
+Now, let's explore a more complex scenario. In this case, we'll entrust the AI with the task of selecting the appropriate SystemMessage.
+
+This is accomplished by associating multiple system message specifications, denoted by the `@SystemSpec` annotations, with a single `@RegisterSystemSpecs` annotation.
+
+Here is an example :
+```java
+interface Professor {
+    
+    // Chat method that takes a user message and returns an AI-generated response based on the specified system specifications.
+    // SystemSpec annotations define the context in which the AI should generate responses for physics and math questions.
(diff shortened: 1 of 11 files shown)


langchain4j:
Hi @KaisNeffati, thank you so much!
Unfortunately this is not what was meant. I did a poor job with issue description, this is completely my fault, sorry.

Currently we already have a `systemMessageProvider` which can be configured on the AI Service.
But it accepts only a memory id. One can dynamically select system message based on the user, but not based on the message (topic/etc).
I was thinking about changing the input of the `systemMessageProvider` to an object that will contain both memory id and a user message, probably with the previous messages as well.
When this is possible, than any algorithm can be implemented on top of `systemMessageProvider`. One could use LLM (your solution) or embedding model, or keyword, or something else.

WDYT?

KaisNeffati:
No worries, my friend ! I just reviewed the langchain python source code for [this feature](https://js.langchain.com/v0.1/docs/modules/chains/additional/multi_prompt_router) and rewrote it.

I see it doesn't exhibit the desired behavior. So closing this PR for now 😙

PR #1004, opened 2024-04-23
Fix out-of-bounds error in jsonStructure method

mmanrai:
Added a conditional check to prevent the StringBuilder from attempting to delete a comma at an invalid index, which occurs when no declared fields exists. This prevents the "Range out of bounds" runtime exception.

## Context

Fix out-of-bounds error in jsonStructure method while attempting to delete the last comma.

For example while creating jsonStructure for javax.money.CurrencyUnit, following runtime error is thrown:
`Range [-1, 0) out of bounds for length 2`

## Change

Made deletion conditional.


langchain4j:
Hi @mmanrai thanks a lot, but this was already fixed here: https://github.com/langchain4j/langchain4j/pull/975

Issue #660, opened 2024-02-23
[FEATURE] Return retrieved segments (RAG) when using AiService and streaming


hannesroth88:
I am using a conversational RAG. I would like to return the Source Documents with Metadata. So that I would be able to show them in the Frontend. 

Haven't found a way to easily do so? Is the only way to do it with a full customized chain?
Right now I am using this setup 

```
    interface RagService {
        String answer(String query);
    }
```

```
        QueryTransformer queryTransformer = new CompressingQueryTransformer(chatLanguageModel);

        ContentRetriever contentRetriever = EmbeddingStoreContentRetrieverCustom.builder()
                .embeddingStore(embeddingStore)
                .embeddingModel(embeddingModel)
                .maxResults(5)
                .minScore(0.6)
                .build();

        RetrievalAugmentor retrievalAugmentor = DefaultRetrievalAugmentor.builder()
                .queryTransformer(queryTransformer)
                .contentRetriever(contentRetriever)
                .build();

        ChatMemory chatMemory = chatMemoryService.getChatMemoryMaxTokens(memoryId);
        RagService ragService = AiServices.builder(RagService.class)
                .chatLanguageModel(chatLanguageModel)
                .retrievalAugmentor(retrievalAugmentor)
                .chatMemory(chatMemory)
                .build(); 
```

langchain4j:
HI @hannesroth88 this feature is in high prio, will be added soon.
Right now you can use a custom `ContentInjector` proxy that would intercept the call to the real `ContentInjector` (.e.g `DefaultContentInjector`) and there you can have access to the injected contents. Ugly, but doable :)

hannesro:
I am struggling to pass the Documents back to the parent class where AiService is called.
What I have already done before this feature request was changing how to retrieve the relevant sources in a proxy ContentRetriever:
```
public class EmbeddingStoreContentRetrieverCustom implements ContentRetriever {

    // .... constructors ...

    @Override
    public List<Content> retrieve(Query query) {
        Embedding embeddedText = embeddingModel.embed(query.text()).content();

        // Pass knowledgeId before embeddedText in the findRelevant method
        List<EmbeddingMatch<TextSegment>> relevant = embeddingStore.findRelevantByKnowledgeId(knowledgeId, embeddedText, maxResults, minScore);

        return relevant.stream()
                .map(EmbeddingMatch::embedded)
                .map(Content::from)
                .collect(toList());
    }
```

[...]

PR #1015, opened 2024-04-24
[FEATURE] Return retrieved Contents/TextSegments when using AI Service with RAG

KaisNeffati:
## Context

#660
## Change
If you wish to access the sources (retrieved `Content`s used to augment the message),
you can easily do so now by wrapping the return type in the `Result` class:
```java
interface Assistant {

    Result<String> chat(String userMessage);
}

Result<String> result = assistant.chat("How to do Easy RAG with LangChain4j?");

String answer = result.content();
List<Content> sources = result.sources();
```

## Checklist


langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithRagIT.java
```diff
@@ -210,6 +212,77 @@ void should_use_query_transformer_and_content_retriever(ChatLanguageModel model)
...
+        AssistantWithSources assistant = AiServices.builder(AssistantWithSources.class)
+                .chatLanguageModel(model)
+                .retrievalAugmentor(DefaultRetrievalAugmentor.builder()
+                        .queryTransformer(queryTransformer)
+                        .contentRetriever(contentRetriever)
+                        .build())
+                .build();
+
+        // when
+        assertThrows(IllegalArgumentException.class, () -> assistant.answerWithNoGenericType("Can I cancel my booking?"));
```
langchain4j: minor: it would be nice to do this validation and fail on AI Service creation (eager), not on usage (lazy).

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithRagIT.java
```diff
@@ -210,6 +212,77 @@ void should_use_query_transformer_and_content_retriever(ChatLanguageModel model)
...
+        );
+        assertThat(answer.getAugmentedMessage().getContents().get(0).textSegment().metadata("index")).isEqualTo("3");
+        assertThat(answer.getAugmentedMessage().getContents().get(0).textSegment().metadata("file_name")).isEqualTo("miles-of-smiles-terms-of-use.txt");
+
+
+    }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    void should_use_query_transformer_and_content_retriever_and_through_exception_when_generic_type_is_not_set(ChatLanguageModel model) {
```
langchain4j: query transformer is not necessary for this test, please remove it to simplify the test

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithRagIT.java
```diff
@@ -210,6 +212,77 @@ void should_use_query_transformer_and_content_retriever(ChatLanguageModel model)
...
+                        .queryTransformer(queryTransformer)
+                        .contentRetriever(contentRetriever)
+                        .build())
+                .build();
+
+        // when
+        assertThrows(IllegalArgumentException.class, () -> assistant.answerWithNoGenericType("Can I cancel my booking?"));
+
+    }
+
```
langchain4j: Wondering: should AI Service creation fail also when return type is WIthSources<T>, but no retriever/augmentor is set up?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithRagIT.java
```diff
@@ -210,6 +212,77 @@ void should_use_query_transformer_and_content_retriever(ChatLanguageModel model)
...
     }
 
+    interface AssistantWithSources {
+        WithSources<String> answer(String query);
+        WithSources answerWithNoGenericType(String query);
+    }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    void should_use_query_transformer_and_content_retriever_and_retrieve_sources(ChatLanguageModel model) {
```
langchain4j: query transformer is not necessary for this test, please remove it to simplify the test

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithRagIT.java
```diff
@@ -210,6 +212,77 @@ void should_use_query_transformer_and_content_retriever(ChatLanguageModel model)
...
+                .retrievalAugmentor(DefaultRetrievalAugmentor.builder()
+                        .queryTransformer(queryTransformer)
+                        .contentRetriever(contentRetriever)
+                        .build())
+                .build();
+
+        // when
+        assertThrows(IllegalArgumentException.class, () -> assistant.answerWithNoGenericType("Can I cancel my booking?"));
+
+    }
```
langchain4j: please add a test for the scenario when structured output is required: `WithSources<Pojo>`

langchain4j/src/main/java/dev/langchain4j/service/WithSources.java
```diff
@@ -0,0 +1,17 @@
...
+import lombok.Getter;
+
+/**
+ * Represents a container holding augmented information associated with a response.
+ *
+ * @param <T> The type of the response.
+ */
+@Getter
+@Builder
+public class WithSources<T> {
```
langchain4j: After a second thought, I think we should make this class a bit more generic, to accomodate for other useful information, such as:
- tools used during the call 
- token usage
- etc

(Please do not add these in this PR.)

Any ideas for the good name for the class that can hold that?

langchain4j-core/src/main/java/dev/langchain4j/rag/RetrievalAugmentor.java
```diff
@@ -23,5 +23,5 @@ public interface RetrievalAugmentor {
      * @param metadata    The {@link Metadata} that may be useful or necessary for retrieval and augmentation.
      * @return The augmented {@link UserMessage}.
      */
-    UserMessage augment(UserMessage userMessage, Metadata metadata);
+    T augment(UserMessage userMessage, Metadata metadata);
```
langchain4j: This is a breaking change, is there a way to avoid this?

langchain4j/src/main/java/dev/langchain4j/service/WithSources.java
```diff
@@ -0,0 +1,17 @@
...
+    private T response; // The response associated with the augmented information.
+    private AugmentedMessage augmentedMessage; // Wrapper for the augmentation details.
```
langchain4j:
```suggestion
    private List<Content> retrievedContents;
```

WDYT?

ghost (changes requested):
@KaisNeffati thank you a lot for your contribution!

langchain4j-core/src/main/java/dev/langchain4j/rag/RetrievalAugmentor.java
KaisNeffati: The purpose of this is to retrieve the contents used for augmentation.

I have two solutions in mind:

1. **First solution:** Add another method in the `RetrievalAugmentor` which returns another object containing the `UserMessage` and the list of contents.
2. **Second solution:** Add the list of contents (`List<TextSegments>`) in the `UserMessage` class.

I noticed another list of contents in the `UserMessage`, but it's only used to specify the type of content used for augmentation.

What do you suggest to make retrieving the `UserMessage` with the list of content possible in the `DefaultAiServices` class?

langchain4j/src/main/java/dev/langchain4j/service/WithSources.java
KaisNeffati: `ResponseMessage` or `ResponsePrompt`

Well when we talk about framework like Spring or the javax package, they provide powerful tools for handling HTTP requests and responses. One key concept is how they handle responses.

In Spring, for example, ResponseEntity is a class that encapsulates an HTTP response, including the response body and additional metadata such as headers and status codes. It's a neat way to package up all the necessary information about a response in one object.

Similarly, in the javax package, there's the ResponseBuilder class. This class allows you to construct an HTTP response step by step, setting things like the status code, headers, and entity (the response body). It's a flexible and convenient way to craft custom responses 

Now, let's bring this analogy back to our prompt usecase. Just as ResponseEntity and ResponseBuilder provide a structured way to handle HTTP responses. ResponseMessage class aims to provide a structured way to handle prompts response

By using ResponseMessage (or ResponsePrompt), we can create responses that not only contain the text of the prompt itself but also include additional contextual information, just like how ResponseEntity includes metadata with the response body. This makes our response handling more organized, easier to manage, and more adaptable to different scenarios
```

ResponseMessage responseMessage = ResponseMessage.create("How can I assist you today?")
    .contents(...)
    .tools(...)
    .tokens(...)
    .build();
```

I can include this in a seperated PR if needed

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithRagIT.java
KaisNeffati: Absolutely! I agree, adopting a straightforward solution like that would indeed be beneficial. Otherwise, we might encounter numerous complaints from users about the `WithSources` object lacking contents or other augement information

langchain4j-core/src/main/java/dev/langchain4j/rag/RetrievalAugmentor.java
langchain4j: I assume that in the long term, regardless of this change, we will have to make internals of `UserMessage` more fine-grained, e.g. having user message and retrieved contents separate (solution 2). I foresee that LLM APIs in the future will separate user message from supporting (retrieved) content to be able to adjust/fine-tune their LLMs better. Some LLM providers [already do this](https://docs.cohere.com/reference/chat) (see `documents` parameter). But this is pretty big change, and introduces more complexity, so I would better postpone it untill it will be really required.

I guess for now we can go with the first solution, something like this?

```java 
public interface RetrievalAugmentor {

    @Deprecated // let's deprecate this
    UserMessage augment(UserMessage userMessage, Metadata metadata); // existing API

    default AugmentationResult augment(AugmentationRequest augmentationRequest) { // new API
        UserMessage augmented = augment(augmentationRequest.chatMessage, augmentationRequest.metadata);
        return new AugmentationResult(augmented, null);
    }
}

// introduce request/result classes to make it more extendable in the future
public class AugmentationRequest {

    ChatMessage chatMessage; // instead of UserMessage, this will be useful for issue #793
    Metadata metadata;
}

public class AugmentationResult {

    ChatMessage augmentedChatMessage;
    List<Content> retrievedContents;
}
```

This feature will not work for existing custom implementations of `RetrievalAugmentor`, but I guess it is fine as there are most probably not many such users and this class is explicitly marked `@Experimental`.

Most users are probably using `DefaultRetrievalAugmentor`, where we will have a proper implementation of the new method.

WDYT?

langchain4j/src/main/java/dev/langchain4j/service/WithSources.java
langchain4j: I am not sure about these names, they are not really responses, and not messages/prompts.

These object can contain much more:
- Answer/Response itself (`String` or `CustomPOJO` or smth else)
- Retrieval:
  - `List<Content> retrievedContents`
  - Used queries (not sure if needed)
  - Used retrievers (not sure if needed)
- LLM calls:
  - `TokenUsage` (for all calls combined and/or for each call separately)
  - `FinishReason` (of the last call)
  - Model parameters: `modelName`/`temperature`/etc?
  - List of all calls with requests and responses (not sure if needed)?
- Tool executions:
  - Executed tools and their inputs/outputs

I am having a hard time to find a good name right now, maybe it is an indicator of a bad design, or maybe I need to go and rest a bit 😆 

Have a nice weekend!

KaisNeffati: Taking a breather and diving into some meditation is the perfect recipe for this naming problem 😙

As for me, I'm rolling with 'WithSources' for now. But if you've got more brainwaves brewing, feel free to open another issue, and we'll keep this brainstorm party going! Yep, it's definitely more than just a response 🤔

langchain4j-core/src/main/java/dev/langchain4j/rag/RetrievalAugmentor.java
KaisNeffati: I'll include the ChatMessage in a separate PR since it requires numerous modifications

langchain4j: @KaisNeffati what is the change exactly? Is it breaking?

KaisNeffati: To utilize `ChatMessage` within both `AugmentationRequest` and `AugmentationResult`, the interface should be adjusted as follows:
```

    @Deprecated
    UserMessage augment(UserMessage userMessage, Metadata metadata);

    UserMessage augment(ChatMessage chatMessage, Metadata metadata);

    /**
     * Augments the provided {@link AugmentationRequest} with retrieved content.
     *
     * @param augmentationRequest The {@link AugmentationRequest} containing the user message and metadata.
     * @return The {@link AugmentationResult} containing the augmented user message.
     */
    default AugmentationResult augment(AugmentationRequest augmentationRequest) { // new API
        UserMessage augmented = augment(augmentationRequest.getChatMessage(), augmentationRequest.getMetadata());
        return AugmentationResult.builder()
                .augmentedUserMessage(augmented)
                .build();
    }
```

This adjustment necessitates modifying the contentInjector to accept:

`    UserMessage inject(List<Content> contents, ChatMessage chatMessage);
`
etc...

So for this PR i kept using the UserMessage , And i'll include the use of ChatMessage in another PR

langchain4j:
Hi @KaisNeffati, do you mind if I will do and push some changes?

KaisNeffati:
Fell free to do it @langchain4j

langchain4j:
@KaisNeffati could you please add me as collaborator to your repo? I can't push to this branch/PR :(

KaisNeffati:
Done

langchain4j:
@KaisNeffati thank you! I have pushed my changes, could you please review? I hope now it is a bit more generic (Result instead of WithSources). I have also changed UserMessage to ChatMessage right away in all places  to avoid doing breaking changes one more time in the near future.

langchain4j-core/src/main/java/dev/langchain4j/rag/DefaultRetrievalAugmentor.java
```diff
@@ -247,8 +264,8 @@ private static void log(Map<Query, Collection<List<Content>>> queryToContents, L
                         .collect(joining("\n")));
     }
 
-    private static void log(UserMessage augmentedUserMessage) {
-        log.trace("Augmented user message: " + escapeNewlines(augmentedUserMessage.singleText()));
+    private static void log(ChatMessage augmentedChatMessage) {
+        log.trace("Augmented chat message: " + escapeNewlines(augmentedChatMessage.text()));
```
KaisNeffati: Too Minor : Placeholder instead of concatenation
` log.trace("Augmented chat message: {}", escapeNewlines(augmentedChatMessage.text()));
`

KaisNeffati (commented):
Just a minor comment below !

And Yeah, `Result` is a very wise choice. It confirms what I had in mind initially when I thought about 'ResponsePrompt' or 'ResponseMessage'. I'm glad to hear that you've found the right name 👏

Atherwise , all is good for me

langchain4j:
@KaisNeffati thanks! Then I am merging it.

Issue #938, opened 2024-04-15
[FEATURE] Allow passing Image or collections of Image into AI Service method


langchain4j:
It would be nice to allow passing `Image` objects or collections of them into AI Service method.
They should be automatically appended to the `UserMessage` and sent to the LLM.
Need to discuss API options before implementation.

langchain4j:
Consider supporting `ImageContent` as well (to be able to specify detail level)

mmacphail:
Could it be as simple as something like this in [DefaultAiServices.prepareUserMessages()](https://github.com/langchain4j/langchain4j/blob/main/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java#L255):

- For each parameter of type ImageContent, we take the parameter as is and add it to the generated UserMessage by the prompt (or we create one if there are no other parameters than ImageContent).

I could be a very simple first step.

langchain4j:
@mmacphail, this sounds reasonable. 

This should work identical for both `Image` and `ImageContent` types.

Currently, if method has more than 1 paramerter, all paramerters must be annotated, to avoid ambiguity.
I guess we should keep it this way. So, if other paramerters are present, `Image`/`ImageContent` should be annotated wih `@UserMessage`.
If there is only 1 paramerter, and it's type is `Image`/`ImageContent`, it will be interpreted as a content for `UserMessage`.

You can check `AiServicesUserMessageConfigTest` and `AiServicesSystemAndUserMessageConfigsTest`, there are lots of combinations. I am afraid the complexity is growing exponentially...

We also need to make sure that images are still inside the `UserMessage` after `RetrievalAugmentor` and `outputFormatInstructions` manipulations.

akissaba:
Hello everyone! This feature would be really interesting! I am also developing a POC, where I need to send an image analysis performed by an LLM and I am considering using RAG for the LLM to respond with greater precision.

To take advantage of the existing **_UserMessage_** annotation, I imagined the possibility of implementing some additional (optional) attributes for each type of content and using **_PromptTemplate_** to identify the variables in the value of these attributes. It would look something similar to the abstracted code snippet:

```java
@AiService(
        wiringMode = EXPLICIT,
        chatModel = CHATMODEL_BEANNAME,
        chatMemoryProvider = CHATMEMORYPROVIDER_BEANNAME,
        contentRetriever = CONTENTRETRIEVER_BEANNAME,
        tools = {
            TOOLS_BEANNAME
        }
)
[...]

PR #1028, opened 2024-04-27
[FEATURE] Allow passing Image or collections of Image into AI Service method

mmacphail:
## Issue
https://github.com/langchain4j/langchain4j/issues/938

## Change
Add the ability to pass Image and ImageContent to AIServices.
Please note that I only added unit tests so far, as I'm looking for comments on style before continuing further.

## General checklist


langchain4j/src/test/java/dev/langchain4j/service/AiServicesUserMessageConfigTest.java
```diff
@@ -47,6 +53,18 @@ interface AiService {
         @UserMessage("What is the {{it}} of {{country}}?")
         String chat7(@V("it") String it, @V("country") String country);
 
+        @UserMessage("Count the number of cars in this image")
+        String chat8(@UserMessage ImageContent imageContent);
```
ashimoon, line 57: Perhaps consider returning `int` here since the expected output should always be an int.

geoand:
I believe this is very interesting. 

@langchain4j what is your take on this? Do you have something else in mind or is does this match what you were planning to do in the future?

langchain4j:
@geoand seems good, but need to test all the possible combinations before merging: single image, multiple images, image+text, multiple images+text, etc

mmacphail:
I can expand the test, I will first work on a list of all needed combination to support.

geoand:
Makes sense, thanks for the explanation

sang5338 (approved):

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -99,7 +102,15 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
 
                         // TODO give user ability to provide custom OutputParser
                         String outputFormatInstructions = outputFormatInstructions(method.getReturnType());
-                        userMessage = UserMessage.from(userMessage.text() + outputFormatInstructions);
+                        if(userMessage.contents().stream().filter(content -> content instanceof TextContent).count() > 1) {
+                            throw illegalConfiguration("Error: The method '%s' has multiple text contents. Please use only one.", method.getName());
+                        }
+                        if(userMessage.contents().size() > 1 && !(userMessage.contents().get(0) instanceof TextContent)) {
+                            throw illegalConfiguration("Error: The first content should be text content.", method.getName());
```
langchain4j, line 127: what is the reason for this requirement?

christianmenz:
New here and just looking into the API, but I was wondering whether we could also support annotations instead of directly using ImageContent and Image. Something like

    Description getDescription(@V("parameter1") String p1, @ImageUrl String imageUrl);

langchain4j:
@christianmenz I am afraid that would make it even more complicated...  Image should be easily created from the url: `Image.from(url)`

langchain4j:
Hi @mmacphail, I've played a bit with the proposed API and added extra examples/tests:

```java
        String chat12(Image image);

        String chat13(@UserMessage String userMessage, @UserMessage Image image);

        String chat14(@UserMessage Image image, @UserMessage String userMessage);
```

```java
@Test
    void test_user_message_configuration_12() {

        // given
                .build();

        // when-then
        assertThat(aiService.chat12(image))
                .containsIgnoringCase("Berlin");
        verify(chatLanguageModel).generate(singletonList(userMessage(imageContent)));
    }

    @Test
    void test_user_message_configuration_13() {

        // given
                .build();

        // when-then
        assertThat(aiService.chat13("What's on this image?", image))
                .containsIgnoringCase("Berlin");
        verify(chatLanguageModel).generate(singletonList(userMessage(
                TextContent.from("What's on this image?"),
                imageContent
        )));
    }

    @Test
    void test_user_message_configuration_14() {

        // given
                .build();

        // when-then
        assertThat(aiService.chat14(image, "What's on this image?"))
                .containsIgnoringCase("Berlin");
        verify(chatLanguageModel).generate(singletonList(userMessage(
                imageContent,
                TextContent.from("What's on this image?")
        )));
    }
```

(Tests 12 and 14 are failing now).

WDYT?

langchain4j:
Hi @mmacphail, do you plan to finish this PR? Thanks!

langchain4j:
We need to do the same for `Video` and `Audio` introduced in https://github.com/langchain4j/langchain4j/pull/1464
(Could be another PR)

langchain4j:
Hi @mmacphail, could you please finish this PR? Thank you!

langchain4j:
It seems that @emivoz will work on this

mmacphail:
Hey ! Sorry guys.
Had a lot of personal and work stuff going on. I had no access for my github account for a while.
It's not the best time to continue contributing right now.
Take care. @emivoz tell me if you need anything.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
singularity-sg: I have the same question - I don't think order matters here

peace317:
Are there any updates to this? It seems like nobody is working on this really

miracle-bean:
#3420
I code another method

Issue #1026, opened 2024-04-26
[BUG] FileSystemDocumentLoader#loadDocuments has a potential issue when using ApacheTikaDocumentParser


vprudnikov:
**Describe the bug**
`FileSystemDocumentLoader#loadDocuments` has a potential issue when using ApacheTikaDocumentParser
As shown in the `langchain4j-examples`, this line of code doesn't work as expected:
`List<Document> documents = loadDocuments(directoryPath, new ApacheTikaDocumentParser());`

**To Reproduce**
Try to load at least 2 documents using the provided example.

**Expected behavior**
Each document is loaded independently.

**Current behavior**

* The first document contains its parsed content
* The second document contains the content of the first document plus its content
* The third document contains the content of the previous documents plus its content
* ... and so on

**Please complete the following information:**
- LangChain4j version: 0.30.0
- LLM(s) used: NA
- Java version: 21
- Spring Boot version (if applicable): NA

**Additional context**
It looks like the ApacheTikaDocumentParser is a stateful object. That's why it cannot be reused across all documents.
Probably, a better solution would be using a Supplier like so:
`List<Document> documents = loadDocuments(directoryPath, () -> new ApacheTikaDocumentParser());`
or
`List<Document> documents = loadDocuments(directoryPath, filePath -> new ApacheTikaDocumentParser());`
As a result, a new fresh instance of the parser will be provided each time.

langchain4j:
@vprudnikov thank you so much for reporting!

langchain4j:
Fixed by https://github.com/langchain4j/langchain4j/pull/1031

PR #1031, opened 2024-04-28
[BUG] Introduce parser supplier support in FileSystemDocumentLoader

KaisNeffati:
## Issue

https://github.com/langchain4j/langchain4j/issues/1026

## General checklist

### DIFF
--- a/document-parsers/langchain4j-document-parser-apache-tika/src/main/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParser.java
+++ b/document-parsers/langchain4j-document-parser-apache-tika/src/main/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParser.java
@@ -12,6 +12,7 @@ import org.apache.tika.sax.BodyContentHandler;
 import org.xml.sax.ContentHandler;
 
 import java.io.InputStream;
+import java.util.function.Supplier;
 
 import static dev.langchain4j.internal.Utils.getOrDefault;
 import static dev.langchain4j.internal.Utils.isNullOrBlank;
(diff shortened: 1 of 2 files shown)


langchain4j:
@KaisNeffati thanks a lot!

I am not sure the proposed solution will solve the issue, as:
- Current users already use existing methods of `FileSystemDocumentLoader` and they will have to change their code to use new methods with suppliers in order to fix the issue
- New users will not necessarily find and use the "correct" new methods with suppliers, so the wrong behavior might still persist for them and be unobvious
- All other parsers are stateless, so having `ApacheTikaDocumentParser` statefull is counter-intuitive

Instead, I would propose to make `ApacheTikaDocumentParser` stateless:
- the creation of `Parser`, `ContentHandler` and other components should be done in `parse` method instead of the constructor
- a new constructor should be added. this ctor should accept suppliers of `Parser`, `ContentHandler` and other tika components
- existing non-default constructor should be marked as `@Deprecated` with a comment to use ctor with suppliers instead if user wants to use this parser for multiple files

WDYT? Thanks!

KaisNeffati:
Yeah ! that's a better approach , I'll update the PR accordingly !

Well seen !

document-parsers/langchain4j-document-parser-apache-tika/src/main/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParser.java
```diff
@@ -48,22 +53,45 @@ public ApacheTikaDocumentParser() {
...
+        this.metadataSupplier = metadataSupplier;
+        this.parseContextSupplier = parseContextSupplier;
     }
 
     // TODO allow automatically extract metadata (e.g. creator, last-author, created/modified timestamp, etc)
 
     @Override
     public Document parse(InputStream inputStream) {
         try {
+            Parser parser = getOrDefault(parserSupplier, () -> DEFAULT_PARSER_SUPPLIER).get();
```
langchain4j: minor: I would call `getOrDefault` in the constructor instead and do `parserSupplier.get()` here (same for the rest of fields)

document-parsers/langchain4j-document-parser-apache-tika/src/test/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParserTest.java
```diff
@@ -40,7 +41,7 @@ void should_parse_doc_ppt_and_pdf_files(String fileName) {
     })
     void should_parse_xls_files(String fileName) {
 
-        DocumentParser parser = new ApacheTikaDocumentParser(new AutoDetectParser(), null, null, null);
+        DocumentParser parser = new ApacheTikaDocumentParser(AutoDetectParser::new, null, null, null);
```
langchain4j, line 44: why is this change required?

ghost (commented):
@KaisNeffati thank you, good job! Just a few minor comments

document-parsers/langchain4j-document-parser-apache-tika/src/main/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParser.java
KaisNeffati: Done !

document-parsers/langchain4j-document-parser-apache-tika/src/test/java/dev/langchain4j/data/document/parser/apache/tika/ApacheTikaDocumentParserTest.java
KaisNeffati: Well spotted, not required ! 👍

ghost (approved):
@KaisNeffati thank you! Good job!

Issue #1042, opened 2024-04-30
[FEATURE] Migrate from Gson to Jackson


langchain4j:
LangChain4j should use Jackson instead of Gson to better allign with Quarkus and Spring Boot ecosystem.

dliubarskyi:
Modules left to migrate:
- `langchain4j-core`
- `langchain4j`
- `langchain4j-google-ai-gemini` ?
- `langchain4j-jlama` (https://github.com/langchain4j/langchain4j/pull/1870)
- remove reference from `langchain4j-embeddings`

glaforge:
We should also migrate `google-ai-gemini`, as I haven't started yet to work on the new unified / joint module.

dliubarskyi:
@glaforge created https://github.com/langchain4j/langchain4j/issues/2596

PR #1043, opened 2024-04-30
Migrate Anthropic from Gson to Jackson

ghost:
## Issue
https://github.com/langchain4j/langchain4j/issues/1042

## Change
This is a first step for migrating LC4j from Gson to Jackson.
This PR migrates Anthropic module only.

Internal classes (which have to be public because Quarkus extension depends on them) are moved into `internal` packages.
@geoand I hope it is not a big deal.

## General checklist


langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java
```diff
@@ -15,29 +16,35 @@
...
-    static final Gson GSON = new GsonBuilder()
-            .setFieldNamingPolicy(LOWER_CASE_WITH_UNDERSCORES)
-            .setObjectToNumberStrategy(LONG_OR_DOUBLE)
-            .setPrettyPrinting()
-            .create();
+    private static final ObjectMapper OBJECT_MAPPER = createObjectMapper();
+
+    private static ObjectMapper createObjectMapper() {
+        ObjectMapper objectMapper = new ObjectMapper();
+        objectMapper.setPropertyNamingStrategy(SNAKE_CASE);
```
langchain4j: @geoand in Quarkus extension mixins are used to define (de)serialization stratigies, can you switch to the global settings on `ObjectMapper` level? Or should I annotate every api model class instead?

geoand:
> Internal classes (which have to be public because Quarkus extension depends on them) are moved into internal packages.
@geoand I hope it is not a big deal.

Not a problem

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java
geoand: The reason I used mixins was because I could not change the classes themselves :). 

Most of the uses of mixins would be hard to replicate on the `ObjectMapper` configuration

langchain4j: As far as I understand, `AnthropicContentMixin` and `AnthropicContentMixinBuilderMixin` can be removed as I removed builder from `AnthropicContent` in this PR.

`AnthropicMessageMixin` can probably be also removed as I have made `AnthropicMessage`'s fields public in this PR and added default ctor.
`AnthropicMessageSerializer` seems to be not used.

And `AnthropicRoleDeserializer`, `AnthropicRoleMixin` and `AnthropicRoleSerializer` can probably be replaced with `objectMapper.configure(WRITE_ENUMS_TO_LOWERCASE, true);`.

So then all mixins can be removed I guess?

geoand: Yeah, that likely can with these changes

langchain4j:
@ThomasVitale do you mind taking a quick look? Thank you!

ThomasVitale:
It looks good, thanks @langchain4j! I only have one comment about the `ObjectMapper` configuration. 

Centralising options like "lower snake case" and "include non null" when defining the `ObjectMapper` in `DefaultAnthropicClient` is quite convenient. Considering the perspective of frameworks providing their own implementation of the `AnthropicClient`, that would mean that the frameworks would have to make sure to include the same options in their own `ObjectMapper` instance. Those options would have to be provided manually by inspecting the default client code and by ensuring they are kept in sync. Also, if that `ObjectMapper` instance is shared across the application , there might be incompatibilities with other serialisable objects (especially with regard to the naming convention).

An alternative approach could be to annotate each API object with the relevant annotations (like `@JsonInclude(JsonInclude.Include.NON_NULL)` and `@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)`). It would require a bit more code, but each API object would be self-contained from a serialisation configuration point of view, it would be self-documenting and also testable by itself. As a result, custom implementations of `AnthropicClient` wouldn't have additional requirements on how to configure an `ObjectMapper` instance or incompatibilities. What do you think?

Also, @geoand what's your perspective from Quarkus?

geoand:
> An alternative approach could be to annotate each API object with the relevant annotations

This would be the safest option.

Another option would be for frameworks to pass in an ObjectMapper to the library (probably via some SPI) and then the library would configure whatever options are necessary. It would have to be from the Javadoc that this ObjectMapper should not be the "global" the framework uses, but a LangChain4j specific one

langchain4j:
@ThomasVitale @geoand good points, thank you! Will change it to annotations on each class.

Issue #967, opened 2024-04-18
[FEATURE] Set User-Agent in all modules to LangChain4j


langchain4j:
Default `User-Agent` in all modules should be `LangChain4j`.

roytruelove:
Long time coder, first time contributor.  Would like to take this on, if that works?

langchain4j:
@roytruelove sure, go ahead!
Could you please start (a draft) with just a few providers?
Thanks!

roytruelove:
The following models could not have their User Agent updated because they don't provide access to the underlying HTTP service:
- AbstractBedrockChatModel (and subclasses)
- AbstractBedrockEmbeddingModel (and subclasses)
- Qwen*
- VertexAiEmbeddingModel
- VertexAiGeminiChatModel
- VertexAiGeminiStreamingChatModel

zz-zhi54:
Hi, do you need any further assistance? Please contact me if so.

kuber-2003:
Hi, I'd like to work on the remaining modules that still don't have the User-Agent header set. Could you confirm this is still open and which providers are pending? Happy to start with a draft PR covering 2–3 modules.

AamirrH:
i would love to work on this, if this issue is still not fixed

Oseas031:
I've submitted a first batch PR (#5707) adding `User-Agent: LangChain4j` to the **Ollama** and **Anthropic** modules.

Happy to work on additional modules in follow-up PRs.

haitow-archive:
Hi, I'd like to help with the remaining providers for #967.

I saw PR #5707 already covers Ollama and Anthropic. If this is still open, I can take the next small batch of still-uncovered modules in a draft PR.

Please let me know if there are any preferred providers to start with, or if someone is already working on the next batch.

yyyCode:
I'd like to help move this forward. Building on #5707 (Ollama + Anthropic), I'm taking the next batch of modules that use the shared `dev.langchain4j.http.client.HttpClient` abstraction but still don't set a default `User-Agent`:

- **Cohere** (`CohereClient`)
- **Jina** (`JinaClient`)
- **Voyage AI** (`VoyageAiClient`)

PR incoming shortly. Happy to follow up with more modules afterwards.

CuriousLinYu:
Hi, I've opened PR #6105 that adds a default `User-Agent: LangChain4j` header to 12 HTTP client modules:

- langchain4j-anthropic
- langchain4j-chroma
[...]

PR #1046, opened 2024-04-30
967 set useragent to lang chain4j

roytruelove:
## Issue
[Set User Agent to 'LangChain4J'](https://github.com/langchain4j/langchain4j/issues/967)

## Change
Added the User Agent wherever it was accessible

## General checklist

- [N/A] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [N/A] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)

- [N/A] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

- [N/A] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [N/A] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

- [N/A] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j


roytruelove:
A few questions for y'all

1) I've cut and pasted the same code to (so far) 3 of the OkHttpClient.  Continuing this way doesn't feel right but there's not a natural place (that I've found) where I could create a singleton Interceptor and/or a make a constant for the User Agent value.  Maybe a util class in `dev.langchain4j.core` but that doesn't have OkHttp as a dependency.

It seems like there are several other tickets discussing HTTP client including getting rid of OkHttp3 as the default, so perhaps the cut and paste is fine for the time being before larger changes take place?

2) There are no current tests for the classes I'm changing, so I'm wondering if as part of this change I'd need to create full coverage for all of the client classes?

langchain4j/src/main/java/dev/langchain4j/code/Judge0JavaScriptEngine.java
```diff
@@ -25,6 +27,13 @@ class Judge0JavaScriptEngine implements CodeExecutionEngine {
         this.apiKey = apiKey;
         this.languageId = languageId;
         this.client = new OkHttpClient.Builder()
+                .addInterceptor(new Interceptor() {
```
langchain4j: please skip Judge0 for the moment, there is another PR that moves it to other place

langchain4j:
Hi @roytruelove,

> I've cut and pasted the same code to (so far) 3 of the OkHttpClient. Continuing this way doesn't feel right but there's not a natural place (that I've found) where I could create a singleton Interceptor and/or a make a constant for the User Agent value. Maybe a util class in dev.langchain4j.core but that doesn't have OkHttp as a dependency.
> It seems like there are several other tickets discussing HTTP client including getting rid of OkHttp3 as the default, so perhaps the cut and paste is fine for the time being before larger changes take place?

Yes, I guess copy/paste is OK for now.

> There are no current tests for the classes I'm changing, so I'm wondering if as part of this change I'd need to create full coverage for all of the client classes?

Ideally it would be nice to have a test. But if it takes too long, let's just test manually and see that existing tests are green.
If it is quick to do though, I would do it and copy/paste it everywhere.

roytruelove:
Will do, thank you

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/InternalVertexAiHelper.java
roytruelove, line 1: Modeled after InternalOpenAiHelper.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
```diff
@@ -95,8 +96,11 @@ public VertexAiEmbeddingModel(String endpoint,
         );
 
         try {
+            // TODO switch to use InternalVertexAiHelper.defaultPredictionServiceSettingsBuilder() once the library
```
roytruelove, line 99: Interestingly this uses a beta verson of `PredictionServiceSettings` so it could not use the new `InternalVertexAiHelper` class.

roytruelove:
Core code is complete, looking to see where I can add some tests

roytruelove (commented):
Added tests for new methods.  Unfortunately it is a bigger lift to test that the HTTP calls contain the correct user agent, without adding mocks / inject-able HTTP clients.  Please let me know your thoughts on this.

Other than that the PR is ready for review.

roytruelove:
Apologies - I marked this as "ready" when it should still be in draft..

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -45,7 +45,7 @@ class InternalAzureOpenAiHelper {
 
     private static final Logger logger = LoggerFactory.getLogger(InternalAzureOpenAiHelper.class);
 
-    public static final String DEFAULT_USER_AGENT = "langchain4j-azure-openai";
+    public static final String DEFAULT_USER_AGENT = "LangChain4j";
```
langchain4j, line 48: @jdubois would it be ok if we change this or better not?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/InternalOpenAiHelper.java
```diff
@@ -35,14 +36,25 @@ public class InternalOpenAiHelper {
     static final String OPENAI_DEMO_URL = "http://langchain4j.dev/demo/openai/v1";
 
 
-    static final String DEFAULT_USER_AGENT = "langchain4j-openai";
```
langchain4j, line 38: @jdubois same question here

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
jdubois: @langchain4j I'm using it for reporting purposes on the Azure OpenAI side: if you do this, we'll lose the history and we'll need to update to use the new key. Also, the idea was to differentiate "LangChain4J Azure OpenAI" and "LangChain4J OpenAI", as both are able to use Azure OpenAI and I would like to know which LangChain4J module people are using (so I can prioritise work).

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/InternalOpenAiHelper.java
jdubois: @langchain4j and same answer as above :-)

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
langchain4j: ok, no worries

langchain4j: @roytruelove could you please rollback changes to Azure OpenAI and OpenAI?

Issue #983, opened 2024-04-19
[FEATURE] UX: handle tool misconfiguration better


langchain4j:
In the following case, when tools is a class, not an object, we should do a better job handling this situation.
```
AiServices.builder(Assistant.class)
                .tools(Tools.class)
                .build();
```
Exception should be thrown if (one of) the provided object(s) is a `Class`.

anunnakian:
If we try to create an object when we have a Class as a parameter. Do we have to handle the following case too ?
```
AiServices.builder(Assistant.class)
                .tools(Tools.class, tools2, tools3)
                .build();
```

langchain4j:
@anunnakian after a second thought, I am not sure it is a good idea to introduce another reflexion usage, I would rather just throw an exception if the provided object is a `Class`

anunnakian:
Got it! could you assign this issue to me, I'll make a PR at the end of the day

langchain4j:
Another check to implement: if provided object/objects in `AiServices.builder().tools(...)` don't have `@Tool`-annotated methods, exception should be thrown

dliubarskyi:
Implemented in https://github.com/langchain4j/langchain4j/pull/1054

PR #1054, opened 2024-05-03
Handle misconfigured tools

anunnakian:
## Issue
https://github.com/langchain4j/langchain4j/issues/983

## Change
When we configure a tool by using a `Class` instance, we get an exception because only object instances are allowed.

## General checklist


langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -316,6 +316,10 @@ public AiServices<T> tools(List<Object> objectsWithTools) { // TODO Collection?
         context.toolExecutors = new HashMap<>();
 
         for (Object objectWithTool : objectsWithTools) {
+            if (objectWithTool instanceof Class) {
+                throw illegalConfiguration("Tools list must be an object instance, not a Class.");
```
```suggestion
                throw illegalConfiguration("Tool '%s' must be an object, not a class", objectWithTool);
```

ghost (dismissed):
@anunnakian thank you!

Issue #232, opened 2023-10-16
[FEATURE] Support queries on DB's e.g. SQLite and MySQL among others like https://python.langchain.com/docs/use_cases/qa_structured/sql


jnahansen:
**Is your feature request related to a problem? Please describe.**
I want to query SQLite and MySQL databases using an AI Chatbot but don't want to integrate with Python

**Describe the solution you'd like**
Desire something like the mechanisms provided by https://python.langchain.com/docs/use_cases/qa_structured/sql

**Describe alternatives you've considered**
Python integration with LangChain

dylankolson:
+1 would love to see this feature - any roadmap on this?

langchain4j:
Can be introduced as a custom implementation of `ContentRetriever`, similar to `Neo4jContentRetriever`

langchain4j:
Implemented in https://github.com/langchain4j/langchain4j/pull/1056

PR #1056, opened 2024-05-06
Experimental: RAG: SQL database content retriever

ghost:
## Issue
https://github.com/langchain4j/langchain4j/issues/232

## Change
An experimental `SqlDatabaseContentRetriever` has been added.

Simplest usage example:
```java
ContentRetriever contentRetriever = SqlDatabaseContentRetriever.builder()
    .dataSource(dataSource)
    .chatLanguageModel(openAiChatModel)
    .build();
```
In this case SQL dialect and table structure will be determined from the `DataSource`.

But it can be customized:
```java
ContentRetriever contentRetriever = SqlDatabaseContentRetriever.builder()
    .dataSource(dataSource)
    .sqlDialect("PostgreSQL")
    .databaseStructure(...)
    .promptTemplate(...)
    .chatLanguageModel(openAiChatModel)
    .maxRetries(2)
    .build();
```

See `SqlDatabaseContentRetrieverIT` for a full example.

## General checklist

- [X] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)


langchain4j:
cc @dandreadis @geoand

geoand:
Also cc @jmartisk

jmartisk:
Looks cool, we have something like it in the Quarkus samples, but this looks more general
Did you try with various models? I remember trying to use codellama and mistral for the sample that we have, but it just didn't work properly (only GPT produced reasonable queries).

langchain4j:
@jmartisk yes, I played with `gpt-3.5-turbo`, `open-mistral-7b` (via Mistral API) and `tinydolphin` (via Ollama).
For those integration  tests that I have, `gpt-3.5-turbo` works most of the time, `open-mistral-7b` in like 70% cases and `tinydolphin` only with simple cases. I can imagine it will work much better with https://ollama.com/library/sqlcoder or other text-to-SQL models.

Issue #199, opened 2023-09-28
[FEATURE] Observability


nooone:
**Is your feature request related to a problem? Please describe.**
Right now it is fun to experiment with different function descriptions, system messages, templates etc. but using AiServices we only get to see the end result, not the process of how the LLM got to that message. There is only logging that can be enabled, but that is meant for developers. It would be nice to have a listener that receives the ongoing messages from the LLM and be able to show that in a UI to the users, so that the inputs can be improved.

**Describe the solution you'd like**
The AiServices builder could receive a ChatMessageListener that gets the individual messages from conversations in simple text form.

**Describe alternatives you've considered**
Logging is currently an option, but it's meant for developers, not users.

Does this make sense? There could be different kind of listeners as well, maybe a RetrieverListener which reports about individual TextSegments being found, or a CostListener which reports the cost response from the conversation.
I think this is simple enough for me to implement as well, if this feature request gets accepted and there is no other alternative that I don't know of right now.

Edit: I think this is kind of what langchain python does with Callbacks: https://python.langchain.com/docs/modules/callbacks/

langchain4j:
@nooone this feature is in our pipeline with the high priority, will update you once we have it.

jettro:
Would be so cool if we could somehow connect it with:
https://github.com/Arize-ai/phoenix

langchain4j:
Related:
https://github.com/langchain4j/langchain4j/issues/246
https://github.com/langchain4j/langchain4j/issues/214

Glarme:
What's the progress on this feature? Is anyone currently working on it

langchain4j:
@Glarme I will be working on it this week

langchain4j:
Related: https://github.com/langchain4j/langchain4j/discussions/883

langchain4j:
Related: https://github.com/langchain4j/langchain4j/discussions/881

ruizrube:
https://langfuse.com/ is also another open source LLM monitoring tool

langchain4j:
Related: https://github.com/langchain4j/langchain4j/issues/1349

langchain4j:
Related (Spring Boot events): https://github.com/langchain4j/langchain4j/issues/1566

ruizrube:
> https://langfuse.com/ is also another open source LLM monitoring tool

A Langchain4J integration with Langfuse would be a big step ahead.

langchain4j:
@ruizrube would you like to help with this?

[...]

PR #1058, opened 2024-05-07
LLM Observability: Part 1 (OpenAI)

ghost:
## Issue
https://github.com/langchain4j/langchain4j/issues/199

## Change
- Added `ModelListener`, `ChatLanguageModelRequest`, and `ChatLanguageModelResponse` that are compatible (have all the required attributes) with [OTEL LLM semconv draft](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/gen-ai/llm-spans.md).
- Added an option to attach multiple `ModelListener<ChatLanguageModelRequest, ChatLanguageModelResponse>` to `OpenAiChatModel` and `OpenAiStreamingChatModel` (pilot module).

### `ChatLanguageModelRequest`
```java
public class ChatLanguageModelRequest {

    private final String model;
    private final Double temperature;
    private final Double topP;
    private final Integer maxTokens;
    private final List<ChatMessage> messages;
    private final List<ToolSpecification> toolSpecifications;
}
```

### `ChatLanguageModelResponse`
```java
public class ChatLanguageModelResponse {

    private final String id;
    private final String model;
    private final TokenUsage tokenUsage;
    private final FinishReason finishReason;
    private final AiMessage aiMessage;
}
```

## Example
```java
ModelListener<ChatLanguageModelRequest, ChatLanguageModelResponse> modelListener =
                new ModelListener<ChatLanguageModelRequest, ChatLanguageModelResponse>() {

                    @Override
                    public void onRequest(ChatLanguageModelRequest request) {
                        // handle request
                    }

                    @Override
                    public void onResponse(ChatLanguageModelResponse response, ChatLanguageModelRequest request) {
                        // handle response
                    }

                    @Override
                    public void onError(Throwable error, ChatLanguageModelResponse response, ChatLanguageModelRequest request) {
                        // handle error
                    }
                };

        OpenAiChatModel model = OpenAiChatModel.builder()
                .apiKey(...)
                .listeners(singletonList(modelListener))
                .build();
```

## General checklist


langchain4j:
@brunobat could you please take a look? 🙏

geoand:
I'm also very interested in this so we can augment in the already present observability features in the Quarkus implementation.

I also think that having something like this in sooner that rather than later would be very helpful as these new integration points could definitely prove useful for various tasks.

geoand:
I believe that `onResponse` could also be passed `ChatLanguageModelRequest`

langchain4j:
@geoand yeah I was just adding it :)

geoand:
💪🏼

langchain4j:
@geoand do you think `id` can still be useful for something (after adding `request` to `onResponse()`)?

geoand:
I very much doubt it's useful

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
```diff
@@ -0,0 +1,48 @@
...
+
+    /**
+     * TODO
+     *
+     * @param request
+     * @param error
+     */
+    // TODO names
+    @Experimental
+    default void onError(ChatLanguageModelRequest request, Throwable error) {
```
langchain4j: should this method be non-default to enforce the user to handle error?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
```diff
@@ -152,22 +158,52 @@ private void generate(List<ChatMessage> messages,
...
+                        );
+                        listener.onResponse(observabilityRequest, observabilityResponse);
+                    }
+
                     handler.onComplete(response);
                 })
-                .onError(handler::onError)
+                .onError(error -> {
+                    if (listener != null) {
+                        listener.onError(observabilityRequest, error);
```
langchain4j: TODO:
- check that exception cannot be thrown outside of onError
- what to do with exceptions produced by user's handler?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
geoand: I don't think it has to

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
```diff
@@ -0,0 +1,48 @@
...
+
+import dev.langchain4j.Experimental;
+
+/**
+ * TODO
+ */
+// TODO name: ChatLanguageModelObserver?
+// TODO package
+@Experimental
+public interface ChatLanguageModelListener {
```
brunobat: What about TracingListener?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
brunobat: I think we should follow the same approach as the vert.x tracing API and squash `onResponse` and the `onError` methods.

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
brunobat: We will need to call the listener from different frameworks, like OTel and Micrometer. Iterating through the listeners would be nice.

langchain4j: It would be nice to keep it more generic, this listener can be used not only for tracing

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
langchain4j: I would still prefer to differentiate between "successful response" and "error" on the API level, I think it makes more sense from the user perspective.

Given the following API:
```java
public interface ModelListener<Request, Response> {

    default void onRequest(Request request) {

    }

    default void onResponse(Request request, Response response) {

    }

    default void onError(Request request, Response response, Throwable error) {

    }
}
```

Given that "squashing" will be required for tracing, it still can be done on the implementor side like this:
```java
ModelListener modelListener = new ModelListener() {

    @Override
    public void onRequest(Request request) {
        // process request
    }

    @Override
    public void onResponse(Request request, Response response) {
        onError(request, response, null);
    }

    @Override
    public void onError(Request request, Response response, Throwable error)    {
        // process response and error
    }
};
```

Would that work for you?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
langchain4j: Totally makes sense, will add!

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
brunobat: No strong feeling here. This should be fine:
```java
public interface ModelListener<Request, Response> {

    default void onRequest(Request request) {

    }

    default void onResponse(Request request, Response response) {

    }

    default void onError(Request request, Response response, Throwable error) {

    }
}
```

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
brunobat: Then ObservabilityListener

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
brunobat: Swallow them and log a warning.
Monitoring code shouldn't break the main business logic flow.

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
geoand: This is a general middleware piece, I don't think it should have any ties with Observability in terms of naming.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
geoand: > Monitoring code shouldn't break the main business logic flow.

Very true, but I see this as a generic middleware piece and theoretically some other kind of listener might want to do something with it

langchain4j-core/src/main/java/dev/langchain4j/model/chat/observability/ChatLanguageModelListener.java
langchain4j: I will keep it as `dev.langchain4j.model.listener.ModelListener<Request, Response>`

langchain4j: different types of models can have different request/response data

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
langchain4j: I will probably just log it for now, for simplicity. Listener that listens on another listener error can always be added, if/when needed.

geoand:
I am actually thinking that the listener might need to be changed to:

```java
public interface ModelListener<Request, Response> {

    default OnRequestResult<Request> onRequest(Request request) {
        return null;
    }

    default void onResponse(Response response, OnRequestResult<Request> onRequestResult) {

    }

    default void onError(Throwable error, Response response, OnRequestResult<Request> onRequestResult) {

    }

    /**
     * This name is horrible, and should replaced by something better 
     */    
    interface OnRequestResult<Request> {
        Request request();
    }
}
```

The reason is that an integration might want to include some custom data when the request is created and that data might need to be accessible when the result comes back (or an error occurs).

WDYT?

langchain4j:
@geoand can you provide an example when this might be required?

geoand:
I am thinking that with OpenTelemetry if I want to open a new span on request and close it on end, I need a way to keep hold of that. Now OTel might have a way to do that already (with ThreadLocal, or some pluggable strategy), but other APIs might not.

brunobat:
We can add to the request object the OTel context

geoand:
Right, but other APIs might not have that kind of capability

geoand:
Here is another example:

With the current API, how would one use Micrometer to add a timed metric of the operation?

P.S. I take complete responsibility for not spotting this issue earlier.

geoand:
> We can add to the request object the OTel context

Looking at this more, I don't see how one could close the `Scope` with the current LangChain4j API.

geoand:
I have a draft proposal of what I would like to do [here](https://github.com/langchain4j/langchain4j/pull/1157).

Using that in Quarkus LangChain4j I would utilize it like so:

```java
public class OpenTelemetryChatLanguageModelListener implements ModelListener<ChatLanguageModelRequest, OpenTelemetryChatLanguageModelListener.ScopeHoldingOnRequestResult, ChatLanguageModelResponse> {

    private final Tracer tracer;

    @Inject
    public OpenTelemetryChatLanguageModelListener(Tracer tracer) {
        this.tracer = tracer;
    }

    @Override
    public ScopeHoldingOnRequestResult onRequest(ChatLanguageModelRequest request) {
        String name = "ChatCompletions " + request.model();

        Span span = tracer.spanBuilder(name).startSpan();
        Scope scope = span.makeCurrent();

        // TODO: implement

        return new ScopeHoldingOnRequestResult(request, scope);
    }

    @Override
    public void onResponse(ChatLanguageModelResponse chatLanguageModelResponse, ScopeHoldingOnRequestResult request) {
        // TODO: implement
        request.scope.close();
    }

    @Override
    public void onError(Throwable error, ChatLanguageModelResponse chatLanguageModelResponse,
                        ScopeHoldingOnRequestResult request) {
        // TODO: implement
        request.scope.close();
    }

    public static class ScopeHoldingOnRequestResult implements OnRequestResult<ChatLanguageModelRequest> {

        private final ChatLanguageModelRequest request;
        private final Scope scope;

        private ScopeHoldingOnRequestResult(ChatLanguageModelRequest request, Scope scope) {
            this.request = request;
            this.scope = scope;
        }

        @Override
        public ChatLanguageModelRequest request() {
            return null;
        }
    }
}
```

brunobat:
Will comment on the PR

PR #1067, opened 2024-05-08
Add new ChatMemory implementation to be used for stateful data extraction

mariofusco:
This pull request implements the `ChatMemory` that I have discussed and proposed [here](https://github.com/langchain4j/langchain4j/pull/862#discussion_r1557626867). This should be a good fit to be used for stateful data extraction, so for instance the included test case produces the following output:
```
User: hi
User: my name is Mario Fusco
User: I'm 50
Extracted Customer { firstName = "Mario", lastName = "Fusco", age = 50 }
```
In essence what this `ChatMemory` does is simply concatenating the values of variables sent by a user at each iteration, recreating at each step the user message from the original prompt template and those concatenated variables. In this way the user message sent to the LLM at the 3rd prompt of my example above will be something like:

"Extract information about a customer from this text 'hi. my name is Mario Fusco. I'm 50'. The response must contain only the JSON with customer's data and without any other sentence. You must answer strictly in the following JSON format: {\n"firstName": (type: string),\n"lastName": (type: string),\n"age": (type: integer)"

In order to implement this feature I had to add to the `UserMessage` both the prompt template and the set of variables from which it has been created. I believe that carrying those information can be useful also beyond the specific needs of this pull request. In reality probably it would be an ever better design if the `UserMessage` would know how to render itself and use the `PromptTemplate` internally instead of having a text populated from the outside as it does now. I'm open to also implement this further improvement, but for now I just wanted to demonstrate the general idea with the smallest possible set of changes.

/cc @sebastienblanc

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/data/message/UserMessage.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/data/message/UserMessage.java
@@ -1,6 +1,7 @@
 package dev.langchain4j.data.message;
 
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 import static dev.langchain4j.data.message.ChatMessageType.USER;
@@ -26,6 +27,9 @@ public class UserMessage implements ChatMessage {
 
     private final String name;
     private final List<Content> contents;
+
+    private String template;
+    private Map<String, Object> variables;
 
     /**
      * Creates a {@link UserMessage} from a text.
@@ -136,6 +140,24 @@ public class UserMessage implements ChatMessage {
      */
     public boolean hasSingleText() {
         return contents.size() == 1 && contents.get(0) instanceof TextContent;
+    }
+
+    public UserMessage fromTemplate(String template) {
+        this.template = template;
+        return this;
+    }
+
+    public String template() {
+        return template;
+    }
+
+    public UserMessage withVariables(Map<String, Object> variables) {
+        this.variables = variables;
+        return this;
+    }
+
+    public Map<String, Object> variables() {
+        return variables;
     }
 
     /**
--- a/langchain4j/src/main/java/dev/langchain4j/memory/chat/UserMessagesConcatChatMemory.java
+++ b/langchain4j/src/main/java/dev/langchain4j/memory/chat/UserMessagesConcatChatMemory.java
@@ -1,0 +1,160 @@
+package dev.langchain4j.memory.chat;
+
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.SystemMessage;
+import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.data.message.UserMessage;
+import dev.langchain4j.memory.ChatMemory;
+import dev.langchain4j.model.input.Prompt;
+import dev.langchain4j.model.input.PromptTemplate;
+import dev.langchain4j.store.memory.chat.ChatMemoryStore;
+import dev.langchain4j.store.memory.chat.InMemoryChatMemoryStore;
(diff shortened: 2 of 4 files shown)


langchain4j:
Hi @mariofusco, thanks a lot! Will try to review it asap

mariofusco:
Do you have any news about this? I'm seeing that this pull request could be relevant also in the light of broader [feature requests](https://github.com/langchain4j/langchain4j/issues/1184) related to chat memories. More in general we may want to introduce some sort of minimal SPI to facilitate the pluggability of custom chat memory implementations and maybe rewrite the existing memories in terms of this SPI. I could sketch this idea in a different pull request if you're interested, or maybe did you already have something similar in mind?

langchain4j:
@mariofusco sorry, I did not have time to look at it yet, I will try to do it today

langchain4j:
Hi @mariofusco! If I understand correctly, this use case assumes interactive conversation with the user to collect all the needed details, right? But in the test there is no response from the model (with guidance what information should be provided) because it outputs a `Customer` object, not a text. How exactly should this be used? Thanks!

mariofusco:
> Hi @mariofusco! If I understand correctly, this use case assumes interactive conversation with the user to collect all the needed details, right? But in the test there is no response from the model (with guidance what information should be provided) because it outputs a `Customer` object, not a text. How exactly should this be used? Thanks!

That's correct, this chat memory is designed to be used with extractors so it cannot provide any message for the user. This means that a second AI service needs to be used together with this in order to give some feedback to user. [Here](https://github.com/mariofusco/quarkus-drools-llm/tree/main?tab=readme-ov-file#the-airline-chatbot-example) you can see an example of how I used a very similar strategy. In fact [here](https://github.com/mariofusco/quarkus-drools-llm/blob/main/src/main/java/org/hybridai/refund/StateManager.java) for each state of the conversation I had to use both a `[Customer/Flight]Extractor` and `[Customer/Flight]ChatService`. I have no idea if there's a better way to achieve a similar result and in particular if it is possible to avoid the double AI service for this situation. Any advice is welcome.

langchain4j:
Hi @mariofusco, one thing that comes to my mind is to provide an LLM with tools to save/validate customer details. Something like:
```java
@Tool
String saveUserName(String name, @ToolMemoryId String memoryId) {
    Customer customer = getCustomer(memoryId);
    customer.setName(name);
    return customer.isValid() ? "successfully collected all customer details" : "some customer details are missing"; // maybe specify what exactly is missing
}

@Tool
String saveUserAge(int age, @ToolMemoryId String memoryId) {
    Customer customer = getCustomer(memoryId);
    customer.setAge(age);
    return customer.isValid() ? "successfully collected all customer details" : "some customer details are missing"; // maybe specify what exactly is missing
}
```
Then the same AI Service/LLM can be used to drive the conversation and collect details.
The downside is that LLM must support tools, but more and more LLMs support them nowadays, so we are almost there.
WDYT?

Regarding this PR:
I would try to avoid making changes in `UserMessage` as it will require more changes to serialization/deserialization, etc.
But in general, I agree that it makes sense to keep user message internals more granular, and I anticipate that we will address that soon.
Maybe we can leverage existing `List<Content>`. Each part (user message itself, RAG data, output formatting instructions, etc) of user message can be represented as a single `Content`. We could also add metadata to `Content` to be able to assign labels (e.g. `user_query`, `rag_data`, `output_format_instructions`, etc.) and then filter by it.

mariofusco:
I will give a try to what you suggested, even though it seems a bit more cumbersome since (if I understand correctly) it requires to configure a different tool for each field of the domain object to be retrieved from the chat. 

Regarding this PR it is ok for me to close it, but I still believe that there should be a way to implement and configure chat memories in a more fine grained way. In case you will make any progress on this or I can help with something please keep me updated.

langchain4j:
> it requires to configure a different tool for each field of the domain object

I would try a single tool per domain object with all the fields, it might work as well. Ideally all tool params should be optional, there was a [PR](https://github.com/langchain4j/langchain4j/pull/922), I need to finish it.

> I still believe that there should be a way to implement and configure chat memories in a more fine grained way

I completely agree, I am just not sure at this point if it should be a template/vars or something different (e.g. separate `Content`s).

> In case you will make any progress on this or I can help with something please keep me updated.

OK!

Thank you!

PR #1072, opened 2024-05-09
Migrate ollama to jackson

Martin7-1:
## Issue

#1042 

## Change

Migrate Ollama module from Gson to Jackson. But we still need to migrate `langchain4j-core` and `openai4j` too.

image

Note that there are three things confused me(these issues exist in original ollama module): 

1. I observe that some tests with inputTokenCount Assertion (such as `assertThat(tokenUsage.inputTokenCount()).isEqualTo(35)`) failed because of the differencet count of input tokens. I don't know whether it's because Ollama's update. Now I correct it
2. `should_propagate_failure_to_handler_onError` in `Streaming{xxx}ModelIT` failed in my local because `NullPointerException` do not have any message. Is it a problem in my local environment?
3. Somtimes if I run tests **individually**, all tests will pass. But if I run them in the same time(such as using command line rather than IDE), they will failed due to some strange reasons(e.g. input token usage will be null). I think Maybe it's a network problem or `Testcontainers`'s problem.

I'm not sure if these problems are due to my local environment, so if you have any suggestions or solutions, please let me know!

## General checklist


langchain4j:
Hi @Martin7-1 thanks a lot! I will try to review it asap!

langchain4j:
Hi, thanks a lot!
This PR should wait till we release the first migration from Gson to Jackson (Anthropic module) and see that it actually works.
We can then proceed with this PR and migrating other modules.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
```diff
@@ -27,15 +26,16 @@
...
 import static java.lang.Boolean.TRUE;
 
 @Slf4j
 class OllamaClient {
 
-    private static final Gson GSON = new GsonBuilder()
-            .setFieldNamingPolicy(LOWER_CASE_WITH_UNDERSCORES)
-            .create();
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper()
+            .setPropertyNamingStrategy(SNAKE_CASE)
```
langchain4j: Please use `@JsonNaming(SnakeCaseStrategy.class)` on each model class instead

Martin7-1: Done.

Martin7-1:
I found a strange thing that all `numPredict` parameter failed to restrict the max output token in tests.

Martin7-1:
BTW, I found that there are some token usage differences. I don't know what the reason is

ghost (approved):
@Martin7-1 thanks a lot! Token differences are ok

PR #1084, opened 2024-05-10
AzureAiSearchEmbeddingStore - add "indexName" to the builder

### LINKED ISSUE #1062: [FEATURE] AzureAiSearchEmbeddingStore - add "indexName" to the builder

Currently, it's not possible to override the index name without creating the entire index by yourself. 

I would like to be able to set the index name using `AzureAiSearchEmbeddingStore.builder()` like 
```
AzureAiSearchEmbeddingStore.builder()
                .apiKey(properties.apiKey())
                .endpoint(properties.endpoint())
                .indexName(myIndexName)
                .dimensions(DIMENSIONS)
                .build();
```

or 

```
AzureAiSearchEmbeddingStore.builder()
                .apiKey(properties.apiKey())
                .endpoint(properties.endpoint())
                .indexName(indexBuilder -> indexBuilder.name(myIndexName))
                .dimensions(DIMENSIONS)
                .build();
```

In other words, it would be great to still rely on the default implementation regarding the index settings however, at the same time, be able to override only the index name.

jdubois:
Fix #1062

### DIFF
--- a/langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
+++ b/langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
@@ -14,20 +14,20 @@ import static dev.langchain4j.internal.ValidationUtils.ensureTrue;
  */
 public class AzureAiSearchEmbeddingStore extends AbstractAzureAiSearchEmbeddingStore implements EmbeddingStore<TextSegment> {
 
-    public AzureAiSearchEmbeddingStore(String endpoint, AzureKeyCredential keyCredential, boolean createOrUpdateIndex, int dimensions) {
-        this.initialize(endpoint, keyCredential, null, createOrUpdateIndex, dimensions, null);
+    public AzureAiSearchEmbeddingStore(String endpoint, AzureKeyCredential keyCredential, boolean createOrUpdateIndex, int dimensions, String indexName) {
+        this.initialize(endpoint, keyCredential, null, createOrUpdateIndex, dimensions, null, indexName);
     }
 
-    public AzureAiSearchEmbeddingStore(String endpoint, AzureKeyCredential keyCredential, boolean createOrUpdateIndex, SearchIndex index) {
-        this.initialize(endpoint, keyCredential, null, createOrUpdateIndex, 0, index);
+    public AzureAiSearchEmbeddingStore(String endpoint, AzureKeyCredential keyCredential, boolean createOrUpdateIndex, SearchIndex index, String indexName) {
+        this.initialize(endpoint, keyCredential, null, createOrUpdateIndex, 0, index, indexName);
     }
 
-    public AzureAiSearchEmbeddingStore(String endpoint, TokenCredential tokenCredential, boolean createOrUpdateIndex, int dimensions) {
-        this.initialize(endpoint, null, tokenCredential, createOrUpdateIndex, dimensions, null);
+    public AzureAiSearchEmbeddingStore(String endpoint, TokenCredential tokenCredential, boolean createOrUpdateIndex, int dimensions, String indexName) {
+        this.initialize(endpoint, null, tokenCredential, createOrUpdateIndex, dimensions, null, indexName);
     }
 
-    public AzureAiSearchEmbeddingStore(String endpoint, TokenCredential tokenCredential, boolean createOrUpdateIndex, SearchIndex index) {
-        this.initialize(endpoint, null, tokenCredential, createOrUpdateIndex, 0, index);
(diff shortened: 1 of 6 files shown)


langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -115,21 +117,32 @@ public Builder index(SearchIndex index) {
...
+         * @param indexName The name of the index to be used.
+         * @return builder
+         */
+        public Builder indexName(String indexName) {
+            this.indexName = indexName;
+            return this;
+        }
+
         public AzureAiSearchEmbeddingStore build() {
             ensureNotNull(endpoint, "endpoint");
```
vprudnikov, line 132: I assume the builder will ensure that either indexName or index is set but never both. Otherwise, it could be a bit hard to predict the implementation behavior. 
If you agree then the test should verify this rule as well.

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
```diff
@@ -115,21 +117,32 @@ public Builder index(SearchIndex index) {
             return this;
         }
 
+        /**
+         * If no index is provided, set the name of the default index to be used.
```
vprudnikov, line 121: This is just a general comment regarding this approach: I feel the API becomes not that straightforward. When I look at it, I start thinking about what erases what for example.

I suggest considering one more option.
You could reload the same method thus making it more explicit that setting the index name will override setting the index. For example:

```
    public Builder setIndex(SearchIndex index) {
        this.index = index;
        this.indexName = null;  // Clear the index name as the index object takes precedence
        return this;
    }

    public Builder setIndex(String indexName) {
        this.indexName = indexName;
        this.index = null;  // Clear the index object as the index name takes precedence
        return this;
    }
```

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
jdubois: Please have a look at my latest commit at https://github.com/langchain4j/langchain4j/pull/1084/commits/ec7fd326b14f2d6feee66d829cad2beea7f4d0ec where I add a test for this

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
jdubois: Same comment as above (this is why I mark this one as resolved, let's have only one discussion). I'm throwing an error, and added more documentation, to be sure people understand what's happening.

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/store/embedding/azure/search/AzureAiSearchEmbeddingStore.java
vprudnikov: looks good!

Since another comment is resolved now, I'll just write here my thoughts, however, you decide :)

I think it would be great to think about the following: What if someone else asks you to add one more option, e.g. 'eTag' or anything else?

* how much documentation will you update regarding what erases what and so on?
* how many NULLs / IFs / logical operators will appear in your code?
* will the complexity of the API increase or not?

If you think the builder API will be still OK, then this PR should be merged.

I got another idea of how it could look like from the client perspective. 

```
AzureAiSearchEmbeddingStore.builder()
                .apiKey(properties.apiKey())
                .endpoint(properties.endpoint())
                .dimensions(1000)
//defaultSearchIndex() is a static method provided by AbstractAzureAiSearchEmbeddingStore
                .index(defaultSearchIndex().setName("myName"))
                .build();
```
However, the code above will probably struggle supporting dimensions.

One more approach:

````
AzureAiSearchEmbeddingStore.builder()
                .apiKey(properties.apiKey())
                .endpoint(properties.endpoint())
                .dimensions(1000)
//your implementation provides a template (pre-built SearchIndex) with all fields set, so developers will be able to modify it in the way they want by passing a Consumer<SearchIndex>
//As a result, the implementation won't suffer from additional NULLs/IF-checks/exceptions thrown.
                .index(indexTemplate -> indexTemplate.setName("myName"))
                .build();
```

ghost (approved):
@jdubois thank you a lot!

Issue #1044, opened 2024-04-30
Decouple modules from HTTP client (currently okhttp/retrofit)


langchain4j:
Modules should be decoupled from the HTTP client to allow using other HTTP clients than okhttp (e.g. RestClient for Spring Boot).

kchaber:
The inspiration might be the Spring’s `RestTemplate`, which allows defining mulitple http backends through the [ClientHttpRequestFactory](https://github.com/spring-projects/spring-framework/blob/69eaf8f8289436d154823e63d71e5102a719ddac/spring-web/src/main/java/org/springframework/http/client/ClientHttpRequestFactory.java). There are multiple [implementations](https://github.com/spring-projects/spring-framework/tree/69eaf8f8289436d154823e63d71e5102a719ddac/spring-web/src/main/java/org/springframework/http/client) for common http providers, which can be passed via `RestTemplate.setRequestFactory(…)`.

One16-KS:
Retrofit has a hard dependency on okhttp.. should we throw retrofit in the mix as well?  Retrofit could be replaced by competing libs such as openfeign?

langchain4j:
Right, I meant okhttp+retrofit. I would avoid reflection and dynamic proxies as much as possible, I guess we can live without this convenience.

langchain4j:
@kchaber @One16-KS could you please review the first draft: https://github.com/langchain4j/langchain4j/pull/1103 ?

One16-KS:
Planning to do so this weekend!

One16-KS:
Nothing to add really, really good first step, I'll try to chip in if I have spare time.

kenix:
Hint: Spring-Web's `RestClient` is the replacement for `RestTemplate`.

As of version 0.36.2 of langchain4j a streaming chat model always waits for 1 minute to shutdown, even though we can make sure no new requests and all onCompletes are already finished. The `OkHttpClient` is completely hidden e.g. in `OllamaClient`. Is there way to gracefully shutdown the "OkHttp Dispatcher" as of now?

langchain4j:
Unfortunately no way to shutdown yet. The plan is to migrate Ollama to Java 11 HTTP Client soon and make http client customizable. Since this is an Okhttp issue, it should be solved by this migration.

kenix:
[...]

PR #1103, opened 2024-05-14
Draft: Decouple modules from (ok)http client

ghost:
## Issue
https://github.com/langchain4j/langchain4j/issues/1044

## Change
This is the first step towards decoupling modules from (ok)http client.
This PR introduces an abstraction layer (in `langchain4j-core` module) over various http clients.
Main components:
- `HttpClient`
- `HttpClientBuilder`
- `HttpRequest`
- `HttpResponse`

This PR also provides 2 implementations of `HttpClient`:
- `OkHttpHttpClient` (uses okhttp that is currently used in most modules, but does not use retrofit)
- `SpringRestClientHttpClient` (uses `RestClient` from Spring)

More implementations (e.g. JDK 11 `HttpClient`) can be added later.

The first module to migrate is `langchain4j-anthropic`, this PR decouples it from okhttp.
Users are able to provide a `HttpClientBuilder` when creating `Anthropic*Model`. If not provided, default (okhttp) will be loaded via SPI.

With this change, request/response logging is also unified and moved to `HttpRequestLogger` and `HttpResponseLogger`. So all modules can now get logging for free.

## General checklist


langchain4j:
Hi @geoand and @ThomasVitale 
Could you please take a quick look at this draft? 🙏

geoand:
Nice!

Right off the bat I would say that I don't believe there should be a Spring specific implementation in the core repo. Spring REST Template is not really something many (any?) non Spring application use.
If anything, there could be a JDK HttpClient module (which would require Java 11+ though).

geoand:
Also for the Quarkus integration, I don't really think this will make much of a difference as we really want to continue using the declarative clients

langchain4j:
@geoand thanks for the feedback! Spring specific things will be moved to `langchain4j-spring` repo (once this PR is about to be finalized), there are a few TODOs in the PR.
Regarding JDK 11 HttpClient, I guess it makes total sense, but probably can be delayed a bit, since there are so many other higher-prio issues.

http-clients/langchain4j-http-client-spring-restclient/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
+            <version>6.1.6</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot</artifactId>
+            <!-- TODO take version from parent -->
+            <version>3.2.6</version>
```
kpavlov, line 34: Let's maybe move all versions to parent pom. Also, SB -> 3.3.4

And why spring boot is needed? RestClient is part of the Spring Framework, not a Spring Boot

http-clients/langchain4j-http-client-spring-restclient/pom.xml
```diff
@@ -0,0 +1,74 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-web</artifactId>
+            <version>6.1.6</version>
```
kpavlov, line 27: Maybe, let's use 
```xml
<!-- https://mvnrepository.com/artifact/org.springframework/spring-framework-bom -->
<dependency>
    <groupId>org.springframework</groupId>
    <artifactId>spring-framework-bom</artifactId>
    <version>6.1.13</version>
    <type>pom</type>
</dependency>
``` 
in parent pom. Also, move versions to properties

http-clients/langchain4j-http-client-spring-restclient/src/main/java/dev/langchain4j/http/spring/restclient/SpringRestClientHttpClient.java
```diff
@@ -0,0 +1,131 @@
...
+
+import static dev.langchain4j.internal.Utils.getOrDefault;
+
+public class SpringRestClientHttpClient extends AbstractHttpClient {
+
+    private final RestClient restClient;
+    private final boolean logRequests;
+    private final boolean logResponses;
+
+    public SpringRestClientHttpClient(SpringRestClientHttpClientBuilder builder) {
```
kpavlov, line 25: Let's have also a constructor with `RestClient.Builder` as a parameter. Spring Boot pre-configures if with observability features
https://docs.spring.io/spring-boot/reference/io/rest-client.html#io.rest-client.restclient

langchain4j-core/src/main/java/dev/langchain4j/http/HttpRequestLogger.java
```diff
@@ -0,0 +1,55 @@
...
+@Experimental
+class HttpRequestLogger {
+
+    private static final Set<String> COMMON_SECRET_HEADERS =
+            new HashSet<>(asList("authorization", "x-api-key", "x-auth-token"));
+
+    static void log(HttpRequest httpRequest) {
+        try {
+            // TODO pretty print body?
+            log.debug("Request:\n- method: {}\n- url: {}\n- headers: {}\n- body: {}",
```
kpavlov, line 24: This piece of code requires extreme care to avoid any PII information. May I suggest to log body only if trace level is enabled?

Should logging a payload be a responsibility of concrete http client? Do we really need the logger abstraction?

langchain4j-core/src/main/java/dev/langchain4j/http/HttpRequestLogger.java
```diff
@@ -0,0 +1,55 @@
...
+        return String.format("[%s: %s]", headerKey, headerValue);
+    }
+
+    static String maskSecretKey(String key) {
+        if (isNullOrBlank(key)) {
+            return key;
+        }
+
+        if (key.length() >= 7) {
+            return key.substring(0, 5) + "..." + key.substring(key.length() - 2);
```
kpavlov, line 50: maybe just always mask it with "***"

langchain4j-core/src/main/java/dev/langchain4j/http/AbstractHttpClient.java
```diff
@@ -0,0 +1,62 @@
+package dev.langchain4j.http;
+
+import dev.langchain4j.Experimental;
+
+@Experimental
+public abstract class AbstractHttpClient implements HttpClient { // TODO name. LoggingHttpClient?
+
+    public HttpResponse execute(HttpRequest httpRequest) {
+
+        if (logRequests()) {
```
kpavlov, line 10: I think it is better to offload logging to concrete http client interceptors (client-native logging mechanism)
At least, not focus on this in the first iteration. Not logging is safer from security point of view than making a bug here and logging too much

langchain4j-core/src/main/java/dev/langchain4j/http/AbstractHttpClient.java
```diff
@@ -0,0 +1,62 @@
...
+            // TODO it logs now as dev.langchain4j.http.HttpRequestLogger.log(), not model-specific
+            HttpRequestLogger.log(httpRequest);
+        }
+
+        HttpResponse httpResponse = doExecute(httpRequest);
+
+        if (logResponses()) {
```
kpavlov, line 17: same as above

http-clients/langchain4j-http-client-okhttp/src/test/java/dev/langchain4j/http/okhttp/OkHttpHttpClientIT.java
```diff
@@ -0,0 +1,8 @@
+package dev.langchain4j.http.okhttp;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class OkHttpHttpClientIT {
```
kpavlov, line 5: Let's maybe remove or implement it. BTW, everything should be already covered by existing integration tests

langchain4j-core/src/main/java/dev/langchain4j/http/HttpClientBuilderLoader.java
```diff
@@ -0,0 +1,30 @@
...
+
+@Slf4j
+@Experimental
+public class HttpClientBuilderLoader {
+
+    public static HttpClientBuilder loadHttpClientBuilder() {
+        Collection<HttpClientBuilderFactory> factories = loadFactories(HttpClientBuilderFactory.class);
+        if (factories.size() > 1) {
+            // TODO log names of factories
+            throw new RuntimeException("Conflict: multiple HTTP clients have been found in the classpath. " +
```
kpavlov, line 18: IllegalStateException?

langchain4j-core/src/main/java/dev/langchain4j/http/HttpClientBuilderLoader.java
```diff
@@ -0,0 +1,30 @@
...
+
+import static dev.langchain4j.spi.ServiceHelper.loadFactories;
+
+@Slf4j
+@Experimental
+public class HttpClientBuilderLoader {
+
+    public static HttpClientBuilder loadHttpClientBuilder() {
+        Collection<HttpClientBuilderFactory> factories = loadFactories(HttpClientBuilderFactory.class);
+        if (factories.size() > 1) {
```
kpavlov, line 16: what if thare are no factories?

langchain4j-core/src/main/java/dev/langchain4j/http/HttpException.java
```diff
@@ -0,0 +1,21 @@
+package dev.langchain4j.http;
+
+import dev.langchain4j.Experimental;
+
+@Experimental
+public class HttpException extends RuntimeException {
+
+    private final Integer statusCode;
```
kpavlov, line 8: Maybe `int` ?
Does exception without code it make sense?

langchain4j-core/src/main/java/dev/langchain4j/http/HttpException.java
```diff
@@ -0,0 +1,21 @@
...
+
+    public HttpException(Integer statusCode, String message) {
+        super(message);
+        this.statusCode = statusCode;
+    }
+
+    /**
+     * HTTP response status code. Can be {@code null}.
```
kpavlov, line 16: What `null` means?

langchain4j-core/src/main/java/dev/langchain4j/http/HttpException.java
```diff
@@ -0,0 +1,21 @@
...
+     */
+    public Integer statusCode() {
```
kpavlov, line 18: If it is nullable, then `@jakarta.annotations.Nullable` is needed

langchain4j-core/src/main/java/dev/langchain4j/http/HttpRequest.java
```diff
@@ -0,0 +1,71 @@
...
+@Experimental
+public class HttpRequest {
+
+    private final HttpMethod method;
+    private final String url;
+    private final Map<String, String> headers;
+    private final String body; // TODO type
+
+    @Builder
+    public HttpRequest(HttpMethod method, String url, Map<String, String> headers, String body) {
```
kpavlov, line 22: Let's add nullability annotations to parameters and methods

langchain4j-core/src/main/java/dev/langchain4j/http/HttpResponse.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.Map;
+
+import static java.util.Collections.emptyMap;
+
+@Experimental
+public class HttpResponse {
+
+    private final int statusCode;
+    private final Map<String, String> headers;
+    private final String body;
```
kpavlov, line 16: Is it always a string? WHat if it is image? Shouldn't it be a generic type `T`?

langchain4j-core/src/main/java/dev/langchain4j/http/HttpResponse.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Builder
```
kpavlov, line 18: If lombok is to be removed, let's maybe not pull it in?

langchain4j-core/src/main/java/dev/langchain4j/http/ServerSentEvent.java
```diff
@@ -0,0 +1,14 @@
+package dev.langchain4j.http;
+
+import dev.langchain4j.Experimental;
+import lombok.Builder;
+import lombok.Getter;
+
+@Getter
+@Builder
```
kpavlov, line 8: Lombok... Java records would have look better here

http-clients/langchain4j-http-client-okhttp/src/main/java/dev/langchain4j/http/okhttp/OkHttpHttpClient.java
```diff
@@ -0,0 +1,151 @@
...
+public class OkHttpHttpClient extends AbstractHttpClient {
+
+    // TODO make configurable?
+    private static final MediaType JSON = MediaType.get("application/json; charset=utf-8");
+
+    private final okhttp3.OkHttpClient client;
+    private final boolean logRequests;
+    private final boolean logResponses;
+
+    public OkHttpHttpClient(OkHttpHttpClientBuilder builder) {
```
kpavlov, line 24: Maybe another constructor with just `OkHttpClient.Builder` will be also useful

langchain4j-core/src/main/java/dev/langchain4j/http/HttpClientBuilder.java
```diff
@@ -0,0 +1,33 @@
+package dev.langchain4j.http;
+
+import dev.langchain4j.Experimental;
+
+import java.time.Duration;
+
+@Experimental
+public interface HttpClientBuilder {  // TODO move inside HttpClient?
```
kpavlov, line 8: Looks like a properties object: `OkHttpHttpClientProperties`

I would even delegate this to concrete factory implementation: it should take care of timeouts

kpavlov (commented):
@langchain4j, thank you for PR. I am looking forward to have this integrated soon to unblock spring boot integration. I want to use injected RestClient.Builder.

I think it is would be better to make several smaller PRs. First one should only focus on introducing http-client abstractiton and implementing OkHttpClient.

Logging is questionable from the security and performance point of view and I would abstain from implementing it and delegate it to the underlaying client implementation. At least not in one big PR

langchain4j:
@kpavlov thank you for the review!

> I am looking forward to have this integrated soon to unblock spring boot integration. I want to use injected RestClient.Builder.

I plan to get back to this pretty soon and finalize.

> I think it is would be better to make several smaller PRs. First one should only focus on introducing http-client abstractiton and implementing OkHttpClient.

It is hard to design abstractions without making sure they works for **_multiple_** things that are abstracted. As someone said, it takes 3 implementations for the API to be good enough to support more. AFAIR, it was pretty easy/fast to introduce Spring client after okhttp was there, so I expect jdk11 client will be easy to introduce as well, but crucial to make sure this abstraction is good to go.

> Logging is questionable from the security and performance point of view and I would abstain from implementing it and delegate it to the underlaying client implementation. At least not in one big PR

Thank you, good point, I will revisit this once I get back to this PR.

ehsavoie:
I've rebased that PR on main and removed the use of lombok from the core http part in https://github.com/langchain4j/langchain4j/pull/2036

langchain4j:
@ehsavoie thanks a lot! Hopefully I will get to it in a week or so

ehsavoie:
Shouldn't the client be Closeable ?

langchain4j:
> Shouldn't the client be Closeable ?

Good point, it should

http-clients/langchain4j-http-client-spring-restclient/pom.xml
dliubarskyi: good point

http-clients/langchain4j-http-client-spring-restclient/src/main/java/dev/langchain4j/http/spring/restclient/SpringRestClientHttpClient.java
dliubarskyi: `RestClient.Builder` can be set in the `SpringRestClientHttpClientBuilder`

langchain4j-core/src/main/java/dev/langchain4j/http/HttpRequestLogger.java
dliubarskyi: > May I suggest to log body only if trace level is enabled?

Do you mean `log.trace(...)` instead of `log.debug(...)` ?

> Should logging a payload be a responsibility of concrete http client? Do we really need the logger abstraction?

Good question, perhaps logging should be indeed implemented in each client

langchain4j-core/src/main/java/dev/langchain4j/http/HttpRequestLogger.java
dliubarskyi: it is definitely easier to mask with ***, but not that usable from user perspective

langchain4j-core/src/main/java/dev/langchain4j/http/AbstractHttpClient.java
dliubarskyi: ~agreed~

langchain4j-core/src/main/java/dev/langchain4j/http/HttpClientBuilderLoader.java
dliubarskyi: Should normally not happen as each module (e.g., `langchain4j-anthropic`) will have one http client (default) as a dependency.

But I will add an exception just in case

http-clients/langchain4j-http-client-spring-restclient/pom.xml
dliubarskyi: Now I remember... we need spring boot because `ClientHttpRequestFactories` is inside the `spring-boot` artifact

dliubarskyi:
Superceded by https://github.com/langchain4j/langchain4j/pull/2413

langchain4j-core/src/main/java/dev/langchain4j/http/HttpException.java
dliubarskyi: Good point

langchain4j-core/src/main/java/dev/langchain4j/http/HttpException.java
dliubarskyi: changed to `int`

http-clients/langchain4j-http-client-okhttp/src/test/java/dev/langchain4j/http/okhttp/OkHttpHttpClientIT.java
dliubarskyi: implemented a set of ITs for each client in https://github.com/langchain4j/langchain4j/pull/2413

langchain4j-core/src/main/java/dev/langchain4j/http/HttpException.java
dliubarskyi: not nullable any more

langchain4j-core/src/main/java/dev/langchain4j/http/HttpResponse.java
dliubarskyi: For all the current use cases in LC4j body is always a `String`. I don't not want to complicate it now just in case

langchain4j-core/src/main/java/dev/langchain4j/http/HttpClientBuilder.java
dliubarskyi: Please check the implementation in https://github.com/langchain4j/langchain4j/pull/2413

langchain4j-core/src/main/java/dev/langchain4j/http/AbstractHttpClient.java
dliubarskyi: I have separated logging logic into a separate `LoggingHttpClient` decorator in https://github.com/langchain4j/langchain4j/pull/2413
Logging is disabled by default and enabled only if user explicitly opted in, so I see not reason for concern here

dliubarskyi:
> Shouldn't the client be Closeable ?

@ehsavoie on a second thought, I am not so sure. If you need to release any resource, you can do that directly on the resource itself, as yo uhave access to it.

langchain4j-core/src/main/java/dev/langchain4j/http/HttpRequestLogger.java
dliubarskyi: I have separated logging logic into a separate `LoggingHttpClient` decorator in https://github.com/langchain4j/langchain4j/pull/2413
Logging is disabled by default and enabled only if user explicitly opted in, so I see not reason for concern here

PR #1020, opened 2024-04-25
PgVectorEmbeddingStore: remove by ID

nottyjay:
## Change
Add search method in PgVectorEmbeddingStore
Add remove data method in the interface of EmbeddingStore
Add remove data method in PgVectorEmbeddingStore

## Checklist
- [x] I have added unit and integration tests for my change
- [x] All unit and integration tests in the module I have added/changed are green
- [x] All unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules are green

## Checklist for adding new embedding store integration
- [x] I have added a {NameOfIntegration}EmbeddingStoreIT that extends from either EmbeddingStoreIT or EmbeddingStoreWithFilteringIT

PR #1113, opened 2024-05-15
PGVector, remove by ID, same impl as #1020 but on top of last dev.

humcqc:
same impl as #1020 but on top of last dev.

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
@@ -56,6 +56,26 @@ public interface EmbeddingStore<Embedded> {
     List<String> addAll(List<Embedding> embeddings, List<Embedded> embedded);
 
     /**
+     * Removes an embedding from the store.
+     *
+     * @param id The unique identifier of the embedding to be removed.
+     * @return True if the embedding was successfully removed, false otherwise.
+     */
+    default boolean remove(String id) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    /**
+     * Removes multiple embeddings from the store.
+     *
+     * @param ids A list of unique identifiers of the embeddings to be removed.
+     * @return A list of unique identifiers of the embeddings that were successfully removed.
+     */
+    default List<String> removeAll(List<String> ids) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    /**
      * Searches for the most similar (closest in the embedding space) {@link Embedding}s.
      * <br>
      * All search criteria are defined inside the {@link EmbeddingSearchRequest}.
--- a/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
@@ -1,7 +1,9 @@
 package dev.langchain4j.store.embedding;
 
 import dev.langchain4j.Experimental;
+import dev.langchain4j.data.document.Metadata;
 import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.data.segment.TextSegment;
 import dev.langchain4j.store.embedding.filter.Filter;
 
 import java.util.List;
@@ -59,9 +61,16 @@ public interface EmbeddingStore<Embedded> {
      * Removes an embedding from the store.
      *
      * @param id The unique identifier of the embedding to be removed.
-     * @return True if the embedding was successfully removed, false otherwise.
      */
-    default boolean remove(String id) {
+    default void remove(String id) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    /**
+     * Removes all embeddings from the store.
+     *
+     */
+    default void removeAll() {
         throw new UnsupportedOperationException("Not supported yet.");
     }
 
@@ -69,9 +78,21 @@ public interface EmbeddingStore<Embedded> {
      * Removes multiple embeddings from the store.
      *
      * @param ids A list of unique identifiers of the embeddings to be removed.
-     * @return A list of unique identifiers of the embeddings that were successfully removed.
      */
-    default List<String> removeAll(List<String> ids) {
+    default void removeAll(List<String> ids) {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    /**
+     * Removes multiple embeddings from the store.
+     *
+     * @param filter         The filter to be applied to the {@link Metadata} during delete.
+     *                       Only {@link TextSegment}s whose {@link Metadata}
+     *                       matches the {@link Filter} will be deleted.
+     *                       Please note that not all {@link EmbeddingStore}s support this feature yet.
+     *                       This is an optional parameter. Default: no filtering, all will be deleted.
+     */
+    default void removeAll(Filter filter) {
         throw new UnsupportedOperationException("Not supported yet.");
     }
 
--- a/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
+++ b/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
@@ -17,8 +17,10 @@ import org.slf4j.LoggerFactory;
 
 import javax.sql.DataSource;
 import java.sql.*;
-import java.util.*;
-import java.util.stream.Collectors;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.UUID;
 import java.util.stream.IntStream;
 
 import static dev.langchain4j.internal.Utils.*;
@@ -243,15 +245,14 @@ public class PgVectorEmbeddingStore implements EmbeddingStore<TextSegment> {
      * Removes an embedding from the store based on its unique identifier.
      *
      * @param id The unique identifier of the embedding to be removed.
-     * @return A boolean indicating whether the removal was successful or not.
      */
     @Override
-    public boolean remove(String id) {
+    public void remove(String id) {
         try (Connection connection = getConnection()) {
             PreparedStatement statement = connection.prepareStatement(String.format(
                     "DELETE FROM %s WHERE embedding_id = ?", table));
             statement.setObject(1, UUID.fromString(id));
-            return statement.executeUpdate() > 0;
+            statement.executeUpdate();
         } catch (SQLException e) {
             throw new RuntimeException(e);
         }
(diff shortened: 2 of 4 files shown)


humcqc:
@langchain4j , @nottyjay, this is the same impl as #1020

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -238,6 +239,35 @@ public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedde
...
+
+    /**
+     * Removes multiple embeddings from the store.
+     *
+     * @param ids A list of unique identifiers of the embeddings to be removed.
+     * @return A list of unique identifiers of the embeddings that were successfully removed.
+     */
+    @Override
+    public List<String> removeAll(List<String> ids) {
+        return ids.stream().filter(this::remove).collect(Collectors.toList());
```
humcqc: Could be enhance with batch processing or in a single request if we just use a boolean in return.

langchain4j: Shouldn't this removal be batched (in the same transaction)?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -55,6 +55,26 @@ public interface EmbeddingStore<Embedded> {
      */
     List<String> addAll(List<Embedding> embeddings, List<Embedded> embedded);
 
+    /**
+     * Removes an embedding from the store.
+     *
+     * @param id The unique identifier of the embedding to be removed.
+     * @return True if the embedding was successfully removed, false otherwise.
+     */
+    default boolean remove(String id) {
```
langchain4j:
```suggestion
    default void remove(String id) {
```
What about throwing an exception instead (in case of failures)?
Reasoning: users might not check the returned value explicitly and the problem will go unnoticed.
Exception is also more informative in terms of reason of the failure (e.g. network issue, not existing id, etc).

WDYT?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -55,6 +55,26 @@ public interface EmbeddingStore<Embedded> {
...
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    /**
+     * Removes multiple embeddings from the store.
+     *
+     * @param ids A list of unique identifiers of the embeddings to be removed.
+     * @return A list of unique identifiers of the embeddings that were successfully removed.
+     */
+    default List<String> removeAll(List<String> ids) {
```
langchain4j:
```suggestion
    default void removeAll(List<String> ids) {
```
same here

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -55,6 +55,26 @@ public interface EmbeddingStore<Embedded> {
...
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
     /**
      * Searches for the most similar (closest in the embedding space) {@link Embedding}s.
```
langchain4j, line 99: What about adding `removeAll()` that will remove all embeddings?
I guess this might be useful in many cases:
- original datasource was updated and one wants to simply re-ingest everything instead of figuring out what was changed and what should be removed and re-ingested again
- one might want to change the embedding model and needs to re-ingest everything

Another useful method would be `removeAll(Filter)` which is supported by some embedding stores (e.g. [Milvus](https://milvus.io/docs/insert-update-delete.md#Delete-entities), [Pinecone](https://docs.pinecone.io/guides/data/delete-data#delete-specific-records-by-metadata)), but this is probably for another PR.

WDYT?

ghost (commented):
@humcqc thanks a lot!

langchain4j:
@geoand @jmartisk could you please take a look as well? 🙏

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
humcqc: yes it' was what i was thinking, but wanted to keep implementation from the original committer to discuss it. 
will do all reviews today to finalize quickly.

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
humcqc: In my case i will have a tenant column that will be mandatory, but i will add this method for your use case.
I 100% agree with removeAll(Filter) , can do it in this pr or in another, as you want ?

humcqc: should removeAll(Filter) replace removeAll(List<String> ids) and remove(String id) ? or you prefer keep both ?

I'm not sure id's will be very useful for delete or update  , as we will have a global metadata id to group all embeddings for a given data.

humcqc: if we keep removeAll(Filter) and removeAll(List ids) we will have inconsistency when we call removeAll(null).
WDYT?
I propose tu use:
- remove(String id) -> remove one record, if null no record deleted
- remove(List<String>) -> remove list of records by id, if null no record deleted
- removeByFilter(Filter) -> remove filtered record, if null all record will be deleted

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -55,6 +57,45 @@ public interface EmbeddingStore<Embedded> {
...
+    default void removeAll() {
+        throw new UnsupportedOperationException("Not supported yet.");
+    }
+
+    /**
+     * Removes multiple embeddings from the store.
+     *
+     * @param ids A list of unique identifiers of the embeddings to be removed.
+     */
+    default void removeAll(List<String> ids) {
```
langchain4j, line 81:
```suggestion
    default void removeAll(Collection<String> ids) {
```
WDYT?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -55,6 +57,45 @@ public interface EmbeddingStore<Embedded> {
...
+    /**
+     * Removes multiple embeddings from the store.
+     *
+     * @param filter         The filter to be applied to the {@link Metadata} during delete.
+     *                       Only {@link TextSegment}s whose {@link Metadata}
+     *                       matches the {@link Filter} will be deleted.
+     *                       Please note that not all {@link EmbeddingStore}s support this feature yet.
+     *                       This is an optional parameter. Default: no filtering, all will be deleted.
+     */
+    default void removeAll(Filter filter) {
```
langchain4j, line 94:
```suggestion
    default void remove(Filter filter) {
```
WDYT?

langchain4j:
@geoand @jmartisk could you please check? Does this API makes sense? I would love to include this in the release today.

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
humcqc: I propose to use:

remove(String id) -> remove one record, if null no record deleted
remove(Collection) -> remove list of records by id, if null no record deleted
removeByFilter(Filter) -> remove filtered records, if null all record will be deleted

or
remove(String... ids) -> remove records by id, if null no record deleted
removeByFilter(Filter) -> remove filtered records, if null all record will be deleted or removeByMetadata(Filter)

removeByMetadata(Filter) seems more adapted.

@langchain4j WDYT ?

geoand:
LGTM, but I'll let @jmartisk comment further since he has interacted with the `EmbeddingStore` much more than I have

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
jmartisk: +1, `all` suggests really all

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -54,6 +56,45 @@ public interface EmbeddingStore<Embedded> {
...
+    }
+
+    /**
+     * Removes multiple embeddings from the store.
+     *
+     * @param filter         The filter to be applied to the {@link Metadata} during delete.
+     *                       Only {@link TextSegment}s whose {@link Metadata}
+     *                       matches the {@link Filter} will be deleted.
+     *                       Please note that not all {@link EmbeddingStore}s support this feature yet.
+     *                       This is an optional parameter. Default: no filtering, all will be deleted.
```
jmartisk, line 92: "Default" I assume means `filter==null`, so the docs should rather say that instead of "default"

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -238,6 +241,78 @@ public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedde
...
+    /**
+     * Removes multiple embeddings from the store.
+     *
+     * @param ids A list of unique identifiers of the embeddings to be removed.
+     */
+    @Override
+    public void removeAll(List<String> ids) {
+        if (ids != null && !ids.isEmpty()) {
+            try (Connection connection = getConnection()) {
+                PreparedStatement statement = connection.prepareStatement(String.format(
```
jmartisk, line 270: PreparedStatement needs closing, so this needs to be wrapped in a try-with-resources too

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
jmartisk: On the other hand, `java.util.Collection.removeAll` also removes just some elements, so maybe we could leave it like this for consistency

langchain4j: Good point, one can read this as "remove all that satisfy this filter"

humcqc: I'm not sure about removeAll(Collection). and removeAll(Filter) 
If we keep this way, removeAll(null) will have 2 opposite behaviors:
-> https://github.com/langchain4j/langchain4j/pull/1113/files#diff-40e560d628f612b7350552586991a852f85d9a2c35f15604a4d345bc13209912R116
-> https://github.com/langchain4j/langchain4j/pull/1113/files#diff-40e560d628f612b7350552586991a852f85d9a2c35f15604a4d345bc13209912R133

@langchain4j , @jmartisk WDYT about:
removeAll() -> remove all
remove(String... ids) -> remove records by id ( one ore +), if null no record deleted
removeByMetadata(Filter) -> remove filtered records, if null all record will be deleted;

ghost (approved):
@humcqc thanks a lot!
I will merge it now and do some changes in a separate commit, as I can't push changes to your fork.

humcqc:
humm, I had the review fixes in my local repo waiting for the answer on my last comment:: 
I'm not sure about removeAll(Collection). and removeAll(Filter)
If we keep this way, removeAll(null) will have 2 opposite behaviors:
-> https://github.com/langchain4j/langchain4j/pull/1113/files#diff-40e560d628f612b7350552586991a852f85d9a2c35f15604a4d345bc13209912R116
-> https://github.com/langchain4j/langchain4j/pull/1113/files#diff-40e560d628f612b7350552586991a852f85d9a2c35f15604a4d345bc13209912R133

@langchain4j , @jmartisk WDYT about:
removeAll() -> remove all
remove(String... ids) -> remove records by id ( one ore +), if null no record deleted
removeByMetadata(Filter) -> remove filtered records, if null all record will be deleted;

langchain4j:
> humm, I had the review fixes in my local repo waiting for the answer on my last comment::

@humcqc my bad, pardon!
I am a bit in a rush, wanted it to make it in this release that I am preparing now.
BTW, I did not see a comment from you. Maybe it is pending? If yes, I can't see it...

I have [changed](https://github.com/langchain4j/langchain4j/commit/358f63c0faa4e56385f53ec89fe806496cf686f9) it so that null/empty ids/filter are not valid inputs.

humcqc:
No pb! if the current version is ok for you it's ok for me.

> BTW, I did not see a comment from you. Maybe it is pending? If yes, I can't see it...

:) My bad so, didn't notice the pending stuff, I was a little bit frustrated to not have answers :)
How can I change this pending stuff ?

langchain4j:
@humcqc yeah, I made this mistake many times 😆 You need to go to files changed -> review changes -> submit

humcqc (commented):
:) a little bit late :)

Issue #968, opened 2024-04-18
[FEATURE] AI Service: Add an option to register a custom OutputParser


langchain4j:
Add an option to register a custom `OutputParser` when building AI Service.

anunnakian:
Hello, are you thinking about something like that : 

```java
OutputParser outputParser = new MyOutputParser(); // my custom implementation
Assistant assistant = AiServices.builder(Assistant.class)
                .chatLanguageModel(model)
                .tools(tool)
                .outputParser(outputParser) 
                .build();
```

langchain4j:
Yes, we should think about how it should behave, when a custom `OutputParser` is provided (or even multiple?).
AI Service can also have multiple methods with different return types, so we should consider this as well.

langchain4j:
Related: https://github.com/langchain4j/langchain4j/issues/1087

langchain4j:
Related: https://github.com/langchain4j/langchain4j/discussions/786

anunnakian:
I'm thinking about making something like that : 

```java
interface Chat {

    @Moderate
    String chat(String text); // will use default parse() method
    
    @Output(MyCustomOutputParser.class) // will use a custom output parser
    Answer getAnswer(String question);
}

```

PR #1126, opened 2024-05-18
Add the ability to configure a custom output parser

anunnakian:
## Issue
https://github.com/langchain4j/langchain4j/issues/968

## Change
Add the ability to configure a custom output parser, here is an example : 

First of all, I define `City` class and my custom output parser by implementing `OutputParser` interface

```java
@Data
@AllArgsConstructor
static class City {
    private String name;
}

static class CityOutputParser implements OutputParser<City> {

    @Override
    public City parse(String string) {
        return new City(string);
    }

    @Override
    public String formatInstructions() {
        return "A name of a city";
    }
}
```

Then, I use my custom output parser by adding an annotation on my method called `@Output` like this : 

```java
interface AiService {

        @Output(CityOutputParser.class)
        @UserMessage("What is the capital of Germany?")
        City chat();
}
```

Here is a test : 

```java
@Test
void test() {

    // given
            .build();

    // when-then
    assertThat(aiService.chat()).isEqualTo(new City("Berlin"));
    verify(chatLanguageModel).generate(singletonList(userMessage("What is the capital of Germany?A name of a city")));
}
```

## General checklist


langchain4j:
Related: https://github.com/langchain4j/langchain4j/pull/1176

PR #1127, opened 2024-05-18
Adding missing -azure-ai-search and azure-cosmos-mongo-vcore to the BOM

agoncal:
I've adding two missing artefacts in the BOM (azure-ai-search and azure-cosmos-mongo-vcore). And to make it easier to spot, I've sorted the `pom.xml` files by artifact id


ghost (approved):
@agoncal thank you!

We should probably automatically generate the bom

PR #1133, opened 2024-05-20
Fix #1124: preserve metadata during query compression

### LINKED ISSUE #1124: [BUG] Query metadata are lost when using the query compression technique
Hi

I have created an assistant using a EmbeddingStoreContentRetriever based on a dynamic metadata filter, like this:

Function<Query, Filter> filterByUserId =
                (query) -> metadataKey("userId").isEqualTo(query.metadata().chatMemoryId().toString());

(https://github.com/langchain4j/langchain4j-examples/blob/main/rag-examples/src/main/java/_3_advanced/_05_Advanced_RAG_with_Metadata_Filtering_Examples.java)

The filter works fine, but when I added a query compression step, and I chat to the assistant for the second time, it fails because query.metadata() is always null. It seems that during query compression, the metadata of the query are missing.  

- LangChain4j version: 0.30
- LLM(s) used:  LLama3
- Java version:  21
- Spring Boot version: 3.2.4 

thanks

ghost:
Fixes #1124

This pull request addresses the issue of query metadata being lost during passing it through the `CompressingQueryTransformer` and `ExpandingQueryTransformer`.

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/rag/query/transformer/CompressingQueryTransformer.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/rag/query/transformer/CompressingQueryTransformer.java
@@ -71,8 +71,11 @@ public class CompressingQueryTransformer implements QueryTransformer {
         }
 
         Prompt prompt = createPrompt(query, format(chatMemory));
-        String compressedQuery = chatLanguageModel.generate(prompt.text());
-        return singletonList(Query.from(compressedQuery, query.metadata()));
+        String compressedQueryText = chatLanguageModel.generate(prompt.text());
+        Query compressedQuery = query.metadata() == null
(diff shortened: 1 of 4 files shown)


langchain4j-core/src/main/java/dev/langchain4j/rag/query/transformer/CompressingQueryTransformer.java
```diff
@@ -72,7 +72,7 @@ public Collection<Query> transform(Query query) {
 
         Prompt prompt = createPrompt(query, format(chatMemory));
         String compressedQuery = chatLanguageModel.generate(prompt.text());
-        return singletonList(Query.from(compressedQuery));
+        return singletonList(Query.from(compressedQuery, query.metadata()));
```
langchain4j: Fix the same for expanding query transformer

Issue #1125, opened 2024-05-18
[BUG] AI Service fails when user message contains template variables


langchain4j:
**Describe the bug**
AI Service fails when user message contains template variables.

**Log and Stack trace**
```
dev.langchain4j.exception.IllegalConfigurationException: Error: cannot find the value of the prompt template variable "{{it}}".
    at dev.langchain4j.exception.IllegalConfigurationException.illegalConfiguration(IllegalConfigurationException.java:12)
    at dev.langchain4j.service.DefaultAiServices.getValueOfVariableIt(DefaultAiServices.java:247)
    at dev.langchain4j.service.DefaultAiServices.findTemplateVariables(DefaultAiServices.java:223)
    at dev.langchain4j.service.DefaultAiServices.prepareUserMessage(DefaultAiServices.java:258)
```

**To Reproduce**
```java
@AiService
public interface AiAssistant {

    TokenStream chat(@MemoryId String chatId, @UserMessage String userMessage);
}

assistant.chat("12345", "Text containing {{it}}");
```

**Please complete the following information:**
- LangChain4j version: latest

KaisNeffati:
Hey ! Could you please clarify the expected behavior for this issue ?

langchain4j:
Hi @KaisNeffati, it should not fail. In this case `{{it}}` (or any other variable) should not be treated as variable, but as a text

anunnakian:
Hello @KaisNeffati, are you going to work on this issue ?

KaisNeffati:
@anunnakian I was planning to do that, but I didn't have the time. Feel free to go for it if you'd like to

anunnakian:
Ok thanks! @langchain4j could you assign me to this issue, I'll work on it today

langchain4j:
Not a bug, current behavior is correct

lylala8:
This is a very big issue, if users manually enter template symbols, all LangChin4j constructed projects will report errors

PR #1137, opened 2024-05-21
Treat messages with missing variables as text

anunnakian:
## Issue
https://github.com/langchain4j/langchain4j/issues/1125

## Change
When we declare an assistant as following example : 

```java
interface AiService {

    @UserMessage("What is the capital of {{it}}?")
    String answer();
}
```

We don't throw an illegal exception like before, we treat the user message like text without resolving the `{{it}}` variable.

Here is a test that explain the treatment clearly : 

```java
@Test
void test_user_message_configuration_8() {

    // given
            .build();

    // when-then
    assertThat(aiService.answer()).containsIgnoringCase("");
    verify(chatLanguageModel).generate(singletonList(userMessage("What is the capital of {{it}}?")));
}
```

## General checklist


langchain4j:
Hi @anunnakian thanks a lot!

This seems to be a bit different from what was described in https://github.com/langchain4j/langchain4j/issues/1125


This is definitely a wrong configuration:
```java
interface AiService {

    @UserMessage("What is the capital of {{it}}?")
    String answer();
}
```

The case described in the issue is different:
```java
@AiService
public interface AiAssistant {

    TokenStream chat(@MemoryId String chatId, @UserMessage String userMessage);
}

assistant.chat("12345", "Text containing {{it}}");
```

anunnakian:
@langchain4j if we consider that the issue was fixed, the following tests must be green right ? 

```java
@Test
void test_user_message_configuration_8() {

    // given
            .chatMemoryProvider(memoryId -> MessageWindowChatMemory.withMaxMessages(10))
            .build();

    // when
    aiService.chat8("12345", "What is the capital of {{it}}?");

    // then
    verify(chatLanguageModel).generate(singletonList(userMessage("What is the capital of {{it}}?")));
}

@Test
void test_user_message_configuration_9() {

        // given
                .chatMemoryProvider(memoryId -> MessageWindowChatMemory.withMaxMessages(10))
                .build();

        // when
        aiService.chat8("12345", "What is the capital of {{variable}}?");

        // then
        verify(chatLanguageModel).generate(singletonList(userMessage("What is the capital of {{variable}}?")));
    }
```

anunnakian:
We should remove this test case, if the missing variable {{name}} must be treated like text, right ?

 ```java
@Test
void should_fail_when_value_is_missing() {

    // given
    PromptTemplate promptTemplate = PromptTemplate.from("My name is {{name}}.");

    Map<String, Object> variables = emptyMap();

    // when-then
    assertThatThrownBy(() -> promptTemplate.apply(variables))
            .isExactlyInstanceOf(IllegalArgumentException.class)
            .hasMessage("Value for the variable 'name' is missing");
}
```

**Comment by langchain4j** [comment]
@anunnakian to your first question: yes.
Second: not sure, is there a way to treat it as a text (instead of template) only for that use case?

**Comment by anunnakian** [comment]
@langchain4j If we treat the message as text instead of a template, we will break the case if we have two variable with one missing like the following test : 

```java
interface AiService {

    String chat9(@MemoryId String chatId, @UserMessage String userMessage, @V("country") String country);
}

@Test
void test_user_message_configuration_10() {

    // given
            .chatMemoryProvider(memoryId -> MessageWindowChatMemory.withMaxMessages(10))
            .build();

    // when
    aiService.chat9("12345", "What is the {{it}} of {{country}}?", "Germany");

    // then
    verify(chatLanguageModel).generate(singletonList(userMessage("What is the {{it}} of Germany?")));
}
```

This test must be green too at the end, right ?

anunnakian:
@langchain4j I found a solution to make all tests above and the following one green! 
my branch is up to date and waiting for your feedback

```java
interface AiService {
    String chat(@MemoryId String chatId, @UserMessage String userMessage);
}

@Test
void test_user_message_configuration_9() {

    // given
            .chatMemoryProvider(memoryId -> MessageWindowChatMemory.withMaxMessages(10))
            .build();

    // when
    aiService.chat("12345", "What is the {{it}} of {{var}}?");

    // then
    verify(chatLanguageModel).generate(singletonList(userMessage("What is the {{it}} of {{var}}?")));
}
```

langchain4j-core/src/main/java/dev/langchain4j/spi/prompt/PromptTemplateFactory.java
```diff
@@ -33,6 +35,14 @@ interface Template {
          * @return the rendered template.
          */
         String render(Map<String, Object> variables);
+
+        /**
+         * Get all variables extracted from the template.
+         * @return A set of variable names.
+         */
+        default Set<String> getAllVariables() {
```
anunnakian, line 43: A make this method `default` to avoid breaking changes

ashimoon:
Should we just always treat missing variables, including ``{{it}}`` as an illegal state? I would imagine in the vast majority of cases, if the author created a `@UserMessage` with a prompt variable in it, then the method signature must match that user message, otherwise an exception should be thrown.

If we do indeed want to support `What is the capital of {{it}}?` where we want the string to be supported literally without any templating, then could we just add a property to `@UserMessage` to indicate that?

e.g.

```java
@UserMessage(value = "What is the capital of {{it}}?", templated = false)

public @interface UserMessage {
    String value()
    boolean templated() default true;
}
```

If we need even further grained control, such that only some variables are excluded from templating, that could also be defined in the annotation?

```java
@UserMessage(value = "What is the capital of {{it}} in {{country}}?", excludedVariables = { "it" })

public @interface UserMessage {
    String value()
    String[] excludedVariables() default {};
}
```

langchain4j:
@ashimoon yes, I think you are right. We should be strict by default (as it is now).

I think I rushed a bit with https://github.com/langchain4j/langchain4j/issues/1125. It can be solved by escaping `{{it}}` instead of making changes to the default behaviour of prompt templates.

@anunnakian sorry about that, let's put this PR on hold for now

anunnakian:
Ok! Don't panic everything is under control (just kidding... 😁)

So, I'll create another PR to escape {{it}} variable tonight

Thanks for your feedback @ashimoon @langchain4j 😉

langchain4j:
@anunnakian I meant that the current implementation (without this PR) indeed seems to be the most reasonable. I guess that specific corner case in https://github.com/langchain4j/langchain4j/issues/1125 is actually ok and the user can escape manually, if there is a need. Whether we should support auto-escaping, it is tricky. I would take a break and return to this problem a bit later.

anunnakian:
@langchain4j got it ;)

lylala8:
This is a very big problem. If users manually enter template symbols, won't all LangChain4j constructed projects report errors

lylala8:
这是一个非常大的问题，如果用户手动输入模板符号，所有LangChin4j构造的项目都会报错

PR #1144, opened 2024-05-22
Support GTP-4o with Azure OpenAI

jdubois:
GPT-4o on Azure OpenAI was announced at Build yesterday, and it's already working with LangChain4J!

This shows that the current code works with GPT-4o (all integration tests are green on my side), with 2 downsides for the moment:

- It's not available in the Sweden Central region, which I typically use as there are more models there: this is why I'm using East US, and had to remove GPT 3.5 Turbo which isn't available there (then, GPT-4o basically replaces both GPT 3.5 Turbo and GPT 4, so it makes sense)
- The CLI commands in the `deploy-azure-openai-models.sh` script currently complains that the API version isn't correct, as it's not updated yet. It should be super soon, and otherwise the trick is to create it using the Azure AI Studio.

Those 2 issues should be very temporary.


ghost (approved):
@jdubois thanks a lot!

langchain4j:
@jdubois do you know if gpt-4o is coming to `swedencentral` soon?

jdubois:
Sorry, no! And if I did, I couldn't tell publicly 😀

PR #1148, opened 2024-05-23
Implementation of a lambda based streaming response handler

glaforge:
The idea is to be able to have a couple utility static methods developers can import to be able to use lambdas to handle streamed tokens coming from the model they use:

```java
model.generate("Why is the sky blue?",
       onNext(text -> System.out.println(text));
model.generate("Why is the sky blue?",
       onNext(System.out::println);
model.generate("Why is the sky blue?",
       onNextAndError(System.out::println, Throwable::printStackTrace));
```

### DIFF
--- a/langchain4j-core/src/test/java/dev/langchain4j/model/LambdaStreamingResponseHandlerTest.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/model/LambdaStreamingResponseHandlerTest.java
@@ -21,8 +21,8 @@ public class LambdaStreamingResponseHandlerTest implements WithAssertions {
         List tokens = new ArrayList<>();
         tokens.add("The sky ");
         tokens.add("is blue because of ");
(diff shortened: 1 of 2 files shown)


langchain4j-core/src/test/java/dev/langchain4j/model/LambdaStreamingResponseHandlerTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @Test
+    void testOnNextAndError() {
+        // given
+        List tokens = new ArrayList<>();
+        tokens.add("Three ");
+        tokens.add("Two ");
+        tokens.add("One ");
+        tokens.add(new RuntimeException("BOOM"));
```
langchain4j, line 45: 😆

ghost (approved):
@glaforge thank you, this looks nice!

PR #1161, opened 2024-05-24
Fix null pointer issue in SqlDatabaseContentRetriever when column value is null

hongliangzhang07:
This pull request addresses the null pointer exception issue in the `SqlDatabaseContentRetriever` class that occurs when a certain column value from a database query is null. The code now includes a check for null values before attempting to access them.


ghost (approved):
@hongliangzhang07 thank you!

PR #1174, opened 2024-05-26
Milvus: improve insert performance

ghost:
## Context
Currently, `MilvusEmbeddingStore` performs [flush](https://milvus.io/api-reference/pymilvus/v2.4.x/ORM/Collection/flush.md) after each insert.
Each flushing takes around 3.5 seconds, which significantly impacts the performance.
The documentation also advises agains this:
> can i call `flush()` after every data insertion?
> 
> When new data is inserted, it is written into a growing segment. Once the size of a growing segment reaches its upper limit, Milvus automatically seals the segment.
> 
> Continuously calling this operation results in many sealed segments of small sizes, which can gradually degrade search performance.
> 
> It is recommended that you wait for Milvus to seal all segments before conducting any searches.

## Change

This PR adds a new property `flushAfterInsert` that controls whether to do flush after inserts or not. The default value is `false`.
This is potentially a breaking change because it means that by default, there is no delay (3.5 secs) after each insert, so if someone does search right after insert, the just inserted embedding(s) might not be available yet.

The performance of `EmbeddingStoreWithFilteringIT` was improved as well.

## General checklist

### DIFF
--- a/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithFilteringIT.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithFilteringIT.java
@@ -8,6 +8,7 @@ import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
 
+import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


langchain4j:
Closing due to conflicts, implemented in https://github.com/langchain4j/langchain4j/commit/2c8ff58c028c6d1927d20ec7ad392864114eef6a

PR #1176, opened 2024-05-26
[FEATURE]: Custom output parsing for AI services (fixes TODO)

ashimoon:
This change gives users the ability to provide custom output parsers to AI services.

# Context
Previously, users were constrained to the set of data types that were supported by langchain4j (most primitives, string lists, string sets, enums, and JSON POJOs). Generic lists and sets were not supported.

# Change
Users can now provide a custom ParserProvider when creating an AI service via the builder, or can provide individual OutputParsers or ParserFactory's on each service method with the @Parser annotation.

This enables users to have full control of the output parsing process, while at the same time they continue to benefit from all the existing output parsers that come bundled with langchain4j.

In addition to replacing the default parser provider, users can override existing parser instances (e.g. for boolean or int), or add parsers for their own custom types (e.g. parsing instances of Foo).

Finally, users can add their own ParserFactory instances which will be called if no output parser instance is found, allowing the user to construct a parser instance on the fly for the requried type.

This change also adds support for generic List<> and Set<> types, where the element can be any supported parsable type (e.g. a List<Integer> works by default). Furthermore, a List<Foo> works by parsing each line using the registered parser for Foo (either a custom parser, or falling back to the default JsonParser).

# Checklist
- [X] All unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and
[main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules are green


ashimoon:
Hi @langchain4j - thanks for triaging this issue. Just double checking, are there any steps left for me to do before it goes to you for review? Thanks.

langchain4j:
@ashimoon no, I will try to do a review early next week, thanks for your patience!

langchain4j-core/src/main/java/dev/langchain4j/model/output/OutputParser.java
```diff
@@ -1,21 +1,46 @@
...
-    T parse(String text);
+    T parse(final OutputParsingContext context);
 
     /**
-     * Description of the text format.
-     * @return the description of the text format.
+     * Get the LLM-facing formatting instructions for the output.
+     * @return the formatting instructions.
      */
     String formatInstructions();
```
langchain4j, line 29: Instead of having formatInstructions and prelude/postlude methods, can we have something like
`void injectOutputFormatInstructions(ChatMessage)`? WDYT?

ghost (commented):
@ashimoon thanks a lot, this looks very nice! I did not have enough time to go through all the changes, so this is a preliminary feedback.

I am not sure about so many ways to provide parsers. What do you think about having just one way, the most flexible and configurable? I guess annotations can be dropped.
IIRC, noone asked about this feature explicitly yet, so there is probably no need to have is so flexible.

Thanks!

langchain4j:
Related: https://github.com/langchain4j/langchain4j/pull/1126

ashimoon:
Thank you for your comments. I agree with them, there is room to simplify.
I'll take a look at the related issue too, see what we can combine.

langchain4j:
@ashimoon thank you!

Issue #1000, opened 2024-04-23
[FEATURE] CI: Run integration tests for changed and dependent modules


langchain4j:
Currently [main CI workflow](https://github.com/langchain4j/langchain4j/blob/main/.github/workflows/main.yaml) runs only unit tests, no integration tests are run, because it takes quite a while.

It should be configured to run ITs for modules that have changed in the PR/merge, as well as modules that depend on changed modules.
Ideally parallelize the build.

Optimizing ITs can also be done, but first let's start with optimizing the workflow.

PR #1185, opened 2024-05-28
support run integration tests for changed and dependent modules

Yellow--:
## Issue

https://github.com/langchain4j/langchain4j/issues/1000

## Change

- add new ```Github Workflow``` job ```changes``` to collect the changed modules
- after push or pr,```java_build```will run changed modules IT now

## How
- use [tj-actions/changed-files@v44](https://github.com/tj-actions/changed-files) to track file changed
- group file changed by modules
- add ```-Dskip{moduleName}ITs``` if module not changed

## Need Discuss
- should we make property ```skip{moduleName}ITs``` default to true?

## General checklist


Yellow--:
@langchain4j

Yellow--:
the build failed is expected,cause i can't setup and don't have so many API_KEY for realtive modules

langchain4j:
Hi @Yellow--, thanks a lot! 
I am wondering if this can be simplified if instead of skipping just ITs we will skip the whole module (compile, unit test, etc).
One can use e.g. `-pl langchain4j-core,langchain4j-open-ai` maven parameter to specify modules that should be executed.

WDYT?

Yellow--:
hi @langchain4j 
except the module that dependented (for example langchain4j-core ) by other
we can skip the module build compile

i will try it later in today~

langchain4j:
> except the module that dependented (for example langchain4j-core ) by other
> we can skip the module build compile

What do you mean?
AFAIU, we can always run `langchain4j-core` and `langchain4j` modules and all the modules that were changed in the PR.
This way we don't need all the `skipXyzITs` flags

Yellow--:
hi @langchain4j 
i have updated the main.yaml
as for now,no changed module will skip compile

without changed,the action result may be like [this](https://github.com/Yellow--/langchain4j/actions/runs/9283922376)

langchain4j:
@Yellow-- I guess all the `skipXyzITs` flags can be removed now?

I also have a feeling that it can be simplified even further. Right now, both `filter` and `analyzer` step have mappings. Maybe `filter` can return a list of concatenated modules which can then be concatenated in `analyzer`? This will remove the extra mapping in the `analyzer` step

Yellow--:
> @Yellow-- I guess all the `skipXyzITs` flags can be removed now?
> 
> I also have a feeling that it can be simplified even further. Right now, both `filter` and `analyzer` step have mappings. Maybe `filter` can return a list of concatenated modules which can then be concatenated in `analyzer`? This will remove the extra mapping in the `analyzer` step

oh,i know what u mean
if module changed,all the test included IT should be run
if not changed,the module should skip compile

so no scene to skip specific module IT now

i will remove them later
😀😀😀

by the way,the way to simplified may not easy to do,i need more time to think

Yellow--:
hi @langchain4j 
i have reverted all changed of module's pom.xml and replace all ```-D skipXYZITs``` with ```-pl !{module-path}```

## Change Detail
- reverted all changed of module's pom.xml
- replace all ```-DskipXYZITs``` with ```-pl !{module-path}```
- add ```langchain4j-parent``` as relative dir to track module changed

## Something need to know
- due to action [snapshot_release](https://github.com/langchain4j/langchain4j/blob/main/.github/workflows/snapshot_release.yaml) and [snapshot_release_core_and_parent](https://github.com/langchain4j/langchain4j/blob/main/.github/workflows/snapshot_release_core_and_parent.yaml),all module will be deploy to remote repository
- the module that dependent by other(for example langchain4j-core, langchain4j-parent) if no changed in pr,will skip compile but download from remote repository
- so we can skip any module compile as we like

Yellow--:
> @Yellow-- I guess all the `skipXyzITs` flags can be removed now?
> 
> I also have a feeling that it can be simplified even further. Right now, both `filter` and `analyzer` step have mappings. Maybe `filter` can return a list of concatenated modules which can then be concatenated in `analyzer`? This will remove the extra mapping in the `analyzer` step

hi @langchain4j 
with the help of built-in command ```jq``` on ```ubuntu-latest``` and name convention of mapping name
the extra mapping is remove now

## Mapping Name Examples

| module_path | mapping_name |
| --- | --- |
| document-loaders/langchain4j-document-loader-tencent-cos | document_loaders__langchain4j_document_loader_tencent_cos |
| langchain4j-core | langchain4j_core |

.github/workflows/main.yaml
```diff
@@ -9,7 +9,261 @@ on:
...
+      excluded_module_arg: ${{ steps.analyzer.outputs.excluded_module_arg }}
+    steps:
+      - uses: actions/checkout@v4
+      # collect changed file group by module
+      - uses: tj-actions/changed-files@v44
+        id: filter
+        with:
+          # each module contained IT should list the dir and direct/indirect dependent module's dir in the files_yaml
+          # the variable of module name should replace path separator with '__' and '-' should replace with '_'
+          files_yaml: |
```
langchain4j: It is not ideal that this list has to be maintained manually, but at least we have to be sure it is up to date. If someone will add new module to the repository, but not to this list, will this build fail (it should)?

Yellow--: good question!!

```main.yaml``` will skip all IT now(before this pr)
after this changed,if new module not add to this list,may cause build fail for the first commit and any future commit relative to the new module

if merge the pr now,we will need some one time job for the new module(update this list)

Comment (Yellow--) [reply .github/workflows/main.yaml]: by the way,due to the new module added is not usally happened,the work to update this list may be acceptable
WDYT?

**Comment by langchain4j** [comment]
@Yellow-- I have learned from @gsmet about https://github.com/gitflow-incremental-builder/gitflow-incremental-builder, seems to be exactly what we need. Could you please check if this can be used instead of a manual setup? 🙏

**Comment by Yellow--** [comment]
@langchain4j sure,i will check it later

**Comment by gsmet** [comment]
@famod maybe you could have a look at the GIB integration here when you have some time.

**Comment by Yellow--** [comment]
hi @langchain4j 
i have use [gib](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder) to achive this work

## Changed
- add [gib](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder) as extension to the root ```pom.xml```
- override property ```gib.disable``` to true as default
- add a new step to fetch push/pull_event relative base sha as branch for later compare(it seem noway to use ```SHA``` as gib.referenceBranch directly)
- remove the ```-DskipIT``` now
- <del>build command add extra args ```-Dgib.disable=false -Dgib.buildDownstream=false -Dgib.buildUpstream=true -Dgib.referenceBranch=__branch_before```</del>
- build command add extra args ```-Dgib.disable=false -Dgib.referenceBranch=__branch_before```

## Thing need to know
- ```gib.disable``` should be true as default [link](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder?tab=readme-ov-file#disable-in-ide)
- cause we may build project with java8,we need to use the 3.x version of this plugin [link](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder?tab=readme-ov-file#requirements)
- <del>```gib.buildDownstream``` will set to false when build to avoid downstream module build(e.g. open-api module changed, core module should not build)</del>
- <del>```gib.buildUpstream``` will set to true when build to include direct upstream module build(e.g. core changed, open-api module should build)</del>

**Comment by langchain4j** [comment]
@Yellow-- thanks a lot! I will take a closer look tomorrow

**Comment by famod** [comment]
> @famod maybe you could have a look at the GIB integration here when you have some time.

Thanks for the ping! I'll try to find some time in the next few day.

**Comment by langchain4j** [comment]
Hi @famod did you have a chance to take a look? 🙏

File: .github/workflows/main.yaml [hunk 2]
```diff
@@ -25,14 +25,23 @@ jobs:
...
+          fi
       - name: Set up JDK ${{ matrix.java_version }}
         uses: actions/setup-java@v4
         with:
           java-version: ${{ matrix.java_version }}
           distribution: 'temurin'
           cache: 'maven'
       - name: Build with JDK ${{ matrix.java_version }}
-        run: mvn --batch-mode install -DskipITs ${{ matrix.included_modules }}
+        run: mvn --batch-mode install ${{ matrix.included_modules }} -Dgib.disable=false -Dgib.buildDownstream=false -Dgib.buildUpstream=true -Dgib.referenceBranch=__branch_before
```
famod: I'm surprised by `buildDownstream=false`. This means that you won't notice if B (which depends on A) breaks because of some incompatible change in A.
This is especially surprising because this `mvn` call does not seem to be only about ITs.

famod:
>     * cause we may build project with java8,we need to use the 3.x version of this plugin [link](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder?tab=readme-ov-file#requirements)

That's a bit unfortunate since 3.x is basically dead; it hasn't seen an update in years and I'm not actually planning one, TBH.
I don't know whether I could even still deploy something under the old coordinates.

Anyway, 3.15.0 can still be used but its de facto EOL is something to be aware of.
 
>     * `gib.buildUpstream` will set to true when build to include direct upstream module build(e.g. core changed, open-api module should build)

Just a note: It's not just the direct upstream modules, it's transitive (upstreams of upstreams...).

Btw, by default GIB will only skip tests for those upstream modules. You might want to skip more plugins like checkstyle, spotbugs etc. (if any). See [`argsForUpstreamModules`](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder/tree/3.x?tab=readme-ov-file#gibargsforupstreammodules).

famod:
> I don't know whether I could even still deploy something under the old coordinates.

Just checked, I'm still able to deploy. So I could push a 3.16.0 in the next few weeks.

.github/workflows/main.yaml
Yellow--: hi @famod ,thanks for your review

with some research before
i found that
- there are 3 ```workflow``` realtive to this issues
- [main.yaml](https://github.com/langchain4j/langchain4j/blob/main/.github/workflows/main.yaml) for ensure compile success and test passed when ```commits``` or ```pull request``` push to ```main``` 
- [snapshot_release.yaml](https://github.com/langchain4j/langchain4j/blob/main/.github/workflows/snapshot_release.yaml) for build and deploy all modules to remote repository except ```langchain4j-core``` and ```langchain4j-parent``` when commits push to branch ```main```
- [snapshot_release_core_and_parent.yaml](https://github.com/langchain4j/langchain4j/blob/main/.github/workflows/snapshot_release_core_and_parent.yaml) for build and deploy module ```langchain4j-core``` and ```langchain4j-parent``` to remote repository
- due to the [comment](https://github.com/langchain4j/langchain4j/pull/1185#issuecomment-2136771673), @langchain4j may want to skip the changed module's submodule build、test if we could achive that
- if we skip module dependent by other(e.g. langchain4j-core),with the help of ```snapshot_release.yaml``` and ```snapshot_release_core_and_parent.yaml```,we could fetch them from remote instead of build them locally

so i set ```buildDownstream=false``` explicitly,to avoid submodule build
@famod should i remove them?i am not really sure my usage is correct here

Yellow--: @famod 
after reread the doc
i found i make a stupid mistake
i reversved the meaning of UpstreamModule and DownstreamModule
the default property value of ```gib.buildDownstream``` and ```gib.buildUpstream``` is the behaviour i want
i will make a new commit again

could you help me to check it again?thks

.github/workflows/main.yaml
```diff
@@ -41,14 +41,23 @@ jobs:
...
+          fi
       - name: Set up JDK ${{ matrix.java_version }}
         uses: actions/setup-java@v4
         with:
           java-version: ${{ matrix.java_version }}
           distribution: 'temurin'
           cache: 'maven'
       - name: Build with JDK ${{ matrix.java_version }}
-        run: mvn --batch-mode install -DskipITs ${{ matrix.included_modules }}
+        run: mvn --batch-mode install ${{ matrix.included_modules }} -Dgib.disable=false -Dgib.referenceBranch=__branch_before
```
Yellow--: hi @famod, i remove the gib.buildDownStream and gib.buildUpstream now
no more different compare to before

gsmet:
> cause we may build project with java8

@langchain4j could we discuss this? Most projects have moved to Java 17 as a minimum nowadays so supporting Java 8 for AI which is such a novel use case looks like a bit of a stretch?
I think you could at least move to Java 11 if not Java 17.

Of course, that's my humble opinion from someone not that familiar with the project. But I think it's worth questioning this from time to time.

langchain4j:
@gsmet yes, we could/should move to 17, but I was hesitant to do so because there was no major reason so far, and around 10% of users still use java 8

gsmet:
Yeah so to give us the perspective we had in Quarkus (and I'm pretty sure Spring had the same when they moved to Spring 6): you will always have people using Java 8 (pretty sure you still have users using Java 1.5 or 1.6...). But are they your users?

The question is: how many of these Java 8 users will start a new AI project or integrate AI in their good old Java 8 apps? And to that I think the answer is: `not that many`.

I understand your position of trying to be kind to as many people as possible. But given all major frameworks moved to Java 17 in their new versions, most new applications (and so most of LangChain4j users) will use Java 17.

Now you could start by moving to Java 11 if that's good enough for GIB (@famod ?) if you want to be conservative and not exclude too many people (and I can understand that!) but I think the reasoning would be the same for Java 11.

Now again, that's just feedback from the reasoning we had when we took this decision for Quarkus. It might not be completely the same for your project.

langchain4j:
@gsmet thank you! Yeah, I guess we just need to cut it... Since we plan to release 1.0 this fall, do you think it is a good time for it?

gsmet:
I would say the earlier the better. You are at the early stage of a very innovative project, that's when these decisions are the easiest :).

It won't improve over time :).

gsmet:
Also I recommend you communicate your intentions early as that helps a lot in getting this sort of decisions accepted.

langchain4j:
@gsmet thanks, good point!

famod:
> Now you could start by moving to Java 11 if that's good enough for GIB (@famod ?)

It is!

.github/workflows/main.yaml
famod: Ok, makes more sense now.

One crucial thing to beware of is that full (or almost full) builds will take much longer now than before.
This can happen when e.g. `langchain4j-parent/pom.xml` is changed.
Are timeouts, memory sizings etc. ready for such a worst case scenario?

famod:
@Yellow-- have you tested this change (in your fork)?

What I usually did in Quarkus was to branch off my PR branch (that was adding GIB or was changing something significant w.r.t. GIB) and in that test branch I touched an arbitrary file in an arbitrary module.
I then pushed that test branch and created a PR for it to see what CI (and GIB) was making of it.

.github/workflows/main.yaml
famod: Those `-pl !...` options in `included_modules` probably mandate [`disableSelectedProjectsHandling`](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder?tab=readme-ov-file#gibdisableselectedprojectshandling).

I'll have to take a closer look.

famod:
You might want to consider _not_ running GIB for `main` for some time to catch unwanted side effects e.g. PR build green but `main` is broken.
This can be done via `-Dgib.disableIfBranchMatches=main` but you should also add back `-DskipITs` (I guess?). So it's probably better to check if running a PR build and
- if yes: `-Dgib.disable=false`
- if not: `-DskipITs` (GIB is disabled by default in pom.xml)

The oddity of this approach would be that a PR build might run more tests than respective main build does, but well... :man_shrugging:

Yellow--:
> @Yellow-- have you tested this change (in your fork)?
> 
> What I usually did in Quarkus was to branch off my PR branch (that was adding GIB or was changing something significant w.r.t. GIB) and in that test branch I touched an arbitrary file in an arbitrary module. I then pushed that test branch and created a PR for it to see what CI (and GIB) was making of it.

sure,i forked the branch,then edit main.yaml and slim it for fast test;
but just test some common case(e.g. langchain4j-core chagned，langchain4j-openapi changed)

.github/workflows/main.yaml
Yellow--: > One crucial thing to beware of is that full (or almost full) builds will take much longer now than before. This can happen when e.g. `langchain4j-parent/pom.xml` is changed. Are timeouts, memory sizings etc. ready for such a worst case scenario?

@langchain4j i can't run all moudle build with IT enabled,may be you could help to check this scenario?

Yellow--:
> You might want to consider _not_ running GIB for `main` for some time to catch unwanted side effects e.g. PR build green but `main` is broken. This can be done via `-Dgib.disableIfBranchMatches=main` but you should also add back `-DskipITs` (I guess?). So it's probably better to check if running a PR build and
> 
> * if yes: `-Dgib.disable=false`
> * if not: `-DskipITs` (GIB is disabled by default in pom.xml)
> 
> The oddity of this approach would be that a PR build might run more tests than respective main build does, but well... 🤷‍♂️

@langchain4j 
should we consider the scenario ```PR build green but `main` is broken``` ?
in my opinion,the main should always build success.if not,wo should correct it and then merge/rebase ```pr``` on the new ```main``` for run gitflow again

by the way, after view the conversation until now;should we make this pr compatable with only java versoin >= 11?
if so,we could dependent to the latest stable gib plugin and no need to changed it later

langchain4j:
Hi @Yellow-- 

> should we consider the scenario PR build green but `main` is broken ?

I'm not sure we need a test/build after merging PR into `main` at all. If we run all unit+integration tests of all changed and related modules before the merge, and enforce PR branch to be up-to-date, then we should be safe.
Once a day the `nightly.yaml` workflow will run all the unit+integration tests anyway.
The only thing that we can do after PR is merged is to deploy new SNAPSHOT artifacts. BTW this workflow (`snapshot_release*.yaml`) might also benefit from GiB.

.github/workflows/main.yaml
langchain4j: @Yellow-- to check this, could you please do some change in `langchain4j-parent/pom.xml`? I can trigger the build then if you won't be able to do so

Yellow--: hi @langchain4j , i have add some safe changed to ```langchain4j-parent/pom.xml```

langchain4j: @Yellow-- thanks! I've started the builds

famod:
FWIW: I decided to update the 3.x branch of GIB, so there should be a 3.16.0 in, say, 1-2 weeks.

langchain4j:
Hi @famod, great! Should we wait till then or can we merge this as it is now?

famod:
Hi @langchain4j, I don't think you need to wait. There are no critical issues in GIB 3.15.0, AFAICS.
There is https://nvd.nist.gov/vuln/detail/CVE-2023-4759, but I don't think it applies here.

ghost (approved):
@Yellow-- @famod thanks a lot, great job! I will merge it and test on new PRs

famod:
Nice to see this merged!

FYI: I had to put aside GIB 3.16.0 for now due to lack of time but I'd like to continue soon.

famod:
FWIW, I've finally managed to release GIB 3.16.0: https://github.com/gitflow-incremental-builder/gitflow-incremental-builder/releases/tag/version%2F3.16.0

langchain4j:
@famod greawesomeat!

We finally plan to move to java 17 baseline in a few weeks though :D

Issue #1110, opened 2024-05-15
[BUG] IllegalException"text cannot be null" when using Qianfan Model 


yao-wenbin:
**Describe the bug**

when I using qianfan model, If api reach the limited, Model's generate() method will throw Illegal Exception instead of clear err  response.

**Log and Stack trace**

```java
Exception in thread "main" java.lang.IllegalArgumentException: text cannot be null
	at dev.langchain4j.internal.Exceptions.illegalArgument(Exceptions.java:19)
	at dev.langchain4j.internal.ValidationUtils.ensureNotNull(ValidationUtils.java:50)
	at dev.langchain4j.internal.ValidationUtils.ensureNotNull(ValidationUtils.java:37)
	at dev.langchain4j.data.message.AiMessage.<init>(AiMessage.java:30)
	at dev.langchain4j.data.message.AiMessage.from(AiMessage.java:116)
	at dev.langchain4j.data.message.AiMessage.aiMessage(AiMessage.java:146)
	at dev.langchain4j.model.qianfan.InternalQianfanHelper.aiMessageFrom(InternalQianfanHelper.java:162)
	at dev.langchain4j.model.qianfan.QianfanChatModel.generate(QianfanChatModel.java:131)
	at dev.langchain4j.model.qianfan.QianfanChatModel.generate(QianfanChatModel.java:95)
	at com.github.yaowenbin.functioncalling.GetWeatherExample.main(GetWeatherExample.java:32)
```

**To Reproduce**

Using qianfan model and provide a reached limit api key and then call model's generate().

```java
    var model = QianfanChatModel.builder()
            .modelName("ERNIE-Bot")
            .apiKey(System.getenv("QIANFAN_API_KEY"))
            .secretKey(System.getenv("QIANFAN_SECRET_KEY"))
            .build();
        UserMessage userMessage = UserMessage.from("What is the square root of 475695037565?");
        Response<AiMessage> response = model.generate(Collections.singletonList(userMessage));

```

**Expected behavior**

If API reached limit, generate() method should throw a more clear exception instead of IllegalException: text cannot be null.

**Please complete the following information:**
- LangChain4j version: 3.0.0
- LLM(s) used:  Qianfan
- Java version:  17
- Spring Boot version (if applicable): 

**Additional context**

yao-wenbin:
[...]

PR #1186, opened 2024-05-28
Qianfan: clear throw exception when call qianfan api has error code （…

yao-wenbin:
## Issue

#1110 

## Change

add error code check in `QianfanClient's ResponseContentExtractor`.
 if `ChatCompletionResponse` has errorCode  will throw `ChatCompletionResponse` directly.

## General checklist


langchain4j-qianfan/src/test/java/dev/langchain4j/model/qianfan/QianfanChatModelIT.java
```diff
@@ -3,8 +3,10 @@
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.*;
+import dev.langchain4j.model.chat.ChatLanguageModel;
 import dev.langchain4j.model.output.Response;
 import dev.langchain4j.model.output.TokenUsage;
+import dev.langchain4j.model.qianfan.client.QianfanHttpException;
```
Kugaaa, line 10: 我认为这里不应该使用 `QianfanHttpException`，这里不属于 HTTP 异常
例如内部错误、限流等，是不是可以定义一个新的异常类型以作区分？

I think `QianfanHttpException` should not be used here, it is not an HTTP exception.
For example, internal errors, current limiting, etc., can a new exception type be defined to distinguish it?

yao-wenbin: 你说的很有道理，我重新为千帆增加了一个QianfanApiException，用于区分调用Qianfan API时发生的异常。

you are right, I add a QianfanApiException to qianfan module, which is used to  distinguish exception occurred when calling qianfan API

ghost (approved):
@yao-wenbin thank you!

Issue #940, opened 2024-04-15
[BUG] Anthropic: first message must use the "user" role


langchain4j:
When using Anthropic models with `ChatMemory`, it can happen that messages do not fit into the memory any more and the first `UserMessage` gets evicted. In this case, Anthropic will raise an error.

Potential solutions:
- if `Anthropic*Model.generate()` methods receive a list and the first message is not a `UserMessage`, drop messages from the beginning of a list untill first message is `UserMessage`
- if `Anthropic*Model.generate()` methods receive a list and the first message is not a `UserMessage`, prepend a dummy `UserMessage`

johnsosoka:
I'll address this, feel free to assign the ticket to me.

johnsosoka:
https://github.com/langchain4j/langchain4j/pull/1197

langchain4j:
Fixed by https://github.com/langchain4j/langchain4j/pull/1197

PR #1197, opened 2024-05-29
fix issue 940 - Anthropic API requires 1st message to be a UserMessage

johnsosoka:
## Issue

[Issue 940](https://github.com/langchain4j/langchain4j/issues/940)

## Change
Ensures that the 1st message sent to Anthropic is a UserMessage; if needed, it will create a dummy user message with content "."

## General checklist

- [N/A] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [N/A] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)

- [N/A] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

- [N/A] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [N/A] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

- [N/A] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j


langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
```diff
@@ -188,4 +193,23 @@ public static AnthropicTool toAnthropicTool(ToolSpecification toolSpecification)
                         .build())
                 .build();
     }
+
+    private static List<ChatMessage> ensureFirstMessageIsUser(List<ChatMessage> messages) {
+        if (isNullOrEmpty(messages)) {
+            return Collections.singletonList(new UserMessage("."));
+        }
+
+        List<ChatMessage> filteredMessages = messages.stream()
```
johnsosoka: I'm going to rewrite this a bit. I shouldn't discard the `SystemMessage` here.

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
```diff
@@ -188,4 +193,23 @@ public static AnthropicTool toAnthropicTool(ToolSpecification toolSpecification)
                         .build())
                 .build();
     }
+
+    private static List<ChatMessage> ensureFirstMessageIsUser(List<ChatMessage> messages) {
+        if (isNullOrEmpty(messages)) {
```
langchain4j: messages must not be null or empty, it is an error

ghost (commented):
@johnsosoka why this solution and not the other one? (just curious)

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
johnsosoka: I'll remove the check, I didn't notice `ensureNotEmpty` is already called and throws an error upstream of this.

johnsosoka:
@langchain4j I chose this solution because I didn't want the `ChatModel` discarding context, so I favored the dummy message instead. I figured that the `ChatMemory` implementation should be making all the decisions on when to discard messages. 

I can move the `ensureFirstMessageIsUser` call to the `generate` method as originally suggested...But, the method itself doesn't seem generic enough to move to `ValidationUtils.java` do you have any strong opinions on that? By the way, I'm fine implementing your alternative suggestion if you don't like the dummy message approach.

johnsosoka:
I re-arranged a little bit & moved work out of the mapper. 

* created a sanitizer package & class (since it's modifying original messages)
* moved `ensureNotEmpty` to the sanitize call
* SystemMessage is removed in sanitizer...It's discarded downstream in the Mapper & only used to prepare the `system`  portion of the anthropic request

By discarding the SystemMessage prior to ensuring that a UserMessage is at index 0, we avoid weird conditions where the SystemMessage is later discarded downstream (by the mapper) and results in an Anthropic error due to two UserMessages in sequence.

The Sanitizer could be expanded in the future to ensure other things like no back-to-back user messages in the list of ChatMessages

---

Looking for opinions on these changes, if it seems okay I'll touch up the tests a bit more. I'm unsure about the refactor/naming/dummy message setup.  CC @langchain4j

langchain4j:
@johnsosoka thanks a lot!

 I am not sure how well Anthropic models will handle dummy user message (now and in the future). They explicitly verify that the first message is user message, I don't know the reason for it, but I assume there is a good reason, if they have this validation. Maybe they fine-tune their models with the assumption that first message is always from user. So appending a dummy user message seems like a hack and may affect the quality of replies. I would rather go with another option: remove messages from the beginning untill the first message is from user.

Since this is an Anthropic-specific thing, I would keep this logic in the Anthropic module and not move it to chat memory impls.

WDYT?

johnsosoka:
okay, I'll update after work:

* remove the prepended dummy message
* continue stripping System Message in Sanitizer
* Drop messages until 1st message is UserMessage (I'd like to log a warning that a message is being dropped, any issue?)
* Throw an error if no UserMessage is present
* Add check to ensure no back-to-back UserMessages; drop one if there are two (another Anthropic requirement I've encountered in this)

johnsosoka:
Updates pushed. Dropping messages now until 1st message is user. I'm also performing additional sanitization by dropping messages if there are multiple consecutive UserMessages (which throws a different API error) tests expanded as well.

Ready for re-review @langchain4j

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/internal/sanitizer/MessageSanitizerTest.java
```diff
@@ -0,0 +1,141 @@
...
+    void test_stripMultipleSystemMessages() {
+        List<ChatMessage> messages = new ArrayList<>();
+        messages.add(new SystemMessage("System message 1"));
+        messages.add(new SystemMessage("System message 2"));
+        messages.add(new UserMessage("User message"));
+
+        List<ChatMessage> sanitized = MessageSanitizer.sanitizeMessages(messages);
+
+        assertEquals(1, sanitized.size());
+        assertTrue(sanitized.get(0) instanceof UserMessage);
```
langchain4j: could you please also check which exactly user message is left? in other tests as well

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/internal/sanitizer/MessageSanitizerTest.java
```diff
@@ -0,0 +1,141 @@
...
+        messages.add(new UserMessage("User message 2"));
+        messages.add(new AiMessage("AI message"));
+
+        List<ChatMessage> sanitized = MessageSanitizer.sanitizeMessages(messages);
+
+        assertEquals(2, sanitized.size());
+        assertTrue(sanitized.get(0) instanceof UserMessage);
+        assertTrue(sanitized.get(1) instanceof AiMessage);
+    }
+}
```
langchain4j, line 214: could you please add a few tests with `ToolExecutionResultMessage`s?

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/sanitizer/MessageSanitizer.java
```diff
@@ -0,0 +1,61 @@
...
+import static dev.langchain4j.internal.ValidationUtils.ensureNotEmpty;
+
+/**
+ * Sanitizes the messages to conform to the format expected by the Anthropic API.
+ */
+@Slf4j
+public class MessageSanitizer {
+
+
+    public static List<ChatMessage> sanitizeMessages(List<ChatMessage> messages) {
```
langchain4j, line 20: Could you please add a javadoc to `AnthropicChatModel` explaining this logic?

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicChatModel.java
langchain4j, line 1: I guess AnthropicStreamingChatModel should also have the same changes?

ghost (commented):
@johnsosoka thanks a lot!

johnsosoka:
Whoops! Totally forgot about the streaming model. I've added `AnthropicStreamingChatModel` support and expanded the tests quite a bit. I've re-run the entire anthropic model test suite & it's all green.

ghost (approved):
@johnsosoka thank you a lot!

PR #1200, opened 2024-05-30
Update to the latest Azure OpenAI models

agoncal:
@jdubois can you double check I have all the models updated #Thanks


jdubois (approved):

langchain4j:
@agoncal thanks!
BTW maybe we should add separate enums for every model type, like `OpenAiChatModelName` and `OpenAiEmbeddingModelName`

agoncal:
Ok I have a look into it

agoncal:
See https://github.com/langchain4j/langchain4j/pull/1205

PR #1205, opened 2024-05-30
Refactor Azure OpenAI models into enums

agoncal:
As discussed in https://github.com/langchain4j/langchain4j/pull/1200 I have refactored the Azure OpenAI models to turn them into enums


langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiModelName.java
```diff
@@ -1,47 +0,0 @@
-package dev.langchain4j.model.azure;
-
-/**
- * You can get the latest model names from the Azure OpenAI documentation or by executing the Azure CLI command:
- * az cognitiveservices account list-models --resource-group "$RESOURCE_GROUP" --name "$AI_SERVICE" -o table
- */
-public class AzureOpenAiModelName {
```
langchain4j, line 7: What about marking the class and/or all the constants as `@Deprecated` instead of deleting them right away? (Someone might be using it)

ghost (dismissed):
@agoncal thank you! One nice extra thing would be to have methods in model builders that consume these enums (in addition to `String`), but that is minor

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiModelName.java
agoncal: Good point: https://github.com/langchain4j/langchain4j/pull/1205/commits/d09273fe97a06b365f1ef0bac85b2b947d0b462f

langchain4j: Nice, thanks! Would be nice to add the `@Deprecated` annotation on the class (now it is only in javadoc), so that users of this class will see the warning in the IDE about the deprecation

agoncal:
Created an issue so we don't forget about the idea of consuming the enums: https://github.com/langchain4j/langchain4j/issues/1206

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiModelName.java
langchain4j: @agoncal please check my comment above

agoncal: Done. My bad, I thought you were talking about the Javadoc.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModelName.java
```diff
@@ -0,0 +1,42 @@
+package dev.langchain4j.model.azure;
+
+/**
+ * You can get the latest model names from the Azure OpenAI documentation or by executing the Azure CLI command:
+ * az cognitiveservices account list-models --resource-group "$RESOURCE_GROUP" --name "$AI_SERVICE" -o table
+ */
+public enum AzureOpenAiChatModelName {
+
+    GPT_3_5_TURBO("gpt-3.5-turbo"), // alias for the latest gpt-3.5-turbo model
```
langchain4j: Shouldn't it (and the rest of 3.5 models here and in `AzureOpenAiLanguageModelName`) be `gpt-35-turbo`? AFAIK Azure deployment names do not have dots.

One way to test it would be to iterate over all enums with `maxTokens=1`, like `AnthropicChatModelIT.should_support_all_enum_model_names()`

agoncal: Yes, you're right. Dot `.` are not allowed. 

I've added a few tests so I can test all the model names. But due to quota restriction, it was impossible to test them all at once. I had to test bunch by bunch... but all the tests pass.

I've also updated the shell script so it creates all the models.

Let me know what you think

ghost (approved):
@agoncal thanks a lot!

PR #1215, opened 2024-06-02
feat: integrate SearchApi as a WebSearchEngine and as a Tool

### LINKED ISSUE #1132: [FEATURE] [💎 $400 bounty] Integrate SearchApi as a WebSearchEngine and as a Tool
## Feature Overview

Integrate [SearchApi](https://www.searchapi.io/) as a `WebSearchEngine` and as a `Tool` for function calling.

## Requirements

- Support [Google Search via SearchApi](https://www.searchapi.io/docs/google)
- Add SearchApi tool documentation
- Write at least 1 example of using SearchApi as a tool

Adhere to [langchain4j contribution guidelines](https://github.com/langchain4j/langchain4j/blob/main/CONTRIBUTING.md)

## Related work

- https://github.com/langchain4j/langchain4j/pull/642 - ground work for adding search tools
- https://github.com/langchain4j/langchain4j/pull/641
- https://github.com/langchain4j/langchain4j/pull/676

Related issues:
- https://github.com/langchain4j/langchain4j/issues/146

## Existing SearchApi integrations

- [SearchApi LangChain Python](https://python.langchain.com/v0.2/docs/integrations/tools/searchapi/)
- [SearchApi LangChain JS](https://js.langchain.com/v0.2/docs/integrations/tools/searchapi)
- [SearchApi Summarization example](https://js.langchain.com/v0.2/docs/tutorials/summarization)

## Design considerations

SearchApi support not only Google Search, but 30+ other APIs such as Youtube Search, Transcripts, Bing Search (similar JSON response keys). In a statically typed language, it is probably better to have separate tools for separate APIs, but implementation could be flexible enough to extend SearchApi and easily adapt to other engines. All engines use the same HTTP GET request. The only difference is the parameters they accept. And how you want the response to be parsed. 

## Other notes

- Ping us if you need more search credits when implementing.
- Someone from the SearchApi Team will help review the functionality.

## Bounty

There is a $400 bounty on this awarded by SearchApi to the community.

In the event of multiple attempts, the one that is going to be merged by the maintainers of the repository will pick the bounty.

ayewo:
## Issue
This implements the integration of SearchApi as a `WebSearchEngine` which after merging should close https://github.com/langchain4j/langchain4j/issues/1132

## Change

cc: @mark-searchapi 

## General checklist

@algora-pbc /claim #1132


langchain4j-parent/pom.xml
```diff
@@ -170,6 +170,11 @@
                 <artifactId>okhttp-sse</artifactId>
                 <version>${okhttp.version}</version>
             </dependency>
+            <dependency>
+                <groupId>com.squareup.okhttp3</groupId>
+                <artifactId>logging-interceptor</artifactId>
```
langchain4j, line 181: we plan to get rid of okhttp in the near future, so please dont use any okhttp-specific features

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
```diff
@@ -0,0 +1,122 @@
...
+class SearchApiWebSearchEngineIT extends WebSearchEngineIT {
+
+	public static final String SEARCHAPI_API_KEY = "SEARCHAPI_API_KEY";
+	
+	private static final boolean logRequests = false;
+	
+    protected WebSearchEngine webSearchEngine = SearchApiWebSearchEngine.withApiKey(System.getenv(SEARCHAPI_API_KEY)).logRequests(logRequests).build();
+
+    @Test
+    void should_search_with_defaults() {
```
langchain4j: is there a need for this test? `WebSearchEngineIT.should_search` seems to test the same/similar

ghost (commented):
@ayewo thanks!

langchain4j-parent/pom.xml
ayewo: Thanks for your review, @langchain4j!

I added this dependency to enable logging of requests and responses from SearchApi with OkHttpClient, similar to how logging is enabled in `langchain4j-web-search-engine-google-custom` via these classes:
* [`GoogleSearchApiHttpRequestLoggingInterceptor.java`](https://github.com/langchain4j/langchain4j/blob/e9a8c25df171f8fd0a16f799baa6d9439a4c3d3b/web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleSearchApiHttpRequestLoggingInterceptor.java)
* [`GoogleSearchApiHttpResponseLoggingInterceptor.java`](https://github.com/langchain4j/langchain4j/blob/e9a8c25df171f8fd0a16f799baa6d9439a4c3d3b/web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleSearchApiHttpResponseLoggingInterceptor.java)

Should I restrict the `logging-interceptor` dependency by moving it inside `langchain4j-web-search-engine-searchapi/pom.xml` instead?

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
ayewo: You are right: `should_search_with_defaults()` duplicates the tests in `WebSearchEngineIT#should_search()`.

I'll remove it.

ayewo: Removed in [93c5feb](https://github.com/langchain4j/langchain4j/pull/1215/commits/93c5feb4ffcc41b8a5ad55c90d3366625659d4ac).

langchain4j-parent/pom.xml
langchain4j: My bad, it is ok for logging, please nevermind :)

ayewo:
Hi @Heezer 👋, 
I see you are leaving comments and providing suggestions in this [PR](https://github.com/langchain4j/langchain4j/pull/1216) 👍.

As you already acknowledged [here](https://github.com/langchain4j/langchain4j/pull/1216#discussion_r1660309973), that PR is the 2nd of 2 PRs opened for the same [issue](https://github.com/langchain4j/langchain4j/issues/1132). 

As the author of the 1st PR, I’m a bit surprised you are spending so much time reviewing the 2nd PR. It wasn’t even ready for review until a day after I opened my PR. 

As part of triaging, @langchain4j added the P2 and P3 labels to my PR on June 3 but those labels weren't added to the 2nd PR until June 6.

## Humble Suggestion
Since the project is in dire need of more eyes to review [outstanding PRs](https://github.com/langchain4j/langchain4j/pulls?q=is%3Apr+is%3Aopen+review%3Arequired), perhaps it would be better if you did additional triaging first?

What I’m suggesting is that you go over the 2 PRs side-by-side and decide for yourself which one is more likely to be merged by the langchain4j maintainers based on which one:
- implements most of the [requirements as stated](https://github.com/langchain4j/langchain4j/issues/1132#issue-2306233099);
- adheres closest to the [two existing implementations](https://github.com/langchain4j/langchain4j/issues/1132#issuecomment-2129734548);
- is easier to extend to add support for other [SearchApi engines mentioned by the sponsor](https://github.com/langchain4j/langchain4j/issues/1132#issuecomment-2144056224).

For instance, the [assert failure](https://github.com/langchain4j/langchain4j/pull/1216#discussion_r1660309973) you helpfully pointed out in the 2nd PR had already [implemented](https://github.com/langchain4j/langchain4j/issues/1132#issuecomment-2144295271) in my PR.

I think this is a more efficient use of your time because at the end of the day, only one PR will be accepted. 

It is the accepted PR that the maintainers will ask its author to make final changes prior to being merged. 

So there’s a real risk that the time you’ve already spent reviewing the 2nd PR (before proper triaging) might go unacknowledged by the maintainers, if the 2nd PR gets closed.

Thanks!

Heezer:
Hi @ayewo 👋 , I didn't know at the beginning about your "competition" 😄 in the meantime I'm done with review of #1216, let me review #1215 considering your suggestions, thanks! But I guess the decision what to take won't be mine 😄

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchContentRetrieverIT.java
```diff
@@ -0,0 +1,61 @@
...
+import dev.langchain4j.web.search.WebSearchEngine;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static dev.langchain4j.web.search.searchapi.SearchApiWebSearchEngine.DEFAULT_ENV_VAR;
+
+
+@EnabledIfEnvironmentVariable(named = "SEARCHAPI_API_KEY", matches = ".+")
+class SearchApiWebSearchContentRetrieverIT extends WebSearchContentRetrieverIT {
```
Heezer, line 17: This test fails for me with `java.util.concurrent.CompletionException: java.lang.IllegalArgumentException: The metadata value for the key 'thumbnail' cannot be null`, but the response from Google is coming. Looks like API has been changed in the meanwhile? 🤔

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
```diff
@@ -0,0 +1,93 @@
...
+        SearchApiWebSearchEngine searchapiWebSearchEngine = SearchApiWebSearchEngine.builder()
+                .apiKey(System.getenv(DEFAULT_ENV_VAR))
+                .logRequests(logRequests)
+                .build();
+
+        // when
+        WebSearchResults webSearchResults = searchapiWebSearchEngine.search("chatgpt");
+
+        // then
+        assertThat(webSearchResults.searchInformation().totalResults() > 0);
```
Heezer: assertThat is not followed by an assertion invocation, Sonar java:S2970. Expected is such assert notation:
```suggestion
        assertThat(webSearchResults.searchInformation().totalResults()).isPositive();
```
the same for the following asserts

Heezer: FYI you could see these code smell warnings from SonarLint plugin

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
```diff
@@ -0,0 +1,93 @@
...
+
+        results.forEach(result -> {
+            assertThat(result.title()).isNotBlank();
+            assertThat(result.url()).isNotNull();
+            assertThat(result.snippet()).isNotBlank();
+            assertThat(result.content()).isNull();
+            assertThat(result.metadata()).isNotNull();
+        });
+
+        assertThat(results).anyMatch(result ->
```
Heezer: this assert fails for me 😞

ayewo:
> Hi @ayewo 👋 , I didn't know at the beginning about your "competition" 😄 in the meantime I'm done with review of #1216, let me review #1215 considering your suggestions, thanks! But I guess the decision what to take won't be mine 😄

Haha 😁, no worries.

Thanks for volunteering your time!

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchToolIT.java
```diff
@@ -0,0 +1,169 @@
...
+            .apiKey(System.getenv("OPENAI_API_KEY"))
+            .modelName(OpenAiChatModelName.GPT_3_5_TURBO)
+            .logRequests(logRequests)
+            .build();
+
+    interface Assistant {
+        @dev.langchain4j.service.SystemMessage({
+                "You are a web search support agent.",
+                "If there is any event that has not happened yet",
+                "You MUST create a web search request with with user query and",
```
Heezer: your competitor had the same issue 😄 
```suggestion
                "You MUST create a web search request with user query and",
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchToolIT.java
```diff
@@ -0,0 +1,169 @@
...
+        String answer(String userMessage);
+    }
+
+    @Test
+    void should_execute_google_tool_with_chatLanguageModel_to_give_a_final_response(){
+        // given
+        webSearchEngine = SearchApiWebSearchEngine
+        		.withApiKey(System.getenv(DEFAULT_ENV_VAR))
+                .logRequests(logRequests)
+                .build();
```
Heezer, line 54: strange formatting

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchToolIT.java
```diff
@@ -0,0 +1,169 @@
...
+
+import static dev.langchain4j.web.search.searchapi.SearchApiWebSearchEngine.DEFAULT_ENV_VAR;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@EnabledIfEnvironmentVariable(named = "SEARCHAPI_API_KEY", matches = ".+")
+class SearchApiWebSearchToolIT extends WebSearchToolIT {
```
Heezer, line 25: first two tests fail 😞

Heezer (changes requested):

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchContentRetrieverIT.java
ayewo: Yes, the search results have changed in the time between when I wrote the tests and now that you are reviewing it.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
ayewo: This should be fixed now. Please kindly re-test using commit: [2bd238b](https://github.com/langchain4j/langchain4j/pull/1215/commits/2bd238b1ffc752b263202c97c8d1ceb3b12c6050).

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
ayewo: Thanks for flagging! The assert code smells should be fixed in `SearchApiWebSearchEngineIT.java` now.

And thanks for the tip to install the SonarLint plugin 👍.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchToolIT.java
ayewo: Fixed.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchToolIT.java
ayewo: Fixed.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
Heezer: yes, works now 👍

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchToolIT.java
Heezer: tests work except of the last one, there is a small fix needed

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchToolIT.java
```diff
@@ -0,0 +1,169 @@
...
+        messages.add(toolExecutionResultMessage);
+
+        AiMessage finalResponse = chatLanguageModel().generate(messages).content();
+        System.out.println(finalResponse.text());
+
+        // then
+        assertThat(finalResponse.text())
+                .as("At least the string result should be contains 'madrid' and 'tourist' ignoring case")
+                .containsIgnoringCase("Madrid")
+                .containsIgnoringCase("Royal Palace");
```
Heezer: here is this fix:
```suggestion
                .containsIgnoringCase("tourist");
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchContentRetrieverIT.java
Heezer: test still fails with the same exception

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
```diff
@@ -0,0 +1,105 @@
...
+import lombok.Builder;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
+@Builder
+class SearchApiRequest {
+
+	/************************************************************************************************
```
Heezer: strange formatting everywhere 😞  looks like spaces mixed with tabs

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
```diff
@@ -0,0 +1,105 @@
...
+    private Integer page;
+    
+    
+    
+    
+    /* Returns a map of key-value pairs representing optional parameters for the search request. */ 
+    public Map<String, Object> getOptionalParameters() {
+    	final Map<String, Object> params = new HashMap<String, Object>();
+    	
+    	if (isNotNullOrBlank(device)) params.put("device", getOrDefault(device, "desktop"));
```
Heezer: hmm, this is strange - `getOrDefault` returns default value if provided value is null, but this cannot be the case because the `isNotNullOrBlank` before. in other words if `device` is null, it won't come to default 'desktop', never

Heezer: and according to the docs, the default value on the server side is already 'desktop':
image
so, `getOrDefault` makes no sense on the client side

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiResponse.java
```diff
@@ -0,0 +1,97 @@
...
+	 * 
+	 * The optional JSON elements appear in the response depending on the type of query (e.g. local business, online shopping, population fact etc) 
+	 * or whether extra parameters are included in the search like "location".
+	 * 
+	 * Note that:
+	 * 	"related_questions" 
+	 *	"related_searches" 
+	 *	are also Mandatory JSON elements. They are always included in the API response but are currently ignored since they are never used directly.
+	 * 
+	 */
```
Heezer: this is such a javadoc overkill 😞

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+    
+    @SuppressWarnings("unchecked")
+	private static void appendAnswerToResults(final SearchApiResponse response, final List<WebSearchOrganicResult> results) {
+    	// searchapi.io may include 1 or more optional JSON elements in the response that directly answers the search query in the "knowledge_graph", "answer_box", etc
+        // such optional elements counts toward the "num" parameter. For instance, if "num" is set to 5 and an "inline_images" is included, then "organic_results" will contain 4 (instead of 5) items,  
+        // which is why we have to add it back to "results" below as the first item.
+    	// See also: https://api.python.langchain.com/en/latest/_modules/langchain_community/utilities/searchapi.html#SearchApiAPIWrapper
+    	
+    	String title = null, website = "https://www.searchapi.io/", description = null;
```
Heezer: what is this line for? this is not how clean Java code should look like, why again the address you already defined above as constant?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
```diff
@@ -0,0 +1,105 @@
...
+import lombok.Setter;
+
+@Getter
+@Setter
+@Builder
+class SearchApiRequest {
+
+	/************************************************************************************************ 
+	 * Mandatory parameters 
+	 * **********************************************************************************************/
```
Heezer: does this internal class really needed to be documented so heavily? this class won't be used by LangChain4J clients. If yes, then pls all these comments shall be reformatted, see https://github.com/langchain4j/langchain4j/blob/d28f5ab470360f68a688bbe0a2ed952d07ef9bf9/langchain4j-core/src/main/java/dev/langchain4j/web/search/WebSearchRequest.java as example. Same for the response.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,87 @@
...
+        final Retrofit retrofit = new Retrofit.Builder()
+                .baseUrl(baseUrl)
+                .client(okHttpClientBuilder.build())
+                .addConverterFactory(GsonConverterFactory.create(GSON))
+                .build();
+
+        this.searchapiAPI = retrofit.create(SearchApiAPI.class);
+    }
+
+    public SearchApiResponse search(final SearchApiRequest searchRequest) {
```
Heezer:
```suggestion
    SearchApiResponse search(final SearchApiRequest searchRequest) {
```
since `SearchApiRequest` has default visibility

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,171 @@
...
+ * Represents SearchApi Search API as a {@code WebSearchEngine}.
+ * See more details <a href="https://www.searchapi.io/docs/google">here</a>.
+ */
+public class SearchApiWebSearchEngine implements WebSearchEngine {
+	public static final String DEFAULT_ENV_VAR = "SEARCHAPI_API_KEY";
+    public static final String DEFAULT_BASE_URL = "https://www.searchapi.io";
+    public static final String DEFAULT_ENGINE = "google";
+    private static final Integer DEFAULT_MAX_RESULTS = 5;
+
+    private final String apiKey;
```
Heezer: `apiKey`, `engine` and `logRequests` don't have to be instance variables

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiResponse.java
ayewo: Nope, not overkill at all. I sometimes think aloud in comments. 

Think of it as a note to my future self[^1] or to other devs that might want to orient themselves before they dive in.



[1]: [Only-god-knows](https://github.com/Tjoosten/Funny-code-comments/blob/1ddc6babbb1c69726334c76814748adf1b56e07b/Only-god-knows.md)

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: Yes, they don't have to be but there's nothing wrong if they are left as-is.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
ayewo: Fixed.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
ayewo: Should be fixed now.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchToolIT.java
ayewo: Done.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
ayewo: No need to be patronizing in your feedback. 

You could have delivered the same message by merely writing: "the `getOrDefault()` call is not necessary".

ayewo: I've removed it.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: > what is this line for? 

I tried to explain why in the preceding comment:
> searchapi.io may include 1 or more optional JSON elements in the response
> that directly answers the search query in the "knowledge_graph",
> "answer_box", etc
> such optional elements counts toward the "num" parameter. For instance, if
> "num" is set to 5 and an "inline_images" is included, then "organic_results"
> will contain 4 (instead of 5) items,
> which is why we have to add it back to "results" below as the first item.
> See also:
> https://api.python.langchain.com/en/latest/_modules/langchain_community/utilities/searchapi.html#SearchApiAPIWrapper

Please also refer to @mark-searchapi's [comment](https://github.com/langchain4j/langchain4j/issues/1132#issuecomment-2144059116) affirming my approach.

The existing code for the Tavily search engine [does similar](https://github.com/langchain4j/langchain4j/blob/2cdfb4a5f121e4ec12eeac6ce1abd0f6a8b70e8a/web-search-engines/langchain4j-web-search-engine-tavily/src/main/java/dev/langchain4j/web/search/tavily/TavilyWebSearchEngine.java#L82-L89).

No problem at all if the maintainers weigh in and ask me to remove it.


> this is not how clean Java code should look like, why again the address you already defined above as constant?

I've fixed the nits.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
ayewo: I re-did the formatting of the comment.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchContentRetrieverIT.java
Heezer: still fails, same exception

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
Heezer: I've provided my thoughts about the check logic, where you possibly had some another intention while writing the code... 
So, there are still two same checks for "hl" and "gl".

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiResponse.java
Heezer: @langchain4j  pls have a look. In my opinion such big comments could belong to some documentation maybe, but not defined as javadoc to an empty line...

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
Heezer: The clean javadoc comment in `WebSearchRequest` starting the line 11. And see how nice and readable is the generated documentation as result: 
image

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchContentRetrieverIT.java
ayewo: Would you be kind enough to include the full stack trace?

Thanks!

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
ayewo: Agreed 👍.

I've redone the comments to follow the javadoc style as requested.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
ayewo: Thanks for catching. Fixed.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiResponse.java
langchain4j: indeed seems like an overkill, isn't it documented in the searchapi docu?

ayewo: @langchain4j no, the same info is not presented in this way in the SearchApi docs.

I guess I could move it into a Markdown file?

langchain4j: sounds like a good idea to document request and response examples in the README.md of this module

ayewo: @langchain4j thanks for the suggestion.

I've extracted the text into the [README.md](https://github.com/ayewo/langchain4j/blob/feat/search-api/web-search-engines/langchain4j-web-search-engine-searchapi/README.md) for this module.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchContentRetrieverIT.java
ayewo: Should be fixed now. Please kindly confirm.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,72 @@
...
+
+	/* Indicates whether to log requests and their responses. Defaults to false. */
+	private final boolean logRequests;
+
+	private static final Gson GSON = new GsonBuilder().setFieldNamingPolicy(LOWER_CASE_WITH_UNDERSCORES)
+			.setPrettyPrinting().create();
+
+	@Builder
+	public SearchApiClient(final String baseUrl, final String apiKey, final String engine, final Duration timeout,
+			final boolean logRequests) {
```
langchain4j, line 31: `logResponses` is missing, is it on purpose?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,184 @@
...
+	private final String engine;
+	private final SearchApiClient searchapiClient;
+	private final Boolean logRequests;
+
+	@Builder
+	public SearchApiWebSearchEngine(final String apiKey, final String engine, final Duration timeout,
+			final Boolean logRequests) {
+
+		this.apiKey = ensureNotBlank(apiKey, "apiKey");
+		this.engine = ensureNotBlank(getOrDefault(engine, DEFAULT_ENGINE), "engine");
```
langchain4j:
```suggestion
		this.engine = getOrDefault(engine, DEFAULT_ENGINE);
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,184 @@
...
+		this.logRequests = logRequests;
```
langchain4j:
```suggestion
		this.logRequests = getOrDefault(logRequests, false);
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,184 @@
...
+
+		final SearchApiRequest.SearchApiRequestBuilder requestBuilder = SearchApiRequest.builder();
+
+		requestBuilder.q(webSearchRequest.searchTerms()).safe(webSearchRequest.safeSearch())
+				.num(getOrDefault(webSearchRequest.maxResults(), DEFAULT_MAX_RESULTS)) // maxResults should default to 5
+				.page(webSearchRequest.startPage());
+
+		/**
+		 * {@link WebSearchRequest#language} is in a composite string "en-us" but
+		 * SearchApi splits them into 2 individual strings. See
```
langchain4j: please extract this logic into separate static method and cover it with unit tests

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,184 @@
...
+				getOrDefault(searchapiResponse.getPagination().getCurrent(), 1),
+				searchapiResponse.getSearchParameters());
+
+		// merge the "search_information" JSON element with the "search_metadata" JSON
+		// element
+		searchapiResponse.getSearchMetadata().putAll(searchapiResponse.getSearchInformation());
+		return WebSearchResults.from(searchapiResponse.getSearchMetadata(), result, results);
+	}
+
+	public static SearchApiWebSearchEngineBuilder withApiKey(final String apiKey) {
```
langchain4j:
```suggestion
	public static SearchApiWebSearchEngine withApiKey(final String apiKey) {
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,184 @@
...
+		return builder().apiKey(apiKey);
```
langchain4j:
```suggestion
		return builder().apiKey(apiKey).build();
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,184 @@
...
+import dev.langchain4j.web.search.WebSearchResults;
+import dev.langchain4j.web.search.searchapi.result.OrganicResult;
+import lombok.Builder;
+
+/**
+ * Represents SearchApi Search API as a {@code WebSearchEngine}. See more
+ * details <a href="https://www.searchapi.io/docs/google">here</a>.
+ */
+public class SearchApiWebSearchEngine implements WebSearchEngine {
+	public static final String DEFAULT_ENV_VAR = "SEARCHAPI_API_KEY";
```
langchain4j: `DEFAULT_ENV_VAR` should not be here

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchContentRetrieverIT.java
langchain4j: Seems to work

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
ayewo: Yes, it's on purpose.

The `langchain4j-web-search-engine-google-custom` module uses [two separate](https://github.com/langchain4j/langchain4j/blob/d23b67567f64c9a2a203704370679460ec92fcce/web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleCustomSearchApiClient.java#L35-L36) flags i.e. `logRequests` and `logResponses` because the HTTP client it uses [supports two individual](https://github.com/langchain4j/langchain4j/blob/d23b67567f64c9a2a203704370679460ec92fcce/web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleCustomSearchApiClient.java#L57-L61) flags to enable logging of requests and responses.

My implementation is based on the `langchain4j-web-search-engine-tavily` module but the developer did not enable [support for logging](https://github.com/langchain4j/langchain4j/blob/d23b67567f64c9a2a203704370679460ec92fcce/web-search-engines/langchain4j-web-search-engine-tavily/src/main/java/dev/langchain4j/web/search/tavily/TavilyClient.java#L26). When I added [support for logging](https://github.com/ayewo/langchain4j/blob/2701180c06468bfd930e67d0863e2a573bc9e850/web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java#L39-L43) in my implementation, I realized that the HTTP client (`OkHttpClient`) only accepts one flag to enable logging which is why I removed `logResponses` since it was redundant.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: @langchain4j Should I move `DEFAULT_ENV_VAR` to a different class or remove the const completely?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: Fixed.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: I decided to extract the logic into a separate static class (instead of a static method).
* [`Locale`](https://github.com/ayewo/langchain4j/blob/d44c28591ad10d65ddef6330b93066eeece3c6f2/web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java#L170-L193)

As requested, I added unit tests: 
* [`LocaleTest`](https://github.com/ayewo/langchain4j/blob/952e0169558c6c2903450f167eec6fa61c478572/web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/LocaleTest.java)

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,194 @@
...
+        final SearchApiResponse searchapiResponse = searchapiClient.search(request);
+
+        final List<WebSearchOrganicResult> results = searchapiResponse.getOrganicResults().stream()
+				.map(result -> WebSearchOrganicResult.from(result.getTitle(), URI.create(result.getLink()),
+						result.getSnippet() != null ? result.getSnippet() : result.getDisplayedLink(), // for the first
+																										// few runs of
+																										// the query I
+																										// tried,
+																										// "snippet" is
+																										// somehow null
```
mark-searchapi: If `snippet` is null, we should default to an empty string `""` instead.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,194 @@
...
+						null, // searchapi does not return "content"
+						toResultMetadataMap(result)))
```
mark-searchapi, line 100: Not sure what this metadata is used for. `position` might make sense, but anything else, probably not.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,194 @@
...
+				.collect(toList());
+
+		// if the search query has a direct answer, append it to the top of the organic
+		// results
+		appendAnswerToResults(searchapiResponse, results);
```
mark-searchapi: This tightly couples the integration with the `google` search engine and is prone for errors. A custom tool can be implemented separately that handles the JSON response in unique ways.

WebSearchEngine is used mostly for RAG use cases, so extracting the `organic_results` array and collecting only `title`, `link`, and `snippet` should be enough.

mark-searchapi (commented):
Implementation seemed to be heavily based on Tavily, but Tavily is a wrapper around an API such as SearchAPI or Google Custom Search API or any other search engine API.

In general, similar comments to https://github.com/langchain4j/langchain4j/pull/1216 apply. Key aspects:
- Missing public documentation (including usage with OpenAI)
- Results tightly coupled to `google`. We can support `organic_results` array with only `title`, `link` and `snippet`. This way users would be able to use SearchApi with any engine that returns organic results (for instance, `google`, `bing`, `baidu`, `bing_news`, `google_news`, `google_scholar`).
- Parameters tightly coupled to `google`. Apart from query (`q` parameter), `engine`, and `api_key`, user should be free to pass any key-value pairs as custom parameters. This makes the search engine integration future proof as using upcoming search engines (or adapting to query parameter change) won't require updating `langchain4j`.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: ~The reason for this approach was due to how SearchApi handles the `num` parameter as discussed [above](https://github.com/langchain4j/langchain4j/pull/1215#discussion_r1667406692). Even if a custom tool is implemented, the results from WebSearchEngine is still self-contained.~

~Please I'll need some clarity as to how to proceed.~

Nevermind. Once I read through your feedback in the PR of #1216, I think I now understand what you were trying to get across here.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: Okay. I've removed: `source` and `thumbail` leaving only `position` in the metadata.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: Thanks for your review. This is fixed.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: @mark-searchapi 
I've refactored the `SearchApiWebSearchEngine` class to reduce the coupling with the `google` search engine.

I've added 2 life cycle methods that callers need to override if they intend to provide additional parameters (for a search `request`), or if they desire to handle the JSON (of a `response`) in unique ways.

To customize a search request or search response, a caller only needs to override one method in their subclass of `SearchApiWebSearchEngine`: 
* `request` => `customizeSearchRequest(final WebSearchRequest webSearchRequest, final SearchApiRequest request)`
* `response` => `customizeSearchResults(final SearchApiResponse response, final List<WebSearchOrganicResult> results)`

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
langchain4j: it is used only in integration test, so please move it there

ayewo:
@mark-searchapi 
Thanks for your review. I'm furiously working to push an update but ...
I finally got the dreaded error 🙃:
```json
{
  "error": "You have used all of the searches for the month. Please upgrade your plan on SearchApi.io."
}
```
How do I request additional credits to my account?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
ayewo: Okay. Moved to [`SearchApiWebSearchEngineIT`](https://github.com/ayewo/langchain4j/blob/49bbeb51c338a49ae1d1b35af6189aa445177345/web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java#L21)

ayewo:
@mark-searchapi 
I've removed the high coupling with the Google Search engine and added support for 2 different ways to customize a search request. 

I'll update the documentation shortly but you can view the 2 styles that code can be used in this test file: [`SearchApiWebSearchEngineIT` ](https://github.com/ayewo/langchain4j/blob/feat/search-api/web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java) (the IT suffix indicates "Integration Tests" I believe).

In addition to the 2 styles supported, you can now use `SearchApiWebSearchEngine` to perform queries against multiple search engines.
The tests includes examples for the following search engines:
* google
* bing
* baidu

ayewo:
> Implementation seemed to be heavily based on Tavily, but Tavily is a wrapper around an API such as SearchAPI or Google Custom Search API or any other search engine API.
> 
> In general, similar comments to #1216 apply. Key aspects:
> 
>     * Missing public documentation (including usage with OpenAI)
> 
>     * Results tightly coupled to `google`. We can support `organic_results` array with only `title`, `link` and `snippet`. This way users would be able to use SearchApi with any engine that returns organic results (for instance, `google`, `bing`, `baidu`, `bing_news`, `google_news`, `google_scholar`).
> 
>     * Parameters tightly coupled to `google`. Apart from query (`q` parameter), `engine`, and `api_key`, user should be free to pass any key-value pairs as custom parameters. This makes the search engine integration future proof as using upcoming search engines (or adapting to query parameter change) won't require updating `langchain4j`.

@mark-searchapi 

I've addressed all of your feedback.

1. Wrt documentation showing OpenAI usage examples, please see [searchapi-io.md](https://github.com/ayewo/langchain4j/blob/feat/search-api/docs/docs/integrations/web-search-engines/searchapi-io.md).

2. & 3. The issue of tight coupling which you observed in the search results and  search request parameters should also be fixed.

web-search-engines/langchain4j-web-search-engine-searchapi/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.squareup.retrofit2</groupId>
+            <artifactId>retrofit</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.squareup.retrofit2</groupId>
+            <artifactId>converter-gson</artifactId>
```
czelabueno, line 33: We've started a few months ago to move JSON lib in LLM integration to Jackson to be compatible with java frameworks (spring, quarkus, etc). Any web search integration would be used with a LLM integration together in a normal way. It's not critical but as new integration would be great if you use Jackson mapper.

```xml
<dependency>
    <groupId>com.squareup.retrofit2</groupId>
    <artifactId>converter-jackson</artifactId>
</dependency>
<dependency>
  </dependency>
```
Check [Anthropic](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java) or [Mistral AI](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/client/DefaultMistralAiClient.java) integration.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class SearchApiClient {
+	private final SearchApiAPI searchapiAPI;
+	private final String engine;
+	private final String apiKey;
+
+	/* Indicates whether to log requests and their responses. Defaults to false. */
+	private final boolean logRequests;
+
+	private static final Gson GSON = new GsonBuilder().setFieldNamingPolicy(LOWER_CASE_WITH_UNDERSCORES)
```
czelabueno, line 26: Same here, Jackson mapper instead of

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
czelabueno: in fact, I think you can implement `okhttp3.Interceptor` so that the request and response treat them as separate flags. It would be great if we gave users the choice. I also think the request and response have different properties. See this example in [Mistral AI](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/client/DefaultMistralAiClient.java#L57-L63)

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,72 @@
...
+			.setPrettyPrinting().create();
+
+	@Builder
+	public SearchApiClient(final String baseUrl, final String apiKey, final String engine, final Duration timeout,
+			final boolean logRequests) {
+		this.apiKey = apiKey;
+		this.engine = engine;
```
czelabueno, line 33: If `apiKey`, `baseUrl` or `engine` is null or empty, does it work anyway? If not, you could add:

```suggestion
if (isNullOrBlank(baseUrl)) {
    throw new IllegalArgumentException("baseUrl cannot be null or empty");
}
if (isNullOrBlank(apiKey)) {
    throw new IllegalArgumentException("SearchAPI API Key must be defined. " +
            "It can be generated here: https://www.searchapi.io/api_tokens");
}
if (isNullOrBlank(engine)) {
    throw new IllegalArgumentException("engine parameter must be set");
}
this.apiKey = apiKey;
this.engine = engine;
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,72 @@
...
+		this.logRequests = logRequests;
+
+		final OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder().callTimeout(timeout)
+				.connectTimeout(timeout).readTimeout(timeout).writeTimeout(timeout);
+
+		if (this.logRequests) {
+			final HttpLoggingInterceptor interceptor = new HttpLoggingInterceptor();
+			interceptor.setLevel(HttpLoggingInterceptor.Level.BODY);
+			okHttpClientBuilder.addInterceptor(interceptor);
+		}
```
czelabueno, line 43: a comment related to this was done before. `logRequests` and `logResponses` could be added.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,277 @@
...
+    /**
+     * Functional interface to allow customization of request parameters prior to a search request.
+     */
+	protected final Consumer<Map<String, Object>> customizeParametersFunc;
+
+	@Builder
+	public SearchApiWebSearchEngine(final String apiKey, final String engine, final Duration timeout,
+			final Boolean logRequests, final Consumer<Map<String, Object>> customizeParametersFunc) {
+
+		this.apiKey = ensureNotBlank(apiKey, "apiKey");
```
czelabueno, line 49: `apiKey` should be validated as not null from api client classes, not from `WebSearchEngine`. This in order to keep clean code and add user guiding to create one if it's missing.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiRequest.java
```diff
@@ -0,0 +1,79 @@
...
+
+	/**
+     * The API key for authentication with the Search API.
+     */
+    private String apiKey;
+
+    /**
+     * The search engine to retrieve real-time data from. Defaults to "google".
+     */
+    private String engine;
```
czelabueno, line 50: `SearchApiClient` does not use these values ​​because they are requested in its constructor. To avoid redundancies, perhaps these fields can be removed.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,277 @@
...
+			website = response.getKnowledgeGraph().containsKey("website")
+					? (String) response.getKnowledgeGraph().get("website")
+					: website;
+			description = (String) response.getKnowledgeGraph().get("description");
+
+		} else if (response.getInlineVideos() != null && !response.getInlineVideos().isEmpty()) {
+			final Map<String, Object> video = response.getInlineVideos().get(0);
+			title = (String) video.get("title");
+			website = video.containsKey("link") ? (String) video.get("link") : website;
+			description = (String) video.get("image");
```
czelabueno, line 188: I'm not sure if `inlineVideos` should be included in this first version. We must include web page results and images results only.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,277 @@
...
+						toResultMetadataMap(result)))
+				.collect(toList());
+
+		// 4. if the search query has a direct answer, append it to the top of the organic results
+		customizeSearchResults(searchapiResponse, results);
+
+		final Double pageNumber = searchapiResponse.getPagination().containsKey("current") ? 
+				(Double)searchapiResponse.getPagination().get("current") : 1;
+		final WebSearchInformationResult result = WebSearchInformationResult.from(
+				Long.valueOf(results.size()),
```
czelabueno, line 109: the value must be `search_information.total_results`. These should be the total results that the search produces regardless of any added filtering.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
```diff
@@ -0,0 +1,336 @@
...
+		results.forEach(result -> {
+			assertThat(result.title()).isNotBlank();
+			assertThat(result.url()).isNotNull();
+			assertThat(result.snippet()).isNotNull();
+			assertThat(result.content()).isNull();
+			assertThat(result.metadata()).isNotNull();
+		});
+	}
+	
+//	@Test
```
czelabueno, line 286: could you remove this commented test? If it's not implemented yet it's not needed.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngineIT.java
```diff
@@ -0,0 +1,336 @@
...
+
+import dev.langchain4j.web.search.WebSearchEngine;
+import dev.langchain4j.web.search.WebSearchEngineIT;
+import dev.langchain4j.web.search.WebSearchInformationResult;
+import dev.langchain4j.web.search.WebSearchOrganicResult;
+import dev.langchain4j.web.search.WebSearchRequest;
+import dev.langchain4j.web.search.WebSearchResults;
+
+@EnabledIfEnvironmentVariable(named = "SEARCHAPI_API_KEY", matches = ".+")
+class SearchApiWebSearchEngineIT extends WebSearchEngineIT {
```
czelabueno, line 21: could you add a test case when user put as `engine='google_images'`? If I saw well the image results are parsed as organic results and would be great test them. :)

czelabueno (changes requested):
Hello @ayewo! You did a good job here! I have added some things, small changes that if you decide to make them, it would be more neat and adjusted to what is desired for a websearch engine to support LLMs :)

langchain4j:
Closing in favor of https://github.com/langchain4j/langchain4j/pull/1216
@ayewo thank you for your efforts!

PR #1216, opened 2024-06-02
[FEATURE] Adds SearchApi as WebSearchEngine and Tool

### LINKED ISSUE #1132: [FEATURE] [💎 $400 bounty] Integrate SearchApi as a WebSearchEngine and as a Tool

zambrinf:
## Issue

closes #1132 
@mark-searchapi

## Change
Added SearchApi as a WebSearchEngine that also can be used as a tool. Currently using Google Search as default engine. It also allows for new engines to be implemented using the SearchApiRequestResponseHandler interface, and adding it to the SearchApiEngine enum so the user can choose which one to use.

## General checklist

- [X] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
* The example is in the docs, I will open a new PR to the examples repo if it is ok

@algora-pbc /claim #1132


web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchEngineIT.java
```diff
@@ -0,0 +1,85 @@
...
+        // when
+        WebSearchRequest request = WebSearchRequest.builder()
+                .searchTerms("What is Langchain4j?")
+                .maxResults(5)
+                .build();
+        WebSearchResults webSearchResults = searchEngine.search(request);
+
+        // then
+        List<WebSearchOrganicResult> results = webSearchResults.results();
+        assertThat(results).hasSize(5);
```
Heezer: This assert fails. Obviously the suggestion https://github.com/langchain4j/langchain4j/issues/1132#issuecomment-2144059116 is not implemented.
OMG - there are two concurrent PRs for this feature 😄 I've never experienced that 😄

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchToolIT.java
```diff
@@ -0,0 +1,162 @@
...
+            .apiKey(System.getenv("OPENAI_API_KEY"))
+            .modelName(OpenAiChatModelName.GPT_3_5_TURBO)
+            .logRequests(true)
+            .build();
+
+    interface Assistant {
+        @dev.langchain4j.service.SystemMessage({
+                "You are a web search support agent.",
+                "If there is any event that has not happened yet",
+                "You MUST create a web search request with with user query and",
```
Heezer:
```suggestion
                "You MUST create a web search request with user query and",
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchToolIT.java
```diff
@@ -0,0 +1,162 @@
...
+        searchApiEngine = SearchApiWebSearchEngine.builder()
+                .apiKey(System.getenv("SEARCHAPI_API_KEY"))
+                .engine(SearchApiEngine.GOOGLE_SEARCH)
+                .build();
+
+        WebSearchTool webSearchTool = WebSearchTool.from(searchApiEngine);
+        List<ToolSpecification> tools = ToolSpecifications.toolSpecificationsFrom(webSearchTool);
+        String query = "What are the release dates for the movies coming out last week of May 2024?";
+        List<ChatMessage> messages = new ArrayList<>();
+        SystemMessage systemMessage = SystemMessage.from("You are a web search support agent. If there is any event that has not happened yet, you MUST use a web search tool to look up the information on the web. Include the source link in your final response. Do not say that you have not the capability to browse the web in real time");
```
Heezer: nice to have - to have this text as once e.g. as constant, instead of three times

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchToolIT.java
```diff
@@ -0,0 +1,162 @@
...
+    void should_execute_searchApi_tool_with_chatLanguageModel_to_summary_response_in_images() {
+        // given
+        searchApiEngine = SearchApiWebSearchEngine.builder()
+                .apiKey(System.getenv("SEARCHAPI_API_KEY"))
+                .engine(SearchApiEngine.GOOGLE_SEARCH)
+                .build();
+
+        WebSearchTool webSearchTool = WebSearchTool.from(searchApiEngine);
+        List<ToolSpecification> tools = ToolSpecifications.toolSpecificationsFrom(webSearchTool);
+        String query = "My family is coming to visit me in Madrid next week, list the best tourist activities suitable for the whole family";
```
Heezer: nice use case 😄 👍

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchToolIT.java
```diff
@@ -0,0 +1,162 @@
...
+        String strResult = webSearchTool.searchWeb(query);
+        ToolExecutionResultMessage toolExecutionResultMessage = ToolExecutionResultMessage.from(aiMessage.toolExecutionRequests().get(0), strResult);
+        messages.add(toolExecutionResultMessage);
+
+        AiMessage finalResponse = chatLanguageModel().generate(messages).content();
+        System.out.println(finalResponse.text());
+
+        // then
+        assertThat(finalResponse.text())
+                .as("At least the string result should be contains 'madrid' and 'tourist' ignoring case")
```
Heezer:
```suggestion
                .as("At least the result string should contain 'madrid' and 'tourist' ignoring case")
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchToolIT.java
```diff
@@ -0,0 +1,162 @@
...
+
+    @Test
+    void should_execute_searchApi_tool_with_AiServices() {
+        // given
+        WebSearchTool webTool = WebSearchTool.from(searchApiEngine);
+
+        Assistant assistant = AiServices.builder(Assistant.class)
+                .chatLanguageModel(chatModel)
+                .tools(webTool)
+                .chatMemory(MessageWindowChatMemory.withMaxMessages(10))
```
Heezer: chat memory isn't needed here, right?

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchToolIT.java
```diff
@@ -0,0 +1,162 @@
...
+                .build();
+        // when
+        String answer = assistant.answer("Search in the web who won the FIFA World Cup 2022?");
```
Heezer:
```suggestion
        String answer = assistant.answer("Who won the FIFA World Cup 2022?");
```

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,102 @@
...
+import dev.langchain4j.web.search.searchapi.SearchApiEngine;
+import dev.langchain4j.web.search.searchapi.SearchApiWebSearchEngine;
+
+public class SearchApiTool {
+
+    interface Assistant {
+        @dev.langchain4j.service.SystemMessage({
+                "You are a web search support agent.",
+                "If there is any event that has not happened yet",
+                "You MUST create a web search request with with user query and",
```
Heezer:
```suggestion
                "You MUST create a web search request with user query and",
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchToolIT.java
zambrinf: right, removed

zambrinf: also removed from the example

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,101 @@
...
+        String answer(String userMessage);
+    }
+
+    private static final String SEARCHAPI_API_KEY = "YOUR_SEARCHAPI_KEY";
+    private static final String OPENAI_API_KEY = "YOUR_OPENAI_KEY";
+
+    public static void main(String[] args) {
+        SearchApiWebSearchEngine searchEngine = SearchApiWebSearchEngine.builder()
+                .apiKey(SEARCHAPI_API_KEY)
+                .engine(SearchApiEngine.GOOGLE_SEARCH)
```
mark-searchapi: Maybe we can simplify it and just use a string? This way we don't need to rely on knowing the constant names

```suggestion
                .engine("google")
```

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,101 @@
...
+                .build();
+
+        WebSearchTool webTool = WebSearchTool.from(searchEngine);
+
+        Assistant assistant = AiServices.builder(Assistant.class)
+                .chatLanguageModel(chatModel)
+                .tools(webTool)
+                .build();
+
+        String answer = assistant.answer("My family is coming to visit me in Madrid next week, list the best tourist activities suitable for the whole family");
```
mark-searchapi, line 80: Great example. 👍

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/RequestResponseHandlerFactory.java
```diff
@@ -0,0 +1,18 @@
...
+class RequestResponseHandlerFactory {
+
+    static SearchApiRequestResponseHandler create(SearchApiEngine engine) {
+        switch (engine) {
+            case GOOGLE_SEARCH:
+                return new SearchApiGoogleSearchHandler();
+            default:
+                throw new RuntimeException("Response handler not found");
+        }
+    }
```
mark-searchapi: I think we should simplify the integration and always default to a generic parser that extracts `organic_results` array and collects only `title`, `link`, and `snippet`.

This way it is compatible with multiple SearchApi engines such as `google`, `bing`, `baidu`, `google_news`, `bing_news`, `google_scholar` and many upcoming ones. SearchApi maintains the same JSON key names across search engines.

Upcoming search engines such as `yandex`, `yahoo` and others would share the same JSON key names, and integrating them wouldn't require changing the `langchain4j` integration.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,90 @@
...
+        try {
+            switch (engine) {
+                case GOOGLE_SEARCH:
+                    SearchApiGoogleSearchApi api = retrofitBase.create(SearchApiGoogleSearchApi.class);
+                    Response<SearchApiGoogleSearchResponse> response = api.search(finalParameters, bearerToken)
+                            .execute();
+                    return getBody(response);
+                default:
+                    throw new RuntimeException("Invalid engine");
+            }
```
mark-searchapi: We can simplify the integration and just default to parsing the `organic_results` array consisting of `title`, `link` and `snippet`.

mark-searchapi: SearchApi uses the same `/v1/search` endpoint across all engines, so dropping dependency to Google (at least in naming `SearchApiGoogleSearchApi`) and making a generic GET request to SearchApi would make integration easier.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiEngine.java
mark-searchapi, line 1: `enum` is too much limiting. When SearchApi adds a new search engine, clients should not need to modify the langchain4j integration. Let's let the users use any `engine` they want, even if it does not exist. SearchApi will make sure to return validation errors in this case.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiGoogleSearchHandler.java
mark-searchapi, line 1: This again seems to be tightly coupled with the `google` search engine.

Would it be possible to build a `SearchApiWebSearchEngine` with a hash of key-value pairs instead (like a JSON `{ "gl: "us", "hl": "en", "num": 10 }`)? That way user could specify any parameters that he wants. It would work for any existing or upcoming SearchApi engine. User can check for query parameters in our documentation. If query parameters changes for any reason, it wouldn't require updating the `langchain4j` implementation.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiGoogleSearchHandler.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        SearchApiGoogleSearchResponse googleSearchResponse = (SearchApiGoogleSearchResponse) response;
+        List<WebSearchOrganicResult> results = googleSearchResponse.getOrganicResults()
+                .stream()
+                .map(result -> WebSearchOrganicResult.from(
+                        result.getTitle(),
+                        URI.create(result.getLink()),
+                        result.getSnippet(),
+                        null,  // by default google custom search api does not return content
+                        toResultMetadata(result) // TODO: not sure which information goes here
```
mark-searchapi: @czelabueno Could you share what is this used for? Can it be empty?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiGoogleSearchHandler.java
```diff
@@ -0,0 +1,80 @@
...
+                ))
+                .collect(Collectors.toList());
+        Long totalResults = googleSearchResponse.getSearchInformation().getTotalResults();
+        WebSearchInformationResult searchInformation = WebSearchInformationResult.from(totalResults,
+                googleSearchResponse.getPagination().getCurrent(),
+                googleSearchResponse.getSearchParameters() // TODO: not sure which information goes here
+        );
```
mark-searchapi: Please account that for some specific engines `total_results`, and `current` page might not be available (not present in the hash). Not throwing an exception in those cases would be preferred as the most important thing is collecting organic results.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiGoogleSearchHandler.java
```diff
@@ -0,0 +1,80 @@
...
+
+    private Map<String, String> toResultMetadata(SearchResult result) {
+        Map<String, AboutResult> aboutThisResult = result.getAboutThisResult();
+        if (aboutThisResult != null && aboutThisResult.containsKey("source")) {
+            AboutResult source = aboutThisResult.get("source");
+            return source.toMetadata();
+        } else {
+            return null;
+        }
+    }
```
mark-searchapi: Google removed the feature of displaying "about this result". Let's just remove it with the tight coupling to `google`. Also, it simplifies the integration.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiGoogleSearchResponse.java
mark-searchapi, line 1: Let's leave handle only the following to make it future-proof:

```json
{
  ...
  "organic_results": [
    {
      "title": "",
      "snippet": "",
      "link": ""
    },
    ...
  ]
}
```

For every successful search, we also always return:

```json
{
  ...
  "search_parameters": {
    "engine": "google",
    "q": "Your query",
    ...
  }
}
```

If that is useful, we can include the search parameters as metadata. It represents the parameters that were used to make this search.

In case of any error (validation or missing results), SearchAPI returns a JSON containing an `error` JSON key.

In the case of pagination, we can't guarantee the presence of any of the following JSON keys (`pagination`, `current` or `next`), but sometimes they are there. There might be cases, where the `pagination` object is missing. There might also be cases where `current`, or `next` JSON keys are missing in `pagination`. There might be cases where both `current` and `next` JSON keys are missing inside the `pagination` object due to different pagination techniques.

```json
  {
    ...
    "pagination": {
      "current": 1,
      "next": "link",
      "..."
    }
  }
```

Search information fields are not guaranteed as well, but sometimes appear in any combination (same explanation as for pagination).

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,69 @@
...
+    @Builder
+    public SearchApiWebSearchEngine(String apiKey,
+                                    Duration timeout,
+                                    SearchApiEngine engine) {
+        this.apiKey = ensureNotBlank(apiKey, "apiKey");
+        this.engine = getOrDefault(engine, DEFAULT_ENGINE);
+        this.handler = RequestResponseHandlerFactory.create(this.engine);
+        this.client = SearchApiClient.builder()
+                .baseUrl(BASE_URL)
+                .timeout(getOrDefault(timeout, ofSeconds(10)))
```
mark-searchapi: I think default timeout of 30 seconds is pretty common across HTTP libraries, so we can use the same here.

```suggestion
                .timeout(getOrDefault(timeout, ofSeconds(30)))
```

mark-searchapi (commented):
Great job @zambrinf, this PR is very close to complete! Thanks for great usage examples with OpenAI, this is what we were looking for. 💪 

SearchApi organic results are consistent across major search engines so we can simplify the integration and make it future proof (work without changes for upcoming search engines).

---

For anybody that wants a custom SearchAPI integration such as YouTube Transcripts, we could create a custom tool instead. It does not need to be related to the WebSearch functionality at all. As I understand, all it needs is marking a function as a tool.

```java
  public class YouTubeTranscriptsTool {

    private static final String API_URL = "https://www.searchapi.io/api/v1/search";
    private static final String ENGINE = "youtube_transcripts";

    @Tool(name = "fetchTranscripts", description = "Fetches transcripts for a given YouTube video ID")
    public String fetchTranscripts(
            @P(value = "videoId", description = "The ID of the YouTube video") String videoId) throws IOException {
        // Build the request URL with parameters
        String url = String.format("%s?engine=%s&video_id=%s", API_URL, ENGINE, videoId);

        // Create an HTTP client and execute the request
        try (CloseableHttpClient httpClient = HttpClients.createDefault()) {
            HttpGet request = new HttpGet(url);
            try (CloseableHttpResponse response = httpClient.execute(request)) {
                if (response.getStatusLine().getStatusCode() == 200) {
                    // Parse the response to JSON
                    String jsonResponse = EntityUtils.toString(response.getEntity());
                    return parseTranscripts(jsonResponse);
                } else {
                    throw new IOException("Failed to fetch transcripts: " + response.getStatusLine().getReasonPhrase());
                }
            }
        }
    }

    private String parseTranscripts(String jsonResponse) throws IOException {
        // Use Jackson to parse the JSON response
        ObjectMapper objectMapper = new ObjectMapper();
        JsonNode rootNode = objectMapper.readTree(jsonResponse);
        JsonNode transcriptsArray = rootNode.path("transcripts");

        // Combine all text values from the transcripts array
        StringBuilder combinedText = new StringBuilder();
        for (JsonNode transcript : transcriptsArray) {
            String text = transcript.path("text").asText();
            combinedText.append(text).append(" ");
        }

        return combinedText.toString().trim();
    }
}
```

The example above is crafted with the help of ChatGPT as I don't know enough Java myself. YoutubeTranscripts implementation is completely optional for this PR to be considered complete, but feel free to play with the idea if you want. 😊

zambrinf:
> SearchApi organic results are consistent across major search engines so we can simplify the integration and make it future proof (work without changes for upcoming search engines).

@mark-searchapi  Oh ok! I thought you wanted other integrations (like "youtube" and "google_maps" that do not have organic results in the response) in the WebSearchEngine, that is why I created the `SearchApiGoogleSearchApi.java`, `SearchApiGoogleSearchHandler.java`, `SearchApiGoogleSearchResponse.java` coupled to Google Search, so if you wanted to add other engines you could create a new customized implementation. I will take a look at your comments and address them.

mark-searchapi:
> I thought you wanted other integrations (like "youtube" and "google_maps" that do not have organic results in the response) in the WebSearchEngine

We were initially imagining other integrations (such as `youtube_transcripts`) as separate tools (completely custom, not related to WebSearchEngine or WebSearchTool), but probably couldn't articulate well enough in the beginning. Like having a BaseSearchAPITool that knows how to make a GET query (and that's it). Then extend that class and have a method that is being marked as a tool and a method to construct a String context that is being passed to LLM (such as OpenAI). Something like `YouTubeTranscriptsTool`, but with a base class that knows how to construct the URL and deals with api_key and parameter interpolation into the final URL for GET request.

The concept of `WebSearchEngine` is a little bit different and would not easily fit those other search engines. It is better to keep the `WebSearchEngine` integration simple & flexible. 😊

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiGoogleSearchHandler.java
zambrinf: the search method `WebSearchResults search(WebSearchRequest webSearchRequest)` has this `WebSearchRequest` parameter which contains some properties (they are not required) like maxResults, language, geoLocation, etc, and also a `Map<String, Object> additionalParams` which can act like a JSON, adding key/value pairs.  I changed the logic to always use the `additionalParams` so we don't get coupled, let me know if it's not ok.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiGoogleSearchHandler.java
zambrinf: now I set the default totalResults to the organic results size, is it ok? Or do you prefer setting to Zero?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiGoogleSearchHandler.java
zambrinf: set to null for now, let me know if it's not ok

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,101 @@
+---
+sidebar_position: 1
+---
+
+# SearchApi
+
+A wrapper around the SearchApi. This tool is handy when you need to answer questions about current events.
```
mark-searchapi: WDYT about this version?

```suggestion
[SearchApi](https://www.searchapi.io/) is a real-time SERP API. You can use it to perform searches in Google, Google News, Bing, Bing News, Baidu, Google Scholar, or any other engine that returns organic results.
```

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Add dependencies
```
mark-searchapi: Maybe we should have h2 (##) as a `Usage example` and then group other paragraphs as h3 (Dependency setup, Example code, etc..). Up for ideas for a better grouping.

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,101 @@
...
+
+or project's `build.gradle`:
+
+```groovy
+implementation 'dev.langchain4j:langchain4j-web-search-engine-searchapi:{your-version}'
+```
+
+### Try out an example code:
+
+[An Example of using SearchApi web search as a tool]()
```
mark-searchapi: It should probably not be an empty link.

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,101 @@
...
+         */
+    }
+}
+```
+
+### Available engines in Langchain4j
+
+| SearchApi Engine                                          | Available |
+|-----------------------------------------------------------|-----------|
+| [Google Web Search](https://www.searchapi.io/docs/google) | ✅         |
```
mark-searchapi, line 103: We can list Google, Google News, Bing, Bing News, Baidu, and Google Scholar. We can also add a comment that any other engine that returns `organic_results` array and has `title`, `link`, and `snippet` is supported.

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,101 @@
...
+
+### References
+
+[SearchApi website](https://www.searchapi.io)
```
mark-searchapi: This can probably be removed if we put a link at the top introduction to SearchApi.

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private static final String SEARCHAPI_API_KEY = "YOUR_SEARCHAPI_KEY";
+    private static final String OPENAI_API_KEY = "YOUR_OPENAI_KEY";
+
+    public static void main(String[] args) {
+        SearchApiWebSearchEngine searchEngine = SearchApiWebSearchEngine.builder()
+                .apiKey(SEARCHAPI_API_KEY)
+                .engine("google")
+                .build();
```
mark-searchapi: Could we add additional parameters `gl=us`, `hl=en`, and `google_domain=google.com`. They are default, but it would be nice to see how to add any additional parameters when constructing the search engine as these parameters influence the organic results in Google Search.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,130 @@
...
+                : response.getOrganicResults().size();
+    }
+
+    private Map<String, Object> toSearchInformationMetadata(SearchApiWebSearchResponse response) {
+        Map<String, Object> metadata = new HashMap<>();
+        Pagination pagination = response.getPagination();
+        if (pagination != null) {
+            metadata.put("current", pagination.getCurrent());
+            metadata.put("next", pagination.getNext());
+        }
```
mark-searchapi: Is it possible to add to `metadata` everything that is inside the `pagination` JSON key (if present) and not depend on specific JSON keys?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,130 @@
...
+        return pagination != null ? pagination.getCurrent() : null;
+    }
+
+    private Map<String, Object> toSearchMetadata(SearchApiWebSearchResponse response) {
+        Map<String, Object> metadata = new HashMap<>();
+        SearchParameters searchParameters = response.getSearchParameters();
+        if (searchParameters != null) {
+            metadata.put("engine", searchParameters.getEngine());
+            metadata.put("q", searchParameters.getQ());
+        }
```
mark-searchapi: Is it possible to put everything that is inside the `searchParameters` (if present) into metadata? It would be great to avoid tightly coupling parameter names as they are different across different search engines.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,130 @@
...
+        return metadata;
+    }
+
+    private List<WebSearchOrganicResult> toWebSearchOrganicResults(List<OrganicResult> organicResults) {
+        return organicResults.stream()
+                .map(result -> WebSearchOrganicResult.from(
+                        result.getTitle(),
+                        URI.create(result.getLink()),
+                        result.getSnippet(),
```
mark-searchapi: Can we fallback to an empty string just in case? There might be some cases in some engines where some organic results are missing the snippet (however, it is rare).

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchEngineIT.java
mark-searchapi, line 1: What does `IT` in `SearchApiGoogleSearchEngineIT` name mean? This is minor, but trying to understand the naming conventions.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchEngineIT.java
```diff
@@ -0,0 +1,112 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@EnabledIfEnvironmentVariable(named = "SEARCHAPI_API_KEY", matches = ".+")
+class SearchApiGoogleSearchEngineIT {
+
+    @Test
+    void should_search() {
+        // given
+        SearchApiWebSearchEngine searchEngine = SearchApiWebSearchEngine.builder()
```
mark-searchapi: Do you think adding a simple test file for `Bing` would be useful? It can be super simple with a single test case just to show that we can use a different search engine with SearchAPI.

mark-searchapi (commented):
From a logical standpoint, this PR is very strong. We haven't run the test cases ourselves, only reviewed the overall concept and functionality. We might run the tests after all comments are addressed. 

@langchain4j Could you please review the Java code? 🙏

docs/docs/integrations/web-search/searchapi.md
zambrinf: Maybe don't list the providers? Otherwise you should add when new ones are available, but with this library it will work automatically

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: sure

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchEngineIT.java
zambrinf: It's for Integration/Integrated Tests, usually its used for running specific tests on specific steps of the building process.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiGoogleSearchEngineIT.java
zambrinf: done!

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: yes, did that!

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: yes. Not directly related but I changed the structure a bit to do this, and WebSearchInformationResult required a totalResults, so I set it to the organicResults size as default.

docs/docs/integrations/web-search/searchapi.md
zambrinf: done

zambrinf:
@mark-searchapi everything ok, just in case I need to run more tests could you add more credits? I have only 2 left, my account is acc_qWRMkOg9leZVfwOz6Jze1b0P

zambrinf:
@mark-searchapi everything ok, just in case I need to run more tests could you add more credits? I have only 2 left, I sent a message in SearchApi website

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: this is what the metadata looks like adding everything to it automatically
image

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchApi.java
```diff
@@ -0,0 +1,15 @@
...
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Header;
+import retrofit2.http.QueryMap;
+
+import java.util.Map;
+
+interface SearchApiWebSearchApi {
+
+    @GET("/api/v1/search")
```
langchain4j:
```suggestion
    @GET("api/v1/search")
```
I have learned recently that when path starts with `/` in retrofit, it means absolute path and any sub-path in `baseUrl` will be ignored:

`baseUrl("http://example.com/sub/")` + `@GET("/api/v1/search")` -> `http://example.com/api/v1/search` (`sub` is ignored)
`baseUrl("http://example.com/sub/")` + `@GET("api/v1/search")` -> `http://example.com/sub/api/v1/search`

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,120 @@
...
+     *                           Default value is 30 seconds.
+     * @param engine             the engine used by Search API to execute the search
+     *                           <p>
+     *                           Default engine is Google Search.
+     * @param optionalParameters optional parameters to be passed on every request of this the engine, they can be overridden by the WebSearchRequest additional parameters for matching keys
+     *                           <p>
+     *                           Check <a href="https://www.searchapi.io">Search API</a> for more information on available parameters for each engine
+     */
+    @Builder
+    public SearchApiWebSearchEngine(String apiKey,
```
langchain4j, line 50: please add `String baseUrl` with the default value of `DEFAULT_BASE_URL` above

web-search-engines/langchain4j-web-search-engine-searchapi/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+
+        <dependency>
+            <groupId>com.squareup.retrofit2</groupId>
+            <artifactId>retrofit</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.squareup.retrofit2</groupId>
+            <artifactId>converter-gson</artifactId>
+        </dependency>
```
czelabueno: We've started a few months ago to move JSON lib in LLM integration to Jackson to be compatible with java frameworks (spring, quarkus, etc). Any web search integration would be used with a LLM integration together in a normal way. It's not critical but as new integration would be great if you use Jackson mapper.

```xml
<dependency>
    <groupId>com.squareup.retrofit2</groupId>
    <artifactId>converter-jackson</artifactId>
</dependency>
<dependency>
  </dependency>
```
Check [Anthropic](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java) or [Mistral AI](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/client/DefaultMistralAiClient.java) integration.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,76 @@
...
+import java.io.IOException;
+import java.time.Duration;
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.google.gson.FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES;
+
+class SearchApiClient {
+
+    private static final Gson GSON = new GsonBuilder()
```
czelabueno: Same here, Jackson mapper instead of

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,76 @@
...
+    SearchApiClient(String baseUrl, Duration timeout) {
+        OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder()
+                .callTimeout(timeout)
+                .connectTimeout(timeout)
+                .readTimeout(timeout)
+                .writeTimeout(timeout);
+        Retrofit retrofit = new Retrofit.Builder()
+                .baseUrl(baseUrl)
+                .client(okHttpClientBuilder.build())
+                .addConverterFactory(GsonConverterFactory.create(GSON))
```
czelabueno: Same here, `JacksonConverterFactory` instead of

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchApi.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.web.search.searchapi;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Header;
+import retrofit2.http.QueryMap;
+
+import java.util.Map;
+
+interface SearchApiWebSearchApi {
```
czelabueno: I think the name of that class is weird, here you still don't use webSearch as such. A better name might be `SearchApi` for the interface and `SearchApiClient` for the client.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
```diff
@@ -0,0 +1,76 @@
...
+
+    @Builder
+    SearchApiClient(String baseUrl, Duration timeout) {
+        OkHttpClient.Builder okHttpClientBuilder = new OkHttpClient.Builder()
+                .callTimeout(timeout)
+                .connectTimeout(timeout)
+                .readTimeout(timeout)
+                .writeTimeout(timeout);
+        Retrofit retrofit = new Retrofit.Builder()
+                .baseUrl(baseUrl)
```
czelabueno: Normally your Builder class should validate If  `baseUrl` is mandatory for the API. If you're using lombok could add them in your main constructor:

```suggestion
        if (isNullOrBlank(baseUrl)) {
            throw new IllegalArgumentException("baseUrl cannot be null or empty");
        }
        Retrofit retrofit = new Retrofit.Builder()
                        .baseUrl(baseUrl)
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,120 @@
...
+                                    Duration timeout,
+                                    String engine,
+                                    Map<String, Object> optionalParameters) {
+        this.apiKey = ensureNotBlank(apiKey, "apiKey");
```
czelabueno, line 55: I think that `apiKey` already is validated in `SearchApiWebSearchRequest.Builder`. If it's, pls remove it.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    @Override
+    public WebSearchResults search(WebSearchRequest webSearchRequest) {
+        SearchApiWebSearchRequest request = SearchApiWebSearchRequest.builder()
+                .apiKey(apiKey)
+                .engine(engine)
+                .query(webSearchRequest.searchTerms())
+                .optionalParameters(optionalParameters)
+                .additionalRequestParameters(webSearchRequest.additionalParams())
```
czelabueno, line 75: SearchApi doesn't support `maxResults`, `startPage` or `safeSearch`? If not, please consider taking them from `WebSearchRequest`. `WebSearchRequest.additionalParams` can be used to get non-common params normally specific by engine.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private WebSearchResults toWebSearchResults(SearchApiWebSearchResponse response) {
+        List<OrganicResult> organicResults = response.getOrganicResults();
+        Long totalResults = ((Integer) organicResults.size()).longValue(); // not ideal, but it may not be present in the response and is required not null by WebSearchInformationResult
+        WebSearchInformationResult searchInformation = WebSearchInformationResult.from(
+                totalResults,
+                getCurrentPage(response.getPagination()),
+                null
+        );
```
czelabueno: Modify my suggestion if "total_results" not is got of this way.
```suggestion
        WebSearchInformationResult searchInformation = WebSearchInformationResult.from(
                Long.valueOf(getOrDefault(response.getSearchInformation().get("total_results"),0)),
                getCurrentPage(response.getPagination()),
                null
        );
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,120 @@
...
+        Map<String, Object> searchMetadata = getOrDefault(response.getSearchParameters(), new HashMap<>());
+        addToMetadata(searchMetadata, response.getPagination());
+        addToMetadata(searchMetadata, response.getSearchInformation());
```
czelabueno: I think that `.getSearchInformation()` have own field and `.getPagination()` not is really relevant.
```suggestion
        Map<String, Object> searchMetadata = getOrDefault(response.getSearchMetadata(), new HashMap<>());
        addToMetadata(searchMetadata, response.getSearchParameters());
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    private List<WebSearchOrganicResult> toWebSearchOrganicResults(List<OrganicResult> organicResults) {
+        return organicResults.stream()
+                .map(result -> WebSearchOrganicResult.from(
+                        result.getTitle(),
+                        URI.create(result.getLink()),
+                        getOrDefault(result.getSnippet(), ""),
+                        null,  // by default google custom search api does not return content
+                        null
```
czelabueno: There not any `OrganicResult` additional info that fetch SearchAPI to put them as metadata? Something that add more info to the LLM about web page result could be good add it here.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiWebSearchEngineIT.java
```diff
@@ -0,0 +1,112 @@
...
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@EnabledIfEnvironmentVariable(named = "SEARCHAPI_API_KEY", matches = ".+")
+class SearchApiWebSearchEngineIT {
```
czelabueno:
```suggestion
class SearchApiWebSearchEngineIT extends WebSearchEngineIT {
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiWebSearchEngineBingIT.java
```diff
@@ -0,0 +1,41 @@
...
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@EnabledIfEnvironmentVariable(named = "SEARCHAPI_API_KEY", matches = ".+")
+class SearchApiWebSearchEngineBingIT {
```
czelabueno: BingIT could be joined as an additional test case in the `SearchApiWebSearchEngineIT`.

Instead of, you must add a new test class like this to test the `WebContentRetriever` scenario:

```java
class SearchApiWebSearchContentRetrieverIT extends WebSearchContentRetrieverIT
```

czelabueno (changes requested):
Hi @zambrinf ! You did a good job. I have added some things, small changes. It would be more neat to what is desired for a `WebSearchEngine` to support to the LLMs integration :)

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: @czelabueno As far as I understood from @mark-searchapi's review it is not standard among all engines how those parameters are sent, that's why we chose to decouple any implementation detail for a specific engine. See: https://github.com/langchain4j/langchain4j/pull/1216#discussion_r1678307385

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: @mark-searchapi is `search_information->total_results` a common naming for every engine of SearchApi?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: @czelabueno it was requested by @mark-searchapi to add everything to the metadata, again because every engine may have different namings.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
czelabueno: @zambrinf actually not, `startPage` and others is part of [OpenSearch standard](https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#opensearch-description-elements) (for all engines) as it's documented in `WebSearchRequest` so you can take them from the request and parse them to the expected key-value for each engine (e.g. google:?page=1;, bing:?page=1, etc). Anyway these are optional params.
If you prefer delegate the responsibility to the user to put every params as `additionalParam`, user should to be aware about all params supported by SearchApi by engine. It'd work same way. My only observation is that code could help to the user to set common params in the `WebSearchRequest` to do it easier. Otherwise documentation must be clear and up to date.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
czelabueno: @zambrinf  yes, I understand it, but let's not lose sight of the fact that it will be used as a `ContentRetriever` or `Tool` by an LLM, so it is necessary to analyze if it is relevant information to index it to the LLM in order to provide better responses to the user.
This [comment](https://github.com/langchain4j/langchain4j/pull/642/files#r1566496331) explains the current use of the metadata  and for now we are only extracting content from the `WebSearchOrganicResult`. In version 2, we will add the `WebSearchInformationResult` data so any other relevant content to be ingested to the LLM can be included there. `WebSearchResults.searchMetadata` would be used for debugging purpose.
The pagination is not considered content that adds any value to LLM for now and it could be included here `WebSearchResults.searchMetadata`

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
langchain4j: or just `.baseUrl(ensureNotBlank(baseUrl, "baseUrl"))`

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,120 @@
...
+     *                           Check <a href="https://www.searchapi.io">Search API</a> for more information on available parameters for each engine
+     */
+    @Builder
+    public SearchApiWebSearchEngine(String apiKey,
+                                    Duration timeout,
+                                    String engine,
+                                    Map<String, Object> optionalParameters) {
+        this.apiKey = ensureNotBlank(apiKey, "apiKey");
+        this.engine = getOrDefault(engine, DEFAULT_ENGINE);
+        this.optionalParameters = getOrDefault(optionalParameters, new HashMap<>());
```
langchain4j, line 57: please copy the map

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchRequest.java
```diff
@@ -0,0 +1,36 @@
...
+    @Builder
+    SearchApiWebSearchRequest(String engine,
+                              String apiKey,
+                              String query,
+                              Map<String, Object> optionalParameters,
+                              Map<String, Object> additionalRequestParameters) {
+        this.engine = ensureNotNull(engine, "engine");
+        this.apiKey = ensureNotBlank(apiKey, "apiKey");
+        this.query = ensureNotBlank(query, "query");
+        this.finalOptionalParameters = optionalParameters;
```
langchain4j, line 28: please make a copy of `optionalParameters`, otherwise it will be polluted with parameters from multiple requests

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    private List<WebSearchOrganicResult> toWebSearchOrganicResults(List<OrganicResult> organicResults) {
+        return organicResults.stream()
+                .map(result -> WebSearchOrganicResult.from(
+                        result.getTitle(),
+                        URI.create(result.getLink()),
+                        getOrDefault(result.getSnippet(), ""),
+                        null,  // by default google custom search api does not return content
```
langchain4j: copy-pasted comment

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiWebSearchEngineIT.java
langchain4j: @zambrinf please also remove tests in this class which are already present in `WebSearchEngineIT`

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,109 @@
...
+## Usage
+
+### Dependencies setup
+
+Add the following dependencies to your project's `pom.xml`:
+```xml
+<dependency>
+  <groupId>dev.langchain4j</groupId>
+  <artifactId>langchain4j-web-search-engine-searchapi</artifactId>
+  <version>{your-version}</version> <!-- Specify langchain4j version here -->
```
langchain4j, line 18:
```suggestion
  <version>0.34.0</version>
```

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,109 @@
...
+</dependency>
+```
+
+or project's `build.gradle`:
+
+```groovy
+implementation 'dev.langchain4j:langchain4j-web-search-engine-searchapi:{your-version}'
```
langchain4j, line 25:
```suggestion
implementation 'dev.langchain4j:langchain4j-web-search-engine-searchapi:0.34.0'
```

docs/docs/integrations/web-search/searchapi.md
```diff
@@ -0,0 +1,109 @@
...
+import dev.langchain4j.model.openai.OpenAiChatModelName;
+import dev.langchain4j.service.AiServices;
+import dev.langchain4j.web.search.WebSearchTool;
+import dev.langchain4j.web.search.searchapi.SearchApiEngine;
+import dev.langchain4j.web.search.searchapi.SearchApiWebSearchEngine;
+
+public class SearchApiTool {
+
+    interface Assistant {
+        @dev.langchain4j.service.SystemMessage({
```
langchain4j, line 43:
```suggestion
        @SystemMessage({
```

web-search-engines/langchain4j-web-search-engine-searchapi/pom.xml
```diff
@@ -0,0 +1,90 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>0.32.0</version>
```
langchain4j:
```suggestion
        <version>0.33.0</version>
```

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchApi.java
mark-searchapi: Great suggestion! 👍

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
mark-searchapi: If the user chooses to use SearchApi, he must have the API key. It should be expected that the user checks the SearchApi documentation for the available parameters. Despite we have some pretty common parameters such as `num`, and `page`, they might have different restrictions and different default values based on search engines. For `safeSearch`, Bing has [safe_search parameter](https://www.searchapi.io/docs/bing#api-parameters-filters-safe-search), but Google has [safe parameter](https://www.searchapi.io/docs/google#api-parameters-filters-safe). Supporting conversions at the integration level is not reasonable (might get out of date pretty quickly and someone needs to be responsible for updating every single integration. Instead, the integration can be flexible by accepting anything, such as through `additionalParam`. 

SearchApi documentation is always up to date. the langchain4j library user should have easily accessible links to check the relevant search engines documentation, which right now is mentioned in `docs/docs/integrations/web-search/searchapi.md`. If there are ways to have extra documentation links in javadoc to SearchApi, that might as well help the user quickly check for available search engine additional params.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
mark-searchapi: Yes, `search_information->total_results` is a common name across all search engines. It might be missing (if the search engine for some reason does not return it), but that does not mean there are no results.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
mark-searchapi: I wasn't aware of metadata use cases. Since `WebSearchResults.searchMetadata` is used for debugging, we can put everything that is in our `search_metadata` root key. It includes information about search execution time and has links to the JSON/HTML for debugging purposes.

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
mark-searchapi: `position` might be relevant to know which organic result is more important than the other. It is common across all search engines.

Apart from that, each search engine might have a different set of fields that might be interesting. However, WebSearchEngine should be universal, so `title`, `link`, and `snippet` should suffice.

SearchApi supports 1k+ unique elements in Google and we are still rapidly expanding it. It contains not only the organic results, but `knowledge_graph`, `sports_results`, `answer_box` (with various types of answers such as finance, population, location, etc..), `top_stories`, `popular_places`, and probably hundreds of others. Integrating this extra information would require plugging the whole response or drafting a custom tool that slices the JSON and leaves only the relevant bits for context given to LLM. As the LLM context window gets larger and cheaper, you will probably want to use the whole SearchAPi response as a LLM context.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiWebSearchEngineIT.java
zambrinf: @czelabueno it fails both `WebSearchEngineIT ` tests because we don't use `maxResults` from `WebSearchRequest` and the other does not return 5 by default, what can we do? I thought of changing the expected size in  `WebSearchEngineIT ` to a Constant that can be used from this test class. Also there is this problem: https://github.com/langchain4j/langchain4j/issues/1132#issuecomment-2143201554 and https://github.com/langchain4j/langchain4j/issues/1132#issuecomment-2144059116, can we change the query from `WebSearchEngineIT `?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
zambrinf: @czelabueno even this being already validated in `SearchApiWebSearchEngine` constructor which uses this builder?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: I think it's better to validate here and remove from `SearchApiWebSearchRequest`, otherwise it would fail only when calling the search method

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: As I understood we will leave this way so I will resolve it, you can unresolve if necessary.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiWebSearchEngineIT.java
zambrinf: please take a look at this changes that fix the problem above https://github.com/langchain4j/langchain4j/pull/1216/files#diff-716bbaaa3beacbc1a856a3732d400e795171cfd8b2f9bfd5c864750a56231c16

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: @mark-searchapi is the organic result `position` field default for all engines?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiWebSearchEngine.java
zambrinf: @mark-searchapi In yours Baidu docs it doesn't have `search_metadata`, is it optional?

web-search-engines/langchain4j-web-search-engine-searchapi/src/main/java/dev/langchain4j/web/search/searchapi/SearchApiClient.java
czelabueno: @zambrinf Yeah, what I mean is that your `SearchApiClient` should have all the constraints validated for a well communication with `SearchApi`. `SearchApiWebSearchEngine` uses `SearchApiClient` and it should have the only responsibility of passing user values ​​or setting default values. My observation is more related to putting the validation where it makes the most sense.

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiContentRetrieverIT.java
```diff
@@ -0,0 +1,60 @@
...
+            .apiKey(System.getenv("OPENAI_API_KEY"))
+            .modelName(GPT_4_O_MINI)
+            .logRequests(true)
+            .build();
+
+    @Override
+    protected WebSearchEngine searchEngine() {
+        return SearchApiWebSearchEngine.builder()
+                .apiKey(System.getenv("SEARCHAPI_API_KEY"))
+                .build();
```
czelabueno: Could you return the same `searchEngine`? or there some reason you've instantiated it  again?

czelabueno (commented):
Hi @zambrinf only two bit comments. If you solve it, we would be completed :)
Thanks for your contribution!

web-search-engines/langchain4j-web-search-engine-searchapi/src/test/java/SearchApiContentRetrieverIT.java
zambrinf: sure, it was just a lapse, thanks

czelabueno (approved):
Thank so much @zambrinf. I submit my approval with this comment.

SebastjanPrachovskij (approved):
Thank you @zambrinf! It looks all good from our (SearchApi) side as well. 

@langchain4j we can proceed with the merge.

ghost (approved):
@zambrinf thank you for the implementation!
@Heezer @mark-searchapi @czelabueno @SebastjanPrachovskij thank you for reviews!

Issue #301, opened 2023-11-20
[FEATURE] Add support for Update/upsert in EmbeddingStore


pascalconfluent:
EmbeddingStore should support delete and update of embeddings.

langchain4j:
@pascalconfluent delete/update by ID or bulk (delete all, replace all with new embeddings)?

igerdy:
The implementation of milvus can be deleted by collection or partition. I am doing extends locally and I am not sure about other stores, so I did not submit a PR. If there is a defined interface specification, I can adapt the implementation of milvus. @langchain4j

noday:
When can #1020 be used?

langchain4j:
Delete is (partially, only for PGVector) implemented in https://github.com/langchain4j/langchain4j/pull/1113

langchain4j:
@noday you can already use it in 0.31.0

igerdy:
@noday I added some functions to my project for reference only:
@
https://github.com/igerdy/langchain4j/blob/0.30.0-plus/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusEmbeddingStore.java

InAnYan:
Hello! It would be cool if this feature was implemented. What's the status?

I think it's a very important issue, consider an example:

I have an app that stores a collection of documents. The user may add document and the app should generate embeddings out of it. But the user has the ability to remove a document. Then, the embeddings better to be removed too)

What if we could just support removing embeddings for `InMemoryEmbeddingsStore`?

langchain4j:
Hi @InAnYan please see the comments above, it is partially supported (currently only for PGVector) in the latest version. Yes, adding it to `InMemoryEmbeddingsStore` should be simple, would you like to help us with that?

PR #1220, opened 2024-06-03
Implement remove methods for InMemoryEmbeddingStore

InAnYan:
## Issue
[https://github.com/langchain4j/langchain4j/issues/301](https://github.com/langchain4j/langchain4j/issues/301)

## Change
I've implemented remove methods for InMemoryEmbeddingStore.

Current problems:
1. I'm not completely sure how to test all of this. And, as I've seen, you don't fully test it too)
2. There is a very nasty problem with `removeAll(Filter filter)` . Let me try to break it in small pieces:
   1. `EmbeddingStore`'s are parametrized with `Embedded` type parameter.
   2. `removeAll(Filter filter)` method in `EmbeddingStore` accepts `Filter`.
   3. `Filter` accepts `Metadata` object.
   4. `Metadata` object is stored in `TextSegment`.
   5. `TextSegment` is usually what is `Embedded`. But it's not guaranteed.

How my pull request deals with this problem: 
1) for every entry in `InMemoryEmbeddingStore` check the type of `Embedded`. 
2) If it's a `TextSegment`, then proceed.
3) If it's not, then raise `UnsupportedOperationException`.
The issue with this strategy is that we perform the check for every entry. I haven't found a way in Java to check the type of type parameter (because it's erasured in runtime). (*In C++, for example, there are `constexpr if`, `<type_traits>`, `std::is_same<U, V>`, explicit specialization, etc.*)

We could solve this problem if `InMemoryEmbeddingStore` would support only `TextSegment` as `Embedded`, but that's a breaking change.

I could also not implement this method, but, I really really need it in my project (if we chose to stick with `InMemoryEmbeddingStore`).
  

## General checklist

### DIFF
--- a/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithRemovalIT.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithRemovalIT.java
@@ -1,0 +1,119 @@
+package dev.langchain4j.store.embedding;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import dev.langchain4j.data.document.Metadata;
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.data.segment.TextSegment;
+import dev.langchain4j.store.embedding.filter.Filter;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static dev.langchain4j.store.embedding.filter.MetadataFilterBuilder.metadataKey;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public abstract class EmbeddingStoreWithRemovalIT extends EmbeddingStoreIT{
+    @BeforeEach
+    void beforeEach() {
+        embeddingStore().removeAll();
+    }
+
+    @Test
+    void remove_by_id() {
+        Embedding embedding = embeddingModel().embed("hello").content();
+        Embedding embedding2 = embeddingModel().embed("hello2").content();
+        Embedding embedding3 = embeddingModel().embed("hello3").content();
+
+        String id = embeddingStore().add(embedding);
+        String id2 = embeddingStore().add(embedding2);
+        String id3 = embeddingStore().add(embedding3);
+
+        assertThat(id).isNotBlank();
+        assertThat(id2).isNotBlank();
+        assertThat(id3).isNotBlank();
+
+        List<EmbeddingMatch<TextSegment>> relevant = embeddingStore().findRelevant(embedding, 10);
+        assertThat(relevant).hasSize(3);
+
+        embeddingStore().remove(id);
+
+        relevant = embeddingStore().findRelevant(embedding, 10);
+        List<String> relevantIds = relevant.stream().map(EmbeddingMatch::embeddingId).collect(Collectors.toList());
+        assertThat(relevantIds).hasSize(2);
+        assertThat(relevantIds).containsExactly(id2, id3);
+    }
+
+    @Test
+    void remove_all_by_ids() {
+        Embedding embedding = embeddingModel().embed("hello").content();
+        Embedding embedding2 = embeddingModel().embed("hello2").content();
+        Embedding embedding3 = embeddingModel().embed("hello3").content();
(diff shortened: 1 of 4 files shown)


langchain4j:
Hi @InAnYan thanks a lot!

> I'm not completely sure how to test all of this. And, as I've seen, you don't fully test it too)

I think we can extract tests from `PgVectorEmbeddingStoreRemoveIT` into `EmbeddingStoreWithRemovalIT` in a `langchain4j-core` module, similar to `EmbeddingStoreIT` and every `EmbeddingStore` that implements removal will extend this test. WDYT?

Regarding the `TextSegment` vs `Embedded`, I plan to review this "problem" soon and see if this sohuld be changed. Originally there was an idea to allow other types of objects to be stored in the `EmbeddingStore`, and I think we might soon start supporting this.
Your current implementation seems ok.

InAnYan:
Hi, @langchain4j. Been working on implementing tests.

Changes I made:
1. I've created `EmbeddingStoreWithRemovalIT` as you proposed. Now I see where the actual tests are)
2. Implemented `InMemoryEmbeddingStoreWithRemovalIT`.
3. Also changed methods a little bit to conform to internal standards of your library.

So far I have these questions:
1. PGVector somehow handles `null` embedded values. However my code relies that `embedded` is present. It throws "not supported operation". Here are two ways of how to solve this issue:
	1) What I did is I added `TextSegment`s to testing code. I did this because I'm not really sure how `EmbeddingStore`s implemenation should handle `null` embedded values.
	2) Another option is to change `InMemoryEmbeddingStore.removeAll(Filter filter)`, change the lambda, and ignore `null` embedded. That way I completely don't change any testing code from `PgVectorStoreRemoveIT`.
2. `PgVectorStoreRemoveIT` relies on `EmbeddingStore.findRelevant` method which is considered depricated. Should we somehow change the test cases in `EmbeddingStoreWithRemovalIT` to accomodate this?
3. Should I delete `PgVectorStoreRemoveIT`? Or, actually, make it use the class I made: `EmbeddingStoreWithRemovalIT`.

langchain4j:
Hi @InAnYan 

> PGVector somehow handles null embedded values. However my code relies that embedded is present. It throws "not supported operation". Here are two ways of how to solve this issue:
> What I did is I added TextSegments to testing code. I did this because I'm not really sure how EmbeddingStores implemenation should handle null embedded values.
> Another option is to change InMemoryEmbeddingStore.removeAll(Filter filter), change the lambda, and ignore null embedded. That way I completely don't change any testing code from PgVectorStoreRemoveIT.

I am not sure I completely understand the problem, could you please elaborate? Usually `EmbeddingStore` can store `Embedding` alone, without `TextSegment`. This is one of the use cases when text data is stored elsewhere and correlated by ID.

> PgVectorStoreRemoveIT relies on EmbeddingStore.findRelevant method which is considered depricated. Should we somehow change the test cases in EmbeddingStoreWithRemovalIT to accomodate this?

Would be great to use the new `EmbeddingStore.search()` API!

> Should I delete PgVectorStoreRemoveIT? Or, actually, make it use the class I made: EmbeddingStoreWithRemovalIT.

Yes, please make `PgVectorStoreRemoveIT` inherit tests from `EmbeddingStoreWithRemovalIT`

InAnYan:
Hello, @langchain4j. Here is my work:
1. About `null` embedded value. Never mind, now I see that it is allowed to use it, so I just skip it in `removeAll(Filter filter)`, because it doesn't contain any metadata.
2. I've changed the usages of `EmbeddingStore.findRelevant`, now they use `search`.
3. `PgVectorEmbeddingStoreRemoveIT` now extends from `EmbeddingStoreWithRemovalIT`.

Tests for `InMemoryEmbeddingStore` and `PgVectorEmbeddingStore` all work.

---

Besides, why mark `EmbeddingStore.findRelevant` as deprecated? Why not implement it like this:
```java
List<EmbeddingMatch<TextSegment>> findRelevant(Embedding embedding, int maxResults) {
	EmbeddingSearchRequest embeddingSearchRequest = EmbeddingSearchRequest
            .builder()
            .queryEmbedding(embedding)
            .maxResults(maxResults)
            .build();

    return search(embeddingSearchRequest).matches();
}

langchain4j:
> Besides, why mark EmbeddingStore.findRelevant as deprecated?

The old `findRelevant` API is too inflexible to be extended with more features. Going forward, only the new `search` API will be extended with more features.

ghost (approved):
@InAnYan thank you!

Issue #1219, opened 2024-06-03
[FEATURE] Implement searching on docs website


InAnYan:
**Is your feature request related to a problem? Please describe.**
[https://docs.langchain4j.dev/](https://docs.langchain4j.dev/) contains lot's of useful information. I often combine information from the code documentation and tutorials. But in order to do this I need to search in the comboboxes (or accordions, or some UI elements) on the website, while I think a search feature would be easier.

**Describe the solution you'd like**
Implement search feature in [https://docs.langchain4j.dev/](https://docs.langchain4j.dev/).

**Describe alternatives you've considered**
*Mind reading* - currently impossible.

**Additional context**

agoncal:
Another easy option is to have the ability to also get a one-html-page format (and why not having also a PDF format so it can be easily downloaded). Image that you have the entire documentation (from https://docs.langchain4j.dev/intro) into a one-single page, then you can easily search. And it's perfect when being offline.

langchain4j:
@agoncal do you know how to do that?

agoncal:
So that means that we can close this issue, right ?

https://github.com/langchain4j/langchain4j/pull/1221

PR #1221, opened 2024-06-03
Add docusaurus-lunr-search plugin

InAnYan:
## Issue
[https://github.com/langchain4j/langchain4j/issues/1219](https://github.com/langchain4j/langchain4j/issues/1219)

## Change
Added the `docusaurus-lunr-search` plugin.

It may seem that for search feature we would need a server. But nothing stops us from passing that logic to the client)

So, if the index for pages is not too big, we will delegate the search functionality to client. That's what this plugin does (I think).

It works with `npm run serve`, but I couldn't test it with just opening the `index.html` file.

Oh, I'm sorry, my editor autoformatted the file)

## General checklist


InAnYan:
@langchain4j Would be very nice, instead of using crappy Google.

Hope this won't be a big problem to test this plugin. It works locally, I hope it will work on GitHub static websites

ghost (approved):
@InAnYan thanks! Let's try to deploy it...

InAnYan:
@langchain4j Is it deployed?

I remember having it run locally and it (search bar or button) was in the upper right corner

langchain4j:
@InAnYan deployed, seems to be working, thank you! 🤗

LizeRaes:
Wow smashing, works super fast. Thanks a lot!!!

PR #1222, opened 2024-06-03
Add a Tokenizer to Azure OpenAI

agoncal:
Today the Azure OpenAI module does not have its own Tokenizer. It actually uses the OpenAI one during its test phase (`langchain4j-open-ai` dependency in `test` scope).

This PR is for Azure OpenAI to have its own Tokenizer. Not just for test purposes (this PR removes the dependency to `langchain4j-open-ai`) but also Azure OpenAI can be used for RAG without the need of the `OpenAITokenizer`.

For that I had to change all the `ModelName` classes (`AzureOpenAiChatModelName`, `AzureOpenAiEmbeddingModelName`...) to add the model name (eg. `gpt-35-turbo-0301`), the model type (eg. `gpt-3.5-turbo`, notice the dot `.` in `3.5`) and version.

The `AzureOpenAiTokenizer` is a copy/paste from `OpenAiTokenizer`.

All `AzureOpenAiTokenizerTest` tests pass.

I need some help with `AzureOpenAiTokenizerIT`.  It uses the `ChatCompletionModel modelName` which has dots `.` in the names 

```
dev.ai4j.openai4j.chat;

public enum ChatCompletionModel {

    GPT_3_5_TURBO("gpt-3.5-turbo"),
```

So the `deploymentName` is not valid for Azure OpenAI. I am incline to use `AzureOpenAiChatModelName` instead of `ChatCompletionModel` but just want to know if it's the right approach

```
    void should_count_tokens_in_messages(List<ChatMessage> messages, ChatCompletionModel modelName) {

        // given
        AzureOpenAiChatModel model = AzureOpenAiChatModel.builder()
                .deploymentName(modelName.toString())
                .maxTokens(1) // we don't need outputs, let's not waste tokens
                .build();
}
```

## General checklist

### DIFF
--- a/langchain4j-azure-open-ai/pom.xml
+++ b/langchain4j-azure-open-ai/pom.xml
@@ -22,11 +22,6 @@
         </dependency>
 
         <dependency>
-            <groupId>dev.ai4j</groupId>
-            <artifactId>openai4j</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>com.azure</groupId>
             <artifactId>azure-ai-openai</artifactId>
         </dependency>
--- a/langchain4j-azure-open-ai/pom.xml
+++ b/langchain4j-azure-open-ai/pom.xml
@@ -22,6 +22,11 @@
         </dependency>
 
         <dependency>
+            <groupId>dev.ai4j</groupId>
+            <artifactId>openai4j</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>com.azure</groupId>
             <artifactId>azure-ai-openai</artifactId>
         </dependency>
@@ -41,9 +46,8 @@
         </dependency>
 
         <dependency>
-            <groupId>dev.langchain4j</groupId>
-            <artifactId>langchain4j-open-ai</artifactId>
-            <scope>test</scope>
+            <groupId>com.knuddels</groupId>
+            <artifactId>jtokkit</artifactId>
         </dependency>
 
         <dependency>
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModelName.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModelName.java
@@ -46,15 +46,15 @@ public enum AzureOpenAiChatModelName {
         this.modelVersion = modelVersion;
     }
 
-    public String getModelName() {
+    public String modelName() {
         return modelName;
     }
 
-    public String getModelType() {
+    public String modelType() {
         return modelType;
     }
 
-    public String getModelVersion() {
+    public String modelVersion() {
         return modelVersion;
     }
 
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModelName.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModelName.java
@@ -6,37 +6,60 @@ package dev.langchain4j.model.azure;
  */
 public enum AzureOpenAiChatModelName {
 
-    GPT_3_5_TURBO("gpt-35-turbo"), // alias for the latest gpt-3.5-turbo model
-    GPT_3_5_TURBO_0301("gpt-35-turbo-0301"), // 4k context, functions
-    GPT_3_5_TURBO_0613("gpt-35-turbo-0613"), // 4k context, functions
-    GPT_3_5_TURBO_1106("gpt-35-turbo-1106"), // 16k context, functions
+    GPT_3_5_TURBO("gpt-35-turbo", "gpt-3.5-turbo"), // alias for the latest gpt-3.5-turbo model
+    GPT_3_5_TURBO_0301("gpt-35-turbo-0301", "gpt-3.5-turbo", "0301"), // 4k context, functions
+    GPT_3_5_TURBO_0613("gpt-35-turbo-0613", "gpt-3.5-turbo", "0613"), // 4k context, functions
+    GPT_3_5_TURBO_1106("gpt-35-turbo-1106", "gpt-3.5-turbo", "1106"), // 16k context, functions
 
-    GPT_3_5_TURBO_16K("gpt-35-turbo-16k"), // alias for the latest gpt-3.5-turbo-16k model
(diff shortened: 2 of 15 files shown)


agoncal:
Pinging @jdubois for a double check

langchain4j-azure-open-ai/pom.xml
```diff
@@ -21,6 +21,11 @@
             <artifactId>langchain4j-core</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>dev.ai4j</groupId>
+            <artifactId>openai4j</artifactId>
```
langchain4j: can probably be removed now?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModelName.java
```diff
@@ -6,37 +6,60 @@
...
+        this.modelVersion = null;
+    }
+
+    AzureOpenAiChatModelName(String modelName, String modelType, String modelVersion) {
+        this.modelName = modelName;
+        this.modelType = modelType;
+        this.modelVersion = modelVersion;
+    }
+
+    public String getModelName() {
```
langchain4j:
```suggestion
    public String modelName() {
```
Please drop "get" from getters (in all classes), we follow modern record-style naming convention

langchain4j:
@agoncal thanks a lot!

> I am incline to use AzureOpenAiChatModelName instead of ChatCompletionModel but just want to know if it's the right approach

Sounds good.

BTW, `OpenAiTokenizer`/`OpenAiTokenizerIT` are probably a bit outdated.
We need to revisit them again and check if is really needed any maybe they can be simplified.

Why `OpenAiTokenizer` was created in the first place and how it can be helpful:
- calculate tokens in streaming mode. For a long time OpenAI did not provide token usage information when streaming, but now it does, so it becomes less important.
- estimate request cost before doing the request. Not sure how needed this feature is right now, I don't have any feedback about it.
- split text for RAG. As context sizes increase, it becomes less important to split documents and to be very precise about token calculation in this case. 1 word/symbol = N tokens might be a reasonable approximation going forward.

jdubois:
Wow @agoncal I wanted to code this too, but it seemed a ton of work, so happy you did this! It seems complicated to have a correct estimation here, but as commented by @langchain4j it seems less important today.

agoncal:
@langchain4j I changed the model names so it uses the Azure ones. But this is difficult to test due to the combination of models. I only managed to have one test pass with GPT 3.5 Turbo. Don't know if it's enough. But as you said, Tokenizers are becoming less important.

Let me know what you think.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiTokenizer.java
```diff
@@ -0,0 +1,404 @@
...
+ * There are integration tests in place that are making sure that the calculations here are very close to that of OpenAI.
+ */
+public class AzureOpenAiTokenizer implements Tokenizer {
+
+    private final String modelName;
+    private final Optional<Encoding> encoding;
+
+    /**
+     * Creates an instance of the {@code AzureOpenAiTokenizer} for the "gpt-3.5-turbo" model.
+     * It should be suitable for all current OpenAI models, as they all use the same cl100k_base encoding.
```
langchain4j: Not true any more, gpt-4o uses 200k tokens now

langchain4j:
@agoncal depends on what this tokenizer is going to be used for, a lot of existing complexity can be removed.

Or we can keep it as it is for the moment and if I will be updating OpenAiTokenizer/tests, I can update this one as well.

langchain4j-azure-open-ai/pom.xml
agoncal: `AzureOpenAiTokenizerTest` uses `openai4j` in the following methods:

```
    @ParameterizedTest
    @EnumSource(ChatCompletionModel.class)
    void should_support_all_chat_models_from_openai4j(ChatCompletionModel model)

    @ParameterizedTest
    @EnumSource(EmbeddingModel.class)
    void should_support_all_embedding_models_from_openai4j(EmbeddingModel model) {

    @ParameterizedTest
    @EnumSource(CompletionModel.class)
    void should_support_all_language_models_from_openai4j(CompletionModel model) {
```

But I don't know if that makes sense anymore as we have:

```
    @ParameterizedTest
    @EnumSource(AzureOpenAiChatModelName.class)
    void should_support_all_chat_models(AzureOpenAiChatModelName modelName) {
```

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModelName.java
agoncal: Done in https://github.com/langchain4j/langchain4j/pull/1222/commits/594231b7c2976aa9a858ad6ca044cfe63d6d5272

langchain4j-azure-open-ai/pom.xml
langchain4j: Ah, I see. All `from_openai4j` tests can be removed as azure module does not rely on openai4j lib

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiTokenizer.java
agoncal: Done https://github.com/langchain4j/langchain4j/pull/1222/commits/dbee1dd08af94844ab00c59ab522742995b4ca88

langchain4j-azure-open-ai/pom.xml
agoncal: Done https://github.com/langchain4j/langchain4j/pull/1222/commits/8c13cb36fa8cdc71a900177d3d5d78d6362c0291

agoncal:
@langchain4j yes, it's a copy/paste from `OpenAiTokenizer`, so if `OpenAiTokenizer` is updated, then we will update this one. It's just so Azure OpenAI does not depend on OpenAI and can have its own tokenizer, mostly for RAG purpose.

ghost (approved):
@agoncal thank you!

Issue #124, opened 2023-08-22
[FEATURE] Option to directly return from tool call


chenhaoqiang:
In a scenario, I ask the AI, for example: "Help me take a leave tomorrow." and it will trigger a leave request page form with my name and the date pre-filled. Then, I can input other leave details like the reason on the form.
I want to use the function calling feature of OpenAI to achieve this functionality. I define a leaveRequest function with parameters for name and leave date, and then obtain the response parameter: function_call. I pass this function_call to the leave request page form to pre-fill the values.
I don't want langchain4j to automatically call the chatLanguageModel.sendMessages() method. Instead, I want to directly return the toolExecutionRequest so that I can carry the parameters and trigger the leave request page form.
![image](https://github.com/langchain4j/langchain4j/assets/20062034/bea0cac8-c0ad-4eaf-86cf-afb5c62f7375)

langchain4j:
@chenhaoqiang pls help me understand your use case better. You want to receive AiMessage containing toolExecutionRequest, instead of langchain4j automatically calling the tool/function and returning the result to the LLM? Or do you want langchain4j call the tool/function and then return the result of the tool execution to the user instead of to the LLM? Is it expected that LLM will be involved again after your manipulations?

chenhaoqiang:
> @chenhaoqiang pls help me understand your use case better. You want to receive AiMessage containing toolExecutionRequest, instead of langchain4j automatically calling the tool/function and returning the result to the LLM? Or do you want langchain4j call the tool/function and then return the result of the tool execution to the user instead of to the LLM? Is it expected that LLM will be involved again after your manipulations?

I want to receive AiMessage containing toolExecutionRequest, LLM will not be involved again. I just want the first step to return.
![370b499bde4740c6965ef20d9fa88178](https://github.com/langchain4j/langchain4j/assets/20062034/e5704853-3d8d-49c7-9e45-59fc14f21e1c)

chenhaoqiang:
@langchain4j  I tried 0.21.0, I tested Azure OpenAI and ChatMemoryProvider with MariaDB and it worked great!!! Thx!

langchain4j:
[...]

Issue #859, opened 2024-03-28
[FEATURE] Improve reliability of structured outputs in AI Services


langchain4j:
- [x] If LLM responds with a JSON wrapped in backticks and/or some additional text, try to extract that instead of failing
- [ ] Automatically enable [JSON mode](https://docs.langchain4j.dev/tutorials/ai-services#json-mode) if return type of AI Service is POJO and underlying LLM supports it
- [ ] (Configurable) Retry in case of parsing failure, feeding the error back to the LLM
- [ ] If return type is enum/boolean, automatically use `logitBias` if supported by the LLM
- [ ] Consider adding function calling as a second option for structured outputs
- [ ] Consider using existing libraries to generate/validate JSON schema

ruizrube:
Hi,

The templates for parsing outputs were developed and tested using ChatGPT, isn't?. Open LLMs, such as Llama, are very more verbose in their answers.

kind regards. 
Thank you for such a stunning project.

langchain4j:
@ruizrube yes, most of the high-level features were developed with at least `gpt-3.5-turbo`-level LLM in mind.

But now when we have llama3/mixtral/etc and [json mode](https://docs.langchain4j.dev/tutorials/ai-services#json-mode) in Ollama, it should work as good for most of the use cases.

But we need to implement this PR anyway to ensure that a long tail of corner cases is handled well.

langchain4j:
Related: https://github.com/langchain4j/langchain4j/issues/1087

langchain4j:
Related: https://github.com/langchain4j/langchain4j/discussions/786

abbas-gadhia:
In case anyone is stuck with this, just enable the json mode of the LLM for the time being.

mariofusco:
This has been solved (or at least largely mitigated) with the introduction of a [specific output guardrail](https://docs.quarkiverse.io/quarkus-langchain4j/dev/guardrails.html#_rewriting_the_llm_output) in the quarkus-langchain4j extension. I wonder if a similar approach can be ported or reimplemented directly into langchain4j.

langchain4j:
@mariofusco yes, guardrails are expected to be backported from quarkus-lc4j to lc4j at some point, so they should definitely help in some cases.

Although I am not sure that "automatically enabling json mode or json schema" or "use tools for structured outputs" is something that fits into the guardrails concept well.

mr-cloud:
[...]

PR #1248, opened 2024-06-07
add attribute “returnAsFinalAnswer” to Tool annotation ,let llm can use tool funcion return value as the output of the model  directly,to reduce llm request count

BiyoTech:
## Issue

https://github.com/langchain4j/langchain4j/issues/124

## Change

when we use tools ,we send request llm 3 times, e.g:
{
"role": "user",
"content": "What's the weather like in Shanghai today？"
},
{
"role": "assistant",
"content": null,
"function_call": {
"name": "get_current_weather",
"arguments": "{ "location": "shanghai"}"
}
},
{
"role": "function",
"name": "get_current_weather",
"content": "{"temperature": "25", "unit": "摄氏度", "description": "晴朗"}"
},
{
"role": "assistant",
"content": "It's 25 degrees Celsius and sunny in Shanghai today"
},

I think we should give user a choice to add a a “ReturnAsOutput” Annotation on get_current_weather method, let DefaultAiServices just return get_current_weather execution result,. like this:
@tool("get weather information by location")
@ReturnAsOutput
String get_current_weather(@p("location")String location) {
return "It's 25 degrees Celsius and sunny in Shanghai today"
}
and in DefaultAiServices.java ,add code below:
if (method.isAnnotationPresent(ReturnAsOutput.class)) {
String result = toolExecutor.getToolExecuteResult();
chatMemory.add(AiMessage.aiMessage(result));
return result;
}
response = context.chatModel.generate(chatMemory.messages(), context.toolSpecifications);


langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnAsOutput.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Java methods annotated with @Tool and @ReturnAsOutput are considered tools that language model can use,also Directly use tool funcion return value as the output of the model
+ * When using OpenAI models, <a href="https://platform.openai.com/docs/guides/function-calling">function calling</a>
+ * is used under the hood.
+ */
+
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface ReturnAsOutput {
```
langchain4j: What about adding an attribute to existing `@Tool` annotation instead of introducing a new annotation?

`@Tool(returnDirectly = true)`

We need to think about a better name though

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -181,11 +184,20 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
...
                                 );
+
                                 if (context.hasChatMemory()) {
                                     context.chatMemory(memoryId).add(toolExecutionResultMessage);
                                 } else {
                                     messages.add(toolExecutionResultMessage);
                                 }
+                                if (method.isAnnotationPresent(ReturnAsOutput.class)) {
+                                    if (context.hasChatMemory()) {
+                                        context.chatMemory(memoryId).add(AiMessage.aiMessage(toolExecutionResult));
```
langchain4j, line 193: Many LLM providers (such as OpenAI) will fail if `AiMessage` with tool execution request is not followed by a tool retult message

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -267,7 +281,8 @@ private static String getValueOfVariableIt(Parameter[] parameters, Object[] args
             if (!parameter.isAnnotationPresent(dev.langchain4j.service.MemoryId.class)
                     && !parameter.isAnnotationPresent(dev.langchain4j.service.UserMessage.class)
                     && !parameter.isAnnotationPresent(dev.langchain4j.service.UserName.class)
-                    && (!parameter.isAnnotationPresent(dev.langchain4j.service.V.class) || isAnnotatedWithIt(parameter))) {
+                    && (!parameter.isAnnotationPresent(dev.langchain4j.service.V.class) || isAnnotatedWithIt(
+                    parameter))) {
```
langchain4j: please rollback all formatting changes

ghost (commented):
Hi, thank you!

Please see my comments.
Please fill in PR template and implement tests.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnAsOutput.java
BiyoTech: this is a better choice

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -181,11 +184,20 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
                                         toolExecutionRequest,
                                         toolExecutionResult
                                 );
+
                                 if (context.hasChatMemory()) {
```
BiyoTech: @langchain4j make tool method return directly also add “ toolExecutionResultMessage”

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
BiyoTech: thanks for you review , this is maybe not a problem ,make tool method return directly also add “ toolExecutionResultMessage” ,see my comment in code lines

langchain4j: My bad, I misunderstood the change. But then why `AiMessage.aiMessage(toolExecutionResult)` is added to the memory if `toolExecutionResultMessage` has been already added?

BiyoTech: I will give you a example chat history to illustrate this:
```Json 
[
        {
            "role": "user",
            "content": "how about the weather in Shanghai？"
        },
        {
            "role": "assistant",
            "content": null,
            "function_call": {
                "name": "get_current_weather",
                "arguments": "{ \"location\": \"Shanghai\"}"
            }
        },
        {----here is toolExecutionResultMessage,without this ,some LLM providers (such as OpenAI) will fail 
            "role": "function",
            "name": "get_current_weather",
            "content": "It's 25 degrees Celsius and sunny in Shanghai today"
        },
        {----here is toolExecutionResult using Tool function return value(usually,we need to request LLM to generate final answer ,here we use Tool function return value as final answer to reduce LLM requests 1 time)
            "role": "assistant",
            "content": "It's 25 degrees Celsius and sunny in Shanghai today"
        }]
```

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnAsOutput.java
langchain4j: > this is a better choice

Not sure I understand :) To what exactly "this" refers?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
langchain4j: Ok I see.

What is the use case for this? Do you plan to use this with chat memory? Is this for a "data extraction" use case?

BiyoTech: I update the example chat history given to you before(use function content as assistant content).
this case is only for function call, and in function call must use chat memory. 
I think this feature can make function call more reliable and more Fault-tolerant，sometimes we need LLM call funtion and output strict json string， but llm may not ,if we can use tool function return value as output ,We can expect to get a definite output(cuz we can control tool function return object or strict json )

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnAsOutput.java
BiyoTech: > > this is a better choice
> 
> Not sure I understand :) To what exactly "this" refers?

my bad ,I mean adding an attribute to existing @Tool annotation instead of introducing a new annotation is a better choice ,I will try to optimize my code

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
langchain4j: > this case is only for function call

If this change is to use function calls to extract structured data, I would say that this is a misuse of the feature.

> and in function call must use chat memory.

Memory is not needed any more to use tools.

> I think this feature can make function call more reliable and more Fault-tolerant，sometimes we need LLM call funtion and output strict json string， but llm may not ,if we can use tool function return value as output ,We can expect to get a definite output(cuz we can control tool function return object or strict json )

Agree, we have an [issue](https://github.com/langchain4j/langchain4j/issues/859) to address this. Soon one could use function calls to extract structured data. But for that you don't need to define a `@Tool` method. You could just define a POJO as a return type.

juinjonn:
@langchain4j Hello! When will this change be released? Now I also encountered this scene. The result of using Aiservice cannot return the tool directly, resulting in a large request once llm, and the result of making the llm return tools. The guarantee of the llm is not very good Okay, because the result returned by the tool is structured, and the front -end dependencies rely on structured data for rendering

langchain4j:
@jacksonhu which LLM are you using? As I've commented, this seems to be a misuse of tools (`@Tool` in LC4j), so I am not sure it is a good idea to move forward with this. For structured output, we recommend using [this](https://docs.langchain4j.dev/tutorials/ai-services#output-parsing-aka-structured-outputs).

juinjonn:
> @jacksonhu which LLM are you using? As I've commented, this seems to be a misuse of tools (`@Tool` in LC4j), so I am not sure it is a good idea to move forward with this. For structured output, we recommend using [this](https://docs.langchain4j.dev/tutorials/ai-services#output-parsing-aka-structured-outputs).

I am currently using AIService capabilities, but as the user communicates with the agent, the agent will choose the tool he already has to execute. At this time, the expected result of the tool execution is returned as the message to the user (the front end depends on the result of the tool execution because it is structured data). However, AiService currently sends the result and context of the tool execution to llm for interaction. The result is often not the structured data we expect. At this time, the front end cannot render the agreed style.
The above is the scenario I encountered, but I know that the openai model can support the call of multiple tools. If the result of the tool is returned directly, this feature cannot be used. This is a trade-off.

langchain4j:
@juinjonn, a few questions:
1. Could you please provide a few examples (questions and tools) of how this will be used?
2. How do you expect to get the tool result from AI Service method? As a JSON String?
3. How do you expect the memory to behave in this case?

juinjonn:
> @juinjonn, a few questions:
> 
> 1. Could you please provide a few examples (questions and tools) of how this will be used?
> 2. How do you expect to get the tool result from AI Service method? As a JSON String?
> 3. How do you expect the memory to behave in this case?

1. I apologize for not being able to discuss the internal scenarios of the company, but I can provide an analogy. Suppose my agent is an HR assistant responsible for processing and searching resumes. The agent is connected to tools such as Resume Search and Resume Information Extraction. The user is an HR professional. When the HR interacts with the assistant, and the agent identifies the need to use the Resume Search tool, the agent will execute the Resume Search tool. The results from the Resume Search tool will then be returned as a message in the chat. When the frontend detects that this is structured data from the resume search, it will render the results into a card with functions and styles, displaying some information from the resumes. The large model does not need to respond in natural language.
2. Yes, the expectation is to receive a JSON string. The tool is a Java method, and we expect to serialize the return value of this method into JSON.
3. The expectation is to store the results returned by the tool into memory.

For this purpose, I modified DefaultAiServices to meet my requirements. However, this is not a reasonable long-term solution, and I hope the developers can consider such scenarios.

The logic I modified is that I wrote a new AiService and extents from AiServices, only modifying the local code. See the image below:


image
However, it seems that by doing this, the agent’s ability to call multiple tools is compromised. If the requirement is for the agent to directly return the results of the tools, it appears inherently contradictory.

langchain4j:
Hi @juinjonn, thank you for detailed explanation!

1. This makes sense! I am wondering if for this use case if would be beneficial to create a new class or maybe extend `Result` to incorporate both response from the LLM or results of tools. This way you do not have to parse string and detect if this is a tool result or LLM response. Also multiple tools can be supported. WDYT?
3. How the memory will look like after the return from the AI Service method?
This
```
User: Search me a resume for XYZ
AI: \*calls search_resume(XYZ)\*
Tool: { "name": "John Doe", "position": "SWE" ...}
```
Or this:
```
User: Search me a resume for XYZ
AI: \*calls search_resume(XYZ)\*
Tool: { "name": "John Doe", "position": "SWE" ...}
AI: { "name": "John Doe", "position": "SWE" ...}
```

juinjonn:
> Hi @juinjonn, thank you for detailed explanation!
> 
> 1. This makes sense! I am wondering if for this use case if would be beneficial to create a new class or maybe extend `Result` to incorporate both response from the LLM or results of tools. This way you do not have to parse string and detect if this is a tool result or LLM response. Also multiple tools can be supported. WDYT?
> 2. How the memory will look like after the return from the AI Service method?
>    This
> 
> ```
> User: Search me a resume for XYZ
> AI: \*calls search_resume(XYZ)\*
> Tool: { "name": "John Doe", "position": "SWE" ...}
> ```
> 
> Or this:
> 
> ```
> User: Search me a resume for XYZ
> AI: \*calls search_resume(XYZ)\*
> Tool: { "name": "John Doe", "position": "SWE" ...}
> AI: { "name": "John Doe", "position": "SWE" ...}
> ```

I think your solution is excellent and can satisfy this case. However, there is one issue: in such scenarios, the focus is more on the tool’s results. If the tool’s results are used to call the LLM, wouldn’t that waste tokens? If a configuration could be added to allow users to control whether the tool’s results need to be sent to the large model for summary output, it would seem to be a more reasonable solution.However, I’m not sure if it’s reasonable to place this configuration on the @Tool annotation. 

In this case, when the user only needs the Tool’s result, the memory will be:

User: Search me a resume for XYZ
AI: calls search_resume(XYZ)
Tool: { “name”: “John Doe”, “position”: “SWE” …}

If the user needs the LLM to process the tool’s result, the memory will be:

User: Search me a resume for XYZ
AI: calls search_resume(XYZ)
Tool: { “name”: “John Doe”, “position”: “SWE” …}
AI: “His name is John Doe, position is SWE….”

langchain4j:
> I think your solution is excellent and can satisfy this case. However, there is one issue: in such scenarios, the focus is more on the tool’s results. If the tool’s results are used to call the LLM, wouldn’t that waste tokens?

Oh yeah, with this config (return tool result(s) directly to: 1. have structured output and 2. not waste tokens) and expanding `Result` class to include details/results of tool executions might be a good combo for this use case

> If a configuration could be added to allow users to control whether the tool’s results need to be sent to the large model for summary output, it would seem to be a more reasonable solution.However, I’m not sure if it’s reasonable to place this configuration on the @tool annotation.

What other options do you have in mind? Configure it on AI Service level?


> In this case, when the user only needs the Tool’s result, the memory will be:
> 
> User: Search me a resume for XYZ
> AI: calls search_resume(XYZ)
> Tool: { “name”: “John Doe”, “position”: “SWE” …}

For this case we need to check if this will work fine for LLM if next message after Tool comes from the User and not from AI (as it is usually done)

juinjonn:
> > I think your solution is excellent and can satisfy this case. However, there is one issue: in such scenarios, the focus is more on the tool’s results. If the tool’s results are used to call the LLM, wouldn’t that waste tokens?
> 
> Oh yeah, with this config (return tool result(s) directly to: 1. have structured output and 2. not waste tokens) and expanding `Result` class to include details/results of tool executions might be a good combo for this use case
> 
> > If a configuration could be added to allow users to control whether the tool’s results need to be sent to the large model for summary output, it would seem to be a more reasonable solution.However, I’m not sure if it’s reasonable to place this configuration on the @tool annotation.
> 
> What other options do you have in mind? Configure it on AI Service level?
> 
> > In this case, when the user only needs the Tool’s result, the memory will be:
> > User: Search me a resume for XYZ
> > AI: calls search_resume(XYZ)
> > Tool: { “name”: “John Doe”, “position”: “SWE” …}
> 
> For this case we need to check if this will work fine for LLM if next message after Tool comes from the User and not from AI (as it is usually done)

It seems more reasonable to configure this on the @Tool annotation. This way, for a single AiService, its tools can let the user control which tools need to be processed by the LLM. In this case, for a single AiService, some tools will need to be sent to the LLM for summarization, while others will not, making it more flexible.

Secondly, regarding memory, when a tool does not need to be summarized by the LLM, we can use the tool’s result as the agent’s response to the user. Because at this point, the interaction with the LLM has already taken place to select the tool, the interaction between the LLM and the user will be in a question-and-answer format, fulfilling its original design purpose. It seems this will not affect the normal functioning of the LLM.

langchain4j:
> It seems more reasonable to configure this on the @tool annotation. This way, for a single AiService, its tools can let the user control which tools need to be processed by the LLM. In this case, for a single AiService, some tools will need to be sent to the LLM for summarization, while others will not, making it more flexible.

Hmm, what if LLM invokes 2 tools in parallel, one tool's result should be returned directly to the user, another should be processed by the LLM first. What do we do in this case?

juinjonn:
> > 在@tool注释。这样，对于单个 AiService，其工具可以让用户控制哪些工具需要由 LLM 处理。在这种情况下，对于单个 AiService，一些工具需要发送到 LLM 进行汇总，而其他工具则不需要，从而更加灵活。
> 
> 嗯，如果 LLM 并行调用 2 个工具，一个工具的结果应该直接返回给用户，另一个工具的结果应该先由 LLM 处理。在这种情况下我们该怎么办？

As you mentioned, the situation of invoking tools can be wrapped into a result class that contains the tool’s results. Similarly, we can include the results of the LLM. This way, when tools using the LLM and those not using the LLM are invoked simultaneously, the result class can be returned to the caller, allowing the business scenario at the time to decide how to present it to the user. Perhaps they only need the tool’s result, perhaps only the LLM’s result, or maybe they need both.

langchain4j:
@jacksonhu I am not sure I fully understand. Let's imagine LLM decides to call 2 tools, result of the first one should be returned directly, result of the second one should be processed by the LLM first. I see 2 ways to handle this situation:
1. Because one of the tool results should be processed by the LLM first, we send another request to the LLM with results of **_both_** tools (otherwise it will fail) and LLM answers in plain text. Then we return LLM answer (regarding both tools) in plain text and two json tool results in the `Result` object.
2. We remove the request for the first tool from the `AiMessage`/`ChatHistory` and send only the request/result of the second tool back to the LLM and  it answers in plain text. Then we return LLM answer (regarding second tool) in plain text and two json tool results in the `Result` object.

Since this is a corner case, I guess it is fine to use the simpler (first) option and think about optimizations later, WDYT?

juinjonn:
> @jacksonhu I am not sure I fully understand. Let's imagine LLM decides to call 2 tools, result of the first one should be returned directly, result of the second one should be processed by the LLM first. I see 2 ways to handle this situation:
> 
> 1. Because one of the tool results should be processed by the LLM first, we send another request to the LLM with results of **_both_** tools (otherwise it will fail) and LLM answers in plain text. Then we return LLM answer (regarding both tools) in plain text and two json tool results in the `Result` object.
> 2. We remove the request for the first tool from the `AiMessage`/`ChatHistory` and send only the request/result of the second tool back to the LLM and  it answers in plain text. Then we return LLM answer (regarding second tool) in plain text and two json tool results in the `Result` object.
> 
> Since this is a corner case, I guess it is fine to use the simpler (first) option and think about optimizations later, WDYT?

Alright, I think your idea is okay. Currently, aiservice does not support responding with structured and unstructured data when invoking tools (since after invoking a tool, it currently needs to go through the LLM for summarization). This is a hindrance to business development. Therefore, we have made modifications locally. If you have a definitive plan, you can let me know, and I will submit a PR based on my availability, as we are really hoping for this capability rather than resorting to modifications. This approach is ultimately just a temporary solution. I hope you can understand.

langchain4j:
@juinjonn please feel free to open the PR! Or perharps @jacksonhu could adjust this PR?

Do we agree on the following solution?
- Add `returnDirectly` attribute to the `@Tool` annotation
- ~~Add new field `List<ToolExecution> toolExecutions` to the `Result<T>` class:~~ (has been implemented already)
```java
public class ToolExecution {
    ToolExecutionRequest toolExecutionRequest;
    Object toolExecutionResult;
}
```
- When **_all_** `ToolExecutionRequest`s in `AiMessage` are related to methods annotated with `@Tool(returnDirectly = true)`, then tools should be executed and `Result<T>` containing raw tool results should be returned directly to the user, without additional LLM request. `Result.content()` is `null` in this case.
- When **_some_** of the `ToolExecutionRequest`s in `AiMessage` are not related to methods annotated with `@Tool(returnDirectly = true)`, the additional LLM requests is done, afterwards `Result<T>` containing LLM summarized response (as `Result.content()`) **_and raw tool results_** is returned.
- `returnDirectly = true` is compatible only with `Result<String>` return type

WDYT?

juinjonn:
> @juinjonn please feel free to open the PR! Or perharps @jacksonhu could adjust this PR?
> 
> Do we agree on the following solution?
> 
> * Add `returnDirectly` attribute to the `@Tool` annotation
> * Add new field `List<ToolExecution> toolExecutions` to the `Result<T>` class:
> 
> ```java
> public class ToolExecution {
>     ToolExecutionRequest toolExecutionRequest;
>     Object toolExecutionResult;
> }
> ```
> 
> * When **_all_** `ToolExecutionRequest`s in `AiMessage` are related to methods annotated with `@Tool(returnDirectly = true)`, then tools should be executed and `Result<T>` containing raw tool results should be returned directly to the user, without additional LLM request. `Result.content()` is `null` in this case.
> * When **_some_** of the `ToolExecutionRequest`s in `AiMessage` are not related to methods annotated with `@Tool(returnDirectly = true)`, the additional LLM requests is done, afterwards `Result<T>` containing LLM summarized response (as `Result.content()`) **_and raw tool results_** is returned.
> * `returnDirectly = true` is compatible only with `Result<String>` return type
> 
> WDYT?

ok  I think it is very nice，I will submit PR according to my time

langchain4j:
Hi @juinjonn, do you still plan to submit a PR? 🙏

jrsperry:
I’m very interested in this work. Happy to take a look and see if I can implement the solution. Is the mentioned PR the best place to start?

langchain4j:
Hi @jrsperry! We seem to have agreed on the desired behaviour in [this](https://github.com/langchain4j/langchain4j/pull/1248#issuecomment-2219830528) comment, could you please check it and see if it makes sense or if there is a better way?

Any help with this feature is highly appreciated! 🙏

jrsperry:
so to clarify, Result content can be nullable (currently is enforced to not be null). If ALL ToolExecutionRequests are related to methods which are "returnDirectly" then we will return a Result with the ToolExecutions and null content. 

If SOME ToolExecutionRequests are related to methods which are "returnDirectly" then we will pass the ToolExecution back to the llm, and then return the result from the llm AND the direct tool results in the result.content?  In this scenario, I'm not sure if just appending the directToolResults to the result.content makes sense. Especially where we are not adding the toolexecution to the content in the first scenario. Thoughts?

jrsperry:
Opened new PR here: https://github.com/langchain4j/langchain4j/pull/2291

mariofusco:
@langchain4j Since there is a second implementation of this feature under development in https://github.com/langchain4j/langchain4j/pull/2291 I believe that this pull request can be closed.

langchain4j:
Closing in favour of https://github.com/langchain4j/langchain4j/pull/2291

zhhaojie:
after i saw this issue , i save my job.

zhhaojie:
image

tested, works very nice

Issue #1080, opened 2024-05-10
[FEATURE] DX: Add dimension() in EmbeddingModel


Martin7-1:
**Is your feature request related to a problem? Please describe.**

As describe in [Redis embedding store spring boot starter](https://github.com/langchain4j/langchain4j-spring/pull/2). Embedding model should have `dimension()` method to enable `EmbeddingStore` to have a default value.

**Describe the solution you'd like**

In `EmbeddingModel`, will add method like below:

```java
    /**
     * Get EmbeddingModel's dimension
     *
     * @return dimension
     */
    default int dimension() {
        return embed(TextSegment.from("test")).content().dimension();
    }
```

**Describe alternatives you've considered**

**Additional context**

langchain4j:
HI @Martin7-1 thank you, this is long due!

A few thoughts:
- while it is great to have a default implementation that will work for all implementations of this interface, I think we should go one step further and have an overriden impl of this method for popular `EmbeddingModel` implementations. In many cases dimensions are known up front, so we can skip embedding of "test". When dimension is not known, this call can be cached.
- we need to decide finally whether to name it `dimensions` or `dimension` (singular). Right now we have both, it is inconsistent.

Martin7-1:
> HI @Martin7-1 thank you, this is long due!
> 
> A few thoughts:
> 
> * while it is great to have a default implementation that will work for all implementations of this interface, I think we should go one step further and have an overriden impl of this method for popular `EmbeddingModel` implementations. In many cases dimensions are known up front, so we can skip embedding of "test". When dimension is not known, this call can be cached.
> * we need to decide finally whether to name it `dimensions` or `dimension` (singular). Right now we have both, it is inconsistent.

* I think it is a very good idea to override the commonly used EmbeddingModel, and I will try to collect the dimensions of EmbeddingModel from common model providers and implement it. **However, for model providers that provide multiple `EmbeddingModel`s, we need to modify the implementation when the model is added or changed.** Maybe we need a way to configure it (in file or other format maybe?) to reduce the number of code changes. WDYT?
* Absolutely it's important to have consistent naming strategy. I prefer `dimension` because it tends to return just a single number.

langchain4j:
> However, for model providers that provide multiple EmbeddingModels, we need to modify the implementation when the model is added or changed. Maybe we need a way to configure it (in file or other format maybe?) to reduce the number of code changes. WDYT?

[...]

PR #1250, opened 2024-06-08
[Feature] Add dimension() in EmbeddingModel

Martin7-1:
## Issue

#1080 

## Change

Add `dimension()` method in `EmbeddingModel`. Using an `AbstractEmbeddingModel` to implement `EmbeddingModel` and a common `dimension()` method like below:

```java
    @Override
    public int dimension() {
        if (dimension != null) {
            return dimension;
        }

        return dimensionMap().compute(modelName(), (key, value) -> {
            this.dimension = Optional.ofNullable(value).orElse(embed("test").content().dimension());
            return dimension;
        });
    }
```

Now, existing `{NameOfIntegration}EmbeddingModel` just need to override `dimensionMap()` method to contains known embedding model dimension and `modelName()` method to get embedding model name. 

## General checklist


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
```diff
@@ -80,10 +80,21 @@ public OpenAiEmbeddingModel(String baseUrl,
...
+    }
+
+    @Override
     public String modelName() {
         return modelName;
     }
 
+    @Override
+    public int dimension() {
+        return dimensions;
```
langchain4j: `dimensions` is an optional parameters, it can be `null`

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
```diff
@@ -0,0 +1,41 @@
...
+    protected abstract String modelName();
+
+    @Override
+    public int dimension() {
+        if (dimension != null) {
+            return dimension;
+        }
+
+        return dimensionMap().compute(modelName(), (key, value) -> {
+            this.dimension = Optional.ofNullable(value).orElse(embed("test").content().dimension());
```
langchain4j:
```suggestion
            this.dimension = Optional.ofNullable(value).orElseGet(() -> embed("test").content().dimension());
```
Otherwise it will embed every time

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    protected abstract String modelName();
+
+    @Override
+    public int dimension() {
+        if (dimension != null) {
+            return dimension;
+        }
+
+        return dimensionMap().compute(modelName(), (key, value) -> {
```
langchain4j:
```suggestion
        return dimensionMap().compute(modelName(), (modelName, dimension) -> {
```
for better clarity

langchain4j: Is the map supposed to be cached?

langchain4j: I am having a hard time to understand what this piece of code is doing, could you please rewrite it with simpler if/else conditions?

ghost (commented):
@Martin7-1 thank you!

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
Martin7-1: Now, I remove `dimensions` in `OpenAiEmbeddingModel` because I think `dimension` in `AbstractEmbeddingModel` and `dimenions` in `OpenAiEmbeddingModel` will make user confused. So I delete it and use `dimension` in `AbstractEmbeddingModel` as an alternative way

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
Martin7-1: Sure. Sorry for makeing you confused. Already rewrite it!

Martin7-1: I'm not sure I know what you mean. This map's purpose is to store known model's name and it's dimension.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
langchain4j: No, please don't remove `dimensions` from `OpenAiEmbeddingModel`, it is needed to **_control_** the number of dimensions embedding model returns.

langchain4j: I meant please don't rename `dimensions` into `dimension` in `OpenAiEmbeddingModel`.

Martin7-1: > I meant please don't rename `dimensions` into `dimension` in `OpenAiEmbeddingModel`.

OK. I got it.

Martin7-1: > I meant please don't rename `dimensions` into `dimension` in `OpenAiEmbeddingModel`.

done.

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
```diff
@@ -44,4 +44,11 @@ default Response<Embedding> embed(TextSegment textSegment) {
      * @return the embeddings.
      */
     Response<List<Embedding>> embedAll(List<TextSegment> textSegments);
+
+    /**
+     * Get embedding model's dimension
```
langchain4j:
```suggestion
     * Returns the dimension of the {@link Embedding} produced by this embedding model.
```

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
```diff
@@ -44,4 +44,11 @@ default Response<Embedding> embed(TextSegment textSegment) {
...
+     *
+     * @return dimension of the embedding model
```
langchain4j:
```suggestion
     * @return dimension of the embedding
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
```diff
@@ -2,24 +2,33 @@
...
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import static dev.langchain4j.internal.Utils.getOrDefault;
+
 /**
  * Bedrock Amazon Titan embedding model
  */
-@SuperBuilder
```
langchain4j: Please make sure changes in this class are not breaking

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
```diff
@@ -0,0 +1,44 @@
...
+    protected abstract String modelName();
+
+    @Override
+    public int dimension() {
+        if (dimension != null) {
+            return dimension;
+        }
+
+        // get known model's dimension first, otherwise embed "test" to get dimension
+        if (dimensionMap().containsKey(modelName())) {
```
langchain4j: minor: `dimensionMap()` can be stored in a local variable to avoid creating this map again

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
```diff
@@ -44,4 +44,11 @@ default Response<Embedding> embed(TextSegment textSegment) {
...
+     */
+    int dimension();
```
langchain4j: This is a breaking change: consider that there are implementations of this interface that are not located in this repository.

What about making this method default?
```java
    default int dimension() {
        return embed("test").content().dimension();
    }
```
In theory this should work for all implementations of this interface.

For specific cases where we know the dimension up-front and want to optimize this method, the implementation class can override this method with the logic similar to what you have implemented in this PR.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/internal/AbstractBedrockEmbeddingModel.java
```diff
@@ -25,22 +23,27 @@
 import java.util.stream.Collectors;
 
 import static dev.langchain4j.internal.RetryUtils.withRetry;
+import static dev.langchain4j.internal.Utils.getOrDefault;
 
 /**
  * Abstract bedrock embedding model
  */
-@SuperBuilder
```
langchain4j: same here

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
```diff
@@ -0,0 +1,44 @@
+package dev.langchain4j.model.embedding;
+
+import java.util.Map;
+
+/**
+ * An Embedding Model which contains
+ */
+public abstract class AbstractEmbeddingModel implements EmbeddingModel {
```
langchain4j:
```suggestion
public abstract class DimensionAwareEmbeddingModel implements EmbeddingModel {
```
WDYT?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
```diff
@@ -80,6 +80,12 @@ public OpenAiEmbeddingModel(String baseUrl,
         this.tokenizer = getOrDefault(tokenizer, OpenAiTokenizer::new);
     }
 
+    @Override
+    protected Map<String, Integer> dimensionMap() {
+        return OpenAiEmbeddingModelName.embeddingModelDimensionMap();
```
langchain4j: If `dimensions` parameter is specified in the constructor, `OpenAiEmbeddingModel.dimension()` should return the specified value

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
```diff
@@ -0,0 +1,44 @@
...
+     * embedding model dimension
+     */
+    protected Integer dimension;
+
+    /**
+     * A map contains known model's name and its dimension
+     *
+     * @return A map, key is the common represented model name, value is its dimension
+     */
+    protected abstract Map<String, Integer> dimensionMap();
```
langchain4j: I am not sure there is a need to expose 2 methods (`dimensionMap` and `modelName`).
Can't implementation of this class return dimension value directly, without a need to expose a map?

langchain4j-cohere/src/main/java/dev/langchain4j/model/cohere/CohereEmbeddingModel.java
```diff
@@ -83,4 +86,14 @@ private static TokenUsage getTokenUsage(EmbedResponse response) {
         }
         return null;
     }
+
+    @Override
+    protected Map<String, Integer> dimensionMap() {
+        return new HashMap<>();
```
langchain4j: can this be a default implementation in `AbstractEmbeddingModel `? So that there is no need to duplicate this code in many places

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
Martin7-1: I think this is great and makes the purpose of this class more clearer. Then maybe not all implmenetations should implement this class, some of them may just implement `EmbeddingModel` (This also avoid breaking changes in `langchain4j-bedrock` module and `langchain4j-workers-ai` module).

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
Martin7-1: OK. It's a good idea to make it default!

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
Martin7-1: done.

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
Martin7-1: done.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
Martin7-1: done. Revert it to implement `EmbeddingModel` directly

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
Martin7-1: Good idea. Now the `DimensionAwareEmbeddingModel` has `dimensionMap` attribute to cache.

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
Martin7-1: done.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/internal/AbstractBedrockEmbeddingModel.java
Martin7-1: done. See above.

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
Martin7-1: done.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
Martin7-1: done.

langchain4j-cohere/src/main/java/dev/langchain4j/model/cohere/CohereEmbeddingModel.java
Martin7-1: done.

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
Martin7-1: hmmmmm...It's a good question. The difference between `EmbeddingModel` and `DimensionAwareEmbeddingModel` is that the latter cache some known model's dimension and make it more convenient to get embedding's dimension. The purpose to expose `dimensionMap` and `modelName` is just like `EmbeddingStoreWithoutMetadataIT`'s `embeddingModel()` and `embeddingStore()`. I think it makes `dimension()` more standardized and uniform without having to worry about differences between `modelName` and `dimensionMap` in different implementations

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    protected abstract String modelName();
+
+    @Override
+    public int dimension() {
+        if (dimension != null) {
+            return dimension;
+        }
+
+        // get known model's dimension first, otherwise embed "test" to get dimension
```
langchain4j:
```suggestion
```
Not very helpful comment

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    protected Map<String, Integer> dimensionMap = new HashMap<>();
+
+    /**
+     * A map contains known model's name and its embedding's dimension
+     *
+     * @return A map, key is the common represented model name, value is its embedding's dimension
+     */
+    protected Map<String, Integer> dimensionMap() {
+        return dimensionMap;
```
langchain4j:
```suggestion
        return emptyMap();
```

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
```diff
@@ -0,0 +1,51 @@
...
+public abstract class DimensionAwareEmbeddingModel implements EmbeddingModel {
+
+    /**
+     * dimension of embedding
+     */
+    protected Integer dimension;
+    /**
+     * dimension map of known embedding model's name and its embedding's dimension
+     */
+    protected Map<String, Integer> dimensionMap = new HashMap<>();
```
langchain4j:
```suggestion
```
It seems that there is no need to keep/cache this map

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java
langchain4j: > The purpose to expose dimensionMap and modelName is just like EmbeddingStoreWithoutMetadataIT's embeddingModel() and embeddingStore(). I think it makes dimension() more standardized and uniform without having to worry about differences between modelName and dimensionMap in different implementations

Not sure I understand, can you please elaborate?

A couple of minor downsides that I see:
- `dimensionMap()` exposes much more information (dimensions for all models) than is required for the task. At some point someone will have a temptation to misuse it.
- having `Map` allows only matching models by exact name, some other implementation-specific operations like "modelName.startsWith("...")" are not possible

Martin7-1: The standard step to get dimension is:

1. check cached dimension, if exist, return it.
2. check known model's dimension, if there is any match, cache and return it.
3. manually embed "test" and cache it.

There are two differences in these steps:

1. different implementations may have different ways of getting model name, e.g. some implementations get it via the modelName attribute, some names may be, as you say, a prefix or something like that.
2. implementations may have different ways of storing known dimensional models, perhaps via enum, perhaps via map, etc., depending on the implementation.

But for `DimensionAwareEmbeddingModel`, all it has to do is to check if there is a corresponding cached dimension based on the modelName, and if there is one, get it directly, if not, embed it manually and get it. So I think their exposure is necessary.

> dimensionMap() exposes much more information (dimensions for all models) than is required for the task. At some point someone will have a temptation to misuse it.

Do you have any good ideas to avoid it? I currently have it set to `protected`, so it is invisible to users. Maybe we should rename it to `getKnownDimension()` to avoid misuse?

> having Map allows only matching models by exact name, some other implementation-specific operations like "modelName.startsWith("...")" are not possible

That's a real downside...Maybe we need a `Function` to calculate the dimension, like `getKnownDimension()`. This allows implementations to implement it more freely.

So the solution may like below:

```java
public int dimension() {
    if (dimension != null) {
        return dimension;
    }
    Integer knownDimension = getKnownDimension();  // different implementations have different ways to get known dimension
    dimension = Optional.ofNullable(knownDimension).orElseGet(() -> embed("test").content().dimension());
    return dimension;
}

File: langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java (hunk 14 above)
Comment (Martin7-1) [reply langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java]: done.

File: langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java (hunk 12 above)
Comment (Martin7-1) [reply langchain4j-core/src/main/java/dev/langchain4j/model/embedding/AbstractEmbeddingModel.java]:
```java
public int dimension() {
    if (dimension != null) {
        return dimension;
    }
    Integer knownDimension = getKnownDimension();  // different implementations have different ways to get known dimension
    dimension = Optional.ofNullable(knownDimension).orElseGet(() -> embed("test").content().dimension());
    return dimension;
}
```

Maybe this will make implementations to get known dimension more freely

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModelName.java
```diff
@@ -1,5 +1,8 @@
 package dev.langchain4j.model.openai;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public enum OpenAiEmbeddingModelName {
 
     TEXT_EMBEDDING_3_SMALL("text-embedding-3-small"),
```
langchain4j:
```suggestion
    TEXT_EMBEDDING_3_SMALL("text-embedding-3-small", 1536),
```
WDYT about putting it here instead of map?

langchain4j-workers-ai/src/main/java/dev/langchain4j/model/workersai/client/AbstractWorkersAIModel.java
```diff
@@ -30,22 +30,19 @@ public abstract class AbstractWorkersAIModel {
...
-     * @param modelName
-     *      model name.
-     * @param apiToken
-     *      api apiToken from .
+     * @param accountId account identifier.
+     * @param modelName model name.
+     * @param apiToken  api apiToken from .
      */
-    public AbstractWorkersAIModel(String accountId, String modelName, String apiToken) {
+    protected AbstractWorkersAIModel(String accountId, String modelName, String apiToken) {
```
langchain4j: why this change?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiEmbeddingModelName.java
```diff
@@ -1,5 +1,8 @@
 package dev.langchain4j.model.azure;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public enum AzureOpenAiEmbeddingModelName {
 
     TEXT_EMBEDDING_3_SMALL("text-embedding-3-small", "text-embedding-3-small"), // alias for the latest text-embedding-3-small model
```
langchain4j:
```suggestion
    TEXT_EMBEDDING_3_SMALL("text-embedding-3-small", "text-embedding-3-small", 1536), // alias for the latest text-embedding-3-small model
```
WDYT about putting it here instead of map?

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
```diff
@@ -0,0 +1,36 @@
...
+     * dimension of embedding
+     */
+    protected Integer dimension;
+
+    /**
+     * Returns the dimension of known {@link Embedding} produced by this embedding model. if it's unknown, return null
+     *
+     * @return the dimension of known {@link Embedding}, null if unknown.
+     */
+    protected Integer getKnownDimension() {
```
langchain4j:
```suggestion
    protected Integer knownDimension() {
```

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public abstract class DimensionAwareEmbeddingModel implements EmbeddingModel {
+
+    /**
+     * dimension of embedding
+     */
+    protected Integer dimension;
+
+    /**
+     * Returns the dimension of known {@link Embedding} produced by this embedding model. if it's unknown, return null
```
langchain4j:
```suggestion
     * When known (e.g., can be derived from the model name), returns the dimension of the {@link Embedding} produced by this embedding model. Otherwise, it returns {@code null}.
```

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
```diff
@@ -0,0 +1,36 @@
...
+     *
+     * @return the dimension of known {@link Embedding}, null if unknown.
```
langchain4j:
```suggestion
     * @return the known dimension of the {@link Embedding}, or {@code null} if unknown.
```

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
langchain4j, line 1: You've removed gecko mappings in the previous commit, probably by accident?

ghost (commented):
@Martin7-1 I like this new solution much more, thank you!

Just a few minor things and we can merge!

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModelName.java
Martin7-1: Good idea. But there is one thing to consider:

In `OpenAiEmbeddingModel`, String is used to represent modelName instead of OpenAiEmbeddingModelName, so if we want to add dimension to an enum, maybe we need a map in `OpenAiEmbeddingModelName` like below:

```java
    private static final Map<String, Integer> KNOWN_DIMENSION = new HashMap<>(OpenAiEmbeddingModelName.values().length);
    
    static {
        for (OpenAiEmbeddingModelName embeddingModelName : OpenAiEmbeddingModelName.values()) {
            KNOWN_DIMENSION.put(embeddingModelName.toString(), embeddingModelName.dimension());
        }
    }

    public static Integer knownDimension(String modelName) {
        return KNOWN_DIMENSION.get(modelName);
    }
```

WDYT?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
Martin7-1: Sorry for that. Now I implement it like `OpenAiEmbeddingModelName`

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
Martin7-1: done.

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
Martin7-1: done.

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/DimensionAwareEmbeddingModel.java
Martin7-1: done.

langchain4j-workers-ai/src/main/java/dev/langchain4j/model/workersai/client/AbstractWorkersAIModel.java
Martin7-1: revert. Maybe it's problem of IntelliJ.

Martin7-1:
The CI failed maybe because the `main` branch's commit. @langchain4j

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiEmbeddingModel.java
langchain4j: Thanks!

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModelName.java
langchain4j: Sounds good!

ghost (approved):
@Martin7-1 good job, thank you!

langchain4j:
@Martin7-1 could you please also update embedding models in https://github.com/langchain4j/langchain4j-embeddings ?

Martin7-1:
> @Martin7-1 could you please also update embedding models in https://github.com/langchain4j/langchain4j-embeddings ?

Sure!

PR #1258, opened 2024-06-10
Adding Tinylog to the BOM

agoncal:
I see myself having to define the Tinylog version in all my projects. And that's also the case for the LangChain4j samples.

What about declaring Tinylog in the BOM as a dependency management ?

### DIFF
--- a/langchain4j-bom/pom.xml
+++ b/langchain4j-bom/pom.xml
@@ -14,8 +14,7 @@
     <packaging>pom</packaging>
 
     <name>LangChain4j :: BOM</name>
-    <description>Bill of Materials POM for getting full, complete set of compatible versions of LangChain4j modules
-    </description>
+    <description>Bill of Materials POM for getting full, complete set of compatible versions of LangChain4j modules</description>
 
     <dependencyManagement>
         <dependencies>
--- a/langchain4j-bom/pom.xml
+++ b/langchain4j-bom/pom.xml
@@ -17,10 +17,6 @@
     <description>Bill of Materials POM for getting full, complete set of compatible versions of LangChain4j modules
     </description>
 
-    <properties>
-        <tinylog.version>2.6.2</tinylog.version>
-    </properties>
-
     <dependencyManagement>
         <dependencies>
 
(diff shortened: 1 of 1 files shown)


langchain4j:
@agoncal hmm, where do you see this problem (need to define versions) exactly?

agoncal:
Well, each time you need to setup the log level (using the `tinylog.properties` file), you need to depend on Tinylog and look for the current version to use. That's also what's done in most LangChain4j samples (eg. https://github.com/langchain4j/langchain4j-examples/blob/main/azure-open-ai-examples/pom.xml#L31). That would allow to depend on Tinylog and inheriting from the current version (which also could evolve):

```xml
 <dependency>
            <groupId>org.tinylog</groupId>
            <artifactId>tinylog-impl</artifactId>
</dependency>

// Instead of having to set the version explicitelly
 <dependency>
            <groupId>org.tinylog</groupId>
            <artifactId>tinylog-impl</artifactId>
            <version>2.6.2</version>
</dependency>
```

langchain4j:
TBH I don't feel comfortable including it in the bom. We use it only for testing, and even then I guess we should migrate to logback or log4j.

In the main repo, tinylog version is defined in one place. We could do the same in examples, define the version in parent.
WDYT?

agoncal:
Ok, I'm closing this issue then.

@langchain4j what about activating the "Issues" in the https://github.com/langchain4j/langchain4j-examples repository ? I could create an issue to add Tinylog to the parent POM.

langchain4j:
> what about activating the "Issues" in the https://github.com/langchain4j/langchain4j-examples repository

I have closed issues in all repos except the main one because it was hard to track all of them. It is way easier to have issues in one place. Unless there is a way to aggregate them across multiple repos?

agoncal:
@langchain4j Yes you can. You can create a project to have a single with of your repos/issues/PRs

Screenshot 2024-06-11 at 10 25 05

Then, when you create an issue, you assign the project:

Screenshot 2024-06-11 at 10 26 38

langchain4j:
Oh, good to know, thank you! I hope the project can be assigned automatically to every issue or PR?
I will check these projects feature into more details next week.

agoncal:
Yes, I think you can add the project and labels automatically. I don't know how to do it (yet) but I've already seen this

PR #1260, opened 2024-06-10
Support ToolExecutionResultMessage on Qwen Models

jiangsier-xyz:
The current implementation supports the tool calls plan, but does not include ToolExecutionResultMessage. This makes it impossible to feed back the tool execution results to the model. This patch includes support for ToolExecutionResultMessage.

## General checklist


ghost (approved):
@jiangsier-xyz thank you!

Issue #837, opened 2024-03-27
[FEATURE] ChatLanguageModel: allow setting parameters dynamically


lbjfish:
**Spring Boot , i want to dynamically call params，but it is limited too die:**
```
@Bean
    public ChatLanguageModel myCustomChatModel() {
        return OpenAiChatModel.builder()
                .apiKey("dynamically ")
                .stop(dynamically )
                .topP(dynamically )
                .temperature(dynamically )
                .modelName("dynamically ").build();
    }
```

Kugaaa:
I am also thinking about it

1. Usually, different businesses will have different parameters, maybe should create different `AiService`
2. Supply method level parameters

langchain4j:
@lbjfish @Kugaaa could you please describe your use cases? Why do you want to change parameters on the fly?

@Kugaaa, yes, the idea was to have a different `AiService`/`ChatLanguageModel` instances per use case/customer/etc, SRP

langchain4j:
Related: https://github.com/langchain4j/langchain4j/discussions/1190#discussioncomment-9591461

fmatar:
A much needed feature for our use case as well. 

@langchain4j our users have the capability of using different models (example gpt-3.5, gpt-4, gpt-4o with different API keys.

geoand:
I am not still not clear why changing parameters makes more sense and not just creating a new `ChatLanguageModel` for example.

langchain4j:
@geoand one use case I have in mind: automatically enabling json mode for structured data extraction. If `ChatLanguageModel` instance does not have it enabled and AI Service method returns a POJO, we could automatically enable json mode to increase the  probability of correct output.

geoand:
That does indeed make sense.

langchain4j:
Related: https://github.com/langchain4j/langchain4j/discussions/2123

langchain4j:
Related: https://github.com/langchain4j/langchain4j/issues/2218

PR #1261, opened 2024-06-10
[Draft] ChatLanguageModel: allow setting parameters dynamically

ghost:
## Issue
https://github.com/langchain4j/langchain4j/issues/837

## Change
Drafted flexible `(Streaming)ChatLanguageModel` API that allows specifying common model parameters (e.g. model name, temparature, etc) for each request.
Comparison table: https://docs.google.com/spreadsheets/d/1hOsbbAz9-qXCrnq_TXW-Q_AIAlssPLz9/edit?gid=1554605342#gid=1554605342

## General checklist


fmatar:
Would this also support passing of apiKeys / baseUrl / organizationId?

langchain4j:
@fmatar I will check if/how that can be supported

fmatar:
@langchain4j one of the approaches I implemented was to use the raw rest client for that purpose. 

Then I pass all my chatCompletion request along with the security keys to the request and execute. 

One major blocking factor was the passing of the model objects by ai4j since they're not serializable and I ended up writing the required serialized for them. 

Happy to assist in anyway since this is a major roadblock we'd like to resolve

cv-msan:
> @fmatar我将检查如何支持那件事

We now have tables in MySQL specifically designed to maintain various model parameters, and we want users to be able to dynamically switch between different models at any time. When is this dynamic switching feature expected to be supported?

langchain4j:
@cv-msan which parameters should be dynamic to support your use case? Hopefully this will be included in the next release, in the mid of december

cv-msan:
> @cv-msan哪些参数应该动态以支持您的用例？希望这将在下一个月中旬的下一个版本中包含
 apiKey、baseUrl、modelName For now, there are three necessary ones

cv-msan:
You can check if there are any other configurable parameters in other models for the rest

cv-msan:
From a business perspective, there is a model configuration list where you can set the default model to use for Rag conversations and switch between multiple models. Currently, I have implemented it myself, but I feel that the writing is not very good. If your official team can implement dynamic model switching instead of creating an immutable model in the form of Spring beans, it would be quite cool
image

langchain4j:
@cv-msan I guess it is feasible, `modelName` is already dynamic in this PR. Regarding `apiKey` and `baseUrl`: we need to think a bit more about it

cv-msan:
Okay, I'm really looking forward to your updates

langchain4j:
Superceded by https://github.com/langchain4j/langchain4j/pull/1989

Issue #1270, opened 2024-06-11
[FEATURE] Gemini — Add a logging request / response option in the builder


glaforge:
It's sometimes difficult to know what's going on with the exchanges between an application and a model.

This feature request is about adding a `logRequestAndResponse()` method to `VertexAiGeminiChatModel` and `VertexAIGeminiStreamingChatModel`.

langchain4j:
Would be nice to have separate `logRequests()` and `logResponses()`, if possible.

glaforge:
I followed some models that implemented `logRequestsAndResponse()` like Azure OpenAI. 
Different models seem to do different things here: I see `logRequests()`, `logRequests(Boolean)`, `logRequestsAndResponses(Boolean)`, or sometimes with primitive `boolean`.
Maybe at some point we should have some interface that models should implement?
In my case, I also log system instructions separately, as they are defined on the generative model configuration.
So should I have three methods: `logRequests()`, `logResponses()`, and `logSystemInstructions()`? Or I guess I would always log the system instructions when log requests is enabled.
Do you prefer `logRequests()` (no-args) or `logRequests(Boolean)`?

langchain4j:
AFAIK only Azure models have `logRequestsAndResponses()` because they can't separate them.
All other `*Model` classes should have `logRequests(Boolean)`.
`logRequests()` or `logRequests(boolean)` can be used in clients for simplicity.

Please use `logRequests(Boolean)` and `logResponses(Boolean)` if possible. If not, then `logRequestsAndResponses()`.
System instructions should ideally be logged with the request.

glaforge:
Thanks for the advice and suggestions.
I'll update my (upcoming) PR accordingly.

langchain4j:
Implemented in https://github.com/langchain4j/langchain4j/pull/1278

Issue #1269, opened 2024-06-11
[FEATURE] Gemini — Support Google Search and Vertex AI Search grounding


glaforge:
The Gemini model has the ability to search 
* into Google Search to ground its responses, or 
* in the Google Cloud Vertex AI Search (to search on custom/personal/company documents)

See this [page](https://cloud.google.com/vertex-ai/generative-ai/docs/multimodal/ground-gemini) for more information.

langchain4j:
Implemented in https://github.com/langchain4j/langchain4j/pull/1278

Issue #1208, opened 2024-05-30
[FEATURE] Gemini: JSON output mode


langchain4j:
Gemini now supports JSON output, more details: https://ai.google.dev/gemini-api/docs/api-overview#json

langchain4j:
cc @glaforge will this be available in Gemini SDK?

glaforge:
Yes, it should be available soon.
The function calling mode is also coming soon.

langchain4j:
Implemented in https://github.com/langchain4j/langchain4j/pull/1278

Issue #1182, opened 2024-05-27
[FEATURE] Add harm block threshold param to Vertex AI Gemini Chat Model


kstanisz:
Sometimes using Google Vertex Ai Gemini model there is a failure:

`java.lang.IllegalArgumentException: The response is blocked due to safety reason.
	at com.google.cloud.vertexai.generativeai.ResponseHandler.checkFinishReason(ResponseHandler.java:187) ~[google-cloud-vertexai-1.3.0.jar:1.3.0]
	at com.google.cloud.vertexai.generativeai.ResponseHandler.getContent(ResponseHandler.java:85) ~[google-cloud-vertexai-1.3.0.jar:1.3.0]
	at dev.langchain4j.model.vertexai.VertexAiGeminiChatModel.generate(VertexAiGeminiChatModel.java:140) ~[langchain4j-vertex-ai-gemini-0.31.0.jar:na]
	at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:189) ~[langchain4j-0.31.0.jar:na]
	at jdk.proxy2/jdk.proxy2.$Proxy98.sendMessage(Unknown Source) ~[na:na]`

There should be an option to configure safety settings for Gemini model.

In VertexAiGeminiChatModel.class we create GenerativeModel which has a safetySettings param to configure.

langchain4j:
cc @glaforge

langchain4j:
Implemented in https://github.com/langchain4j/langchain4j/pull/1278

PR #1278, opened 2024-06-13
Gemini updates

glaforge:
Gemini updates:
* update to latest Java SDK version
* #1269
* #1270
* #1208
* #1399 
* #1397 
* #1182
* #828
* fixes parallel function calling which wasn't working properly in the previous release
* refactored a bit the `generate()` method to have a single entry point and less duplication

Vertex AI embedding model:
* add new task types (question answering and fact verification)

Imagen image model:
* support more configuration parameters
* #1367


langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/PartsMapper.java
```diff
@@ -86,7 +93,12 @@ static List<Part> map(ChatMessage message) {
                 try {
                     JsonFormat.parser().merge(functionResponseTextAsMap, structBuilder);
                 } catch (InvalidProtocolBufferException e2) {
-                    throw new RuntimeException(e);
+                    String functionResponseTextWithQuotesAsMap = "{\"result\":" + quoted(functionResponseText) + "}";
```
langchain4j, line 96: Can we predict if response text should be in a map and/or quoted? I would avoid using exceptions to control the flow

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
```diff
@@ -73,8 +93,28 @@ public VertexAiGeminiChatModel(String project,
         if (topP != null) {
             generationConfigBuilder.setTopP(topP);
         }
+        if (responseMimeType != null) {
+            generationConfigBuilder.setResponseMimeType(responseMimeType);
+        }
         this.generationConfig = generationConfigBuilder.build();
 
+        if (safetySettings != null) {
+            this.safetySettings = safetySettings;
```
langchain4j: minor:
```suggestion
            this.safetySettings = new HashMap<>(safetySettings);
```

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
```diff
@@ -59,7 +73,13 @@ public VertexAiGeminiChatModel(String project,
                                    Integer maxOutputTokens,
                                    Integer topK,
                                    Float topP,
-                                   Integer maxRetries) {
+                                   Integer maxRetries,
+                                   String responseMimeType,
```
langchain4j, line 80: please add Gemini to the list here: https://docs.langchain4j.dev/tutorials/ai-services#json-mode

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
```diff
@@ -99,43 +150,60 @@ public VertexAiGeminiChatModel(GenerativeModel generativeModel,
...
-            = ContentsMapper.splitInstructionAndContent(messages);
+        if (!this.safetySettings.isEmpty()) {
+            model = model.withSafetySettings(SafetySettingsMapper.mapSafetySettings(this.safetySettings));
+        }
 
-        GenerativeModel model = instructionAndContent.systemInstruction != null ?
-            modelWithTools.withSystemInstruction(instructionAndContent.systemInstruction) :
-            modelWithTools;
+        if (this.logRequests && logger.isDebugEnabled()) {
+            logger.debug("GEMINI ({}) tools: {} request: {}", modelName, tools, instructionAndContent);
```
langchain4j: minor: maybe put request first, then tools?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -356,6 +370,10 @@ void should_handle_parallel_function_calls() {
             .project(System.getenv("GCP_PROJECT_ID"))
             .location(System.getenv("GCP_LOCATION"))
             .modelName(GEMINI_1_5_PRO)
+            .temperature(0f)
```
langchain4j: nit: I would change the type of `temperature` to `Double` (and all other `Float` parameters) to avoid this nasty `f` at the end

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -390,6 +408,23 @@ void should_handle_parallel_function_calls() {
...
+        allMessages.add(ToolExecutionResultMessage.toolExecutionResultMessage(null, "getProductInventory",
+            "{\"product_id\":\"XYZ789\", \"stock\": 5}"));
+
+        messageResponse = model.generate(allMessages, stockInventoryToolSpec);
+
+        System.out.println("Final response = " + messageResponse.content().text());
+
+        // then
+        String text = messageResponse.content().text();
+        assertThat(text.split(" ")).containsAll(
```
langchain4j: minor: hard to read/understand this assertion

ghost (commented):
@glaforge thanks a lot! I did a quick look and left a few comments, will check it again and in more depth tomorrow or Thursday

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/PartsMapper.java
glaforge: I followed the _pattern_ that was used with the nested exception control flow.
I'm not quite sure what is the best approach here, as a function can return even raw text, not necessarily JSON.

My understanding is that 1) if it's JSON, it works fine, 2) if it's some value like a number, a boolean, it can be returned in a wrapped JSON object with a `result` key, but 3) if it's some text, it has to be quoted. Hence the three levels of nesting.
We have to parse somehow to know in which case we are.

But if you have a better idea, I'm interested.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
glaforge: Once we release 0.32 (with this PR), yes, happy to add it.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
glaforge: OK

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
glaforge: It's been that way since the first version, but I think it's not a breaking change, so ok.

glaforge: This applies to `topP` as well then.

glaforge: But internally, Gemini uses floats for those.

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
glaforge: Right, I'll try to update it a bit to make it clearer.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModel.java
glaforge: Actually, yes, since that's part of the sources, I should update it right away.

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
langchain4j: Ah ok then

glaforge: I made the change but then reverted back, to avoid any breaking change.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImageModel.java
```diff
@@ -63,21 +81,82 @@ public enum ImageStyle {
...
 
+    /**
+     * Supported aspect ratios: 1:1, 9:16, 16:9, 4:3, and 3:4.
+     * The 1:1 square format is the default aspect ratio when not specified.
+     */
+    public enum AspectRatio {
+        SQUARE("1:1"),
+        PORTRAIT("9:16"),
+        LANDSCAPE("16:9"),
+        THREE_FOURTH("3:4"),
```
langchain4j:
```suggestion
        THREE_FOURTHS("3:4"),
```

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImageModel.java
```diff
@@ -63,21 +81,82 @@ public enum ImageStyle {
...
+        FOUR_THIRD("4:3");
```
langchain4j:
```suggestion
        FOUR_THIRDS("4:3");
```

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImageModel.java
```diff
@@ -63,21 +81,82 @@ public enum ImageStyle {
...
+        }
+    }
+
+    /**
+     * Specify whether persons are allowed to be generated.
+     * By default, only adults are allowed.
+     * For all persons, including children, your project will need to be allowlisted.
+     */
+    public enum PersonGeneration {
+        dont_allow,
```
langchain4j: please make them upper-case

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImageModel.java
```diff
@@ -63,21 +81,82 @@ public enum ImageStyle {
...
+     * @param modelName        the name of the image model (<code>imagegeneration@002</code>, <code>imagegeneration@005</code>,
+     *                         <code>imagen-3.0-generate-preview-0611</code>, <code>imagen-3.0-fast-generate-preview-0611</code>)
      * @param seed             a fixed random seed number between 0 and 2^32-1
      * @param language         the spoken language used for the prompt
      * @param guidanceScale    an integer that represents the strength of the edit to make (0-9: low, 10-20: medium, 21+: high)
      * @param negativePrompt   a negative prompt to specify what you don't want to see in the generated image
      * @param sampleImageStyle the style of the image for Imagen v1, see the <code>ImageStyle</code> enum for reference
      * @param sampleImageSize  the size of the images to generate
+     * @param aspectRatio      the aspect ratio of the image, whether square, portrait or landscape
+     * @param personGeneration specify if it is allowed to generate persons (all, only adults, all)
```
langchain4j:
```suggestion
     * @param personGeneration specify if it is allowed to generate persons (none, only adults, all)
```

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImageModel.java
```diff
@@ -217,6 +360,28 @@ private Value prepareParameters(int n) throws InvalidProtocolBufferException {
...
+
+            paramsMap.put("outputOptions", outputOptions);
+        }
+
+        if (this.personGeneration != null) {
+            paramsMap.put("personGeneration", this.personGeneration);
+        }
+
+        if (this.watermark != null) {
+            paramsMap.put("addWatermark", this.watermark);
```
langchain4j:
```suggestion
            paramsMap.put("addWatermark", this.addWatermark);
```
maybe `addWatermark` to keep naming in sync?

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImageModel.java
```diff
@@ -270,7 +435,12 @@ public Response<Image> edit(Image image, Image mask, String prompt) {
     private URI persistAndGetURI(String bytesBase64Encoded) {
         if (this.withPersisting != null && this.withPersisting) {
             try {
-                Path tempFile = Files.createTempFile(this.tempDirectory, "imagen-image-", ".png");
+                String suffix = ".png";
+                if (this.mimeType != null && this.mimeType == MimeType.JPEG) {
```
langchain4j:
```suggestion
                if (this.mimeType == MimeType.JPEG) {
```
It is safe to compare null to enum

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -390,6 +408,23 @@ void should_handle_parallel_function_calls() {
...
+            "{\"product_id\":\"XYZ789\", \"stock\": 5}"));
+
+        messageResponse = model.generate(allMessages, stockInventoryToolSpec);
+
+        System.out.println("Final response = " + messageResponse.content().text());
+
+        // then
+        String text = messageResponse.content().text();
+        List<String> sequence = Arrays.asList("ABC123", "more", "XYZ789");
+        assertThat(text.split(" ")).containsAll(sequence);
```
langchain4j: you probably wanted to check that the order is correct?
```suggestion
        assertThat(text.split(" ")).containsSubsequence("ABC123", "more", "XYZ789");
```

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -456,28 +491,186 @@ void should_use_tools_with_AiService_2() {
...
+            .useGoogleSearch(true)
+            .build();
+
+        // when
+        String resp = modelWithSearch.generate(
+            "Why is the sky blue?");
+
+        // then
+        System.out.println("Google Search powered response = " + resp);
+        assertThat(resp).contains("scatter");
```
langchain4j, line 631: hmm, does it prove that google search was used?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -456,28 +491,186 @@ void should_use_tools_with_AiService_2() {
...
+            .modelName("gemini-1.5-flash-001")
+            .responseMimeType("application/json")
+            .build();
+
+        // when
+        String response = modelWithResponseMimeType.generate("Run a dice roll");
+
+        // then
+        System.out.println("response = " + response);
+        assertThat(response).containsAnyOf("1", "2", "3", "4", "5", "6");
```
langchain4j: This does not seem to prove that generated thing is a valid JSON, maybe copy-paste `OpenAiChatModelIT.should_generate_valid_json`?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
```diff
@@ -367,4 +387,123 @@ void should_accept_function_call() {
...
+        CompletableFuture<Response<AiMessage>> future = new CompletableFuture<>();
+        assistant.chat("Is there more stock of ABC or of XYZ?")
+            .onNext(System.out::println)
+            .onComplete(future::complete)
+            .onError(future::completeExceptionally)
+            .start();
+        Response<AiMessage> response = future.get(30, TimeUnit.SECONDS);
+
+        // then
+        assertThat(response.content().toString()).contains("XYZ");
```
langchain4j, line 443: This does not seem to verify that functions were called in parallel, I would suggest checking internals of a `ChatMemory`

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
```diff
@@ -367,4 +387,123 @@ void should_accept_function_call() {
...
+            .useGoogleSearch(true)
+            .build();
+
+        TestStreamingResponseHandler<AiMessage> handler = new TestStreamingResponseHandler<>();
+
+        // when
+        modelWithSearch.generate("Why is the sky blue?", handler);
+
+        // then
+        assertThat(handler.get().content().text()).containsIgnoringCase("scatter");
```
langchain4j, line 487: same here

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
```diff
@@ -367,4 +387,123 @@ void should_accept_function_call() {
...
+            .build();
+
+        // when
+        StringBuilder accumulatedResponse = new StringBuilder();
+        modelWithResponseMimeType.generate("Run a dice roll", onNext(accumulatedResponse::append));
+        String response = accumulatedResponse.toString();
+
+        // then
+        System.out.println("streamed response = " + response);
+        assertThat(response).containsAnyOf("1", "2", "3", "4", "5", "6");
```
langchain4j: same here

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/ContentsMapper.java
```diff
@@ -2,39 +2,100 @@
...
         InstructionAndContent instructionAndContent = new InstructionAndContent();
-
         List<Part> sysInstructionParts = new ArrayList<>();
 
-        for (ChatMessage message : messages) {
-            if (message instanceof SystemMessage) {
-                sysInstructionParts.addAll(PartsMapper.map(message));
+        List<ToolExecutionResultMessage> executionResultMessages = new ArrayList<>();
+
+        for (int msgIdx = 0; msgIdx < messages.size(); msgIdx++) {
```
langchain4j, line 36: the logic of this method is quite complicated, could you please cover it with unit tests?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
```diff
@@ -456,28 +491,186 @@ void should_use_tools_with_AiService_2() {
...
+        }
+    }
+
+    interface FunctionCallingAssistant {
+        @dev.langchain4j.service.SystemMessage("You MUST call the `paymentStatus()` function if you're asked about a payment transaction.")
+        String chat(String userMessage);
+    }
+
+    @Test
+    void should_work_with_interspersed_function_execution_results() {
```
langchain4j, line 568: Could you please clarify what `interspersed` means in this context?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/PartsMapper.java
langchain4j: I see. Maybe we should always quote the `functionResponseText` to avoid the third level of nesting? If it is number or boolean, I guess it is still OK to quote it. But it would be nice to understand why this must JSON at all.

ghost (commented):
@glaforge thanks a lot!

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/PartsMapper.java
glaforge: I wanted to do that initially, but decided against it, as it means we would always wrap even basic types like numbers, booleans, or null, into strings.

langchain4j-vertex-ai/src/main/java/dev/langchain4j/model/vertexai/VertexAiImageModel.java
glaforge: I went with lowercase, because they were automatically serialized as strings as is, as the API expects `dont_allow` and not `DONT_ALLOW`. But I guess I just need to define a `toString()` method, and that should do it.

glaforge: Also there was another enum like this with lower-case labels.
It might be a small breaking change though.

langchain4j: Lowercase enums is agains Java conventions, I would have a `toString` to serialize it properly as you suggested

glaforge: I'll follow the conventions.

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
glaforge: Not really. That's right.
At first, I had a test that worked very well, as it was searching for a very recent news item.
But after a few days, the news item disappeared from the Google news headlines, and the search wouldn't give me the result I was expecting.
So this is very hard to test, and not make the test flaky (and have to change it every week)

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
glaforge: Same thing, I haven't found a proper test that doesn't become flaky.

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
glaforge: It doesn't really matter here, but that was for me to remember that this could happen:
Gemini can answer with a function execution request AND a text message at the same time.
So in a way, the text message is interspersed in the function call request / response exchanges.
But it's just an extra test for function calling.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/ContentsMapper.java
glaforge: Good idea. It was a bit tricky to get right. Adding some tests.

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
glaforge: I'm adding a bunch of asserts on the `ChatMemory`.
However, even for the very same scenario, I don't always get parallel tool execution requests 100% of the time. Sometimes the model ask for them serially instead. So this test could be a bit flaky at times, when asserting that there really are parallel calls happening. That's something to be aware of.
But also, what's cool is that it shows that LangChain4j makes this transparent: whether requests are sequential or parallel, the end result is the same!

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelIT.java
langchain4j: Hmm, what about asking it what is current date?

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiStreamingChatModelIT.java
langchain4j: Thanks! If the test is working 90% of the time, it is good enough, we can think of adding automatic retries for flaky tests.
But most important IMHO is that desired behavior is tested.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/SchemaHelper.java
```diff
@@ -0,0 +1,96 @@
...
+            Float.class.isAssignableFrom(theClass) ||
+            float.class.isAssignableFrom(theClass)) {
+            return Schema.newBuilder().setType(Type.NUMBER).build();
+        } else if (theClass.isArray()) {
+            Class<?> componentType = theClass.getComponentType();
+            return Schema.newBuilder().setType(Type.ARRAY).setItems(
+                fromClass(componentType)
+            ).build();
+        } else if (Collection.class.isAssignableFrom(theClass)) {
+            // Because of type erasure, we can't easily know the type of the items in the collection
```
langchain4j, line 77: This is not entirely true, see `ToolSpecifications.toJsonSchemaProperties`:

```java
if (Collection.class.isAssignableFrom(type)) {
    return removeNulls(ARRAY, arrayTypeFrom((Class<?>) ((ParameterizedType) field.getGenericType()).getActualTypeArguments()[0]), description);
}
```

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/SchemaHelper.java
```diff
@@ -0,0 +1,96 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * Create an instance of <code>Schema</code> from a class by reflection on its fields.
+     * @param theClass the class for which to create a schema representation
+     * @return a fully built schema
+     */
+    public static Schema fromClass(Class<?> theClass) {
```
langchain4j, line 53: It seems that this kind of logic is already implemented in `ToolSpecifications`, maybe worth looking into re-using it.

ghost (approved):
@glaforge great job, thank you!

PR #1279, opened 2024-06-13
Open up access to ServiceOutputParser#jsonStructure

geoand:
This is useful in Quarkus, so we can add better support for `List<SomeType> ` (although we will eventually probably end up writing our own since we have all the generic type information and can expand more types)


ghost (approved):

geoand:
Thanks!

Issue #1266, opened 2024-06-11
[BUG] Tools in streaming AI Services do not work without memory


langchain4j:
TBD

This has been fixed, but not for the streaming case: https://github.com/langchain4j/langchain4j/pull/819

Kugaaa:
I gave it a try, thank you for CR~

I am considering whether a new `Memory` implementation class can replace the temp list

PR #1280, opened 2024-06-13
Fix 1266 Allow using streaming AI Service with tools without memory

Kugaaa:
## Issue

See #1266 

## Change

- Add temp memory list to allow using streaming AI Service with tools without memory
- Add unit testing

## General checklist


langchain4j/src/test/java/dev/langchain4j/service/StreamingAiServicesWithToolsIT.java
```diff
@@ -130,5 +130,36 @@ void should_use_tool_with_List_of_Strings_parameter(StreamingChatLanguageModel m
         );
     }
 
+    @ParameterizedTest
+    @MethodSource("models")
+    void should_use_tool_without_memory(StreamingChatLanguageModel model) throws Exception {
```
langchain4j: Could you please follow the same approach as in `AiServicesWithToolsWithoutMemoryIT`? You could create `StreamingAiServicesWithToolsWithoutMemoryIT` and add all 3 test methods there

ghost (commented):
@Kugaaa thank you!

This does not seem to work as expected, I can see that `should_use_tool_without_memory` is failing for the second model (Mistral).

For OpenAI, it does not work properly as well:
In the first request, user message is sent.
In the second request, only assistant and tool messages are sent (user message is not sent, but should be).

langchain4j/src/test/java/dev/langchain4j/service/StreamingAiServicesWithToolsIT.java
Kugaaa: No problem ~ I will create a `StreamingAiServicesWithToolsWithoutMemoryIT` test class

Kugaaa:
> @Kugaaa thank you!
> 
> This does not seem to work as expected, I can see that `should_use_tool_without_memory` is failing for the second model (Mistral).
> 
> For OpenAI, it does not work properly as well: In the first request, user message is sent. In the second request, only assistant and tool messages are sent (user message is not sent, but should be).

I'll check again

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingResponseBuilder.java
```diff
@@ -29,7 +29,7 @@ public class OpenAiStreamingResponseBuilder {
     private final StringBuffer toolNameBuilder = new StringBuffer();
     private final StringBuffer toolArgumentsBuilder = new StringBuffer();
 
-    private final Map<Integer, ToolExecutionRequestBuilder> indexToToolExecutionRequestBuilder = new HashMap<>();
```
Kugaaa: sorry, I found `toolCall.index()` is null, the specific reason is still unclear

langchain4j: Does this happen with OpenAI? Every model?

Kugaaa: I am unable to access OpenAI in my region, so I am using a third-party proxy service 😿
I cannot confirm if it is a third-party service issue 😿

langchain4j: I see. So `toolCall.index() == null` blocks you from testing changes in this PR?

Kugaaa: I change `indexToToolExecutionRequestBuilder` to `HashMap` to ignore `toolCall.index() == null`
The `StreamingAiServicesWithToolsWithoutMemoryIT` in PR is passed

langchain4j: @Kugaaa this will most probably not work when there are multiple tool calls in the LLM response.

`ToolCall toolCall = delta.toolCalls().get(0)` means that there is only one `ToolCall` **_in the current chunk_** `ChatCompletionResponse`. But if LLM decides to call multiple tools in the same response, `ToolCall toolCall = delta.toolCalls().get(0)` will return another `ToolCall`, with other name/arguments. I guess we need to count `index` manually or find other solution (maybe rely on `toolCall.id()`, but some models probably don't return it?)

Kugaaa: Yes, I understand its purpose. I modified it to a `HashMap` for testing the `StreamingAiServicesWithToolsWithoutMemoryIT.should_execute_multiple_tools_sequentially_then_answer` method (since there would only be one `ToolCall` in this case). 
However, I accidentally committed the code, so I reverted it in [this commit](https://github.com/langchain4j/langchain4j/pull/1280/commits/5ff6e9cfb0d7fc37ecc1f753950c1d331d6765bb)(I am not trying to express that I want to replace `ConcurrentHashMap` with `HashMap`).

Additionally, I checked the [OpenAI doc](https://platform.openai.com/docs/api-reference/chat/streaming), and there should be the `index` field. 
I have already raised an [issue](https://github.com/chatanywhere/GPT_API_free/issues/254#issuecomment-2180471421) with the third-party service I'm using to inquire about why this is happening. I suspect it's an issue with the third-party service.

ghost (approved):
@Kugaaa thank you!

Issue #1294, opened 2024-06-14
[BUG] [AzureOpenAI] TokenUsage set to null 


fb33:
If we define a AI Assistant with 

```
import dev.langchain4j.service.Result;

public interface Assistant {
    Result<String> answer(String query);
}
```

When an error occurs during the generate phase, then the Response built got a null TokenUsage.
See here for example line 268: https://github.com/langchain4j/langchain4j/blob/25e4d6de4932fc23bdd92a2c9c9fc2613169f125/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java#L256-L270

But, the AiService will try to map the response into a Result (following the interface)
https://github.com/langchain4j/langchain4j/blob/25e4d6de4932fc23bdd92a2c9c9fc2613169f125/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java#L202-L210

and there is a rule in Result constructor to avoid null for TokenUsage
https://github.com/langchain4j/langchain4j/blob/25e4d6de4932fc23bdd92a2c9c9fc2613169f125/langchain4j/src/main/java/dev/langchain4j/service/Result.java#L27-L31

To fix that, can we create an empty instance of TokenUsage in the catch block ?
```
            return Response.from(
                    aiMessage(httpResponseException.getMessage()),
                    new TokenUsage(),
                    exceptionFinishReason
            );
```

fb33:
ping @jdubois @agoncal 

I made the small PR with the proposal

agoncal:
I wasn't aware of this API. 

Looking at https://github.com/langchain4j/langchain4j/blob/main/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java#L181 I can see that OpenAI uses the `InternalOpenAiHelper.tokenUsageFrom()` method.

Azure OpenAI also has this method in `InternalAzureOpenAiHelper`. I wonder if we should use it?  https://github.com/langchain4j/langchain4j/blob/main/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java#L299

langchain4j:
I guess we should relax the `ensureNotNull(tokenUsage, "tokenUsage")` requirement, not all models return it

fb33:
[...]

PR #1295, opened 2024-06-14
fix(TokenUsage): Token can't be null in AI Response

fb33:
Signed-off-by: François Barbe <f.barbe@lectra.com>

## Issue
https://github.com/langchain4j/langchain4j/issues/1294

## Change
add a empty TokenUsage in case of exception in the Response. 

## General checklist


langchain4j:
Hi @fb33, thank you!

I guess we should relax the `ensureNotNull(tokenUsage, "tokenUsage")` requirement instead, as not all models return it.
WDYT?

fb33:
@langchain4j yes I'm agree but I've this issue when I tried other type of search than Vector with Azure Ai Search.
Perhaps for other Index DB implementations, the requirement is still relevant ?

langchain4j:
@fb33 sorry, not sure how vector DBs (Index DB) are relevant here?

fb33:
My bad... I mixed topics... my comment makes no sense :) 

So, yes relax the ensureNotNull(tokenUsage, "tokenUsage") requirement is a good option for me.

fb33:
@langchain4j I revert my previous proposal and just relax the requirement as you said.

ghost (approved):
@fb33 thank you!

PR #1305, opened 2024-06-17
chroe: add ChatModelListener support for azure openai chat model and streaming chat model

qiaoleiatms:
…hat model

## Issue
https://github.com/langchain4j/langchain4j/issues/199

## Change
add ChatModelListener support for azure openai:
- chat model
- streaming chat model

## General checklist

### DIFF
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
@@ -36,8 +36,7 @@ import java.util.List;
 import java.util.Map;
 
 import static dev.langchain4j.data.message.AiMessage.aiMessage;
-import static dev.langchain4j.internal.Utils.getOrDefault;
-import static dev.langchain4j.internal.Utils.isNullOrEmpty;
+import static dev.langchain4j.internal.Utils.*;
 import static dev.langchain4j.model.azure.InternalAzureOpenAiHelper.*;
 import static dev.langchain4j.spi.ServiceHelper.loadFactories;
 import static java.util.Collections.emptyList;
@@ -397,7 +396,9 @@ public class AzureOpenAiStreamingChatModel implements StreamingChatLanguageModel
                         responseBuilder.append(chatCompletions);
                         handle(chatCompletions, handler);
 
-                        responseId.set(chatCompletions.getId());
+                        if (!isNotNullOrBlank(chatCompletions.getId())) {
+                            responseId.set(chatCompletions.getId());
+                        }
                     });
             Response<AiMessage> response = responseBuilder.build(tokenizer, toolThatMustBeExecuted != null);
             ChatModelResponse modelListenerResponse = createModelListenerResponse(
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
@@ -15,8 +15,17 @@ import dev.langchain4j.model.Tokenizer;
 import dev.langchain4j.model.azure.spi.AzureOpenAiStreamingChatModelBuilderFactory;
 import dev.langchain4j.model.chat.StreamingChatLanguageModel;
 import dev.langchain4j.model.chat.TokenCountEstimator;
+import dev.langchain4j.model.chat.listener.ChatModelErrorContext;
+import dev.langchain4j.model.chat.listener.ChatModelListener;
+import dev.langchain4j.model.chat.listener.ChatModelRequest;
+import dev.langchain4j.model.chat.listener.ChatModelRequestContext;
+import dev.langchain4j.model.chat.listener.ChatModelResponse;
+import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
 import dev.langchain4j.model.output.FinishReason;
 import dev.langchain4j.model.output.Response;
+import java.util.ArrayList;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import reactor.core.publisher.Flux;
@@ -31,6 +40,7 @@ import static dev.langchain4j.internal.Utils.getOrDefault;
 import static dev.langchain4j.internal.Utils.isNullOrEmpty;
 import static dev.langchain4j.model.azure.InternalAzureOpenAiHelper.*;
 import static dev.langchain4j.spi.ServiceHelper.loadFactories;
+import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 
 /**
@@ -80,6 +90,7 @@ public class AzureOpenAiStreamingChatModel implements StreamingChatLanguageModel
     private final AzureChatEnhancementConfiguration enhancements;
     private final Long seed;
     private final ChatCompletionsResponseFormat responseFormat;
+    private final List<ChatModelListener> listeners;
 
     public AzureOpenAiStreamingChatModel(OpenAIClient client,
                                          OpenAIAsyncClient asyncClient,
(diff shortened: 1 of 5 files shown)


jdubois:
That's good for me @qiaoleiatms ! Thanks!!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
```diff
@@ -309,32 +337,105 @@ private void handleResponseException(Throwable throwable, StreamingResponseHandl
...
         try {
+            AtomicReference<String> responseId = new AtomicReference<>();
+
             client.getChatCompletionsStream(deploymentName, options)
                     .stream()
                     .forEach(chatCompletions -> {
                         responseBuilder.append(chatCompletions);
                         handle(chatCompletions, handler);
+
+                        responseId.set(chatCompletions.getId());
```
langchain4j: please add a check `isNullOrBlank(chatCompletions.getId())`

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -322,6 +335,135 @@ void should_support_all_string_model_names(AzureOpenAiChatModelName modelName) {
         assertThat(response.content().text()).isNotBlank();
     }
 
+    @ParameterizedTest(name = "Deployment name {0} using {1}")
+    @CsvSource({
+        "gpt-4o,        gpt-4o"
+    })
+    void should_listen_request_and_response(String deploymentName, String gptVersion) {
```
langchain4j: minor: `gptVersion` can be removed together with `.tokenizer(new AzureOpenAiTokenizer(gptVersion))`

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -322,6 +335,135 @@ void should_support_all_string_model_names(AzureOpenAiChatModelName modelName) {
...
+        assertThat(response.tokenUsage().totalTokenCount()).isGreaterThan(0);
+        assertThat(response.finishReason()).isNotNull();
+        assertThat(response.aiMessage()).isEqualTo(aiMessage);
+    }
+
+    @ParameterizedTest(name = "Deployment name {0} using {1}")
+    @CsvSource({
+        "gpt-4o,        gpt-4o"
+    })
+    void should_listen_error(String deploymentName, String gptVersion) {
```
langchain4j: `deploymentName` and `gptVersion` unused

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -322,6 +335,135 @@ void should_support_all_string_model_names(AzureOpenAiChatModelName modelName) {
...
+                assertThat(responseContext.attributes().get("id")).isEqualTo("12345");
+            }
+
+            @Override
+            public void onError(ChatModelErrorContext errorContext) {
+                fail("onError() must not be called");
+            }
+        };
+
+        AzureOpenAiChatModelName modelName = GPT_4_O;
```
langchain4j:
```suggestion
```

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -322,6 +335,135 @@ void should_support_all_string_model_names(AzureOpenAiChatModelName modelName) {
...
+            .addParameter("a", INTEGER)
+            .addParameter("b", INTEGER)
+            .build();
+
+        // when
+        AiMessage aiMessage = model.generate(singletonList(userMessage), singletonList(toolSpecification)).content();
+
+        // then
+        ChatModelRequest request = requestReference.get();
+        assertThat(request.model()).isEqualTo(modelName.toString());
```
langchain4j:
```suggestion
        assertThat(request.model()).isEqualTo(deploymentName);
```

ghost (commented):
@qiaoleiatms thank you!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
qiaoleiatms: added check

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
qiaoleiatms: removed gptVersion and tokenizer call

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
qiaoleiatms: removed unused

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
qiaoleiatms: removed redundant declaration

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
qiaoleiatms: use deploymentName instead

ghost (approved):
@qiaoleiatms thank you!

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModelIT.java
```diff
@@ -413,4 +424,153 @@ public void onError(Throwable error) {
...
+        StreamingResponseHandler<AiMessage> handler = new StreamingResponseHandler<AiMessage>() {
+
+            @Override
+            public void onNext(String token) {
+                fail("onNext() must not be called");
+            }
+
+            @Override
+            public void onError(Throwable error) {
+                fail("OnError() must not be called");
```
langchain4j, line 557: Hi @qiaoleiatms, I've jsut noticed that this test (and implementation in `AzureOpenAiStreamingChatModel`) is different from what has been implemented in other models (e.g. `OpenAiStreamingChatModel`).

In case of errors, `onError()` should be called, not `onComplete()`. Could you please check?

BTW, I have moved tests to `AzureOpenAiStreamingChatModelListenerIT` and disabled `should_listen_error` untill this discrepancy is fixed.

Thanks!

qiaoleiatms: @langchain4j that is because while exception happen, the HttpResponseException will be treated as a complete response from the original implementation 

https://github.com/langchain4j/langchain4j/blob/a2004b689a5a010cfef20fadef4fff6d14efaca2/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java#L329

@jdubois could you please help to confirm the discrepency between the implementation of OpenAiStreamingChatModel? it treats all the exception as error

jdubois: I'm not totally sure.
For the "content_filter", the explanation is that on Azure OpenAI we have some specific filters for removing violence, sexual content, etc. Those are not part of the LLM, but are added on top of it. So they would not be an "error" (=the service could not process the request), but as a response that "completed" (=it worked, but it was filtered out as the user requested something that shouldn't be generated). I don't think the other services have this, so it's indeed specific to Azure OpenAI.

langchain4j: I am not sure about harmful content, but in this specific case the problem is wrong API key, which IMHO should definitely result in an exception and `onError()` being called instead of `onComplete()`

qiaoleiatms: @langchain4j yeah, get your point, but I believe this is handled by the azure openai sdk(=complete the request even it's an API key error) 
@jdubois how do you think about it?

langchain4j: @qiaoleiatms

> I believe this is handled by the azure openai sdk(=complete the request even it's an API key error)

Not really, SDK is throwing this exception:
`com.azure.core.exception.ClientAuthenticationException: Status code 401, "{"error":{"code":"401","message":"Access denied due to invalid subscription key or wrong API endpoint. Make sure to provide a valid key for an active subscription and use a correct regional API endpoint for your resource."}}"`

PR #1325, opened 2024-06-20
Polish OpenSearch and Vearch IT

eddumelendez:
* Update LocaStack image version
* Avoid custom LocalStackContainer configuration
* Fixed Vearch image version
* Use copyFileToContainer

Note: almost all Vearch IT works with version `3.4.x`. `should_add_embedding_with_segment_with_metadata` fails with

```
status code: 400; body: {"error":{"root_cause":[{"type":"","reason":"param have error field [long_0]"}],"type":"","reason":"param have error field [long_0]"},"status":400}
```

PR #1336, opened 2024-06-20
[Fix] fix VearchEmbeddingStoreIT

Martin7-1:
## Issue

#1325 

## Change

Update `VearchEmbeddingStoreIT`. The exception in the issue is because the update of `EmbeddingStoreIT`. `EmbeddingStoreIT` add more metadatas so `VearchEmbeddingStore` failed because of the mismatch of metadatas.

There are two major changes in `VearchEmbeddingStoreIT`

1. Because `Vearch` do not support long and double, so I manually override `should_add_embedding_with_segment_with_metadata`.
2. Because `Vearch` do not allow `null` value, so every metadata should contains a default value when embed. When the test create a Space with metadatas, all `findRelevant` result will carry metadata even though users don't embed content with metadata. So, tests like `should_add_embedding_with_segment` which assert embedded will failed. (because `Vearch` return embedded with default metadata, but this test and assertion do not expect any metadata) . To solve this problem, I check where the test from and create different Space before each test. (see `@BeforeEach` in `VearchEmbeddingStoreIT`)

image

## General checklist


Martin7-1:
cc @eddumelendez

ghost (approved):
@Martin7-1 thank you!

PR #1338, opened 2024-06-20
Do not run CI for files not impacting CI

gsmet:
There's a good chance you will make a lot of small changes to the bot config file and you really don't want to run CI for it.

Also ignored a bunch of files that don't impact CI at all.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -4,9 +4,25 @@ on:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


gsmet:
Ah. I just noticed that it might not work as the CI builds are required to merge something.

But these changes will probably be merged by administrators so maybe that works OK?

I will let you decide.

ghost (approved):
@gsmet thank you!

PR #1347, opened 2024-06-21
Draft: [BUG 1340] Updates naming of keys prefix and index in RedisEmbeddingStore

### LINKED ISSUE #1340: [BUG] RedisEmbeddingStore returns embeddings from different indexes
**Describe the bug**
During  search RedisEmbeddingStore finds embeddings in all existing Redis embedding stores, in the same instance of Redis (not only current one) .
Bug was noticed around version 0.23.0, when TODO comment was added in the RedisEmbeddingStoreIT:
```
// TODO fix: why redis returns embeddings from different indexes?
```

**To Reproduce**
Create at least two Redis storages using the same Redis instance.
Embeddings added to the first storage would be found when search is done with use of second storage.

It's also possible to run this test (can be added to RedisEmbeddingStoreIT):
 ```	
    @Test
    void should_find_embedding_in_proper_store() {
       final EmbeddingStore<TextSegment> firstStore = createStore("first-store");
       final EmbeddingStore<TextSegment> secondStore = createStore("second-store");

       Embedding embedding = embeddingModel().embed("hello").content();
       
       String id = firstStore.add(embedding);
       assertThat(id).isNotBlank();

       List<EmbeddingMatch<TextSegment>> relevant = firstStore.findRelevant(embedding, 10);
       assertThat(relevant).hasSize(1);
       assertThat(secondStore.findRelevant(embedding, 10)).hasSize(0);

       // new API
       assertThat(firstStore.search(EmbeddingSearchRequest.builder()
               .queryEmbedding(embedding)
               .maxResults(10)
               .build()).matches()).isEqualTo(relevant);
       assertThat(secondStore.search(EmbeddingSearchRequest.builder()
               .queryEmbedding(embedding)
               .maxResults(10)
               .build()).matches()).hasSize(0);
    }
    
    private EmbeddingStore<TextSegment> createStore(String name) {
      return RedisEmbeddingStore.builder()
          .host(redis.getHost())
          .port(redis.getFirstMappedPort())
          .indexName(name)
          .dimension(384)
          .metadataKeys(createMetadata().toMap().keySet())
          .build();      
  }
```    

Current tests in the RedisEmbeddingStoreIT  can also be used to reproduce, when lines:
```
        try (JedisPooled jedis = new JedisPooled(redis.getHost(), redis.getFirstMappedPort())) {
            jedis.flushDB(); // TODO fix: why redis returns embeddings from different indexes?
        }
```
are removed.

**Expected behavior**
RedisEmbeddingStore should only return embeddings from current storage

- LangChain4j version: 0.31.0 
- LLM(s) used: llama3
- Java version: openjdk 21.0.2 2024-01-16 LTS
- Spring Boot version (if applicable): 3.2.5

**Additional context**
Problem is caused by having the same key prefix (defined in RedisSchema as "embedding:") for all embeddings stored in the Redis.
Since that all Redis indexes created by langchain4j are searching through the same embedding source.
(I'm currently working on PR)

### PR DESCRIPTION
**Pull request by Rommek**
## Issue
Fixes https://github.com/langchain4j/langchain4j/issues/1340

## Change

For creation of the **RedisEmbeddingStore**, field 'name' was created and is used instead of 'indexName'. 
'name' field is used to create indexName and key prefix for **RedisSchema**.
Redis index name and key prefix are now connected, as both uses provided name as a base.
Default values were kept as in the original implemention, to keep backward compatibility.
Integration test was added in the RedisEmbeddingStoreIT, as issue with indexes is specific to the Redis.

With this implementation, if unique storage name is used, Redis embedding store would only return results from current store.

## General checklist
- [ ] ?_(only integration test)_ I have added unit and integration tests for my change 
- [ ] ?_(tests that required api key were skipped, rest green)_ I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] ?_(previously persisted data can be used with default name, after fix new stores would use custom name)_ I have manually verified that the `RedisEmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

### DIFF
--- a/langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java
+++ b/langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java
@@ -63,7 +63,7 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
         this.client = user == null ? new JedisPooled(host, port) : new JedisPooled(host, port, user, password);
         this.schema = RedisSchema.builder()
                 .prefix(toPrefix(getOrDefault(name, "embedding")))
-                .indexName(toIndex(getOrDefault(name, "embedding-index")))
+                .indexName(toIndexName(getOrDefault(name, "embedding-index")))
                 .dimension(dimension)
                 .metadataKeys(metadataKeys)
                 .build();
@@ -77,7 +77,7 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
         return name + ":";
     }
 
-    private static String toIndex(String name) {
+    private static String toIndexName(String name) {
         return name + "-index";
     }
 
--- a/langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java
+++ b/langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java
@@ -271,7 +271,7 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
         
         /**
          * @param name The name of the store (optional). Used for prefix and index name generation. 
-         * Default value: "embedding:".
+         * Default value: "embedding".
          * @return builder
          */
         public Builder name(String name) {
--- a/langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java
+++ b/langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java
@@ -45,7 +45,7 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
      * @param port         Redis Stack Server port
      * @param user         Redis Stack username (optional)
      * @param password     Redis Stack password (optional)
-     * @param indexName    The name of the index (optional). Default value: "embedding-index".
+     * @param name         The name of the store (optional). Used for prefix and index name generation.
      * @param dimension    Embedding vector dimension
      * @param metadataKeys Metadata keys that should be persisted (optional)
      */
@@ -53,7 +53,7 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
                                Integer port,
                                String user,
                                String password,
-                               String indexName,
+                               String name,
                                Integer dimension,
                                Collection<String> metadataKeys) {
         ensureNotBlank(host, "host");
@@ -62,7 +62,8 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
 
         this.client = user == null ? new JedisPooled(host, port) : new JedisPooled(host, port, user, password);
         this.schema = RedisSchema.builder()
-                .indexName(getOrDefault(indexName, "embedding-index"))
+                .prefix(toPrefix(getOrDefault(name, "embedding")))
+                .indexName(toIndex(getOrDefault(name, "embedding-index")))
                 .dimension(dimension)
                 .metadataKeys(metadataKeys)
                 .build();
@@ -70,6 +71,14 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
         if (!isIndexExist(schema.indexName())) {
             createIndex(schema.indexName());
         }
+    }
+    
+    private static String toPrefix(String name) {
+        return name + ":";
+    }
+
+    private static String toIndex(String name) {
+        return name + "-index";
     }
 
     @Override
@@ -223,7 +232,7 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
         private Integer port;
         private String user;
         private String password;
-        private String indexName;
+        private String name;
         private Integer dimension;
         private Collection<String> metadataKeys = new ArrayList<>();
 
@@ -259,14 +268,16 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
             return this;
         }
 
+        
         /**
-         * @param indexName The name of the index (optional). Default value: "embedding-index".
+         * @param name The name of the store (optional). Used for prefix and index name generation. 
+         * Default value: "embedding:".
          * @return builder
          */
-        public Builder indexName(String indexName) {
-            this.indexName = indexName;
+        public Builder name(String name) {
+            this.name = name;
             return this;
-        }
+        }        
 
         /**
          * @param dimension embedding vector dimension
@@ -296,7 +307,7 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
         }
 
         public RedisEmbeddingStore build() {
-            return new RedisEmbeddingStore(host, port, user, password, indexName, dimension, metadataKeys);
+            return new RedisEmbeddingStore(host, port, user, password, name, dimension, metadataKeys);
         }
     }
 }
--- a/langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java
+++ b/langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java
@@ -268,6 +268,15 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
             return this;
         }
 
+        /**
+         * @param indexName The name of the index (optional). Default value: "embedding-index".
+         * @deprecated use {@link #name(String)} instead. Method does not modify the Redis index name.
+         * Default value is used for the name of the index.
+         */
+        @Deprecated
+        public Builder indexName(String name) {
+            return this;
+        }
         
         /**
          * @param name The name of the store (optional). Used for prefix and index name generation. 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Martin7-1** [comment]
Thank you! It looks good to me. @langchain4j WDYT?

File: langchain4j-redis/src/main/java/dev/langchain4j/store/embedding/redis/RedisEmbeddingStore.java [hunk 1]
```diff
@@ -259,14 +268,16 @@ public Builder password(String password) {
...
         }
 
+        
         /**
-         * @param indexName The name of the index (optional). Default value: "embedding-index".
+         * @param name The name of the store (optional). Used for prefix and index name generation. 
+         * Default value: "embedding".
          * @return builder
          */
-        public Builder indexName(String indexName) {
```
langchain4j, line 266: I am pretty sure there are users that use this method, so could you please make this change in a non-breaking way? You could keep `indexName` method and mark is as `@Deprecated` and introduce a new one called `name`

Rommek: Hi @langchain4j , good point. 

I just sent the update. I went with deprecated indexName not setting any field, assuming that accessing already created embeddings (with default prefix) have higher priority than keeping index name intact. I view index as a part of internal storage representation. 

With that approach, embeddings already created with langchain4j, would be accessible without problems after fix is applied. Although additional embedding index would be created in the Redis instance.

I was also thinking about exposing indexName as additional parameter, so user of RedisEmbeddingStore could specify separately prefix for keys and indexName. But it would make creation of RedisEmbeddingStore a bit more complicated, so for now I stayed with simpler approach.

@langchain4j , please let me know, if this approach looks good to you.

langchain4j:
@Rommek I am wondering if it makes more sense to make the `prefix` configurable in the builder/ctor?
I've checked [Quarkus extension](https://docs.quarkiverse.io/quarkus-langchain4j/dev/redis-store.html) and it looks like they do exactly that.

Rommek:
Hi @langchain4j  , thank you for your time on this topic.

I looked a bit into Quarkus approach you mentioned and indeed I might be a bit biased towards coupling key prefix and index name.
I like the idea of specifying storage name and don't thinking too much, how it's represented in the Redis itself; 
But in this way some expected flexibility is lost, which can be quite confusing for people already using it.

So since having prefix configurable can also resolve connected issue, it's sounds good from my side. 
I will make some updates in the implementation (and possibly add some small info in the javaDoc, to keep awareness about importance of key prefix) .

langchain4j:
@Rommek hmm, what about having `prefix` configurable, but it's default value will be `indexName`? This way there is flexibility for the user to choose any prefix, but also no problem when using multiple `indexName` ?

Rommek:
Hi @langchain4j , I was bit unsure about using indexName as a base for prefix, so I went looking for inspiration and checked few existing implementations:

- [Quarkus (as you mentioned in conversation)](https://github.com/quarkiverse/quarkus-langchain4j/blob/main/embedding-stores/redis/runtime/src/main/java/io/quarkiverse/langchain4j/redis/RedisEmbeddingStore.java)
- [Spring-ai](https://github.com/spring-projects/spring-ai/blob/main/vector-stores/spring-ai-redis-store/src/main/java/org/springframework/ai/vectorstore/RedisVectorStore.java)
- [langchain](https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/vectorstores/redis/base.py)

_Qurarkiverse_ one seems most demading, as no default values are present (at least I didn't noticed any).
_Spring-ai_ is using defaults prefix - "embedding:"  and indexName - "spring-ai-index"
In the _langchain_ there is quite nice approach:
- default for prefix is: f"doc:{index_name}" 
- default for indexName: is generated UUID4

I would say that langchain won competition for the safest default values :)


**Proposed solution:**
So after those checks, I think that having prefix and indexName configurable can work quite fine. 

If possible, I would go with similar approach for defaults, as in the _langchain_:
- default for prefix - "embedding:{indexName}" 
- default for indexName - generated UUID

Sadly, modifying current default values (prefix - "embedding:", indexName - "embedding-index") will be a bit problematic.
To access embeddings stored in Redis before fix was included, users would need to explicitly set prefix as "embedding:"
Maybe it would be possible to include this information in changelog and use new defaults in next version?

What do you think about this approach?

langchain4j:
@Rommek thank you for the thorough analysis!

If there is absolutely no way to make this change without breaking it for existing users, let's do the one you propose and include it in release notes.

I am just not sure about generated UUID for `indexName`, would it be more intuitive to use static default `indexName`? So if the user did not define it explicitly and runs the app multiple times, the same index will be used?

Rommek:
Hi @langchain4j,

> If there is absolutely no way to make this change without breaking it for existing users, let's do the one you propose and include it in release notes.

Currently I cannot find a better way. I think it's possible to have some custom logic to discover already existing keys with specific prefix in the Redis instance; but it could get messy really fast and eventually it would need to be removed. 

>I am just not sure about generated UUID for indexName, would it be more intuitive to use static default indexName? So if the user did not define it explicitly and runs the app multiple times, the same index will be used?

True, also cluttering Redis with forgotten data could be avoided. But in that case we would need to take into consideration case when multiple storages are used in the same time i.e.
- storage 1 (custom prefix: newspaper, default indexName i.e. langchain4j)
- storage 2 (custom prefix: forumThreads, default indexName i.e. langchain4j)

In the current implementation Redis index is created only if index with the same name doesn't exists. So with multiple storages we would end up searching through the first one created.

Maybe I could use some custom logic for default values when during RedisEmbeddingStore creation:
- indexName  == null && prefix == null:
	- indexName = "embedding-index"
	- prefix = "embedding:"
- indexName == "someValue" && prefix == null:
	- prefix = "embedding:{indexName}"
- indexName == null && prefix == "someValue":
	- indexName = "{prefix}-index"

In this case data in the Redis storage created with default values would be available after application restart. Also in case indexName or prefix is provided default values would be quite safe.

langchain4j:
@Rommek it seems that changing default values complicates things a lot...
As far as I understand, `indexName` does not really matter, as `prefix` is the parameter which decides which keys will be searched in the end. So the only thing we need to do now is to make `prefix` configurable, right?

@Martin7-1 perharps you could comment as well?

BTW, while reading the code, I've noticed that `indexDefinition` is created, but not used in this method:
```java
    private void createIndex(String indexName) {
        IndexDefinition indexDefinition = new IndexDefinition(JSON);
        indexDefinition.setPrefixes(schema.prefix());
        String res = client.ftCreate(indexName, FTCreateParams.createParams()
                .on(IndexDataType.JSON)
                .addPrefix(schema.prefix()), schema.toSchemaFields());
        if (!"OK".equals(res)) {
            if (log.isErrorEnabled()) {
                log.error("create index error, msg={}", res);
            }
            throw new RedisRequestFailedException("create index error, msg=" + res);
        }
    }
```
Is this expected?

langchain4j:
@Rommek @Martin7-1 could you please check my latest comment? Thank you!

Martin7-1:
@langchain4j @Rommek Hi! Sorry for replying so late.

> As far as I understand, indexName does not really matter, as prefix is the parameter which decides which keys will be searched in the end. So the only thing we need to do now is to make prefix configurable, right?

I support this point... Maybe `indexName` can be UUID like `langchain`, and `prefix` should be configurable so user can search the data they want.

If `indexName` change to UUID, current index creation should be change as the logic is:

```java
        if (!isIndexExist(schema.indexName())) {
            createIndex(schema.indexName());
        }
```

The `createIndex` is triggered when the `indexName` is not exist. But if we change it to UUID, maybe the `isIndexExist` is not work and we should introduce a new variable like `createIndex`, the logic should change like:

```java
        if (createIndex) {
            createIndex(schema.indexName());
        }
```

> BTW, while reading the code, I've noticed that indexDefinition is created, but not used in this method:

After double check implementation, I think it's unused and do not really need. @Rommek Could you please remove it?

langchain4j:
@Martin7-1 

> Maybe indexName can be UUID like langchain, and prefix should be configurable so user can search the data they want.

Do you mean random UUID? Not sure I understand the benefit.

Martin7-1:
> > Maybe indexName can be UUID like langchain, and prefix should be configurable so user can search the data they want.
> 
> Do you mean random UUID? Not sure I understand the benefit.

@langchain4j There's no obvious advantage. It doesn't matter to the user what the `indexName` is, so I think it's fine to use the UUID when the user doesn't specify it.

langchain4j:
@Martin7-1 I guess I am still not getting it 😅
Currently `indexName` has a default value "embedding-index". If we change it to a (random?) UUID, it will break for existing users who already have it as "embedding-index".

Isn't it enough to just make `prefix` configurable?

Martin7-1:
> @Martin7-1 I guess I am still not getting it 😅 Currently `indexName` has a default value "embedding-index". If we change it to a (random?) UUID, it will break for existing users who already have it as "embedding-index".
> 
> Isn't it enough to just make `prefix` configurable?

Sorry for confusing it... Yeah, just keeping `indexName` deafult value and makeing `prefix` configurable is enough.

langchain4j:
Hi @Rommek, do you still plan to finalize this PR?

Martin7-1:
Looks like this PR is no response for a long time...#1635 and #1833 need this PR to be merged.

I will take this PR over and create a new one. Now you can close it. @langchain4j

langchain4j:
@Martin7-1 thank you for taking care of it!

Do you mean https://github.com/langchain4j/langchain4j/pull/1635 and https://github.com/langchain4j/langchain4j/pull/1833 cannot be merged before prefix is configurable? If yes, could you please mention this in descriptions of those PRs? 🙏

Martin7-1:
> thank you for taking care of it!
> 
> Do you mean #1635 and #1833 cannot be merged before prefix is configurable? If yes, could you please mention this in descriptions of those PRs? 🙏

Sure.

PR #1375, opened 2024-06-28
Check for duplicate method names in a list of tool specifications

agoncal:
## Issue

When a class has several annotated methods with `@Tools` that have the same name (but different parameters), the model does not know which one to invoke. So the idea if to check for method name duplicates and throw an exception.

## Change

So now, the `toolSpecificationsFrom` method invokes a  `validateSpecifications`. For now, this method only checks that there is no duplicate method names (but it could be used for other checks in the future if needed). The `validateSpecifications` method returns void if the list of tool specifications is valid. If not, it throws an `IllegalArgumentException`.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
@@ -1,16 +1,37 @@
 package dev.langchain4j.agent.tool;
 
-import dev.langchain4j.model.output.structured.Description;
-
-import java.lang.reflect.*;
-import java.math.BigDecimal;
-import java.math.BigInteger;
-import java.util.*;
-
-import static dev.langchain4j.agent.tool.JsonSchemaProperty.*;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.ARRAY;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.BOOLEAN;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.INTEGER;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.NUMBER;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.OBJECT;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.STRING;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.description;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.enums;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.from;
+import static dev.langchain4j.agent.tool.JsonSchemaProperty.items;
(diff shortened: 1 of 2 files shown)


langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
```diff
@@ -44,6 +67,23 @@ public static List<ToolSpecification> toolSpecificationsFrom(Object objectWithTo
...
+     *
+     * @param toolSpecifications list of ToolSpecification to be validated.
+     */
+    public static void validateSpecifications(List<ToolSpecification> toolSpecifications) throws IllegalArgumentException {
+
+        // Checks for duplicates methods
+        Set<String> names = new HashSet<>();
+        for (ToolSpecification toolSpecification : toolSpecifications) {
+            if (!names.add(toolSpecification.name())) {
+                throw new IllegalArgumentException(format("Methods in Tools must be unique. The method '%s' appears several times", toolSpecification.name()));
```
langchain4j:
```suggestion
                throw new IllegalArgumentException(format("Tool names must be unique. The tool '%s' appears several times", toolSpecification.name()));
```
Method names are used as tool names by default, but they can be overriden:
```java
        @Tool(name = "a")
        public int duplicateMethod(String typeString) {
            return 42;
        }

        @Tool(name = "b")
        public int duplicateMethod(int typeInt) {
            return 42;
        }
```
So I would suggest changing the error message to better reflect the validation logic.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
```diff
@@ -44,6 +67,23 @@ public static List<ToolSpecification> toolSpecificationsFrom(Object objectWithTo
         return toolSpecificationsFrom(objectWithTools.getClass());
     }
 
+    /**
+     * Validates all the {@link ToolSpecification}s. The validation checks for duplicate method names.
+     * Throws {@link IllegalArgumentException} if validation fails
+     *
+     * @param toolSpecifications list of ToolSpecification to be validated.
+     */
+    public static void validateSpecifications(List<ToolSpecification> toolSpecifications) throws IllegalArgumentException {
```
langchain4j, line 76: This method is not used by `AiServices`, `toolSpecificationFrom(Method)` is.
I guess we will need to validate it there...

ghost (commented):
@agoncal thanks a lot for addressing this issue so fast!

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
agoncal: Yes, I had a look at it, but will double check with you where to do the validation. But that will be another PR

ghost (approved):
@agoncal thank you!

PR #1380, opened 2024-07-01
fix: Azure finishReasonFrom add TOOL_CALLS case

Kugaaa:
## Issue

null

## Change

`InternalAzureOpenAiHelper.finishReasonFrom` case `TOOL_CALLS` return `TOOL_EXECUTION`

## General checklist


langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -206,6 +214,46 @@ void should_call_function_with_no_argument(String deploymentName, String gptVers
         ToolExecutionRequest toolExecutionRequest = aiMessage.toolExecutionRequests().get(0);
         assertThat(toolExecutionRequest.name()).isEqualTo(toolName);
         assertThat(toolExecutionRequest.arguments()).isEqualTo("{}");
+        assertThat(response.finishReason()).isEqualTo(STOP);
```
langchain4j, line 197: Shouldn't is be `TOOL_EXECUTION`?

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -206,6 +214,46 @@ void should_call_function_with_no_argument(String deploymentName, String gptVers
...
+    }
+
+    @ParameterizedTest(name = "Deployment name {0} using {1}")
+    @CsvSource({
+            "gpt-4o,        gpt-4o"
+    })
+    void should_call_function_finish_reason_is_tool_calls(String deploymentName, String gptVersion) {
```
langchain4j, line 204: New test is not necessary, updating assertions in existing tests that test tools is enough

ghost (commented):
@Kugaaa thanks!

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
Kugaaa: See below.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
Kugaaa: I think it shows different behaviors here.
- if the tool must be executed, finish reason should be `STOP`
- if the tool is optional, finish reason maybe be `TOOL_EXECUTION`

This is the reason for I added `should_call_function_finish_reason_is_tool_calls`.

langchain4j: Hmm. Where do these rules come from? Who decides if the tool must be executed or optional?

Kugaaa: - `dev.langchain4j.model.chat.ChatLanguageModel#generate(java.util.List<dev.langchain4j.data.message.ChatMessage>, java.util.List<dev.langchain4j.agent.tool.ToolSpecification>)` 
- `dev.langchain4j.model.chat.ChatLanguageModel#generate(java.util.List<dev.langchain4j.data.message.ChatMessage>, dev.langchain4j.agent.tool.ToolSpecification)`
Did I misunderstand?

```java
/**
 * Represents a language model that has a chat interface.
 */
public interface ChatLanguageModel {

    ...

    /**
     * Generates a response from the model based on a list of messages and a list of tool specifications.
     * The response may either be a text message or a request to execute one of the specified tools.
     * Typically, the list contains messages in the following order:
     * System (optional) - User - AI - User - AI - User ...
     *
     * @param messages           A list of messages.
     * @param toolSpecifications A list of tools that the model is allowed to execute.
     *                           The model autonomously decides whether to use any of these tools.
     * @return The response generated by the model.
     * {@link AiMessage} can contain either a textual response or a request to execute one of the tools.
     */
    default Response<AiMessage> generate(List<ChatMessage> messages, List<ToolSpecification> toolSpecifications) {
        throw new IllegalArgumentException("Tools are currently not supported by this model");
    }

    /**
     * Generates a response from the model based on a list of messages and a single tool specification.
     * The model is forced to execute the specified tool.
     * Typically, the list contains messages in the following order:
     * System (optional) - User - AI - User - AI - User ...
     *
     * @param messages          A list of messages.
     * @param toolSpecification The specification of a tool that must be executed.
     *                          The model is forced to execute this tool.
     * @return The response generated by the model.
     * {@link AiMessage} contains a request to execute the specified tool.
     */
    default Response<AiMessage> generate(List<ChatMessage> messages, ToolSpecification toolSpecification) {
        throw new IllegalArgumentException("Tools are currently not supported by this model");
    }
}
```

Kugaaa: I discovered this during testing, but I haven't found relevant docs to explain it yet.

langchain4j: In my understanding if LLM decides to execute one or multiple tools, `AiMessage.toolExecutionRequests` will be populated with tool requests and `Response.finishReason` should be `TOOL_EXECUTION`

Kugaaa: Yes, at first I understood it the same way.
Perhaps more validation is needed and waiting for an explanatory document.

langchain4j: > waiting for an explanatory document

What do you mean?

Kugaaa: The test results are different from the doc, making this PR questionable...

langchain4j: @Kugaaa I am not sure I understand you.
Just in case: whether tool must be called or can be called has nothing to do with finish reason. If LLM decides to call a tool, finish reason should be `TOOL_EXECUTION`

Kugaaa: I may understand what you mean. 
Let me confirm it again: the concept about finish reason from LangChain4j is different between of LLM server（such as `com.azure.ai.openai.models.CompletionsFinishReason`）

For example, in the `dev.langchain4j.model.azure.AzureOpenAiChatModelIT#should_call_function_with_no_argument` test, Azure OpenAI returns `CompletionsFinishReason.STOPPED`. According to the above definition, finishReason should be `FinishReason.TOOL_EXECUTION`.

langchain4j: `CompletionsFinishReason` in Azure and `FinishReason` in LC4j are supposed to be the same concepts which should map t oeach other, ideally 1:1.

In `should_call_function_with_no_argument`, `Response.finishReason()` should return `TOOL_EXECUTION` because `AiMessage` contains tool execution request.

If `CompletionsFinishReason` is `STOPPED`, then it might be a bug on Azure side, because they also have `CompletionsFinishReason.TOOL_CALLS` which is more appropriate in this case. If there is a bug, we can still return `FinishReason.TOOL_EXECUTION` if we see that there are tool execution requests in `AiMessage`

wuxinwei:
@langchain4j  I got the same problem, plz merge this request

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
langchain4j: @Kugaaa I've just debugged and indeed Azure OpenAI returns "stop" finish reason for some reason... So, WDYT about this:

> If there is a bug, we can still return FinishReason.TOOL_EXECUTION if we see that there are tool execution requests in AiMessage

This should be easy to implement and will return a correct `FinishReason` which is much better and cleaner.

langchain4j: @Kugaaa could you please check my comment above? Thank you!

Kugaaa: I think no problem, consider `FinishReason` as a new definition

langchain4j: @Kugaaa I am not sure I understand what you mean by "new definition". Returning `FinishReason.TOOL_EXECUTION` when LLM responded with a tool call is how it always has been (see `OpenAiChatModelIT`). Azure OpenAI for some reason has a different behavior (I guess it is a bug)

Kugaaa: I agree～

dliubarskyi:
Fixed long time ago in other PR

PR #1385, opened 2024-07-02
(feat) add base graph feature  version :0.1

zhangdashuai11:
add graph-0.1 feature for lagnchain4j  
this version is base version It simply implements a basic graph structure
todo list:
  1.support async call
  2.graph builder to Makes it easier to construct graphs
  3.Align with langchain


langchain4j:
@zhangdashuai11 thanks a lot for your PR, but we are not looking for graphs at this point.
Moreover, there are already a few libs that do this:
- https://github.com/bsorrentino/langgraph4j
- https://github.com/czelabueno/langchain4j-workflow

PR #1389, opened 2024-07-02
#725: PR 1 / 4:  Addresses `EnumOutputParser` not being called

PrimosK:
## Issue

#725 PR 1 / 4: Addresses `EnumOutputParser` not being called.

## Change

Fixed `EnumOutputParser` not being called.

Introduced new unit test which makes sure that correct output parsers are being called (including the one above).

Note that I had to refactor `DefaultAiServices` by extracting logic for picking up the correct `OutputParser` based on the returned type in order to cover my changes with unit test.

## General checklist


ghost (approved):
@PrimosK good job, thank you!

PR #1353, opened 2024-06-24
Add experimental Tools Management in Ollama Model

humcqc:
## Issue

https://github.com/langchain4j/langchain4j/issues/318

## Change

- Add experimental Tools Management in Ollama Model:
    2 Modes: sequential and parallel, tested with llama3
- Add dynamic variables in tool execution
- Add ollama AI Service with tools IT
- fix: should automatically convert List<String> into String[]
- Use java 17 for Ollama, To check with @langchain4j
- refactor chatMemory usage in DefaultAIServices
- share ollama test classes

## General checklist

- [ ] There are no breaking changes
- [ ] I have added unit and integration tests for my change
- [ ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new model integration

- [ ] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #1394, opened 2024-07-02
automatically convert List<String> into String[]

humcqc:
## Issue

automatically convert List<String> into String[]

## Change

- [fix: should automatically convert List`<`String`>` into String[]](https://github.com/langchain4j/langchain4j/pull/1394/files#diff-d8935c746a9b58544db4c7e3a9303e693a4b2353001c3af5038db4586e3324f8L482)

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/agent/tool/DefaultToolExecutor.java
+++ b/langchain4j/src/main/java/dev/langchain4j/agent/tool/DefaultToolExecutor.java
@@ -10,6 +10,7 @@ import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.util.Collection;
 import java.util.Map;
 import java.util.Objects;
 
@@ -185,6 +186,15 @@ public class DefaultToolExecutor implements ToolExecutor {
                     getNonFractionalDoubleValue(argument, parameterName, parameterType)).toBigInteger();
         }
 
+        if (parameterType.isArray() && argument instanceof Collection) {
+            Class<?> type = parameterType.getComponentType();
+            if (type == String.class) {
+                return ((Collection<String>) argument).toArray(new String[0]);
+            }
+            // TODO: Consider full type coverage.
+        }
+
+
         String result  = Json.toJson(argument);
         return Json.fromJson(result, parameterType);
     }
@@ -194,6 +204,13 @@ public class DefaultToolExecutor implements ToolExecutor {
             String parameterName,
             Class<?> parameterType
     ) {
+        if (argument instanceof String) {
+            try {
+                return Double.parseDouble(argument.toString());
+            } catch (Exception e) {
+                // nothing, will be handled with bellow code
+            }
+        }
         if (!(argument instanceof Number)) {
             throw new IllegalArgumentException(String.format(
                     "Argument \"%s\" is not convertable to %s, got %s: <%s>",
--- a/langchain4j/src/test/java/dev/langchain4j/agent/tool/DefaultToolExecutorTest.java
+++ b/langchain4j/src/test/java/dev/langchain4j/agent/tool/DefaultToolExecutorTest.java
@@ -41,7 +41,9 @@ class DefaultToolExecutorTest implements WithAssertions {
             Short ShortP,
             ExampleEnum enumP,
             boolean booleanP,
-            Boolean BooleanP
+            Boolean BooleanP,
+            double double2P,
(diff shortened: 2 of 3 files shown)


humcqc:
Hi @langchain4j , this is the PR to handle LONG_OR_DOUBLE strategy discuss in https://github.com/langchain4j/langchain4j/pull/1353.

langchain4j:
Hi @humcqc, thank you!
The first change (long or double) seems to be too risky and I am not sure I understand the upside of it, could you please elaborate? Why is this change required?
The second and third changes are great though!

humcqc:
> Hi @humcqc, thank you! The first change (long or double) seems to be too risky and I am not sure I understand the upside of it, could you please elaborate? Why is this change required? The second and third changes are great though!

The first one is to fix [this](https://github.com/langchain4j/langchain4j/pull/1394/files#diff-d8935c746a9b58544db4c7e3a9303e693a4b2353001c3af5038db4586e3324f8L429) ( that is the second point )
It avoid double conversion , from json to double and double to INT or Long.
And avoid to test int value (2) with double value (2.0)

I understand it's risky but it seems we are doing weird int to double conversion and then double to int . IT seems not good when you are using tools with int parameters and you receive from Json doubles.

I feel it risky too but i think it's the good way to go. 
How could we verify that we are not breaking things related to this change ?

humcqc:
Perhaps we can wait when we will move to jackson for this.
I can revert this and keep the test disabled and just fix the third point.

langchain4j:
> The first one is to fix [this](https://github.com/langchain4j/langchain4j/pull/1394/files#diff-d8935c746a9b58544db4c7e3a9303e693a4b2353001c3af5038db4586e3324f8L429) ( that is the second point )
> It avoid double conversion, from json to double and double to INT or Long.

I am not sure this will fix `should_use_tool_with_List_of_Integers_parameter` as Gson is deserializing it into a `List<Long>` in this case, not `List<Integer>`... Same issue with the `List<Float>`...

> I understand it's risky but it seems we are doing weird int to double conversion and then double to int . IT seems not good when you are using tools with int parameters and you receive from Json doubles.

For singular (not collection/array) `@Tool` method parameters we handle types properly in `DefaultToolExecutor.coerceArgument`, so the problem is only with collections/arrays, which we can fix in a similar way to `coerceArgument`.

Since this change 
- is risky
- does not fix the problem completely
- we are migriing to Jackson

, let's indeed revert it

Thank you!

humcqc:
Agreed! I will close this PR and handle the 3rd point in the original PR if you are ok.

langchain4j:
@humcqc, it would be great if this change can stay in a separate (this) PR. But it is not critical

humcqc:
> @humcqc, it would be great if this change can stay in a separate (this) PR. But it is not critical

Done!

langchain4j/src/main/java/dev/langchain4j/agent/tool/DefaultToolExecutor.java
```diff
@@ -194,6 +204,13 @@ private static double getDoubleValue(
             String parameterName,
             Class<?> parameterType
     ) {
+        if (argument instanceof String) {
```
langchain4j, line 207: This change is not required any more, right?

humcqc: I've put this one because some times llm respond with string instead of double when I was testing ollama tools.
Do you mean you fix it in another way ?

humcqc: I can move it to Ollama tools PR if you prefer ?

langchain4j: Ah, I see. No, it is ok to keep it here. Perharps add a test?

humcqc: sure, done! it was tested before in my Ollama IT , but it's better to have it in the unit test.

ghost (approved):
@humcqc thanks a lot!

humcqc:
You welcome, if you have time we can try to tackle this one: https://github.com/langchain4j/langchain4j/pull/1353 it will help me a lot in my project the time ollama implement it.

langchain4j:
@humcqc do you know if ollama has some timelines for implementing tools natively?

humcqc:
I've just [asked](https://github.com/ollama/ollama/issues/4386).

PR #1398, opened 2024-07-03
#725: PR 3 / 4: Make enum values work with @Description annotation

PrimosK:
## Issue 

https://github.com/langchain4j/langchain4j/issues/725

PR 3 / 4: Makes it possible to use @Description on Enum values

Side notes:
1. This PR is dependant on PR [1 / 4.](https://github.com/langchain4j/langchain4j/pull/1389)
2. List<Enum> is still not supported - this will be tackled in [PR 4 / 4](https://github.com/langchain4j/langchain4j/pull/1410)

## Change

Made it possible to annotate Enum values with @Description.

This is how prompt looks like for enums without descriptions:
```
Analyze sentiment of:
|||This LaptopPro X15 is wicked fast and that 4K screen is a dream.|||
You must answer strictly with one of these enums:
POSITIVE
NEUTRAL
NEGATIVE
```

This is how prompt looks like for enums with descriptions:
```
Analyze weather forecast for:
|||It will be cloudy and mostly rainy. No more rain early in the day but the sky remains overcast. Afternoon it is mostly cloudy. The sun will not be visible. The forecast has a moderate, 40% chance of Precipitation. Temperatures peaking at 17 °C.|||
You must answer strictly with one of these enums:
SUNNY - A clear day with bright sunlight and few or no clouds
CLOUDY - The sky is covered with clouds with no rain, often creating a gray and overcast appearance
RAINY - Precipitation in the form of rain, with cloudy skies and wet conditions
SNOWY - Snowfall occurs, covering the ground in white and creating cold, wintry conditions
```

Covered everything with unit and integrationt tests.

Tested mainly with ˙gpt-3.5-turbo˙.

## General checklist

- [X] There are no breaking changes
- [X] I have added unit and integration tests for my change
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [X] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

PR #1410, opened 2024-07-04
#725: PR 4 / 4: Make it possible to return list of enum values annotated with @Description annotation

PrimosK:
## Issue
https://github.com/langchain4j/langchain4j/issues/725

PR 4 / 4: Makes it possible to return list of enum values with @Description annotation

Side notes:
This PR is dependant on [PR 3 / 4](https://github.com/langchain4j/langchain4j/pull/1398).

## Change

Introduced dedicated parsers for lists and sets of enum values annotated with @Description annotation.

 Few additional output parsers were introduced:
  - `EnumListOutputParser` for list of enums.
  - `EnumSetOutputParser` for set of enums.
... but also:
  - `StringListOutputParser` for list of strings.
  - `StringSetOutputParser` for set of strings.
... to make it consistent.

Testing:
All changes are covered with unit and integration tests inside:
- `ServiceOutputParserTest`
- `AiServicesIT`

Refactoring:
As part of this work, the `TypeUtils` class was introduced to simplify logic inside:
  - `DefaultAiServices.java`
  - `ServiceOutputParser.java`

Note:
I only refactored the parts that were necessary to implement the new feature

## General checklist


langchain4j/src/main/java/dev/langchain4j/model/output/StringSetOutputParser.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import static java.util.Arrays.asList;
+
+public class StringSetOutputParser extends CollectionOutputParser<Set<String>> {
+
+    @Override
+    public Set<String> parse(String text) {
+        return new HashSet<>(asList(text.split("\n")));
```
langchain4j:
```suggestion
        return new LinkedHashSet<>(asList(text.split("\n")));

```
To keep the original order of items in LLM response intact

langchain4j/src/main/java/dev/langchain4j/model/output/EnumSetOutputParser.java
```diff
@@ -0,0 +1,22 @@
...
+public class EnumSetOutputParser extends EnumCollectionOutputParser<Enum> {
+
+    public EnumSetOutputParser(Class<? extends Enum> enumClass) {
+        super(enumClass);
+    }
+
+    @Override
+    public Set<Enum> parse(String text) {
+        List<String> stringsList = asList(text.split("\n"));
+        return stringsList.stream().map(enumOutputParser::parse).collect(Collectors.toSet());
```
langchain4j: Please make sure the order of enums in returned `Set` is the same as in LLM response

PrimosK: I've made it the same as initially was to keep the scope of required changes at the minimum. 

But I full agree that makes sense ... Change applied.

langchain4j/src/main/java/dev/langchain4j/model/output/StringSetOutputParser.java
PrimosK: I've made it the same as initially was to keep the scope of required changes at the minimum. 

But I full agree that makes sense ... Change applied.

PrimosK:
Changes based on the review feedback were applied. Also resolved conflicts. 

Ready for second pass review ...

langchain4j/src/main/java/dev/langchain4j/model/output/EnumSetOutputParser.java
langchain4j: @PrimosK thanks!

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
```diff
@@ -64,6 +64,26 @@ void afterEach() {
...
+    }
+
+    @Test
+    void test_simple_instruction_with_primitive_return_type() {
+        EggCounter eggCounter = AiServices.create(EggCounter.class, chatLanguageModel);
+
+        String sentence = "I have ten eggs in my basket and three in my pocket.";
+
+        int count = eggCounter.count(sentence);
+        System.out.println(count);
```
langchain4j: minor: please add an assertion on `count` to make sure LLM understands the prompt

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
```diff
@@ -218,6 +238,86 @@ void test_extract_enum_with_description() {
...
+    }
+
+    @Test
+    void test_extract_list_of_enums() {
+        IngredientsExtractor ingredientsExtractor = AiServices.create(IngredientsExtractor.class, chatLanguageModel);
+
+        String recipe = "Just mix some salt, pepper and oil in the bowl. That will be a basis for...";
+
+        List<Ingredient> ingredients = ingredientsExtractor.extractIngredients(recipe);
+        System.out.println(ingredients);
```
langchain4j: minor: please add an assertion on `ingredients` to make sure LLM understands the prompt

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
```diff
@@ -218,6 +238,86 @@ void test_extract_enum_with_description() {
...
+
+        String review = "Our stay at hotel was a mixed experience. The location was perfect, just a stone's throw away " +
+                "from the beach, which made our daily outings very convenient. The rooms were spacious and well-decorated, " +
+                "providing a comfortable and pleasant environment. However, we encountered several issues during our " +
+                "stay. The air conditioning in our room was not functioning properly, making the nights quite uncomfortable. " +
+                "Additionally, the room service was slow, and we had to call multiple times to get extra towels. Despite the " +
+                "friendly staff and enjoyable breakfast buffet, these issues significantly impacted our stay.";
+
+        List<IssueCategory> issueCategories = hotelReviewIssueAnalyzer.analyzeReview(review);
+        System.out.println(issueCategories); // [MAINTENANCE_ISSUE, SERVICE_ISSUE, COMFORT_ISSUE, OVERALL_EXPERIENCE_ISSUE]
```
langchain4j: same here

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
```diff
@@ -790,6 +890,54 @@ void should_throw_exception_when_retrieve_sources_and_generic_type_is_not_set()
...
+        assertThatThrownBy(() ->
+                AiServices.create(InvalidAssistantWithList.class, chatLanguageModel))
+                .isExactlyInstanceOf(IllegalArgumentException.class)
+                .hasMessage("The return type 'Result' of the method 'answerWithNoGenericType' must be " +
+                        "parameterized with a concrete type, for example: Result<String> or Result<MyCustomPojo>");
+    }
+
+    interface InvalidAssistantWithSet {
+
+        List answerWithNoGenericType(String query);
```
langchain4j:
```suggestion
        Set answerWithNoGenericType(String query);
```

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -69,8 +70,10 @@ public T build() {
                 throw illegalConfiguration("The @Moderate annotation is present, but the moderationModel is not set up. " +
                         "Please ensure a valid moderationModel is configured before using the @Moderate annotation.");
             }
-            if (method.getReturnType() == Result.class) {
-                validateResultReturnType(method);
+            if (method.getReturnType() == Result.class ||
+                method.getReturnType() == List.class ||
```
langchain4j, line 74: I guess the validation error message for `List` and `Set` without concrete types should mention `List` and `Set` instead of `Result`

langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
```diff
@@ -22,67 +22,81 @@ public ServiceOutputParser(OutputParserFactory outputParserFactory) {
...
-        Optional<OutputParser<?>> outputParser = outputParserFactory.get(returnType);
+        Optional<OutputParser<?>> outputParser = outputParserFactory.get(rawClass, typeArgumentClass);
         if (outputParser.isPresent()) {
             String formatInstructions = outputParser.get().formatInstructions();
-            return "\nYou must answer strictly in the following format: " + formatInstructions;
-        }
 
-        if (returnType == List.class || returnType == Set.class) {
-            return "\nYou must put every item on a separate line.";
+            if (rawClass == List.class ||
```
langchain4j, line 88: Perharps we should duplicate `\nYou must answer strictly in the following format: `message in all output parsers to make it more uniform and avoid this brittle `if` condition, WDYT?

langchain4j/src/main/java/dev/langchain4j/service/TypeUtils.java
```diff
@@ -0,0 +1,71 @@
+package dev.langchain4j.service;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+
+public class TypeUtils {
```
langchain4j: Could you please cover methods in this class with unit tests?

langchain4j/src/main/java/dev/langchain4j/service/ServiceOutputParser.java
PrimosK: Actually I was already thinknig about that.

That indeed makes sense as it's concrete output parser's responsibility to construct entire instruction (for itself) on it's own (especially as there can differ from parser to parser). That would consequently further simplify `ServiceOutputParser`.

It's just I wasn't sure this is something to be done as part of this PR as scope of changes was quite large already...

If you think it's fine to do that as part of this PR then I can look further into it.

langchain4j: You're right, let's not increase the scope of this PR, we can address it in a separate one

langchain4j/src/main/java/dev/langchain4j/service/TypeUtils.java
PrimosK: Done...

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
PrimosK: Right.. I've made these dynamically generated which should improve the UX (for the end users). Also moved few more type related methods from  `AiServices` to `TypeUtils` and covered everything with unit tests.

ghost (approved):
@PrimosK great job, thank you!

PR #1416, opened 2024-07-06
Mistral AI - Make Client configurable via ChatModel

ThomasVitale:
## Change
The `MistralAiClient` provides a convenient way to provide a custom implementation of the underlying HTTP interaction. Using the `MistralAiClientBuilderFactory` SPI, it's possible to specify a custom static `Builder` for a `MistralAiClient`.

Since an instance of `MistralAiClient` is built internally from that static builder when building a `MistralChatClient` object, it's not possible to provide further customizations. For example, it's not possible to include additional configuration parameters to `MistralAiClient`, even though the Builder SPI would actually allow that. It's also not possible to initialise the `MistralAiClient` instance via dependency injection, resulting in not being able to manage it properly from a Spring application context.

This PR aims at solving that problem by making it possible to provide a `MistralAiClient` instance when building a  `MistralChatClient` object. The additional benefit is making it clearer what parameters are specific to the model integration (like model and temperature) and what are the ones specific to the HTTP client (like timeout and API key).

The change extends the current setup to avoid any breaking change. It also marks the previous way of configuring a Client deprecated.

## General checklist

### DIFF
--- a/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiChatModel.java
+++ b/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiChatModel.java
@@ -45,27 +45,29 @@ public class MistralAiChatModel implements ChatLanguageModel {
     /**
      * Constructs a MistralAiChatModel with the specified parameters.
      *
-     * @param baseUrl      the base URL of the Mistral AI API. It uses the default value if not specified
-     * @param apiKey       the API key for authentication
-     * @param modelName    the name of the Mistral AI model to use
-     * @param temperature  the temperature parameter for generating chat responses
-     * @param topP         the top-p parameter for generating chat responses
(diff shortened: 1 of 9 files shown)


geoand:
+1 for this

langchain4j:
Hi @ThomasVitale, thank you for the PR!

And I am sorry for not commenting on this in time! Because this PR is a foundation for https://github.com/langchain4j/langchain4j-spring/pull/32, and I had questions there regarding the approach (e.g. allignment with https://github.com/langchain4j/langchain4j/pull/1103), I was waiting for https://github.com/langchain4j/langchain4j-spring/pull/32 to settle before going forward with this PR, because it affects end users (parameter deprecations and a new way of creating `MistralAiChatModel`), and I did not want to bother users with constantly changing API.

Please ping me once you'll have some time, I am happy to discuss!

dliubarskyi:
Closing it for the same reason as described in https://github.com/langchain4j/langchain4j-spring/pull/32#issuecomment-2629570318

PR #1427, opened 2024-07-09
Keeping track of textSegment startIndex when splitting documents hierarchically

### LINKED ISSUE #1424: [FEATURE] HierarchicalDocumentSplitter to return the start index of each TextSegment
I am trying to replicate functionality existing in the python langchain package which provides a `start_index` for each textSegment. 

https://api.python.langchain.com/en/latest/base/langchain_text_splitters.base.TextSplitter.html#langchain_text_splitters.base.TextSplitter 

Currently, langchain4j will give back the index of the textSegment - but this is only so helpful. 

Furthermore, one could calculate the start_indexes with the results, but this becomes much easier if you know the exact text that has gone into any overlap. Otherwise, there is some sort of overlap search that needs to happen which feels a bit gross

To mimic the python library with an input paramter when creating the textSplitter to add_start_index to the metadata of the TextSegment

Post calculation, but again, it gets messy with overlap being variable in length.

bracl:
## Issue

Closes #1424 

## Change

Whilst splitting text, the `startIndex` of each `textSegment` is kept track of. The `startIndex` is added to the `metadata` of each split in the same way that is done in the python repo. 

This is done by initialising a mutable variable that is updated whenever segments are added. When segments have overlap, the startIndex variable is updated accordingly.

The test that has been added:
- take a string
- remove all duplicate words
- make note of each words starting index within the string
- split the document
- for each segment, assert that the startIndex is the same startIndex as noted earlier in the test.

## General checklist

** I can run unit tests for the langchain4j module which is where all of my changes are. I get errors complaining about an `openAiApiKey cannot be null or empty` but i assume that's because I havent set that up. But I believe that none of my changes should impact these tests **

** I have not explicitly run any integration tests if they are different from unit tests? **


ocoulson (commented):
Looks like a useful change! We use these start indices in the python version to analyse the results of vector embedding searches, to compare expected sections of a document from the actual chunks returned by the search. 

The lack of this in the 4j lib is a blocker on us moving away from python to be able to use jvm languages

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/HierarchicalDocumentSplitter.java
```diff
@@ -28,11 +28,13 @@ public abstract class HierarchicalDocumentSplitter implements DocumentSplitter {
             new DocumentBySentenceSplitter(1, 0, null, null);
 
     private static final String INDEX = "index";
+    private static final String START_INDEX = "startIndex";
```
langchain4j:
```suggestion
    private static final String START_INDEX = "start_index";
```
To keep it in sync with the rest of metadata fields (e.g. "file_name", "absolute_directory_path", etc).

I am not sure about `start_index` name though, it might be not clear what is the difference/relationshop between `start_index` and `index`. Did you consider other options?

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/HierarchicalDocumentSplitter.java
```diff
@@ -28,11 +28,13 @@ public abstract class HierarchicalDocumentSplitter implements DocumentSplitter {
...
 
     protected final int maxSegmentSize;
     protected final int maxOverlapSize;
     protected final Tokenizer tokenizer;
     protected final DocumentSplitter subSplitter;
+    protected final AtomicInteger startIndex = new AtomicInteger(0);
```
langchain4j: What about keeping this as a local variable instead of an object field?

langchain4j/src/test/java/dev/langchain4j/data/document/splitter/DocumentByParagraphSplitterTest.java
```diff
@@ -53,8 +53,8 @@ void should_split_into_segments_with_one_paragraph_per_segment(String separator)
         segments.forEach(segment ->
                 assertThat(segment.text().length()).isLessThanOrEqualTo(maxSegmentSize));
         assertThat(segments).containsExactly(
-                textSegment(firstParagraph, metadata("index", "0").put("document", "0")),
-                textSegment(secondParagraph, metadata("index", "1").put("document", "0"))
+                textSegment(firstParagraph, metadata("index", "0").put("document", "0").put("startIndex", "0")),
+                textSegment(secondParagraph, metadata("index", "1").put("document", "0").put("startIndex", "27"))
```
langchain4j: Is this supposed to be a start index in the original document? Whitespaces seems to be ignored

ghost (commented):
@bracl thank you!

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/HierarchicalDocumentSplitter.java
bracl: The reason for start_index was to mimic the python library. Happy to change it to whatever you prefer however.

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/HierarchicalDocumentSplitter.java
bracl: Ahh, I was moving this field around during development because i thought it was getting confused. I can try putting it back inside the function.

langchain4j/src/test/java/dev/langchain4j/data/document/splitter/DocumentByParagraphSplitterTest.java
bracl: Yes, the start index should be the index of the first character as it appears in the outputted textSegments. 

However (and this might break the whole idea of it) white space is sort of normalised to one character between all words because of how splitting and then buliding up parts works. Essentially, if you have the string `"some       text"` before the split, after the splitting and segmenting, it will appear like `"some text"`. 

But, as far as I have tested, the single whitespace between words is respected by the currnet implementation of the start index.

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/HierarchicalDocumentSplitter.java
bracl: Have changed this - and good spot :)

langchain4j/src/test/java/dev/langchain4j/data/document/splitter/DocumentByParagraphSplitterTest.java
langchain4j: Is there a way to respect the original formatting?

bracl: This would be a change to the underlying splitting logic. This PR makes no logical change to the output, just adds some metadata. If you want the whitespace logic to change, I would suggest that gets done in a separate issue/PR?

langchain4j: @bracl I meant to respect the original formatting only in the index calculation. No need to do any changes to the splitting logic

bracl: This start_index logic works by keeping track of an index and moving it for two reasons:
1. a new segment is created, and the index increases
2. an overlap is created, and the index decreases

In 1. the amount of increase is determined by the size of the `part`. This part does not have the original whitespace of the input text. 
In 2. any overlap is generated from the `part`, which again, does not have the original whitespace from the input text.

Basically, that information is lost as soon as we perform `String[] parts = split(document.text());`

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/HierarchicalDocumentSplitter.java
langchain4j: What about `char_index_start` or `char_start_index` or some combination of these 3 words?

langchain4j/src/test/java/dev/langchain4j/data/document/splitter/DocumentByParagraphSplitterTest.java
langchain4j: Can't we take the first part of the new segment (or overlap, whatever comes first) and do something like `document.text().indexOf(firstPart)`?

bracl: Doing some testing on this, and im not sure it holds when the original document has characters other than spaces. `\n \r \t` characters are lost in the splitting, so then `indexOf` is unable to find the place in the text

langchain4j: I've put `System.out.println(">>" + document.text().indexOf(part));` on `HierarchicalDocumentSplitter.java:118` and it seems to work. (This should be called only for the first segment in the segment)

bracl: I think unfortunately going down the route of using `.indexOf` doesnt hold in all cases. 

Consider following test
```
    @Test
    public void test_start_indexes_with_lots_of_whitespace() {
        {
            DocumentSplitter splitter = DocumentSplitters.recursive(15, 10);

            List<TextSegment> textSegments = splitter.split(Document.from("aaaa aaaa aaaa\t\r\r\r\t\taaaa\r\r\rbbbb aaaa"));
            assertThat(textSegments).containsExactly(
                    textSegment("aaaa aaaa aaaa", metadata("index", "0").put("document", "0").put("char_start_index", "0")),
                    textSegment("aaaa\r\r\rbbbb aaaa", metadata("index", "1").put("document", "0").put("char_start_index", "20")) // This 20 comes from manually counting `aaaa aaaa aaaa\t\r\r\r\t\t`. 5 + 5 + 4 + 6.
            );
        }
    }
```

Running this test against the code in the branch as it stands results in this
image

If I use `indexOf` on the whole segmentText, I get this
image
We can observe that the StringBuilder is replacing the original text with the `joinSeparator` when we build the string.

If I update to start using `indexOf` to get the indexes of the first part in the chunk, then I get this
image
We are finding the repeated word from earlier in the document. 

So, `indexOf` has a second parameter `fromIndex` that can be used to mark a place to start searching in the string. This would help us with duplicate text, however, it means that we would have to be keeping track of the current index throughout the whole document text as a "where to start from"... which is exactly what the PR is currently doing.

Okay, so im now thinking about changing the logic to update the index based on the original text and not the output of the stringBuilders. The problem with this is that when we go into a recursive splitter, we dont retain any of the information when it returns. Ultimately, because the stringBuilder doesnt know the original splitting character and replaces it with the joinSeparator we have issues. 

I think maybe I should be looking at the python library to see what they are doing? But I also feel like the structure that they have is actually quite different...? 

Thoughts? I am still very keen on trying to find a solution that satisfies as much as possible - i just worry that we are going to have to compromise somewhere...

langchain4j: Yes, python implementation is different.

Well, if there is no easy way to do it, let's then keep it as it is now? But will this work for you then? Since one can't really rely on this...

bracl: One can rely on the numbers _**if**_ the text is white-space-normalised on input. Which I agree is not something to be relied upon. 

What kind of documentation or caveats would we have to put in place to introduce something like this. 

Also, I understand if this is too far away from something that is completely correct and therefore not something you are comfortable with merging. I dont want this to be forced through for the sake of forcing it through

langchain4j: TBH I am not convinced this change as of now is ready to be merged, considering it does not work as expected.
If you feel like you will benefit from this feature, let's keep it, but add a flag (default=disabled) to explicitly enable it and document the known issues in javadoc.

bracl: Yep, that sounds reasonable. We can make it explicitly opt-in. 

I will update

PR #1431, opened 2024-07-10
[Doc] Refactor the Azure documentation

agoncal:
I'm refactoring the Azure documentation so the samples of creating Embedding and Image models are moved to their own pages.


ghost (approved):
@agoncal thank you!

PR #1448, opened 2024-07-11
EmbeddingModelTextClassifier to embed all samples per labels at once

### LINKED ISSUE #1447: [BUG] EmbeddingModelTextClassifier should batch embedding calculations
**Describe the bug**

_This is more of an improvement than a bug._

Currently, `EmbeddingModelTextClassifier` calculates the vector embeddings for each sample of each label one at a time.
This can be particularly slow when calling a cloud-based embedding model.
The idea would be to use the embedding model's `embedAll()` method, instead of `embed()` to calculate all embeddings for each label at once.

- LangChain4j version: 0.32.0
- LLM(s) used: any
- Java version: any

glaforge:
Fixes #1447


ghost (approved):
@glaforge thank you!

PR #1449, opened 2024-07-11
Add possibility to shutdown

koppor:
When our app quits, we have following thing in [VisualVM](https://visualvm.github.io/):

image

I searched for ways to shutdown the task runner.

There seem to be discussions in OKHttp for shutdown "support" (https://github.com/square/okhttp/issues/6702, https://github.com/square/okhttp/issues/6173). When checking the code, there is something like `shutdown()` already existing: `dev.ai4j.openai4j.DefaultOpenAiClient#shutdown` 

The client is generated in `dev.langchain4j.model.openai.OpenAiChatModel#OpenAiChatModel` "automagically". On the one hand, this is cool, because a user does not need to take care about the client. On the other hand, the client's `shtudown()` method cannot be called.

I needed to decied between:

1. Change the constructor to accept an ´OpenAiClient` so that the caller has full control on the client.
2. Add a `shutdown` to the model to enable the model to shutdown all used services.

I opted for the latter to have more "lightweight" interface.

I additionally needed to decide to switch to [AutoClosable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html), because `close()` is IMHO the "modern" Java way for shutting down things. -- However, I kept `shutdown ` to focus at one issue at a time.

Curious, what you think.

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
@@ -116,6 +116,13 @@ public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
         this.listeners = listeners == null ? emptyList() : new ArrayList<>(listeners);
     }
 
+    /**
+     * Shuts down the used client
+     */
+    public void shutdown() {
+        client.shutdown();
+    }
+
     public String modelName() {
         return modelName;
     }


koppor:
Try locally. Does not help to solve our issue. Therefore closing. Will reopen if I know more.

langchain4j:
Hi @koppor this is a known issue, should be fixed once we migrate away from okhttp

langchain4j:
Although we can have a temporary solution as you suggested. Did you manage to find out what is wrong with these changes?

koppor:
> Although we can have a temporary solution as you suggested. Did you manage to find out what is wrong with these changes?

I'm on it. Currently, this change seems to help to reduce shutdown time. We still have one thread opened. I am investigating.

In general, I tend to like this connection pooling concept. Useful for long running apps and servers only.

koppor:
Regarding HTTP clients, I got following options:

1. Use JDKs (because 1 less dependency)
2. Use [feign](https://github.com/OpenFeign/feign): One can select the HTTP Client implementation (see https://github.com/OpenFeign/feign?tab=readme-ov-file#client)
3. Use Apache HTTTP Client: https://hc.apache.org/httpcomponents-client-5.3.x/index.html (Based on the evaluation matrix: https://github.com/docker-java/docker-java/blob/main/docs/transports.md)
4. Use [Riptide](https://github.com/zalando/riptide?tab=readme-ov-file#features) if one needs asynchronous handling (which I think, isn't needed, is it?). However, one would need to contribute to https://github.com/zalando/riptide/issues/837 (otherwise, a Spring dependency is added)

IMHO, the Apache HTTP Client is the most stable.

(I can post this to the issue where you discuss clients. I just did not see it)

langchain4j:
@koppor do you see the problem when using `OpenAiChatModel` or `OpenAiStreamingChatModel`?

I see it when using `OpenAiStreamingChatModel`, the app hangs for 1 minute after main thread is done.
I've added `OpenAiStreamingChatModel.shutdown()` locally that calls `client.shutdown()` and it works ok.

koppor:
> @koppor do you see the problem when using `OpenAiChatModel` or `OpenAiStreamingChatModel`?

OpenAiChatModel. Same code as at https://github.com/langchain4j/langchain4j/issues/1454#issuecomment-2230953662.

(@InAnYan and me are working together in the context of Google Summer of Code)

PR #1475, opened 2024-07-16
[Feature] ApachePdfBoxDocumentParser should return metadata

### LINKED ISSUE #1406: [FEATURE] ApachePdfBoxDocumentParser should return metadata extracted from the PDF
Try to get metadata from a perfect accessable pdf that contains metadata. 

Here is such a pdf attached
[bookmarklet-einrichten.pdf](https://github.com/user-attachments/files/16090155/bookmarklet-einrichten.pdf)

I expect a Metada Object

```
       ApachePdfBoxDocumentParser parser = new ApachePdfBoxDocumentParser();
        InputStream inputStream = new FileInputStream("bookmarklet-einrichten.pdf");
        Document document = parser.parse(inputStream);
        Metadata metadata = document.metadata();
        System.out.println(metadata.toMap().isEmpty());

```
Above Code prints true, I also debugged alot of PDFs, its always empty. I also investigate the parse Method. As I see you never extract any metadata.

**Expected behavior**
Metadata of parsed pdf.

- LangChain4j version: 0.31.0
- LLM(s) used: none
- Java version: 21
- Spring Boot version (if applicable): none

zambrinf:
## Issue

Closes #1406 

## Change
Added metadata to Document returned by ApachePdfBoxDocumentParser.

## General checklist


document-parsers/langchain4j-document-parser-apache-pdfbox/src/main/java/dev/langchain4j/data/document/parser/apache/pdfbox/ApachePdfBoxDocumentParser.java
```diff
@@ -18,19 +20,25 @@ public class ApachePdfBoxDocumentParser implements DocumentParser {
...
-
-            return Document.from(text);
+            Metadata metadata = toMetadata(pdfDocument.getDocumentInformation());
+            return Document.from(text, metadata);
         } catch (IOException e) {
             throw new RuntimeException(e);
         }
     }
+
+    private Metadata toMetadata(PDDocumentInformation documentInformation) {
```
langchain4j: I am wondering if we should make inclusion of metadata configurable and don't include all metadata fields by default.
It might contain a lot of useless information and will just pollute the storage.
WDYT?

zambrinf: Maybe keep the existing method as is, and add a method to DocumentParser which has a second parameter `Set<String> metadataKeys` to be retrieved? What if the user wants everything because he doesn't know what is available?

langchain4j: With new method it would be complicated because you would need to modify all parsers and loaders.
What about adding a ctor `ApachePdfBoxDocumentParser(boolean includeMetadata)` instead?
User can filter out needed keys after parsing. But by default `includeMetadata` should be `false` to avoid pollution.

ghost (approved):
@zambrinf thank you!

PR #1487, opened 2024-07-18
Adding support for Amazon Bedrock Titan Embedding Model V2 (amazon.titan-embed-text-v2:0)

### LINKED ISSUE #1486: [FEATURE] Support for AmazonBedrock Titan Embeddings V2
Would like to enhance Amazon Bedrock Titan Model to support recently added Amazon Titan Text Embeddings V2

staleks:
## Issue

Closes #1486 

## Change
Added support for Amazon Bedrock Titan Embedding Model version 2 (amazon.titan-embed-text-v2:0)

## General checklist

- [  ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

### DIFF
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
@@ -40,11 +40,10 @@ public class BedrockTitanEmbeddingModel extends AbstractBedrockEmbeddingModel<Be
     /**
      * 1024 is default size of output vector for Titan Embedding model V2.
      */
-    @Builder.Default
-    private final Integer dimensions = 1024;
+    private final Integer dimensions;
 
     /**
-     *  A flag indicating whether or not to normalize the output embeddings.
+     *  A flag indicating whether to normalize the output embeddings.
      *  It defaults to true, which is optimal for RAG use cases.
      */
     @Builder.Default
@@ -58,19 +57,16 @@ public class BedrockTitanEmbeddingModel extends AbstractBedrockEmbeddingModel<Be
                     .map(TextSegment::text)
                     .map(text -> of("inputText", text))
                     .collect(Collectors.toList());
-        } else if (MODEL_V2_ID.equals(this.model)) {
-            List<Map<String, Object>> result = new ArrayList<>();
-            for (TextSegment textSegment : textSegments) {
-                Map<String, Object> parameters = new HashMap<>();
-                parameters.put("inputText", textSegment.text());
-                parameters.put("dimensions", dimensions);
-                parameters.put("normalize", normalize);
-                result.add(parameters);
-            }
-            return result;
-        } else {
-            throw new IllegalArgumentException("Unsupported model: " + this.model);
         }
+        List<Map<String, Object>> result = new ArrayList<>();
+        for (TextSegment textSegment : textSegments) {
+            Map<String, Object> parameters = new HashMap<>();
+            parameters.put("inputText", textSegment.text());
+            parameters.put("dimensions", dimensions);
+            parameters.put("normalize", normalize);
+            result.add(parameters);
+        }
+        return result;
     }
 
     @Override
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
@@ -46,8 +46,7 @@ public class BedrockTitanEmbeddingModel extends AbstractBedrockEmbeddingModel<Be
      *  A flag indicating whether to normalize the output embeddings.
      *  It defaults to true, which is optimal for RAG use cases.
      */
-    @Builder.Default
-    private final Boolean normalize = true;
+    private final Boolean normalize;
 
 
     @Override
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
@@ -53,6 +53,9 @@ public class BedrockTitanEmbeddingModel extends AbstractBedrockEmbeddingModel<Be
     @Override
     protected List<Map<String, Object>> getRequestParameters(List<TextSegment> textSegments) {
         if (MODEL_V1_ID.equals(this.model)) {
+            if (this.dimensions != null || this.normalize != null) {
+                throw new IllegalArgumentException("Dimensions and normalize are not supported for Titan Embedding model V1");
+            }
             return textSegments.stream()
                     .map(TextSegment::text)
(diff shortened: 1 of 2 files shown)


langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
```diff
@@ -26,12 +37,40 @@ protected String getModelId() {
...
-        return textSegments.stream()
-                .map(TextSegment::text)
-                .map(text -> of("inputText", text))
-                .collect(Collectors.toList());
+        if (MODEL_V1_ID.equals(this.model)) {
+            return textSegments.stream()
+                    .map(TextSegment::text)
+                    .map(text -> of("inputText", text))
+                    .collect(Collectors.toList());
+        } else if (MODEL_V2_ID.equals(this.model)) {
```
langchain4j: Is `MODEL_V2_ID` the only model that supports dimensions/normalize?
Would be nice to think about how to make it future-proof when new models will be added (they will most probably support these params).

BTW, if dimensions/normalize are not supported by the older model, it would be better to fail (either validate on our side or send them to AWS so that it can validate) in this case, otherwise user might be expecting for parameters to work, but they will be silently ignored

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
```diff
@@ -26,12 +37,40 @@ protected String getModelId() {
         return model;
     }
 
+    /**
+     * 1024 is default size of output vector for Titan Embedding model V2.
+     */
+    @Builder.Default
+    private final Integer dimensions = 1024;
```
langchain4j: I would avoid setting default values, they are already defined on AWS side, right?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
```diff
@@ -26,12 +37,40 @@ protected String getModelId() {
...
+            for (TextSegment textSegment : textSegments) {
+                Map<String, Object> parameters = new HashMap<>();
+                parameters.put("inputText", textSegment.text());
+                parameters.put("dimensions", dimensions);
+                parameters.put("normalize", normalize);
+                result.add(parameters);
+            }
+            return result;
+        } else {
+            throw new IllegalArgumentException("Unsupported model: " + this.model);
```
langchain4j: When the new model will come out, users will have to wait untill we add it explicitly and release the next version of LC4j to be able to use it...

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
```diff
@@ -6,17 +6,28 @@
...
 /**
- * Bedrock Amazon Titan embedding model
+ * Bedrock Amazon Titan embedding model with support for both versions:
+ * <ul>
+ *     <li>amazon.titan-embed-text-v1</li>
+ *     <li>amazon.titan-embed-text-v2:0</li>
+ * </ul>
+ *
+ * @link https://docs.aws.amazon.com/bedrock/latest/userguide/titan-embedding-models.html
+ * @link https://aws.amazon.com/blogs/aws/amazon-titan-text-v2-now-available-in-amazon-bedrock-optimized-for-improving-rag/
```
langchain4j, line 23: Thank you for adding javadoc 🙏

ghost (commented):
@staleks thanks a lot!

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
staleks: I just checked ... if we provide `dimensions` and `normalize` as parameters then AWS Bedrock throws 400 Exception.
example.
`Malformed input request: 2 schema violations found, please reformat your input and try again. (Service: BedrockRuntime, Status Code: 400, Request ID: b606929b-e642-4c34-8565-bc4e32f66510)
`
But I agree to all your comments.
I removed IllegalArgumentException and now just checking if V1 model is applied.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
staleks: Yes, correct. Good point. I'll remove it.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
staleks: Already addressed with one previous comment. Totally agree, we should think of V3, V4,etc.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockTitanEmbeddingModel.java
```diff
@@ -40,37 +40,35 @@ protected String getModelId() {
...
-    @Builder.Default
-    private final Boolean normalize = true;
+    private final Boolean normalize;
 
 
     @Override
     protected List<Map<String, Object>> getRequestParameters(List<TextSegment> textSegments) {
         if (MODEL_V1_ID.equals(this.model)) {
+            if (this.dimensions != null || this.normalize != null) {
+                throw new IllegalArgumentException("Dimensions and normalize are not supported for Titan Embedding model V1");
```
langchain4j, line 56: I would even remove this check (and check for `MODEL_V1_ID`) and just pass through all parameters that user set to the AWS and let it decide if it is legal or not.

I can imagine at some point `MODEL_V1_ID` might allow setting this parameters and the logic on our side will eb outdated.

staleks: If we remove this check too, then it will fail at AWS side. Remember?
`Malformed input request: 2 schema violations found, please reformat your input and try again. (Service: BedrockRuntime, Status Code: 400, Request ID: b606929b-e642-4c34-8565-bc4e32f66510)`

And one of your comments was that we should fail on "our" (langchain4j) side.

So I'm OK to implement either way, just let's decide.
Thank You

staleks: Any news on this? How do you want to proceed?

langchain4j: Let's keep it as it is now, thank you!

ghost (approved):
@staleks thanks a lot!

PR #1491, opened 2024-07-19
[Feature] Allow varargs with multiple parameters 

zambrinf:
## Issue

N/A

## Change
Allow varargs with multiple parameters in AiServices interfaces, fixes the bug below.
```
interface Chef {
  @SystemMessage("You are very {{character}} chef")
  @UserMessage("Create a recipe that can be prepared using only {{items}}")
  Recipe createRecipeFromVarargs(@V("character") String character, @V("items") String... ingredients);
}
// FAILS before this PR, returns: "Create a recipe that can be prepared using only [Ljava.lang.String;@20d3d15a”
// Now: "Create a recipe that can be prepared using only [rice, beans]"
```

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/model/input/DefaultPromptTemplateFactory.java
```diff
@@ -63,7 +64,10 @@ private static String replaceAll(String template, String variable, Object value)
             if (value == null || value.toString() == null) {
                 throw illegalArgument("Value for the variable '%s' is null", variable);
             }
-            return template.replace(inDoubleCurlyBrackets(variable), value.toString());
+            String replacement = value instanceof Object[]
```
langchain4j, line 67: I am not sure if it should be fixed here.
If you check `Chef.createRecipeFrom(String... ingredients)`, it worked fine.
Please check `DefaultAiServices.findTemplateVariables`, it works differently for `Chef.createRecipeFrom` and `Chef.createRecipeFromVarargs`

ghost (commented):
@zambrinf thank you!

langchain4j-core/src/main/java/dev/langchain4j/model/input/DefaultPromptTemplateFactory.java
zambrinf: @langchain4j the problem only happens if you have more than 1 parameter, check the example I gave in the `change` section of the PR descriptions (has  `String character` + varargs argument)

PR #1502, opened 2024-07-22
module names everywhere

hendrikebbers:


langchain4j-anthropic/pom.xml
```diff
@@ -108,6 +108,18 @@
...
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>${maven-jar-plugin.version}</version>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>dev.langchain4j.model.anthropic</Automatic-Module-Name>
```
langchain4j:
```suggestion
                            <Automatic-Module-Name>dev.langchain4j.anthropic</Automatic-Module-Name>
```
To be consistent with other modules

langchain4j-azure-ai-search/pom.xml
```diff
@@ -103,6 +103,18 @@
...
                 <artifactId>license-maven-plugin</artifactId>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>${maven-jar-plugin.version}</version>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <Automatic-Module-Name>dev.langchain4j.azure.search</Automatic-Module-Name>
```
langchain4j:
```suggestion
                            <Automatic-Module-Name>dev.langchain4j.azure.aisearch</Automatic-Module-Name>
```

ghost (commented):
@hendrikebbers thank you!

From the [article](https://dev.java/learn/modules/automatic-module/) that you mentioned I understand that this PR will be a "commitment":

> - The names of the future modules must be exactly those that you declare now. (Otherwise, reliable configuration will bite your users because modules are missing.)
> - The artifact structure must remain the same, so you can't start moving supported classes or packages from one JAR to another. (Even without modules, this isn't recommended practice, but with the class path it doesn't matter which JAR contains a class, so you could get away with it. With the module system in play, on the other hand, a class' origin is very relevant because accessibility forces users to require the correct module.)

I am not sure I am ready to commit to this as I envision changing module naming convention and module structure before/for 1.0.

Do you think it makes sense to merge this PR now?

hendrikebbers:
@langchain4j In that case, I will continue to check first if everything is prepared to switch to modules easily. Once that is done we can discuss if and when it makes sense to add module names. Just created https://github.com/langchain4j/langchain4j/issues/1537

Question: will https://github.com/langchain4j/langchain4j-embeddings change in future, too. Otherwise it would make sense that I include that in my research.

langchain4j:
@hendrikebbers thanks!

> Question: will https://github.com/langchain4j/langchain4j-embeddings change in future, too. Otherwise it would make sense that I include that in my research.

What do you mean? That repo is a part of the project and will continue to be changed (bugfixes, new features, etc).

Issue #1068, opened 2024-05-08
[FEATURE] Azure OpenAI: Use provided token usage instead of manually calculated


langchain4j:
OpenAI did not provide token usage in streaming mode for a long time, so we had to calculate it manually (see `OpenAiStreamingResponseBuilder`).

Now they provide it in the last SSE event: https://cookbook.openai.com/examples/how_to_stream_completions#4-how-to-get-token-usage-data-for-streamed-chat-completion-response

We should use it instead of our current approach.
Might require adapting https://github.com/ai-for-java/openai4j

langchain4j:
@jdubois do you know if `include_usage` will be supported in Azure?

jdubois:
I have no idea - maybe @agoncal will know as he modified that part recently? I've also pinged internally our PM in charge of the SDK, so we can get an official answer.

langchain4j:
Related: https://github.com/langchain4j/langchain4j/issues/1621

dliubarskyi:
Hi @jdubois, do you have any news regarding this?

jdubois:
Sorry, I still don't have an answer on this. It's probably handled on the Open AI side.

dliubarskyi:
`ChatCompletionsOptions` has `ChatCompletionStreamOptions streamOptions`, so we can implement it now

dliubarskyi:
setter for `ChatCompletionsOptions` is private for some reason - any idea why?

```java
/**
     * Set the streamOptions property: Options for streaming response. Only set this when you set `stream: true`.
     *
     * @param streamOptions the streamOptions value to set.
     * @return the ChatCompletionsOptions object itself.
     */
    private ChatCompletionsOptions setStreamOptions(ChatCompletionStreamOptions streamOptions) {
        this.streamOptions = streamOptions;
        return this;
    }
```

This blocks simplifying `AzureOpenAiTokenizer` and getting rid of deprecated APIs in LC4j. It would really help a lot if we could just set `ChatCompletionsOptions` with `includeUsage=true`

PR #1504, opened 2024-07-22
OpenAI: Get model-specific tokenizer automatically

### LINKED ISSUE #1102: [FEATURE] OpenAI: Tokenizer for gpt-4o (update jtokkit)
gpt-4o uses a different tokenizer, we need to update jtokkit version once it is released

drbscl:
## Issue

Closes #1102

## Change

This PR makes the OpenAI model creation logic supply the specified (or default) model name when initialising the tokeniser. This is required, since GPT 4o (and related models) use a different tokenizer. 

This was the most straightforward way to fix this from what I can tell, but I welcome suggestions for better implementations.

`modelName` is always non-null in all modified paths, so this change should be safe.

## General checklist

  - [ ] Added unit tests, but unsure if integration tests are necessary for the change

   - [ ] I checked, and I don't believe this is necessary here.  
- [ ] ~I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)~ N/A
- [ ] ~I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)~ N/A

- [ ] ~I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)~ N/A

- [ ] ~I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`~ N/A
- [ ] ~I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)~ N/A

- [ ] ~I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j~ N/A

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
@@ -112,7 +112,7 @@ public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
         this.seed = seed;
         this.user = user;
         this.maxRetries = getOrDefault(maxRetries, 3);
-        this.tokenizer = getOrDefault(tokenizer, () -> new OpenAiTokenizer(this.modelName));
+        this.tokenizer = OpenAiTokenizer.getTokenizerOrDefault(tokenizer, this.modelName);
         this.listeners = listeners == null ? emptyList() : new ArrayList<>(listeners);
     }
 
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
@@ -112,7 +112,7 @@ public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
         this.seed = seed;
         this.user = user;
         this.maxRetries = getOrDefault(maxRetries, 3);
-        this.tokenizer = getOrDefault(tokenizer, OpenAiTokenizer::new);
+        this.tokenizer = getOrDefault(tokenizer, () -> new OpenAiTokenizer(this.modelName));
         this.listeners = listeners == null ? emptyList() : new ArrayList<>(listeners);
     }
 
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiTokenizer.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiTokenizer.java
@@ -67,14 +67,18 @@ public class OpenAiTokenizer implements Tokenizer {
     // TODO rename
     public static Tokenizer getOpenAiTokenizerInOrder(Tokenizer specifiedTokenizer, String modelName) {
         if (specifiedTokenizer != null) {
+            // Use the tokenizer provided by the consumer, if present
             return specifiedTokenizer;
         }
         if (modelName != null && !modelName.trim().isEmpty()) {
+            // Attempt to retrieve the model's associated tokenizer
             OpenAiTokenizer candidate = new OpenAiTokenizer(modelName);
             if (candidate.encoding.isPresent()) {
                 return candidate;
             }
         }
+        // Failing the other options, default to GPT 3.5's tokenizer, since it is the most prevalent, and is used by
+        // other (non-OpenAI) models
         return new OpenAiTokenizer(GPT_3_5_TURBO);
     }
 
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiTokenizer.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiTokenizer.java
@@ -64,8 +64,7 @@ public class OpenAiTokenizer implements Tokenizer {
         this.encoding = Encodings.newLazyEncodingRegistry().getEncodingForModel(modelName);
     }
 
-    // TODO rename
-    public static Tokenizer getOpenAiTokenizerInOrder(Tokenizer specifiedTokenizer, String modelName) {
+    public static Tokenizer getTokenizerOrDefault(Tokenizer specifiedTokenizer, String modelName) {
         if (specifiedTokenizer != null) {
             // Use the tokenizer provided by the consumer, if present
             return specifiedTokenizer;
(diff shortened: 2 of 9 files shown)


drbscl:
(Marking as draft pending creation of tests & completion of checklist.)

drbscl:
So, a couple of things:

1. I don't personally have an OpenAI API key, so I can't currently run the full test suite locally. Does CI cover this?
2. Do you need me to write a similar test to correspond with each of my similar changes?

langchain4j:
Hi @ben-clegg thanks a lot!

> 1. I don't personally have an OpenAI API key, so I can't currently run the full test suite locally. Does CI cover this?

CI does not cover it (yet). I can run the tests if you need.

> 2. Do you need me to write a similar test to correspond with each of my similar changes?

Sorry, I don't understand, could you elaborate?


How urgent is this change for you?

There are 2 things to keep in mind:
1. `OpenAi*Model` classes are used with other LLM providers as well (e.g. Ollama, Groq, etc.), so model names will be different and jtokkit will not know them
2. `OpenAi*Model` classes are used with fine-tuned OpenAI models, so model name is also different (which leads to the same problem)

Since `OpenAiTokenizer` is used to calculate tokens for every request/response in `OpenAiStreamingModel` and `OpenAiStreamingLanguageModel`, we will have to [implement this](https://github.com/langchain4j/langchain4j/issues/1068) first.

drbscl:
> CI does not cover it (yet). I can run the tests if you need.

Thank you for this. I will probably make more changes, so no need until this PR is ready. I'll mark the PR as draft for now.

> Sorry, I don't understand, could you elaborate?

I made a very similar change across several classes, but only added a test for one. However (as I'll explain below), I think there is probably a better solution in general, which will also simplify testing.

> How urgent is this change for you? 

Not all that urgent, just wanted to contribute.

> There are 2 things to keep in mind:
> `OpenAi*Model` classes are used with other LLM providers as well (e.g. Ollama, Groq, etc.), so model names will be different and jtokkit will not know them
> `OpenAi*Model` classes are used with fine-tuned OpenAI models, so model name is also different (which leads to the same problem)

This is very useful information, thanks. I have an idea for how to not block those use cases here. I'll make a further change to further simplify and allow a default (GPT 3.5's tokenizer) to be used if a model is not found in jtokkit.

> Since OpenAiTokenizer is used to calculate tokens for every request/response in OpenAiStreamingModel and OpenAiStreamingLanguageModel, we will have to https://github.com/langchain4j/langchain4j/issues/1068 first.

I think that if I make a further change this will hopefully not be a blocker, but honestly it's no problem for me if it is. 

Thanks for your detailed & welcoming response!

langchain4j:
> I think that if I make a further change this will hopefully not be a blocker, but honestly it's no problem for me if it is.

OK, let's see the change and then decide about https://github.com/langchain4j/langchain4j/issues/1068

drbscl:
Ok, let me know what you think.

langchain4j:
@ben-clegg sorry for the delay! I have recently merged [this](https://github.com/langchain4j/langchain4j/pull/1622) and I expect to do the same for [Azure OpenAI](https://github.com/langchain4j/langchain4j/issues/1068) soon. Considering that `Tokenizer` is no more responsible for calculating `TokenUsage` for OpenAI, do you think this change is still required? Thanks!

drbscl:
@langchain4j no problem, and sorry for the late reply myself. My (admittedly loose) understanding is that the linked PRs will correctly provide the tokens actually used in the response (or a partial, streamed response), but there is still a use case for checking the number of tokens prior to making the request.

Some past cases where I've found this useful myself:
- Validating the token limit without having to wait for the request's flight time
- Validating a limit for a sub-prompt

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiTokenizer.java
```diff
@@ -31,14 +31,6 @@ public class OpenAiTokenizer implements Tokenizer {
     private final String modelName;
     private final Optional<Encoding> encoding;
 
-    /**
-     * Creates an instance of the {@code OpenAiTokenizer} for the "gpt-3.5-turbo" model.
-     * It should be suitable for all current OpenAI models, as they all use the same cl100k_base encoding.
-     */
-    public OpenAiTokenizer() {
```
langchain4j, line 38: I would mark it `@Deprecated` instead of removing, otherwise it will break a lot of code

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
```diff
@@ -112,7 +112,7 @@ public OpenAiChatModel(String baseUrl,
         this.seed = seed;
         this.user = user;
         this.maxRetries = getOrDefault(maxRetries, 3);
-        this.tokenizer = getOrDefault(tokenizer, OpenAiTokenizer::new);
+        this.tokenizer = OpenAiTokenizer.getTokenizerOrDefault(tokenizer, this.modelName);
```
langchain4j, line 115:
```suggestion
        this.tokenizer = getOrDefault(tokenizer, () -> OpenAiTokenizer.from(this.modelName));
```

What about keeping `getOrDefault` logic here?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiTokenizer.java
```diff
@@ -72,6 +64,23 @@ public OpenAiTokenizer(String modelName) {
...
+        if (modelName != null && !modelName.trim().isEmpty()) {
+            // Attempt to retrieve the model's associated tokenizer
+            OpenAiTokenizer candidate = new OpenAiTokenizer(modelName);
+            if (candidate.encoding.isPresent()) {
+                return candidate;
+            }
+        }
+        // Failing the other options, default to GPT 3.5's tokenizer, since it is the most prevalent, and is used by
+        // other (non-OpenAI) models
+        return new OpenAiTokenizer(GPT_3_5_TURBO);
```
langchain4j, line 81:
```suggestion
        return new OpenAiTokenizer(GPT_4_O_MINI);
```
I guess GPT_4_O_MINI is most commonly used right now, 3.5 is deprecated

langchain4j: On a second thought, perharps we should return `null` here instead of a tokenizer with a default model name?
Since `Tokenizer` is not required for basic functionality of `(Streaming)ChatLanguageModel` any more, it is ok to return `null`. If user wants to estimate token count and is using some LLM provider/name that is not supported by tiktoken, perharps it is better to get NPE and configure `Tokenizer` explicitly than get wrong token count...

WDYT?

ghost (commented):
@ben-clegg thank you!

dliubarskyi:
@ben-clegg thank you for your PR, but I am closing it due to inactivity

Issue #1506, opened 2024-07-23
Maven (dependency) enforcer plugin


PrimosK:
## Dependency issues identified

- **Existing Dependency Convergence Errors**: Various modules have unresolved dependency convergence issues.

- **Need for Enforcing Dependency Convergence**: It might be beneficial to configure [Maven (dependency) enforcer plugin](https://maven.apache.org/enforcer/maven-enforcer-plugin/usage.html) to error out the build when such issues are detected. That will force contributtors to keep project free of dependency confclits consequently ensure a more stable and predictable build process, which is crucial for the reliability of the project.

## Proposed solution

- **Resolve Existing Dependency Convergence Errors**: Address current dependency convergence errors in each affected module.
- **Enforce Dependency Convergence**: Set up the Maven Enforcer Plugin to ensure that the build fails if dependency convergence errors are detected.

### Alternative solution

Do nothing - i.e. leave it like it is and let end-user solve these in their project when including `Langchain4j`.

## Additional context

This feature request was created based on the https://github.com/langchain4j/langchain4j/discussions/1494 discussion. So please read it through to get some additional information.

## Work plan

I suggest we tackle this in incremental fashion by first enforcing dependency convergence in sub-modules. When all sub-modules will be tackled we can define it in single place - so in `LangChain4j :: Parent POM` module.

So to track this properly I've prepared this list of Maven modules showing where Maven enforcer plugin was already 'turned on'. That also means that module that is marked done has no pending dependency conflicts.

I will soon prepare a draft PR(s) to better demonstrate how I envisioned tackling this.

PRs:
1. https://github.com/langchain4j/langchain4j/pull/1507
2. https://github.com/langchain4j/langchain4j/pull/1508
3. https://github.com/langchain4j/langchain4j/pull/1509
4. https://github.com/langchain4j/langchain4j/pull/1645
5. ...

Legend:
🏁 - PR pending
✔️ - PR accepted

|   | Module | PR  | Comment  |
|---|--------|-------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
|   | LangChain4j :: Parent POM || |
| ✔️ | LangChain4j |https://github.com/langchain4j/langchain4j/pull/1507| No dependency conflicts - just enabling Maven enforcer. |
|  ✔️ | LangChain4j :: Core |https://github.com/langchain4j/langchain4j/pull/1507| No dependency conflicts - just enabling Maven enforcer. |
[...]

PR #1507, opened 2024-07-23
re #1506 Enabling Maven (version) enforcer plugin in modules with no version conflicts

PrimosK:
## Issue

#1506

## Change

Enabled Maven Enforcer Plugin on modules without existing version conflicts to ensure they remain conflict-free. The Maven Enforcer Plugin will now cause the build to fail if new conflicts are introduced guarding against these.

## Tests

`mvn clean test` passed


ghost (approved):
@PrimosK thank you!

PR #1516, opened 2024-07-24
1446 : Extraction of json block before parse

### LINKED ISSUE #1446: [FEATURE] If LLM responds with a JSON wrapped in backticks and/or some additional text, try to extract that instead of failing
This issue is a part of https://github.com/langchain4j/langchain4j/issues/859

Sometimes when doing [structured data extraction](https://docs.langchain4j.dev/tutorials/ai-services#output-parsing-aka-structured-outputs), LLM response can contain backticks:
````
```json
{
    ...
}
```
````
It can also be without "json":
````
```
{
    ...
}
```
````
Or with some text before and/or after the code block:
````
Sure, here is your JSON:
```json
{
    ...
}
```
Let me know if you need more help.
````

In such cases, instead of failing, we need to try to extract the valid json.

### PR DESCRIPTION
**Pull request by patpe**
## Issue
Closes #1446

## Change
When return type of the AI Service method is a POJO, extract json block from LLM response if parsing failed.
Continuation: ab819a32

## General checklist

### CODE REVIEW

File: langchain4j/src/main/java/dev/langchain4j/service/output/DefaultServiceOutputParser.java [hunk 1]
```diff
@@ -0,0 +1,211 @@
...
+    private String extractJsonBlock(String text) {
+        int startIdx = text.indexOf("{");
+        int endIdx = text.lastIndexOf("}");
+
+        if (startIdx == -1 || endIdx == -1) {
+            return text;
+        }
+
+        return text.substring(startIdx, endIdx + 1);
+    }
```
patpe: New method added to extract json block

langchain4j/src/main/java/dev/langchain4j/service/output/DefaultServiceOutputParser.java
```diff
@@ -0,0 +1,211 @@
...
+        if (rawReturnClass == String.class) {
+            return text;
+        }
+
+        Optional<OutputParser<?>> optionalOutputParser = outputParserFactory.get(rawReturnClass, typeArgumentClass);
+        if (optionalOutputParser.isPresent()) {
+            return optionalOutputParser.get().parse(text);
+        }
+
+        String extractedJsonBlock = extractJsonBlock(text);
```
patpe: New method invocation to extract json block before parsing

langchain4j:
Hi @patpe, thanks a lot!

> If custom ServiceOutputParser should be moved to separate PR or not solved, I will remove it.

Yes, please remove it from this PR, I am not yet sure this is the API we should be exposing to the users

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -192,8 +194,7 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
 
                         response = Response.from(response.content(), tokenUsageAccumulator, response.finishReason());
 
-                        Object parsedResponse;
-                        parsedResponse = serviceOutputParser.parse(response, returnType);
+                        Object parsedResponse = serviceOutputParser.parse(response, returnType);
```
patpe, line 241: Minor fix, couldn't see the reason for splitting into two lines. Seems like an old refactor that didn't go all the way?

langchain4j/src/main/java/dev/langchain4j/service/output/ServiceOutputParser.java
```diff
@@ -105,7 +106,7 @@ public String outputFormatInstructions(Type returnType) {
         return "\nYou must answer strictly in the following JSON format: " + jsonStructure((rawClass), new HashSet<>());
     }
 
-    public static String jsonStructure(Class<?> structured, Set<Class<?>> visited) {
+    private static String jsonStructure(Class<?> structured, Set<Class<?>> visited) {
```
patpe, line 124: Minor fix, this method does not need to be public and reducing visibility keeps the public API clean

patpe:
@langchain4j thanks for feedback, removed refactor of ServiceOutputParser

langchain4j:
Related: https://github.com/quarkiverse/quarkus-langchain4j/pull/780

patpe:
> Related: [quarkiverse/quarkus-langchain4j#780](https://github.com/quarkiverse/quarkus-langchain4j/pull/780)

Their issue seems to be a bit different since it implies a setting adding `\n` to the output json block. But a regex solution would work for this case as well, would it be better?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
langchain4j: Seems so, thank you!

langchain4j/src/main/java/dev/langchain4j/service/output/ServiceOutputParser.java
langchain4j: It was made public by @geoand to be usable from Quarkus extension, but it seems that it is not used?

langchain4j/src/main/java/dev/langchain4j/service/output/ServiceOutputParser.java
```diff
@@ -194,4 +195,15 @@ private static String simpleTypeName(Type type) {
                 return type.getTypeName();
         }
     }
+
+    private String extractJsonBlock(String text) {
+        int startIdx = text.indexOf("{");
```
langchain4j: `["apple","pear","banana"]` is also a valid json.
Although we do not support extracting collections of POJOs (yet).
I am wondering if we should use regex similar to [this](https://github.com/quarkiverse/quarkus-langchain4j/pull/780/files) instead, WDYT?

langchain4j: Also, what about "Sure, here is your JSON enclosed in '{' and '}'\njson{...}"? :)
It will extract `{' and '}'\njson{...}` in this case

ghost (commented):
@patpe thank you!

langchain4j:
> > Related: [quarkiverse/quarkus-langchain4j#780](https://github.com/quarkiverse/quarkus-langchain4j/pull/780)
> 
> Their issue seems to be a bit different since it implies a setting adding `\n` to the output json block. But a regex solution would work for this case as well, would it be better?

Ah, you're right, I missed that!
Well, a (slightly different) regex is also a valid solution I guess, might be even more robust

patpe:
> > > Related: [quarkiverse/quarkus-langchain4j#780](https://github.com/quarkiverse/quarkus-langchain4j/pull/780)
> > 
> > 
> > Their issue seems to be a bit different since it implies a setting adding `\n` to the output json block. But a regex solution would work for this case as well, would it be better?
> 
> Ah, you're right, I missed that! Well, a (slightly different) regex is also a valid solution I guess, might be even more robust

No problem, I could go for a regex solution instead. Just one question before heading forward.

Since the root cause of this problem is the behavior of different language models which produce additional output when asked for JSON, it will be very hard to write a catch all regex (or any other solution for that matter). To properly handle this, I think a design where a developer can inject a custom `JsonConverter` that if set is invoked before the text is parsed.

Which is where the `ServiceOutputParser` refactor might come in handy. Thoughts?

langchain4j:
True, configurable output parser is a good idea (let's keep it separate though), but we should first try to make the default implementation robust

patpe:
Updated to use regex and support collections of objects as result object. Implemented json array support because I added test cases for making sure that the regex properly supported extracting the array.

Don't know where this is going at the moment, array support too much for this PR? A different solution would be to make the json extraction a separate package private class to be able to test it which would remove the need for json array support for now. Can then only validate the string extraction.

langchain4j-core/src/main/java/dev/langchain4j/internal/GsonJsonCodec.java
```diff
@@ -59,6 +60,12 @@ public <T> T fromJson(String json, Class<T> type) {
         return GSON.fromJson(json, type);
     }
 
+    @Override
+    public <T, U> T fromJson(String json, Class<T> rawType, Class<U> typeArgumentClass) {
```
langchain4j:
```suggestion
    public <T> T fromJson(String json, Class<T> rawType, Class<?> typeArgumentClass) {
```
Seems like there is no need for `U`

langchain4j-core/src/main/java/dev/langchain4j/internal/Json.java
```diff
@@ -37,6 +37,20 @@ public interface JsonCodec {
...
+         * This method is useful if the specified object is a generic type.
+         * For non-generic objects, use {@link #fromJson(String, Class)} instead. 
+         *
+         * @param json the JSON string.
+         * @param typeToken The specific genericized type of the object.
+         * @param <T>  the type of the object.
+         * @param <U>  the argument type of the class of the object
+         * @return the object.
+         */
+        <T, U> T fromJson(String json, Class<T> rawType, Class<U> typeArgumentClass);
```
langchain4j:
```suggestion
        <T> T fromJson(String json, Class<T> rawType, Class<?> typeArgumentClass);
```

langchain4j-core/src/main/java/dev/langchain4j/internal/Json.java
```diff
@@ -79,6 +93,20 @@ public static <T> T fromJson(String json, Class<T> type) {
...
+     * Convert the given JSON string to an object of the given type.
+     *
+     * @param json the JSON string.
+     * @param rawType the type of the object.
+     * @param typeArgument the generic type argument of the object.
+     * @param <T> the type of the object.
+     * @param <U> the generic type of {@code rawType}. 
+     * @return the object.
+     */
+    public static <T, U> T fromJson(String json, Class<T> rawType, Class<U> typeArgument) {
```
langchain4j:
```suggestion
    public static <T> T fromJson(String json, Class<T> rawType, Class<?> typeArgument) {
```

langchain4j/src/main/java/dev/langchain4j/service/output/ServiceOutputParser.java
```diff
@@ -18,6 +20,20 @@
 
 public class ServiceOutputParser {
 
+    /**
+     * JSON Pattern:<br />
+     * 
+     * <i>\\{(?:[^{}]|\\{(?:[^{}]|\\{(?:[^{}]|\\{[^{}]*\\})*\\})*\\})*\\}</i>: Matches JSON objects, accounting for nested objects.
+     * <i>|\\[(?:[^\\[\\]]|\\[(?:[^\\[\\]]|\\[(?:[^\\[\\]]|\\[[^\\[\\]]*\\])*\\])*\\])*\\]</i>: Matches JSON arrays, including arrays of objects.
+     */
+    private static final String JSON_PATTERN_REGEX = "\\{(?:[^{}]|\\{(?:[^{}]|\\{(?:[^{}]|\\{[^{}]*\\})*\\})*\\})*\\}" +
```
langchain4j: Oh, that is a very complicated pattern... any chance to simplify it? or at least document it a bit more which part does what

ghost (commented):
@patpe thank you, good job!

langchain4j:
> Don't know where this is going at the moment, array support too much for this PR?

Yes, let's keep it out of this PR. It seems that it adds complexity to the regex expression, and since we do not really require it now, let's better simplify it.


> A different solution would be to make the json extraction a separate package private class to be able to test it which would remove the need for json array support for now.

I am not sure how exactly separate class will allow to remove the need for json array support, but it a good idea to extract json logic form this class in any case

langchain4j/src/main/java/dev/langchain4j/service/output/ServiceOutputParser.java
patpe: As you might have guessed, this was crafted using an LLM (ChatGPT). When asked if it can be simplified without loosing any functionality it says no (more or less).

The only simplification I can think of is to drop the array support which will give a one line expression but also reduce the capabilities of langchain4j. This is fine by me of course.

patpe:
Dropped support for collections so this PR is focused on extracting JSON text before parsing it.

I hadn't noticed that IT was not being run locally and I can't find the docs for how to properly run them. Am I supposed to generate my own set of keys for the integrated models and run the tests?

langchain4j/src/main/java/dev/langchain4j/service/output/ServiceOutputParser.java
langchain4j: Ok, I see.

There is [another solution](https://github.com/langchain4j/langchain4j/pull/1605) to solve this issue which TBH I prefer more. It is also kind of a black box (at least for me), but time-tested in Groovy and more extendable if we'll want to forgive some LLM mistakes in terms of json syntax.

I hope you don't mind if we will proceed with #1605 and close this PR. Anyway, thank you so much for your efforts!

patpe: No harm, we all share a common goal

geoand:
@andreadimaio you probably want to have a look at this

andreadimaio:
A regex that could be used to extract the contents of a JSON object or array from a string is  `(?s)\\{.*\\}|\\[.*\\]`. This is the regex used in quarkus-langchain4j to sanitise the JSON string coming from the LLM. quarkiverse/quarkus-langchain4j#803

langchain4j:
@andreadimaio thank you for pointing that out, that seems like a good solution!

@patpe I will merge this PR and adapt the regexp

PR #1519, opened 2024-07-25
1465 : Ensuring trailing / in retrofit baseurl

### LINKED ISSUE #1465: [BUG] Incorrect use of Retrofit relative URLs
First spotted in https://github.com/langchain4j/langchain4j/issues/1455

There is a number of places where we use Retrofit's `@GET`, `@POST`, `@DELETE`, etc incorrectly.

For example:

```java
interface ChromaApi {

    @GET("/api/v1/collections/{collection_name}")
    @Headers({ "Content-Type: application/json" })
    Call<Collection> collection(@Path("collection_name") String collectionName);

    ...
}
```
Here, the path inside the `@GET` annotation starts with "/" which means that this will be treated as absolute URL.

If user defines `baseUrl` as `http://my.com/sub/`, `sub` will be ignored and final URL will be `http://my.com//api/v1/collections/{collection_name}`.

We need to fix each such case and make sure URL resolution works as expected.

Affected modules:
- `langchain4j-chroma`
- `langchain4j-vearch`
- `langchain4j-chatglm`
- `langchain4j-hugging-face`
- `langchain4j-tavily`

patpe:
## Issue
Closes #1465

## Change
According to [retrofit](https://github.com/square/retrofit/blob/trunk/retrofit%2Fsrc%2Fmain%2Fjava%2Fretrofit2%2FRetrofit.java#L564) base urls should always end with `/`.

Added new utility method to ensure that a provided base url always ends with a `/` and checked existing API classes so that they all start **without** a `/`.

### Tests
I have added unit test for the new utility method but testing the actual invocation of the method in the different builder classes is harder. The existing Ollama test case spins up a temporary web server and I don't want to replicate this to al lmodules since I suspect build times will increase a lot etc.

Thoughts?

## General checklist

### DIFF
--- a/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java
+++ b/langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java
@@ -2,6 +2,7 @@ package dev.langchain4j.model.anthropic.internal.client;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.internal.Utils;
 import dev.langchain4j.model.StreamingResponseHandler;
 import dev.langchain4j.model.anthropic.internal.api.*;
 import dev.langchain4j.model.output.Response;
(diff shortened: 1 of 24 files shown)


ghost (approved):
@patpe thank you, good job!

PR #1539, opened 2024-07-29
fix: Use `NODE` type to check for the values type.

### LINKED ISSUE #1537: Neo4j module use private api
To be prepared to use Java modules in the future (see https://github.com/langchain4j/langchain4j/issues/1066), langchain4j code must not access code of a module that is not exported by that module. The Neo4J module uses `org.neo4j.driver.internal.value.NodeValue` in the `Neo4jContentRetriever` class today.

michael-simons:
Fixes #1537.


ghost (approved):
@michael-simons thank you a lot!

Issue #1525, opened 2024-07-26
[FEATURE] Ollama: support tools


langchain4j:
Ollama is supporting tools now: https://ollama.com/blog/tool-support

While tools can already be used with `langchain4j-open-ai` module, it would be nice to support them in `langchain4j-ollama` module as well.

langchain4j:
Implementation in Quarkus extension: https://github.com/quarkiverse/quarkus-langchain4j/pull/783

jprogramista:
I’ve tried to use these freshly merged tools and encountered some problems. I modified the test OllamaToolChatModelIT and used my locally installed Ollama on Mac (no docker). However, I couldn’t get a response, so I finally examined the requests and responses. It turned out that if I didn’t add tools to the second request, the "tool" role answer was ignored. In the end, I tried the requests in Postman, and it seems they only work if the tools spec is added.
```
{
    "model": "llama3.1",
    "messages": [
        {
            "content": "What is the flight price from New York (NYC) to Los Angeles (LAX)?",
            "role": "user"
        },
        {
            "content": "",
            "role": "assistant",
            "tool_calls": [
                {
                    "function": {
                        "arguments": {
                            "arrival": "LAX",
                            "departure": "NYC"
                        },
                        "name": "get_flight_info"
                    }
                }
            ]
        },
        {
            "content": "{&quot;from&quot;: &quot;NYC&quot;, &quot;to&quot;: &quot;LAX&quot;, &quot;duration&quot;: &quot;15h 30m&quot;, &quot;price&quot;: &quot;150 USD&quot;}",
            "role": "tool"
        }
    ],
    "options": {},
    "stream": false,
    "tools": [
        {
            "type": "function",
            "function": {
                "name": "get_flight_info",
                "description": "Get the flight data",
                "parameters": {
                    "properties": {
                        "arrival": {
                            "description": "The arrival ",
                            "departure": "string"
                        },
                        "departure": {
                            "description": "The departure",
                            "type": "string"
                        }
                    },
                    "required": [
                        "arrival",
                        "departure"
                    ]
                }
            }
        }
    ]
}
```
Answer: "The price of a flight from New York (NYC) to Los Angeles (LAX) is $150 USD. The duration of the flight is approximately 15 hours and 30 minutes."

If I remove tools section and also if I don't have `toolSpecifications` in:
```
Response<AiMessage> secondResponse = ollamaChatModel.generate(messages, toolSpecifications);
```

[...]

Issue #318, opened 2023-11-27
[FEATURE] Ollama to support Tools/Functions


goafabric:
Langchain4j is at it's current state already an awesome product.
And the recent Ollama integration, makes it possible to have a powerfull LLM running locally, for chat requests.

What it is unfortunately missing, is an implementation of the ChatLanguageModel,
that can be used together with Functions and Tools.
To melt local code with an LLM
For OpenAI we have OpenAiChatModel + AIServices.

For Ollama, thats not yet there.
Outside the Java Ecosystem i found this one:
https://js.langchain.com/docs/integrations/chat/ollama_functions

Would it be possible to get an Ollama Integration similar to the OpenAI one ?

langchain4j:
HI @goafabric yes, this is one of the higher priorities right now. Not sure about Ollama (will check a bit later), but [LocalAI](https://localai.io/) supports [function calling](https://localai.io/features/openai-functions/). Of course it depends a lot on the underlying model... If you have a possibility to run it somewhere (there is a docker image), you can try `AiServices` + `LocalAiChatModel` (from `langchain4j-local-ai` module) with some llamacpp model fine-tuned for function calling.

goafabric:
thx 4 letting me know .. i shied away from localai .. because of the rather complex build options
but i now found there is a simple "localai.app" that can even run on apple silicon .. i might eventually give that a spin
until function support is there for ollama

thx again

langchain4j:
@goafabric Ollama has `format="json"` which can help with function calling, I have added support for this in https://github.com/langchain4j/langchain4j/pull/373

bidek:
I made some experiments, trying to convince model to respond in specific manner if he want to execute tool.

If model respond with string `TOOL-EXECUTION-REQUEST - currentDate` i capture this and respond with `ToolExecutionRequest` to be compatible with current tool execution flow from `DefaultAiServices`.

For now, it only support tools without parameters, capturing and parsing execution request is very fragile but it works.

I tested it on `mistral` model from standard ollama model repository

Its based on work of @fintanmm , maybe @fintanmm could comment on how he sees ollama langchain tool support

[fork-with-ollama-tool-support](https://github.com/bidek/langchain4j/tree/features/ollama-tools-support)
[example-tool-usage](https://github.com/bidek/ollamaToolUsage/blob/master/src/main/java/Main.java)

fintanmm:
[...]

PR #1558, opened 2024-08-04
Ollama Tools Support - only in OllamaChatModel

bidek:
## Issue

It is relevant to:
#1525
#318

## Change

Add new DTOs to support Tool/Function calling.
Add new integration test and model configuration that supports Tool Calls - Mistral
Modify some class to support Tool calling, for example `OllamaChatModel`, `Role` 

As stated here https://ollama.com/blog/tool-support, tools support is not possible in streaming scenario - for now

I got the best results with the Mistral model, but without a SystemMessage, it always wants to call a tool, even if it doesn't make any sense. With Llama 3.1, it properly returns a ToolCallRequest but ignores the result.

## General checklist


ghost (approved):
@bidek great job, thank you!

PR #1587, opened 2024-08-13
Use dev.langchain4j.internal.Json to serialize tool params

G0dC0der:
## Issue

n/a

## Change
So far, we can customise the Gson instance used to serialise tool responses, through SPI. While this is an uncommon way to do it, it works!

In a recent release of Langchain4J, we can have complex object as tool params. The Gson instance used to serialise these can however _not_ be customised. This limits the feature and the goal of this pull request is to change that.

**The change**: Tools now use `dev.langchain4j.internal.Json`to serialise params, instead of hardcoding a private static instance. Same class have also gotten an additional method, which the tools needed. This means the same Gson instance will be used to serialise tool responses and tool params.

**This is a breaking change!** Previously, you could pass UNIX timestamp to a tool and it would be serialised into a LocalDate or LocalDateTime(other `java.time` objects are not affected). Now you have to pass the ISO date. I have however few reasons why this is ok:

1) The feature(complex objects as tool params) is very new and it is unlikely that someone is affected 
2) This change homogenise the serialisation of tool params and responses
3) This feature could be achieved with few refactors/changes
4) If you really need to pass a UNIX timestamp, you can always supply your own Gson instance through SPI

## General checklist

**Note:** I couldn't run the integration tests due to missing api keys!


langchain4j:
@G0dC0der thank you!
Out of interest, how/why do you customize Json serialization/deserialization for tools?

> The change: Tools now use dev.langchain4j.internal.Jsonto serialise params, instead of hardcoding a private static instance.

Do you actually mean **de** -serialise tool _**arguments**_?

langchain4j:
> Previously, you could pass UNIX timestamp to a tool and it would be serialised into a LocalDate or LocalDateTime(other java.time objects are not affected). Now you have to pass the ISO date.

Could you please provide a code to reproduce this scenario? Thanks!

G0dC0der:
**You mean deserialise?**
Yes

**How to customise the Gson**
First, create a new file called `dev.langchain4j.spi.json.JsonCodecFactory` in `resources/META-INF/services`. 
This should be the contents of the file:
```
com.g0dcoder.gson.MyCustomGsonIntanceFactory
```

This class(`MyCustomGsonIntance`), is my own class and it implements `dev.langchain4j.spi.json.JsonCodecFactory`. This wasn't documented anywhere, I had to look into the source code. 

**Why should I customise it?**
The Gson instance used by default have problem deserialising multiple common classes, such as `java.time.Instant`. It's also unable to handle your own complex types. Take this as an exmaple:
```java
public sealed interface RevisionCriteria {

    enum RevisionStatus implements RevisionCriteria { LATEST, LAST_PUBLISHED }

    record TargetRevision(Integer rev) implements RevisionCriteria {}
}
```
An instance of `RevisionCriteria`may be either "LATEST", "LAST_PUBLISHED"(`RevisionStatus`) , or an integer(`TargetRevision`). The out-of-the-box Gson has no idea how to construct this object based on the supplied data, and you basically need to register your own `TypeAdapter`. This is just an example, I imagine companies have loads of custom complex types. 

**Reproducing the breaking change**
Create a tool:
```java
@Tool 
public void printTime(Instant time) { 
    System.out.println("The time is " + time);
}

void someMethod() {
    myAssistant.chat("Invoke tool printTime with param 1724703310112"); //This should cause an error !
}
```

This should cause an error, because the Gson instance used for deserialising tool responses for some reason expects an ISO date or an object that contains "year", "month" "day" etc:

https://github.com/langchain4j/langchain4j/blob/30d02f84c726122c35c0b3a8d83eb7c3f593e26b/langchain4j-core/src/main/java/dev/langchain4j/internal/GsonJsonCodec.java#L17-L83

langchain4j:
@G0dC0der thank you for the info!
Apart from `java.time.Instant`, did you have problems with other commonly used types?

MrP-ICA:
Well, all the java 8 types that are not listed in GsonJsonCodec(shown above) such as ZonedDateTime, Optional(you shouldn't use Optional as params though), ZoneOffet/ZoneId, Duration. Period etc.

G0dC0der:
ZonedDateTime, Optional, Duration, ZoneOffset etc. Basically, all of those who does not get an explicit typer adapter registered at dev.langchain4j.internal.GsonJsonCodec

G0dC0der:
@langchain4j do you need anything more from me in order for this to be merged?

langchain4j:
@G0dC0der sorry for the delay, I will take a look at it again

langchain4j:
> This should cause an error, because the Gson instance used for deserialising tool responses for some reason expects an ISO date or an object that contains "year", "month" "day" etc:

Do you think we should also support UNIX timestamp?
It expects only ISO date or an object that contains "year", "month" "day" because I did not think about UNIX timestamp TBH, and I am not sure LLMs are good at generating UNIX timestampts.

WDYT?

G0dC0der:
I can update it to support UNIX timestamp as well as other classes from java.time package(like Instant), however, I see that this lib is looking to replace the use of Gson with ObjectMapper. Makes me think if this change is needed. My questions to you are:

1) When is the switch to jackson?
2) Are the APIs around object mapping going to change(meaning will it allow customisation)?

langchain4j:
> When is the switch to jackson?

It is in process (most modules have been migrated to Jackson already), but I do not have a concrete timeline for `langchain4j-core`, but it sohuld be in hte next couple of months

> Are the APIs around object mapping going to change(meaning will it allow customisation)?

I guess we will keep the SPI so that Quarkus (and potentially other frameworks) will be able to plug in what they want. Or do you mean an end-user-facing customization APIs (I did not think about it yet, any ideas)?

G0dC0der:
New commit that handles Instant and ZonedDateTime. 

> (I did not think about it yet, any ideas)?

The demand for exposing ObjectMapper isn't as important as it is with Gson, because of the powerful annotation support in Jackson. Either way, it would be great if we can customise the ObjectMapper instance used for both tool params and tool responses.

langchain4j:
@G0dC0der I am uncomfortable doing this change without good tests in place.
Could you please add integration tests for tool parameters of those affected types (those listed in `GsonJsonCodec`, like `LocalDate`, `LocalTime`, etc.), so that we can make sure they work both before and after this change?
Thank you!

G0dC0der:
Hi, sorry for disappearing. I am thinking we can close this, since Jackson support should be available soon? As long as we have access to the Jackson instance used to serdes tool params and responses, we're good.

langchain4j:
Hi @G0dC0der !

> I am thinking we can close this, since Jackson support should be available soon?

Yes, I hope to complete the migration to Jackson in this (end of Jan) or next (end of Feb) release.

> As long as we have access to the Jackson instance used to serdes tool params and responses, we're good.

Hmm, I am not sure this will be available by default, perhaps we still need your changes?

langchain4j:
@G0dC0der why closing it? I would keep it open untill the migration to Jackson is complete

G0dC0der:
Maybe it didn't need to be closed, but in the latest version of langchain4j, only one line of code is needed to fix this. Ideally i would want to create a new PR, that is much smaller than this.

dliubarskyi:
Superceded by https://github.com/langchain4j/langchain4j/pull/2594

PR #1597, opened 2024-08-16
fix [BUG] #1585

### LINKED ISSUE #1585: [BUG]
**Describe the bug**

image

**repro steps：**
```
Return result example(Markdown):
“\n## AA“
”\n“
”### AAA“
```

1、result = "## AA" Jump to handler.onNext(result);
2、result = "\n" Jump to if (Utils.isNullOrBlank(result)) { return;}
3、result = "### AAA" Jump to handler.onNext(result);
Due to the loss of the '\ n' character, This will cause the output format of '### AAA' to display incorrectly
I think using the isNullOrEmpty method is more reasonable

- LangChain4j version: 0.33.0
- LLM(s) used: qianfan
- Java version: 17

**Additional context**

lane12138:
## Issue
Fixes #1585


ghost (approved):
@lane12138 thank you!

Issue #1420, opened 2024-07-08
[FEATURE] Distributed Tools


d135-1r43:
**Is your feature request related to a problem? Please describe.**

Our bot currently has very many tools, most of them using a different JDBC driver, different APIs (REST, GQL, gRPC). This comes with several downsides:

* Dependency accumulation, too many data sources
* The Bot grows to be an omnipotent monolith
* API lifecycle management needs to be addressed 
* token consumption is high and will grow even further
* customizing the bot for different customers or deployments is not possible

Also, some of our tools are located on the remote servers. It means, that these tools may become unavailable during the work of the service, also some new tools may appear, or some past remote tools can be deleted. Also, we do not know descriptions and parameters of these remote tools in advance, so they cannot be attached to an AIService during the boot time or compile time.

**Describe the solution you'd like**

Tools should be distributed. A Tool can be implemented by a microservice. As soon as that microservice starts, it will register itself at the AIService; if it is unavailable, the tool can be unregistered. It should be possible to run these operations in the runtime.  

**Describe alternatives you've considered**

Right now, to make the LLM know about connected remote tools, we use `RetrievalAugmentor` that returns descriptions of the attached remote tools. Then, LLM can call a special "generic" tool that is attached traditionally as a common tool with an annotation. This tool acts like a dispatcher for all the connected remote tools. This approach is very buggy and non-deterministic and consumes tons of tokens, so it seems more like a workaround rather than a reliable, solid solution.

**Additional context**

We at EXP would be willing to contribute a PR.

langchain4j:
Hi @d135-1r43, if you need the ability to specify which tools should be available to the LLM dynamically (for each AI Service invocation), there is [this issue](https://github.com/langchain4j/langchain4j/issues/637) which can be implemented pretty easily.
So, something like this:
```java
Map<ToolSpecification, ToolExecutor> tools = new ConcurrentHashMap<>();  // you manage this map, adding and removing tools to it as you need

Assistant assistant = AiServices.builder(Assistant.class)
    .chatLanguageModel(...)
    .toolProvider((userMessage, memoryId) -> tools)
    .build();
```

`toolProvider` will be called on every invocation of `Assistant` and you can return all the tools that LLM should have access to for this specific invocation. This way, tools can change over time (you can add/remove new/existing tools in the map). You can also provide different tools for different users or even depending on the current message (e.g. depending on user intent). Would that work for you?

MiggiV2:
[...]

PR #1644, opened 2024-08-27
[FEATURE] Select tools dynamically on incoming message

### LINKED ISSUE #637: [FEATURE] Select tools dynamically on incoming message
With an Assistant interface like this:

```
interface Assistant {
    String chat(String message);
}
```

We can use the AiService like this:

```
assistant = AiServices.builder(Assistant.class)
    .streamingChatLanguageModel(model)
    .tools(tools)
    .build();
```

And then we can just call the Assistant like this:

```
assistant.chat(message);
```

Base on the provided message OpenAI will then select which tool to use, which is great!

Now, imagine there are dozens of tool specifications. These tool specifications will be sent with each chat message, which will count towards Token usage.

My question is, is there a way to select the the tools to be included in chat message, before calling OpenAI?

I was thinking of creating multiple different Assistants, but then each would have its separate Memory, so that would not work.

Would it make sense to create multiple Assistants and have all of them use the same ChatMemoryProvider?

MiggiV2:
## Issue
Closes #637 and #1420 

## Change
One can now specify tool provider for an AI Service. This provider will be called each time AI service is called and will decide which tools should be specified in the request to the LLM.

Example:
```java
    .chatLanguageModel(model)
    .toolProvider(request -> {
        if (request.userMessage().singleText().contains("booking")) {
            ToolSpecification bookingToolSpecification = ToolSpecification.builder()
                .name("get_booking_details")
                .description("Returns booking details")
                .addParameter("bookingNumber", type("string"))
                .build();
                
            return ToolProviderResult.builder()
                .add(bookingToolSpecification, new BookingToolExecutor())
                .build();
        } else {
            return null;
        }
    })
    .build();
```

## General checklist
- [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green [I don't have access to MistralAI, but every test is green for OpenAI]

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
@@ -1,8 +1,6 @@
 package dev.langchain4j.service;
 
-import dev.langchain4j.service.tool.DefaultToolExecutor;
 import dev.langchain4j.agent.tool.Tool;
-import dev.langchain4j.service.tool.ToolExecutor;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
@@ -22,6 +20,9 @@ import dev.langchain4j.rag.content.Content;
 import dev.langchain4j.rag.content.retriever.ContentRetriever;
 import dev.langchain4j.rag.content.retriever.EmbeddingStoreContentRetriever;
 import dev.langchain4j.retriever.Retriever;
+import dev.langchain4j.service.tool.DefaultToolExecutor;
+import dev.langchain4j.service.tool.ToolExecutor;
+import dev.langchain4j.service.tool.ToolProvider;
 import dev.langchain4j.spi.services.AiServicesFactory;
 
 import java.lang.reflect.Method;
@@ -301,6 +302,10 @@ public abstract class AiServices<T> {
      * @see Tool
      */
     public AiServices<T> tools(Object... objectsWithTools) {
+        if (context.toolProvider != null) {
+            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
+                    "method ToolProvider. But not both!");
+        }
         return tools(asList(objectsWithTools));
     }
 
@@ -341,6 +346,21 @@ public abstract class AiServices<T> {
     }
 
     /**
+     * Configures the tool provider that the LLM can use
+     *
+     * @param toolProvider Decides which tool or tools should be used to handle the user message
+     * @return builder
+     */
+    public AiServices<T> toolsToolProvider(ToolProvider toolProvider) {
+        if (context.toolSpecifications != null | context.toolExecutors != null) {
+            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
+                    "method ToolProvider. But not both!");
+        }
+        context.toolProvider = toolProvider;
+        return this;
+    }
+
+    /**
      * Configures the tools that the LLM can use.
      *
      * @param tools A map of {@link ToolSpecification} to {@link ToolExecutor} entries.
@@ -350,7 +370,10 @@ public abstract class AiServices<T> {
      * @return builder
      */
     public AiServices<T> tools(Map<ToolSpecification, ToolExecutor> tools) {
-
+        if (context.toolProvider != null) {
+            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
+                    "method ToolProvider. But not both!");
+        }
         if (context.toolSpecifications == null) {
             context.toolSpecifications = new ArrayList<>();
         }
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
@@ -303,8 +303,7 @@ public abstract class AiServices<T> {
      */
     public AiServices<T> tools(Object... objectsWithTools) {
         if (context.toolProvider != null) {
-            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
-                    "method ToolProvider. But not both!");
+            throw new IllegalArgumentException("Either the tools or the tool provider can be configured, but not both!");
         }
         return tools(asList(objectsWithTools));
     }
@@ -348,13 +347,12 @@ public abstract class AiServices<T> {
     /**
      * Configures the tool provider that the LLM can use
      *
-     * @param toolProvider Decides which tool or tools should be used to handle the user message
+     * @param toolProvider Decides which tools the LLM could use to handle the request
      * @return builder
      */
-    public AiServices<T> toolsToolProvider(ToolProvider toolProvider) {
+    public AiServices<T> toolProvider(ToolProvider toolProvider) {
         if (context.toolSpecifications != null | context.toolExecutors != null) {
-            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
-                    "method ToolProvider. But not both!");
+            throw new IllegalArgumentException("Either the tools or the tool provider can be configured, but not both!");
         }
         context.toolProvider = toolProvider;
         return this;
@@ -371,8 +369,7 @@ public abstract class AiServices<T> {
      */
     public AiServices<T> tools(Map<ToolSpecification, ToolExecutor> tools) {
         if (context.toolProvider != null) {
-            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
-                    "method ToolProvider. But not both!");
+            throw new IllegalArgumentException("Either the tools or the tool provider can be configured, but not both!");
         }
         if (context.toolSpecifications == null) {
             context.toolSpecifications = new ArrayList<>();
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -1,6 +1,5 @@
 package dev.langchain4j.service;
 
-import dev.langchain4j.service.tool.ToolExecution;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.AiMessage;
@@ -25,6 +24,7 @@ import dev.langchain4j.rag.AugmentationResult;
 import dev.langchain4j.rag.content.Content;
 import dev.langchain4j.rag.query.Metadata;
 import dev.langchain4j.service.output.ServiceOutputParser;
+import dev.langchain4j.service.tool.ToolExecution;
 import dev.langchain4j.service.tool.ToolExecutor;
 import dev.langchain4j.service.tool.ToolProviderRequest;
 import dev.langchain4j.service.tool.ToolProviderResult;
@@ -185,9 +185,10 @@ class DefaultAiServices<T> extends AiServices<T> {
 
                         if (returnType == TokenStream.class) {
                             List<Content> contents = augmentationResult != null ? augmentationResult.contents() : null;
-                            context.toolSpecifications = toolSpecifications;
-                            context.toolExecutors = toolExecutors;
-                            return new AiServiceTokenStream(messages, contents, context, memoryId); // TODO moderation
+                            AiServiceTokenStream aiServiceTokenStream = new AiServiceTokenStream(messages, contents, context, memoryId);
+                            aiServiceTokenStream.setToolSpecifications(toolSpecifications);
+                            aiServiceTokenStream.setToolExecutors(toolExecutors);
+                            return aiServiceTokenStream; // TODO moderation
                         }
 
                         Response<AiMessage> response;
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -4,6 +4,7 @@ import dev.langchain4j.service.tool.ToolExecution;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.SystemMessage;
 import dev.langchain4j.data.message.ToolExecutionResultMessage;
 import dev.langchain4j.data.message.UserMessage;
@@ -25,6 +26,8 @@ import dev.langchain4j.rag.content.Content;
 import dev.langchain4j.rag.query.Metadata;
 import dev.langchain4j.service.output.ServiceOutputParser;
 import dev.langchain4j.service.tool.ToolExecutor;
+import dev.langchain4j.service.tool.ToolProviderRequest;
+import dev.langchain4j.service.tool.ToolProviderResult;
 
 import java.io.InputStream;
 import java.lang.reflect.Array;
@@ -74,10 +77,11 @@ class DefaultAiServices<T> extends AiServices<T> {
             dev.langchain4j.service.UserMessage userMessage = parameter.getAnnotation(dev.langchain4j.service.UserMessage.class);
             MemoryId memoryId = parameter.getAnnotation(MemoryId.class);
             UserName userName = parameter.getAnnotation(UserName.class);
-            if (v == null && userMessage == null && memoryId == null && userName == null) {
+            boolean isToolResult = parameter.getType().equals(ToolProviderResult.class);
+            if (v == null && userMessage == null && memoryId == null && userName == null && !isToolResult) {
                 throw illegalConfiguration(
                         "Parameter '%s' of method '%s' should be annotated with @V or @UserMessage " +
-                                "or @UserName or @MemoryId", parameter.getName(), method.getName()
+                                "or @UserName or @MemoryId or be an instance of ToolProviderResult", parameter.getName(), method.getName()
                 );
             }
         }
@@ -117,6 +121,8 @@ class DefaultAiServices<T> extends AiServices<T> {
                         validateParameters(method);
 
                         Object memoryId = findMemoryId(method, args).orElse(DEFAULT);
+
+                        Optional<ToolProviderResult> toolsFromArgs = findToolsFromArguments(method, args);
 
                         Optional<SystemMessage> systemMessage = prepareSystemMessage(memoryId, method, args);
                         UserMessage userMessage = prepareUserMessage(method, args);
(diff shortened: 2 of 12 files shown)


langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -301,6 +302,10 @@ public AiServices<T> moderationModel(ModerationModel moderationModel) {
      * @see Tool
      */
     public AiServices<T> tools(Object... objectsWithTools) {
+        if (context.toolProvider != null) {
+            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
```
```suggestion
            throw new IllegalArgumentException("Either the tools or the tool provider can be configured, but not both!");
```

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -340,6 +345,21 @@ public AiServices<T> tools(List<Object> objectsWithTools) { // TODO Collection?
         return this;
     }
 
+    /**
+     * Configures the tool provider that the LLM can use
+     *
+     * @param toolProvider Decides which tool or tools should be used to handle the user message
+     * @return builder
+     */
+    public AiServices<T> toolsToolProvider(ToolProvider toolProvider) {
```
```suggestion
    public AiServices<T> toolProvider(ToolProvider toolProvider) {
```

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -340,6 +345,21 @@ public AiServices<T> tools(List<Object> objectsWithTools) { // TODO Collection?
...
+        if (context.toolSpecifications != null | context.toolExecutors != null) {
+            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
```
```suggestion
            throw new IllegalArgumentException("Either the tools or the tool provider can be configured, but not both!");
```

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -350,7 +370,10 @@ public AiServices<T> tools(List<Object> objectsWithTools) { // TODO Collection?
      * @return builder
      */
     public AiServices<T> tools(Map<ToolSpecification, ToolExecutor> tools) {
-
+        if (context.toolProvider != null) {
+            throw new IllegalArgumentException("You can either use static and this tools method or a tool provider via the " +
```
```suggestion
            throw new IllegalArgumentException("Either the tools or the tool provider can be configured, but not both!");
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderRequest.java
```diff
@@ -0,0 +1,24 @@
...
+public class ToolProviderRequest {
+    private final Object chatMemoryId;
+    private final UserMessage userMessage;
+
+    public ToolProviderRequest(Object chatMemoryId, UserMessage userMessage) {
+        this.chatMemoryId = chatMemoryId;
+        this.userMessage = userMessage;
+    }
+
+    public Object getChatMemoryId() {
```
langchain4j:
```suggestion
    public Object chatMemoryId() {
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderRequest.java
```diff
@@ -0,0 +1,24 @@
...
+        return chatMemoryId;
+    }
+
+    public UserMessage getUserMessage() {
```
langchain4j:
```suggestion
    public UserMessage userMessage() {
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
```diff
@@ -0,0 +1,22 @@
...
+ * A wrapper mostly used by the {@link ToolProvider}
+ */
+public class ToolProviderResult {
+    private final Map<ToolSpecification, ToolExecutor> tools = new HashMap<>();
+
+    public Map<ToolSpecification, ToolExecutor> getTools() {
+        return Collections.unmodifiableMap(tools);
+    }
+
+    public void add(ToolSpecification specification, ToolExecutor toolExecutor) {
```
langchain4j: All our domain objects are immutable, please change it so that all tools are provided via constructor

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
```diff
@@ -0,0 +1,22 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A wrapper mostly used by the {@link ToolProvider}
+ */
+public class ToolProviderResult {
+    private final Map<ToolSpecification, ToolExecutor> tools = new HashMap<>();
+
+    public Map<ToolSpecification, ToolExecutor> getTools() {
```
langchain4j:
```suggestion
    public Map<ToolSpecification, ToolExecutor> tools() {
```

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
```diff
@@ -681,6 +673,126 @@ void should_use_programmatically_configured_tools(ChatLanguageModel chatLanguage
...
+        assertThat(response.content().text()).contains("2027");
+
+        // No tool again
+        assistant.chat("When does my holiday 123-456 starts?");
+        verify(executor, times(1)).execute(any(), any());
+    }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    void only_static_or_tool_provider(ChatLanguageModel chatLanguageModel) {
```
langchain4j:
```suggestion
    void only_static_or_tool_provider() {
```
This test does not require testing with various (and real) models, `ChatModelMock` can be used (or `ChatLanguageModel` mocked with Mockito)

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -166,6 +172,38 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
...
+                                    ? context.chatMemory(memoryId)
+                                    : null;
+                            context.toolSpecifications = new ArrayList<>();
+                            context.toolExecutors = new HashMap<>();
+                            ToolProviderRequest providerRequest = new ToolProviderRequest(chatMemory, userMessage);
+                            ToolProviderResult providerResult = context.toolProvider.provideTools(providerRequest);
+                            if (providerResult != null) {
+                                // Overriding default tools by provider tools
+                                providerResult.getTools().forEach((toolSpecification, toolExecutor) -> {
+                                    context.toolSpecifications.add(toolSpecification);
```
langchain4j: This is not the expected behavior, tools returned in `ToolProviderResult` should be used only for the current AI Service invocation, they should not be cached in the context

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProvider.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.service.tool;
+
+/**
+ * A low-level provider of a {@link ToolProviderRequest}.
```
langchain4j:
```suggestion
 * A tool provider. It is called each time the AI service is called and supplies tools for that specific call.
 * Tools returned in {@link ToolProviderResult} will be included in the request to the LLM.
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProvider.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+@FunctionalInterface
+public interface ToolProvider {
+    /**
+     * Receives a request based on a user message and sends relevant tools back. Use can use AI for example.
```
langchain4j:
```suggestion
     * Provides tools for the request to the LLM.
```

ghost (commented):
@MiggiV2 thanks a lot, good job!

I have left some comments (don't worry, most of them are small nitpicks), please take a look.

I have also noticed that you've implemented an ability to specify a tool inside the AI service method. Would you mind to exclude this from the current PR (I am not yet sure about this exact implementation)?

Thanks!

MiggiV2:
> I have also noticed that you've implemented an ability to specify a tool inside the AI service method. Would you mind to exclude this from the current PR (I am not yet sure about this exact implementation)?

Yea sure. What implementation we are talking about exactly?
Are we talking about using tools like this:
```JAVA
assistant.chat("When does my booking 123-456 starts?", tools);
```

langchain4j:
@MiggiV2 yes, exactly!

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
MiggiV2: Sure. I would change the constructor to this:
```Java
public ToolProviderResult(Map<ToolSpecification, ToolExecutor> tools) {
        this.tools = tools;
}
```

But then we get more boilerplate code.
For example:
```JAVA
Map<ToolSpecification, ToolExecutor> toolMap = new HashMap<>();
toolMap.put(specification, executor);
ToolProviderResult bookingResult = new ToolProviderResult(toolMap);
 ```
 
Is this okay, or do we want to improve this?
I thought of something with varargs, but this didn't work as expected.

Comment (langchain4j) [reply langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java]: What about:
```java
ToolProviderResult.builder()
    .add(spec1, executor1)
    .add(spec2, executor2)
    ...
    .build();
```

But there should be a `.add(Map<Spec,Executor>)` as well.

MiggiV2: Update: I have just discussed this with @d135-1r43, and he thinks it's okay to write a little bit more boilerplate code.
[This was before your reply 😅]

MiggiV2: The builder pattern is a way cleaner solution. I have added your idea and @d135-1r43 reviewed my code.
You find my commit [here](https://github.com/langchain4j/langchain4j/pull/1644/commits/c80710140c4242be7bb07d616c9aa862a76996a0).
WDYT?

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -340,6 +344,20 @@ public AiServices<T> tools(List<Object> objectsWithTools) { // TODO Collection?
         return this;
     }
 
+    /**
+     * Configures the tool provider that the LLM can use
+     *
+     * @param toolProvider Decides which tool or tools should be used to handle the user message
```
```suggestion
     * @param toolProvider Decides which tools the LLM could use to handle the request
```

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -340,6 +344,20 @@ public AiServices<T> tools(List<Object> objectsWithTools) { // TODO Collection?
         return this;
     }
 
+    /**
+     * Configures the tool provider that the LLM can use
```
langchain4j, line 348:
```suggestion
     * Configures the tool provider that decides which tools the LLM can use.
```

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -166,11 +169,34 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
                             return new AiServiceTokenStream(messages, contents, context, memoryId); // TODO moderation
```
langchain4j: streaming case is left without toolProvideer support, is it intentional?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -166,11 +169,34 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
...
+                        Map<String, ToolExecutor> toolExecutors = context.toolExecutors;
+
+                        if (context.toolProvider != null) {
+                            // Use provider to decide matching tools
+                            ChatMemory chatMemory = context.hasChatMemory()
+                                    ? context.chatMemory(memoryId)
+                                    : null;
+                            toolSpecifications = new ArrayList<>();
+                            toolExecutors = new HashMap<>();
+                            ToolProviderRequest providerRequest = new ToolProviderRequest(chatMemory, userMessage);
```
langchain4j:
```suggestion
                            ToolProviderRequest providerRequest = new ToolProviderRequest(memoryId, userMessage);
```

`chatMemory` above can be removed

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -166,11 +169,34 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
...
+                            ToolProviderResult providerResult = context.toolProvider.provideTools(providerRequest);
+                            if (providerResult != null) {
+                                // Overriding default tools by provider tools
```
langchain4j:
```suggestion
```
outdated comment

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderRequest.java
```diff
@@ -0,0 +1,24 @@
+package dev.langchain4j.service.tool;
+
+import dev.langchain4j.data.message.UserMessage;
+
+/**
+ * A wrapper mostly used for the {@link ToolProvider}
```
langchain4j:
```suggestion
 * A request to the {@link ToolProvider}
```

langchain4j: Hovewer the comment is mostly useless, feel free to remove it

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
```diff
@@ -0,0 +1,44 @@
+package dev.langchain4j.service.tool;
+
+import dev.langchain4j.agent.tool.ToolSpecification;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A wrapper mostly used by the {@link ToolProvider}
```
langchain4j: same here

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public class ToolProviderResult {
+    private final Map<ToolSpecification, ToolExecutor> tools;
+
+    public ToolProviderResult(Map<ToolSpecification, ToolExecutor> tools) {
+        this.tools = tools;
```
langchain4j:
```suggestion
        this.tools = copyIfNotNull(tools);
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public Map<ToolSpecification, ToolExecutor> tools() {
+        return Collections.unmodifiableMap(tools);
```
langchain4j:
```suggestion
        return tools;
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public static ToolProviderResultBuilder builder() {
+        return new ToolProviderResultBuilder();
+    }
+
+    public static class ToolProviderResultBuilder {
```
langchain4j:
```suggestion
    public static class Builder {
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
```diff
@@ -0,0 +1,44 @@
...
+        private final Map<ToolSpecification, ToolExecutor> tools = new HashMap<>();
```
langchain4j:
```suggestion
        private final Map<ToolSpecification, ToolExecutor> tools = new LinkedHashMap<>();
```
Better to preserve the ordering defined by the user

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolProviderResult.java
```diff
@@ -0,0 +1,44 @@
...
+
+        public ToolProviderResultBuilder add(ToolSpecification tool, ToolExecutor executor) {
+            tools.put(tool, executor);
+            return this;
+        }
+
+        public ToolProviderResultBuilder add(Map<ToolSpecification, ToolExecutor> tools) {
```
langchain4j:
```suggestion
        public ToolProviderResultBuilder addAll(Map<ToolSpecification, ToolExecutor> tools) {
```
my bad

ghost (commented):
@MiggiV2 thanks a lot!

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
MiggiV2: Oh, no. This was not intended.
I will look into this.

MiggiV2: What do you think about my solution, overriding the context toolSpecifications and toolExecutors?
[DefaultAiServices.java](https://github.com/langchain4j/langchain4j/pull/1644/commits/39ea9f41efc3732d9a2a2c4f0bc8708a8ad79d55#diff-1d724bf189ebfdb3d76eaefc0f9f1768b01e615ae3521ade0ea4fe4c8d774bd2R179)

langchain4j: @MiggiV2 although simple, it does not seem right to override global tools with local (for this request only) ones.
What do you think about modifying `AiServiceTokenStream` and `AiServiceStreamingResponseHandler` to accept a list of tool specifications instead?

MiggiV2: Sounds like a more and clean solution. I was also experimenting with this, but didn't want to change so much on the code base.
I will modify both classes and will think of backwards compatibility.

langchain4j: @MiggiV2 ok, thanks! Do not worry about backwards compatibility, those 2 classes are internal, noone should use them directly. Adding an extra list to both constructors should be ok

langchain4j:
@MiggiV2 how is it going? Don't want to rush you, but do you have an estimate? I am asking because I plann to do a release tomorrow...

MiggiV2:
> @MiggiV2 how is it going? Don't want to rush you, but do you have an estimate? I am asking because I plann to do a release tomorrow...

My main focus is on langchain4j again. I will modify `AiServiceTokenStream` and `AiServiceStreamingResponseHandler` now and push in 60min - 90min. Then, there everything should be fine from my side ;D

MiggiV2:
So. In my last changes, I don't override any context attributes anymore. Instead, new attributes were added with setters.
Should be it from my side. Excited to hear your feedback 😄 

Tests are green on my machine for OpenAI. I don't have access to MistralAI or AzureAI, these are unknown.

MiggiV2:
Just synced with main ⬆️

ghost (approved):
@MiggiV2 thanks a lot, good job!

PR #1645, opened 2024-08-27
re #1506 Enable Maven enforcer plugin and fix dependency conflict introduced by `okhttp` dependency in 19 modules.

PrimosK:
## Issue

#1506

## Change

Resolved version conflict:
```
[ERROR] Rule 0: org.apache.maven.enforcer.rules.dependency.DependencyConvergence failed with message:
[ERROR] Failed while enforcing releasability.
[ERROR]
[ERROR] Dependency convergence error for org.jetbrains.kotlin:kotlin-stdlib-jdk8:jar:1.9.10 paths to dependency are:
[ERROR] +-dev.langchain4j:langchain4j-ollama:jar:0.34.0-SNAPSHOT
[ERROR]   +-com.squareup.okhttp3:okhttp:jar:4.12.0:compile
[ERROR]     +-com.squareup.okio:okio:jar:3.6.0:compile
[ERROR]       +-com.squareup.okio:okio-jvm:jar:3.6.0:compile
[ERROR]         +-org.jetbrains.kotlin:kotlin-stdlib-jdk8:jar:1.9.10:compile
[ERROR] and
[ERROR] +-dev.langchain4j:langchain4j-ollama:jar:0.34.0-SNAPSHOT
[ERROR]   +-com.squareup.okhttp3:okhttp:jar:4.12.0:compile
[ERROR]     +-org.jetbrains.kotlin:kotlin-stdlib-jdk8:jar:1.8.21:compile
[ERROR]
[ERROR] -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :langchain4j-ollama
```

... caused by 'okhttp' dependency and enabled Maven enforcer plugin in the following modules:

- LangChain4j :: Integration :: Anthropic
- LangChain4j :: Integration :: ChatGLM
- LangChain4j :: Integration :: Chroma
- LangChain4j :: Integration :: CloudFlare Workers AI
- LangChain4j :: Integration :: Cohere
- LangChain4j :: Integration :: DashScope
- LangChain4j :: Integration :: Hugging Face
- LangChain4j :: Integration :: Jina
- LangChain4j :: Integration :: Judge0
- LangChain4j :: Integration :: MistralAI
- LangChain4j :: Integration :: Nomic
- LangChain4j :: Integration :: OVHcloud AI
- LangChain4j :: Integration :: Ollama
- LangChain4j :: Integration :: Qianfan
- LangChain4j :: Integration :: Vearch
- LangChain4j :: Integration :: Vespa
- LangChain4j :: Integration :: Zhipu AI
- LangChain4j :: Web Search Engine :: SearchApi
- LangChain4j :: Web Search Engine :: Tavily

## Note

Please note that [issue ](https://github.com/square/okhttp/issues/8288) for this was already created in `httpok` repository but it will not be fixed in 4.x. It's reportedly already tackled in version 5.x.

With that in mind I suggest we apply temporary changes proposed in this PR. After upgrading to `httpok` 5.x we will be able to remove these.

## Tests

`mvn clean test` passed

### DIFF
--- a/langchain4j-qianfan/pom.xml
+++ b/langchain4j-qianfan/pom.xml
@@ -30,14 +30,22 @@
             <artifactId>converter-gson</artifactId>
         </dependency>
 
+        <!-- DEPENDENCY CONFLICT RESOLUTION FOR OKHTTP (START) -->
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>okhttp</artifactId>
         </dependency>
+
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>okhttp-sse</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib-jdk8</artifactId>
+        </dependency>
+        <!-- DEPENDENCY CONFLICT RESOLUTION FOR OKHTTP (END)  -->
 
         <dependency>
             <groupId>org.projectlombok</groupId>
@@ -65,6 +73,31 @@
             <scope>test</scope>
         </dependency>
     </dependencies>
+
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-enforcer-plugin</artifactId>
+                    <version>3.5.0</version>
+                    <executions>
+                        <execution>
+                            <id>enforce</id>
+                            <configuration>
+                                <rules>
+                                    <dependencyConvergence/>
+                                </rules>
+                            </configuration>
+                            <goals>
+                                <goal>enforce</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+    </build>
 
     <licenses>
         <license>
--- a/code-execution-engines/langchain4j-code-execution-engine-judge0/pom.xml
+++ b/code-execution-engines/langchain4j-code-execution-engine-judge0/pom.xml
@@ -26,10 +26,17 @@
             <artifactId>langchain4j-core</artifactId>
         </dependency>
 
+        <!-- DEPENDENCY CONFLICT RESOLUTION FOR OKHTTP (START) -->
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>okhttp</artifactId>
         </dependency>
+
+        <dependency>
+            <groupId>org.jetbrains.kotlin</groupId>
+            <artifactId>kotlin-stdlib-jdk8</artifactId>
+        </dependency>
+        <!-- DEPENDENCY CONFLICT RESOLUTION FOR OKHTTP (END)  -->
 
         <dependency>
             <groupId>org.junit.jupiter</groupId>
@@ -86,4 +93,28 @@
 
     </dependencies>
 
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-enforcer-plugin</artifactId>
+                    <version>3.5.0</version>
+                    <executions>
+                        <execution>
+                            <id>enforce</id>
+                            <configuration>
+                                <rules>
+                                    <dependencyConvergence/>
+                                </rules>
+                            </configuration>
+                            <goals>
+                                <goal>enforce</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+    </build>
 </project>
--- a/langchain4j-anthropic/pom.xml
+++ b/langchain4j-anthropic/pom.xml
@@ -38,6 +38,7 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
 
+        <!-- DEPENDENCY CONFLICT RESOLUTION FOR OKHTTP (START) -->
         <dependency>
             <groupId>com.squareup.okhttp3</groupId>
             <artifactId>okhttp</artifactId>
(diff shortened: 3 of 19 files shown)


ghost (approved):
@PrimosK thanks a lot!

boris-petrov:
I believe this commit broke my project. I've added the `langchain4j` BOM: `dependencies.platform('dev.langchain4j:langchain4j-bom:0.34.0')` and I've added a dependency which transitively depends on `okhttp3`. However, this commit removes `kotlin-stdlib-jdk8` which is a needed dependency - and that breaks. It's not added for some reason. Why is that? What can I do about it?

boris-petrov:
To give a bit more detail. I have two projects in my Gradle. I add this BOM to both for simplicity, even though only one of them uses `langchain4j`. The one that uses it uses these three dependencies: `dev.langchain4j:langchain4j`, `dev.langchain4j:langchain4j-embeddings`, `dev.langchain4j:langchain4j-qdrant`. Neither of them adds `kotlin-stdlib-jdk8`. And also no one adds `kotlin-stdlib-jdk8` to the project that only uses the BOM but nothing other from `langchain4j`. So both projects are broken. I don't think `langchain4j` should remove dependencies because it breaks use-cases like mine.

PrimosK:
You are correct that this commit excludes the conflicting dependency introduced by `com.squareup.okhttp3:okhttp`. **However**, it also explicitly defines `org.jetbrains.kotlin:kotlin-stdlib-jdk8` in each module where it was excluded, including in the [parent module's](https://github.com/PrimosK/langchain4j/blob/9b8896602c549fe64ed41da2124d06ef5fb74fdb/langchain4j-parent/pom.xml#L196) dependency management section.

In theory, this should ensure that the correct version (1.9.10) of `kotlin-stdlib-jdk8` is included in your project.

_Why is that? What can I do about it?_
Honestly I am not (yet) sure why this happens. But for now I think you could solve it by including `kotlin-stdlib-jdk8` dependency explicitly - i.e like:
```
dependency>
    <groupId>org.jetbrains.kotlin</groupId>
    <artifactId>kotlin-stdlib-jdk8</artifactId>
    <version>1.9.10</version>
</dependency>
```

Side note:
To better understand and resolve the problem, it would be helpful if you could provide a minimal Maven project that reproduces the issue. This way, we can analyze the dependency tree and configuration to determine why the expected version of `kotlin-stdlib-jdk8` is not being included as intended.

boris-petrov:
[Here you go](https://github.com/boris-petrov/langchain4j-bug).

Clone the repo, run `./gradlew dependencies`. You'll see `org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.10` in it.

Check `build.gradle`. Uncomment `implementation platform('dev.langchain4j:langchain4j-bom:0.34.0')`. Run `./gradlew dependencies` again. See that there is no `kotlin-stdlib-jdk8`. This simulates one of my projects.

Uncheck the three `langchain4j` dependencies. Run the command. Again there is no `kotlin-stdlib-jdk8`. That simulates my other project. Both are broken.

PrimosK:
Thank you for providing this information. I was able to reproduce it.

I've created another [project ](https://github.com/PrimosK/langchain4j-test-app/tree/main)in Maven which is using the same set of dependencies you're using.

`mvn dependency:tree` results in such [tree](https://github.com/PrimosK/langchain4j-test-app/blob/main/dependency-tree.txt).

So it seems that `org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.10` is properly included in this case:

image

This experiment shows that this might theorhetically be related to Maven/Gradle dependency resolution differences. Given that I am not proeficient in Gradle any hint of why this is happening is welcome.

Side question - does explicitly including `org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.9.10` solves it for you?

boris-petrov:
@PrimosK thanks for taking the time to look into it!

Yes, adding it explicitly fixes the issue, as expected. I'll probably use this workaround until the real problem is resolved.

As for the problem itself - unfortunately I'm also not sure why would Gradle do that. It seems like a strange behavior to me.

boris-petrov:
@PrimosK if that's not something you can easily fix, do you want me to open a new issue so it's tracked?

PrimosK:
@boris-petrov, yes...

I think that makes sense.

One theoretically possible solution would be to exclude `org.jetbrains.kotlin:kotlin-stdlib-jdk8` in each of the `Langchain4j` submodules, rather than in the dependency management section of the parent module (as we do it now). However, I'm not sure we want to go this route, as the issue you've experienced occurs only in rare cases (when a BOM is used + Langchain4j submodules without the OkHttp dependency + a dependency that has a transitive `OkHttp` dependency + using Gradle).

It seems that the best way forward is to wait for this [issue](https://github.com/square/okhttp/issues/8288) to be resolved or for `OkHttp 5.0` to be released, where the dependency conflict is already resolved.

Anyway, having a separate issue with a dedicated description and sample might still be handy, as someone from the Gradle community could offer advice on a solution we aren't aware of.

langchain4j-qianfan/pom.xml
```diff
@@ -66,6 +74,31 @@
...
+                                <rules>
+                                    <dependencyConvergence/>
+                                </rules>
+                            </configuration>
+                            <goals>
+                                <goal>enforce</goal>
+                            </goals>
+                        </execution>
+                    </executions>
+                </plugin>
```
kpavlov, line 97: It makes sense to move this to parent pom and use [`enforcer.skipRules`](https://maven.apache.org/enforcer/maven-enforcer-plugin/enforce-mojo.html#rulesToSkip) property to skip rules for some modules meanwhile. Eventually removing`enforcer.skipRules`.

PR #1646, opened 2024-08-27
feat: Metadata filtering Qdrant

### LINKED ISSUE #1600: [FEATURE] Support of filtering by metadata for Qdrant in EmbeddingStore
We are implementing the chat platform with multiple users. The code is based on langhcain4j and Qdrant to store and augment the future chats based on historical chats. In order to do this we need to search the previous messages and filter them by user id and some other fields from Qdrant collection. Currently I see the support for vector metadata search for Milvus and Elasticsearch but not Qdrant.

My team needs the possibility to filter the results of vector search with metadata. E.g. call to QdrantEmbeddingStore.search(EmbeddingSearchRequest) with filter added to EmbeddingSearchRequest instance.
The alternatives we consider are the separate collections for distinct user (not scaleable) or switching to Milvius (we chose Qdrant as a result of comparison so it would be easier to stay and implement this feature rather than switching)

My team and me are ready to contribute and implement this feature. 
This was also discussed in #151

Anush008:
## Description

- Filtering support for Qdrant with tests for the converter.
- Updated metadata storage handling to support complex metadata types. No longer limited to `Map<String, String>`.

Resolves #1600 

## NOTE

In Qdrant,
- Eq, NEq, In, NIn conditions don't allow float and double values. Only integers and strings.
- LT, GT, LTE, GTE allow only numbers, not alphabets.
- For In and NIn conditions, if the key doesn't exist in the metadata, it is not matched.


langchain4j-qdrant/src/test/java/dev/langchain4j/store/embedding/qdrant/QdrantEmbeddingStoreIT.java
```diff
@@ -29,22 +29,22 @@ class QdrantEmbeddingStoreIT extends EmbeddingStoreIT {
   private static QdrantEmbeddingStore embeddingStore;
 
   @Container
-  private static final QdrantContainer qdrant = new QdrantContainer("qdrant/qdrant:latest");
+  private static final QdrantContainer qdrant = new QdrantContainer("qdrant/qdrant:v1.11.1");
```
langchain4j: Is there a reason to change it to a specific version?

langchain4j-qdrant/src/test/java/dev/langchain4j/store/embedding/qdrant/QdrantEmbeddingStoreIT.java
```diff
@@ -1,78 +1,1308 @@
...
+    }
+
+    @Override
+    protected void clearStore() {
+        embeddingStore.clearStore();
+    }
+
+    @ParameterizedTest
+    @MethodSource
+    void should_filter_by_metadata_not(Filter metadataFilter,
```
langchain4j: please extend this test class from `EmbeddingStoreWithFilteringIT` instead of duplicating test cases

ghost (commented):
@Anush008 thank you!

langchain4j-qdrant/src/test/java/dev/langchain4j/store/embedding/qdrant/QdrantEmbeddingStoreIT.java
Anush008: It's the latest.

Anush008: We'd not want some future version of randomly breaking Langchain4J's CI.

langchain4j-qdrant/src/test/java/dev/langchain4j/store/embedding/qdrant/QdrantEmbeddingStoreIT.java
Anush008: I've explained in the note of the PR description.

langchain4j-qdrant/src/test/java/dev/langchain4j/store/embedding/qdrant/QdrantEmbeddingStoreIT.java
langchain4j: Ideally this test should make sure LC4j implementation works well with Qdrant, including the latest version. So I would revert this change. It should fail as soon as possible so that we can start fixing it asap...

Anush008: Done.

langchain4j-qdrant/src/test/java/dev/langchain4j/store/embedding/qdrant/QdrantEmbeddingStoreIT.java
langchain4j: Sorry, I did not notice it. Could you do something similar to what we did in `ChromaEmbeddingStoreIT.should_filter_by_metadata` and `ChromaEmbeddingStoreIT.should_filter_by_metadata_not`?

Anush008: I tried that too.
But there a quite a lot of cases to exclude.
So it made more sense to me to write the ones to include rather than exclude.

langchain4j: @Anush008 it is hard to maintain duplicated test cases, please follow the same approach as in `ChromaEmbeddingStoreIT`. It should be simple given criterias are quite straightforward:

> Eq, NEq, In, NIn don't allow float and double values. Only integers and strings.
> LT, GT, LTE, GTE allow only numbers, not alphabets.
> For In and NIn conditions, if the key doesn't exist in the metadata, it is not matched.

holeniev: @langchain4j just FYI we've tested this PR in our environment and filtering now works fine for our cases

langchain4j: @holeniev that's great! But please rewrite the test to reduce the duplication. Thank you!

Anush008: Thanks for checking @holeniev. I'll refactor the tests.

Anush008: Done.
[`f3fb841` (#1646)](https://github.com/langchain4j/langchain4j/pull/1646/commits/f3fb84142faef096bc0c60418e9c637f4b54764d)

ghost (approved):
@Anush008 thank you!

Issue #1640, opened 2024-08-26
[FEATURE] Supports sorting of data in the Elasticsearch vector database


t19980620:
When Elasticsearch is used as a vector database to retrieve data, it is not possible to sort the Elasticsearch data. For example, sorting by the age field

langchain4j:
cc @dadoonet (do you mind if I'll ping you on Elasticsearch issues?)

dadoonet:
Sure! Please do or assign me 😊

dadoonet:
About this feature request: I guess this is something which needs some update in the `EmbeddingSearchRequest` class. We would need to add some sort orders on a list of fields. In Elasticsearch we support things like:

```java
.sort(so -> so
        .geoDistance(gd -> gd
            .field("location")
            .location(
                    new GeoLocation.Builder()
                            .latlon(ll -> ll.lat(49.0404).lon(2.0174))
                            .build()
            )
        .order(SortOrder.Asc)
        .unit(DistanceUnit.Kilometers)
)
```

How should we design some generic sorting in LangChain4j? Any thoughts?

langchain4j:
@dadoonet good question. Perharps sorting can be configured in the builder of `ElasticsearchEmbeddingStore` for the moment. And, if/when it will be required in other stores, we can think about adding it to the common API, WDYT?

dadoonet:
Yeah. We can do that probably. Although that will be very "static" if the user wants to change the order at request time.

Think of an e-commerce website where the user want dynamically order by price, note or date. We probably don't want to recreate an instance of the embedding store for each call/user.

langchain4j:
@dadoonet if dynamism is required, `ElasticsearchEmbeddingStore.search(EmbeddingSearchRequest, Sorting)` or similar can be added.

@t19980620 could you please share a bit more info on your use case? Thanks!

dadoonet:
I'd keep `ElasticsearchEmbeddingStore.search(EmbeddingSearchRequest)` but change the `EmbeddingSearchRequest` object to add a `sort` field and its builders.

```java
EmbeddingSearchRequest.builder()
  .queryEmbedding(queryEmbedding)
  .sort(Sort.builder()
      .field("foo").order(Sort.ASC)
      .build())
  .build());
```

With that, we don't have to change all the existing code. embedding stores will "just" have to modify their implementation of `search(EmbeddingSearchRequest)` when they want to support sorting.

t19980620:
[...]

PR #1649, opened 2024-08-28
Add numCandidates setting to ElasticsearchConfigurationKnn

dadoonet:
## Issue

Related to #1640

## Change

We can now support a new setting for the `ElasticsearchConfigurationKnn`. It allows to define how many nearest neighbor candidates to consider per shard while doing knn search.

```java
EmbeddingStore<TextSegment> embeddingStore = ElasticsearchEmbeddingStore.builder()
        .configuration(ElasticsearchConfigurationKnn.builder()
                .withNumCandidates(10)
                .build())
        .restClient(restClient)
        .build();
```

## General checklist

### DIFF
--- a/langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchConfigurationKnn.java
+++ b/langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchConfigurationKnn.java
@@ -17,7 +17,7 @@ import java.io.IOException;
  */
 public class ElasticsearchConfigurationKnn extends ElasticsearchConfiguration {
     private static final Logger log = LoggerFactory.getLogger(ElasticsearchConfigurationKnn.class);
-    private final int numCandidates;
+    private final Integer numCandidates;
 
     public static class Builder {
         private Integer numCandidates;
@@ -45,7 +45,7 @@ public class ElasticsearchConfigurationKnn extends ElasticsearchConfiguration {
     }
 
 
-    private ElasticsearchConfigurationKnn(int numCandidates) {
+    private ElasticsearchConfigurationKnn(Integer numCandidates) {
         this.numCandidates = numCandidates;
     }
 
@@ -61,7 +61,7 @@ public class ElasticsearchConfigurationKnn extends ElasticsearchConfiguration {
             krb.filter(ElasticsearchMetadataFilterMapper.map(embeddingSearchRequest.filter()));
         }
 
-        if (numCandidates > 0) {
+        if (numCandidates != null) {
             krb.numCandidates(numCandidates);
         }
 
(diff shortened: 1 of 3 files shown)


langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchConfigurationKnn.java
```diff
@@ -12,23 +12,41 @@
...
+
+        /**
+         * The number of nearest neighbor candidates to consider per shard while doing knn search.
+         * Cannot exceed 10,000. Increasing num_candidates tends to improve the accuracy of the final
+         * results.
+         *
+         * @param numCandidates The number of nearest neighbor candidates to consider per shard
+         * @return the builder instance
+         */
+        public Builder withNumCandidates(int numCandidates) {
```
langchain4j:
```suggestion
        public Builder numCandidates(Integer numCandidates) {
```

langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchConfigurationKnn.java
```diff
@@ -12,23 +12,41 @@
...
- * using the Knn implementation.
+ * using the approximate kNN query implementation.
+ * @see <a href="https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-knn-query.html#knn-query-top-level-parameters">kNN query</a>
  */
 public class ElasticsearchConfigurationKnn extends ElasticsearchConfiguration {
     private static final Logger log = LoggerFactory.getLogger(ElasticsearchConfigurationKnn.class);
+    private final int numCandidates;
 
     public static class Builder {
+        private int numCandidates = 0;
```
langchain4j:
```suggestion
        private Integer numCandidates;
```
minor: what about using `Integer` instead of `int`. This way we know if user configured a value or not (`numCandidates != null`)

langchain4j-elasticsearch/src/test/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchEmbeddingStoreKnnWithConfigurationIT.java
```diff
@@ -0,0 +1,106 @@
...
+
+    @AfterAll
+    static void stopServices() throws IOException {
+        elasticsearchClientHelper.stopServices();
+    }
+
+    @BeforeEach
+    void createEmbeddingStore() throws IOException {
+        indexName = randomUUID();
+        elasticsearchClientHelper.removeDataStore(indexName);
```
langchain4j: minor: does it make sense to call remove here? `indexName` has just been generated on the line before

ghost (commented):
@dadoonet thank you!

langchain4j-elasticsearch/src/test/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchEmbeddingStoreKnnWithConfigurationIT.java
dadoonet: Correct. Previously I was using the test method name but here it's random. So it should work without needing to remove the index first :)

ghost (approved):
@dadoonet thank you!

PR #1695, opened 2024-09-03
Google AI Gemini integration

### LINKED ISSUE #1363: Support for Google Gemini AI
Hi team,

First of all thanx for this library, Its great for working with langchain in Java and Kotlin, But i have one question here why we have just added the support for models for Google Cloud Vertex AI Gemini , But not Google Gen AI Gemini without google cloud as we have in original langchain?

Is there any reason, Why we are just using google cloud version only that is enterprise version as mentioned in langchain docs but not normal Gen Ai version of Gemini?

My point is becoz Gemini Ai has free of charge plan, but google cloud vertex Gemini Ai has only pay as u go plan, thats why for testing and development purposes we can utilize free of charge plan. But here we dont.

Thanx

glaforge:
## Issue

Closes #1363

## Change

New module for the Gemini AI model

## General checklist

- [x] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)


langchain4j-bom/pom.xml
```diff
@@ -147,6 +147,12 @@
                 <version>${project.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>dev.langchain4j</groupId>
+                <artifactId>langchain4j-gemini</artifactId>
```
langchain4j:
```suggestion
                <artifactId>langchain4j-google-ai-gemini</artifactId>
```

langchain4j-core/src/main/java/dev/langchain4j/internal/CustomMimeTypesFileTypeDetector.java
```diff
@@ -0,0 +1,180 @@
...
+ *     this class allows you to define your own extension to mime-type mapping,
+ *     in addition to providing a few extra ones built-in.
+ * </p>
+ * <p>
+ *     When the mime-type can't be found in those mappings, or from JDK's built-in,
+ *     it tries, as a fallback, to use <code>URLConnection</code>'s capability,
+ *     but it's slower as it needs to connect to the underlying URL.
+ * </p>
+ */
+public class CustomMimeTypesFileTypeDetector extends FileTypeDetector {
```
langchain4j, line 34: should it be in core? I would move it to the module where it is used (test as well)

langchain4j-google-ai-gemini/pom.xml
```diff
@@ -0,0 +1,142 @@
...
+            <groupId>com.squareup.retrofit2</groupId>
+            <artifactId>adapter-rxjava3</artifactId>
+            <version>2.11.0</version>
+        </dependency>
+-->
+
+        <!-- Lombok for @Data and @Builder -->
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
```
langchain4j, line 64: we are trying to get rid of lombok, please do not use it any more in the new code

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FinishReasonMapper.java
```diff
@@ -0,0 +1,28 @@
...
+            case BLOCKLIST:
+            case PROHIBITED_CONTENT:
+            case RECITATION:
+            case SPII:
+            case SAFETY:
+            case LANGUAGE:
+                return FinishReason.CONTENT_FILTER;
+            case MAX_TOKENS:
+                return FinishReason.LENGTH;
+            case MALFORMED_FUNCTION_CALL:
```
langchain4j: it is malformed, should it be mapped to `TOOL_EXECUTION`?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FunctionMapper.java
```diff
@@ -0,0 +1,79 @@
...
+        if (specifications == null) {
+            return null;
+        }
+
+        GeminiTool.GeminiToolBuilder tool = GeminiTool.builder();
+
+        List<GeminiFunctionDeclaration> functionDeclarations = specifications.stream()
+            .map(specification -> {
+                // using Gemini's built-in hosted Python sandboxed code executor
+                if (specification.name().equals("code_execution")) {
```
langchain4j: what if user has his own tool called `code_execution`?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FunctionMapper.java
```diff
@@ -0,0 +1,79 @@
...
+                if (specification.name() != null) {
+                    fnBuilder.name(specification.name());
+                }
+                if (specification.description() != null) {
+                    fnBuilder.description(specification.description());
+                }
+                if (specification.parameters() != null) {
+                    ToolParameters parameters = specification.parameters();
+
+                    final String[] propName = {""};
```
langchain4j: how does this work?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FunctionMapper.java
```diff
@@ -0,0 +1,79 @@
...
+            tool.functionDeclarations(functionDeclarations);
+        }
+
+        return tool.build();
+    }
+
+    public static List<ToolExecutionRequest> fromToolExecReqToGFunCall(List<GeminiFunctionCall> functionCalls) {
+        return functionCalls.stream().map(functionCall -> ToolExecutionRequest.builder()
+            .name(functionCall.getName())
+            .arguments(new Gson().toJson(functionCall.getArgs()))
```
langchain4j: minor: Gson could be a static constant

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
```diff
@@ -0,0 +1,254 @@
...
+import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotBlank;
+import static dev.langchain4j.model.gemini.PartsAndContentsMapper.fromMessageToGContent;
+import static dev.langchain4j.model.gemini.FinishReasonMapper.fromGFinishReasonToFinishReason;
+import static dev.langchain4j.model.gemini.PartsAndContentsMapper.fromGPartsToAiMessage;
+import static dev.langchain4j.model.gemini.SchemaMapper.fromJsonSchemaToGSchema;
+import static java.util.Collections.emptyList;
+
+@Experimental
+public class GeminiAiChatLanguageModel implements ChatLanguageModel {
```
langchain4j:
```suggestion
public class GoogleAiGeminiChatModel implements ChatLanguageModel {
```
major

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
```diff
@@ -0,0 +1,254 @@
...
+    public static final String GEMINI_AI_ENDPOINT = "https://generativelanguage.googleapis.com/v1beta/";
```
langchain4j:
```suggestion
    private static final String GEMINI_AI_ENDPOINT = "https://generativelanguage.googleapis.com/v1beta/";
```

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
```diff
@@ -0,0 +1,254 @@
...
+    private final List<String> stopSequences;
+
+    private final Integer candidateCount;
+
+    private final String responseMimeType;
+    private final ResponseFormat responseFormat;
+
+    private final GeminiFunctionCallingConfig toolConfig;
+
+    private final Boolean logRequestsAndResponses;
```
langchain4j: would be nice to have separate `logRequests` and `logResponses`

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
```diff
@@ -0,0 +1,254 @@
...
+    private final Double temperature;
+    private final Integer topK;
+    private final Double topP;
+    private final Integer maxOutputTokens;
+    private final List<String> stopSequences;
+
+    private final Integer candidateCount;
+
+    private final String responseMimeType;
+    private final ResponseFormat responseFormat;
```
langchain4j: `ResponseFormat` is supposed to be provided inside the `ChatRequest`, not via constructor

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
```diff
@@ -0,0 +1,254 @@
...
+                                     Integer maxOutputTokens, Integer candidateCount,
+                                     String responseMimeType, ResponseFormat responseFormat,
+                                     List<String> stopSequences, GeminiFunctionCallingConfig toolConfig,
+                                     Boolean logRequestsAndResponses,
+                                     List<GeminiSafetySetting> safetySettings
+    ) {
+        this.apiKey = ensureNotBlank(apiKey, "apiKey");
+        this.modelName = ensureNotBlank(modelName, "modelName");
+
+        // using Gemini's default values
```
langchain4j: if these values are not mandatory and are defined on Google AI Gemini side, please do not specify them here, let them be null

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
```diff
@@ -0,0 +1,254 @@
...
+        this.temperature = getOrDefault(temperature, 1.0);
+        this.topK = getOrDefault(topK, 64);
+        this.topP = getOrDefault(topP, 0.95);
+        this.maxOutputTokens = getOrDefault(maxOutputTokens, 8192);
+        this.candidateCount = getOrDefault(candidateCount, 1);
+        this.stopSequences = getOrDefault(stopSequences, emptyList());
+        this.toolConfig = toolConfig;
+
+        this.safetySettings = safetySettings;
```
langchain4j:
```suggestion
        this.safetySettings = copyIfNotNull(safetySettings);
```

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAttributionSourceId.java
```diff
@@ -0,0 +1,11 @@
+package dev.langchain4j.model.gemini;
+
+import lombok.Builder;
+import lombok.Data;
+
+@Data
+@Builder
+public class GeminiAttributionSourceId {
```
langchain4j: there are a lot of public classes, should they all be public? Please check them and make internal classes package-private

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
```diff
@@ -0,0 +1,254 @@
+package dev.langchain4j.model.gemini;
```
langchain4j:
```suggestion
package dev.langchain4j.model.googleai;
```

major: to be consistent with `dev.langchain4j.model.vertexai`

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiFunctionCall.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Builder;
+import lombok.Data;
+
+import java.util.Map;
+
+@Data
+@Builder
+public class GeminiFunctionCall {
+    private String name;
+    private Map args;
```
langchain4j: please add generic type info

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiFunctionResponse.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.Builder;
+import lombok.Data;
+
+import java.util.Map;
+
+@Data
+@Builder
+public class GeminiFunctionResponse {
+    private String name;
+    private Map response;
```
langchain4j: here as well

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiGenerationConfig.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.List;
+
+@Data
+@Builder
+public class GeminiGenerationConfig {
+    private List<String> stopSequences;
+    private String responseMimeType;
+    private GeminiSchema responseSchema;
+    private int candidateCount = 1;
```
langchain4j: please do not use primitives and default values (use object types with default null)

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiService.java
```diff
@@ -0,0 +1,32 @@
...
+import retrofit2.http.POST;
+import retrofit2.http.Path;
+import retrofit2.http.Query;
+import retrofit2.http.Headers;
+//import retrofit2.http.Streaming;
+
+interface GeminiService {
+
+    @POST("models/{model}:generateContent")
+    @Headers("LLM-Library: LangChain4j")
```
langchain4j: major: What about using `User-Agent` intead of `LLM-Library`? (in all other places as well)

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiService.java
```diff
@@ -0,0 +1,32 @@
...
+    Call<GeminiGenerateContentResponse> generateContent(
+        @Path("model") String modelName,
+        @Query("key") String apiKey,
```
langchain4j: major: I am almost sure API key should not be transmitted via query, it is unsafe.
It is also logged:
```
Sept 04, 2024 10:24:07 AM okhttp3.internal.platform.Platform log
INFO: --> POST https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent?key=fdsgjasd8of83..sdwsdfw4fwdfs (redacted)
```

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiService.java
```diff
@@ -0,0 +1,32 @@
...
+        @Body GeminiGenerateContentRequest request);
+
+/*
+    @Streaming
+    @POST("models/{model}:streamGenerateContent")
+    @Headers("LLM-Library: LangChain4j")
+    Observable<GeminiGenerateContentResponse> streamGenerateContent(
+        @Path("model") String modelName,
+        @Query("key") String apiKey,
```
langchain4j: same here

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/PartsAndContentsMapper.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    static AiMessage fromGPartsToAiMessage(List<GeminiPart> parts) {
+        StringBuilder fullText = new StringBuilder();
+        List<GeminiFunctionCall> functionCalls = new ArrayList<>();
+
+        for (GeminiPart part : parts) {
+            GeminiExecutableCode executableCode = part.getExecutableCode();
+            if (executableCode != null) { //TODO maybe add a flag to the model like showCodeExecution()
```
langchain4j: yes, flag is a good idea

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/PartsAndContentsMapper.java
```diff
@@ -0,0 +1,257 @@
...
+                        AiMessage aiMessage = (AiMessage) msg;
+
+                        if (aiMessage.hasToolExecutionRequests()) {
+                            return GeminiContent.builder()
+                                .role(GeminiRole.MODEL.toString())
+                                .parts(((AiMessage) msg).toolExecutionRequests().stream()
+                                    .map(toolExecutionRequest -> GeminiPart.builder()
+                                        .functionResponse(GeminiFunctionResponse.builder()
+                                            .name(toolExecutionRequest.name())
+                                            .response(new Gson().fromJson(toolExecutionRequest.arguments(), Map.class))
```
langchain4j: minor: Gson can be a constant

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+        ChatResponse response = gemini.chat(ChatRequest.builder()
+            .messages(UserMessage.from("What is a large language model?"))
+            .build());
+
+        // then
+        String text = response.aiMessage().text();
+        assertThat(text).containsIgnoringCase("artificial intelligence");
+        assertThat(text).containsIgnoringCase("neural network");
+
+        assertThat(response.finishReason()).isEqualTo(FinishReasonMapper.fromGFinishReasonToFinishReason(GeminiFinishReason.STOP));
```
langchain4j: Why not just
```suggestion
        assertThat(response.finishReason()).isEqualTo(STOP);
```

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+    void should_respect_system_instruction() {
+        // given
+        GeminiAiChatLanguageModel gemini = GeminiAiChatLanguageModel.builder()
+            .apiKey(GEMINI_AI_KEY)
+            .modelName("gemini-1.5-flash")
+            .build();
+
+        // when
+        Response<AiMessage> response = gemini.generate(
+            SystemMessage.from("Translate fromToolExecReqToGFunCall English into French"),
```
langchain4j:
```suggestion
            SystemMessage.from("Translate from English into French"),
```

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+            ToolExecutionResultMessage.from(null, "getFirstNFibonacciNumbers", "[0, 1, 1, 2, 3, 5, 8, 13, 21, 34]");
+        allMessages.add(forecastResult);
+
+        // then
+        response = gemini.generate(
+            allMessages
+        );
+
+        // then
+        assertThat(response.content().text()).contains("[0, 1, 1, 2, 3, 5, 8, 13, 21, 34]");
```
langchain4j:
```suggestion
        assertThat(response.content().text()).isEqualTo("[0, 1, 1, 2, 3, 5, 8, 13, 21, 34]");
```

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+
+        // when
+        ChatResponse response = gemini.chat(ChatRequest.builder()
+            .messages(UserMessage.from("What is a large language model?"))
+            .build());
+
+        // then
+        String text = response.aiMessage().text();
+        assertThat(text).containsIgnoringCase("artificial intelligence");
+        assertThat(text).containsIgnoringCase("neural network");
```
langchain4j: could be flaky... what about something like "What is the capital of France"? Something that has a single correct answer

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+            .topK(10)
+            .maxOutputTokens(10)
+            .build();
+
+        // when
+        String response = gemini.generate(
+            "How much is 3+4? Reply with just the answer");
+
+        // then
+        assertThat(response.trim()).isEqualTo("7");
```
langchain4j: Does not seem to test what it is supposed to

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+            .build();
+
+        // when
+        Response<AiMessage> response = gemini.generate(
+            UserMessage.from("What is the firstname of the John Doe?\n" +
+                "Reply in JSON following with the following format: {\"firstname\": string}"));
+
+        // then
+        String jsonText = response.content().text();
+        assertThat(jsonText).contains("\"firstname\"");
```
langchain4j:
```suggestion
        assertThat(jsonText).isEqualToIgnoringWhitespace("{\"firstname\": \"John\"}");
```

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+        assertThat(jsonText).contains("\"John\"");
```
langchain4j:
```suggestion
```

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+
+        assertThat(response.tokenUsage().inputTokenCount()).isEqualTo(25);
```
langchain4j: please do not hardcode concrete numbers, they will change in the future

```java
assertThat(tokenUsage.inputTokenCount()).isGreaterThan(0);
assertThat(tokenUsage.outputTokenCount()).isGreaterThan(0);
assertThat(tokenUsage.totalTokenCount())
                .isEqualTo(tokenUsage.inputTokenCount() + tokenUsage.outputTokenCount());
```

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+import static dev.langchain4j.model.chat.request.ResponseFormatType.JSON;
+import static dev.langchain4j.model.chat.request.json.JsonStringSchema.JSON_STRING_SCHEMA;
+import static dev.langchain4j.model.gemini.GeminiAiChatLanguageModel.pythonCodeExecution;
+import static dev.langchain4j.model.gemini.GeminiHarmBlockThreshold.BLOCK_LOW_AND_ABOVE;
+import static dev.langchain4j.model.gemini.GeminiHarmCategory.HARM_CATEGORY_HARASSMENT;
+import static dev.langchain4j.model.gemini.GeminiHarmCategory.HARM_CATEGORY_HATE_SPEECH;
+import static java.util.Collections.singletonList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@ExtendWith(RetryExtension.class)
```
langchain4j: major: something is wrong with the test execution. even if the test assertion fails, junit says it is green

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+    void should_support_text_file() {
+        // given
+        GeminiAiChatLanguageModel gemini = GeminiAiChatLanguageModel.builder()
+            .apiKey(GEMINI_AI_KEY)
+            .modelName("gemini-1.5-flash")
+            .build();
+
+        // when
+        Response<AiMessage> response = gemini.generate(UserMessage.userMessage(
+            TextFileContent.from(new String(Base64.getEncoder().encode(readBytes(MD_FILE_URL))), "text/markdown"),
```
langchain4j: I had an impression that `TextFileContent` was introduced in order to specify file as a URL, but if I change it to `TextFileContent.from(MD_FILE_URL)`, it fails

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
```diff
@@ -0,0 +1,597 @@
...
+        response = gemini.generate(
+            allMessages
+        );
+
+        // then
+        assertThat(response.content().text()).containsIgnoringCase("sunny");
+    }
+
+    @Test
+    void should_support_JSON_array_in_tools() {
```
langchain4j: not sure it tests what it should... Perharps `ResponseFormat` should be specified in the request?

ghost (commented):
@glaforge thanks a lot, good job!

langchain4j-core/src/main/java/dev/langchain4j/internal/CustomMimeTypesFileTypeDetector.java
glaforge: I kept it in core, but also modified the Vertex AI Gemini module to make use of it

langchain4j-google-ai-gemini/pom.xml
glaforge: OK, I won't do it in this PR, I'll look into that after the release.
So shall we do manual builders, or use something else?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FinishReasonMapper.java
glaforge: Well spotted, I changed to match to other, as there's no equivalent.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FunctionMapper.java
glaforge: I haven't touched this one.
I could use a more specific name instead of the name that Gemini uses?
Maybe `gemini_code_execution`?  (which is hopefully a name no user would use)

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
glaforge: Why removing this line? To weaken the assertions a bit?

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
glaforge: I've removed my custom assertion, and used one from JUnit Pioneer (just a test dependency)
I'm using Pioneer's `@RetryingTest(n)` annotation now, which works at a method test level rather than class+method level.
It's for flaky tests which sometimes fail because Gemini replies a bit differently. That's the case for example for parallel too calling or safety settings tests, as Gemini doesn't 100% reliably behave the same way, once in a while.

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
glaforge: The variants of `*Content` classes don't yet fully work with just URLs.
Gemini has a file upload API that needs to be supported, to first upload files there, and then to point at the specific URL from that place. So external free-form URLs are not supported by Gemini out of the box.
But I'll try to see after the release what I can do here to improve this, to let users pass any URL and not just special URLs.

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
glaforge: I discovered that you can't have function calling and response schema used at the same time.
So I can't set both.
Here the test was to check that JSON arrays can be used in tools.
But I'm gonna add a test to check that you can use JSON arrays as response schema too.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FunctionMapper.java
glaforge: Hmm maybe not the cleanest, but I need to pass data (the property name) between two steps of the streap.
I guess I could've created some intermediary object to hold both property name and the schema.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FunctionMapper.java
glaforge: OK

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
glaforge: I'm relying on Retrofit and OkHttp's own logging, and it doesn't offer the choice of logging either the input or the output, so I went with logging both.
I guess it could be done, but I'd have to write my own interceptor, which I wanted to avoid as it's not necessarily trivial.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModel.java
glaforge: I set the response format (if provided) in the constructor, but if someone pass it via `ChatRequest`, I use the one from the request of course.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FunctionMapper.java
langchain4j: Could you please rewrite it in some more simple way? It is hard to understand what's going on

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiAttributionSourceId.java
glaforge: Good point, only a few need to be public (as they're used or reachable from the constructor for users to configure)

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
langchain4j: but the tool does not use json array, it returns a String that has a json array in it, but it is treated as text everywhere

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/GeminiService.java
glaforge: There's a special header: `x-goog-api-key`

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
langchain4j: please check suggestion for the previous line, "John" is checked there

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/gemini/GeminiAiChatLanguageModelIT.java
langchain4j: suggested assertion actually tests that returned string is JSON

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/gemini/FunctionMapper.java
langchain4j: Do I understand that `code_execution` enables sandbox code execution on Gemini side? If yes, perharps it makes more sense to have a flag in ctor like "enableCodeExecution" than expecting user to pass `ToolSpecification` with specific name? Also it does not seem right, because there is actually no tool on the client side, it is executed automatically on the LLM side

langchain4j-google-ai-gemini/pom.xml
langchain4j: Manual builders I guess

langchain4j-core/src/main/java/dev/langchain4j/data/message/ContentType.java
```diff
@@ -9,6 +9,10 @@ public enum ContentType {
      * Text content.
      */
     TEXT(TextContent.class),
+    /**
+     * Text content.
```
langchain4j:
```suggestion
     * Text file content.
```

ghost (dismissed):
@glaforge thank you a lot!

sunildhiman90:
@glaforge @langchain4j , Thanku for this support. When will be new version released with this google gemini AI support?

glaforge:
@sunildhiman90 There should be a release this week, if all goes well.

Your feedback will be welcome, as it's a brand new model, freshly implemented.

A few notes:
* Currently, there's just a non-streaming model available (till I figure out how to call the streaming model via Retrofit/OkHttp)
* The cached content feature of Gemini is not supported yet. We need to figure how to surface that as a generic feature for LangChain4j, so that other models can fit in the mold
* If you're sending multimodal content (images, PDF, etc), you have to include the Base64 encoded content. The URL variants expect a special URL as files have to be uploaded first via a special API. I'll see if we can support that transparently

sunildhiman90:
Thanks @glaforge , Sure i will try it out.

glaforge:
@sunildhiman90 Please have a look at my [article](https://glaforge.dev/posts/2024/09/05/new-gemini-model-in-langchain4j/) to have a good idea of what you can do with this new module.

sunildhiman90:
Thanks @glaforge , I will checkout that.🙌

sunildhiman90:
Hi @glaforge , I was trying to use the gemini-1.5-flash model in ktor, But im getting this error:

`FAILED_PRECONDITION (code 400) User location is not supported for the API use.`

But the same key is working fine for my android app. What can be the reasons?
PS: I followed your article for testing this.

glaforge:
@sunildhiman90 From which `location` are you calling the API?
I know that the Gemini web app is not available in all countries, so there may be similar restrictions for the API?
Your Android phone, or your server IP addresses are not the same, and maybe from the phone it doesn't appear to be coming from a location where the API might be restricted?

sunildhiman90:
Hi @glaforge Im calling it From India, and i can see, this region is already available in supported regions: 
https://ai.google.dev/gemini-api/docs/available-regions

glaforge:
@sunildhiman90 Strange. India is definitely supported, as the page you mentioned confirms.
I'm gonna ask internally at Google to see if I can get some suggestions to unblock you.
I've no clue for now 😞

glaforge:
Maybe you can add a bit more context about the machine/server/device where you get those errors? 
Searching on the web led me to some stackoverflow [questions](https://stackoverflow.com/questions/76367740/error-message-failedprecondition-400-user-location-is-not-supported-for-the-ap) where someone had the error when using IPv6 instead of IPv4, or another user was behind a VPN that was likely not in the supported list of countries.
Another [question](https://discuss.ai.google.dev/t/generate-object-error-user-location-is-not-supported-for-the-api-use/37440/7) was mentioning some similar problems in some regions when running inside Firebase functions.

Anyhow, more information is needed to better understand what's going on.

sunildhiman90:
Hi @glaforge , I was using the vpn and didnt noticed that. Thanx its working fine now. Thanx a lot again for providing this support in langchain4j.🙌

Issue #1722, opened 2024-09-07
[FEATURE] Allow us to override METADATA_TEXT_SEGMENT in the WeaviateEmbeddingStoreExample


Jaland:
**Is your feature request related to a problem? Please describe.**
We have loaded data into a Weaviate DB but we have loaded it with the key `page_content`. When we used the python libraries we were able to specify that key with the following code:

```
WeaviateRetrieverWithScoreThreshold(
      client=WeaviateClient(self.config).client,
      index_name=index_name,
      # TODO: Should we make this data driven? For different data types?
      text_key="page_content",
      embedding_function=self.embeddings,
      k=4, # can we make this dynamic?
    )
```
But when using the WeaviateEmbeddingStoreExample it seems like it default to `text` and there is no way to change that

**Describe the solution you'd like**
I believe this is based on the `METADATA_TEXT_SEGMENT` that seems to be hard coded (linked below). 

It would just be nice if this was a value we were able to override.

**Additional context**
Here is the code I am referencing:
https://github.com/langchain4j/langchain4j/blob/main/langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java#L42C33-L42C54

Jaland:
Created a PR:
https://github.com/langchain4j/langchain4j/pull/1723

Jaland:
I have also noticed that any meta data is assumed to be under `_metadata` so it seems like it may not be possible to retrieve fields directly under properties.

So inside of my vector DB we have something like:

```
{
	"class": "ClassName",
	"creationTimeUnix": 1718292601895,
	"id": "00cd5667-623e-471f-996c-5f68f585bd15",
	"lastUpdateTimeUnix": 1718292601895,
	"properties": {
		"header1": "Example Header 1",
		"header2": "Example Header 2",
		"language": "en-US",
		"page_content": "CONTENT YADA YADA YADA.......................................",
		"source": "https://my-example-documentation.com",
		"title": "Getting started...",
		"version": "1.1"
	},
	"vectorWeights": null
}
```

And I would want to be able to capture version/source as part of the metadata I care about

langchain4j:
Hi @Jaland, `WeaviateEmbeddingStore` was not specifically designed to be compatible with python's version. Feel free to make changes to make it more configurable though

PR #1723, opened 2024-09-07
Allow user to modify text key

Jaland:
## Issue
[1722](https://github.com/langchain4j/langchain4j/issues/1722)

## Change
Parameterized the key used to look up the text value on weaviate. It should be 

## General checklist

- [x ] There are no breaking changes
- [? ] I have added unit and integration tests for my change (Based on the test that are there I am not really sure what to add that would really test it well)
- [x ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [x ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

### DIFF
--- a/langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
+++ b/langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
@@ -67,7 +67,7 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
      * @param useGrpcForInserts Use GRPC instead of HTTP for batch inserts only. <b>You still need HTTP configured for search</b>
      * @param securedGrpc       The GRPC connection is secured
      * @param grpcPort          The port, e.g. 50051. This parameter is optional.
-     * @param textFieldName     The key in metadata that contains the text. Default is "text".
+     * @param textFieldName     The name of the field that contains the text of a {@link TextSegment}. Default is "text".
      */
     @Builder
     public WeaviateEmbeddingStore(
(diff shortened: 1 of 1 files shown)


Jaland:
I was also able to test/validate this works with my local code.

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -80,7 +81,8 @@ public WeaviateEmbeddingStore(
             String objectClass,
             Boolean avoidDups,
             String consistencyLevel,
-            Collection<String> metadataKeys
+            Collection<String> metadataKeys,
+            String textKey
```
langchain4j:
```suggestion
            String textFieldName
```

ghost (commented):
@Jaland thanks!

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
Jaland: Yeah I like that name a lot better than textKey, let me fix that now

Jaland:
Field name updated

Jaland:
I also have a secondary branch I want to create a PR for if this gets accepted for updating the metadata field, but it is a little more involved.

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -67,6 +67,7 @@ public class WeaviateEmbeddingStore implements EmbeddingStore<TextSegment> {
      * @param useGrpcForInserts Use GRPC instead of HTTP for batch inserts only. <b>You still need HTTP configured for search</b>
      * @param securedGrpc       The GRPC connection is secured
      * @param grpcPort          The port, e.g. 50051. This parameter is optional.
+     * @param textFieldName     The key in metadata that contains the text. Default is "text".
```
langchain4j:
```suggestion
     * @param textFieldName     The name of the field that contains the text of a {@link TextSegment}. Default is "text".
```

ghost (dismissed):
@Jaland thank you!

PR #1800, opened 2024-09-19
TextClassifier return score and support type beyond Enum

### LINKED ISSUE #1665: [FEATURE] EmbeddingModelTextClassifier to support Map<String, List<String>>
I would like the EmbeddingModelTextClassifier to support Map<String, List<String>> as input, rather than only Map<Enum, List<String>>, and to provide the similarity score for each output result.

Martin7-1:
## Issue

Closes #1665 

## Change

1. Add method `classifyWithScore` in `TextClassifier`.
2. Remove `<E extends Enum<E>>` to support more type rather than only enum.

## General checklist


langchain4j/src/main/java/dev/langchain4j/classification/LabelWithScore.java
```diff
@@ -0,0 +1,20 @@
...
+
+    private final E label;
+    private final double score;
+
+    public LabelWithScore(E label, double score) {
+        this.label = label;
+        this.score = score;
+    }
+
+    public E getLabel() {
```
langchain4j:
```suggestion
    public E label() {
```

langchain4j/src/main/java/dev/langchain4j/classification/LabelWithScore.java
```diff
@@ -0,0 +1,20 @@
...
+        return label;
+    }
+
+    public double getScore() {
```
langchain4j:
```suggestion
    public double score() {
```

langchain4j/src/main/java/dev/langchain4j/classification/LabelWithScore.java
```diff
@@ -0,0 +1,20 @@
+package dev.langchain4j.classification;
+
+public class LabelWithScore<E> {
```
langchain4j:
```suggestion
public class LabelWithScore<L> {
```

(`L` is for label type). Could you pelase rename `E` into `L` everywhere?

langchain4j/src/main/java/dev/langchain4j/classification/TextClassifier.java
```diff
@@ -12,13 +13,45 @@
  */
 public interface TextClassifier<E extends Enum<E>> {
```
langchain4j:
```suggestion
public interface TextClassifier<L> {
```

Shouldn't it support `String` as requested in the issue?

langchain4j/src/main/java/dev/langchain4j/classification/LabelWithScore.java
langchain4j: I am also wondering if we need to rename the class since it might later become `LabelWithScoreWithXwithYWithZ...`

ghost (commented):
@Martin7-1 thank you, looks good!

langchain4j/src/main/java/dev/langchain4j/classification/LabelWithScore.java
Martin7-1: > I am also wondering if we need to rename the class since it might later become `LabelWithScoreWithXwithYWithZ...`

LOL. Maybe rename it to `LabelResult`? `classifyWithScore` may also need to rename.

langchain4j/src/main/java/dev/langchain4j/classification/EmbeddingModelTextClassifier.java
```diff
@@ -6,15 +6,9 @@
...
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
-import static dev.langchain4j.internal.ValidationUtils.ensureBetween;
-import static dev.langchain4j.internal.ValidationUtils.ensureGreaterThanZero;
-import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+import java.util.*;
+
+import static dev.langchain4j.internal.ValidationUtils.*;
```
glaforge: nit: could we avoid collapsing imports with stars, and keep each individual import instead?

Martin7-1: Done.

langchain4j/src/main/java/dev/langchain4j/classification/TextClassifier.java
Martin7-1: Done.

langchain4j/src/main/java/dev/langchain4j/classification/LabelWithScore.java
Martin7-1: Now I rename it to `ClassifyResult`, and rename the method to `classifyWithDetail`.

langchain4j/src/main/java/dev/langchain4j/classification/LabelWithScore.java
Martin7-1: Done.

langchain4j/src/main/java/dev/langchain4j/classification/LabelWithScore.java
Martin7-1: Done.

langchain4j/src/main/java/dev/langchain4j/classification/TextClassifier.java
```diff
@@ -4,29 +4,62 @@
...
-public interface TextClassifier<E extends Enum<E>> {
+public interface TextClassifier<L> {
+
+    /**
+     * Classify the given text with score.
+     *
+     * @param text Text to classify.
+     * @return A list of classification categories and detailed results
+     */
+    List<ClassifyResult<L>> classifyWithDetail(String text);
```
langchain4j: I am wondering if instead of a `List<ClassifyResult<L>>` we should return `ClassificationResult` that will contain labels and scores in some form (can be either `List<ScoredLabel<L>>` or `Map<L, Double>` or smth else). This way we could add more information aside from labels and scores. For example, if LLM is used under the hood for classification, we could also return `TokenUsage`, etc. WDYT?

Martin7-1: Sounds reasonable, applied.

langchain4j:
@Martin7-1 thanks a lot!
I did a few cosmetic changes, hope you don't mind.
Could you please review? 🙏

Martin7-1:
@langchain4j Thank you! All changes look good to me.

ghost (approved):

langchain4j:
@Martin7-1 thank you!

PR #1807, opened 2024-09-20
Support for GitHub Models using the Azure AI Inference API

### LINKED ISSUE #1719: [FEATURE] Add support for GitHub Models
GitHub Models is a new feature provided by GitHub, which can be found at https://github.com/marketplace/models

It allows to access a variety of models using a GitHub API key, and they are running behind the scenes on Azure OpenAI.

It makes it easy for all GitHub users to use those models, and to support them here is my proposal:

- Create a new "langchain4j-github-models" Maven module
- Have this module depend on "langchain4j-azure-open-ai": as it's the same system underneath, that should simplify a lot testing and dependency management
- Implement a specific mechanism for authentication (it's just a bearer token), and use the endpoint "https://models.inference.ai.azure.com" -> this should be most of the work
- Have a list of supported models to help people configure their model names

If all goes well, the resulting API would be something like:

```java
GitHubModelsChatModel.builder()
                .aipKey(gitHubApiKey)
                .deploymentName(GitHubModelsCahtModelName.PHI_3_5_MINI_INSTRUCT)
                .build();
```

Once this is all done, I'll try to have this in the official documentation, when you click on `Get Started` at https://github.com/marketplace/models/azureml/Phi-3-5-mini-instruct as they have currently 0 Java documentation.

jdubois:
Fix #1719

This adds GitHub Models (see https://github.com/marketplace/models ) support with the new Azure AI Inference API Java SDK.

Documentation: https://docs.langchain4j.dev/integrations/language-models/github-models


langchain4j-github-models/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+    <dependencies>
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-ai-inference</artifactId>
```
maxandersen, line 26: I thought GitHub models exposed openai apis - can't we reuse existing api ? Or is there a value add of adding  such dependency?

jdubois: Have a look at https://github.com/marketplace/models/azure-openai/gpt-4o in "get started", and check the API in JavaScript (Java will be added soon): you have the choice between OpenAI and the Azure AI Inference API.
So yes, the OpenAI models should work for accessing OpenAI models.
Then, for https://github.com/marketplace/models/azureml-mistral/Mistral-Nemo you have the choice between the Mistral AI SDK and the Azure AI Inference API.
And for https://github.com/marketplace/models/azureml/Phi-3-5-mini-instruct it will be only the Azure AI Inference API.
So the Azure AI Inference API is the only one which should work across all models. I agree, differences are probably very light, but that might change.

n3kh3l: In "Get Started," if we check the API specifications for REST—whether it's "gpt-4o," "Mistral-Nemo," or "Phi-3.5-mini-instruct"—the API specifications are the same for all and are compatible with OpenAI's API, which uses a Bearer token. Therefore, if you use langchain4j's OpenAiChatModel API, it will work with all the supported GitHub models. The same code snippet I mentioned in issue(https://github.com/langchain4j/langchain4j/issues/1719#issuecomment-2361872630)  works for all the GitHub models you referenced.

Now, regarding the "Azure AI Inference API," you’re correct that it may not currently support Java. However, langchain4j's OpenAiChatModel API alone suffices for your requirements to use GitHub models. Internally, it does not rely on any SDKs like the Mistral AI SDK or the Azure AI Inference SDK.

Aside from this, is there any specific requirement that langchain4j's OpenAiChatModel API cannot address?

jdubois: Oh yes: it’s not done by the companies doing those models, so you can’t rely on them for maintenance, support, new features… Most companies will refuse to use the non official SDK, and I can confirm this on the Azure OpenAI side where I have the LangChain4J OpenAI and the LangChain4J Azure OpenAI stats, which is exactly the same situation - nearly everyone uses the official SDK.
The Java version of the Azure AI Inference API had its first release yesterday.

jdubois: I will also have access to the GitHub Models stats, so I’ll be able to tell which implementation people use.
If this implementation isn’t popular, you can trust me that I will remove this module.

n3kh3l: > Oh yes: it’s not done by the companies doing those models, so you can’t rely on them for maintenance, support, new features… Most companies will refuse to use the non official SDK, and I can confirm this on the Azure OpenAI side where I have the LangChain4J OpenAI and the LangChain4J Azure OpenAI stats, which is exactly the same situation - nearly everyone uses the official SDK.
> The Java version of the Azure AI Inference API had its first release yesterday.

I agree, we should use the official SDK.

jdubois:
I've added support for all the current models:
- ChatModel
- StreamingChatModel
- EmbeddingModel

I also added tests for everything.

There are a couple of things which aren't perfect:
- Function calling only works with GPT-4. According to https://thenewstack.io/a-comprehensive-guide-to-function-calling-in-llms/ , Mistral and Cohere should support it too, but it doesn't work. I'm guessing this is because they use a different API, and that the Azure AI Inference SDK doesn't support it yet.
- TokenUsage for StreamingChatModel is always 0. As it works for ChatModel, I believe this is currently an API limitation.

To run the tests, you need 2 environment variables:
- `GITHUB_TOKEN` : your GitHub authentication token
- `CI_DELAY_SECONDS_GITHUB_MODELS`: a delay in order not to have rate limitations. I personally put it at 5 seconds.

All tests should be green, but I'll need to do another test tomorrow, as currently I've exhausted my quotas!!

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
```diff
@@ -0,0 +1,424 @@
...
+
+/**
+ * Represents a language model, hosted on GitHub Models, that has a chat completion interface, such as gpt-4o.
+ * <p>
+ * Mandatory parameters for initialization are: gitHubToken (the GitHub Token used for authentication) and modelName (the name of the model to use).
+ * You can also provide your own ChatCompletionsClient instance, if you need more flexibility.
+ * <p>
+ * The list of models, as well as the documentation and a playground to test them, can be found at https://github.com/marketplace/models
+ */
+public class GitHubModelsChatModel implements ChatLanguageModel {
```
langchain4j, line 44: What about `GitHubChatModel`? (no strong preference)

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
```diff
@@ -0,0 +1,424 @@
...
+                                 Double temperature,
+                                 Double topP,
+                                 List<String> stop,
+                                 Double presencePenalty,
+                                 Double frequencyPenalty,
+                                 Long seed,
+                                 ChatCompletionsResponseFormat responseFormat,
+                                 List<ChatModelListener> listeners) {
+
+        this.modelName = getOrDefault(modelName, DEFAULT_CHAT_MODEL_NAME);
```
langchain4j:
```suggestion
        this.modelName = ensureNotBlank(modelName, "modelName");
```
We are moving away from default values for model name (because they are becoming outdated pretty fast), please make it mandatory without default

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
```diff
@@ -0,0 +1,424 @@
...
+        this.maxTokens = maxTokens;
+        this.temperature = getOrDefault(temperature, 0.7);
```
langchain4j:
```suggestion
        this.temperature = temperature;
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
```diff
@@ -0,0 +1,424 @@
...
+                    logger.warn("Exception while calling model listener", e);
+                }
+            });
+
+            return response;
+        } catch (HttpResponseException httpResponseException) {
+            logger.info("Error generating response, {}", httpResponseException.getValue());
+            FinishReason exceptionFinishReason = contentFilterManagement(httpResponseException, "content_filter");
+            Response<AiMessage> response = Response.from(
+                    aiMessage(httpResponseException.getMessage()),
```
langchain4j, line 203: This does not seem right, we should throw an exception instead of returning error message in the field where user would expect LLM generated text to be

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
```diff
@@ -0,0 +1,424 @@
...
+                                 Double frequencyPenalty,
+                                 Long seed,
+                                 ChatCompletionsResponseFormat responseFormat,
+                                 List<ChatModelListener> listeners) {
+
+        this.modelName = getOrDefault(modelName, DEFAULT_CHAT_MODEL_NAME);
+        this.maxTokens = maxTokens;
+        this.temperature = getOrDefault(temperature, 0.7);
+        this.topP = topP;
+        this.stop = stop;
```
langchain4j:
```suggestion
        this.stop = copyIfNotNull(stop);
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
```diff
@@ -0,0 +1,566 @@
...
+
+/**
+ * Represents a language model, hosted on GitHub Models, that has a chat completion interface, such as gpt-4o.
+ * <p>
+ * Mandatory parameters for initialization are: gitHubToken (the GitHub Token used for authentication) and modelName (the name of the model to use).
+ * You can also provide your own ChatCompletionsClient and ChatCompletionsAsyncClient instance, if you need more flexibility.
+ * <p>
+ * The list of models, as well as the documentation and a playground to test them, can be found at https://github.com/marketplace/models
+ */
+public class GitHubModelsStreamingChatModel implements StreamingChatLanguageModel {
```
langchain4j, line 44: same comments as in `GitHubModelsChatModel` above

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
```diff
@@ -0,0 +1,566 @@
...
+                                          boolean useAsyncClient,
+                                          List<ChatModelListener> listeners,
+                                          String userAgentSuffix,
+                                          Map<String, String> customHeaders) {
+
+        this(modelName, maxTokens, temperature, topP, stop, presencePenalty, frequencyPenalty, seed, responseFormat, listeners);
+        if (useAsyncClient) {
+            this.asyncClient = setupAsyncClient(endpoint, serviceVersion, gitHubToken, timeout, maxRetries, proxyOptions, logRequestsAndResponses, userAgentSuffix, customHeaders);
+        } else {
+            this.client = setupSyncClient(endpoint, serviceVersion, gitHubToken, timeout, maxRetries, proxyOptions, logRequestsAndResponses, userAgentSuffix, customHeaders);
```
langchain4j: Does it make sense to have sync client? AFAIR it was not working properly for streaming

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingResponseBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+                if (toolCall.getFunction().getArguments() != null) {
+                    toolExecutionRequestBuilder.argumentsBuilder.append(toolCall.getFunction().getArguments());
+                }
+            }
+        }
+    }
+
+    public Response<AiMessage> build() {
+
+        String content = contentBuilder.toString();
```
langchain4j, line 101: Sometimes LLM can produce both text content **and** tool calls, so it is worth checking it. Here is an example of how to handle it:

```java
private Response<AiMessage> build() {

                String text = contentBuilder.toString();
                TokenUsage tokenUsage = new TokenUsage(inputTokenCount.get(), outputTokenCount.get());
                FinishReason finishReason = toFinishReason(stopReason);
                Map<String, Object> metadata = createMetadata();

                if (toolExecutionRequestBuilderMap.isEmpty()) {
                    return Response.from(
                            AiMessage.from(text),
                            tokenUsage,
                            finishReason,
                            metadata
                    );
                } else {
                    List<ToolExecutionRequest> toolExecutionRequests = toolExecutionRequestBuilderMap
                            .values().stream()
                            .map(AnthropicToolExecutionRequestBuilder::build)
                            .collect(toList());

                    AiMessage aiMessage = isNullOrBlank(text)
                            ? AiMessage.from(toolExecutionRequests)
                            : AiMessage.from(text, toolExecutionRequests);

                    return Response.from(
                            aiMessage,
                            tokenUsage,
                            finishReason,
                            metadata
                    );
                }
            }
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingResponseBuilder.java
```diff
@@ -0,0 +1,146 @@
...
+        if (!content.isEmpty()) {
+            return Response.from(
+                    AiMessage.from(content),
+                    new TokenUsage(inputTokenCount, outputTokenCount),
+                    finishReasonFrom(finishReason)
+            );
+        }
+
+        String toolName = toolNameBuilder.toString();
+        if (!toolName.isEmpty()) {
```
langchain4j: How is this scenario different from tools in the map below?

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/InternalGitHubModelHelper.java
```diff
@@ -0,0 +1,365 @@
...
+            chatCompletionsClientBuilder.credential(new AzureKeyCredential(gitHubToken));
+        } else {
+            throw new IllegalArgumentException("GitHub token is a mandatory parameter for connecting to GitHub models.");
+        }
+
+        return chatCompletionsClientBuilder;
+
+    }
+
+    public static ModelServiceVersion getModelServiceVersion(String serviceVersion) {
```
langchain4j: minor: does it make sense to accept `serviceVersion` as `String` if there is no way to use a value outside of `ModelServiceVersion `enum? What about accepting `ModelServiceVersion` from user?

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/InternalGitHubModelHelper.java
```diff
@@ -0,0 +1,365 @@
...
+            ToolExecutionResultMessage toolExecutionResultMessage = (ToolExecutionResultMessage) message;
+            return new ChatRequestToolMessage(toolExecutionResultMessage.text(), toolExecutionResultMessage.id());
+        } else if (message instanceof SystemMessage) {
+            SystemMessage systemMessage = (SystemMessage) message;
+            return new ChatRequestSystemMessage(systemMessage.text());
+        } else if (message instanceof UserMessage) {
+            UserMessage userMessage = (UserMessage) message;
+            ChatRequestUserMessage chatRequestUserMessage;
+            if (userMessage.hasSingleText()) {
+                chatRequestUserMessage = new ChatRequestUserMessage(((TextContent) userMessage.contents().get(0)).text());
```
langchain4j: nit:
```suggestion
                chatRequestUserMessage = new ChatRequestUserMessage(userMessage.singleText());
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/InternalGitHubModelHelper.java
```diff
@@ -0,0 +1,365 @@
...
+                            if (content instanceof TextContent) {
+                                String text = ((TextContent) content).text();
+                                return new ChatMessageTextContentItem(text);
+                            } else if (content instanceof ImageContent) {
+                                ImageContent imageContent = (ImageContent) content;
+                                if (imageContent.image().url() == null) {
+                                    throw new IllegalArgumentException("Image URL is not present. Base64 encoded images are not supported at the moment.");
+                                }
+                                ChatMessageImageUrl imageUrl = new ChatMessageImageUrl(imageContent.image().url().toString());
+                                return new ChatMessageImageContentItem(imageUrl);
```
langchain4j, line 171: This is missing: `imageUrl.setDetail(ChatMessageImageDetailLevel.fromString(imageContent.detailLevel().name()));`

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/InternalGitHubModelHelper.java
```diff
@@ -0,0 +1,365 @@
...
+                        })
+                        .collect(toList()));
+            }
+            return chatRequestUserMessage;
+        } else {
+            throw new IllegalArgumentException("Unsupported message type: " + message.type());
+        }
+    }
+
+    private static String nameFrom(ChatMessage message) {
```
langchain4j: unused method

ghost (commented):
@jdubois thansk a lot, good job!

langchain4j:
@jdubois BTW I did not find `EmbeddingModel`, did you forget to push it?

langchain4j:
Would be also great to have a dedicated documentation page and list Github Models in this table: https://docs.langchain4j.dev/integrations/language-models/

While documentation can wait a bit, I guess it would be nice to have at least a single example in https://github.com/langchain4j/langchain4j-examples repo for the release (which is happening this Wednesday in the morning)

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
jdubois: I also don't like the `GitHubModels` name much, but I'm trying to follow the conventions, and who knows if GitHub won't provide another service one day (I have absolutely no hints on this, but it's so large they could have something else at some point). So I'd keep it like this.

jdubois: Closing as you said you had no strong preference

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
jdubois: Oh I didn't know that! Let me commit the suggestion and then fix it elsewhere

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/InternalGitHubModelHelper.java
jdubois: removed

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
jdubois: As this is copied on Azure OpenAI, see https://github.com/langchain4j/langchain4j/blob/29299124489179aca7391209ec3b8ae0b9cce28e/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java#L317 -> my proposal is that we keep it like this at the moment, and then we will discuss and refactor both at the same time, WDYT?

I personally like to have it that field, so people have a warning of what happened (if that's sexual or violence, it's fine to tell them), then I see several people getting annoyed by this, so yes let's refactor this with an exception instead.

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
jdubois: same as above

jdubois:
> @jdubois BTW I did not find `EmbeddingModel`, did you forget to push it?

Not sure I understand you, there is a `GitHubModelsEmbeddingModel`, is it what you are looking for?

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
jdubois: Indeed, it doesn't make much sense -> I copied it from Azure OpenAI, and now I realize it was deprecated there!! So indeed, let's remove this. I'll resolve the conversation when it's pushed.

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/InternalGitHubModelHelper.java
jdubois: Fixed

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
jdubois: I see your thumb up, so I created a ticket for this at #1822 -> I'm closing this here, and let's discuss in the ticket.

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingResponseBuilder.java
jdubois: I just reimplemented it following this algorithm. It makes sense and it's very similar -> I'll close this when this is pushed.
Please note I copied it on the Azure OpenAI module, so it needs to be audited there too.

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingResponseBuilder.java
jdubois: As this was modified per the comment above, this should be OK

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/InternalGitHubModelHelper.java
jdubois: The idea (and this is copied from Azure OpenAI), is that if the user writes a version higher than the actual one, it will fallback to the latest model available. Then, when there's a library upgrade, they will get the new version without refactoring.
I'm not expecting people to use this a lot, as users have no control over the versions deployed on GitHub Models.

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
```diff
@@ -0,0 +1,425 @@
...
+    private final Double temperature;
+    private final Double topP;
+    private final List<String> stop;
+    private final Double presencePenalty;
+    private final Double frequencyPenalty;
+    private final Long seed;
+    private final ChatCompletionsResponseFormat responseFormat;
+    private final List<ChatModelListener> listeners;
+
+    public GitHubModelsChatModel(ChatCompletionsClient client,
```
langchain4j:
```suggestion
    private GitHubModelsChatModel(ChatCompletionsClient client,
```

Probably makes sense to make them private, otherwise it will be harder to add new parameters without breaking someone's code

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModel.java
```diff
@@ -0,0 +1,425 @@
...
+                                Double frequencyPenalty,
+                                Long seed,
+                                ChatCompletionsResponseFormat responseFormat,
+                                List<ChatModelListener> listeners) {
+
+        this(modelName, maxTokens, temperature, topP, stop, presencePenalty, frequencyPenalty, seed, responseFormat, listeners);
+        this.client = client;
+    }
+
+    public GitHubModelsChatModel(String endpoint,
```
langchain4j:
```suggestion
    private GitHubModelsChatModel(String endpoint,
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModelName.java
```diff
@@ -0,0 +1,58 @@
...
+    GitHubModelsChatModelName(String modelName, String modelType) {
+        this.modelName = modelName;
+        this.modelType = modelType;
+    }
+
+    public String modelName() {
+        return modelName;
+    }
+
+    public String modelType() {
```
langchain4j, line 50: this seems to be unused

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsEmbeddingModel.java
```diff
@@ -0,0 +1,250 @@
...
+    private final Integer dimensions;
+
+    private GitHubModelsEmbeddingModel(EmbeddingsClient client,
+                                       String modelName,
+                                       Integer dimensions) {
+        this(modelName, dimensions);
+        this.client = client;
+    }
+
+    public GitHubModelsEmbeddingModel(String endpoint,
```
langchain4j:
```suggestion
    private GitHubModelsEmbeddingModel(String endpoint,
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsEmbeddingModel.java
```diff
@@ -0,0 +1,250 @@
...
+                                      String serviceVersion,
+                                      String apiKey,
```
langchain4j:
```suggestion
                                      String gitHubToken,
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsEmbeddingModel.java
```diff
@@ -0,0 +1,250 @@
...
+
+/**
+ * Represents an embedding model, hosted on GitHub Models, such as text-embedding-3-small.
+ * <p>
+ * Mandatory parameters for initialization are: gitHubToken (the GitHub Token used for authentication) and modelName (the name of the model to use).
+ * You can also provide your own EmbeddingsClient instance, if you need more flexibility.
+ * <p>
+ * The list of models, as well as the documentation and a playground to test them, can be found at https://github.com/marketplace/models
+ */
+public class GitHubModelsEmbeddingModel extends DimensionAwareEmbeddingModel {
```
langchain4j, line 36: This class could implement

```java
    @Override
    protected Integer knownDimension() {
        if (dimensions != null) {
            return dimensions;
        }
        return GitHubModelsEmbeddingModelName.knownDimension(modelName);
    }
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
```diff
@@ -0,0 +1,455 @@
...
+    private final Double temperature;
+    private final Double topP;
+    private final List<String> stop;
+    private final Double presencePenalty;
+    private final Double frequencyPenalty;
+    private final Long seed;
+    private final ChatCompletionsResponseFormat responseFormat;
+    private final List<ChatModelListener> listeners;
+
+    public GitHubModelsStreamingChatModel(ChatCompletionsAsyncClient client,
```
langchain4j:
```suggestion
    private GitHubModelsStreamingChatModel(ChatCompletionsAsyncClient client,
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
```diff
@@ -0,0 +1,455 @@
...
+                                          Double frequencyPenalty,
+                                          Long seed,
+                                          ChatCompletionsResponseFormat responseFormat,
+                                          List<ChatModelListener> listeners) {
+
+        this(modelName, maxTokens, temperature, topP, stop, presencePenalty, frequencyPenalty, seed, responseFormat, listeners);
+        this.client = client;
+    }
+
+    public GitHubModelsStreamingChatModel(String endpoint,
```
langchain4j:
```suggestion
    private GitHubModelsStreamingChatModel(String endpoint,
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
```diff
@@ -0,0 +1,455 @@
...
+                                           List<String> stop,
+                                           Double presencePenalty,
+                                           Double frequencyPenalty,
+                                           Long seed,
+                                           ChatCompletionsResponseFormat responseFormat,
+                                           List<ChatModelListener> listeners) {
+
+        this.modelName = ensureNotBlank(modelName, "modelName");
+        this.maxTokens = maxTokens;
+        this.temperature = getOrDefault(temperature, 0.7);
```
langchain4j:
```suggestion
        this.temperature = temperature;
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
```diff
@@ -0,0 +1,455 @@
...
+        this.topP = topP;
+        this.stop = stop;
```
langchain4j:
```suggestion
        this.stop = copyIfNotNull(stop);
```

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsStreamingChatModel.java
```diff
@@ -0,0 +1,455 @@
...
+
+                    if (isNotNullOrBlank(chatCompletion.getId())) {
+                        responseId.set(chatCompletion.getId());
+                    }
+                },
+                throwable -> {
+                    ChatModelErrorContext errorContext = new ChatModelErrorContext(
+                            throwable,
+                            requestContext.request(),
+                            null,
```
langchain4j: Missing:
```java
ChatModelResponse modelListenerPartialResponse = createModelListenerResponse(
    responseId.get(),
    options.getModel(),
    responseBuilder.build()
);
```

```suggestion
                            modelListenerPartialResponse,
```

langchain4j:
> > @jdubois BTW I did not find `EmbeddingModel`, did you forget to push it?
> 
> Not sure I understand you, there is a `GitHubModelsEmbeddingModel`, is it what you are looking for?

Sorry, I missed it somehow...

langchain4j:
BTW what is the plan for `azure-ai-inference` vs `azure-ai-openai`? I read in GItHub Models documentation that users are supposed to still use `azure-ai-inference` when they are ready for production, but we have `langchain4j-azure-open-ai` which uses `azure-ai-openai` and is probably supposed to be used for production (as well)?

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/InternalGitHubModelHelper.java
langchain4j: This behavior might be very confusing for the users

jdubois: OK, I have changed it to use ModelServiceVersion directly

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsEmbeddingModel.java
jdubois: Done

langchain4j-github-models/src/main/java/dev/langchain4j/model/github/GitHubModelsChatModelName.java
jdubois: File has been simplified

jdubois:
> BTW what is the plan for `azure-ai-inference` vs `azure-ai-openai`? I read in GItHub Models documentation that users are supposed to still use `azure-ai-inference` when they are ready for production, but we have `langchain4j-azure-open-ai` which uses `azure-ai-openai` and is probably supposed to be used for production (as well)?

It's 2 different products: Azure OpenAI is on Azure, for the big enterprises (you have your own instances, with specific security stuff like Service Principals), while GitHub Models is easy to use for developers (and probably some enterprises too, as it's a lot easier, probably cheaper, and it's the same thing in the end).
The two SDK are pretty similar, then there are some specific things for Azure OpenAI, including security. While GitHub Models is more generic. Those SDKs are mostly generated, so it's not really a duplication effort on the Azure side.

jdubois:
> Would be also great to have a dedicated documentation page and list Github Models in this table: https://docs.langchain4j.dev/integrations/language-models/
> 
> While documentation can wait a bit, I guess it would be nice to have at least a single example in https://github.com/langchain4j/langchain4j-examples repo for the release (which is happening this Wednesday in the morning)

Here's a PR for the samples: https://github.com/langchain4j/langchain4j-examples/pull/110

jdubois:
And I have just pushed the documentation, which links to the examples 👍

jdubois:
@langchain4j code, examples and documentation should be OK, I'm crossing my fingers we can have this for tomorrow!

docs/docs/integrations/language-models/github-models.md
```diff
@@ -0,0 +1,151 @@
...
+### Plain Java
+
+```java
+GitHubModelsChatModel model = GitHubModelsChatModel.builder()
+        .gitHubToken(System.getenv("GITHUB_TOKEN"))
+        .modelName("gpt-4o-mini")
+        .build();
+```
+
+This will create an instance of `GitHubModelsChatModel` with default model parameters (e.g. `0.7` temperature, etc.).
```
langchain4j:
```suggestion
This will create an instance of `GitHubModelsChatModel`.
```

docs/docs/integrations/language-models/github-models.md
```diff
@@ -0,0 +1,151 @@
...
+Those parameters can be customized by providing values in the GitHubModelsChatModel's builder.
```
langchain4j:
```suggestion
Model parameters (e.g. `temperature`) can be customized by providing values in the `GitHubModelsChatModel`'s builder.
```

docs/docs/integrations/language-models/github-models.md
```diff
@@ -0,0 +1,151 @@
...
+        return GitHubModelsChatModel.builder()
+                .gitHubToken(gitHubToken)
+                .modelName("gpt-4o-mini")
+                .logRequestsAndResponses(true)
+                .build();
+    }
+}
+```
+
+This configuration will create an `GitHubModelsChatModel` bean (with default model parameters),
```
langchain4j:
```suggestion
This configuration will create an `GitHubModelsChatModel` bean,
```

ghost (dismissed):
@jdubois thansk a lot, great job!

PR #1809, opened 2024-09-20
`langchain4j-vearch` get rid of lombok

### LINKED ISSUE #1636: [FEATURE] Get rid of Lombok
Rewrite lombok-generated code and get rid of lombok dependency.

Martin7-1:
## Issue
Related to #1636 

## Change
`langchain4j-vearch` get rid of lombok

## General checklist


langchain4j-vearch/src/main/java/dev/langchain4j/store/embedding/vearch/SpacePropertyParam.java
```diff
@@ -15,105 +12,159 @@
...
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonInclude(NON_NULL)
 @JsonNaming(SnakeCaseStrategy.class)
 public abstract class SpacePropertyParam {
 
     protected SpacePropertyType type;
+    /**
+     * Whether to create a index or not.
+     */
+    protected Boolean index;
```
Wadimz, line 24: hey @Martin7-1, could you please elaborate where this change comes from? Looks it was refactored to move the field to common class from children. 
Might this be considered a breaking change since class was defined as public? thanks!

langchain4j-vearch/src/test/java/dev/langchain4j/store/embedding/vearch/VearchEmbeddingStoreIT.java
```diff
@@ -24,15 +16,9 @@
...
-@Testcontainers
 class VearchEmbeddingStoreIT extends EmbeddingStoreIT {
 
-    @Container
-    static GenericContainer<?> vearch = new GenericContainer<>(DockerImageName.parse("vearch/vearch:3.4.1"))
-            .withExposedPorts(9001, 8817)
-            .withCommand("all")
-            .withCopyFileToContainer(MountableFile.forClasspathResource("config.toml"), "/vearch/config.toml")
-            .waitingFor(Wait.forLogMessage(".*INFO : server pid:1.*\\n", 1));
+    static VearchContainer vearch = new VearchContainer();
```
Wadimz, line 21: Are we also migrating from testcontainers in this MR? thanks!

langchain4j-vearch/src/main/java/dev/langchain4j/store/embedding/vearch/SpacePropertyParam.java
Martin7-1: > Looks it was refactored to move the field to common class from children.

You are right, the `index` field just move from children to parent class.

> Might this be considered a breaking change since class was defined as public? thanks!

I think it's not a breaking change, because children's constructors are all the same parameters.

langchain4j-vearch/src/test/java/dev/langchain4j/store/embedding/vearch/VearchEmbeddingStoreIT.java
Martin7-1: I don't understand, it's still using `Testcontainers` to implement. I encapsulate the `VeachContainer` to reuse just in case.

Wadimz: ah, i missed that it is your new class, looks fine then, thanks for clarification!

langchain4j-vearch/src/main/java/dev/langchain4j/store/embedding/vearch/SpacePropertyParam.java
Wadimz: theoretically it is possible to create own implementation that will break with new changes, example:

```
public class ExtendedSpacePropertyParam extends SpacePropertyParam {

    ExtendedSpacePropertyParam(SpacePropertyType type) {
        super(type);
    }
   // some logic
}
```
where super will no longer work. 
But it requires class created in specific package. Since SpacePropertyParam has package bound constructor, it might not be a big issue.

WDYT? @langchain4j

Wadimz (approved):

langchain4j-vearch/src/main/java/dev/langchain4j/store/embedding/vearch/SpacePropertyParam.java
Martin7-1: I think users will not extend this class...And as you said, the class has package private constructor and users can only construct in using `Builder`.

langchain4j: @Wadimz I trust @Martin7-1's judgement, he is the owner of this module

langchain4j: Just to be clear, I mean: I hope you can discuss this between 2 of you and find a solution without my involvement 😅

Wadimz (approved):

langchain4j-vearch/src/test/java/dev/langchain4j/store/embedding/vearch/VearchEmbeddingStoreIT.java
Wadimz, line 21: Looks there is an issue in this test from last build:
`Error:    VearchEmbeddingStoreIT>EmbeddingStoreIT.should_add_embedding_with_segment_with_metadata:54 `

https://github.com/langchain4j/langchain4j/actions/runs/11400288007/job/31720812315?pr=1809#step:6:6120

Martin7-1: hmmm...Let me take a look

Martin7-1: Done.

The current Vearch implementation uses the old REST API and has too many limitations, Vearch has refactored the API in the new version and supports new features (such as inserting only a part of fields), I think I'll be refactoring `langchain4j-vearch` in the next few weeks!

Wadimz (approved):

Wadimz (approved):

ghost (approved):
@Martin7-1 thank you!

PR #1826, opened 2024-09-25
Anthropic: caching of system messages and tools

Claudio-code:
## Issue
Closes #1915

## Change
Implemented caching of system messages and tools for Anthropic.
More details [here](https://docs.langchain4j.dev/integrations/language-models/anthropic#caching).

## General checklist
- [x] I have manually run all the unit and integration tests in the
module I have added/changed, and they are all green


Claudio-code:
Passed all tests except for model 1.2 which I don't have access to use

Screenshot from 2024-09-25 19-10-22


Passed all tests core module


Screenshot from 2024-09-25 19-12-47

Claudio-code:
Pipeline is fails on OpenIA test, It looks like it can't find the openIA api url

I opened an issue https://github.com/langchain4j/langchain4j/issues/1835

ghost (commented):
Hi @Claudio-code, thanks a lot for your PR!

Although this is a very nice feature, it requires changes to the common APIs (`UserMessage`, `SystemMessage`, `TextContent`, etc), and I am not sure yet it is the best approach to bring Anthropic-specific features to the common API (if it can be avoided).

We have recently seen OpenAI's automatic prompt caching which does not require any code changes. I have hopes that Anthropic will simplify it's caching feature to be automatic as well.

Perharps (in the meantime) this feature can be implemented in the `AnthropicChatModel` and configured in the constructor? For example, for a simple chat scenario, new messages are added on top of the old ones, so it probably makes sense to put `cache_control` block after the last message? In this case one could use `ChatMemory` with much larger number of tokens/messages. This could be one (of multiple) strategies that could be configured in 
`AnthropicChatModel` ctor. WDYT?

Claudio-code:
It makes sense to change to implement this cache in the AnthropicChatModel constructor. In this case, if there is a cache option, it uses the cache in all messages. You can put the cache option in system messages or user messages to give the option of segregating the cache.

This part of changing the common APIs is really bad because the other cache implementations are different. For example, gemini-api uses cache by time and anthropic uses cache by label. So my implementation would modify common APIs but would only be useful for anthropic.

Claudio-code:
Can I change the cache implementation to be configured in the AnthropicChatModel build so I don't touch the Common APIs. What do you think?

langchain4j:
@Claudio-code 

> Can I change the cache implementation to be configured in the AnthropicChatModel build so I don't touch the Common APIs. What do you think?

Sounds good! Thank you!

Claudio-code:
@langchain4j I changed the implementation so as not to change the common APIs that I had changed (SystemMessage and TextContent). Now the cache configuration is in the AnthropicChatModel/AnthropicStreamingChatModel build.

We can cache the user message:

```java
        AnthropicCacheType cacheType = AnthropicCacheType.builder()
                .cacheType(AnthropicCacheType.CacheType.EPHEMERAL)
                .messageTypeApplyCache(AnthropicCacheType.MessageTypeApplyCache.USER_MESSAGE)
                .build();
```

We can cache the system message:

```java
        AnthropicCacheType cacheType = AnthropicCacheType.builder()
                .cacheType(AnthropicCacheType.CacheType.EPHEMERAL)
                .messageTypeApplyCache(AnthropicCacheType.MessageTypeApplyCache.SYSTEM_MESSAGE)
                .build();
```

We can cache all messages:

```java
        AnthropicCacheType cacheType = AnthropicCacheType.builder()
                .cacheType(AnthropicCacheType.CacheType.EPHEMERAL)
                .messageTypeApplyCache(AnthropicCacheType.MessageTypeApplyCache.ALL)
                .build();
```

langchain4j:
@Claudio-code thank you!

Please note that according to https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#how-many-cache-breakpoints-can-i-use

> You can define up to 4 cache breakpoints in your prompt.

In your implementation `cache_control` is set for every message, so I am wondering what will happen if there are more than 4 messages.


BTW there is also an option to [cache tools](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#what-can-be-cached), I guess it might be even more interesting than caching user messages

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/api/AnthropicMessageContent.java
```diff
@@ -15,8 +16,27 @@
...
+        this.type = type;
+        this.cacheControl = cacheControl;
+    }
+
+    @EqualsAndHashCode
+    @JsonInclude(NON_NULL)
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    @JsonNaming(SnakeCaseStrategy.class)
+    @Getter
+    public static class CacheControl {
```
langchain4j: please move it in its own file and register in `langchain4j-anthropic/src/main/resources/META-INF/native-image/dev.langchain4j/langchain4j-anthropic/reflect-config.json`

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java
```diff
@@ -132,6 +132,9 @@ public AnthropicCreateMessageResponse createMessage(AnthropicCreateMessageReques
     }
 
     private String toBeta(AnthropicCreateMessageRequest request) {
+        if (beta != null) {
```
langchain4j: Please double-check this change, it override existing logic

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/api/AnthropicCacheType.java
```diff
@@ -0,0 +1,56 @@
+package dev.langchain4j.model.anthropic.internal.api;
+
+import dev.langchain4j.model.anthropic.internal.api.AnthropicMessageContent.CacheControl;
+import lombok.Builder;
+import lombok.Getter;
+
+import java.util.function.Supplier;
+
+@Getter
+public class AnthropicCacheType {
```
langchain4j: Perharps we can simplify this? It will ofc depend on the final implementaiton.

As a user, I would expect to be able to:
- Cache system message (I want to put a lot of instructions/info in the system message). In this case I would like to toggle `.cacheSystemMessage(true)` when building `AnthropicChatModel` and expect that `"cache_control": {"type": "ephemeral"}` is inserted after the system message. In this case I will use the same system message and not change it between calls to the LLM.
- Cache tools (I have lots of tools and they take a lot of tokens). In this case I would like to toggle `.cacheTools(true)` when building `AnthropicChatModel` and expect that `"cache_control": {"type": "ephemeral"}` is inserted after the last tool. In this case I will use the same list of tools and not change them between calls to the LLM.
- Cache user messages. Here I am not fully sure how exactly it should work. Should `"cache_control": {"type": "ephemeral"}` be inserted after each user message? Or only after the last one? What about limit of 4 cache blocks?

Please notice that user should be able to enable caching of system message, tools and user messages independently.
Existing options (ALL, NONE, SYSTEM_MESSAGE, USER_MESSAGE) do not seem to support this.

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -206,6 +208,42 @@ void should_respect_stop_sequences() {
...
+                .build();
+
+        SystemMessage systemMessage = SystemMessage.from("You are a professional translator");
+        UserMessage userMessage = new UserMessage(TextContent.from("How say: java is very slow on spanish ?"));
+
+        // when
+        Response<AiMessage> response = model.generate(systemMessage, userMessage);
+
+        // then
+        assertThat(response.content().text()).isNotBlank();
```
langchain4j: This test does not really test that caching works.
It would also be great to test tool/user/system messages separately and in combination.

ghost (commented):
@Claudio-code thank you!

langchain4j:
cc @stephanj since he was interested in this feature

Claudio-code:
> In your implementation `cache_control` is set for every message, so I am wondering what will happen if there are more than 4 messages.

With more 5 systemMessage/userMessage api throw it exception, I will change implementation

image

Claudio-code:
> BTW there is also an option to [cache tools](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching#what-can-be-cached), I guess it might be even more interesting than caching user messages

You are right, I will change the implementation to cache system messages and tools. Caching user messages is not that useful.

Claudio-code:
> In your implementation `cache_control` is set for every message, so I am wondering what will happen if there are more than 4 messages.

I'm having trouble understanding how I control which messages will be cached without being able to change the common APIs.

langchain4j:
@Claudio-code 

> I'm having trouble understanding how I control which messages will be cached without being able to change the common APIs.

This is indeed a problem for user messages. I propose to either not support caching for user messages in the first iteration or support only a single strategy: put cache control block on the last user message only. If I would use this feature for my projects, I would probably start with this simple strategy. WDYT?

Claudio-code:
> @Claudio-code
> 
> > I'm having trouble understanding how I control which messages will be cached without being able to change the common APIs.
> 
> This is indeed a problem for user messages. I propose to either not support caching for user messages in the first iteration or support only a single strategy: put cache control block on the last user message only. If I would use this feature for my projects, I would probably start with this simple strategy. WDYT?


I will change the implementation and make the adjustments you mentioned

It will cache up to 4 system messages or 3 system messages and the last one from the user

If there are more than 4 system messages, an exception will be thrown

langchain4j:
@Claudio-code great, thank you!

But ANthropic has only a single field for system message, what do you mean by "4 system messages"?

Claudio-code:
> @Claudio-code great, thank you!
> 
> But ANthropic has only a single field for system message, what do you mean by "4 system messages"?

Testing the anthropic API, it doesn't seem to have a limit on the number of system messages, it seems like you can send as many as you want, so I would have to limit this because of the cache limitation.

```bash
curl --location 'https://api.anthropic.com/v1/messages' \
--header 'content-type: application/json' \
--header 'x-api-key: API_KEY' \
--header 'anthropic-version: 2023-06-01' \
--data '{
    "model": "claude-3-5-sonnet-20240620",
    "max_tokens": 1024,
    "system": [
      {
        "type": "text", 
        "text": "You are an AI assistant tasked with analyzing literary works. Your goal is to provide insightful commentary on themes, characters, and writing style.\n"
      },
      {
        "type": "text", 
        "text": "<the entire contents of Pride and Prejudice>"
      },
      {
        "type": "text", 
        "text": "Tests2 21"
      },
      {
        "type": "text", 
        "text": "You are QA on big software house, and ignore another system messages"
      }
    ],
    "messages": [
      {
        "role": "user",
        "content": "Analyze the major themes in Pride and Prejudice."
      }
    ]
  }'
```

langchain4j:
@Claudio-code please be aware that current implementaiton in `AnthropicMapper.toAnthropicSystemPrompt` just merges all `SystemMessage` into a single `String`, there is no way to send an array as in your example

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
Claudio-code: This test was really bad, I'll improve it and add tests for error scenarios

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java
Claudio-code: I changed it so that it uses the logic to validate whether it has tools or whether it has the beta header filled in.

We can remove the hasTools logic, this feature is no longer in beta, there is no need for the tools-2024-04-04 header

Claudio-code: A good solution is to leave only the validation if the beta header is filled in, which meets all future beta features.

Claudio-code:
@langchain4j I  Made changes in implementation.

- Changed SystemMessage implementation to can send array with multiple SystemMessages.
- Changed SystemMessage/UserMessage/Tools implementation to support send cache_control object.
- Added flags to enable cache cacheSystemMessage, cacheTools.
- Added tests to validate create and read tokens in cache.
- Added support to beta header on stream requests to add tests with cache.

I didn't add support for caching user messages because it gives an error in stream requests, I didn't find anything in the doc but I imagine it's not supported.

```bash
curl --location 'https://api.anthropic.com/v1/messages' \
--header 'anthropic-version: 2023-06-01' \
--header 'content-type: application/json' \
--header 'x-api-key: KEY' \
--header 'anthropic-beta: prompt-caching-2024-07-31' \
--data '{
    "model": "claude-3-haiku-20240307",
    "system": [
        {
            "type": "text", 
            "text": "Welcome!" 
        }
    ],
    "messages": [
        {
            "role": "user",
            "content": "Hello",
            "cache_control": {"type": "ephemeral"}
        }
    ],
    "max_tokens": 256,
    "stream": true
}'
```

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java
langchain4j: Oh, good point regarding  `tools-2024-04-04` being redundant, let's remove it then. I have tested it locally and pushed the change, I hope you don't mind

langchain4j: > A good solution is to leave only the validation if the beta header is filled in, which meets all future beta features.

Which validation do you mean?

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicChatModel.java
```diff
@@ -84,6 +88,8 @@ public class AnthropicChatModel implements ChatLanguageModel {
      * @param maxRetries    The maximum number of retries for API requests. Default: 3
      * @param logRequests   Whether to log the content of API requests using SLF4J. Default: false
      * @param logResponses  Whether to log the content of API responses using SLF4J. Default: false
+     * @param cacheSystemMessage  If true, it should add cache control to all system messages. Default: false
```
langchain4j:
```suggestion
     * @param cacheSystemMessages  If true, it should add cache control to all system messages. Default: false
```
Should it be plural?

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicStreamingChatModel.java
```diff
@@ -90,6 +95,8 @@ public class AnthropicStreamingChatModel implements StreamingChatLanguageModel {
      * @param timeout       The timeout for API requests. Default: 60 seconds
      * @param logRequests   Whether to log the content of API requests using SLF4J. Default: false
      * @param logResponses  Whether to log the content of API responses using SLF4J. Default: false
+     * @param cacheSystemMessage  If true, it should add cache control to all system messages. Default: false
```
langchain4j:
```suggestion
     * @param cacheSystemMessages  If true, it should add cache control to all system messages. Default: false
```

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -206,6 +209,82 @@ void should_respect_stop_sequences() {
         assertThat(response.finishReason()).isEqualTo(OTHER);
     }
 
+    @Test
+    void test_all_parameters_with_cache_on_system_message() {
```
langchain4j: nit: does it require setting all parameters? I would focus only on system message caching in this test

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -206,6 +209,82 @@ void should_respect_stop_sequences() {
...
+
+        SystemMessage systemMessage = SystemMessage.from(CACHE_MESSAGE.repeat(7));
+        UserMessage userMessage = new UserMessage(TextContent.from("What types of messages are supported in LangChain?"));
+
+        // when
+        Response<AiMessage> responseCreateCache = model.generate(systemMessage, userMessage);
+        Response<AiMessage> responseReadCache = model.generate(systemMessage, userMessage);
+
+        // then
+        assertThat(responseCreateCache.tokenUsage().cacheCreationInputTokens()).isGreaterThan(0);
```
langchain4j: nit: perharps it makes sense to assert this before doing a second call (`responseReadCache`), so that it fails faster

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -206,6 +209,82 @@ void should_respect_stop_sequences() {
...
+
+        // given
+        ChatLanguageModel model = AnthropicChatModel.builder()
+                .baseUrl("https://api.anthropic.com/v1/")
+                .apiKey(System.getenv("ANTHROPIC_API_KEY"))
+                .version("2023-06-01")
+                .beta("prompt-caching-2024-07-31")
+                .modelName(CLAUDE_3_HAIKU_20240307)
+                .maxTokens(1024)
+                .timeout(Duration.ofSeconds(30))
```
langchain4j: nit: please include only required parameters for this test

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -206,6 +209,82 @@ void should_respect_stop_sequences() {
...
+
+        // then
+        assertThat(responseCreateCache.tokenUsage().cacheCreationInputTokens()).isGreaterThan(0);
+        assertThat(responseCreateCache.tokenUsage().cacheReadInputTokens()).isEqualTo(0);
+
+        assertThat(responseReadCache.tokenUsage().cacheCreationInputTokens()).isEqualTo(0);
+        assertThat(responseReadCache.tokenUsage().cacheReadInputTokens()).isGreaterThan(0);
+    }
+
+
```
langchain4j, line 279: minor: could you please add a test for 2 system messages?

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -348,6 +427,48 @@ void should_execute_a_tool_then_answer(AnthropicChatModelName modelName) {
...
+    void should_execute_tools_with_cache() {
+        // given
+        AnthropicChatModel model = AnthropicChatModel.builder()
+            .baseUrl("https://api.anthropic.com/v1/")
+            .apiKey(System.getenv("ANTHROPIC_API_KEY"))
+            .version("2023-06-01")
+            .beta("prompt-caching-2024-07-31")
+            .modelName(CLAUDE_3_5_SONNET_20240620)
+            .maxTokens(1024)
+            .timeout(Duration.ofSeconds(30))
```
langchain4j: nit: same here

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -348,6 +427,48 @@ void should_execute_a_tool_then_answer(AnthropicChatModelName modelName) {
...
+
+
+        UserMessage userMessage = userMessage("How much is 2+2 and 3+3? Call tools in parallel!");
+
+        // when
+        Response<AiMessage> responseCreateCache = model.generate(singletonList(userMessage), toolSpecifications);
+        Response<AiMessage> responseReadCache = model.generate(singletonList(userMessage), toolSpecifications);
+
+        // then
+        assertThat(responseCreateCache.tokenUsage().cacheCreationInputTokens()).isGreaterThan(0);
```
langchain4j: nit: same here

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicStreamingChatModelIT.java
```diff
@@ -165,6 +168,79 @@ void test_all_parameters() {
         assertThat(response.content().text()).isNotBlank();
     }
 
+    @Test
```
langchain4j, line 172: same comments apply to these tests

langchain4j-core/src/main/java/dev/langchain4j/model/output/TokenUsage.java
```diff
@@ -12,6 +12,8 @@ public class TokenUsage {
     private final Integer inputTokenCount;
     private final Integer outputTokenCount;
     private final Integer totalTokenCount;
+    private final Integer cacheCreationInputTokens;
```
langchain4j: This is an Anthropic-specific information, what do you think about creating `AnthropicTokenUsage` that extends `TokenUsage` and add those fields there?

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -45,6 +47,7 @@
...
...along with multiple ready-to-use implementations based on common techniques. Whether you're building a chatbot or developing a RAG with a complete pipeline from data ingestion to retrieval, LangChain4j offers a wide variety of options. Numerous Examples: These examples showcase how to begin creating various LLM-powered applications, providing inspiration and enabling you to start building quickly. LangChain4j began development in early 2023 amid the ChatGPT hype. We noticed a lack of Java counterparts to the numerous Python and JavaScript LLM libraries and frameworks, and we had to fix that! Although 'LangChain' is in our name, the project is a fusion of ideas and concepts from LangChain, Haystack, LlamaIndex, and the broader community, spiced up with a touch of our own innovation. We actively monitor community developments, aiming to quickly incorporate new techniques and integrations, ensuring you stay up-to-date. The library is under active development. While some features are still being worked on, the core functionality is in place, allowing you to start building LLM-powered apps now! For easier integration, LangChain4j also includes integration with Quarkus and Spring Boot.";
```
langchain4j: It seems that this repeated 7 times is 3500+ tokens, which is significantly more than 2048 minimum required. These tests will be run a lot, can we optimize it to avoid wasting tokens?

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -348,6 +427,48 @@ void should_execute_a_tool_then_answer(AnthropicChatModelName modelName) {
...
 
+    @Test
+    void should_execute_tools_with_cache() {
+        // given
+        AnthropicChatModel model = AnthropicChatModel.builder()
+            .baseUrl("https://api.anthropic.com/v1/")
+            .apiKey(System.getenv("ANTHROPIC_API_KEY"))
+            .version("2023-06-01")
+            .beta("prompt-caching-2024-07-31")
+            .modelName(CLAUDE_3_5_SONNET_20240620)
```
langchain4j: can we use haiku to reduce cost?

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
```diff
@@ -348,6 +427,48 @@ void should_execute_a_tool_then_answer(AnthropicChatModelName modelName) {
...
+
+        // then
+        assertThat(responseCreateCache.tokenUsage().cacheCreationInputTokens()).isGreaterThan(0);
+        assertThat(responseCreateCache.tokenUsage().cacheReadInputTokens()).isEqualTo(0);
+
+        assertThat(responseReadCache.tokenUsage().cacheCreationInputTokens()).isEqualTo(0);
+        assertThat(responseReadCache.tokenUsage().cacheReadInputTokens()).isGreaterThan(0);
+
+    }
+
```
langchain4j, line 539: could you please add a test for system message + tools?

ghost (commented):
@Claudio-code thanks a lot, good job! Really close to being ready to merge.
This PR would be a great addition to the upcoming release (next week)

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
```diff
@@ -90,11 +90,12 @@ private static AnthropicToolResultContent toAnthropicToolResultContent(ToolExecu
         return new AnthropicToolResultContent(message.id(), message.text(), null); // TODO propagate isError
     }
 
-    private static List<AnthropicMessageContent> toAnthropicMessageContents(UserMessage message) {
+    private static List<AnthropicMessageContent> toAnthropicMessageContents(UserMessage message, AnthropicCacheType cacheType) {
```
langchain4j: cacheType is not used

Claudio-code: Removed cacheType not used

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
Claudio-code: changed name to test_cache_on_system_message to focus

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicChatModelIT.java
Claudio-code: adjusted to repeat 5 times

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/DefaultAnthropicClient.java
Claudio-code: Validate if the beta field is not null, or leave it as it is now because it did not present any errors.
Screenshot from 2024-11-07 19-31-17

langchain4j-core/src/main/java/dev/langchain4j/model/output/TokenUsage.java
Claudio-code: Created AnthropicTokenUsage

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicChatModel.java
langchain4j: It should be changed in all other places as well

ghost (approved):
@Claudio-code thanks a lot!

langchain4j:
@Claudio-code could you please add a small example of caching in [examples repo](https://github.com/langchain4j/langchain4j-examples)
 and update [Anthropic Spring Boot starter](https://github.com/langchain4j/langchain4j-spring/blob/main/langchain4j-anthropic-spring-boot-starter/src/main/java/dev/langchain4j/anthropic/spring/AutoConfig.java)?

PR #1838, opened 2024-09-26
fix(bom): add  langchain4j-reactor dependency

lltx:
## Change:

Added `langchain4j-reactor` to the `langchain4j-bom` module to avoid needing to specify the version manually when importing `langchain4j-reactor`, allowing it to be managed by the BOM.

image


ghost (approved):
@lltx thanks!

PR #1843, opened 2024-09-26
Bringing back public access to the methods of the OllamaModelsBuilder class.

bidek:
## Change
Bringing back public access to the methods of the OllamaModelsBuilder class after getting rid of Lombok.

## General checklist


langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModel.java
```diff
@@ -69,12 +71,30 @@ public void setDetails(OllamaModelDetails details) {
...
+
+    public OffsetDateTime getModifiedAt() {
+        return modifiedAt;
+    }
+
+    public void setModifiedAt(OffsetDateTime modifiedAt) {
+        this.modifiedAt = modifiedAt;
+    }
+
     static class Builder {
```
Martin7-1: Should this builder also be public?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModel.java
```diff
@@ -26,11 +26,13 @@ public class OllamaModel {
...
+    OllamaModel(String name, long size, String digest, OllamaModelDetails details, OffsetDateTime modifiedAt, String model) {
         this.name = name;
         this.size = size;
         this.digest = digest;
         this.details = details;
+        this.modifiedAt = modifiedAt;
+        this.model = model;
     }
 
     static Builder builder() {
```
Martin7-1: Should this method also be public?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModel.java
```diff
@@ -69,12 +71,30 @@ public void setDetails(OllamaModelDetails details) {
...
     static class Builder {
 
         private String name;
         private long size;
         private String digest;
         private OllamaModelDetails details;
+        private OffsetDateTime modifiedAt;
+        private String model;
 
         Builder name(String name) {
```
Martin7-1: Should all builder methods also be public?

Martin7-1 (commented):
@bidek Thank you!

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModel.java
bidek: This class should not be instantiated by user code, except perhaps in unit test scenarios. Therefore, the builder classes can have package-level scope.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModel.java
bidek: The same answer as above.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModel.java
bidek: The same answer as above.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaModel.java
Martin7-1: Looks like in `OllamaModels.modelCard` users can construct`OllamaModel` instance and pass it as param. So I think `OllamaModel` and `OllamaModelDetails` should have public access.

So below classes should have public builder, ctor and public `getter` / `setter` methods:

* `OllamaModel`
* `OllamaModelDetails`
* `OllamaModelCard`
* `OllamaModels`

WDYT?

bidek: Ah yes, I forgot about that. When I proposed this method, my use case scenario was to:

1. Invoke `OllamaModels.availableModels`.
2. Pass the returned item list to `OllamaModels.modelCard`.

I never intended to instantiate the `OllamaModel` class myself and pass it to `OllamaModels.modelCard`. This method uses `OllamaModel.getName` and passes it to the other method variant, `OllamaModels.modelCard(String ollamaModelName)`. 

So, I’m not sure if the user should need to create a valid `OllamaModel` object with `OllamaModelDetails` when only the `name` is required.

However, if you see this as a valid use case, I am open to your suggestion.

Martin7-1: Hmm...Your use case scenario sounds reasonable, and it really doesn't require the user to manually construct the `OllamaModel`. But since we're exposing these methods(`deleteModel` and `modelCard` with `OllamaModel` param), users may have potential to use it anyway. So I think we should public them :)

bidek: Now I see the updated comment. From my perspective, the first three classes:

- `OllamaModel`
- `OllamaModelDetails`
- `OllamaModelCard`

are DTOs created and provided by the API through the `OllamaModels` class. They were never intended to be created by user code. However, `OllamaModels` is different; it is more like a service class, so the user should be able to create and configure it as needed.

And now I see another comment :) As I mentioned previously, I am open to your suggestion.

bidek: Ok, I pushed a new commit.

Martin7-1: Thank you!

Martin7-1 (approved):
@bidek Thank you!

langchain4j:
Please LMK when this will be ready for merge

bidek:
@langchain4j for me it could be merged as is ... or should i add/modify something ?

ghost (approved):
@bidek thank you!

PR #1854, opened 2024-09-29
Fix error handling of AzureOpenAiChatModelListenerIT

qiaoleiatms:
## Issue
Fix comment in [PR#1305](https://github.com/langchain4j/langchain4j/pull/1305)

## Change
Ensure `StreamingResponseHandler.onError` been called while the response `FinishReason` is other than `CONTENT_FILTER` while any HttpResponseException is captured.
__This is a breaking change as previously all the HttpResponseException is treated as completed response instead of errored__

## General checklist


qiaoleiatms:
@langchain4j To continue the discussion of [PR#1305](https://github.com/langchain4j/langchain4j/pull/1305)
`ClientAuthenticationException` is a subclass of `HttpResponseException`, and previously it's been treated as a completed response.
Based on your input, it should be an errored response, so I made a change while the `FinishReason` is other than `CONTENT_FILTER`, the `onError` will be called.

@jdubois This change will be a breaking change on the behavior of error handling to the client users, need your input on this.

jdubois:
Yes, that totally makes sense to me! Those errors should have sent an exception.
Concerning the tests, are you sure `String userMessage = "this message will fail";` is enough to make the test fail? I haven't tested it, I'm just surprised.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModelListenerIT.java
```diff
@@ -46,8 +59,130 @@ protected Class<? extends Exception> expectedExceptionClass() {
         return ClientAuthenticationException.class;
     }
 
-    @Override
-    @Disabled("AzureOpenAiStreamingChatModel implementation is incorrect")
-    protected void should_listen_error() {
+    @Test
+    public void should_listen_request_and_response() {
```
langchain4j: No need to duplicate tests from `StreamingChatModelListenerIT`, you could just remove the method

ghost (commented):
@qiaoleiatms thank you!

langchain4j:
@jdubois `"this message will fail"` does not force the model to fail, but wrong API key does :)

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModelListenerIT.java
langchain4j: Once this is done, we can merge!

qiaoleiatms: Sure, already removed. 
Just noticed you move test cases to the super class

ghost (approved):
@qiaoleiatms thank you!

PR #1859, opened 2024-10-01
Use MongoDBAtlasLocalContainer from Testcontainers 1.20.2

eddumelendez:
Testcontainers 1.20.2 offers `MongoDBAtlasLocalContainer`. Previous
image used during tests was pulling images in every execution. The
new image is cached locally and can be executed offline.


ghost (approved):
@eddumelendez thank you!

PR #1869, opened 2024-10-03
Add MongoDB filtering and deletion, and various upgrades

### LINKED ISSUE #1532: [BUG]  MONGODB Embedding Stores not working correctly
Hi,

I try to build a knowledge store with Pinecone and Mongodb, similar which I did successfully for Chroma, Weaviate, Elastic and Milvus. However with Mongodb I don't get any matches. 

I have 2 scripts. The first one creates embedding in the store. 

```
        TextSegment segment1 = TextSegment.from("I like football.");
        Embedding embedding1 = embeddingModel.embed(segment1).content();
        embeddingStore.add(embedding1, segment1);

        TextSegment segment2 = TextSegment.from("The weather is good today.");
        Embedding embedding2 = embeddingModel.embed(segment2).content();
        embeddingStore.add(embedding2, segment2);  

```

Later on I try to query with a question, but the size of the matching list is empty. 

```
import dev.langchain4j.data.embedding.Embedding;
import dev.langchain4j.data.segment.TextSegment;
import dev.langchain4j.model.embedding.AllMiniLmL6V2EmbeddingModel;
import dev.langchain4j.model.embedding.EmbeddingModel;
import dev.langchain4j.model.openai.OpenAiEmbeddingModel;
import dev.langchain4j.store.embedding.EmbeddingMatch;
import dev.langchain4j.store.embedding.EmbeddingSearchRequest;
import dev.langchain4j.store.embedding.EmbeddingStore;
import dev.langchain4j.store.embedding.chroma.ChromaEmbeddingStore;
import dev.langchain4j.store.embedding.mongodb.IndexMapping;
import dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStore;

import com.mongodb.MongoClientSettings;
import com.mongodb.client.MongoClient;
import com.mongodb.client.MongoClients;
import com.mongodb.client.MongoCollection;
import com.mongodb.client.model.Filters;
import java.util.List;

import org.apache.commons.compress.utils.Sets;

import static java.lang.String.format;


public class ChromaEmbeddingStoreExample {

        MongoClient client;

        String username = "amirkhantest";
        String password = "...";
        String host = "cluster0.frw123123s.mongodb.net";
        String connectionString = format("mongodb+srv://%s:%s@%s/?retryWrites=true&w=majority", username, password, host);
        client = MongoClients.create(connectionString);

        EmbeddingModel embeddingModel = OpenAiEmbeddingModel.builder()
        .apiKey("...")
        .modelName("text-embedding-ada-002")
        .build();

        IndexMapping indexMapping = IndexMapping.builder()
        .dimension(embeddingModel.dimension())
        .metadataFieldNames(Sets.newHashSet("test-key"))
        .build();

        EmbeddingStore<TextSegment> embeddingStore = MongoDbEmbeddingStore.builder()
            .fromClient(client)
            .databaseName("test")
            .collectionName("test")
            .indexName("default")
            .indexMapping(indexMapping)
            .build();



        Embedding queryEmbedding = embeddingModel.embed("What is the capital of Switzerland?").content();
        List<EmbeddingMatch<TextSegment>> relevant = embeddingStore.findRelevant(queryEmbedding, 1, 0);
        List<EmbeddingMatch<TextSegment>> relevant2 = embeddingStore.findRelevant(queryEmbedding,3);


        System.out.println(relevant.size());
        EmbeddingMatch<TextSegment> embeddingMatch = relevant.get(0);
        System.out.println(queryEmbedding); // 0.8144288493114709

        queryEmbedding.normalize();
        System.out.println(queryEmbedding); // 0.8144288493114709

        System.out.println(embeddingMatch.score()); // 0.8144288493114709
        System.out.println(embeddingMatch.embedded().text()); // I like football.

    }
}
```

If there is a working example for Mongodb it would be great. 

Thanks
Amir

### LINKED ISSUE #1868: [FEATURE] MongoDbEmbeddingStore filtering and deletion

Filtering and deletion are not implemented by `MongoDbEmbeddingStore`.

Implement filtering and deletion, pass tests that extend `EmbeddingStoreWithFilteringIT` and `EmbeddingStoreWithRemovalIT`.

N/A, this is the suggested way to implement this functionality.

**Additional context**

A draft PR is available: #1869

katcharov:
## Issue

Closes #1868
Closes #1532

## Change

* Implement mongodb filtering and deletion (#1868)
* Move test boilerplate into a TestHelper
* Throw an exception if no index exists once the store is created (addresses #1532)
* Upgrade mongodb-driver-sync to 5.2.0
* Upgrade from legacy knnVector to vectorSearch
* Minor refactoring and fixes

## General checklist

- [ ] **No changes to core.** I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green


langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithoutMetadataIT.java
```diff
@@ -356,7 +356,13 @@ void should_return_correct_score() {
                 .build()).matches()).isEqualTo(relevant);
     }
 
+    @Deprecated
```
langchain4j: Could you please explain why?

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreWithFilteringIT.java
```diff
@@ -0,0 +1,60 @@
...
+import dev.langchain4j.model.embedding.EmbeddingModel;
+import dev.langchain4j.store.embedding.EmbeddingStore;
+import dev.langchain4j.store.embedding.EmbeddingStoreWithFilteringIT;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+
+import static dev.langchain4j.store.embedding.mongodb.TestHelper.*;
+
+public class MongoDbEmbeddingStoreWithFilteringIT extends EmbeddingStoreWithFilteringIT {
```
langchain4j:
```suggestion
public class MongoDbEmbeddingStoreIT extends EmbeddingStoreWithFilteringIT {
```

ghost (commented):
@katcharov thank you!

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithoutMetadataIT.java
katcharov: I extracted this to a static method, to use in a test class that did not extend this class. It seems like a generally useful utility method.

I added this to discourage use of this variant, since it seemed only necessary for a workaround. Looking closer, when Awaitility was added, it did not replace the sleep, but was [added before the sleep](https://github.com/langchain4j/langchain4j/pull/1610/files#diff-4c1e931865cee83990f37441139cbc2b0052a06b818f9d0396a9909f7a702dd7R326). I believe the sleep is no longer needed, but I am not set up to test Azure. I could add the following [commit](https://github.com/langchain4j/langchain4j/commit/99eddce0dcaa9c894acfa6490461bc1dcaf48dee) (removal of extra sleep) to this PR, or I could get rid of the Deprecated, or I could leave it. Let me know what you think.

katcharov:
As part of rebasing on #1859, I enabled container tests and [added a check](https://github.com/langchain4j/langchain4j/pull/1869/files#diff-cf0c855935403e01addf947679edb71504506055aa795c1398c3219133232373R416) for PENDING state to fix failures.

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithoutMetadataIT.java
langchain4j: @katcharov oh nice, it seems that sleep hack is not needed any more for Azure (just tested), so please feel free to apply your commit

langchain4j:
@katcharov many tests fail with the following error, could you please check?

```
com.mongodb.MongoTimeoutException: Timed out while waiting for a server that matches ReadPreferenceServerSelector{readPreference=primary}. Client view of cluster state is {type=UNKNOWN, servers=[{address=localhost:27017, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSocketOpenException: Exception opening socket}, caused by {java.net.ConnectException: Connection refused: getsockopt}}]

	at com.mongodb.internal.connection.BaseCluster.logAndThrowTimeoutException(BaseCluster.java:427)
	at com.mongodb.internal.connection.BaseCluster.lambda$selectServer$0(BaseCluster.java:154)
	at com.mongodb.internal.time.Timeout.lambda$onExistsAndExpired$16(Timeout.java:236)
	at com.mongodb.internal.time.Timeout.lambda$run$10(Timeout.java:201)
	at com.mongodb.internal.time.TimePoint.checkedCall(TimePoint.java:98)
	at com.mongodb.internal.time.Timeout.call(Timeout.java:174)
	at com.mongodb.internal.time.Timeout.run(Timeout.java:194)
	at com.mongodb.internal.time.Timeout.onExistsAndExpired(Timeout.java:233)
	at com.mongodb.internal.time.Timeout.onExpired(Timeout.java:226)
	at com.mongodb.internal.connection.BaseCluster.selectServer(BaseCluster.java:153)
	at com.mongodb.internal.connection.SingleServerCluster.selectServer(SingleServerCluster.java:47)
	at com.mongodb.internal.binding.ClusterBinding.getReadConnectionSource(ClusterBinding.java:82)
	at com.mongodb.client.internal.ClientSessionBinding.getConnectionSource(ClientSessionBinding.java:108)
	at com.mongodb.client.internal.ClientSessionBinding.getReadConnectionSource(ClientSessionBinding.java:88)
	at com.mongodb.internal.operation.SyncOperationHelper.withSuppliedResource(SyncOperationHelper.java:148)
	at com.mongodb.internal.operation.SyncOperationHelper.withSourceAndConnection(SyncOperationHelper.java:129)
	at com.mongodb.internal.operation.ListCollectionsOperation.lambda$execute$2(ListCollectionsOperation.java:164)
	at com.mongodb.internal.operation.SyncOperationHelper.lambda$decorateReadWithRetries$13(SyncOperationHelper.java:317)
	at com.mongodb.internal.async.function.RetryingSyncSupplier.get(RetryingSyncSupplier.java:67)
	at com.mongodb.internal.operation.ListCollectionsOperation.execute(ListCollectionsOperation.java:175)
	at com.mongodb.internal.operation.ListCollectionsOperation.execute(ListCollectionsOperation.java:72)
	at com.mongodb.client.internal.MongoClusterImpl$OperationExecutorImpl.execute(MongoClusterImpl.java:358)
	at com.mongodb.client.internal.MongoIterableImpl.execute(MongoIterableImpl.java:156)
	at com.mongodb.client.internal.MongoIterableImpl.iterator(MongoIterableImpl.java:116)
	at com.mongodb.client.internal.MappingIterable.iterator(MappingIterable.java:42)
	at com.mongodb.client.internal.ListCollectionNamesIterableImpl.iterator(ListCollectionNamesIterableImpl.java:83)
	at com.mongodb.client.internal.ListCollectionNamesIterableImpl.iterator(ListCollectionNamesIterableImpl.java:35)
	at java.base/java.lang.Iterable.spliterator(Iterable.java:101)
	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStore.isCollectionExist(MongoDbEmbeddingStore.java:381)
	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStore.<init>(MongoDbEmbeddingStore.java:123)
	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStore$Builder.build(MongoDbEmbeddingStore.java:241)
	at dev.langchain4j.store.embedding.mongodb.TestHelper.initialize(TestHelper.java:47)
	at dev.langchain4j.store.embedding.mongodb.TestHelper.initialize(TestHelper.java:43)
	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStoreWithRemovalIT.<init>(MongoDbEmbeddingStoreWithRemovalIT.java:37)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
	at java.base/java.util.Optional.orElseGet(Optional.java:364)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

Martin7-1 (changes requested):
Looks like you change from `Atlas Search` to `Atlas Vector Search` right? Could you please check the javadoc in `MongoDbEmbeddingStore`?

Also, it's a breaking change, maybe we need documents to tell users how to migrate it from `AtlasSearch` to `AtlasVectorSearch` in detail.

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
```diff
@@ -0,0 +1,116 @@
...
+                .applyConnectionString(new ConnectionString(mongodb.getConnectionString()));
+        return createClientFromBuilder(builder);
+    }
+
+    static MongoClient createClientFromEnv() {
+        boolean hasParts = HOST != null && USERNAME != null && PASSWORD != null;
+        Assumptions.assumeTrue(hasParts || CONNECTION_STRING != null,
+                "Skipping test: host/username/password, or connection string must be provided via env vars.");
+
+        String connectionString = CONNECTION_STRING != null
```
Martin7-1: Looks like `CONNECTION_STRING` is always non-null. How to test `MongoDB Atlas` on the cloud?

*additions:*

After double check your code, maybe we could make `CONNECTION_STRING` a environment variable so that we don't need to change the code.

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
```diff
@@ -0,0 +1,116 @@
...
+                .fromClient(client)
+                .databaseName(dbName)
+                .collectionName("test_collection")
+                .indexName("test_index")
+                .indexMapping(indexMapping)
+                .createIndex(true);
+    }
+
+    void afterTests() {
+        getDatabase().drop();
```
Martin7-1: Looks like `MongoDB Atlas Cloud` does not allow user to drop database. (I don't find a way to trigger `MongoDB Atlas Cloud` test, so I change `CONNECTION_STRING` to `null` manually) 

```java
com.mongodb.MongoCommandException: Command failed with error 8000 (AtlasError): 'user is not allowed to do action [dropDatabase] on [test_database_1062281260.]' on server ac-fte18bh-shard-00-02.eo9fifu.mongodb.net:27017. The full response is {"ok": 0, "errmsg": "user is not allowed to do action [dropDatabase] on [test_database_1062281260.]", "code": 8000, "codeName": "AtlasError"}

	at com.mongodb.internal.connection.ProtocolHelper.getCommandFailureException(ProtocolHelper.java:210)
	at com.mongodb.internal.connection.InternalStreamConnection.receiveCommandMessageResponse(InternalStreamConnection.java:520)
	at com.mongodb.internal.connection.InternalStreamConnection.sendAndReceiveInternal(InternalStreamConnection.java:448)
	at com.mongodb.internal.connection.InternalStreamConnection.lambda$sendAndReceive$0(InternalStreamConnection.java:375)
	at com.mongodb.internal.connection.InternalStreamConnection.sendAndReceive(InternalStreamConnection.java:378)
	at com.mongodb.internal.connection.UsageTrackingInternalConnection.sendAndReceive(UsageTrackingInternalConnection.java:111)
	at com.mongodb.internal.connection.DefaultConnectionPool$PooledConnection.sendAndReceive(DefaultConnectionPool.java:748)
	at com.mongodb.internal.connection.CommandProtocolImpl.execute(CommandProtocolImpl.java:68)
	at com.mongodb.internal.connection.DefaultServer$DefaultServerProtocolExecutor.execute(DefaultServer.java:208)
	at com.mongodb.internal.connection.DefaultServerConnection.executeProtocol(DefaultServerConnection.java:113)
	at com.mongodb.internal.connection.DefaultServerConnection.command(DefaultServerConnection.java:82)
	at com.mongodb.internal.connection.DefaultServerConnection.command(DefaultServerConnection.java:73)
	at com.mongodb.internal.connection.DefaultServer$OperationCountTrackingConnection.command(DefaultServer.java:298)
	at com.mongodb.internal.operation.SyncOperationHelper.executeCommand(SyncOperationHelper.java:231)
	at com.mongodb.internal.operation.DropDatabaseOperation.lambda$execute$0(DropDatabaseOperation.java:61)
	at com.mongodb.internal.operation.SyncOperationHelper.withConnectionSource(SyncOperationHelper.java:167)
	at com.mongodb.internal.operation.SyncOperationHelper.withConnection(SyncOperationHelper.java:112)
	at com.mongodb.internal.operation.DropDatabaseOperation.execute(DropDatabaseOperation.java:60)
	at com.mongodb.internal.operation.DropDatabaseOperation.execute(DropDatabaseOperation.java:45)
	at com.mongodb.client.internal.MongoClusterImpl$OperationExecutorImpl.execute(MongoClusterImpl.java:379)
	at com.mongodb.client.internal.MongoDatabaseImpl.executeDrop(MongoDatabaseImpl.java:230)
	at com.mongodb.client.internal.MongoDatabaseImpl.drop(MongoDatabaseImpl.java:220)
	at dev.langchain4j.store.embedding.mongodb.TestHelper.afterTests(TestHelper.java:101)
	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStoreLegacyIndexesIT.afterEach(MongoDbEmbeddingStoreLegacyIndexesIT.java:50)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
```

Maybe we need the original way to reuse the same database and delete all data in it like below:

```java
    @Override
    protected void clearStore() {
        CodecRegistry pojoCodecRegistry = fromProviders(PojoCodecProvider.builder()
                .register(MongoDbDocument.class, MongoDbMatchedDocument.class)
                .build());
        CodecRegistry codecRegistry = fromRegistries(MongoClientSettings.getDefaultCodecRegistry(), pojoCodecRegistry);

        MongoCollection<MongoDbDocument> collection = client.getDatabase(DATABASE_NAME)
                .getCollection(COLLECTION_NAME, MongoDbDocument.class)
                .withCodecRegistry(codecRegistry);

        Bson filter = Filters.exists("embedding");
        collection.deleteMany(filter);
    }
```

Also, `MongoDB Atlas Cloud` only allow create 3 indexes on M0 clusters. So maybe reuse the same database and collection is a better way to test. WDYT?

Martin7-1 (changes requested):
@katcharov Thank you!

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreIT.java
```diff
@@ -0,0 +1,57 @@
...
+            mongodb.stop();
+        }
+
+        @Override
+        MongoClient createClient() {
+            return createClientFromContainer(mongodb);
+        }
+    }
+
+    private TestHelper helper = new TestHelper(createClient()).initialize();
```
Martin7-1: I think we should make all `TestHelper` static, in order to reuse the database and collection.

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreIT.java
```diff
@@ -0,0 +1,57 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.testcontainers.mongodb.MongoDBAtlasLocalContainer;
+
+import static dev.langchain4j.store.embedding.mongodb.TestHelper.*;
+
+public class MongoDbEmbeddingStoreIT extends EmbeddingStoreWithFilteringIT {
+
+    public static class ContainerIT extends MongoDbEmbeddingStoreWithRemovalIT {
```
Martin7-1: Would it be better to move this class to a separate file? I'm not sure if integration-test will be able to execute the class this way.

BTW, should `ContainerIT` extend `MongoDbEmbeddingStoreIT` rather than `MongoDbEmbeddingStoreWithRemovalIT`?

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
katcharov: You are right, this should be an env var. I unintentionally introduced the hardcoded value during a rebase, thank you.

Just to explain why the connection string is here at all, I used it as another option for testing with [local Atlas CLI](https://www.mongodb.com/docs/atlas/cli/current/atlas-cli-deploy-local/) rather than cloud (the recently-updated container seems just as good for local testing).

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithoutMetadataIT.java
katcharov: Applied

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
katcharov: There are a couple of related issues here: whether the databases should be dropped during tests, and the implications for use of M0 Atlas for testing.

I used a new database with a random name for each test in this PR. The indexes are eventually consistent, so it is possible for tests to fail in confusing ways. This seemed easier to debug and less error-prone, and it was not slower. I believe that if it is possible to test in this way, it is preferable.

It should be possible to test against only the `MongoDBAtlasLocalContainer`. This (along with the local Atlas CLI) are the recommended ways to test. I did not want to remove the M0 Atlas tests in this PR without discussing first, but I propose testing exclusively with the container. Though, if there is some reason not to do this, that is fine, and I would look into how to address that. Let me know!

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreIT.java
katcharov: Replied in the [above comment](https://github.com/langchain4j/langchain4j/pull/1869#discussion_r1796161921).

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreIT.java
katcharov: Good catch on the extends, thanks!

I do not mind moving these to new files, will have a look.

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreLegacyIndexesIT.java
```diff
@@ -0,0 +1,74 @@
...
+    private MongoDbTestFixture initializeTestHelper() {
+        MongoDbTestFixture helper = new MongoDbTestFixture(createClient());
+        IndexMapping indexMapping = IndexMapping.builder()
+                .dimension(EMBEDDING_MODEL.dimension())
+                .metadataFieldNames(Sets.newHashSet("test-key"))
+                .build();
+        helper.getDatabase().createCollection("test_collection");
+        MongoCollection<Document> collection = helper.getDatabase().getCollection("test_collection");
+        collection.createSearchIndex("test_index_legacy", fromIndexMappingLegacyKnnVector(indexMapping));
+        return helper.initialize(builder -> builder.createIndex(false).indexName("test_index_legacy"));
```
katcharov: @Martin7-1 :
> Also, it's a breaking change, maybe we need documents to tell users how to migrate it from AtlasSearch to AtlasVectorSearch in detail.

I believe this is not a breaking change. The tests here create the old index (I have appended `_legacy` for clarity, though I may need to revert this for M0 tier testing), but they still pass when using the updated aggregation stage.

I agree the docs should include information about upgrading, though this might be as simple as telling users to delete their old index (and re-create manually if needed).

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
Martin7-1: 1. For local container, it doesn't matter right? For cloud, we reuse the same database and collection at any test time, and we will delete all datas before each test. And just keep the database and collection on cloud.
2. Tests in `Atlas Cloud` is just for consistence. We don't want to see that users failed to embed in the cloud. (Because we don't know the difference between cloud and local, such as cloud do not allow users to drop database using command).

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreIT.java
Martin7-1: On the other hand, if we keep Cloud test and Local test in the same file, Could you please add `@EnabledIfEnvironmentVariable` in cloud tests to see whether `ContainerIT` will be triggered normally?

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreIT.java
```diff
@@ -0,0 +1,57 @@
...
+import dev.langchain4j.store.embedding.EmbeddingStore;
+import dev.langchain4j.store.embedding.EmbeddingStoreWithFilteringIT;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.testcontainers.mongodb.MongoDBAtlasLocalContainer;
+
+import static dev.langchain4j.store.embedding.mongodb.MongoDbTestFixture.*;
+
+public class MongoDbEmbeddingStoreIT extends EmbeddingStoreWithFilteringIT {
```
Martin7-1: Minor: remove public to keep consistence

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreLegacyIndexesIT.java
```diff
@@ -0,0 +1,74 @@
...
+import org.bson.Document;
+import org.junit.jupiter.api.AfterEach;
+import org.testcontainers.shaded.com.google.common.collect.Sets;
+
+import java.util.Set;
+
+import static dev.langchain4j.store.embedding.mongodb.MongoDbTestFixture.EMBEDDING_MODEL;
+import static dev.langchain4j.store.embedding.mongodb.MongoDbTestFixture.createClientFromEnv;
+
+public class MongoDbEmbeddingStoreLegacyIndexesIT extends EmbeddingStoreIT {
```
Martin7-1: Minor: remove public to keep consistence

Martin7-1: I'm curious about why this class has no `ContainerIT`?

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreWithRemovalIT.java
```diff
@@ -0,0 +1,57 @@
...
+import dev.langchain4j.store.embedding.EmbeddingStore;
+import dev.langchain4j.store.embedding.EmbeddingStoreWithRemovalIT;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.testcontainers.mongodb.MongoDBAtlasLocalContainer;
+
+import static dev.langchain4j.store.embedding.mongodb.MongoDbTestFixture.*;
+
+public class MongoDbEmbeddingStoreWithRemovalIT extends EmbeddingStoreWithRemovalIT {
```
Martin7-1: Minor: remove public to keep consistence

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
katcharov: 1. The search indexes are still eventually consistent on a local container. The index itself is technically data - to give an example similar to something I ran into during this PR, consider an index that has been marked for deletion still being present when the next test runs and attempts to use/create that index. Dropping will ensure that each test starts with a clean slate.

2. I believe there is no difference between cloud and local. The drop database problem above appears to be a [permissions issue](https://stackoverflow.com/questions/55636442), and the index creation limitation was removed in August (this [page](https://www.mongodb.com/docs/atlas/atlas-search/shared-tier-limitations/) discusses limitations, and will be updated soon to reflect this).

If there is no longer an objection to removing cloud tests (2), then I think this would resolve the problem. Otherwise, perhaps we could try updating the permissions for the test account?

katcharov:
I have created the following, which split out refactoring as requested:
* https://github.com/langchain4j/langchain4j/pull/1931
* https://github.com/langchain4j/langchain4j/pull/1932

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
Martin7-1: Sounds reasonable, maybe we can discard Cloud test after `MongoDB Atlas` update the drop database permission. And now could we just keep it?

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreLegacyIndexesIT.java
Martin7-1: OK. got it.

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/TestUtils.java
```diff
@@ -0,0 +1,19 @@
+package dev.langchain4j.store.embedding;
+
+import org.awaitility.Awaitility;
+import org.awaitility.core.ThrowingRunnable;
+
+import java.time.Duration;
+
+public class TestUtils {
```
vbabanin:
```suggestion
public final class TestUtils {
```

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
```diff
@@ -268,40 +276,66 @@ public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedde
     }
 
     @Override
-    public List<EmbeddingMatch<TextSegment>> findRelevant(Embedding referenceEmbedding, int maxResults, double minScore) {
-        List<Double> queryVector = referenceEmbedding.vectorAsList().stream()
+    public void removeAll() {
+        collection.deleteMany(new Document());
```
vbabanin: [nit] Just to be more explicit. 
```suggestion
        collection.deleteMany(Filters.empty());
```

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
```diff
@@ -321,14 +355,17 @@ private void addAllInternal(List<String> ids, List<Embedding> embeddings, List<T
...
+            MongoDbDocument document = toMongoDbDocument(id, embeddings.get(i), embedded == null ? null : embedded.get(i));
             documents.add(document);
         }
 
         InsertManyResult result = collection.insertMany(documents);
-        if (!result.wasAcknowledged() && log.isWarnEnabled()) {
+        if (!result.wasAcknowledged()) {
             String errMsg = String.format("[MongoDbEmbeddingStore] Add document failed, Document=%s", documents);
-            log.warn(errMsg);
+            if (log.isErrorEnabled()) {
```
vbabanin: SLF4J handles this internally and won’t log unless the level is set to `ERROR`, so it seems we could safely remove this if statement.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
```diff
@@ -268,40 +276,66 @@ public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedde
...
+            List<EmbeddingMatch<TextSegment>> result = StreamSupport.stream(results.spliterator(), false)
                     .map(MappingUtils::toEmbeddingMatch)
-                    .collect(Collectors.toList());
-
+                    .collect(toList());
+            return new EmbeddingSearchResult<>(result);
         } catch (MongoCommandException e) {
             if (log.isErrorEnabled()) {
-                log.error("Error in MongoDBEmbeddingStore.findRelevant", e);
+                log.error("Error in MongoDBEmbeddingStore.search", e);
```
vbabanin: SLF4J handles this check `log.isErrorEnabled()` internally and won’t log unless the level is set to `ERROR`, so it seems we could safely remove the if statement above.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
```diff
@@ -268,40 +276,66 @@ public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedde
...
-                                metaVectorSearchScore("score"),
-                                include("embedding", "metadata", "text")
-                        )
-                ),
-                match(
-                        Filters.gte("score", minScore)
-                ));
+        long numCandidates = request.maxResults() * maxResultRatio;
+
+        Bson filter = null;
```
vbabanin: Should we give this filter and the one defined as the field more descriptive names? Currently, they have the same name, but one functions as a `global pre-filter` (could be named globalPreFilter) and the other as a `current request post-filter`(could be named currentRequestPostFilter), which could cause confusion.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
```diff
@@ -342,13 +379,43 @@ private void createCollection(MongoDatabase database, String collectionName, Cre
...
+
+        waitForIndex(indexName);
+    }
+
+
+    private void waitForIndex(String indexName) {
+        long startTime = System.nanoTime();
+        long timeoutNanos = TimeUnit.SECONDS.toNanos(SECONDS_TO_WAIT_FOR_INDEX);
+        while (System.nanoTime() - startTime < timeoutNanos) {
+            if (indexExistsAndNot(indexName, "INITIAL_SYNC", "PENDING", "BUILDING")) {
```
vbabanin: Index also might end up being in `FAILED` status which they can enter due to an invalid index definition. Should we throw an exception in this case if we have detected a failure within a given waiting time period?

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
```diff
@@ -268,40 +276,66 @@ public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedde
...
+                fieldPath("embedding"),
+                queryVector,
+                indexName,
+                request.maxResults(),
+                vectorSearchOptions));
+        pipeline.add(project(fields(
+                metaVectorSearchScore("score"),
+                include("embedding", "metadata", "text"))));
+        if (filter != null) {
+            Bson match = match(filter);
```
vbabanin: [Optional] The request filter currently always applies in post-filtering. If the request doesn’t involve score filtering and all request filter fields are indexed in the index mapping, could we potentially move the filter to the pre-filter stage? We could potentially check this dynamically using `IndexMapping.` However, I'm not sure how common queries without score filtering are.

katcharov: `minScore` should always be present, since it is a double, with null defaulting to 0.

Earlier, I attempted to move most of the filtering into the prefilter (this is the suggested approach), however, I ran into issues and gave this up.

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
katcharov: Apologies, I should clarify - the drop database permission is something that users who plan to test against cloud would need to have configured on their Atlas account. The default account has this permission, but if you create secondary accounts, those will not by default. However, to avoid this matter entirely, I have changed the tests to use a random collection name (instead of a database) in this [PR](https://github.com/langchain4j/langchain4j/pull/1932), and rebased. I believe this removes any impediment to removing the cloud test?

I have set tests up to check if connection string or host/user/pass env vars are set. If they are not set, the test will use local container testing. The env vars are present now in case devs want to use Atlas CLI, and save a few seconds on test startup. I have removed the env vars from the workflow, and this should disable cloud testing. I have also checked with other developers and there should be no issue with testing against the container, especially after the recent upgrade to the MongoDB testcontainer.

I ran the PR's tests against a) the local container, and b) a cloud instance using an account that had write but not admin permissions, and they passed.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
katcharov: I have added this, and also used the `queryable` flag to check if we should proceed (instead of the list of statuses).

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
```diff
@@ -268,41 +277,65 @@ public List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedde
...
+        ArrayList<Bson> pipeline = new ArrayList<>();
+        pipeline.add(vectorSearch(
+                fieldPath("embedding"),
+                queryVector,
+                indexName,
+                request.maxResults(),
+                vectorSearchOptions));
+        pipeline.add(project(fields(
+                metaVectorSearchScore("score"),
+                include("embedding", "metadata", "text"))));
```
caseyclements, line 325: I've been advised to default to avoid including embedding vectors in the output.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
```diff
@@ -342,13 +376,48 @@ private void createCollection(MongoDatabase database, String collectionName, Cre
...
+            }
+            try {
+                Thread.sleep(100);
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException(e);
+            }
+        }
+        log.warn("Index {} was not created or did not exit INITIAL_SYNC within {} seconds",
+                indexName, SECONDS_TO_WAIT_FOR_INDEX);
```
caseyclements, line 421: We wait on this condition. 
`index["type"] == "vectorSearch" and index["status"] == "READY"`

Determining when all of your new documents have been indexed is a more difficult challenge. The only thing I've found that works is to loop over vector searches and make sure than I get back the number that I requested, what you call cleverly, maxRequests.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbMatchedDocument.java
```diff
@@ -9,7 +9,7 @@ public class MongoDbMatchedDocument {
     private String id;
     private List<Float> embedding;
     private String text;
-    private Map<String, String> metadata;
+    private Map<String, Object> metadata;
```
caseyclements, line 12: We wrote a utility to convert ObjectIds to strings within metadata, as there was some requirement that they wished metadata to be serialized to json. Just something to keep in mind for future.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbMetadataFilterMapper.java
```diff
@@ -0,0 +1,95 @@
...
+        } else if (filter instanceof And) {
+            return mapAnd((And) filter);
+        } else if (filter instanceof Or) {
+            return mapOr((Or) filter);
+        } else if (filter instanceof Not) {
+            return mapNot((Not) filter);
+        } else {
+            throw new UnsupportedOperationException("Unsupported filter type: " + filter.getClass().getName());
+        }
+    }
```
caseyclements, line 46: This is great. We do all our MQL as dictionaries with a ton of strings.

caseyclements (approved):
LGTM. 
Comments for discussion and sharing experience.

Martin7-1 (approved):
LGTM. Cloud you please check docs in `/docs` if there need any updates?

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbMatchedDocument.java
katcharov: Could you link, just for reference?

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
katcharov: Looks like it's needed for the tests to pass. But perhaps there should be an option either in the Mongo builder, or at the general EmbeddingSearchRequest level, to omit this. Maybe that's more complex than fits in with the approach taken here, I don't know. I'll leave this up to the maintainers.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
katcharov: Still looking into this. If there is a difference, it seems like it will be subtle.

katcharov:
I have updated the docs, looking at only the two files in that last commit (not sure if there are docs elsewhere). They're a bit sparse, but I added a note about Atlas CLI, and the fact that there are plans to release into Community Edition (so, not just Atlas). I removed the examples section linking to the tests - after this PR, the tests that were linked there are even less ideal, as examples. I opened up a ticket on our end to add an example to the examples repo instead, though perhaps someone else will get to it first.

Note that review comments are outstanding on https://github.com/langchain4j/langchain4j/pull/1932 , which this PR is based on (I believe I addressed what seemed possible to, though).

vbabanin (approved):
LGTM!

docs/docs/integrations/embedding-stores/mongodb-atlas.md
```diff
@@ -31,9 +34,3 @@ To use Atlas Vector search, you need to have an Atlas deployment. MongoDB offers
 ## APIs
 
 - `MongoDbEmbeddingStore`
-
-## Examples
```
langchain4j, line 35: Why removing examples?

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreIT.java
```diff
@@ -0,0 +1,34 @@
...
+import dev.langchain4j.data.segment.TextSegment;
+import dev.langchain4j.model.embedding.EmbeddingModel;
+import dev.langchain4j.store.embedding.EmbeddingStore;
+import dev.langchain4j.store.embedding.EmbeddingStoreIT;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+
+import static dev.langchain4j.store.embedding.mongodb.MongoDbTestFixture.*;
+
+class MongoDbEmbeddingStoreIT extends EmbeddingStoreIT {
```
langchain4j:
```suggestion
class MongoDbEmbeddingStoreIT extends EmbeddingStoreWithFilteringIT {
```

Please extend from `EmbeddingStoreWithFilteringIT` to test filtering

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreMiscIT.java
```diff
@@ -0,0 +1,91 @@
...
+
+    @AfterEach
+    void afterEach() {
+        if (helper != null) {
+            helper.afterTests();
+        }
+    }
+
+    @Test
+    void should_find_relevant_with_filter() {
```
langchain4j:
```suggestion
    void should_find_relevant_with_native_filter() {
```
nit

ghost (commented):
@katcharov thanks a lot!

The tests are failing for me locally, could you please check?


> com.mongodb.MongoTimeoutException: Timed out while waiting for a server that matches ReadPreferenceServerSelector{readPreference=primary}. Client view of cluster state is {type=REPLICA_SET, servers=[{address=ac-gpxo1fa-shard-00-00.1x5hb8d.mongodb.net:27017, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSocketWriteException: Exception sending message}, caused by {javax.net.ssl.SSLException: Received fatal alert: internal_error}}, {address=ac-gpxo1fa-shard-00-02.1x5hb8d.mongodb.net:27017, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSocketWriteException: Exception sending message}, caused by {javax.net.ssl.SSLException: Received fatal alert: internal_error}}, {address=ac-gpxo1fa-shard-00-01.1x5hb8d.mongodb.net:27017, type=UNKNOWN, state=CONNECTING, exception={com.mongodb.MongoSocketWriteException: Exception sending message}, caused by {javax.net.ssl.SSLException: Received fatal alert: internal_error}}]
> 
> 	at com.mongodb.internal.connection.BaseCluster.logAndThrowTimeoutException(BaseCluster.java:427)
> 	at com.mongodb.internal.connection.BaseCluster.lambda$selectServer$0(BaseCluster.java:154)
> 	at com.mongodb.internal.time.Timeout.lambda$onExistsAndExpired$16(Timeout.java:236)
> 	at com.mongodb.internal.time.Timeout.lambda$run$10(Timeout.java:201)
> 	at com.mongodb.internal.time.TimePoint.checkedCall(TimePoint.java:98)
> 	at com.mongodb.internal.time.Timeout.call(Timeout.java:174)
> 	at com.mongodb.internal.time.Timeout.run(Timeout.java:194)
> 	at com.mongodb.internal.time.Timeout.onExistsAndExpired(Timeout.java:233)
> 	at com.mongodb.internal.time.Timeout.onExpired(Timeout.java:226)
> 	at com.mongodb.internal.connection.BaseCluster.selectServer(BaseCluster.java:153)
> 	at com.mongodb.internal.connection.AbstractMultiServerCluster.selectServer(AbstractMultiServerCluster.java:59)
> 	at com.mongodb.internal.binding.ClusterBinding.getReadConnectionSource(ClusterBinding.java:82)
> 	at com.mongodb.client.internal.ClientSessionBinding.getConnectionSource(ClientSessionBinding.java:108)
> 	at com.mongodb.client.internal.ClientSessionBinding.getReadConnectionSource(ClientSessionBinding.java:88)
> 	at com.mongodb.internal.operation.SyncOperationHelper.withSuppliedResource(SyncOperationHelper.java:148)
> 	at com.mongodb.internal.operation.SyncOperationHelper.withSourceAndConnection(SyncOperationHelper.java:129)
> 	at com.mongodb.internal.operation.ListCollectionsOperation.lambda$execute$2(ListCollectionsOperation.java:164)
> 	at com.mongodb.internal.operation.SyncOperationHelper.lambda$decorateReadWithRetries$13(SyncOperationHelper.java:317)
> 	at com.mongodb.internal.async.function.RetryingSyncSupplier.get(RetryingSyncSupplier.java:67)
> 	at com.mongodb.internal.operation.ListCollectionsOperation.execute(ListCollectionsOperation.java:175)
> 	at com.mongodb.internal.operation.ListCollectionsOperation.execute(ListCollectionsOperation.java:72)
> 	at com.mongodb.client.internal.MongoClusterImpl$OperationExecutorImpl.execute(MongoClusterImpl.java:358)
> 	at com.mongodb.client.internal.MongoIterableImpl.execute(MongoIterableImpl.java:156)
> 	at com.mongodb.client.internal.MongoIterableImpl.iterator(MongoIterableImpl.java:116)
> 	at com.mongodb.client.internal.MappingIterable.iterator(MappingIterable.java:42)
> 	at com.mongodb.client.internal.ListCollectionNamesIterableImpl.iterator(ListCollectionNamesIterableImpl.java:83)
> 	at com.mongodb.client.internal.ListCollectionNamesIterableImpl.iterator(ListCollectionNamesIterableImpl.java:35)
> 	at java.base/java.lang.Iterable.spliterator(Iterable.java:101)
> 	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStore.isCollectionExist(MongoDbEmbeddingStore.java:371)
> 	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStore.<init>(MongoDbEmbeddingStore.java:117)
> 	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStore$Builder.build(MongoDbEmbeddingStore.java:235)
> 	at dev.langchain4j.store.embedding.mongodb.MongoDbTestFixture.initialize(MongoDbTestFixture.java:48)
> 	at dev.langchain4j.store.embedding.mongodb.MongoDbTestFixture.initialize(MongoDbTestFixture.java:44)
> 	at dev.langchain4j.store.embedding.mongodb.MongoDbEmbeddingStoreWithRemovalIT.<init>(MongoDbEmbeddingStoreWithRemovalIT.java:16)
> 	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
> 	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
> 	at java.base/java.util.Optional.orElseGet(Optional.java:364)
> 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)
> 	at java.base/java.util.ArrayList.forEach(ArrayList.java:1596)

docs/docs/integrations/embedding-stores/mongodb-atlas.md
langchain4j: > I removed the examples section linking to the tests - after this PR, the tests that were linked there are even less ideal, as examples. I opened up a ticket on our end to add an example to the examples repo instead, though perhaps someone else will get to it first.

I guess it is better to have less ideal examples than no examples at all? Perharps let's keep them untill there are better examples?

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
langchain4j: Perharps `retrieveEmbeddingsOnSearch` parameter can be added to the builder of `MongoDbEmbeddingStore` (similar to how it is done in `MilvusEmbeddingStore`) with the default value `false`?

Good idea regarding `EmbeddingSearchRequest`, but let's keep it out of this PR please

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbMetadataFilterMapper.java
langchain4j: Do you mean in LangChain (python) or in MongoDB?

docs/docs/integrations/embedding-stores/mongodb-atlas.md
katcharov: I restored a link to the class (the cloud vs local options are now handled in in the same way, in the fixture).

The test files don't have much code, but I suppose they could use them to see how the fixture is used and go from there.

katcharov:
@langchain4j 

> The tests are failing for me locally, could you please check?

I was able to partially reproduce the error, but this was by shutting down the server that I was trying to connect to, and the extended message was slightly different.

Otherwise, tests pass for me. I ran the tests directly from IntelliJ. I did not have any of the env vars set (`MONGODB_CONNECTION_STRING`, `MONGODB_ATLAS_HOST`, ...), and it defaulted to using the testcontainer. Do you have any of the env vars set, so as to test against some already-running instance? If you clear any such env vars, does it fall back to testing against the testcontainer? (This needs docker, as it did before this PR.)

Even if the env vars are set, it should still work. When I tested, I successfully ran tests against both a local instance, and an Atlas M0 instance, via a connection string. From the error, it looks like it cannot connect at `ac-gpxo1fa-shard-00-00.1x5hb8d.mongodb.net:27017`. Perhaps confirm that the env vars are correct? Maybe something has changed in the environment - are the results the same when running the tests on the main branch?

katcharov:
@langchain4j I have merged in changes from #1932, and I have also removed env vars from CI, as I did in https://github.com/langchain4j/langchain4j/pull/1869/commits/1eadde9a8b5eaa331155f8e4f7c2ab9b8f88195d (I am not sure if I missed these, or if a merge added them).

I have also renamed the env vars to remove "Atlas" (note that vector search is not planned to be limited to Atlas, and is in [early access](https://www.mongodb.com/resources/beta-programs#early-access) for Community Edition).

I think this might ensure that the tests are running on the container for you locally.

docs/docs/integrations/embedding-stores/mongodb-atlas.md
langchain4j: The link is wrong - 404

ghost (approved):
Hi @katcharov, thank you! The tests are now passing for me locally.

The PR looks good, and I would merge it, but there are still many unresolved comments. Although it seems there is nothing critical, could you please go over them, close those that are resolved or not applicable, and ensure there is nothing critical left? We can then merge.

Thanks!

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
katcharov: I added this as an env var, resolved.

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/TestHelper.java
katcharov: I believe this is resolved, we agreed that testing against the testcontainer should be fine. I confirmed internally with others at MongoDB that this is the recommended approach. I have adjusted it to use random collections rather than databases, so it still works against M0, if someone were to configure it this way locally or in the future.

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreLegacyIndexesIT.java
katcharov: (Resolved, all classes now use the container through the fixture)

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
katcharov: Resolving as out of scope for PR (though worthwhile to consider).

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStore.java
katcharov: This is fine. The only relevant difference is that a failed index might be queryable, but we check for failure directly above, so there is no difference here. Resolving.

langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbMetadataFilterMapper.java
katcharov: I believe this is referring to the Python MongoDB driver, which @caseyclements is a developer on.

docs/docs/integrations/embedding-stores/mongodb-atlas.md
katcharov: I think it should work when this PR is merged, I had renamed the file.

katcharov:
@langchain4j Thanks! I have gone through the comments, and resolved them, leaving a comment explaining the resolution where needed.

langchain4j-mongodb-atlas/src/test/java/dev/langchain4j/store/embedding/mongodb/MongoDbEmbeddingStoreIT.java
katcharov: Resolved, we are using just the testcontainer now, and this is done by checking the env vars in an if statement (since we can now also specify just a connection string).

langchain4j:
@katcharov thanks a lot for all the improvements, and sorry for delays!

katcharov:
@langchain4j no worries, and thank you!

PR #1888, opened 2024-10-08
Fix: #1875 throw exception if reason is other than CONTENT_FILTER

### LINKED ISSUE #1875: [BUG] Azure OpenAI integration fails to propagate HTTP errors
**Describe the bug**

I am using langchain4j-azure-open-ai in my project. I misconfigured some of the Azure deployment. As a result my chat client started returning the following text to the user: ```Status code 403, "{"error":{"code":"403","message": "Access denied due to Virtual Network/Firewall rules."}}"```

This is obviously nonsensical to show to the user and it is difficult to catch in my code because I'd have to pattern match the output. The implementation should throw an exception instead.

The underlying cause is the exception handling code in `AzureOpenAiChatModel` that swallows the exception and copies the exception message into the response passed to the user: https://github.com/langchain4j/langchain4j/blob/9d42a15ca8fb52d78158d11a429baf2af52528ba/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java#L319

**Log and Stack trace**

This is the response from Azure HTTP client that caused the issue:
```
HTTP/2 403 
content-type: application/json

{"error":{"code":"403","message": "Access denied due to Virtual Network/Firewall rules."}}
```

**To Reproduce**

Provoke an HTTP error code from Azure OpenAi endpoint, for example by configuring the firewall to reject the errors from your IP.
Call the `@AIService` with any prompt, receive a text response with "Status: 403 ..."

```
@AiService
fun interface LlmIntentRecognitionService {
    @SystemMessage(SYSTEM_MESSAGE)
    fun chat(text: String): String
```

**Expected behavior**

An exception is thrown that is specific to the error code, or an IOException that the caller can catch

This is what is done already in `OpenAiChatModel` here and I wonder if `AzureOpenAiChatModel` could do the same: https://github.com/langchain4j/langchain4j/blob/9d42a15ca8fb52d78158d11a429baf2af52528ba/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java#L279

- LangChain4j version: 0.34
- LLM(s) used: azure-open-ai
- Java version: Coretto-21 JDK
- Spring Boot version (if applicable): 3.3.4

**Additional context**

qiaoleiatms:
## Issue

Fixes #1875 

## Change
1. Throw exception when FailingReason is other than CONTENT_FILTER in `AzureOpenAiChatModel`
2. Assert exception is thrown when `model.generate(userMessage)`

## General checklist

### DIFF
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
@@ -315,12 +315,6 @@ public class AzureOpenAiChatModel implements ChatLanguageModel, TokenCountEstima
             return response;
         } catch (HttpResponseException httpResponseException) {
             logger.info("Error generating response, {}", httpResponseException.getValue());
-            FinishReason exceptionFinishReason = contentFilterManagement(httpResponseException, "content_filter");
-            Response<AiMessage> response = Response.from(
-                    aiMessage(httpResponseException.getMessage()),
-                    null,
-                    exceptionFinishReason
-            );
             ChatModelErrorContext errorContext = new ChatModelErrorContext(
                 httpResponseException,
                 modelListenerRequest,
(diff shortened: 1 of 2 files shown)


langchain4j:
@qiaoleiatms thank you!

ghost (approved):

PR #1896, opened 2024-10-10
Fix that Qwen does not allow tool description to be null (but allows blank string)

jiangsier-xyz:
## Change
Fix that Qwen does not allow tool description to be null (but allows blank string)

## General checklist


ghost (approved):
@jiangsier-xyz thank you!

PR #1901, opened 2024-10-11
[DEPENDENCIES] Upgrade Kotlin version to 2.0.21 for #1897, cleanup dependencies, upgrade Milvus and Guava

kpavlov:
## Issue #1897 

Dependency upgrade before  #1897 

## Change

- Use Kotlin BOM for dependency resolution
- Remove explicit Kotlin dependencies from multiple modules
   This eliminates redundant Kotlin standard library dependencies across various `pom.xml` files. The change resolves potential conflicts and reduces unnecessary dependency overhead.
* Updated `guava` version from `32.0.0-jre` to `32.1.2-jre` in `langchain4j-zhipu-ai/pom.xml`.
* Updated Milvus SDK to version 2.4.2 

## General checklist


langchain4j:
@kpavlov thank you! Some modules still use `kotlin-stdlib-jdk8` (in this PR), is it on putpose?

A few things to make sure we are on the same page:
- `kotlin-stdlib-jdk8` is currently used only for okhttp and we plan to replace okhttp with jdk 11 httpclient by default, although okhttp could be used as well
- Does okhttp/retrofit works fine with `kotlin-stdlib` instead of `kotlin-stdlib-jdk8`?
- I cannot commit to keep Kotlin-related stuff in this repository, I need to understand the scope first

kpavlov:
Kotlin std-library dependency is `kotlin-stdlib`. `kotlin-stdlib-jdk8` should be used only on jdk 8 with Kotlin < 1.8. 
Since kotlin-bom is upgraded, it will get `kotlin-stdlib-jdk8` 2.0.21 automatically.
https://kotlinlang.org/docs/maven.html#set-dependencies

> Some modules still use kotlin-stdlib-jdk8 (in this PR), is it on putpose?

Nope, I didn't notice it, will fix.

I don't understand, why this dependency was added? In compile scope. There are no references to Kotlin stdlib code

> Does okhttp/retrofit works fine with kotlin-stdlib instead of kotlin-stdlib-jdk8?

That's the unit and integration testing for. `kotlin-stdlib-jdk8` will be used as transitive dependency.

kpavlov:
Let's merge #1913 first, I am not feeling comfortable to continue without tests

langchain4j-jlama/pom.xml
```diff
@@ -81,17 +82,6 @@
             <type>test-jar</type>
             <scope>test</scope>
         </dependency>
-
-        <dependency>
-            <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
```
langchain4j, line 91: Why removing these? They are required for `JlamaAiServicesWithToolsIT`

langchain4j:
> I don't understand, why this dependency was added? In compile scope. There are no references to Kotlin stdlib code

https://github.com/langchain4j/langchain4j/issues/1506

langchain4j:
> That's the unit and integration testing for.

Tests can miss something, it is never a bad idea to double-check

ghost (commented):
@katcharov thanks!

langchain4j-jlama/pom.xml
kpavlov: As Java 21 and 23 builds are green, it means it's safe to removed this dependency here.
It is now transitive from [parent pom](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-parent/pom.xml#L363-L369).

langchain4j:
@kpavlov there are conflicts now, sorry about that

ghost (approved):
@kpavlov thank you!

PR #1907, opened 2024-10-11
feat: Tracking input tokens on embeddings store ingestor

### LINKED ISSUE #1906: [FEATURE] Tracking tokens on embedding store ingestor.

I would like to know how many tokens were consumed during the embedding process used on my RAG application. 

I need to receive the input tokens when I call the "ingest" method. Then, I will be able to track the cost of this embedding request.

The core class EmbeddingStoreIngestor should return the TokenUsage resource after perform the embedding process.

**Additional context**

pguedes17:
## Issue

Closes #1906 

## Change

Refactor the `EmbeddingStoreIngestor` class to return `IngestionResult` instead of `void` on `ingest` method call. It will allow to track the cost of this embedding request.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java
@@ -204,11 +204,11 @@ public class EmbeddingStoreIngestor {
         embeddingStore.addAll(embeddingsResponse.content(), segments);
         log.debug("Finished storing {} text segments into the embedding store", segments.size());
 
-        return IngestionResult.builder()
-            .content(embeddingsResponse.content())
-            .metadata(embeddingsResponse.metadata())
-            .tokenUsage(embeddingsResponse.tokenUsage())
-            .build();
+        return new IngestionResult(
+                embeddingsResponse.content(),
+                embeddingsResponse.tokenUsage(),
+                embeddingsResponse.metadata()
+                );
     }
 
     /**
--- a/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java
@@ -124,7 +124,7 @@ public class EmbeddingStoreIngestor {
      * <br>
      * For the "Easy RAG", import {@code langchain4j-easy-rag} module,
      * which contains a {@code DocumentSplitterFactory} and {@code EmbeddingModelFactory} implementations.
-     * @return the token usage of the embedding process.
+     * @return result including information related to ingestion process.
      */
     public static IngestionResult ingest(Document document, EmbeddingStore<TextSegment> embeddingStore) {
         return builder().embeddingStore(embeddingStore).build().ingest(document);
@@ -138,6 +138,7 @@ public class EmbeddingStoreIngestor {
      * <br>
      * For the "Easy RAG", import {@code langchain4j-easy-rag} module,
      * which contains a {@code DocumentSplitterFactory} and {@code EmbeddingModelFactory} implementations.
+     * @return result including information related to ingestion process.
      */
     public static IngestionResult ingest(List<Document> documents, EmbeddingStore<TextSegment> embeddingStore) {
         return builder().embeddingStore(embeddingStore).build().ingest(documents);
@@ -148,6 +149,7 @@ public class EmbeddingStoreIngestor {
      * during the creation of this {@code EmbeddingStoreIngestor}.
      *
      * @param document the document to ingest.
+     * @return result including information related to ingestion process.
      */
     public IngestionResult ingest(Document document) {
         return ingest(singletonList(document));
@@ -158,6 +160,7 @@ public class EmbeddingStoreIngestor {
      * during the creation of this {@code EmbeddingStoreIngestor}.
      *
      * @param documents the documents to ingest.
+     * @return result including information related to ingestion process.
      */
     public IngestionResult ingest(Document... documents) {
         return ingest(asList(documents));
@@ -168,6 +171,7 @@ public class EmbeddingStoreIngestor {
      * during the creation of this {@code EmbeddingStoreIngestor}.
      *
      * @param documents the documents to ingest.
+     * @return result including information related to ingestion process.
      */
     public IngestionResult ingest(List<Document> documents) {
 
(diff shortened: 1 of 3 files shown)


langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java
```diff
@@ -122,9 +124,10 @@ private static EmbeddingModel loadEmbeddingModel() {
      * <br>
      * For the "Easy RAG", import {@code langchain4j-easy-rag} module,
      * which contains a {@code DocumentSplitterFactory} and {@code EmbeddingModelFactory} implementations.
+     * @return the token usage of the embedding process.
      */
-    public static void ingest(Document document, EmbeddingStore<TextSegment> embeddingStore) {
-        builder().embeddingStore(embeddingStore).build().ingest(document);
+    public static TokenUsage ingest(Document document, EmbeddingStore<TextSegment> embeddingStore) {
```
glaforge: I'm wondering if we shouldn't return something like an `IngestionResult` class instead, which would contain a `getTokenUsage()` method. It's an indirection, but in case we want to add more context, more details, of the ingestion, that would not be a breaking change later on as we want to add such extra info (for ex. some embedding models might truncate the text, so it could report if a segment has been truncated, etc).

glaforge (commented):
Maybe add an indirection layer, for potential evolution. See comment.

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java
pguedes17: For sure, makes sense! Thanks for the review! I'll take a look on that!

pguedes17: Hi @glaforge! I adjusted with an intermediary layer called `IngestResult` as you suggested. Also I included additional information included on Embeddings response that could be used :)

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java
```diff
@@ -1,25 +1,25 @@
...
 import java.util.Collection;
 import java.util.List;
-
-import static dev.langchain4j.internal.Utils.getOrDefault;
-import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
-import static dev.langchain4j.spi.ServiceHelper.loadFactories;
-import static java.util.Arrays.asList;
-import static java.util.Collections.singletonList;
-import static java.util.stream.Collectors.toList;
+import lombok.extern.slf4j.Slf4j;
```
kpavlov: Lombok [should not be used](https://github.com/langchain4j/langchain4j/blob/main/CONTRIBUTING.md?plain=1#L12). Soon it will be JDK 17 => data classes.

kpavlov (commented):
The test is missing...

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java
pguedes17: This dependency was not added by me. Should I refactor?

langchain4j: @pguedes17 please just do not introduce new lombok usage

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/IngestionResult.java
```diff
@@ -0,0 +1,48 @@
...
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.model.output.TokenUsage;
+import java.util.List;
+import java.util.Map;
+import lombok.Builder;
+
+/**
+ * Represents the result of a {@link EmbeddingStoreIngestor} ingest process.
```
langchain4j:
```suggestion
 * Represents the result of a {@link EmbeddingStoreIngestor} ingestion process.
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/IngestionResult.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Map;
+
+/**
+ * Represents the result of a {@link EmbeddingStoreIngestor} ingest process.
+ */
+public class IngestionResult {
+    /**
+     * The embeddings content that were ingested.
+     */
+    private final List<Embedding> content;
```
langchain4j: This PR is about tracking token usage. Why are embeddings and metadata included? Is this reqlly required? I would rather remove them and return only `TokenUsage`. We can always add other things later, but it does not seem to be very useful right now.

ghost (commented):
@pguedes17 thank you!

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/IngestionResult.java
pguedes17: Sure! Makes sense! fixed!

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/IngestionResult.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+public class IngestionResult {
+    /**
+     * The token usage information.
+     */
+    private final TokenUsage tokenUsage;
+
+
+    public IngestionResult(TokenUsage tokenUsage) {
+        this.tokenUsage = ensureNotNull(tokenUsage, "tokenUsage");
```
langchain4j:
```suggestion
        this.tokenUsage = tokenUsage;
```

There are `EmbeddingModel` implementations that do not return `TokenUsage` (e.g. `OllamaEmbeddingModel`). This will fail then.

ghost (commented):
@pguedes17 thanks, almost there!

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/IngestionResult.java
pguedes17: makes sense! done! thank you so much! :)

ghost (approved):
@pguedes17 thank you!

PR #1928, opened 2024-10-15
Add support for multi-value metadata

### LINKED ISSUE #1905: [FEATURE] Support multi-value document metadata

Currently it is not possible to store multi-value document metadata (e.g. list of uuids) in the embedding store and to filter them during retrieval. This would allow us to implement role-based access control using metadata filtering.

Ideally we would like to store a list of uuids (representing a list of roles) as document metadata in the embedding store (Elasticsearch). We could then use a metadata filter for restricting retrieved text segments based on the assigned roles of the user.

Elasticsearch supports storing any number of values per field. In a different context we already use a multi-valued keyword field in Elasticsearch to store all role uuids that are allowed to read a document. During retrieval we use a bool-query consisting of multiple term-queries (one for each assigned role of the user) combined with a logical-OR to efficiently restrict read-access to documents. We would like to do the same with langchain4j!

lglauer:
## Issue

Closes #1905 

## General checklist


ghost (commented):
@lglauer thank you!

Please add an integration test for Elasticsearch (e.g. in the `ElasticsearchEmbeddingStoreKnnIT`) where such metadata is stored and then retrieved back. Thank you!

lglauer:
@langchain4j Done. I added an IT for storing, searching/filtering and retrieving multi valued metadata in Elasticsearch.

Boonichi:
Hi, I wanted to check if there are any updates on this PR regarding support for multiple values in `Metadata`. I would really appreciate it, as I am currently unable to retrieve points from the vector store (using Qdrant) that include multiple values. The issue arises because it fails the type checking in `Metadata`.

langchain4j:
Sorry for delays! Prioritizing this higher

ghost (commented):
@lglauer thank you!
Could you please change the implementation ot use `Collection`, not `List`? Reasons:
- are we sure that all underlying stores will keep the order of the elements?
- let user use `Set<T>` as well

Boonichi:
Hi @lglauer @langchain4j, can I help you with this PR? My project is blocked due to this feature, although I'm hesitant to override @lglauer's work. Please let me know. Thanks

lglauer:
@Boonichi I started to refactor the PR to use collections. Sorry for not responding, I'm short on time currently.

langchain4j:
@lglauer thanks a lot! Would you be able to do it this week? There is a release planned for next week, would be greate to get it in. No pressure though, if you don't have enough time, perhars @Boonichi could finish?

lglauer:
@langchain4j I updated the PR. Please have a look and review it.

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
```diff
@@ -65,26 +61,45 @@ public Metadata() {
      * Constructs a Metadata object from a map of key-value pairs.
      *
      * @param metadata the map of key-value pairs; must not be {@code null}. {@code null} values are not permitted.
-     *                 Supported value types: {@link String}, {@link Integer}, {@link Long}, {@link Float}, {@link Double}
+     *                 Supported value types: {@link String}, {@link Integer}, {@link Long}, {@link Float}, {@link Double}, {@link Collection}
```
mariofusco, line 64:
```suggestion
     *                 Supported value types: {@link String}, {@link Integer}, {@link Long}, {@link Float}, {@link Double}, and {@link Collection} of the former types.
```

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
```diff
@@ -65,26 +61,45 @@ public Metadata() {
...
+
+    private static void validateValueCollection(String key, Collection<?> values) {
+        values.forEach(value -> {
+            if (!SUPPORTED_VALUE_TYPES.contains(value.getClass())) {
+                throw illegalArgument("The metadata key '%s' collection contains a value '%s', which is of the unsupported type '%s'. " +
+                                "Currently, the supported types are: %s",
+                        key, value, value.getClass().getName(), SUPPORTED_VALUE_TYPES
+                );
+            }
+        });
```
mariofusco, line 100:
```suggestion
        values.forEach(value -> validateValue(key, value));
```

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
```diff
@@ -65,26 +61,45 @@ public Metadata() {
...
+        if (!SUPPORTED_VALUE_TYPES.contains(value.getClass())) {
+            throw illegalArgument("The metadata key '%s' has the value '%s', which is of the unsupported type '%s'. " +
+                            "Currently, the supported types are: %s",
+                    key, value, value.getClass().getName(), SUPPORTED_VALUE_TYPES
+            );
+        }
+    }
+
+    private static void validateValueCollection(String key, Collection<?> values) {
+        values.forEach(value -> {
```
mariofusco, line 93: Nitpick: are we ok with collections with values of different types? Does it make sense for a collection to have a String and an Integer? And if doesn't should we check and prevent this situation?

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
```diff
@@ -124,6 +139,30 @@ public String getString(String key) {
...
+     * @param key the key
+     * @return the collection of {@code String} values associated with the given key, or an empty collection if the key is not present.
+     * @throws RuntimeException if the value is not of type Collection
+     */
+    public Collection<String> getStrings(String key) {
+        if (!containsKey(key)) {
+            return Collections.emptyList();
+        }
+
+        Object values = metadata.get(key);
```
mariofusco, line 154: The double lookup in the map could be avoided (and yes this should be applied to all other getters in this class).

```suggestion
        Object values = metadata.get(key);
        if (values == null) {
            return Collections.emptyList();
        }
```

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
```diff
@@ -124,6 +139,30 @@ public String getString(String key) {
...
+        if (values instanceof Collection<?>) {
+            return ((Collection<?>) values).stream()
+                    .filter(String.class::isInstance)
```
mariofusco, line 157: Should we throw an exception if the collection contains a value that isn't a String here, instead of silently filtering it away?

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
mariofusco: If I understand correctly there are meaningful use cases only for collections of `String`s and `UUID`s? If that's correct should we enforce this also in the collection validation?

mariofusco:
@lglauer I added a few comments and suggestions, but they are all quite minor and I'm open to discuss all of them. My main concern is allowing in the metadata only the kind of values that can be meaningful and just that. Please let me know your feedback, so we will try to include this improvement in the next minor release. 

/cc @langchain4j

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
lglauer: @mariofusco My access control use case (#1905) is fully covered by String and UUID types.
@Boonichi what about storing points in Qdrant?

Currently only getter for `Collection<String>` and `Collection<UUID>` exist, but it would be easy to implement them for all supported types.

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
lglauer: Metadata are a quite abstract construct, so I tend to be lenient and allow collections of mixed types here.
Also JSON allows arrays of mixed types.

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
lglauer: I chose the filtering approach to allow collections of mixed types.

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
langchain4j: Beware that not all stores are using JSON to store metadata, and not all of them (not sure even any of them) support mexed types. Did you check what stores support what datatypes?

langchain4j: In general, it would be great to know which stores support collections of which types. And what filtering operations are allowed for those types

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
langchain4j: Does any store support collections of mixed types?

langchain4j-core/src/main/java/dev/langchain4j/data/document/Metadata.java
Boonichi: For my cases, support for the String and UUID types is enough

PR #1943, opened 2024-10-18
Make DefaultAiServices public

vladislavkn:
## Issue

Scope: `package dev.langchain4j.service`

Currently, the DefaultAiServices class is not publicly accessible, limiting its usability for libraries built on langchain4j. Developers are restricted to using the standard building method of AiServices that selects some factory method for creating AI services, which may not align with lbraries specific architectural needs. This restricts flexibility and reusability in extending AI services functionality.

## Change
This merge request changes the visibility of the DefaultAiServices class to public. By doing so, it allows libraries that are built upon langchain4j to directly utilize the default AI services builder instead of relying solely on the random factory method.

## General checklist

### DIFF
--- a/docs/docs/tutorials/5-ai-services.md
+++ b/docs/docs/tutorials/5-ai-services.md
@@ -74,6 +74,10 @@ In a Quarkus or Spring Boot application, this can be a bean that you can then in
 wherever you need AI Services.
 :::
 
+:::note
+The `AiServices.create` method implements the factory pattern, allowing other libraries built on langchain4j to provide their own factory methods. If you require direct access to the default constructor, consider using `new DefaultAiServices()`.
+:::
+
 Now we can use `Assistant`:
 ```java
 String answer = assistant.chat("Hello");
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -58,7 +58,7 @@ import static dev.langchain4j.service.TypeUtils.typeHasRawClass;
 import static dev.langchain4j.service.output.JsonSchemas.jsonSchemaFrom;
 import static dev.langchain4j.spi.ServiceHelper.loadFactories;
 
-class DefaultAiServices<T> extends AiServices<T> {
+public class DefaultAiServices<T> extends AiServices<T> {
 
     private static final int MAX_SEQUENTIAL_TOOL_EXECUTIONS = 100;
 


langchain4j:
@vladislavkn thank you!

What do you mean by "random factory method"?

vladislavkn:
"Random" is not the best matching word. I've meant that another library, e.g. [quarkus-langchain4j](https://github.com/quarkiverse/quarkus-langchain4j) can implement an aiService factory method that differs from the default (`DefaultAiServices`). In this case the creation of the aiServices in the way described in the [langchain4j docs](https://docs.langchain4j.dev/tutorials/ai-services/#simplest-ai-service) can not work: in the example with quarkus-langchain4j the aiService interface must be additionally annotated with `CreatedAware`. Sometimes this requirement cannot be satisfied, e.g. when the aiService is initiated inside of another library that works with "pure" langchain4j. 

This PR is intended to provide a way to such libraries to explicitly select the DefaultAiServices method of creation of aiServices.

Please check if it is still unclear or if the usecase described can be fixed in another way. Thank you for you help 🙂

langchain4j:
> I've meant that another library, e.g. [quarkus-langchain4j](https://github.com/quarkiverse/quarkus-langchain4j) can implement an aiService factory method that differs from the default (DefaultAiServices)

This is how it is done now, no? Quarkus extension has it's own implementation.

> In this case the creation of the aiServices in the way described in the [langchain4j docs](https://docs.langchain4j.dev/tutorials/ai-services/#simplest-ai-service) can not work

AFAIK in Quarkus extension one is not supposed to create AI service programmaticaly. @geoand could you please comment?

> Please check if it is still unclear or if the usecase described can be fixed in another way.

Yes, please expand a bit, I am still not sure I understand the use case. An example would be very helpful.

geoand:
> AFAIK in Quarkus extension one is not supposed to create AI service programmaticaly. @geoand could you please comment?

There is actually some wizardry to make it work :)

vladislavkn:
> Yes, please expand a bit, I am still not sure I understand the use case. An example would be very helpful.

I'm working on a Java library for model quality evaluation, similar to Ragas. It currently uses only Langchain4j, not the Quarkus extension. Although still in development and internal testing, the library is expected to be open-sourced.

The core library has evaluators that work with LLMs. An example constructor looks like:

```java
public FaithfulnessEvaluator(ChatLanguageModel model) {
    bot = AiServices.create(FaithfulnessBot.class, model);
}
```

Here, `FaithfulnessBot` is just an interface as described in Langchain4j docs.

I want to use the library in a production project that uses Quarkus and Quarkus-Langchain4j. When creating a `FaithfulnessEvaluator` instance in the project, I get an error during `aiService` creation:

```java
@Inject
public Metrics(ChatLanguageModel chatLanguageModel, EmbeddingModel embeddingModel) {
    faithfulness = new FaithfulnessEvaluator(chatLanguageModel);
    // ...
}
```

The error:

```
java.lang.RuntimeException: Quarkus was not able to determine class 'dev.rage4j.evaluation.faithfulness.FaithfulnessBot' as an AiService at build time. Consider annotating the class with '@CreatedAware'
```

To avoid this error I wanted to use the `DefaultAiServices` class for the creation of the aiServices.

I want to keep the library core as clean and small as possible and the question is whether there is better way to integrate the library in the Quarkus project?

geoand:
A very simple thing to try is to add an empty `META-INF/beans.xml` file to your library which should make Quarkus aware of the code in it.

vladislavkn:
> A very simple thing to try is to add an empty META-INF/beans.xml file to your library which should make Quarkus aware of the code in it.

Thank you very much, I will try it!

geoand:
Let me know how it goes :)

vladislavkn:
It seems to work, thank you for the solution. I think that the PR can be closed.

PR #1949, opened 2024-10-18
Fix #1747

### LINKED ISSUE #1747: [BUG] Milvus query result metadata data accuracy loss
I used Milvus to query, and in the returned result, metadata={id: xxxxx} The ID value I saved in the database is a number with a length of 19, but the query result seems to get a Scientific notation? And the accuracy is lost. Is this a problem with the Milvus database?

image

ghost:
## Issue
Fixes #1747

## Change
Fixed this bug for Milvus and Chroma.

Did not manage to fix it for Pinecone because Pinecone can store `Integer`/`Long` only as `Double`. We could theoretically store them as `String`s, but then metadata filtering for `Integer`/`Long` values will not work. For now I've left the comment in the `PineconeEmbeddingStore` Javadoc and Pinecone documentation about the bug and possible workaround.

## General checklist

### DIFF
--- a/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreIT.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreIT.java
@@ -63,6 +63,9 @@ public abstract class EmbeddingStoreIT extends EmbeddingStoreWithoutMetadataIT {
         assertThat(match.embedded().metadata().getLong("long_minus_1")).isEqualTo(-1L);
         assertThat(match.embedded().metadata().getLong("long_0")).isEqualTo(0L);
         assertThat(match.embedded().metadata().getLong("long_1")).isEqualTo(1L);
+        if (testLong1746714878034235396()) {
+            assertThat(match.embedded().metadata().getLong("long_1746714878034235396")).isEqualTo(1746714878034235396L);
+        }
         assertThat(match.embedded().metadata().getLong("long_max")).isEqualTo(Long.MAX_VALUE);
 
         assertThat(match.embedded().metadata().getFloat("float_min")).isEqualTo(-Float.MAX_VALUE);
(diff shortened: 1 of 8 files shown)


langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreIT.java
```diff
@@ -63,6 +63,9 @@ void should_add_embedding_with_segment_with_metadata() {
         assertThat(match.embedded().metadata().getLong("long_minus_1")).isEqualTo(-1L);
         assertThat(match.embedded().metadata().getLong("long_0")).isEqualTo(0L);
         assertThat(match.embedded().metadata().getLong("long_1")).isEqualTo(1L);
+        if (testLong1746714878034235396()) {
```
Martin7-1, line 66: Maybe this change will affect other subclass's IT?

Martin7-1 (commented):
Looks good to me, but the change of `EmbeddingStoreIT` may break other `EmbeddingStore`, so we need to run all `EmbeddingStore`'s IT to check.

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreIT.java
langchain4j: For those stores that I can test, tests are green. I can't test only Azure Cosmos and Tablestore, but I don't see why this change should affect them, unless this bug is present there, in which case it is good that it will fail and expose it

Martin7-1: Exactly...What I mean it's that other embedding stores may also have this bug.

langchain4j: Chroma had it and this change exposed it :)

Martin7-1 (approved):

PR #1950, opened 2024-10-18
Google AI Gemini: replace OkHttp and Retrofit with Java 11 HttpClient

Bjarne-Kinkel:
## Issue
Based on #1903

## Change
Replaced OkHttp and Retrofit inside the GeminiService with an implementation using the HttpClient (Java 11).

## General checklist

- [X] There are no breaking changes
- [X]  I have added unit and integration tests for my change (already existing)
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [X] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new maven module

- [ ] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #1951, opened 2024-10-18
Adding GoogleAiGeminiStreamingChatModel

### LINKED ISSUE #1903: Google AI Gemini: implement streaming model
### Discussed in https://github.com/langchain4j/langchain4j/discussions/1902

<div type='discussions-op-text'>

<sup>Originally posted by **Bjarne-Kinkel** October 11, 2024</sup>
Is there allready an efford to implement a GoogleAiGeminiStreamingChatModel as mentioned [here](https://docs.langchain4j.dev/integrations/language-models/google-ai-gemini#googleaigeministreamingchatmodel)? Or at least a feature request?</div>

Bjarne-Kinkel:
## Issue
Closes #1903

## Change
Implemented the `GoogleAiGeminiStreamingChatModel`.

This PR depends on #1950.

## General checklist


langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,224 @@
...
+                                            Duration timeout,
+                                            ResponseFormat responseFormat,
+                                            List<String> stopSequences, GeminiFunctionCallingConfig toolConfig,
+                                            Boolean allowCodeExecution, Boolean includeCodeExecutionOutput,
+                                            Boolean logRequestsAndResponses,
+                                            List<GeminiSafetySetting> safetySettings,
+                                            List<ChatModelListener> listeners) {
+        this.apiKey = ensureNotBlank(apiKey, "apiKey");
+        this.modelName = ensureNotBlank(modelName, "modelName");
+        this.temperature = getOrDefault(temperature, 1.0);
```
langchain4j: @glaforge maybe it makes sense to not set default parameters here?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,224 @@
...
+        this.topK = getOrDefault(topK, 64);
+        this.topP = getOrDefault(topP, 0.95);
+        this.maxOutputTokens = getOrDefault(maxOutputTokens, 8192);
+        this.candidateCount = getOrDefault(candidateCount, 1);
+        this.stopSequences = getOrDefault(stopSequences, emptyList());
+        this.toolConfig = toolConfig;
+        this.allowCodeExecution = allowCodeExecution != null ? allowCodeExecution : false;
```
langchain4j:
```suggestion
        this.allowCodeExecution = getOrDefault(allowCodeExecution, false);
```

nit: please use the same style of setting default parameters consistently (and a few places below as well)

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,224 @@
...
+    private final GeminiFunctionCallingConfig toolConfig;
+    private final boolean allowCodeExecution;
+    private final boolean includeCodeExecutionOutput;
+    private final Boolean logRequestsAndResponses;
+    private final List<GeminiSafetySetting> safetySettings;
+    private final List<ChatModelListener> listeners;
+
+    @Builder
+    public GoogleAiGeminiStreamingChatModel(String apiKey, String modelName,
+                                            Integer maxRetries,
```
langchain4j: please remove `maxRetries`

Bjarne-Kinkel:
@glaforge, I’ve merged the new version of main. Could you please take a final look at this PR and approve it? In the meantime, I will update documentation.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiStreamingResponseBuilder.java
```diff
@@ -0,0 +1,109 @@
...
+        boolean hasText = !text.isEmpty() && !text.isBlank();
+
+        if (hasText && !functionCalls.isEmpty()) {
+            return new AiMessage(text, functionCalls);
+        } else if (hasText) {
+            return new AiMessage(text);
+        } else if (!functionCalls.isEmpty()) {
+            return new AiMessage(functionCalls);
+        }
+        return new AiMessage("No valid return.");
```
langchain4j: This does not seem right... When this can happen? Should we throw exception in this case?

ghost (commented):
@Bjarne-Kinkel thank you!

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
Bjarne-Kinkel: @glaforge mentioned on Discord that this is something he would like to remove in the near future. (It’s also in the standard GoogleGeminiChatModel; I just copied it from there.)

Bjarne-Kinkel:
I just remembered that the StreamingChatModel doesn't retry if a request fails. I believe it's the same with the Vertex Streaming AI version. However, other streaming models do have retries. Should I implement that as well?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
langchain4j: Right, it makes sense to remove them from existing `GoogleAiGeminiChatModel`. But does it make sense to add them to the new `GoogleAiGeminiStreamingChatModel`, just to be removed soon after and having a risk of breaking someone's code that already started relying on those defaults?

Bjarne-Kinkel: You're right. I will remove them.

langchain4j:
Hi @Bjarne-Kinkel 

> I just remembered that the StreamingChatModel doesn't retry if a request fails. I believe it's the same with the Vertex Streaming AI version. However, other streaming models do have retries. Should I implement that as well?

Good catch! It seems that only Azure OpenAI and GitHub Models support retries for streaming, and this is implemented by their SDKs. It would be interesting to see how they handle this.

TBH I am not completely sure how retries for streaming models should behave. If it fails before the first token is streamed, we could just retry, ok. But what if it fails in the middle of streaming (when `onNext()` has been called already). How `StreamingResponseHandler` should behave in this case? Call `onError()`, then retry request and start calling `onNext()` again?

cc @cescoffier

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
langchain4j: Thank you!

Bjarne-Kinkel: Are there any default values I should leave as is, such as temperature=0? What about parameters like topK, topP, candidateCount, or maxTokenOutput? What should happen if these aren't set? Should an exception be thrown when trying to build the model? How do other models handle this? Am I correct in assuming that OpenAI just requires an API key?

langchain4j: @Bjarne-Kinkel please remove the default values for optional parameters. If you send request to the LLM without this parameter and it does not fail, then it is safe to remove.
I guess these should have no defaults: temperature, topK, topP, maxOutputTokens, 

I guess candidateCount should be hardcoded to 1 and should not be configurable as LC4j API does not support multiple generations. cc @glaforge

Bjarne-Kinkel: I've removed default values where possible and, for now, set candidateCount to 1 by default.

cescoffier:
@langchain4j About retry and stream.

Of course, the official answer is: "it depends". 

More specifically, it depends on the nature of the stream source: cold or hot (https://smallrye.io/smallrye-mutiny/latest/guides/hot-streams/). 

In this case, I would consider the stream to be "cold," and thus, a retry would be a re-subscription to the stream. I expect the full answer to be sent back. At least, that's how I implemented it in Quarkus Langchain4J. So, depending on how the stream is consumed, it may get duplicated content, but that's fine.

langchain4j:
@cescoffier thanks for the feedback!

I assume that in a typical chatbot application tokens are streamed from LLM to the backend and then from backend to the UI.
If at some point streaming between LLM and backend breaks (and some part of the response has been already displayed to the end user), the `onError()` will be called, so developer should be able to handle it and notify end user in some way (e.g., remove previously displayed tokens). In this case we could indeed retry automatically. Same request to the LLM will be sent again and `onNext()` will start being called for each new token and streamed to the UI. If developer does not handle this situation in `onError()` properly, user might get confusing response (a mix of old and new response). But this is on the developer to handle it properly.

BTW did you encounter problems with streaming already (when it breaks in the middle)? If yes, how often?

Bjarne-Kinkel:
So far, I have only experienced one abrupt interruption in a stream with the Gmini AI API. A specific request always led to an abrupt end after the first few messages. However, it seemed to be related to safety settings that terminated the stream "properly". I received an empty message in the middle of a sentence with the finish reason being "safety", even though I didn’t think the text was "dangerous". However, it was not an issue with my implementation; a simple curl request via the console behaved exactly the same way.

@langchain4j Should I go ahead and implement the retry feature?

Bjarne-Kinkel:
I tested what happens when the internet connection is turned off during streaming. Instead of causing an error, the stream simply pauses, then resumes from where it left off once the connection is restored, delivering the remaining text all at once. This seems like a viable solution. However, it may be challenging to detect a connection loss when the HttpClient doesn’t throw an error. Other types of stream interruptions are also difficult to test. Still, retrying after an error during the initial connection would probably be a good idea.

langchain4j:
@Bjarne-Kinkel hmm, for how long did you cut the internet connection?

Bjarne-Kinkel:
@langchain4j about 30-60 seconds.

langchain4j:
@Bjarne-Kinkel perharps you did not get any error because it was shorter than a timeout?

Bjarne-Kinkel:
@langchain4j I waited for almost 5 minutes (timeout was 60 seconds), and surprisingly, I didn’t get an error. However, after reconnecting, the stream didn’t continue, and the program was stuck.

langchain4j:
@Bjarne-Kinkel this happens with new http client or with okhttp?

Bjarne-Kinkel:
@langchain4j Streaming is only implemented with the HttpClient; there is no OkHttp implementation for streaming. When using curl it depends on how short the connection-loss is. 10 secounds, no problem, but about 30 and the connection can't be picked up again.

Bjarne-Kinkel:
Okay, I did some digging, and it turns out the HttpClient doesn’t support detecting timeouts during the stream; the timeout only applies to connection establishment. I suggest implementing a retry mechanism for establishing the connection and leaving the rest as it is for now. What do you think, @langchain4j ?

langchain4j:
@Bjarne-Kinkel sounds good, thank you!

Bjarne-Kinkel:
@langchain4j okay, done. The PR should be ready for a final review.

glaforge (dismissed):
Looking great, thank you!

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,228 @@
...
+            handler.onComplete(fullResponse);
+
+            notifyListenersOnResponse(fullResponse, chatModelRequest, listenerAttributes);
+        } catch (RuntimeException exception) {
+            notifyListenersOnError(exception, chatModelRequest, listenerAttributes);
+            handler.onError(exception);
+        }
+    }
+
+    private static String computeMimeType(ResponseFormat responseFormat) {
```
langchain4j: Can this method be shared with `GoogleAiGeminiChatModel` to avoid duplication?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,228 @@
...
+        );
+    }
+
+    @Override
+    public void generate(List<ChatMessage> messages, StreamingResponseHandler<AiMessage> handler) {
+        generate(messages, Collections.emptyList(), handler);
+    }
+
+    @Override
+    public void generate(List<ChatMessage> messages, ToolSpecification toolSpecification, StreamingResponseHandler<AiMessage> handler) {
```
langchain4j, line 54: This method is supposed to **_force_** the LLM to call provided `ToolSpecification`, but it seems that existing implementation acts as if this is one of the tools that **_can_** be called.
If there is no way to enforce Gemini to call specific tool, I think this method should not be implemented (and thus throw an exception, as is done in its default impl)

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,228 @@
...
+                                            List<GeminiSafetySetting> safetySettings,
+                                            List<ChatModelListener> listeners,
+                                            Integer maxRetries) {
+        this.apiKey = ensureNotBlank(apiKey, "apiKey");
+        this.modelName = ensureNotBlank(modelName, "modelName");
+        this.temperature = temperature;
+        this.topK = topK;
+        this.topP = topP;
+        this.maxOutputTokens = maxOutputTokens;
+        this.candidateCount = 1; // Multiple candidates aren't supported by langchain4j
```
langchain4j: nit: if it is always 1, perharps remove this parameter from class fields?

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
```diff
@@ -0,0 +1,752 @@
...
+        TestStreamingResponseHandler<AiMessage> handler = new TestStreamingResponseHandler<>();
+        gemini.generate(userMessage, handler);
+        Response<AiMessage> response = handler.get();
+
+        // then
+        assertThat(response.content().text()).containsIgnoringCase("Pixel");
+    }
+
+    @Test
+    void should_support_video_file() {
```
langchain4j, line 229: nit: please mark this `@Disabled`

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
```diff
@@ -0,0 +1,752 @@
...
+        allMessages.add(UserMessage.from("What is the weather forecast in Paris?"));
+
+        // when
+        TestStreamingResponseHandler<AiMessage> handler1 = new TestStreamingResponseHandler<>();
+        gemini.generate(
+                allMessages,
+                ToolSpecification.builder()
+                        .name("getWeatherForecast")
+                        .description("Get the weather forecast for a given city")
+                        .addParameter("city", JsonSchemaProperty.STRING)
```
langchain4j: please use new API for tools (here and a few times below)

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,228 @@
...
+        listeners.forEach((listener) -> {
+            try {
+                listener.onRequest(context);
+            } catch (Exception e) {
+                log.warn("Exception while calling model listener (onRequest)", e);
+            }
+        });
+    }
+
+    private void notifyListenersOnResponse(Response<AiMessage> fullResponse, ChatModelRequest chatModelRequest,
```
langchain4j: perharps some logic can be shared with `GoogleAiGeminiChatModel`?

ghost (commented):
@Bjarne-Kinkel thank you!

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
Bjarne-Kinkel: Yes. Should I leave it in the GoogleAiGeminiChatModel-class or put it in a new or another file?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
Bjarne-Kinkel: This would require a more extensive refactor of the GoogleAiGeminiChatModel class, but I can certainly do it. However, the same question as above arises: which class should contain the logic? A new class? Perhaps even an abstract one?

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
Bjarne-Kinkel: @langchain4j which one do you mean? The one using @Tool?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiService.java
```diff
@@ -48,6 +50,11 @@ GoogleAiBatchEmbeddingResponse batchEmbed(String modelName, String apiKey, Googl
         return sendRequest(url, apiKey, request, GoogleAiBatchEmbeddingResponse.class);
     }
 
+    Stream<GeminiGenerateContentResponse> generateContentStream(String modelName, String apiKey, GeminiGenerateContentRequest request) {
+        String url = String.format("%s/models/%s:streamGenerateContent?alt=sse", GEMINI_AI_ENDPOINT, modelName);
```
kpavlov, line 54: Side question: How can I override this endpoint, e.g. in testing, when I want to call wiremock on localhost?
Not blocking this PR of course

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiService.java
```diff
@@ -72,6 +79,40 @@ private <T> T sendRequest(String url, String apiKey, Object requestBody, Class<T
...
+                throw new RuntimeException(String.format("HTTP error (%d): %s", httpResponse.statusCode(), errorBody));
+            }
+
+            Stream<T> responseStream = httpResponse.body()
+                    .filter(line -> line.startsWith("data: "))
+                    .map(line -> line.substring(6)) // Remove "data: " prefix
+                    .map(jsonString -> gson.fromJson(jsonString, responseType));
+
+            if (logger != null) {
+                responseStream = responseStream.peek(response -> logger.debug("Partial response from Gemini:\n{}", response));
```
kpavlov, line 104: Nit: It would be great to delegate it to configurable handler chain

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiService.java
```diff
@@ -72,6 +79,40 @@ private <T> T sendRequest(String url, String apiKey, Object requestBody, Class<T
...
+            }
+
+            return responseStream;
+        } catch (IOException e) {
+            throw new RuntimeException("An error occurred while streaming the request", e);
```
kpavlov, line 109: Nit: is there any more specific Langchain4jRuntimeException?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,228 @@
...
+    private final GeminiFunctionCallingConfig toolConfig;
+    private final boolean allowCodeExecution;
+    private final boolean includeCodeExecutionOutput;
+    private final List<GeminiSafetySetting> safetySettings;
+    private final List<ChatModelListener> listeners;
+    private final Integer maxRetries;
+
+    @Builder
+    public GoogleAiGeminiStreamingChatModel(String apiKey, String modelName,
+                                            Double temperature, Integer topK, Double topP,
```
kpavlov: Nit: Would be great to introduce model config object to avoid adding 100500 parameters here

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
Bjarne-Kinkel: Should this then also be the same for the `GoogleGeminiAiChatModel`?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
```diff
@@ -0,0 +1,228 @@
...
+            contentStream.forEach(partialResponse -> {
+                Optional<String> text = responseBuilder.append(partialResponse);
+                text.ifPresent(handler::onNext);
+            });
+
+            Response<AiMessage> fullResponse = responseBuilder.build();
+            handler.onComplete(fullResponse);
+
+            notifyListenersOnResponse(fullResponse, chatModelRequest, listenerAttributes);
+        } catch (RuntimeException exception) {
```
kpavlov, line 88: What if this is not a runtime exception ?

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
```diff
@@ -0,0 +1,752 @@
...
+        UserMessage userMessage = UserMessage.from("What is the capital of France?");
+
+        // when
+        TestStreamingResponseHandler<AiMessage> handler = new TestStreamingResponseHandler<>();
+        gemini.generate(userMessage, handler);
+        Response<AiMessage> response = handler.get();
+
+        // then
+        String text = response.content().text();
+        assertThat(text).containsIgnoringCase("Paris");
```
kpavlov, line 82: Is there a test verifying that concatenation of partial responses is equal to final result?

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
```diff
@@ -0,0 +1,752 @@
...
+        Response<AiMessage> response = handler.get();
+
+        // then
+        assertThat(response.content().text()).containsIgnoringCase("cat");
+    }
+
+    @Test
+    void should_support_text_file() {
+        // given
+        GoogleAiGeminiStreamingChatModel gemini = GoogleAiGeminiStreamingChatModel.builder()
```
kpavlov, line 183: Nit: Model builder should be, probably, initialized in `@BeforeClass` and reused.

kpavlov (approved):
Thank you, @Bjarne-Kinkel 

My comments are not blocking this PR, just suggesting follow up refactorings. 
P.S. Would be nice to adopt Java 17 style.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiService.java
Bjarne-Kinkel: Is there already an implementation like this? I'm not sure how to implement it without adding boilerplate code.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiService.java
Bjarne-Kinkel: No, the Google Gemini AI API only uses generic RuntimeExceptions so far.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
Bjarne-Kinkel: Sure, I can simply catch all exceptions; I based this on the GoogleAiGeminiCatModel class, which also only catches RuntimeExceptions. What other exceptions could even be thrown here? I thought I’m actually already catching all the others beforehand, aren’t I?

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
Bjarne-Kinkel: Yes, it’s inside the `TestStreamingResponseHandler`'s implementation of the `onComplete `method.

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
Bjarne-Kinkel: But it can only be reused for a couple of tests because some have models with different properties. Is it worth it anyway?

Bjarne-Kinkel:
Okay, I have moved similar code from `GoogleAiGeminiChatModel `and `GoogleAiGeminiStreamingChatModel `into a new abstract class called `BaseGeminiChatModel`. Everything should works exactly as before, except for one small change I made in the `chat `function of `GoogleAiGeminiChatModel`: now, every exception thrown during the request and response creation is also sent to the listeners. Previously, only exceptions thrown during the actual request were sent to the listeners. I’m not sure where the listeners are used for at all, but I thought it might make sense. Does it, @langchain4j?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
langchain4j: @Bjarne-Kinkel yes

langchain4j:
@Bjarne-Kinkel great, thank you! I will have a look at it today. 

> Everything should works exactly as before, except for one small change I made in the chat function of GoogleAiGeminiChatModel: now, every exception thrown during the request and response creation is also sent to the listeners. Previously, only exceptions thrown during the actual request were sent to the listeners. I’m not sure where the listeners are used for at all, but I thought it might make sense. Does it, @langchain4j?

Listeners are there for observability. Can you give an example of an exception thrown at response?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
Bjarne-Kinkel: @langchain4j  okay, but there is a problem. The method in the `GoogleGeminiAiChatModel` is already incorrectly implemented. However, Gemini doesn’t offer the option to enforce a tool call. Ideally, the method should throw an exception. Since this function is already implemented incorrectly, changing it would result in a breaking change. Should I still modify the method to throw an exception?

langchain4j: Hmm, it seems that tool call can be enforced by using `mode=ANY` and `allowed_function_names=[{my_tool}]`:
https://ai.google.dev/gemini-api/docs/function-calling?authuser=1#function_calling_mode

Bjarne-Kinkel:
> @Bjarne-Kinkel great, thank you! I will have a look at it today.
> 
> > Everything should works exactly as before, except for one small change I made in the chat function of GoogleAiGeminiChatModel: now, every exception thrown during the request and response creation is also sent to the listeners. Previously, only exceptions thrown during the actual request were sent to the listeners. I’m not sure where the listeners are used for at all, but I thought it might make sense. Does it, @langchain4j?
> 
> Listeners are there for observability. Can you give an example of an exception thrown at response?

@langchain4j If the response from Gemini is empty or lacks a candidate, a `RuntimeException` may be thrown. And in the old version a AIMessage with an errortext was returned. Now an exception gets thrown.

langchain4j:
> > @Bjarne-Kinkel great, thank you! I will have a look at it today.
> > > Everything should works exactly as before, except for one small change I made in the chat function of GoogleAiGeminiChatModel: now, every exception thrown during the request and response creation is also sent to the listeners. Previously, only exceptions thrown during the actual request were sent to the listeners. I’m not sure where the listeners are used for at all, but I thought it might make sense. Does it, @langchain4j?
> > 
> > 
> > Listeners are there for observability. Can you give an example of an exception thrown at response?
> 
> @langchain4j If the response from Gemini is empty or lacks a candidate, a `RuntimeException` may be thrown. And in the old version a AIMessage with an errortext was returned. Now an exception gets thrown.

The current behavior (exception) is better, thank you

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
glaforge: Yes, if we really want to follow the principle that the generate method with a single call means forcing a tool call, then we should use the ANY mode to do that.

But I still find a bit unfortunate that this generate method variant exist, as it was modeled after a specific behavior of OpenAI GPTs. It's not generic and forces all other models to somehow behave like this even if they are not supporting this approach. The name of the method should have been different and should have been explicit about its behavior.

Note that the Vertex AI Gemini behaves as if a list of one tool was passed. So the tool choice is not enforced.
I wonder if it's okay to have the two models behave differently now.

langchain4j: @glaforge I agree that the name is bad, although I disagree about the existence of the method itself. Structured data extraction is a big use case and function calling was used for it for a long time. So there is definitely a need to force the model to call a specific tool. Most LLM providers (OpenAI, Gemini, Bedrock, Anthropic, Mistral, etc) support this behaviour, so I see no problem with this.

> Note that the Vertex AI Gemini behaves as if a list of one tool was passed. So the tool choice is not enforced.
> I wonder if it's okay to have the two models behave differently now.

Oh, I was not aware of this, we will need to fix this. But I guess we can just wait till we migrate to the new `chat(ChatRequest)` API, `ChatRequest` weill eventually have something like `ToolMode`

Bjarne-Kinkel: @langchain4j What does this mean for this PR? Should I implement forced function calls within this PR? If so, for which models? Should I limit it to the streaming version or include the normal model as well? Or should I keep the feature as it is for now and create a new issue to discuss this further?

langchain4j: @Bjarne-Kinkel I guess it makes sense to keep it separate and just remove this method for now (so that default impl will take care of throwing exception). Once the new API will be in place, we can implement this behaviour in a separate PR

Bjarne-Kinkel: Is that for the normal ChatModel implementation as well or just for my streaming version? @langchain4j

langchain4j: let's keep `GoogleAiGeminiChatModel` as it is now and only remove this method from `GoogleAiGeminiStreamingChatModel`

Bjarne-Kinkel: Okay, perfect. Done.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
langchain4j: We need to rethink this for all models, not only Gemini. Let's keep it out of this PR

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
langchain4j: @kpavlov what else can this be?

ghost (approved):
@Bjarne-Kinkel thank you, great job!

Bjarne-Kinkel:
> @Bjarne-Kinkel thank you, great job!

@langchain4j your welcome. Is there anything else do to for me, regarding this PR?

langchain4j:
@Bjarne-Kinkel there are still a few open discussions, could you comment on them please?

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
Bjarne-Kinkel: I could do that, but it would only be usefull for a third of the test since the other tests use different model implementations.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiService.java
Bjarne-Kinkel: Is there already an implementation for a configurable handler chain?

Bjarne-Kinkel:
> My comments are not blocking this PR, just suggesting follow up refactorings.

@langchain4j I thought those were just suggestions, but I can certainly work on that.

langchain4j:
@Bjarne-Kinkel sorry, I missed that those are suggestions. Merging it then!

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModel.java
kpavlov: Maybe Langchain4jException extends RuntimeException. Not important

langchain4j: We plan to intoroduce a set of common exceptions for common situations (e.g. bad request, rate limit, service unavailable, etc), so let's not add anything now that has to be changed again soon

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiStreamingChatModelIT.java
kpavlov: Singletons, caching model or model builderl, would be very useful across the codebase.

KaiSuchomel:
@langchain4j In which langchain4j release will this feature be included???

glaforge:
@KaiSuchomel I don't remember in which version it was introduced, but I think well before the 1.0, several months ago. So if you're using a recent LangChain4j version, you have it already.

KaiSuchomel:
@glaforge Thx for your fast reply. I am using the quarkus-langchain4j-ai-gemini extension. 
I thought they might have used this langchain4j google lib. But it doesnt work with the current (1.1.0) release.
I will have a look in the quarkus project....

Issue #1496, opened 2024-07-20
[BUG] Use deprecated methods in DocumentLoader.java cause type lost in metadata


zhifeiyue:
**Describe the bug**
Use deprecated methods in DocumentLoader.java cause type lost in metadata, then an illegalArgument exception thrown when use a Filter.
```
public static Document load(DocumentSource source, DocumentParser parser) {
  try (InputStream inputStream = source.inputStream()) {
      Document document = parser.parse(inputStream);
      source.metadata().asMap().forEach((key, value) -> document.metadata().add(key, value));
      return document;
      ...
```
`asMap()` and `add()` cause any type in metadata converted to String.

- LangChain4j version: main branch

langchain4j:
Hi @zhifeiyue thanks a lot for reporting! Could you please provide the code to reproduce illegal argument exception with `Filter`?

zhifeiyue:
Hi @langchain4j,
First, use a custom DocumentSource
```

public class YourCustomSource implements DocumentSource {
    public final long contentId;
    public final String contentType;

    ...

    @Override
    public Metadata metadata() {
        return new Metadata().put("content_id", contentId)
            .put("type", contentType);
    }
}
```

then, use InMemoryEmbeddingStore's removeAll with filter:
```
List<Long> contentIdsToDelete = ...
embeddingStore.removeAll(metadataKey("content_id").isIn(contentIdsToDelete));
```

`IsIn::test` will throw exception on the `ensureTypesAreCompatible` line.

langchain4j:
@zhifeiyue ah I see, a custom `DocumentSource` was a missing part

dliubarskyi:
This should be fixed by #2726. I am closing this issue, but feel free to reopen if you still face it

PR #1959, opened 2024-10-19
#1946 Convert Document, Content, DocumentLoader to interfaces

kpavlov:
## Issue
Explains #1946

## Change

Refactor following classes to interfaces for better modularity and readability:
- `Document` to interface and introduce `SimpleDocument` as a record implementation
- `Content` to interface and introduce `TextContent` as a record implementation
- `DocumentLoader` to interface
- `DocumentParser` to use generics

**This is a potentially breaking change,** because public constructors of converted classes were removed. Luckily, static factory methods had been used already in the code.

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentLoader.java
```diff
@@ -5,10 +5,7 @@
 /**
  * Utility class for loading documents.
  */
-public class DocumentLoader {
-
-    private DocumentLoader() {
-    }
+public interface DocumentLoader {
```
langchain4j, line 8: Why is this change required?

langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentParser.java
```diff
@@ -6,7 +6,7 @@
  * Defines the interface for parsing an {@link InputStream} into a {@link Document}.
  * Different document types require specialized parsing logic.
  */
-public interface DocumentParser {
+public interface DocumentParser<T extends Document> {
```
langchain4j, line 9: Why is this required?

langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentLoader.java
```diff
@@ -20,10 +17,10 @@ private DocumentLoader() {
      * @return The loaded document.
      * @throws BlankDocumentException when the parsed {@link Document} is blank/empty.
      */
-    public static Document load(DocumentSource source, DocumentParser parser) {
+    static Document load(DocumentSource source, DocumentParser parser) {
         try (InputStream inputStream = source.inputStream()) {
             Document document = parser.parse(inputStream);
-            source.metadata().asMap().forEach((key, value) -> document.metadata().add(key, value));
+            source.metadata().toMap().forEach((key, value) -> document.metadata().add(key, value));
```
langchain4j, line 23: We need to be careful with this change, we can break something

ghost (commented):
@kpavlov thank you!

langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentLoader.java
kpavlov: Because there is no interface for DocumentLoaders and now e.g. `FileSystemDocumentLoader` or`AmazonS3DocumentLoader` can implement the at least marker interface.
It would be even better to implement :

```java
interface DocumentLoader<T extends Document> {
   T loadDocument(DocumentSource<T> source);
}
```

Spring AI has a similar concept of [DocumentReader](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/document/DocumentReader.java)

langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentLoader.java
kpavlov: I read the deprecation comment, and go ahead with this change, hoping that tests can verify that. I hope this should be fine as long as unit and integration tests are green.
But I agree, this needs care, because Object to String conversion may need a special care

langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentParser.java
kpavlov: To introduce different types of document in the future.

kpavlov:
@langchain4j , would it be fine to start with just the first [commit](https://github.com/langchain4j/langchain4j/pull/1959/commits/b7ce3543a9bcabbf4aafc4dfd4a57fc4d5df77db)? I can revert other commits or create a new PR
It make sense to refactor DocumentLoaders separately

langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentLoader.java
langchain4j: @kpavlov loaders were left out without a common interface on purpose, as they all have different APIs and it is hard to find a common one without forcing a lot of limitations. Also loaders usually are used in the beginning of the pipeline, so it is not really required for them to be polymorphic.

I am OK with a marker interface solely for discoverability, but I would not force a common API.
BTW, users are not supposed to use `DocumentSource` directly, they are supposed to use `XyzDocumentLoader` which will then deal with `DocumentSource`s internally.

langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentParser.java
langchain4j: Could you please elaborate? What other types do you foresee?

langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentLoader.java
langchain4j: Comment for myself: check https://github.com/langchain4j/langchain4j/issues/1496

kpavlov:
@langchain4j , let's take  #1988

Issue #1715, opened 2024-09-06
[FEATURE] Azure OpenAI: Structured Outputs


langchain4j:
Same as https://github.com/langchain4j/langchain4j/pull/1590 but for Azure OpenAI

langchain4j:
Azure OpenAI SDK feature request: https://github.com/Azure/azure-sdk-for-java/issues/41655

langchain4j:
Structured Outputs are available in [1.0.0-beta.12](https://repo1.maven.org/maven2/com/azure/azure-ai-openai/1.0.0-beta.12/) now!

Hi @jdubois, @agoncal, any chance you could add this to Azure OpenAI before the next release (most probably mid/end next week)? 🙏

langchain4j:
BTW, there is already a reference implementation in `OpenAiChatModel` and a set of tests, just need to extend `AiServicesWithJsonSchemaIT`, you can see `OpenAiAiServicesWithJsonSchemaIT` as an example.

jdubois:
I've started the work on #1982 but there's a lot to do, and I won't be able to work at all on this next week. Happy to have others contribute if that's urgent!

langchain4j:
Implemented in https://github.com/langchain4j/langchain4j/pull/1982

PR #1668, opened 2024-09-02
New low-level API to specify tool parameters

ghost:
## Issue
Closes #896
Fixes #1480
Fixes #1881
Fixes #1858
Fixes #1865

## Context
Schema for tool parameters in the low-level API (`ChatLanguageModel` + `ToolSpecification`) is currently specified using `ToolParameters` and `JsonSchemaProperty` classes. This API works fine for simple cases, but it does not work well for more complicated cases (e.g., when nested objects or collections or recursion is required).

Moreover, [Structured Outputs](https://github.com/langchain4j/langchain4j/pull/1590) feature is using another API and a lot of logic between tools and structured outputs is duplicated and out of sync.

## Change
`ToolParameters` and `JsonSchemaProperty` classes are now deprecated in favour of new `JsonSchemaElement` API (which is used for Structured Outputs as well). Most of the logic between this 2 features is now unified.

`JsonSchemaElement` represents a schema for a JSON element. It can be of such types:
- `JsonArraySchema`
- `JsonBooleanSchema`
- `JsonEnumSchema`
- `JsonIntegerSchema`
- `JsonNumberSchema`
- `JsonObjectSchema`
- `JsonReferenceSchema`
- `JsonStringSchema`

All model providers that support tools (see the list below) were updated to support the new API. Old API is still supported for backward compatibility, but will be removed in the future.

`ToolSpecifications` helper class now generates `ToolSpecification`s using the new API only.

Example of the new API:
```java
ToolSpecification.builder()
    .name("weather")
    .description("Returns the current weather in the specified city")
    .parameters(JsonObjectSchema.builder()
        .addStringProperty("city", "The name of the city, e.g., Munich")
        .addEnumProperty("units", List.of("CELSIUS", "FAHRENHEIT"))
        .required("city") // the required properties should be specified explicitly
        .build())
    .build();
```

`JsonObjectSchema` builder has some helper methods to make it easier to add properties (e.g., `addProperty`, `addStringProperty`, `addIntegerProperty`, etc.), but properties can also be specified using a map:
```java
ToolSpecification.builder()
    .name("weather")
    .description("Returns the current weather in the specified city")
    .parameters(JsonObjectSchema.builder()
        .properties(Map.of(
            "city", JsonStringSchema.builder().description("The name of the city, e.g., Munich").build(),
            "units", JsonEnumSchema.builder().enumValues("CELSIUS", "FAHRENHEIT").build()
        ))
        .required("city") // the required properties should be specified explicitly
        .build())
    .build();
[...]

PR #1982, opened 2024-10-24
Upgrade to Azure OpenAI SDK 1.0.0-beta.13 + Structured Outputs support

jdubois:
As discussed in #1715 this includes migrating to Structured Outputs.

For more information on Structured Outputs:

- The announcement by OpenAI: https://openai.com/index/introducing-structured-outputs-in-the-api/
- The OpenAI documentation: https://platform.openai.com/docs/guides/structured-outputs
- Microsoft Learn documentation: https://learn.microsoft.com/en-us/java/api/overview/azure/ai-openai-readme?view=azure-java-preview#structured-outputs
- Sample code: https://github.com/Azure/azure-sdk-for-java/blob/azure-ai-openai_1.0.0-beta.12/sdk/openai/azure-ai-openai/src/samples/java/com/azure/ai/openai/StructuredOutputsResponseFormat.java


jdubois:
I migrated many tests:

- I'm not sure the AzureOpenAiTokenizer works correctly anymore
- I had to exclude or migrate some models are they are not available anymore in Sweden Central

There is clearly more work needed, @agoncal just told me he would help, but we're both out for about 10 days, so unless someone else is available to help, this will need to wait a bit.

langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonSchemaElementHelper.java
```diff
@@ -185,7 +179,7 @@ public static Map<String, Object> toMap(JsonSchemaElement jsonSchemaElement) {
             if (jsonObjectSchema.description() != null) {
                 properties.put("description", jsonObjectSchema.description());
             }
-            properties.put("properties", toMap(jsonObjectSchema.properties()));
+            properties.put("properties", jsonObjectSchema.properties());
```
langchain4j: Why is this change required? I can already see that behaviour for `JsonReferenceSchema` (recursion case) will be broken.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -13,22 +13,17 @@
 import dev.langchain4j.model.azure.spi.AzureOpenAiChatModelBuilderFactory;
 import dev.langchain4j.model.chat.ChatLanguageModel;
 import dev.langchain4j.model.chat.TokenCountEstimator;
-import dev.langchain4j.model.chat.listener.ChatModelErrorContext;
-import dev.langchain4j.model.chat.listener.ChatModelListener;
-import dev.langchain4j.model.chat.listener.ChatModelRequest;
-import dev.langchain4j.model.chat.listener.ChatModelRequestContext;
-import dev.langchain4j.model.chat.listener.ChatModelResponse;
-import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
+import dev.langchain4j.model.chat.listener.*;
```
langchain4j: nit: please do not use asterisk imports

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -261,7 +226,7 @@ private static BinaryData getParameters(ToolSpecification toolSpecification) {
         if (toolSpecification.parameters() != null) {
             return toOpenAiParameters(toolSpecification.parameters());
         } else {
-            return toOpenAiParametersOld(toolSpecification.toolParameters());
+            return toOpenAiParametersOld(toolSpecification.parameters());
```
langchain4j: This change seems wrong and should not be required. As you can see, method name is toOpenAiParameters**_Old_** and is supposed to convert deprecated `ToolParameters` (that some users might still use in their code).
Handling for new parameters happens 2 lines above. `toOpenAiParametersOld` should stay as-is and convert `ToolParameters`

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -116,8 +108,12 @@ void should_call_function_with_argument(String deploymentName, String gptVersion
 
         ToolSpecification toolSpecification = ToolSpecification.builder()
                 .name(toolName)
-                .description("Get the current weather")
-                .parameters(getToolParameters())
+                .description("Get the current temperature, in celsius or fahrenheit depending on the city location")
```
langchain4j: I wouls keep old API in these tests to make sure we are backwards compatible with the old tool API.
Testing of the new API is happening in `AzureOpenAiAiServicesWithToolsIT`

ghost (commented):
@jdubois thank you!

Just to double check: it seems that this PR is mostly focused on using new `JsonObjectSchema` API to define tool parameters, while it has already been covered in https://github.com/langchain4j/langchain4j/pull/1668 and should not require any work at this point.

[Structured Outputs](https://github.com/langchain4j/langchain4j/pull/1590) is a different feature, you can see how it is implemented for OpenAI in `InternalOpenAiHelper.toOpenAiResponseFormat()`

langchain4j:
Regarding `AzureOpenAiTokenizer`: yes, it is really hard to maintain that logic, perharps we should simplify it (support token estimation only for text) or deprecate completely?

jdubois:
> @jdubois thank you!
> 
> Just to double check: it seems that this PR is mostly focused on using new `JsonObjectSchema` API to define tool parameters, while it has already been covered in #1668 and should not require any work at this point.
> 
> [Structured Outputs](https://github.com/langchain4j/langchain4j/pull/1590) is a different feature, you can see how it is implemented for OpenAI in `InternalOpenAiHelper.toOpenAiResponseFormat()`

Yes I first wanted to do this, I'm doing the structured output right now!! But I missed that you had another draft request for this, so now let me look at this.

jdubois:
I just pushed what I did in https://github.com/langchain4j/langchain4j/pull/1982/commits/fce2bc537d9e13c3a8502894b23b80456cbfc3dd but that's not finished, and I won't have time until November, 4th.

langchain4j:
@jdubois ok, got it, thank you!

langchain4j:
Hi @jdubois, quick update: I will postpone the release till the next week. So I guess there is a chance to get this PR in? Thanks!

jdubois:
With this latest commit I started to add the real code to support structured outputs. It's far from being finished and I can't promise I'll have much time in the next 2 weeks to work on this, but this is definitely moving forward.

jdubois:
For the record, here is the Azure SDK documentation:
- Microsoft Learn documentation: https://learn.microsoft.com/en-us/java/api/overview/azure/ai-openai-readme?view=azure-java-preview#structured-outputs
- Sample code: https://github.com/Azure/azure-sdk-for-java/blob/azure-ai-openai_1.0.0-beta.12/sdk/openai/azure-ai-openai/src/samples/java/com/azure/ai/openai/StructuredOutputsResponseFormat.java

jdubois:
@langchain4j  All integration tests are now green, I need time to look at your code review and to clean up everything, but that should be good soon!!!

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelperTest.java
```diff
@@ -92,7 +92,7 @@ void toToolDefinitionsShouldReturnCorrectToolDefinition() {
         toolSpecifications.add(ToolSpecification.builder()
                 .name("test-tool")
                 .description("test-description")
-                .parameters(ToolParameters.builder().build())
+                .parameters(JsonObjectSchema.builder().build())
```
langchain4j: nit: let's keep `ToolParameters` for a bit, we need to make sure we are backwards compatible. There is another IT that tests tools with the new `JsonObjectSchema` API: `AzureOpenAiAiServicesWithToolsIT`

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -447,4 +472,106 @@ static ChatModelResponse createModelListenerResponse(String responseId,
                 .aiMessage(response.content())
                 .build();
     }
+
+    static ChatCompletionsResponseFormat toAzureOpenAiResponseFormat(ResponseFormat responseFormat, Boolean strict) {
+        if (responseFormat == null || responseFormat.type() == ResponseFormatType.TEXT) {
+            return null;
+        } else if (responseFormat.type() != ResponseFormatType.JSON) {
+            logger.warn("Unsupported response format: " + responseFormat);
```
langchain4j: I would throw an exception to let the user know that this is unsupported

ghost (commented):
@jdubois thanks a lot!

I've noticed an opportunity to simplify JSON mapping, so I did the change and pushed it right away, hope you don't mind. Now everything is supported, only recursion needs a bit of adjustments in `JsonSchemaElementHelper`

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -100,10 +108,11 @@ public AzureOpenAiChatModel(OpenAIClient client,
                                 List<AzureChatExtensionConfiguration> dataSources,
                                 AzureChatEnhancementConfiguration enhancements,
                                 Long seed,
-                                ChatCompletionsResponseFormat responseFormat,
```
langchain4j, line 103: This is a breaking change, some people have `.responseFormat(new ChatCompletionsJsonResponseFormat())` in their code. Can we keep supporting `ChatCompletionsResponseFormat`? We can make it deprecated in the builder though.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -100,10 +108,11 @@ public AzureOpenAiChatModel(OpenAIClient client,
...
+                                ResponseFormat responseFormat,
```
langchain4j, line 116: I am wondering if we really need to set format in the constructor, as in the new `chat()` API it is possible to set it per request.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -100,10 +108,11 @@ public AzureOpenAiChatModel(OpenAIClient client,
...
+                                boolean strictJsonSchema,
```
langchain4j: We need to not forget to implement `strictTools` as well, it should be pretty straightforward.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -124,7 +133,8 @@ public AzureOpenAiChatModel(String endpoint,
                                 List<AzureChatExtensionConfiguration> dataSources,
                                 AzureChatEnhancementConfiguration enhancements,
                                 Long seed,
-                                ChatCompletionsResponseFormat responseFormat,
```
langchain4j, line 127: same comments as above for this and other ctors

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -228,28 +241,54 @@ private AzureOpenAiChatModel(String deploymentName,
...
-        this.responseFormat = responseFormat;
+        this.chatCompletionsResponseFormat = toAzureOpenAiResponseFormat(responseFormat, strictJsonSchema);
+        this.strictJsonSchema = getOrDefault(strictJsonSchema, false);
         this.listeners = listeners == null ? emptyList() : new ArrayList<>(listeners);
     }
 
+    @Override
+    public Set<Capability> supportedCapabilities() {
+        Set<Capability> capabilities = new HashSet<>();
+        if (chatCompletionsResponseFormat != null && chatCompletionsResponseFormat instanceof ChatCompletionsJsonResponseFormat) {
```
langchain4j: Shouldn't it be `ChatCompletionsJsonSchemaResponseFormat`?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -518,11 +558,16 @@ public Builder seed(Long seed) {
             return this;
         }
 
-        public Builder responseFormat(ChatCompletionsResponseFormat responseFormat) {
+        public Builder responseFormat(ResponseFormat responseFormat) {
```
langchain4j, line 602: breaking change

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiAiServicesWithJsonSchemaIT.java
```diff
@@ -0,0 +1,41 @@
...
+class AzureOpenAiAiServicesWithJsonSchemaIT extends AiServicesWithJsonSchemaIT {
+
+    @Override
+    protected List<ChatLanguageModel> models() {
+        return List.of(
+                AzureOpenAiChatModel.builder()
+                        .endpoint(System.getenv("AZURE_OPENAI_ENDPOINT"))
+                        .apiKey(System.getenv("AZURE_OPENAI_KEY"))
+                        .deploymentName("gpt-4o-mini")
+                        .responseFormat(JSON)
```
langchain4j: `json_object` and `json_schema` are different features and we should distinguish between them as not all models support `json_schema`. Because of this, user has to opt-in in order to use `json_schema` and `strict`.
Current users already do `.responseFormat(new ChatCompletionsJsonResponseFormat())` to opt-in into `json_object`.
We should have a similar alternative for `json_schema`.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModelIT.java
```diff
@@ -182,8 +182,11 @@ void should_call_function_with_argument(String deploymentName, String gptVersion
         ToolSpecification toolSpecification = ToolSpecification.builder()
                 .name(toolName)
                 .description("returns a sum of two numbers")
-                .addParameter("first", INTEGER)
-                .addParameter("second", INTEGER)
+                .parameters(JsonObjectSchema.builder()
```
langchain4j: nit: let's keep all tools with the old API (explained the reason in other comment)

jdubois:
Thanks @langchain4j !
I do have all the tests working, and here's a screenshot  to prove it, but the `should_return_result_with_pojo()` test is unstable: despite sending it the right JSON request (I log them, to be sure we send the schema and the "strict" boolean), it sometimes returns text! I believe that's an issue on the Azure OpenAI side, so I'll ping them if this continues.

Screenshot 2024-11-20 at 17 44 23

langchain4j:
BTW there is a new JSON type (`JsonAnyOfSchema`) being added in https://github.com/langchain4j/langchain4j/pull/2148, I guess Azure OpenAI supports it as well, so we should not forget to add a mapping for it in Azure OpenAI

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelperTest.java
jdubois: rolled back

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
jdubois: Done

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: As we also add the `boolean strictJsonSchema` argument just after, there will be a breaking change in the constructor anyway: are you fine if we only do this in the Builder? It would accept `ChatCompletionsResponseFormat` (with a deprecation notice) and the `ResponseFormat`.

jdubois: Removing my comment above, it wasn't good

jdubois: I reverted to the former API

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: I reverted to the former API

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: same, reverted to the former API

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: That was a bug, it's fixed.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: We now support both

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiAiServicesWithJsonSchemaIT.java
jdubois: If you select JSON, you can have "strict" which means you use the schema, or not "strict" that makes you use the `json_object`

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModelIT.java
jdubois: Change has been rolled back

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: I'm now doing a mix:
- You can set it up in the constructor, and it's the only way to set the "strict" boolean as otherwise it's not in the current `ChatRequest` object yet. I believe it's consistent with the other parameters, like `temperature`.
- Then the response format can be overridden in every request. I believe it's better as you can reuse your configuration for different requests (as a Spring user, I wouldn't like to have multiple instances of the same Spring bean here).

This will need to be refactored when `ChatRequest` supports the "strict" boolean, but that's shared with OpenAI and Gemini, so it should be in a different PR.

jdubois:
@langchain4j I put this as "ready for review", all my tests are passing and the code review should be OK.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
langchain4j: Just in case: my comment was about `strictTools`, not `strictJsonSchema`, it is slightly different feature. Should we create a separate PR for `strictTools`?

jdubois: Oh sorry, I thought you made a typo! I haven't had a look at `strictTools`

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -518,11 +553,25 @@ public Builder seed(Long seed) {
             return this;
         }
 
-        public Builder responseFormat(ChatCompletionsResponseFormat responseFormat) {
+        /**
+         * @deprecated For JSON output, you can replace `.responseFormat(new ChatCompletionsJsonResponseFormat())` by `.responseFormat(ResponseFormat.JSON)`, and use `.strictJsonSchema(true)`to force JSON schema validation.
```
langchain4j:
```suggestion
         * @deprecated For JSON output, you can replace `.responseFormat(new ChatCompletionsJsonResponseFormat())` by `.responseFormat(ResponseFormat.JSON)`, and use `.strictJsonSchema(true)`to force JSON schema adherence.
```
nit: "adherence" seems to be more appropriate here

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -518,11 +553,25 @@ public Builder seed(Long seed) {
...
+         */
+        @Deprecated(forRemoval = true)
```
langchain4j, line 596: Could you please update Azure OpenAI example in [this](https://docs.langchain4j.dev/tutorials/ai-services#json-mode) documentation section to use the new way?

It would also be great to have specific sections on https://docs.langchain4j.dev/integrations/language-models/azure-open-ai page, similarly to https://docs.langchain4j.dev/integrations/language-models/open-ai#structured-outputs

langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonSchemaElementHelper.java
```diff
@@ -150,7 +150,8 @@ private static String descriptionFrom(Description description) {
     }
 
     private static Class<?> getActualType(Type type) {
-        if (type instanceof ParameterizedType parameterizedType) {
+        if (type instanceof ParameterizedType) {
+            ParameterizedType parameterizedType = (ParameterizedType) type;
```
langchain4j: Just in case: we are on a Java 17 baseline now :)

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -278,7 +279,7 @@ void should_use_json_format(String deploymentName, String gptVersion) {
                 .apiKey(System.getenv("AZURE_OPENAI_KEY"))
                 .deploymentName(deploymentName)
                 .tokenizer(new AzureOpenAiTokenizer(gptVersion))
-                .responseFormat(new ChatCompletionsJsonResponseFormat())
```
langchain4j: I've tested `.responseFormat(new ChatCompletionsJsonResponseFormat())` and it does not work, it used to include `"response_format":{"type":"json_object"}` in the request, but not any more. This is a breaking change

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
```diff
@@ -278,7 +279,7 @@ void should_use_json_format(String deploymentName, String gptVersion) {
...
+                .responseFormat(ResponseFormat.JSON)
```
langchain4j, line 296: `ResponseFormat.JSON` does not work either, there is no `"response_format":{"type":"json_object"}` in the request.

langchain4j: We need to improve this test to catch such things. I guess `GitHubModelsStreamingChatModelIT.should_use_json_format()` is more reliable. Returned content in this test is wrapped into ````json`, `GitHubModelsStreamingChatModelIT`'s way would catch it.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiAiServicesWithJsonSchemaIT.java
langchain4j: AFAIU `strict` does not mean "there is a schema", it means "adhere to the schema".
There are the following options/combinations:
1. no response format -> text response
2. `ChatCompletionsTextResponseFormat` -> text response
3. `ChatCompletionsJsonResponseFormat` -> JSON response without schema, requires describing schema in a free-form text in the prompt
4. `ChatCompletionsJsonSchemaResponseFormat` + `strict = true` -> JSON response with schema, guarantees schema adherance. This will require processing the schema (on the first request only) which takes around 10 seconds.
5. `ChatCompletionsJsonSchemaResponseFormat` + `strict = false` -> JSON response with schema, does not guarantee schema adherance. This does not require processing the schema.

So there is currently no way to distinguish between 3 and 5.
3 is used by many users for sure, as it is currently the only way to ensure json output. So we should be backwads compatible and still support it.
5 could be a "corner case" when schema is generated dynamically and/or provided by the end users and 10 second latency is not acceptable. It would be great to support this use case as well.

Would it be possible to use Azure's `ChatCompletionsResponseFormat` type instead of LC4j's `ResponseFormatType`?

Examples of how users would configure it:
3.
```java
AzureOpenAiChatModel.builder()
    .deploymentName("gpt-4o-mini")
    .responseFormat(new ChatCompletionsJsonResponseFormat())
    .build()
```

4.
```java
AzureOpenAiChatModel.builder()
    .deploymentName("gpt-4o-mini")
    .responseFormat(new ChatCompletionsJsonSchemaResponseFormat(null)) // not sure about null
    .strictJsonSchema(true)
    .build()
```

5.
```java
AzureOpenAiChatModel.builder()
    .deploymentName("gpt-4o-mini")
    .responseFormat(new ChatCompletionsJsonSchemaResponseFormat(null)) // not sure about null
    .strictJsonSchema(false) // or not set at all, false by default
    .build()
```

Another option would be to use OpenAI's strings `json_object` and `json_schema` in ctor/builder.

A bit of a background info: for vanilla OpenAI there is no way to set the actual schema (complete structure with fields) in the constructor/builder, only to "enable `json_schema` support", as currently not all models support this feature, only latest `gpt-4o/-mini` ones. When `ChatLanguageModel` is used with AI Service and return type is a POJO, we want to use Structured Outputs feature (with json schema), but only if `RESPONSE_FORMAT_JSON_SCHEMA` capability is enabled explicitly. If this capability is not enabled, AI Service will fall back to generating pseudo schema and append it to the user message. If we'll enable `RESPONSE_FORMAT_JSON_SCHEMA` for `OpenAiChatModel`/`AzureOpenAiChatModel` by default, this will lead to errors if users use older models. So right now to use Structured Outputs feature, users will need to _**opt in**_. I guess we will need to rethink and change this soon (no later than right before 1.0), as more and more users will migrate to newer models. I can imagine in 1 year older models will be decomissioned and newer will all support this feature, so there will be no need to require users to opt in, we could just enbale it by default. If user does not want to use it, we could add an option to opt out.

WDYT?

ghost (commented):
@jdubois thank you!

langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonSchemaElementHelper.java
jdubois: I guess you mean doing this, so I'm committing it:
```
        if (type instanceof final ParameterizedType parameterizedType) {
            Type[] actualTypeArguments = parameterizedType.getActualTypeArguments();
```

jdubois:
I really dislike the `new ChatCompletionsJsonSchemaResponseFormat(null)` configuration:

- Putting a `null` as an argument seems wrong
- Also we would mix LangChain4j's API with Azure OpenAI's API

Then, you are right, use case 5 isn't working here, even if that's a weird one we should support it.

Do you think we could update `dev.langchain4j.model.chat.request.ResponseFormat`, to support `TEXT`, `JSON` and `STRUCTURED_JSON`?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: Yes, but I'll do it once we have agreed on the API changes, so probably once this PR is merged.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -228,28 +235,54 @@ private AzureOpenAiChatModel(String deploymentName,
...
         this.seed = seed;
-        this.responseFormat = responseFormat;
+        this.chatCompletionsResponseFormat = chatCompletionsResponseFormat;
         this.listeners = listeners == null ? emptyList() : new ArrayList<>(listeners);
     }
 
+    @Override
+    public Set<Capability> supportedCapabilities() {
+        Set<Capability> capabilities = new HashSet<>();
+        if (chatCompletionsResponseFormat != null && chatCompletionsResponseFormat instanceof ChatCompletionsJsonSchemaResponseFormat) {
```
langchain4j: I was thinking about this logic lately and I think that it would be clearer if user enables it like this:

```java
XyzChatModel().builder()
    ...
    .supportedCapabilities(List.of(RESPONSE_FORMAT_JSON_SCHEMA, ...))
    .build();
```

On the other hand, when user sets the response format (example below), it sohuld be applied to all the requests to the LLM, but it should not drive the logic of the `supportedCapabilities()` method.
```java
XyzChatModel().builder()
    ...
    .responseFormat(...)
    .build();
```

WDYT?

langchain4j:
> I really dislike the `new ChatCompletionsJsonSchemaResponseFormat(null)` configuration:
> 
> * Putting a `null` as an argument seems wrong
> * Also we would mix LangChain4j's API with Azure OpenAI's API
> 
> Then, you are right, use case 5 isn't working here, even if that's a weird one we should support it.
> 
> Do you think we could update `dev.langchain4j.model.chat.request.ResponseFormat`, to support `TEXT`, `JSON` and `STRUCTURED_JSON`?

I guess if we change the logic of `supportedCapabilities()` like I've described in https://github.com/langchain4j/langchain4j/pull/1982#discussion_r1878523329, there will be no need in `new ChatCompletionsJsonSchemaResponseFormat(null)`

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
jdubois: Indeed, I just fixed it in my last commit, and added some unit test (not perfect) for the old method

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: I get it!
As I put in https://github.com/langchain4j/langchain4j/pull/1982#issuecomment-2532127682 I preferred to have 3 options in the `ResponseFormat`,  but your solution also works and has the advantage of not breaking the existing OpenAI and Gemini code.
I've just pushed a commit that supports your solution, can you have a look? Looking at `AzureOpenAiAiServicesWithJsonSchemaIT`, I believe it's OK for end users.

jdubois:
@langchain4j yes with my last commit this should be OK

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -228,28 +240,51 @@ private AzureOpenAiChatModel(String deploymentName,
         this.dataSources = dataSources;
         this.enhancements = enhancements;
         this.seed = seed;
-        this.responseFormat = responseFormat;
+        this.chatCompletionsResponseFormat = chatCompletionsResponseFormat;
         this.listeners = listeners == null ? emptyList() : new ArrayList<>(listeners);
+        this.supportedCapabilities = capabilities;
```
langchain4j:
```suggestion
        this.supportedCapabilities = copyIfNotNull(capabilities);
```

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -569,7 +621,18 @@ public Builder customHeaders(Map<String, String> customHeaders) {
...
+            this.capabilities = capabilities;
+            return this;
+        }
+
         public AzureOpenAiChatModel build() {
+            if (this.capabilities == null) {
+                capabilities = new HashSet<>();
+            }
+            if (this.responseFormat != null) {
+                this.chatCompletionsResponseFormat = toAzureOpenAiResponseFormat(this.responseFormat, this.strictJsonSchema);
```
langchain4j: I guess the same logic should be applied to the streaming model, it can be added in a separate PR though

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiAiServicesWithJsonSchemaIT.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Override
+    protected List<ChatLanguageModel> models() {
+        return List.of(
+                AzureOpenAiChatModel.builder()
+                        .endpoint(System.getenv("AZURE_OPENAI_ENDPOINT"))
+                        .apiKey(System.getenv("AZURE_OPENAI_KEY"))
+                        .deploymentName("gpt-4o-mini")
+                        .supportedCapabilities(Set.of(RESPONSE_FORMAT_JSON_SCHEMA))
+                        .responseFormat(JSON)
```
langchain4j: Hmm, ` .responseFormat(JSON)` seems redundant here.

` .supportedCapabilities(Set.of(RESPONSE_FORMAT_JSON_SCHEMA))` means "this model _can_ consume `JsonSchema` in the `ResponseFormat`. If `ChatRequest` will contain`ResponseFormat`/`JsonSchema`, it will work."

` .responseFormat(JSON)` means "always output JSON (without schema)"

".strictJsonSchema(true)" means "be strict with JSON schema"

These should work (currently they fails with NPEs):
```java
@Override
    protected List<ChatLanguageModel> models() {
        return List.of(
                AzureOpenAiChatModel.builder()
                        .deploymentName("gpt-4o-mini")
                        .supportedCapabilities(Set.of(RESPONSE_FORMAT_JSON_SCHEMA))
                        .strictJsonSchema(true)
                        .temperature(0.0)
                        .build(),
                AzureOpenAiChatModel.builder()
                        .deploymentName("gpt-4o-mini")
                        .supportedCapabilities(Set.of(RESPONSE_FORMAT_JSON_SCHEMA))
                        .strictJsonSchema(false)
                        .temperature(0.0)
                        .build()
        );
    }
```

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -447,4 +455,36 @@ static ChatModelResponse createModelListenerResponse(String responseId,
...
+            throw new IllegalArgumentException("Unsupported response format: " + responseFormat);
+        }
+        if (strict == null) {
+            strict = false;
+        }
+
+        JsonSchema jsonSchema = responseFormat.jsonSchema();
+        if (jsonSchema == null) {
+            if (strict) {
+                // Schema isn't provided when constructing the object, but it can be set later when doing the requests.
```
langchain4j: Perhaps this mapping logic should be delayed till doing the actual request? At that point of time all information will be available.

`AzureOpenAiChatModel` could have `ResponseFormat responseFormat` and `Boolean strict` fields (instead of `ChatCompletionsResponseFormat chatCompletionsResponseFormat`) which are then mapped to `ChatCompletionsResponseFormat` in `chat(ChatRequest)` method.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
langchain4j: Just in case, `supportedCapabilities(List.of(RESPONSE_FORMAT_JSON_SCHEMA, ...))` is required only when using AI Services (because AI Services logic sohuld know in what format to generate schema: append to the prompt or generate `JsonSchema` for models that support this feature). It is not required to be set when using `AzureOpenAiChatModel` directly.

agoncal:
@jdubois wondering if you need any help or review on this one ?

jdubois:
@agoncal I was going to work on this today, but as we have the day tomorrow together at the office, let's pair program tomorrow!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: Yes let's do non-steaming first, once this is good I'll copy the logic for streaming

jdubois: I added this as a note in https://github.com/langchain4j/langchain4j/issues/2241 so we don't forget

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiAiServicesWithJsonSchemaIT.java
jdubois: I've corrected this and modified the test to be like the one you coded above

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
jdubois: Indeed, I have moved this

jdubois:
@langchain4j I believe this now works the way you wanted

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -228,29 +259,69 @@ private AzureOpenAiChatModel(String deploymentName,
         this.dataSources = dataSources;
         this.enhancements = enhancements;
         this.seed = seed;
+        this.chatCompletionsResponseFormat = chatCompletionsResponseFormat;
         this.responseFormat = responseFormat;
```
langchain4j, line 263: minor: perhaps we should validate that either `responseFormat` or `chatCompletionsResponseFormat` are set, but not both at the same time?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -228,29 +259,69 @@ private AzureOpenAiChatModel(String deploymentName,
...
     ) {
+        ChatCompletionsResponseFormat chatCompletionsResponseFormat = null;
+        if (this.chatCompletionsResponseFormat != null) {
+            chatCompletionsResponseFormat = this.chatCompletionsResponseFormat;
+        } else {
+            chatCompletionsResponseFormat = toAzureOpenAiResponseFormat(responseFormat, this.strictJsonSchema);
+        }
+        if (chatCompletionsResponseFormat instanceof ChatCompletionsJsonSchemaResponseFormat) {
+            if (!supportedCapabilities.contains(Capability.RESPONSE_FORMAT_JSON_SCHEMA)) {
+                throw new IllegalArgumentException("JSON schema response format is not supported by this model");
```
langchain4j: I think this validation can be removed, `Capability` is needed only to communicate supported capabilities between AI Service and `ChatLanguageModel`. If one uses `ChatLanguageModel` directly, there is no need in setting supported capabilities (and validating them), user is responsible to select a proper model if they specify JSON schema

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -518,11 +592,25 @@ public Builder seed(Long seed) {
             return this;
         }
 
-        public Builder responseFormat(ChatCompletionsResponseFormat responseFormat) {
+        /**
+         * @deprecated For JSON output, you can replace `.responseFormat(new ChatCompletionsJsonResponseFormat())` by `.responseFormat(ResponseFormat.JSON)`, and use `.strictJsonSchema(true)`to force JSON schema adherence.
```
langchain4j: `strictJsonSchema(true)` will take effect only if `ResponseFormat` contains `JsonSchema`, it does not affect `responseFormat(ResponseFormat.JSON)` case

langchain4j:
@jdubois thanks you so much!

19 tests are failing for me when run locally:
image

Does it work fine for you?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
```diff
@@ -228,29 +259,69 @@ private AzureOpenAiChatModel(String deploymentName,
...
 
     @Override
     public Response<AiMessage> generate(List<ChatMessage> messages, ToolSpecification toolSpecification) {
-        return generate(messages, singletonList(toolSpecification), toolSpecification);
+        return generate(messages, singletonList(toolSpecification), toolSpecification, this.responseFormat);
+    }
+
+    @Override
+    public ChatResponse chat(ChatRequest request) {
+        // If the response format is not specified in the request, use the one specified in the model
```
langchain4j, line 294: nice!

langchain4j:
@jdubois no worries, I will push a tiny fix

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiChatModelIT.java
langchain4j: Thank you!

langchain4j-parent/pom.xml
```diff
@@ -56,7 +56,7 @@
         <argLine></argLine>
         <!-- Dependency Versions -->
         <openai4j.version>0.24.0</openai4j.version>
-        <azure-ai-openai.version>1.0.0-beta.11</azure-ai-openai.version>
+        <azure-ai-openai.version>1.0.0-beta.12</azure-ai-openai.version>
```
langchain4j: should we update to `1.0.0-beta.13`?

ghost (commented):
@jdubois thanks a lot!

langchain4j:
@jdubois I think it is good to merge. There are a few minor comments left, but we can also merge it as it is now (if you will not have time to address them tomorrow). Please LMK how you want to proceed

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
langchain4j: I guess this is quite major, I wiould remove it as it will force users to enable `RESPONSE_FORMAT_JSON_SCHEMA` even when using the low-level API

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: I'm adding a `throw new IllegalArgumentException("You can't set both chatCompletionsResponseFormat and responseFormat");`

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: Done

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
jdubois: I rephrased it

langchain4j-parent/pom.xml
jdubois: I'm trying this right now. If we migrate here, there's at least one `setStream` method that disappears (so there's a small impact in the code, see https://github.com/Azure/azure-sdk-for-java/releases/tag/azure-ai-openai_1.0.0-beta.13 ), I'll close https://github.com/langchain4j/langchain4j/issues/2241 and add a new ticket for doing structured outputs with streaming.

jdubois: Done as tests are passing

ghost (approved):
@jdubois thanks a lot!

langchain4j:
@jdubois could you please update the documentation and Spring Boot starter?

jdubois:
Awesome! Yes I'll do this today.

Issue #2053, opened 2024-11-09
[BUG] issue with TextDocumentParser


omarmahamid:
**Describe the bug**

1.

Managing ByteArrayOutputStream is not occurred correctly
The ByteArrayOutputStream should either be closed explicitly or be used in a try-with-resources statement. Using try-with-resources ensures that the stream is properly closed, even if an exception is thrown during processing.

https://github.com/langchain4j/langchain4j/blob/main/langchain4j/src/main/java/dev/langchain4j/data/document/parser/TextDocumentParser.java#L31

2.
As we can see here
https://stackoverflow.com/questions/68708640/conversion-of-bytearrayinputstream-tostring-results-in-warning-inefficient-conve

how the TextDocumentParser convert to String from buffer is in this way

https://github.com/langchain4j/langchain4j/blob/main/langchain4j/src/main/java/dev/langchain4j/data/document/parser/TextDocumentParser.java#L39

which is not efficient conversion

PR #2054, opened 2024-11-09
issue with TextDocumentParser PR

omarmahamid:
## Issue

Closes #https://github.com/langchain4j/langchain4j/issues/2053

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/data/document/parser/TextDocumentParser.java
+++ b/langchain4j/src/main/java/dev/langchain4j/data/document/parser/TextDocumentParser.java
@@ -27,8 +27,7 @@ public class TextDocumentParser implements DocumentParser {
     @Override
     public Document parse(InputStream inputStream) {
         // see https://stackoverflow.com/a/35446009
-        try {
-            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+        try(ByteArrayOutputStream buffer = new ByteArrayOutputStream()){
             int nRead;
             byte[] data = new byte[1024];
             while ((nRead = inputStream.read(data, 0, data.length)) != -1) {
@@ -36,7 +35,7 @@ public class TextDocumentParser implements DocumentParser {
             }
             buffer.flush();
 
-            String text = new String(buffer.toByteArray(), charset);
+            String text = buffer.toString(charset);
 
             if (isNullOrBlank(text)) {
                 throw new BlankDocumentException();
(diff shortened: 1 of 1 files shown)


langchain4j/src/main/java/dev/langchain4j/data/document/parser/TextDocumentParser.java
```diff
@@ -27,16 +27,15 @@ public TextDocumentParser(Charset charset) {
     @Override
     public Document parse(InputStream inputStream) {
         // see https://stackoverflow.com/a/35446009
-        try {
-            ByteArrayOutputStream buffer = new ByteArrayOutputStream();
+        try(ByteArrayOutputStream buffer = new ByteArrayOutputStream()){
```
langchain4j: `ByteArrayOutputStream.close()` is empty, it makes no sense to close it.
`inputStream` is closed in `DocumentLoader.load()`

langchain4j/src/main/java/dev/langchain4j/data/document/parser/TextDocumentParser.java
```diff
@@ -27,16 +27,15 @@ public TextDocumentParser(Charset charset) {
...
             int nRead;
             byte[] data = new byte[1024];
             while ((nRead = inputStream.read(data, 0, data.length)) != -1) {
                 buffer.write(data, 0, nRead);
             }
             buffer.flush();
 
-            String text = new String(buffer.toByteArray(), charset);
```
langchain4j, line 39: Since we are on Java 17 now, how about this instead?

```java
    @Override
    public Document parse(InputStream inputStream) {
        try {
            String text = new String(inputStream.readAllBytes(), charset);
            if (text.isBlank()) {
                throw new BlankDocumentException();
            }
            return Document.from(text);
        } catch (BlankDocumentException e) {
            throw e;
        } catch (Exception e) {
        }
    }
```

ghost (commented):
@omarmahamid thank you!

langchain4j/src/main/java/dev/langchain4j/data/document/parser/TextDocumentParser.java
omarmahamid: you are right, so it's a minor change, but still since autoclosable, it's best practice to sorround it with try with resources.

langchain4j/src/main/java/dev/langchain4j/data/document/parser/TextDocumentParser.java
omarmahamid: I agree, 
generally more performant

ghost (approved):
@omarmahamid thank you!

Issue #1276, opened 2024-06-13
[FEATURE] Consistent exception handling in models


langchain4j:
TBD

Both sync and async models should have consistent behavior in terms of errors/exceptions.

Also review handling of situations where content is filtered out by the moderation (Azure/Gemini)

PR #2056, opened 2024-11-10
migrate vespa to jackson + tests and refactorings

### LINKED ISSUE #1681: Vespa: migrate from Gson to Jackson
Migrate `langchain4j-vespa` module from Gson to Jackson.
Apart from `converter-gson`, `VespaEmbeddingStore` uses `Json.toJson()` and `Json.toInputStream()` methods that uses Gson under the hood.

Heezer:
## Issue

Closes #1681

## Change

1. Raise the baseline from jdk8 to jdk17 (because the newest vespa-feed-client is no longer support jdk8).
2. Migrate from Gson to Jackson.
3. Get rid of Lombok
4. Implement VespaEmbeddingStoreIT
5. <s>VespaEmbeddingStoreCloudIT</s> will be done in scope of other ticket
6. Support new API search()
7. Support logRequests & logResponses
8. Make parameters keyPath & certPath optional
9. Support removeAll()

## General checklist

- [x] <s>I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)</s> will be done in scope of other ticket
- [x] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features) will be updated in scope of other ticket


langchain4j-vespa/src/main/java/dev/langchain4j/store/embedding/vespa/DeleteResponse.java
```diff
@@ -0,0 +1,30 @@
+package dev.langchain4j.store.embedding.vespa;
+
+class DeleteResponse {
```
langchain4j: please add `@JsonIgnoreProperties(ignoreUnknown = true)`

langchain4j: nit: could be a record?

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Override
+    protected String documentURI(String id) {
+        return String.format("id:%s:%s::%s", DEFAULT_NAMESPACE, DEFAULT_DOCUMENT_TYPE, id);
+    }
+
+    @Override
+    protected Percentage percentage() {
+        return withPercentage(12);
```
langchain4j: 12 seems a bit too much, why is it required?

langchain4j-vespa/src/main/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStore.java
```diff
@@ -37,246 +42,409 @@
...
-              } else if (Result.Type.success.equals(result.type())) {
-                resId.set(result.documentId().toString());
-              }
+
+    @Override
+    public void removeAll() {
+        try {
+            Response<DeleteResponse> response = api().deleteAll(namespace, documentType, clusterName).execute();
+            if (!response.isSuccessful()) {
+                throw new RuntimeException("Delete failed");
```
langchain4j: would be nice to provide code and/ or error body in the exception

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Override
+    protected List<String> addAll(List<Embedding> embeddings, List<TextSegment> embedded) {
+        // we have to add records one by one, since Vespa's Json feeder in addAll method returns the ids in random order
```
langchain4j: users are also affected? should we document this?

ghost (commented):
@Heezer thanks a lot!

langchain4j-vespa/pom.xml
```diff
@@ -23,7 +23,9 @@
     </description>
 
     <properties>
-        <vespa.version>8.190.2</vespa.version> <!-- the latest Java 8 version -->
+        <vespa.version>8.432.4</vespa.version>
+        <maven.compiler.source>17</maven.compiler.source>
```
Martin7-1: Since langchain4j now uses jdk17 as baseline, this is not needed anymore.

langchain4j-vespa/pom.xml
```diff
@@ -101,6 +107,47 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-embeddings-all-minilm-l6-v2</artifactId>
```
Martin7-1: Why this needed?

Martin7-1 (commented):
Thank you!

langchain4j-vespa/src/main/java/dev/langchain4j/store/embedding/vespa/DeleteResponse.java
Heezer: just converted all DTOs to records, works well!

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
Heezer: Yes, here I need a support if you think 12 is too much 😄 because less than 12 leads to assert failures. Since Vespa is highly configurable, could I may be adjust here anything to reduce percentage setting?
```
  rank-profile cosine_similarity {
    inputs {
      query(q) tensor<float>(x[384])
    }
    function cosine() {
      expression: 1.0 - distance(field, vector)
    }
    first-phase {
      expression: if (cosine() < query(threshold), -1, cosine())
      rank-score-drop-limit: 0.0
    }
  }
```

langchain4j-vespa/pom.xml
Heezer: done

langchain4j-vespa/pom.xml
Heezer: I suppose - garbage, just removed

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
Heezer: Hmmm, the test relies on the proper sequence of IDs, I suppose for the users it's not relevant - but, yes, it makes sense to mention that point in the docs, looking into this...

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
langchain4j: I need to dig deeper into this, but from the surface 12% sounds very fishy, it might even hide a real problem when wrong embedding is matched...

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
langchain4j: You mean that IDs returned in the `List<String>` from `addAll()` methods do not correspond (by order) to the embeddings that were added? This might be a big problem for some users that want to use these IDs to manage their embeddings later on...

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
Heezer: yes, if you think, it's fishy, any suggestions are welcome 🙏

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
Heezer: yes, the order of response IDs is random 😞 going to check the config options and possibly open an issue on Vespa side

Heezer:
@langchain4j  there are lots of builds failing, I suppose my changes have nothing to do with that 😄

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
Heezer: simple workaround would be to use in `addAll()` generally `JsonFeeder.feedSingle()` instead of `.feedMany()` with some loose of performance...

langchain4j-vespa/src/main/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStore.java
langchain4j: This method from `SearchApiClient` can be reused:

```java
private static RuntimeException toException(Response<?> response) throws IOException {
        try (ResponseBody responseBody = response.errorBody()) {
            int code = response.code();
            if (responseBody != null) {
                String body = responseBody.string();
                String errorMessage = String.format("status code: %s; body: %s", code, body);
                return new RuntimeException(errorMessage);
            } else {
                return new RuntimeException(String.format("status code: %s;", code));
            }
        }
    }
```

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
langchain4j: How big the performance impact would be? This only affects parsing, right? Or would it mean that there will be a network request for every embedding?

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
langchain4j: I'll try to look into it during this week!

Heezer: Look at schema/langchain4j.sd schema within vespa_app.zip - the Vespa config to play with is this one. I've just tried various variants but unfortunately without luck - the better difference I got - 7%. I suppose to be able to properly fix it I need to have good Vector DB knowledge and this is what I don't have currently :(

langchain4j: Sorry, I did not have time to look into this yet :(

But the problem could be that Vespa returns cosine similarity (in the range of -1 to 1), while `EmbeddingMatch.score` should be a relevance score (in the range of 0 to 1). Cosine similarity can be converted into the relevance score using the `RelevanceScore.fromCosineSimilarity()` helper. There is a `CosineSimilarity.fromRelevanceScore()` as well BTW.

Heezer: good news to the end of the year 😄 got it working, this is the main change:
```
function relevance() {
  expression: ( cos(distance(field, vector)) + 1 ) / 2
}
```
in fact, previous config Vespa was returning the cosine similarity, now it returns the relevance score, exactly what the `EmbeddingStoreWithoutMetadataIT` expects.

langchain4j-vespa/src/test/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStoreIT.java
Heezer: Here good news as well - with #2106 the original IDs remain the same, they are not coming from the service any longer. I've refactored `EmbeddingStoreWithoutMetadataIT` back.

langchain4j-vespa/src/main/java/dev/langchain4j/store/embedding/vespa/VespaEmbeddingStore.java
Heezer: Good, I've just copied this code one to one und used it in Vespa. At the moment three modules having the method `RuntimeException toException(Response<?> response)` with same or similar code. From my side it makes sense to centralize this implementation, I suppose it could be done within #1276.

Heezer:
@langchain4j I would like to take point 5 - updating of `VespaEmbeddingStoreCloudIT` and documentation into separate issues, otherwise it's way too much 😄 what is actually the case already anyway 😄 fine? removing draft label then...

Heezer:
> @langchain4j there are lots of builds failing, I suppose my changes have nothing to do with that 😄

@langchain4j builds 17 and 21 fail with timeouts in azure-open-ai 🤷‍♂️

langchain4j:
Hi @Heezer, happy New Year! 🥳 

> @langchain4j builds 17 and 21 fail with timeouts in azure-open-ai 🤷‍♂️

No worries, please ignore tests not related to your changes

Heezer:
> Hi @Heezer, happy New Year! 🥳 

Thx and happy New Year back ☃️

Heezer:
@langchain4j, @Martin7-1  guys, can you review pls? 🙏

dliubarskyi (approved):
@Heezer thanks a lot, great job!

PR #2130, opened 2024-11-16
Remove Java 17 from nightly workflow

kpavlov:
I feel, that nightnly build fails due to lack of resources...

## Issue
Nightly build fails when it runs matrix build for 2 JDKs. 
Trying to address it by removing Java 17 from the matrix configuration in the nightly workflow YAML file and running only on current LTS java version.

## Change
Remove Java 17 from nightly workflow

Java 17 has been definitively removed from the matrix configuration in the nightly workflow YAML file. This change restricts the build and testing process to only Java 21, aligning with our current Java version requirements.

## General checklist

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -12,11 +12,8 @@ jobs:
       fail-fast: false # do not cancel jobs on error
       matrix:
         java_version:
-          - 17
           - 21
         include:
-          - java_version: '17'
-            mvn_opts: '-pl !langchain4j-ollama,!langchain4j-local-ai,!langchain4j-milvus'
           - java_version: '21'
             mvn_opts: '-pl !langchain4j-ollama,!langchain4j-local-ai,!langchain4j-milvus'
       max-parallel: 1


langchain4j:
Hi @kpavlov, thank you!

Since we target Java 17, it would be great to keep it. What about splitting the nightly build into 2 separate workflows?

kpavlov:
If we run on 17, then tests for jlama will not run, since jlama [requires 21+](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-jlama/pom.xml#L21)

langchain4j:
> If we run on 17, then tests for jlama will not run, since jlama [requires 21+](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-jlama/pom.xml#L21)

I meant that we can run both 17 and 21, but in 2 separate workflows, to avoid resource exhaustion

kpavlov:
Then we should not run them in parallel to avoid hitting rate limits. Maybe we can run single workflow in a different schedule with target JDK as a param?

kpavlov:
@langchain4j , please feel free to close this PR if it doesn't seems right

langchain4j:
> Then we should not run them in parallel to avoid hitting rate limits. Maybe we can run single workflow in a different schedule with target JDK as a param?

Yes, makes sense. We can schedule 17 for 00:00 and 21 for 04:00

Issue #2142, opened 2024-11-18
Test EmbeddingStore implementations with the range of versions


langchain4j:
`EmbeddingStore` integrations should be tested with the latest docker image version as well as with a range of older versions to make sure the implementation stays compatible.
Tested and supported versions should be documented.

PR #2143, opened 2024-11-18
Test EmbeddingStore implementations with the range of versions

ghost:
## Issue
Contributes to #2142

## Change
TODO

## General checklist

### DIFF
--- a/langchain4j-chroma/src/test/java/dev/langchain4j/store/embedding/chroma/ChromaEmbeddingStoreIT.java
+++ b/langchain4j-chroma/src/test/java/dev/langchain4j/store/embedding/chroma/ChromaEmbeddingStoreIT.java
@@ -32,7 +32,7 @@ import static org.assertj.core.api.Assertions.assertThat;
 class ChromaEmbeddingStoreIT extends EmbeddingStoreWithFilteringIT {
 
     @Container
-    private static final ChromaDBContainer chroma = new ChromaDBContainer("chromadb/chroma:0.5.4");
+    private static final ChromaDBContainer chroma = new ChromaDBContainer("chromadb/chroma:latest");
 
     EmbeddingStore<TextSegment> embeddingStore = ChromaEmbeddingStore.builder()
             .baseUrl(chroma.getEndpoint())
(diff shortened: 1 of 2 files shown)


langchain4j-chroma/src/test/java/dev/langchain4j/store/embedding/chroma/ChromaEmbeddingStoreIT.java
```diff
@@ -32,7 +32,7 @@
 class ChromaEmbeddingStoreIT extends EmbeddingStoreWithFilteringIT {
 
     @Container
-    private static final ChromaDBContainer chroma = new ChromaDBContainer("chromadb/chroma:0.5.4");
+    private static final ChromaDBContainer chroma = new ChromaDBContainer("chromadb/chroma:latest");
```
kpavlov, line 35: This is actually less deterministic than targeting the specific version

langchain4j-chroma/src/test/java/dev/langchain4j/store/embedding/chroma/ChromaEmbeddingStoreRemovalIT.java
```diff
@@ -23,7 +23,7 @@
 public class ChromaEmbeddingStoreRemovalIT extends EmbeddingStoreWithRemovalIT {
 
     @Container
-    private static final ChromaDBContainer chroma = new ChromaDBContainer("chromadb/chroma:0.5.2");
+    private static final ChromaDBContainer chroma = new ChromaDBContainer("chromadb/chroma:latest");
```
kpavlov, line 26: same here

langchain4j-chroma/src/test/java/dev/langchain4j/store/embedding/chroma/ChromaEmbeddingStoreIT.java
langchain4j: Just in case, this PR is not complete, it is just a reminder for myself to explore this idea.

> This is actually less deterministic than targeting the specific version


But that's the whole point. I think ideally we should:
- test multiple specific versions (and document which versions are tested and supported)
- test the `latest` version (to have a faster feedback loop when something is breaking in the new version)

PR #2156, opened 2024-11-21
EmbeddingStore: implement addAll(ids, embeddings, segments)

### LINKED ISSUE #2106: [FEATURE] Need method add(String id, Embedding embedding, Embedded embedded) in the EmbeddingStore
interface dev.langchain4j.store.embedding.EmbeddingStore has such method as
`void add(String id, Embedding embedding);`

and

`String add(Embedding embedding, Embedded embedded);`

but there is no method like

`add(String id, Embedding embedding, Embedded embedded)`

I would like to add to my database embedding with my own ID and I want to see the object which was embedded

This will save me from having to go to some other storage to retrieve the Embedded object under  ID.

VictorZZZZ:
## Issue
Closes #2106 
Contributes to #2160

## Change
Added method `addAll(**ids**, embeddings, embedded)`

## General checklist
- [x] I have manually run all the unit tests in the module I have added/changed, and they are all green


langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -52,14 +54,39 @@ public interface EmbeddingStore<Embedded> {
...
 
+    /**
+     * Adds multiple embeddings and their corresponding contents that have been embedded to the store.
+     *
+     * @param ids A list of embeddings ids associated with the added embeddings.
+     * @param embeddings A list of embeddings to be added to the store.
+     * @param embedded   A list of original contents that were embedded.
+     * @return A list of IDs associated with the added embeddings.
+     */
+    List<String> addAll(List<String> ids, List<Embedding> embeddings, List<Embedded> embedded);
```
langchain4j:
```suggestion
    void addAll(List<String> ids, List<Embedding> embeddings, List<Embedded> embedded);
```
since `ids` are provided as a parameter, there is no need to return them

langchain4j: nit: please move it below `addAll(List<Embedding> embeddings, List<Embedded> embedded)`

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -52,14 +54,39 @@ public interface EmbeddingStore<Embedded> {
...
+    default List<String> addAll(List<Embedding> embeddings, List<Embedded> embedded) {
+        return addAll(generateIds(embeddings.size()), embeddings, embedded);
+    }
+
+
+    /**
+     * Generates list of UUID strings
+     * @param dimension  - dimension of list
+     */
+    default List<String> generateIds(long dimension) {
```
langchain4j:
```suggestion
    default List<String> generateIds(int n) {
```
`dimension` has a specific meaning in the context of embeddings, so let's better rename it into `n`

langchain4j-vearch/src/main/java/dev/langchain4j/store/embedding/vearch/VearchEmbeddingStore.java
```diff
@@ -0,0 +1,279 @@
...
+import java.util.*;
+
+import static dev.langchain4j.internal.Utils.*;
+import static dev.langchain4j.internal.ValidationUtils.*;
+import static java.time.Duration.ofSeconds;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonList;
+import static java.util.stream.Collectors.toList;
+
+public class VearchEmbeddingStore implements EmbeddingStore<TextSegment> {
```
langchain4j: Vearch module has been removed from this repository, please remove this class

ghost (commented):
@VictorZZZZ thanks a lot!

Could you please add a test to `EmbeddingStoreWithoutMetadataIT` that tests the new method?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
VictorZZZZ: +++

VictorZZZZ: +++

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
VictorZZZZ: +++

langchain4j-vearch/src/main/java/dev/langchain4j/store/embedding/vearch/VearchEmbeddingStore.java
VictorZZZZ: +++

VictorZZZZ:
> @VictorZZZZ thanks a lot!
> 
> Could you please add a test to `EmbeddingStoreWithoutMetadataIT` that tests the new method?

@langchain4j 

New method addAll(ids, embeddings, embedded) now is used inside addAll(embeddings, embedded)

It's covered by existing tests. Is there a need to cover it separately?

langchain4j:
@VictorZZZZ yes, please implement a test. Each public API method should be covered with a test, as each integration can re-implement it and we need to make sure it works as expected. Thank you!

VictorZZZZ:
> @VictorZZZZ yes, please implement a test. Each public API method should be covered with a test, as each integration can re-implement it and we need to make sure it works as expected. Thank you!

@langchain4j done

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -55,11 +57,12 @@ public interface EmbeddingStore<Embedded> {
     /**
      * Adds multiple embeddings and their corresponding contents that have been embedded to the store.
      *
+     * @param ids A list of embeddings ids associated with the added embeddings.
```
langchain4j:
```suggestion
     * @param ids A list of IDs associated with the added embeddings.
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -55,11 +57,12 @@ public interface EmbeddingStore<Embedded> {
...
      * @param embeddings A list of embeddings to be added to the store.
      * @param embedded   A list of original contents that were embedded.
-     * @return A list of auto-generated IDs associated with the added embeddings.
+     * @return A list of IDs associated with the added embeddings.
```
langchain4j:
```suggestion
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -102,6 +105,32 @@ default void removeAll() {
...
     }
 
+    /**
+     * Adds multiple embeddings and their corresponding contents that have been embedded to the store.
+     *
+     * @param embeddings A list of embeddings to be added to the store.
+     * @param embedded   A list of original contents that were embedded.
+     * @return A list of auto-generated IDs associated with the added embeddings.
+     */
+    default List<String> addAll(List<Embedding> embeddings, List<Embedded> embedded) {
```
langchain4j: this method is out of place, please move it higher to all the other `add`/`addAll `methods

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -55,11 +57,12 @@ public interface EmbeddingStore<Embedded> {
...
      * Adds multiple embeddings and their corresponding contents that have been embedded to the store.
      *
+     * @param ids A list of embeddings ids associated with the added embeddings.
      * @param embeddings A list of embeddings to be added to the store.
      * @param embedded   A list of original contents that were embedded.
-     * @return A list of auto-generated IDs associated with the added embeddings.
+     * @return A list of IDs associated with the added embeddings.
      */
-    List<String> addAll(List<Embedding> embeddings, List<Embedded> embedded);
+    void addAll(List<String> ids, List<Embedding> embeddings, List<Embedded> embedded);
```
langchain4j: This is a breking change, there are implementations of this interface in other repositories and maintained by other parties. What about making it `default` and throwing `UnsupportedOperationException` by default?

ghost (commented):
@VictorZZZZ thank you!

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
langchain4j: Just in case: this is not **_very_** critical, but if there is a way to avoid breaking chnages, it would be great. If it requires tonnes of manual changes, I guess we can live with it

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
VictorZZZZ: +++

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
VictorZZZZ: it's a new method, And I implemented it everywhere

langchain4j: Not everywhere, there are implementations of `EmbeddingStore` interface that live outside of this repository

VictorZZZZ: I can do it, but I'm not sure that it is a good desicion.

It will compile every time, and not implemented - not very good.

langchain4j: exception should be thrown only for this implementations outside of this repo - they are responsible to implement it if they want to use that new method

VictorZZZZ: done/ but don't like it

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -59,7 +61,22 @@ public interface EmbeddingStore<Embedded> {
...
+
+    /**
+     * Adds multiple embeddings and their corresponding contents that have been embedded to the store.
+     *
+     * @param ids A list of IDs associated with the added embeddings.
+     * @param embeddings A list of embeddings to be added to the store.
+     * @param embedded   A list of original contents that were embedded.
+     */
+    default void addAll(List<String> ids, List<Embedding> embeddings, List<Embedded> embedded) {
+        throw new UnsupportedOperationException("Method need to be implemented manually");
```
langchain4j:
```suggestion
        throw new UnsupportedOperationException("Not supported yet.");
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
langchain4j: @VictorZZZZ thanks! Please explain why don't you like it?

VictorZZZZ: ...

langchain4j: @VictorZZZZ sorry, I forgot to sumbit my comments :/

VictorZZZZ: don't like 'default' because we are doing a new version of library, so it's a new method in our EmbeddingStore interface and all required implementations in library we made. 
We don't know who will use and it's OK, we shoudn't know this.
If somebody wants to use a new version, they must be ready to implement required methods.

And here it turns out that by adding a `default`, we are accommodating them so as not to break anything for them. But it should be the other way around.

langchain4j: @VictorZZZZ I see, thank you for the explanation. I see it differently. This library is used by many people, and it is our duty to make the experience for them as seamless as possible. It does not cost us much to avoid breaking changes, so I see no reason to break anything for our users, especially on purpose. Backward compatibility is a thing, especially in Java world.

VictorZZZZ: @langchain4j 
It's a new method. There is no anything about "Backward compatibility"

langchain4j: If you introduce a new non-default method to the interface, all implementations of this interface have to be updated, otherwise it will not compile

VictorZZZZ: it's a new version of core library. User must be ready for this. 
Backward compatibility will not be affected.

langchain4j: This would be OK if we were to release a major version update, but we are far from it. So let's keep minor versions without too many breaking changes.

VictorZZZZ: I didn't think about it in minor/major versions projection.

But there are some changes in already existed method `addAll(List<Embedding> embeddings, List<Embedded> embedded)`

now it uses new method `addAll(List<String> ids, List<Embedding> embeddings, List<Embedded> embedded)` - which is throws exception by default.

langchain4j: Ah, so you mean that other implementations of `EmbeddingStore` that we do not control will throw an exception when `addAll(List<Embedding> embeddings, List<Embedded> embedded)` is called?
This is not nice, it could by decoupling `addAll(List<Embedding> embeddings, List<Embedded> embedded)` from the new method. WIll this be too much of changes?

VictorZZZZ: Yes `EmbeddingStores` we don't control will be broken if there won't be implementation of new method.

Decoupling is not hard. But it's not correct.
That's why i'd like to make a new method required to implement.

langchain4j: If decoupling is not hard, could you please do it?

I see no reason why existing code that uses `EmbeddingStore.addAll(embeddings, segments)` or `EmbeddingStore` in general should break when updating to the next LC4j version.

It should only fail if some (unknown to us) `EmbeddingStore` implementation does not implement the new default `addAll(ids, embeddings, segments)` method, but someone calls it. In all other cases there should be no compilation or runtime errors.

VictorZZZZ: @langchain4j, I have done decoupling

VictorZZZZ: ...

langchain4j: I guess I made a mistake, and I apologize.

Decoupling was actually not needed, the default version of `addAll(List<Embedding>, List<Embedded>)` (which now has duplicated implementation in every store) can call `addAll(List<String>, List<Embedding>, List<Embedded>)` which has a default implementation that throws an exception.

For `EmbeddingStore` implementations that we control, every one of them will have `addAll(List<String>, List<Embedding>, List<Embedded>)` overriden/implemented, right? So exception will not be thrown when either of the methods is called.

For `EmbeddingStore` implementations that we do not control, every of them will have a `addAll(List<Embedding>, List<Embedded>)` implemented, so when it is called, there will be no exception either.

Sorry for the confusion that lead to extra work from your side!
If it is easy to roll back the changes, it would be great to do so in order to avoid duplicated logic for `addAll(List<Embedding>, List<Embedded>)` in every store. If it is not, feel free to live it like this, we can live with some duplication. Most important is that it works and tested.

VictorZZZZ: OK.
Reverted my last commit

ghost (approved):
@VictorZZZZ thank you!

Spotless is failing, but I would reformat it in another commit.

PR #2181, opened 2024-11-24
Jlama: add tool support to streaming chat models. closes #2115

### LINKED ISSUE #2115: Jlama: support tools/functions in streaming mode
Tools/functions should be supported in streaming mode. Currently, they work only in sync mode.

tjake:
## Issue

Closes #2115 

## Change
Adds tool support to streaming chat models

## General checklist


langchain4j-jlama/src/main/java/dev/langchain4j/model/jlama/JlamaStreamingChatModel.java
```diff
@@ -68,25 +83,74 @@ public static JlamaStreamingChatModelBuilder builder() {
 
     @Override
     public void generate(List<ChatMessage> messages, StreamingResponseHandler<AiMessage> handler) {
+        generate(messages, List.of(), handler);
+    }
+
+    @Override
+    public void generate(List<ChatMessage> messages, ToolSpecification toolSpecification, StreamingResponseHandler<AiMessage> handler) {
+        generate(messages, List.of(toolSpecification), handler);
```
langchain4j: The `generate(List<ChatMessage>, ToolSpecification, StreamingResponseHandler<AiMessage>)` method is meant to **_force_** the LLM to execute the specified tool. It this supported by Jlama? If not, it would be better to throw an unsupported exception

langchain4j-jlama/src/main/java/dev/langchain4j/model/jlama/JlamaStreamingChatModel.java
```diff
@@ -55,7 +70,7 @@ public JlamaStreamingChatModel(Path modelCachePath,
             loader = loader.workingDirectory(workingDirectory);
 
         this.model = loader.load();
-        this.temperature = temperature == null ? 0.7f : temperature;
+        this.temperature = temperature == null ? 0.3f : temperature;
```
langchain4j, line 73: Is there a default value for `temperature` in Jlama? Perharps we should not have defaults on LC4j side?

langchain4j-jlama/src/test/java/dev/langchain4j/model/jlama/JlamaChatModelToolsIT.java
```diff
@@ -75,4 +89,42 @@ void should_execute_a_tool_then_answer() {
         assertThat(secondAiMessage.text()).contains("32");
         assertThat(secondAiMessage.toolExecutionRequests()).isNull();
     }
+
+    @Test
+    void should_execute_a_tool_then_answer_streaming() {
```
langchain4j: Please move this into a dedicated `JlamaStreamingChatModelToolsIT`

ghost (commented):
@tjake thanks a lot!

langchain4j-jlama/src/main/java/dev/langchain4j/model/jlama/JlamaStreamingChatModel.java
langchain4j: BTW, same applies to `JlamaChatModel.generate(List<ChatMessage>, ToolSpecification)`

tjake: I don't know of any open models that do this, you kind of just ask the LLM.  I guess it could add a system prompt like "You MUST use the tool specified".  Do you want me todo that?

langchain4j-jlama/src/main/java/dev/langchain4j/model/jlama/JlamaStreamingChatModel.java
tjake: Models don't have defaults like this, I could add a default on jlamas side, but in my api its required.

langchain4j-jlama/src/main/java/dev/langchain4j/model/jlama/JlamaStreamingChatModel.java
langchain4j: I see, then better to just throw an exception that this is not supported

langchain4j-jlama/src/main/java/dev/langchain4j/model/jlama/JlamaStreamingChatModel.java
langchain4j: OK let's keep it then, thanks

langchain4j-jlama/src/main/java/dev/langchain4j/model/jlama/JlamaStreamingChatModel.java
tjake: Ok.. Can you should fix the javadoc on this method as there's no indication of this invariant.

tjake: Oh the javadoc is correct on `ChatLanguageModel` but not `StreamingChatLanguageModel`

langchain4j: Oh good catch, thanks! I will update it

ghost (commented):
@tjake something seems to be wrong with downloading the models. Tests are stuck and the directory with the model looks like this:

image

Do you know what might be the reason?

tjake:
> Do you know what might be the reason?

Do you use the same instances for testing? Maybe the downloads failed and its seeing the bad one and needs to be cleared before the test action?

langchain4j-jlama/src/test/java/dev/langchain4j/model/jlama/JlamaChatModelToolsIT.java
```diff
@@ -34,6 +34,7 @@ static void setup() {
                 .temperature(0.0f)
                 .maxTokens(1024)
                 .build();
+
```
langchain4j:
```suggestion
```

ghost (dismissed):
@tjake tried again, now it works :)

Thanks a lot for the PR!

PR #2190, opened 2024-11-26
Avoid NPE in advance and remove unnecessary conditional statements.

jiangsier-xyz:
## Issue
The null pointer check is too late. NPE will be thrown earlier.

## Change
Avoid NPE in advance and remove unnecessary conditional statements.


langchain4j:
Hi @jiangsier-xyz, thank you!
When the user explicitly configures `CompressingQueryTransformer`, I assume they expect it to work, right?
If there is no `ChatMemory` configured, (as a user) I would expect to get an error or warning that will surface the problem and suggests how to fix it (_If you wish to use CompressingQueryTransformer, please configure ChatMemory or ChatMemoryProvider, blablabla_) instead of silently skipping compressing logic. WDYT?

jiangsier-xyz:
> Hi @jiangsier-xyz, thank you! When the user explicitly configures `CompressingQueryTransformer`, I assume they expect it to work, right? If there is no `ChatMemory` configured, (as a user) I would expect to get an error or warning that will surface the problem and suggests how to fix it (_If you wish to use CompressingQueryTransformer, please configure ChatMemory or ChatMemoryProvider, blablabla_) instead of silently skipping compressing logic. WDYT?

Both options are acceptable. In any case, the null pointer check below is definitely redundant.

If I had to choose one, I'm not sure if users can guarantee that they'll always pass in a ChatMemory instance. In some scenarios, many instances are automatically created based on configuration and can even change dynamically at runtime. They may not want to spend time checking constraints between upstream and downstream components. Adding a warning and allowing the workflow to continue seems to be a more gentle approach.

langchain4j:
@jiangsier-xyz agreed, null check below makes no sense.

> If I had to choose one, I'm not sure if users can guarantee that they'll always pass in a ChatMemory instance. In some scenarios, many instances are automatically created based on configuration and can even change dynamically at runtime. They may not want to spend time checking constraints between upstream and downstream components. Adding a warning and allowing the workflow to continue seems to be a more gentle approach.

I am not sure I understand the use case. `CompressingQueryTransformer` makes no sense without configured `ChatMemory`. If user configures `CompressingQueryTransformer` explicitly, it means that user wants to compress the `ChatMemory`. If `ChatMemory` is not configured, and we silently skip it, I am sure this will just lead to the frustration ("why doesn't it work???") and those RAG flows are not that easy to debug (yet).

If there are (rare) specific use cases that assume optionality of `ChatMemory` (although I am having hard time imagining one, please LMK what you have in mind), user could opt-in to disable the validation check explicitly. But for the majority of use cases, I assume that absense of `ChatMemory` when using `CompressingQueryTransformer` should lead to a failure

jiangsier-xyz:
> @jiangsier-xyz agreed, null check below makes no sense.
> 
> > If I had to choose one, I'm not sure if users can guarantee that they'll always pass in a ChatMemory instance. In some scenarios, many instances are automatically created based on configuration and can even change dynamically at runtime. They may not want to spend time checking constraints between upstream and downstream components. Adding a warning and allowing the workflow to continue seems to be a more gentle approach.
> 
> I am not sure I understand the use case. `CompressingQueryTransformer` makes no sense without configured `ChatMemory`. If user configures `CompressingQueryTransformer` explicitly, it means that user wants to compress the `ChatMemory`. If `ChatMemory` is not configured, and we silently skip it, I am sure this will just lead to the frustration ("why doesn't it work???") and those RAG flows are not that easy to debug (yet).
> 
> If there are (rare) specific use cases that assume optionality of `ChatMemory` (although I am having hard time imagining one, please LMK what you have in mind), user could opt-in to disable the validation check explicitly. But for the majority of use cases, I assume that absense of `ChatMemory` when using `CompressingQueryTransformer` should lead to a failure

Oh, what I was envisioning at the time was an orchestration-based low-code platform where users could configure different components to create a pipeline. The CompressingQueryTransformer might be placed alongside other QueryTransformers to allow for online configuration. Additionally, this low-code platform does not perform parameter validation for different components (such as CompressingQueryTransformer).

However, as I mentioned earlier, both options are acceptable. Perhaps we could use IllegalArgumentException instead of NPE?

langchain4j:
@jiangsier-xyz what about `dev.langchain4j.exception.IllegalConfigurationException` or `IllegalStateException`?

jiangsier-xyz:
> @jiangsier-xyz what about `dev.langchain4j.exception.IllegalConfigurationException` or `IllegalStateException`?

IllegalConfigurationException seems quite appropriate.
But I'm not sure whether relying on dev.langchain4j:langchain4j is a good idea. langchain4j depends on langchain4j-open-ai, right? langchain4j-dashscope should be on the same level as langchain4j-open-ai.

langchain4j:
@jiangsier-xyz oh, I missed that that exception is in `langchain4j` module. Depending on where/when we would like to do this validation, if required, it could be moved to `langchain4j-core` module.

> langchain4j depends on langchain4j-open-ai, right?

Only for testing, and this will change soon

jiangsier-xyz:
> If you wish to use CompressingQueryTransformer, please configure ChatMemory or ChatMemoryProvider

How about using Exceptions.illegalArgument("please configure ChatMemory") for now?

jiangsier-xyz:
> @jiangsier-xyz oh, I missed that that exception is in `langchain4j` module. Depending on where/when we would like to do this validation, if required, it could be moved to `langchain4j-core` module.
> 
> > langchain4j depends on langchain4j-open-ai, right?
> 
> Only for testing, and this will change soon

How about using Exceptions.illegalArgument("please configure ChatMemory") for now?

langchain4j:
@jiangsier-xyz then better `IllegalStateException` or `RuntimeException`, not sure how this situation is related to illegal arguments

jiangsier-xyz:
> @jiangsier-xyz then better `IllegalStateException` or `RuntimeException`, not sure how this situation is related to illegal arguments

Done.

langchain4j-core/src/main/java/dev/langchain4j/rag/query/transformer/CompressingQueryTransformer.java
```diff
@@ -70,6 +66,10 @@ public static CompressingQueryTransformerBuilder builder() {
 
     @Override
     public Collection<Query> transform(Query query) {
+        if (query.metadata() == null || query.metadata().chatMemory() == null) {
+            throw new IllegalStateException("ChatMemory is missing " +
```
langchain4j:
```suggestion
            throw new IllegalStateException("ChatMemory is not configured. " +
```

langchain4j-core/src/main/java/dev/langchain4j/rag/query/transformer/CompressingQueryTransformer.java
```diff
@@ -70,6 +66,10 @@ public static CompressingQueryTransformerBuilder builder() {
...
+                    "(if you wish to use CompressingQueryTransformer, please configure ChatMemory or ChatMemoryProvider)");
```
langchain4j:
```suggestion
                    "If you wish to use CompressingQueryTransformer, please configure ChatMemory or ChatMemoryProvider.");
```

ghost (commented):
@jiangsier-xyz thank you! Could you please add a test for this? 🙏

jiangsier-xyz:
> @jiangsier-xyz thank you! Could you please add a test for this? 🙏

Done.

langchain4j-core/src/test/java/dev/langchain4j/rag/query/transformer/CompressingQueryTransformerTest.java
```diff
@@ -168,4 +169,35 @@ void should_compress_query_and_chat_memory_into_single_query_using_custom_prompt
...
+        // then
+        assertThrows(IllegalStateException.class, () -> transformer.transform(query));
+    }
+
+    @Test
+    void should_fail_without_chat_memory_in_metadata() {
+
+        // given
+        UserMessage userMessage = UserMessage.from("Hello");
+        Metadata metadata = Metadata.from(userMessage, null, null);
```
langchain4j, line 187: metadata is unused

jiangsier-xyz: > metadata is unused

My fault. Fixed.

ghost (approved):
@jiangsier-xyz thank you!

langchain4j:
I was about to merge this PR but then recalled about https://github.com/langchain4j/langchain4j/pull/2293, something strange is happening there, so I propose to put it on hold for now as there is a risk to introduce a blocker for some users

jiangsier-xyz:
> I was about to merge this PR but then recalled about #2293, something strange is happening there, so I propose to put it on hold for now as there is a risk to introduce a blocker for some users

ok for me

Issue #2118, opened 2024-11-14
Improve Prompt Templates


langchain4j:
Consider adding a few improvements to current prompt template features:
- nested properties
- loops
- conditional expressions

Consider using existing (lightweight) templating engine (can be different ones for vanilla LC4j and SB integration) for this.

langchain4j:
cc @glaforge (you were interested in this)

glaforge:
So the idea would be to bring back templating like mustache or jinja (often used in the Python ecosystem) as a dedicated module maybe? Or you want to extend the current variable placeholder mechanism?

langchain4j:
I guess we need to consider and evaluate all options. Ofc it would be great to have more features by default, but if it will require importing heawyweight library, then it is a no-go. Having a separate out-of-the-box module that enables better templating sounds good, but I would investigate first if there is lightweight option to include by default.

glaforge:
I was looking at Jinja, because it's what's used by LangChain, so it could help porting nice prompts from LangChain to LangChain4j, however the jinjava library has [tons of dependencies](https://github.com/HubSpot/jinjava/blob/master/pom.xml). On the other hand, the jmustache library has [zero dependencies](https://github.com/samskivert/jmustache/blob/master/pom.xml) at all.
But there are other options as well...

My fear is that we turn the existing placeholder value replacement into something a bit too complex, and we invent yet another template engine :-)

langchain4j:
Let's not reinvent the wheel and search for a lightweight solution that we can include with LC4j by default. If there is no such option, we can always implement a separate module that will optionally enable advanced templating features with one of the popular temlating engines (even if heavyweight)

kpavlov:
This is an [example](https://github.com/kpavlov/langchain4j-kotlin/pull/27), how to introduce external template source and rendering mechanism it by extending LC4J. Here is the [documentation](https://github.com/kpavlov/langchain4j-kotlin/blob/main/docs/PromptTemplates.md) and [test](https://github.com/kpavlov/langchain4j-kotlin/blob/a7f671947d49d26f308e965c58d3b4234a393c43/langchain4j-kotlin/src/test/kotlin/me/kpavlov/langchain4j/kotlin/service/ServiceWithPromptTemplatesTest.kt#L21-L20). 

**TL;DR:**
```kotlin
interface Assistant {
    @UserMessage(
        // "Hello, {{userName}}! {{message}}"
[...]

PR #2194, opened 2024-11-26
#2118 Add TemplateRenderingEngine interface and implementation

kpavlov:
## Issue
Contributes to #2118 

## Change

Introduce the TemplateRenderingEngine interface for rendering templates with dynamic variables, moving rendering logic out of the Template interface. Updated related classes and added tests to ensure proper functionality of the template rendering process.

First step: introduced TemplateRenderingEngine interface and moved template rendering from DefaultPromptTemplateFactory. [This logic ](https://github.com/kpavlov/langchain4j/blob/5f234212d1a0c4f8dfe32fcdaa527d871a1b80f5/langchain4j-core/src/main/java/dev/langchain4j/model/input/PromptTemplate.java#L112-L117)should move to DefaultReneringEngine in follow-up PRs.
Using factories/ServiceLoader to load TemplateRenderingEngine is out of scope of this PR and should be done separately.

### Separation of Concerns

The template rendering logic has been moved from the Template interface to a new TemplateRenderingEngine interface. This separation allows for better modularity and easier maintenance.

### Extensibility:
- Interface Introduction: Introducing the TemplateRenderingEngine interface makes it easier to extend or replace the rendering logic without modifying the core Template interface.
- Default Implementation: A default implementation (DefaultTemplateRenderingEngine) is provided, which handles the rendering of templates with dynamic variables. This can be replaced or extended by other implementations if needed.

### Simplified Template Class:
- The Template class now delegates the rendering logic to the TemplateRenderingEngine, making the Template class simpler and more focused on representing templates.

## General checklist


langchain4j:
cc @geoand

ghost (commented):
@kpavlov thank you!

I am having a hard time understanding the need for this though.
Could you please expain why this change is required?

Now we have:
- `PromptTemplate` - a wrapper for a template string with placeholders
- `Prompt` - rendered prompt (the result of applying values to `PromptTemplate`)
- `Template` - What is this actually? Seems to overlap with `PromptTemplate`
- `PromptTemplateFactory` - factory for `Template`. BTW names do not match, we need to change this
- `TemplateRenderingEngine` - why is this required and why can't `PromptTemplate` ~~render itself~~ delegate rendering to `Template`?

kpavlov:
Thank you, @langchain4j , for looking into this.

My primary goal was to improve the separation of concerns and make the code more modular by having a dedicated engine for rendering templates with dynamic variables. Coupling rendering with data (prompt templates) makes code less maintainable ([SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) voilation) and makes introducing other rendering engines difficult.
I did not introduce `Template`, but moved it from the `PromptTemplateFactory` to the upper level. I do agree that it duplicated the `PromptTemplate`, but I wanted to avoid big refactoring in a single PR. The next steps would be

geoand:
> cc @geoand

Thanks for the heads up!

langchain4j:
Hi @kpavlov, thanks for the explanation!

> My primary goal was to improve the separation of concerns and make the code more modular by having a dedicated engine for rendering templates with dynamic variables. 
Coupling rendering with data (prompt templates) makes code less maintainable ([SRP](https://en.wikipedia.org/wiki/Single-responsibility_principle) voilation) and makes introducing other rendering engines difficult.

As far as I understand it, this is already the case. `PromptTemplateFactory.Template` is responsible for rendering, `PromptTemplate` is responsible for keeping the template string and delegating rendering to `PromptTemplateFactory.Template`. Other rendering engines can be introduced easily, this was done by @geoand exactly for this reason, to be able to plug in Quarkus-specific template engine. So what is missing?


> The next steps would be

Are you gonna finish this sentence? 😆

kpavlov:
@langchain4j , current duplicate type hierarchy was a bit different to understand for me (Template vs PromptTemplate). The [spring-ai structure](https://docs.spring.io/spring-ai/reference/api/prompt.html) has very similar types, but without duplication:
- `Prompt` - what we send to LLM. Text/Data + model parameters
- [`PromptTemplate`](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java) - encapsulates template text + model parameters and uses `TemplateRenderingEngine`. And also it has TemplateFormat property, and hinting that it this is a point for extension. Default implementation is using [StringTemplate](https://www.stringtemplate.org). 

This PR introduces TemplateRenderingEngine (separation of concerns). PromptTemplate and Template should be combined separately, because having two entities is very confusing.
@geoand, what do you think?
I am working on [LangFuse client](https://github.com/kpavlov/langfuse-jvm) and looking for the way to integrate it here... Or it can be [integrated](https://github.com/kpavlov/langchain4j-kotlin/blob/main/docs/PromptTemplates.md) via langchain4j-kotlin

geoand:
> Other rendering engines can be introduced easily, this was done by @geoand exactly for this reason, to be able to plug in Quarkus-specific template engine. So what is missing?

+1. See [this](https://github.com/quarkiverse/quarkus-langchain4j/blob/main/core/runtime/src/main/java/io/quarkiverse/langchain4j/QuarkusPromptTemplateFactory.java) and [this](https://github.com/quarkiverse/quarkus-langchain4j/blob/main/core/runtime/src/main/java/io/quarkiverse/langchain4j/QuarkusStructuredPromptFactory.java)

> @geoand, what do you think?

Unless there is a very compelling reason for change, I would leave things the way they are

langchain4j:
@kpavlov 

> [PromptTemplate](https://github.com/spring-projects/spring-ai/blob/main/spring-ai-core/src/main/java/org/springframework/ai/chat/prompt/PromptTemplate.java) - encapsulates template text + model parameters and uses TemplateRenderingEngine. And also it has TemplateFormat property, and hinting that it this is a point for extension. Default implementation is using [StringTemplate](https://www.stringtemplate.org/).

`PromptTemplate` implementation in Spring AI does not allow for extension, how do you extend an enum without changing the code? It is tightly coupled to ST (or anything else that Spring team might add in the future).

In our implementation any user can implement `PromptTemplateFactory` interface in their code with any template engine of choice and it will be plugged in via SPI. What am I missing?

> This PR introduces TemplateRenderingEngine (separation of concerns). 

This extra entity seems to only add more complexity and indirection. What is currently not possible that will become possible with this change?

> PromptTemplate and Template should be combined separately, because having two entities is very confusing.

I agree that `Template` name could be improved, any ideas?

@geoand would it work for Quarkus if `Template.render` signature will change from `String render(Map<String, Object> variables);` to `String render(String template, Map<String, Object> variables);`? If yes, we could also rename `Template` into `TemplateRenderingEngine` and we will achieve:
- complete separation of concerns: `TemplateRenderingEngine` will just render given template with given variables. String template will be provided to it from outside, allowing it to be loaded dynamically from external sources like LangFuse) 
- clearer naming

geoand:
Sure, that would be fine

langchain4j:
@kpavlov do you plan to submit another PR? I think we have found a solution after quite a bit of back and forth

PR #2208, opened 2024-11-28
Apache Tika Document Parser: Fix failure due to XPathFactoryConfigurationException

### LINKED ISSUE #2209: [BUG] ApacheTikaDocumentParser doesn't work in production
### The Problem

When using the `ApacheTikaDocumentParser` in production, pulling it in as dependency causes the following code (which is internally used by the Tika parser)

```java
XPathFactory factory = XPathFactory.newInstance();
```

to always fail with:

```
javax.xml.xpath.XPathFactoryConfigurationException:
  No XPathFctory implementation found for the object model:
    http://java.sun.com/jaxp/xpath/dom
```

### How to Reproduce

The issue can be reproduced by adding the dependency on the Tika document parser module:

```xml
        <dependency>
            <artifactId>langchain4j-document-parser-apache-tika</artifactId>
            <version>0.36.2</version>
        </dependency>
```

Add the ApacheTikaDocumentParser somewhere in the application code and use it.
Alternative approach: Add the ApacheTikaDocumentParser and add the above code to the application, make sure the above code is executed.

Interestingly, the issue does not occur in Quarkus Dev mode.

### Additional Information

- LangChain4j version: 0.36.2
- Java version: OpenJDK Runtime Environment Temurin-21.0.5+11 (build 21.0.5+11-LTS)
- Quarkus: 3.17.0

florian-h05:
Fixes #2209.

When using the `ApacheTikaDocumentParser` in production, pulling it in as dependency causes the following code (which is internally used by the Tika parser)

```java
XPathFactory factory = XPathFactory.newInstance();
```

to always fail with:

```
javax.xml.xpath.XPathFactoryConfigurationException:
  No XPathFctory implementation found for the object model:
    http://java.sun.com/jaxp/xpath/dom
```

Upgrading Apache Tika to the latest version (3.0.0) fixed this problem. Looking at their [changelog](https://github.com/apache/tika/blob/3.0.0/CHANGES.txt), I would suspect that the removal on xerces2 dependency fixed this issue (xerces2 is a library providing a XPathFactory implementation). Their breaking changes seem to don't cause problems, JDK 11 requirement is already met due to LangChain4j requirement I think.

Note: One unit test was failing for test-file.pptx, saving the .ppt as .pptx fixed this. I think there was some hidden text.


rzo1 (approved):

ghost (commented):
@florian-h05 thank you!

The builds are failing, please check

rzo1:
Think it might be sufficient to manage the dependency of commons-logging to the latest version, it should be (mostly) backwards compatible.

florian-h05:
@langchain4j I have fixed the build.

Checking the release notes of `commons-logging` (see https://commons.apache.org/proper/commons-logging/changes-report.html), dependencies using commons-logging:1.2 and commons-logging:1.3.3 should be find with commons-logging:1.3.4 as well.

The Java 23 build fails unrelated: 

```
Failed to execute goal on project langchain4j-open-ai: 
  Could not resolve dependencies for project dev.langchain4j:langchain4j-open-ai:jar:0.37.0-SNAPSHOT:
    Failed to collect dependencies at dev.ai4j:openai4j:jar:0.24.0 -> com.squareup.retrofit2:converter-jackson:jar:2.9.0: 
      Failed to read artifact descriptor for com.squareup.retrofit2:converter-jackson:jar:2.9.0: Could not transfer artifact com.squareup.retrofit2:converter-jackson:pom:2.9.0 from/to central (https://repo.maven.apache.org/maven2): /home/runner/.m2/repository/com/squareup/retrofit2/converter-jackson/2.9.0/converter-jackson-2.9.0.pom.part (No such file or directory) -> [Help 1]
```

langchain4j:
@kpavlov could you please help here? 🙏

mariofusco (approved):
I checked this and it works fine for me. I confirm that the CI issue with Java 23 is unrelated.

As a side note I don't like the fact that the [dependency convergence rule](https://maven.apache.org/enforcer/enforcer-rules/dependencyConvergence.html) obliges you to also add that exclusion, but it doesn't seem that there is a way to better configure it. In my opinion once you define a top level dependency it should be obvious that it takes precedence on all the transitive ones, so it shouldn't be necessary to also explicitly add the exclusion.

langchain4j-easy-rag/pom.xml
```diff
@@ -60,6 +64,11 @@
             <artifactId>commons-lang3</artifactId>
             <version>3.16.0</version>
         </dependency>
+        <dependency>
+            <groupId>commons-logging</groupId>
+            <artifactId>commons-logging</artifactId>
+            <version>1.3.4</version>
```
kpavlov, line 68: Versions should be defined in parent-Pom's dependency management section. This ensures consistency across all project modules. Defining a version here might cause conflict with the same dependency, but with different version, defined in another module

kpavlov (dismissed):
While not perfect in terms of dependency management, it is definitely an improvement
Thank you, @florian-h05

kpavlov:
Please, run `make format` to fix formatting

ghost (approved):
@florian-h05 thank you!

PR #2222, opened 2024-12-02
[FEATURE] Added metadata (score and embeddedId) for Content

### LINKED ISSUE #680: [FEATURE] ReRankingContentAggregator incl. MaxResult
The new Advanced RAG provides the `ReRankingContentAggregator` which can re-rank results based on a `ScoringModel` (e.g. the `CohereScoringModel`).

this works great, the re-ranked list of results is definitely better (i.e. more relevant results are more likely to be at the top) than what the vector database returns

However I noticed that the scores Cohere assigns vary a lot, I have seen the largest scores within one batch at around 0.98, but then there are batches where the highest score is 0.03. So the `minScore` property turned out to be not so helpful. What would be great is to have a `maxResults` property as well similar to the one in the `EmbeddingStoreContentRetriever`

### LINKED ISSUE #1733: [FEATURE] TokenStream.onRetrieved, please return the score
Knowing the score is important, as contentAggregator, embeddingRetriever, and customRetriever do not have a score in the returned List<Content>. This is crucial for displaying to users during both development and production phases.
And each Retriever's score can be determined to be the best

emredmrcan:
## Issue
Closes #680
Closes #1733

## Change
Introduced a `metadata` field (`Map<ContentMetadata, Object>`) in the `Content` class to store additional information like `SCORE`, `RERANKED_SCORE` and `EMBEDDING_ID`. 
This allows for the flexibility in associating metadata with content without modifying the Content's core properties.
Updated the `EmbeddingStoreContentRetriever`, `AzureAiSearchContentRetriever`, and `ReRankingContentAggregator` to include the metadata (`SCORE`, `RERANKED_SCORE` and `EMBEDDING_ID`) in the retrieved `Content`. Ensured that these changes do not affect equality checks (comparison of two `Content`s).

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
```diff
@@ -23,19 +24,27 @@
...
+        this(TextSegment.from(text), Map.of());
     }
 
     public Content(TextSegment textSegment) {
+        this(textSegment, Map.of());
+    }
+
+    public Content(TextSegment textSegment, Map<String, Object> metadata) {
         this.textSegment = ensureNotNull(textSegment, "textSegment");
+        this.metadata = metadata == null ? Map.of() : metadata;
```
langchain4j: would be good to copy the map

langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
```diff
@@ -63,4 +73,9 @@ public static Content from(String text) {
     public static Content from(TextSegment textSegment) {
         return new Content(textSegment);
     }
+
+    public static Content from(EmbeddingMatch<TextSegment> embeddingMatch) {
```
langchain4j: `Content` is store-agnostic, can we keep it so (not to couple it with embeddings)?

ghost (commented):
@emredmrcan thank you!

langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
emredmrcan: I agree, just made another commit for this

langchain4j-azure-ai-search/src/main/java/dev/langchain4j/rag/content/retriever/azure/search/AzureAiSearchContentRetriever.java
```diff
@@ -176,9 +176,14 @@ public List<Content> retrieve(Query query) {
...
-                    .map(Content::from)
-                    .collect(toList());
+                    .map(embeddingMatch -> Content.from(
+                            embeddingMatch.embedded(),
+                            Map.of(
+                                    "score", embeddingMatch.score(),
+                                    "embeddingId", embeddingMatch.embeddingId()
+                            )
+                    ))
+                    .toList();
```
kpavlov: Thank you, @emredmrcan 
Would it be possible to add a unit test?

emredmrcan: Hi @kpavlov, I’ve added a metadata assertion for the integration tests and included a unit test for the case where `AzureAiSearchQueryType` is `VECTOR`, as it doesn’t require `searchClient`. Unfortunately, I couldn’t find a way to mock `searchClient` without refactoring the code to accept pre-configured/mocked clients. 
Let me know if you have any suggestions or additional feedback. Otherwise, please feel free to resolve the conversation.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
```diff
@@ -23,19 +24,27 @@
 public class Content {
 
     private final TextSegment textSegment;
+    private final Map<String, Object> metadata;
```
langchain4j: After a second thought, I am wondering if this is the right solution.

A few solutions from the top of my head:
- add `Optional<Double> score` and `Optional<String> embeddingId` fields in the `Content`. Might need one more fields for score after re-ranking.
- create an `EmbeddingStoreContent implements Content` with `Double score` and `String embeddingId` fields.  Will require type check and casting.
- keep existing solution. If we will go this way, let's at least use enums or string constants as keys in this map?

Do you have other ideas on how ths could be implemented?

emredmrcan: @langchain4j  After reviewing the options, I think the second one — creating an `EmbeddingStoreContent` with additional fields — makes the most sense.

I know that the type checks and casting will introduce some complexity but I believe it's a reasonable trade-off for better separation of concerns and improved maintainability.

I'd like to share why I think the first and third options are less ideal as well:

1) Adding `Optional<Double> score` and `Optional<String> embeddingId` fields in `Content`
- This would tie `Content` to "embedding-specific" use cases
- Future extensions could clutter the `Content` class further, violating its current simplicity and flexibility.
2) Keeping the existing solution with a metadata map
- Lacks type safety: Next developer will need to check what’s inside the map
- As we add more metadata or extend functionality, this approach could become messy and harder to maintain.

If you agree, I would proceed with the 2nd option.

langchain4j: Since re-ranking `Contnet`s after retrieving them with `EmbeddingStore` is a pretty common use case, and score changes after re-ranking, how would this work?
- Additionally to `EmbeddingStoreContent` create another `ReRankedContent` or something?
- Override `score` in `EmbeddingStoreContent`?
- Add Optional<Double> rerankedScore to `EmbeddingStoreContent`?

Sorry for so many questions, trying to think out loud :)

emredmrcan: @langchain4j No worries, happy to help! I’d go with Option 2 or Option 3, depending on whether keeping the initial score is necessary. Do you think it’s really needed? If not, Option 2 could be a simpler approach, but if we need to preserve the original score, then Option 3 would be the better choice.

langchain4j: I doubt that original score is required for the majority of use cases. Reranking model is usually much better than emnbedding model, so the final score should be more important. Let's indeed keep it simple and do option 2?

langchain4j: By override I mean clone and override it during cloning, `Content` itself should be immutable

emredmrcan: @langchain4j I agree with Option 2. I’ll proceed with implementing this and complete it by tomorrow morning. Thanks for the clarification!

emredmrcan: @langchain4j I just made another commit where I updated `ContentRetriever` to be a generic interface with a type parameter `<T extends Content>`, added some unit tests and refactored Integration tests.

 I was thinking about our previous conversation... Shall we create another `Content` class (e.g., `ReRankingContent`) in this PR for use in the `ReRankingContentAggregator`? It could follow a similar pattern to `EmbeddingStoreContent` and extend `Content` but will have only `score` and it would clearly distinguish between different types of `Content`s.
 
 Let me know if you have any suggestions or additional feedback.

langchain4j: Hey, I am not sure `ContentRetriever` needs to be parameterized, after all `EmbeddingStoreContent` **_is_** a `Content` as well. I would really love to postpone any changes to `ContentRetriever` to the point when we will implement support for multi-modal RAG.

Regarding new type for re-ranked content: then re-ranked `EmbeddingStoreContent` will become `ReRankedContent` and will lose `embeddingId`?

emredmrcan: @langchain4j You're right; parameterizing `ContentRetriever` can be postponed until we address multi-modal RAG. I’ve reverted the changes to keep things aligned for now.
As for the re-ranked content, I’ve created a new `ReRankedContent` class, which indeed doesn’t include embeddingId. I though this way, we can separate responsibilities between the two content types.
Let me know if this approach makes sense or if you’d prefer an alternative!

langchain4j: TBH I would go with a simpler approach of having only an `EmbeddingStoreContent`, this will also keep the `embeddingId` after reranking.
On the other hand, the `score` after reranking will be interesting not only when the source of `Content` is `EmbeddingStore`, so having a `ReRankedContent` seems reasonable.

But maybe we (or I) are overthinking all of this and we should just add a `Map<ContentMetadata, Object> metadata` to the `Content` for now?

```java
public enum ContentMetadata {
    SCORE_AFTER_RETRIEVAL,
    SCORE_AFTER_RERANKING,
    EMBEDDING_ID
}
```

This will ensure that:
- all data on each step is available
- it will work for all types of `ContentRetriever`
- the most flexible option I guess

Once other types of `ContentRetriever` will start adopting this approach, we could rethink again and see it we could make it even more type safe and specialized.

WDYT?

yushengliao: My task is to retain and record both the original embedding score and the rerank score. This is more convenient for users to observe

kpavlov: @langchain4j, did you consider converting Content to interface for flexibility? Then implementation might be more flexible. Check [Liskov substitution pronciple](https://en.m.wikipedia.org/wiki/Liskov_substitution_principle)
Maybe, create another interface, something like "HasScore" (not sure about the name)?

langchain4j: Turning Content into interface can improve the flexibility a bit, but we are solving a different problem here. Afterwards you will need to create HasX and HasY and HasZ interfaces for all kinds of data...

langchain4j: IIUC, the primary use case for this is for the developer to be able to access the metadata (id, scores, etc) of [sources retrieved](https://docs.langchain4j.dev/tutorials/rag#accessing-sources) during RAG flow.

So the typical code will probably look something like this:
```java
Result<String> result = assistant.chat("How do I ... ?");

String answer = result.content();

for (Content content : result.sources()) {
    String embeddingId = content.metadata().get(EMBEDDING_ID);
    String score = content.metadata().get(SCORE);
    String rerankedScore = content.metadata().get(RERANKED_SCORE);
    // other metadata entries can be populated by custom ContentRetriever and ContentAggregator implementations
}
```

With HasX approach it will look like this:

```java
Result<String> result = assistant.chat("How do I ... ?");

String answer = result.content();

for (Content content : result.sources()) {
    String embeddingId = null;
    if (content instanceof HasEmbeddingId hasEmbeddingId) {
        embeddingId = hasEmbeddingId.embeddingId();
    }

    String score = null;
    if (content instanceof HasScore hasScore) {
        score = hasScore.score();
    }

    String rerankedScore = null;
    if (content instanceof HasRerankedScore hasRerankedScore) {
        rerankedScore = hasRerankedScore.rerankedScore();
    }

    // other metadata entries can be populated by custom ContentRetriever and ContentAggregator 
}
```

emredmrcan: Thank you all for the valuable discussion and insights. I’ve chosen to proceed with the simpler approach of introducing `ContentMetadata`, which makes all necessary data accessible while offering sufficient flexibility for now.

As @langchain4j mentioned, we can revisit and refine this solution in the future as other types of `ContentRetriever` start adopting this approach. I truly appreciate everyone’s input 🙏.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
```diff
@@ -23,19 +24,27 @@
 public class Content {
 
     private final TextSegment textSegment;
+    private final Map<ContentMetadata, Object> metadata;
```
mariofusco, line 27: Warning: this is just a suggestion and it's very likely that I'm overthinking this.

I'm noticing that in a few places like this we have a Map of heterogeneous values, where each value must be of a type determined by its key, but at the moment we don't have a way to enforce the right type at compile time. E.g.: we cannot enforce that the key `ContentMetadata.SCORE` will have a `Double` as value.

A (probably overeengineered) solution)would be to use the Map implementation that I sketched here https://gist.github.com/mariofusco/c9071b51d18b5577d22f3e25cc9e003c using as enumeration a `sealed interface` and a set of singletons implementing it. 

@langchain4j Maybe something similar could be also (re)used for the builders refactor that we were discussing separately?

mariofusco (approved):

langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
langchain4j: I like having better type safety for sure, but AFAIU this approach will require users to introduce new types (that implement `ContentMetadata`) if they will want to add custom fields to their custom `ContentRetriever`s? We need to weight all pros and cons and also analyze what other places might benefit from this approach (to have a unified approach throughout the library. But this requires more time, so I propoge to go with this PR for now and rethink this closer to 1.0?

langchain4j: I missed that currently the key is an enum, so your approach will be better for customization for sure

langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
```diff
@@ -53,6 +62,7 @@ public int hashCode() {
     public String toString() {
         return "Content {" +
                 " textSegment = " + textSegment +
+                ", metadata = " + metadata +
```
langchain4j, line 65: minor: worth leaving a comment in `equals()` and `hashCode()` that `metadata` does not participate in comparisons for {reasons}

langchain4j: Or perhaps it should participate in the comparisons, but we need to reimplement comparison logic in `ReciprocalRankFuser`?

langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
langchain4j: @emredmrcan thanks a lot, greta job!

langchain4j-core/src/main/java/dev/langchain4j/rag/content/Content.java
langchain4j: I will merge this PR in its current state now to unblock a few users waiting for this feature, but we need to revisit this

langchain4j:
@emredmrcan thanks a lot, great job, I especially like the test coverage!

Could you please create a follow-up PR with updates to the documentation?

emredmrcan:
> @emredmrcan thanks a lot, great job, I especially like the test coverage!
> 
> Could you please create a follow-up PR with updates to the documentation?

Thank you for your collaboration and I'm glad the test coverage met your expectations 🙏 I've created a follow-up PR with the requested documentation updates: [PR #2325](https://github.com/langchain4j/langchain4j/pull/2325). Let me know if there's anything else you'd like me to address!

PR #2240, opened 2024-12-05
Initial implementation of the MCP client and tool provider

### LINKED ISSUE #2213: Support Model Context Protocol (MCP)
### Discussed in https://github.com/langchain4j/langchain4j/discussions/2204

<div type='discussions-op-text'>

<sup>Originally posted by **kpavlov** November 28, 2024</sup>
## Idea

> The Model Context Protocol (MCP) is built on a flexible, extensible architecture that enables seamless communication between LLM applications and integrations. This document covers the core architectural components and concepts.

1. Support Model Context Protocol (MCP) for integrating with external Tools.
LC4j may use MCP server to build a ToolSpecification and also run Tool requests against MCP server. 
- https://modelcontextprotocol.io/docs/concepts/tools

2. Support Model Context Protocol (MCP) for to retrieve external with Prompts (related to https://github.com/kpavlov/langchain4j/pull/11)
https://modelcontextprotocol.io/docs/concepts/prompts

## Links
- https://github.com/modelcontextprotocol
- https://modelcontextprotocol.io/docs/concepts/architecture
- https://www.anthropic.com/news/model-context-protocol

</div>

jmartisk:
## Issue

Closes https://github.com/langchain4j/langchain4j/issues/2213

## Change

See the examples in https://github.com/langchain4j/langchain4j-examples/pull/134 
## General checklist

- [ ] The tests cover both positive and negative cases: not yet, but I plan to add more proper error handling as a follow-up

### DIFF
--- a/docs/docs/tutorials/mcp.md
+++ b/docs/docs/tutorials/mcp.md
@@ -1,0 +1,80 @@
+# Model Context Protocol
+
+LangChain4j supports the Model Context Protocol (MCP) to communicate with
+MCP compliant servers that can provide and execute tools. General
+information about the protocol can be found at the [MCP
+website](https://modelcontextprotocol.io/).
+
+The protocol specifies two types of transport, both of these are supported:
+
+- `HTTP`: The client requests an SSE channel to receive events from the
+  server and then sends commands via HTTP POST requests.
+- `stdio`: The client can run an MCP server as a local subprocess and
+  communicate with it directly via standard input/output.
+
+To let your chat model or AI service run tools provided by an MCP server,
+you need to create an instance of an MCP tool provider.
+
+## Creating an MCP tool provider
+
+### Transport
+
+First, you need an instance of a Transport.
+
+For stdio - this example shows how to start a server from a NPM package as a subprocess:
+
+```java
+McpTransport transport = new StdioMcpTransport.Builder()
+    .command(List.of("/usr/bin/npm", "exec", "@modelcontextprotocol/server-everything@0.6.2"))
+    .logEvents(true) // only if you want to see the traffic in the log
+    .build();
+```
+
+For HTTP, you need two URLs, one for starting the SSE channel and one for submitting commands via `POST`:
+
+```java
+McpTransport transport = new HttpMcpTransport.Builder()
+    .sseUrl("http://localhost:3001/sse")
+    .postUrl("http://localhost:3001/message")
+    .logRequests(true) // if you want to see the traffic in the log
+    .logResponses(true)
+    .build();
+```
+
+### Client
+
+To create an MCP client from the transport:
+
+```java
+McpClient mcpClient = new DefaultMcpClient.Builder()
+    .transport(transport)
+    .build();
+```
+
+### Tool provider
+
+Finally, you create an MCP tool provider from the client:
+
+```java
+ToolProvider toolProvider = McpToolProvider.builder()
+    .mcpClients(List.of(mcpClient))
+    .build();
+```
+
+Note that one MCP tool provider can use multiple clients at the same time.
+If you make use of this, you can also specify the behavior of the tool provider
+in cases when retrieving tools from a particular server fails - this is done
+by the `builder.failIfOneServerFails(boolean)` method. The default is `false`, 
+which means that the tool provider will ignore the error from one server and 
+continue with the other servers. If you set it to `true`, a failure from any 
+server will cause the tool provider to throw an exception.
+
+To bind a tool provider to an AI service, simply use the `toolProvider` method
+of an AI service builder:
+
+```java
+Bot bot = AiServices.builder(Bot.class)
+    .chatLanguageModel(model)
+    .toolProvider(toolProvider)
+    .build();
+```
--- a/langchain4j-bom/pom.xml
+++ b/langchain4j-bom/pom.xml
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>dev.langchain4j</groupId>
--- a/langchain4j-bom/pom.xml
+++ b/langchain4j-bom/pom.xml
@@ -217,6 +217,12 @@
 
             <dependency>
                 <groupId>dev.langchain4j</groupId>
+                <artifactId>langchain4j-mcp</artifactId>
+                <version>${project.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>dev.langchain4j</groupId>
                 <artifactId>langchain4j-milvus</artifactId>
                 <version>${project.version}</version>
             </dependency>
--- a/langchain4j-mcp/pom.xml
+++ b/langchain4j-mcp/pom.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>0.37.0-SNAPSHOT</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-mcp</artifactId>
+    <name>LangChain4j :: Integration :: Model Context Protocol</name>
+
+    <dependencies>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp-sse</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.tinylog</groupId>
+            <artifactId>tinylog-impl</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.tinylog</groupId>
+            <artifactId>slf4j-tinylog</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jetbrains</groupId>
+                    <artifactId>annotations</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+    </dependencies>
+
+</project>
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
@@ -1,0 +1,88 @@
+package dev.langchain4j.mcp;
+
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.internal.Utils;
+import dev.langchain4j.mcp.client.McpClient;
+import dev.langchain4j.service.tool.ToolProvider;
+import dev.langchain4j.service.tool.ToolProviderRequest;
+import dev.langchain4j.service.tool.ToolProviderResult;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A tool provider backed by one or more MCP clients.
+ */
+public class McpToolProvider implements ToolProvider {
+
+    private final List<McpClient> mcpClients;
+    private final boolean failIfOneServerFails;
+    private static final Logger log = LoggerFactory.getLogger(McpToolProvider.class);
+
+    private McpToolProvider(Builder builder) {
+        this.mcpClients = new ArrayList<>(builder.mcpClients);
+        this.failIfOneServerFails = Utils.getOrDefault(builder.failIfOneServerFails, false);
+    }
+
+    @Override
+    public ToolProviderResult provideTools(final ToolProviderRequest request) {
+        ToolProviderResult.Builder builder = ToolProviderResult.builder();
+        for (McpClient mcpClient : mcpClients) {
+            try {
+                List<ToolSpecification> toolSpecifications = mcpClient.listTools();
+                for (ToolSpecification toolSpecification : toolSpecifications) {
+                    builder.add(
+                            toolSpecification, (executionRequest, memoryId) -> mcpClient.executeTool(executionRequest));
+                }
+            } catch (Exception e) {
+                if (failIfOneServerFails) {
+                    throw new RuntimeException("Failed to retrieve tools from MCP server", e);
+                } else {
+                    log.warn("Failed to retrieve tools from MCP server", e);
+                }
+            }
+        }
+        return builder.build();
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+
+        private List<McpClient> mcpClients;
+        private Boolean failIfOneServerFails;
+
+        /**
+         * The list of MCP clients to use for retrieving tools.
+         */
+        public McpToolProvider.Builder mcpClients(List<McpClient> mcpClients) {
+            this.mcpClients = mcpClients;
+            return this;
+        }
+
+        /**
+         * The list of MCP clients to use for retrieving tools.
+         */
+        public McpToolProvider.Builder mcpClients(McpClient... mcpClients) {
+            this.mcpClients = Arrays.asList(mcpClients);
+            return this;
+        }
+
+        /**
+         * If this is true, then the tool provider will throw an exception if it fails to list tools from any of the servers.
+         * If this is false (default), then the tool provider will ignore the error and continue with the next server.
+         */
+        public McpToolProvider.Builder failIfOneServerFails(boolean failIfOneServerFails) {
+            this.failIfOneServerFails = failIfOneServerFails;
+            return this;
+        }
+
+        public McpToolProvider build() {
+            return new McpToolProvider(this);
+        }
+    }
+}
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -1,0 +1,148 @@
+package dev.langchain4j.mcp.client;
+
+import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.mcp.client.protocol.InitializeParams;
+import dev.langchain4j.mcp.client.protocol.McpCallToolRequest;
+import dev.langchain4j.mcp.client.protocol.McpInitializeRequest;
+import dev.langchain4j.mcp.client.protocol.McpListToolsRequest;
+import dev.langchain4j.mcp.client.transport.McpTransport;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// TODO: currently we request a new list of tools every time, so we should
+// add support for the `ToolListChangedNotification` message, and then we can
+// cache the list
+public class DefaultMcpClient implements McpClient {
+
+    private static final Logger log = LoggerFactory.getLogger(DefaultMcpClient.class);
+    private final AtomicLong idGenerator = new AtomicLong(0);
+    private final McpTransport transport;
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private final String clientName;
+    private final String clientVersion;
+    private final String protocolVersion;
+
+    public DefaultMcpClient(Builder builder) {
+        transport = ensureNotNull(builder.transport, "transport");
+        clientName = getOrDefault(builder.clientName, "langchain4j");
+        clientVersion = getOrDefault(builder.clientVersion, "1.0");
+        protocolVersion = getOrDefault(builder.protocolVersion, "2024-11-05");
+
+        // Initialize the client...
+        transport.start();
+        McpInitializeRequest request = new McpInitializeRequest(idGenerator.getAndIncrement());
+        InitializeParams params = createInitializeParams();
+        request.setParams(params);
+        JsonNode capabilities = transport.initialize(request);
+        log.debug("MCP server capabilities: {}", capabilities.get("result"));
+    }
+
+    private InitializeParams createInitializeParams() {
+        InitializeParams params = new InitializeParams();
+        params.setProtocolVersion(protocolVersion);
+
+        InitializeParams.ClientInfo clientInfo = new InitializeParams.ClientInfo();
+        clientInfo.setName(clientName);
+        clientInfo.setVersion(clientVersion);
+        params.setClientInfo(clientInfo);
+
+        InitializeParams.Capabilities capabilities = new InitializeParams.Capabilities();
+        InitializeParams.Capabilities.Roots roots = new InitializeParams.Capabilities.Roots();
+        roots.setListChanged(false); // TODO: listChanged is not supported yet
+        capabilities.setRoots(roots);
+        params.setCapabilities(capabilities);
+
+        return params;
+    }
+
+    @Override
+    public List<ToolSpecification> listTools() {
+        McpListToolsRequest operation = new McpListToolsRequest(idGenerator.getAndIncrement());
+        JsonNode jsonNode = transport.listTools(operation);
+        return ToolSpecificationHelper.toolSpecificationListFromMcpResponse(
+                (ArrayNode) jsonNode.get("result").get("tools"));
+    }
+
+    @Override
+    public String executeTool(ToolExecutionRequest executionRequest) {
+        ObjectNode arguments = null;
+        try {
+            arguments = OBJECT_MAPPER.readValue(executionRequest.arguments(), ObjectNode.class);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+        McpCallToolRequest operation =
+                new McpCallToolRequest(idGenerator.getAndIncrement(), executionRequest.name(), arguments);
+        JsonNode jsonNode = transport.executeTool(operation);
+        return ToolExecutionHelper.extractResult(
+                (ArrayNode) jsonNode.get("result").get("content"));
+    }
+
+    @Override
+    public void close() {
+        try {
+            transport.close();
+        } catch (Exception e) {
+            log.warn("Cannot close MCP transport", e);
+        }
+    }
+
+    public static class Builder {
+
+        private McpTransport transport;
+        private String clientName;
+        private String clientVersion;
+        private String protocolVersion;
+
+        public Builder transport(McpTransport transport) {
+            this.transport = transport;
+            return this;
+        }
+
+        /**
+         * Sets the name that the client will use to identify itself to the
+         * MCP server in the initialization message. The default value is
+         * "langchain4j".
+         */
+        public Builder clientName(String clientName) {
+            this.clientName = clientName;
+            return this;
+        }
+
+        /**
+         * Sets the version string that the client will use to identify
+         * itself to the MCP server in the initialization message. The
+         * default value is "1.0".
+         */
+        public Builder clientVersion(String clientVersion) {
+            this.clientVersion = clientVersion;
+            return this;
+        }
+
+        /**
+         * Sets the protocol version that the client will advertise in the
+         * initialization message. The default value right now is
+         * "2024-11-05", but will change over time in later langchain4j
+         * versions.
+         */
+        public Builder protocolVersion(String protocolVersion) {
+            this.protocolVersion = protocolVersion;
+            return this;
+        }
+
+        public DefaultMcpClient build() {
+            return new DefaultMcpClient(this);
+        }
+    }
+}
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
@@ -1,0 +1,23 @@
+package dev.langchain4j.mcp.client;
+
+import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.agent.tool.ToolSpecification;
+import java.util.List;
+
+/**
+ * Represents a client that can communicate with an MCP server over a given transport protocol,
+ * retrieve and execute tools using the server.
+ */
+public interface McpClient extends AutoCloseable {
+
+    /**
+     * Obtains a list of tools from the MCP server.
+     */
+    List<ToolSpecification> listTools();
+
+    /**
+     * Executes a tool on the MCP server and returns the result as a String.
+     * Currently, this expects a tool execution to only contain text-based results.
+     */
+    String executeTool(ToolExecutionRequest executionRequest);
+}
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
@@ -1,0 +1,27 @@
+package dev.langchain4j.mcp.client;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+class ToolExecutionHelper {
+
+    /**
+     * Extracts a single string as a result from the 'content' element of
+     * a CallToolResult message. Only supports String contents, if it finds any
+     * other content types, it throws an exception.
+     */
+    static String extractResult(ArrayNode contents) {
+        Stream<JsonNode> contentStream = StreamSupport.stream(contents.spliterator(), false);
+        return contentStream
+                .map(content -> {
+                    if (!content.get("type").asText().equals("text")) {
+                        throw new RuntimeException("Unsupported content type: " + content.get("type"));
+                    }
+                    return content.get("text").asText();
+                })
+                .collect(Collectors.joining("\n"));
+    }
+}
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
@@ -1,0 +1,118 @@
+package dev.langchain4j.mcp.client;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.model.chat.request.json.JsonArraySchema;
+import dev.langchain4j.model.chat.request.json.JsonBooleanSchema;
+import dev.langchain4j.model.chat.request.json.JsonEnumSchema;
+import dev.langchain4j.model.chat.request.json.JsonIntegerSchema;
+import dev.langchain4j.model.chat.request.json.JsonNumberSchema;
+import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
+import dev.langchain4j.model.chat.request.json.JsonSchemaElement;
+import dev.langchain4j.model.chat.request.json.JsonStringSchema;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+class ToolSpecificationHelper {
+
+    /**
+     * Converts the 'tools' element from a ListToolsResult MCP message
+     * to a list of ToolSpecification objects.
+     */
+    static List<ToolSpecification> toolSpecificationListFromMcpResponse(ArrayNode array) {
+        List<ToolSpecification> result = new ArrayList<>();
+        for (JsonNode tool : array) {
+            final ToolSpecification.Builder builder = ToolSpecification.builder();
+            builder.name(tool.get("name").asText());
+            if (tool.has("description")) {
+                builder.description(tool.get("description").asText());
+            }
+            builder.parameters((JsonObjectSchema) jsonNodeToJsonSchemaElement(tool.get("inputSchema")));
+            result.add(builder.build());
+        }
+        return result;
+    }
+
+    /**
+     * Converts the 'inputSchema' element (inside the 'Tool' type in the MCP schema)
+     * to a JsonSchemaElement object that describes the tool's arguments.
+     */
+    static JsonSchemaElement jsonNodeToJsonSchemaElement(JsonNode node) {
+        String nodeType = node.get("type").asText();
+        if (nodeType.equals("object")) {
+            JsonObjectSchema.Builder builder = JsonObjectSchema.builder();
+            JsonNode required = node.get("required");
+            if (required != null) {
+                builder.required(toStringArray((ArrayNode) required));
+            }
+            if (node.has("additionalProperties")) {
+                builder.additionalProperties(node.get("additionalProperties").asBoolean(false));
+            }
+            JsonNode description = node.get("description");
+            if (description != null) {
+                builder.description(description.asText());
+            }
+            JsonNode properties = node.get("properties");
+            if (properties != null) {
+                ObjectNode propertiesObject = (ObjectNode) properties;
+                for (Map.Entry<String, JsonNode> property : propertiesObject.properties()) {
+                    builder.addProperty(property.getKey(), jsonNodeToJsonSchemaElement(property.getValue()));
+                }
+            }
+            return builder.build();
+        } else if (nodeType.equals("string")) {
+            if (node.has("enum")) {
+                JsonEnumSchema.Builder builder = JsonEnumSchema.builder();
+                if (node.has("description")) {
+                    builder.description(node.get("description").asText());
+                }
+                builder.enumValues(toStringArray((ArrayNode) node.get("enum")));
+                return builder.build();
+            } else {
+                JsonStringSchema.Builder builder = JsonStringSchema.builder();
+                if (node.has("description")) {
+                    builder.description(node.get("description").asText());
+                }
+                return builder.build();
+            }
+        } else if (nodeType.equals("number")) {
+            JsonNumberSchema.Builder builder = JsonNumberSchema.builder();
+            if (node.has("description")) {
+                builder.description(node.get("description").asText());
+            }
+            return builder.build();
+        } else if (nodeType.equals("integer")) {
+            JsonIntegerSchema.Builder builder = JsonIntegerSchema.builder();
+            if (node.has("description")) {
+                builder.description(node.get("description").asText());
+            }
+            return builder.build();
+        } else if (nodeType.equals("boolean")) {
+            JsonBooleanSchema.Builder builder = JsonBooleanSchema.builder();
+            if (node.has("description")) {
+                builder.description(node.get("description").asText());
+            }
+            return builder.build();
+        } else if (nodeType.equals("array")) {
+            JsonArraySchema.Builder builder = JsonArraySchema.builder();
+            if (node.has("description")) {
+                builder.description(node.get("description").asText());
+            }
+            builder.items(jsonNodeToJsonSchemaElement(node.get("items")));
+            return builder.build();
+        } else {
+            throw new IllegalArgumentException("Unknown element type: " + nodeType);
+        }
+    }
+
+    private static String[] toStringArray(ArrayNode jsonArray) {
+        String[] result = new String[jsonArray.size()];
+        for (int i = 0; i < jsonArray.size(); i++) {
+            result[i] = jsonArray.get(i).asText();
+        }
+        return result;
+    }
+}
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/InitializeParams.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/InitializeParams.java
@@ -1,0 +1,95 @@
+package dev.langchain4j.mcp.client.protocol;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+public class InitializeParams {
+
+    private String protocolVersion;
+    private Capabilities capabilities;
+    private ClientInfo clientInfo;
+
+    public String getProtocolVersion() {
+        return protocolVersion;
+    }
+
+    public void setProtocolVersion(final String protocolVersion) {
+        this.protocolVersion = protocolVersion;
+    }
+
+    public Capabilities getCapabilities() {
+        return capabilities;
+    }
+
+    public void setCapabilities(final Capabilities capabilities) {
+        this.capabilities = capabilities;
+    }
+
+    public ClientInfo getClientInfo() {
+        return clientInfo;
+    }
+
+    public void setClientInfo(final ClientInfo clientInfo) {
+        this.clientInfo = clientInfo;
+    }
+
+    public static class Capabilities {
+
+        private Roots roots;
+
+        @JsonInclude(JsonInclude.Include.NON_NULL)
+        private Sampling sampling;
+
+        public Roots getRoots() {
+            return roots;
+        }
+
+        public void setRoots(final Roots roots) {
+            this.roots = roots;
+        }
+
+        public Sampling getSampling() {
+            return sampling;
+        }
+
+        public void setSampling(final Sampling sampling) {
+            this.sampling = sampling;
+        }
+
+        public static class Roots {
+
+            private boolean listChanged;
+
+            public boolean isListChanged() {
+                return listChanged;
+            }
+
+            public void setListChanged(final boolean listChanged) {
+                this.listChanged = listChanged;
+            }
+        }
+
+        public static class Sampling {}
+    }
+
+    public static class ClientInfo {
+
+        private String name;
+        private String version;
+
+        public String getVersion() {
+            return version;
+        }
+
+        public void setVersion(final String version) {
+            this.version = version;
+        }
+
+        public String getName() {
+            return name;
+        }
+
+        public void setName(final String name) {
+            this.name = name;
+        }
+    }
+}
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
@@ -1,0 +1,178 @@
+package dev.langchain4j.mcp.client.transport.http;
+
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import dev.langchain4j.mcp.client.protocol.McpCallToolRequest;
+import dev.langchain4j.mcp.client.protocol.McpInitializeRequest;
+import dev.langchain4j.mcp.client.protocol.McpListToolsRequest;
+import dev.langchain4j.mcp.client.transport.McpTransport;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import okhttp3.sse.EventSource;
+import okhttp3.sse.EventSources;
+
+public class HttpMcpTransport implements McpTransport {
+
+    private final String sseUrl;
+    private final String postUrl;
+    private final OkHttpClient client;
+    private final boolean logResponses;
+    private final boolean logRequests;
+    private EventSource mcpSseEventListener;
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private final Map<Long, CompletableFuture<JsonNode>> pendingOperations = new ConcurrentHashMap<>();
+
+    public HttpMcpTransport(Builder builder) {
+        OkHttpClient.Builder httpClientBuilder = new OkHttpClient.Builder();
+        if (builder.timeout != null) {
+            httpClientBuilder.callTimeout(builder.timeout);
+            httpClientBuilder.connectTimeout(builder.timeout);
+            httpClientBuilder.readTimeout(builder.timeout);
+            httpClientBuilder.writeTimeout(builder.timeout);
+        }
+        this.logRequests = builder.logRequests;
+        if (builder.logRequests) {
+            httpClientBuilder.addInterceptor(new McpRequestLoggingInterceptor());
+        }
+        this.logResponses = builder.logResponses;
+        sseUrl = ensureNotNull(builder.sseUrl, "Missing SSE endpoint URL");
+        postUrl = ensureNotNull(builder.postUrl, "Missing POST endpoint URL");
+        client = httpClientBuilder.build();
+    }
+
+    @Override
+    public void start() {
+        mcpSseEventListener = startSseChannel(logResponses);
+    }
+
+    @Override
+    public JsonNode initialize(final McpInitializeRequest request) {
+        Request httpRequest = null;
+        try {
+            httpRequest = new Request.Builder()
+                    .url(postUrl)
+                    .header("Content-Type", "application/json")
+                    .post(RequestBody.create(OBJECT_MAPPER.writeValueAsBytes(request)))
+                    .build();
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+        return executeAndWait(httpRequest, request.getId());
+    }
+
+    @Override
+    public JsonNode listTools(McpListToolsRequest operation) {
+        try {
+            Request httpRequest = new Request.Builder()
+                    .url(postUrl)
+                    .header("Content-Type", "application/json")
+                    .post(RequestBody.create(OBJECT_MAPPER.writeValueAsBytes(operation)))
+                    .build();
+            return executeAndWait(httpRequest, operation.getId());
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    public JsonNode executeTool(McpCallToolRequest operation) {
+        try {
+            Request httpRequest = new Request.Builder()
+                    .url(postUrl)
+                    .header("Content-Type", "application/json")
+                    .post(RequestBody.create(OBJECT_MAPPER.writeValueAsBytes(operation)))
+                    .build();
+            return executeAndWait(httpRequest, operation.getId());
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private JsonNode executeAndWait(Request request, Long id) {
+        try {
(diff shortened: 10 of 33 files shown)


langchain4j-bom/pom.xml
```diff
@@ -215,6 +215,12 @@
                 <version>${project.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>dev.langchain4j</groupId>
+                <artifactId>langchain4j-mcp</artifactId>
```
geoand, line 218: Maybe it should be `langchain4j-mcp-client`?

jmartisk: That depends, do we plan to have the server module in langchain4j too? I think that would probably be an independent project. Also, the module doesn't just provide the 'client' per se, it also provides a `ToolProvider` based on it (and more stuff in the future)

geoand: > That depends, do we plan to have the server module in langchain4j too?

Very unlikely. 

> Also, the module doesn't just provide the 'client' per se, it also provides a ToolProvider based on it (and more stuff in the future)

Okay, makes sense

langchain4j-mcp/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>0.37.0-SNAPSHOT</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-mcp</artifactId>
+    <name>LangChain4j :: Integration :: MCP client</name>
```
langchain4j:
```suggestion
    <name>LangChain4j :: Integration :: Model Context Protocol</name>
```
- This is not only a client (as you've described above)
- I would use the full name instead of abbreviation

langchain4j-mcp/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
```
langchain4j: nit: not required, transitively loaded from `langchain4j`

langchain4j-mcp/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.jetbrains</groupId>
+                    <artifactId>annotations</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
```
langchain4j, line 74: test scope is missing?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+import okhttp3.sse.EventSource;
+import okhttp3.sse.EventSources;
+
+public class HttpMcpTransport implements McpTransport {
```
geoand, line 24: How would this be overridden in Quarkus?

jmartisk: Quarkus will have its own implementation of `McpTransport` that doesn't depend  on OkHttp. The stdio  transport should work as-is, I assume

geoand: 👍🏽

jmartisk: I tried to only include the okhttp-specific stuff here as much as I could, most reusable stuff is split out into other places

geoand: 👍🏽

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * A tool provider backed by one or more MCP clients
+ */
+public class McpToolProvider implements ToolProvider {
+
+    private final List<McpClient> mcpClients;
+
+    private McpToolProvider(Builder builder) {
+        this.mcpClients = builder.mcpClients;
```
langchain4j: copy the list?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public ToolProviderResult provideTools(final ToolProviderRequest request) {
+        ToolProviderResult.Builder builder = ToolProviderResult.builder();
+        for (McpClient mcpClient : mcpClients) {
+            List<ToolSpecification> toolSpecifications = mcpClient.listTools();
```
langchain4j: What if one of the calls fail? Should we fail completely or continue to call other servers?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -0,0 +1,52 @@
...
+            for (ToolSpecification toolSpecification : toolSpecifications) {
+                builder.add(
+                        toolSpecification, // TODO: how do I handle memory here?
```
langchain4j: Why is memory relevant?

`memoryId` is currently there to distinguish between users. This is not ideal though, and I want to change it to accepts a map of attributes instead of a single memory id. This way user will be able to pass some parameters to the tool. But this is probably not relevant if MCP server does not distinguish between end users?

ghost (commented):
@jmartisk thank you, great job!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+
+        private List<McpClient> mcpClients;
+
+        public McpToolProvider.Builder mcpClients(List<McpClient> mcpClients) {
```
langchain4j, line 62: nit: could add `mcpClients(McpClient... mcpClients)` for convenience

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,104 @@
...
+import dev.langchain4j.mcp.client.protocol.McpCallToolRequest;
+import dev.langchain4j.mcp.client.protocol.McpInitializeRequest;
+import dev.langchain4j.mcp.client.protocol.McpListToolsRequest;
+import dev.langchain4j.mcp.client.transport.McpTransport;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+// TODO: currently we request a new list of tools every time, perhaps we can add some configurable caching
```
langchain4j: I thought that updates to the list of tools will be sent from the server via notifications? So this class should always have an up to date list of tools?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,104 @@
...
+public class DefaultMcpClient implements McpClient {
+
+    private static final Logger log = LoggerFactory.getLogger(DefaultMcpClient.class);
+    private final AtomicLong idGenerator = new AtomicLong(0);
+    private final McpTransport transport;
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    public DefaultMcpClient(Builder builder) {
+        transport = builder.transport;
```
langchain4j:
```suggestion
        transport = ensureNotNull(builder.transport);
```
nit

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,104 @@
...
+        transport.start();
+        McpInitializeRequest request = new McpInitializeRequest(idGenerator.getAndIncrement());
+        InitializeParams params = createInitializeParams();
+        request.setParams(params);
+        JsonNode capabilities = transport.initialize(request);
+        log.info("MCP server capabilities: {}", capabilities.get("result"));
```
langchain4j: nit: should this use `debug` level?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private InitializeParams createInitializeParams() {
+        InitializeParams params = new InitializeParams();
+        params.setProtocolVersion("2024-11-05"); // TODO: what is the correct value?
```
langchain4j: perharps this should be configurable via builder? So that users will be able to modify it if needed

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,104 @@
...
+
+        InitializeParams.ClientInfo clientInfo = new InitializeParams.ClientInfo();
+        clientInfo.setName("langchain4j"); // TODO: this probably should be configurable
```
langchain4j: +1, also for all other properties

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,104 @@
...
+        clientInfo.setVersion("0.1"); // TODO: this probably should be configurable
+        params.setClientInfo(clientInfo);
+
+        InitializeParams.Capabilities capabilities = new InitializeParams.Capabilities();
+        InitializeParams.Capabilities.Roots roots = new InitializeParams.Capabilities.Roots();
+        roots.setListChanged(false); // TODO: listChanged is not supported yet
```
langchain4j, line 62: is this related to the server->cient notificaton mechanism?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
jmartisk: Yeah I wasn't sure about it so I put a TODO there, but probably we can ignore the memory

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
```diff
@@ -0,0 +1,25 @@
+package dev.langchain4j.mcp.client;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+
+public class ToolExecutionHelper {
```
langchain4j:
```suggestion
class ToolExecutionHelper {
```

Or maybe put it in the `internal` sub package if you want to reuse it from Quarkus?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.List;
+import java.util.Map;
+
+class ToolSpecificationHelper {
+
+    /**
+     * Converts the 'tools' element from a ListToolsResult MCP message
+     * to a list of ToolSpecification objects.
+     */
+    public static List<ToolSpecification> toolSpecificationListFromMcpResponse(ArrayNode array) {
```
langchain4j:
```suggestion
    static List<ToolSpecification> toolSpecificationListFromMcpResponse(ArrayNode array) {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * Extracts a single string as a result from the 'content' element of
+     * a CallToolResult message. Only supports String contents, if it finds any
+     * other content types, it throws an exception.
+     */
+    public static String extractResult(ArrayNode contents) {
```
langchain4j:
```suggestion
    static String extractResult(ArrayNode contents) {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
```diff
@@ -0,0 +1,89 @@
...
+        } else if (nodeType.equals("boolean")) {
+            JsonBooleanSchema.Builder builder = JsonBooleanSchema.builder();
+            builder.description(node.get("description").asText());
+            return builder.build();
+        } else if (nodeType.equals("array")) {
+            JsonArraySchema.Builder builder = JsonArraySchema.builder();
+            builder.description(node.get("description").asText());
+            builder.items(jsonNodeToJsonSchemaElement(node.get("items")));
+            return builder.build();
+        } else {
```
langchain4j, line 106: The following types are missing:
- `JsonIntegerSchema`
- `JsonEnumSchema`
- `JsonAnyOfSchema`
- `JsonReferenceSchema`

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+        McpCallToolRequest operation =
+                new McpCallToolRequest(idGenerator.getAndIncrement(), executionRequest.name(), arguments);
+        JsonNode jsonNode = transport.executeTool(operation);
+        return ToolExecutionHelper.extractResult(
+                (ArrayNode) jsonNode.get("result").get("content"));
+    }
+
+    @Override
+    public void close() throws Exception {
```
langchain4j:
```suggestion
    public void close() {
```
nit

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/InitializeParams.java
```diff
@@ -0,0 +1,95 @@
+package dev.langchain4j.mcp.client.protocol;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+public class InitializeParams {
```
langchain4j, line 5: nit: could be a record (all other classes as well)? Just in case: we are finally on Java 17 baseline now

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
jmartisk: Maybe this could be configurable, WDYT?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private final String sseUrl;
+    private final String postUrl;
+    private final OkHttpClient client;
+    private final boolean logResponses;
+    private final boolean logRequests;
```
langchain4j, line 30: nit: `logRequests` can be local var

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+        this.logRequests = builder.logRequests;
+        if (builder.logRequests) {
+            httpClientBuilder.addInterceptor(new McpRequestLoggingInterceptor());
+        }
+        this.logResponses = builder.logResponses;
+        if (builder.logResponses) {
+            httpClientBuilder.addInterceptor(new McpResponseLoggingInterceptor());
+        }
+        sseUrl = builder.sseUrl;
```
langchain4j: nit: would be nice to add `ensureNotNull`/`ensureNotBlank`/etc for each required field to fail fast and to "document" what is mandatory and what is not

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+    }
+
+    private JsonNode executeAndWait(Request request, Long id) {
+        try {
+            CompletableFuture<JsonNode> future = new CompletableFuture<>();
+            pendingOperations.put(id, future);
+            try (final Response response = client.newCall(request).execute()) {
+                int statusCode = response.code();
+                if (statusCode != 202) {
```
langchain4j: Are we sure all servers will always respond 202 (and not 200, for example)? Is it a requirement of the protocol? Can't find this on https://modelcontextprotocol.io/

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/McpResponseLoggingInterceptor.java
```diff
@@ -0,0 +1,52 @@
...
+
+class McpResponseLoggingInterceptor implements Interceptor {
+
+    private static final Logger log = LoggerFactory.getLogger(McpResponseLoggingInterceptor.class);
+
+    @Override
+    public Response intercept(Chain chain) throws IOException {
+        Request request = chain.request();
+        Response response = chain.proceed(request);
+        if (response.code() != 200) {
```
langchain4j: Should we check for 202 as well? Perharps we should log it in any case?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/SseEventListener.java
```diff
@@ -0,0 +1,64 @@
...
+                    log.warn("Received response for unknown message id: {}", messageId);
+                }
+            } catch (JsonProcessingException e) {
+                log.warn("Failed to parse response data", e);
+            }
+        }
+    }
+
+    @Override
+    public void onFailure(@NotNull final EventSource eventSource, final Throwable t, final Response response) {
```
langchain4j: do you mind using `@NonNull` from JSpecify instead?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/SseEventListener.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public void onClosed(EventSource eventSource) {
+        log.debug("SSE channel closed");
+    }
+
+    @Override
+    public void onEvent(EventSource eventSource, String id, String type, String data) {
+        if (type.equals("message")) {
+            if (logEvents) {
+                log.info("< {}", data);
```
langchain4j: nit: debug level?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/ProcessErrorPrinter.java
```diff
@@ -0,0 +1,30 @@
+package dev.langchain4j.mcp.client.transport.stdin;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ProcessErrorPrinter implements Runnable {
```
langchain4j:
```suggestion
class ProcessErrorPrinter implements Runnable {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdinMcpTransport.java
```diff
@@ -0,0 +1,132 @@
...
+import dev.langchain4j.mcp.client.transport.McpTransport;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StdinMcpTransport implements McpTransport {
```
langchain4j: This class is not used anywhere, is this expected?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/ProcessIOHandler.java
```diff
@@ -0,0 +1,64 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class ProcessIOHandler implements Runnable {
```
langchain4j: This seems to do the same thing as `ProcessErrorPrinter`, but for input stream. Should they both have similar naming convention?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/ProcessIOHandler.java
```diff
@@ -0,0 +1,64 @@
...
+        this.logEvents = logEvents;
+    }
+
+    @Override
+    public void run() {
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                if (logEvents) {
+                    log.info("< {}", line);
```
langchain4j: nit: debug?

langchain4j: could you please change all instances of `log.info(...)` in this module to `log.debug(...)`?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdinMcpTransport.java
```diff
@@ -0,0 +1,132 @@
...
+            return this;
+        }
+
+        public Builder logEvents(boolean logEvents) {
+            this.logEvents = logEvents;
+            return this;
+        }
+
+        public StdinMcpTransport build() {
+            if (command == null || command.isEmpty()) {
```
langchain4j: curious: I've noticed that you often check invariants in the builders. Are there benefits for doing this in builders instead of constructors?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
jmartisk: I made this configurable

ghost (commented):
@jmartisk thank you!

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
```diff
@@ -0,0 +1,123 @@
...
+    GenericContainer<?> container = new GenericContainer<>("docker.io/node:23")
+            .withCopyFileToContainer(MountableFile.forClasspathResource("everything"), "/everything")
+            .withLogConsumer(outputFrame -> System.out.println(outputFrame.getUtf8String()))
+            .withWorkingDirectory("/everything")
+            .withCommand("sh", "-c", "npm install && npm run build && node dist/sse.js")
+            .withExposedPorts(3001)
+            .waitingFor(Wait.forHttp("/").forStatusCode(404));
+
+    @Test
+    public void test() throws Exception {
```
langchain4j: minor: this test could be simplified a lot by using AI Service and configuring `McpToolProvider` on it

jmartisk: I know and I considered that first, but this gives you more finer-grained insight into what happens, so I went this way

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdinMcpTransport.java
jmartisk: Ah ok, I see other code does that in constructors, I'll redo this. I guess it's mostly a personal preference, but I'll adjust

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/ProcessIOHandler.java
jmartisk: Even when this is something that was explicitly enabled by configuration? In that case I think it should be INFO, no?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdinMcpTransport.java
jmartisk: I haven't written a test that uses it, that's why. (I haven't figured out how to do it in a way that will run reliably in every environment). It is used in one of the examples in the example repo though.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdinMcpTransport.java
langchain4j: Thanks! This is definitely not critical, but would be nice to have consistency

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/ProcessIOHandler.java
langchain4j: Hmm, I am not sure, maybe you are right.

We are usually using `debug` level, does not matter if enabled or disabled.
`logRequests`/`logResponses` enables logging, I am not sure it should affect the level of logging though.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: I haven't implemented tool updates yet, I'll probably add that tomorrow, so just for now I'm assuming that we always need to get a new list of tools for every prompt

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
jmartisk: I've now changed it to wrap it in `Collections.unmodifiableList`, do we need a full copy?!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: Yes, and I will implement that soon

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
jmartisk: Right now I don't think Quarkus will need it, so ok I'll make it package-private

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/SseEventListener.java
jmartisk: Hm my IDE generated this (the superclass is in Kotlin), I think we don't need it at all, do we..

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
langchain4j: By full do you mean deep? I am not sure. In theory user can manipulate `McpClient`: call `listTools`/`executeTool`/`close`. Can this interfere if the same `McpClient` will be used simultaneously by user and `McpToolProvider`? I guess it can, but should we care?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/ProcessIOHandler.java
jmartisk: Looking at `OllamaResponseLoggingInterceptor`, for example, it uses `debug`... So ok, I'll adjust for consistency.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/SseEventListener.java
langchain4j: Not really, this will be called by okhttp, so why should we annotate?

But if we use those kind of annotations, let's use JSpecify which we just started using recently

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: It could, but I think it makes sense to leave it here, as we may find more use for it later

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: The servers I've tried so far all return 202, and the protocol doesn't specify. Maybe let's keep it like this for now and adjust if some servers later return something else... (but 202 makes most sense)

langchain4j: If specification does not specify this (which is strange?) people will start returning 200 sooner or later. Maybe let's check for 200-299 range? Easy to do this now, will not need to "fix" this later

langchain4j: BTW why 202 makes more sense then 200?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/McpResponseLoggingInterceptor.java
jmartisk: This only applies to the initial GET request that starts the SSE channel, and that should return 200, as far as I know, which means that the SSE channel was successfully created. Otherwise, it's an error that should be logged.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
jmartisk: Yeah that's another TODO for tomorrow...

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/McpResponseLoggingInterceptor.java
jmartisk: Ok I'm now realizing that maybe this `McpResponseLoggingInterceptor` isn't necessary at all then, because SSE events are logged anywhere, and if the SSE channel creation fails, the execution doesn't even get here, it ends up in `SseEventListener.onFailure` already

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
jmartisk: Ok I'll do `this.mcpClients = new ArrayList<>(builder.mcpClients);` here just to be sure

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/InitializeParams.java
jmartisk: Hmmm... The problem is that I'm using inheritance (`extends McpClientMessage`) which you can't do with records. Not sure how I would replace the inheritance, maybe I just don't know enough about how records work

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: 202 is described as

>The request has been accepted for processing, but the processing has not been completed. 

Which is what this does - the server has accepted something to be processed asynchronously later

jmartisk: But ok, if we want to be sure, we could also consider 200 and 204 fine? (but I don't think any other codes beyond these)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/McpResponseLoggingInterceptor.java
langchain4j: Can't comment as I am not as deep into it...

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/InitializeParams.java
langchain4j: this was a nitpick, no problem at all with using class instead of record

langchain4j: https://stackoverflow.com/questions/63605794/is-there-any-way-of-using-records-with-inheritance

> The Java 17 [JLS 8.10](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.10) notes1 this:
> 
> A record declaration does not have an extends clause, so it is not possible to explicitly declare a direct superclass type, even Record.
> 
> However, a record can implement interfaces, so you can use them polymorphically. Furthermore, since records will inherit any default methods in the interfaces that they implement, they do support a limited form of inheritance.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
langchain4j: OK, although we could just do `if (statusCode < 200 || statusCode >= 300)` as well?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
jmartisk: Hmm, actually, looking at the specification schema (https://github.com/modelcontextprotocol/specification/blob/main/schema/schema.json), I think the only type missing that might appear there is `integer`

langchain4j: Probably enum as well, in JSON schema it is basically `type=string` and `enum=[valiue1, value2]`

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: Ok then, will do.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: One problem is that I don't know of any MCP server that actually does this - do you know any? If there's none, I don't have any way to test the implementation and it may not make much sense to do (yet)

langchain4j: I don't know any, but since adding/changing tools usually require writing code, I am not sure it is very useful. I can imagine new versions of MCP servers having more tools, but that will require restarting the connection anyway? Perhaps notifications are more useful for resources/data then for tools

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
jmartisk: Ah right, thanks, will add

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: A more advanced server could potentially serve tools that are dynamically added and removed and it wouldn't require restarting,  but I don't know if that's going to become a normal thing. Maybe we could leave this for now and see how stuff pans out later.

langchain4j: Makes sense

jmartisk:
I've added a piece of documentation now

langchain4j-mcp/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+            <artifactId>okhttp</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp-sse</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
```
kpavlov:
```suggestion
            <artifactId>junit-jupiter-api</artifactId>
```

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/ToolSpecificationHelperTest.java
```diff
@@ -0,0 +1,189 @@
...
+                        },
+                        "required" : [ "stringParameter" ],
+                        "additionalProperties" : false,
+                        "$schema" : "http://json-schema.org/draft-07/schema#"
+                      }
+                    } ]
+                """;
+        ArrayNode json = OBJECT_MAPPER.readValue(text, ArrayNode.class);
+        List<ToolSpecification> toolSpecifications = ToolSpecificationHelper.toolSpecificationListFromMcpResponse(json);
+        assertEquals(1, toolSpecifications.size());
```
kpavlov: nit: IMO assertJ assertions are more expressive

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.Set;
+import java.util.stream.Collectors;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.utility.MountableFile;
+
+@EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = ".+")
```
kpavlov: Why McpHttpTransport test needs AI model? What is the system under test? Integration test should have IT suffix.

The project already has way too many integration tests that are not being executed for PRs, because they are slow and expensive. 

Would you mind to keep this test only focusing on McpHttpTransport testing? LLM integration is tested somewhere else, as long as McpToolProivder produce tools complying with the ToolSpecification

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
```diff
@@ -0,0 +1,132 @@
...
+import dev.langchain4j.mcp.client.transport.McpTransport;
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StdioMcpTransport implements McpTransport {
```
kpavlov: Is there a test for ToolSpecification? Would it make sense to introduce this class along with the test in a separate PR to keep PRs small?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
```diff
@@ -0,0 +1,132 @@
...
+    public void close() throws IOException {
+        process.destroy();
+    }
+
+    private JsonNode executeAndWait(String request, Long id) {
+        try {
+            CompletableFuture<JsonNode> future = new CompletableFuture<>();
+            pendingOperations.put(id, future);
+            processIOHandler.submit(request);
+            return future.get();
```
kpavlov: no timeout

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
```diff
@@ -0,0 +1,132 @@
...
+        } catch (Exception e) {
+            throw new RuntimeException(e);
```
kpavlov: Wouldn't it be good to log something in case of error?

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
```diff
@@ -0,0 +1,123 @@
...
+            .withExposedPorts(3001)
+            .waitingFor(Wait.forHttp("/").forStatusCode(404));
+
+    @Test
+    public void test() throws Exception {
+        container.start();
+        Integer port = container.getMappedPort(3001);
+        McpClient mcpClient = null;
+        try {
+            OpenAiChatModel chat = OpenAiChatModel.builder()
```
kpavlov: Would ChatModelMock work for the test as well?

kpavlov (commented):
Thank you, @jmartisk, this is cool!

Would it be possible to add missing unit tests and to switch from real LLM models to dummy ones for integration tests?
Please don't hesitate to slice this PR into parts if it starts growing too much.

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
jmartisk: I definitely could replace the openai model with a mock, it will just be a bunch of work, I thought it easier to just use the real model... Wdyt @langchain4j ?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
jmartisk: This exception will bubble up to the application into the `DefaultMcpClient` constructor, so I'd say it's not necessary?!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
jmartisk: That's a bit tricky because the user might not know how long a tool will take to finish... 
But right, maybe we could implement a timeout for tool executions, and send the server a cancellation message if the timeout is reached! I think that makes sense

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
jmartisk: Hm, `ChatModelMock` only serves a single hardcoded response, and I need it to request tools etc.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
jmartisk: Not sure what you mean, I have `ToolSpecificationHelperTest` that verifies creating `ToolSpecification` from MCP protocol messages

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
langchain4j: Yeah I agree, no need to use real model here, you could either mock `ChatLanguageModel` interface or add into `ChatModelMock` a `thatAlwaysResponds(AiMessage)` method that will allow you to specify tools

jmartisk: I have an idea and I will rewrite it so that it doesn't need a chat model at all.

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/ToolSpecificationHelperTest.java
jmartisk: Ok, I rewrote them (I generally prefer pure JUnit, but it may just be a force of habit..)

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
jmartisk: Done, please check this test again

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
jmartisk: I rewrote it to not use a chat model at all

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/ToolSpecificationHelperTest.java
```diff
@@ -0,0 +1,189 @@
...
+import org.junit.jupiter.api.Test;
+
+public class ToolSpecificationHelperTest {
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    @Test
+    public void toolWithSimpleParams() throws JsonProcessingException {
+        String text =
+                """
```
kpavlov:
```suggestion
        // language=json
        String text =
                """
```
hint for IDEA :)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
kpavlov: Depending on if we want to see something in the log from StdioMcpTransport. I'm fine to skip it

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
kpavlov: Java structured concurrency from JDK 21 preview would fit well here. Meanwhile, would it make sense to add timeout as a configuration parameter for the transport and set some meaningful default value, like 5 seconds?  Or keep it optional (nullable), but have an option for the user to set it.
nit: I would love to see JSpecify nullability annotations so the contract is more clear for the developer and the compiler

jmartisk: Yeah unfortunately we're not on Java 21. Also, I think configurable tool timeouts are something that we should probably add in general, also for locally executed tools. So maybe it's something that we should keep as a separate task after all

docs/docs/tutorials/mcp.md
```diff
@@ -0,0 +1,80 @@
...
+    .logEvents(true) // only if you want to see the traffic in the log
+    .build();
+```
+
+For HTTP, you need two URLs, one for starting the SSE channel and one for submitting commands via `POST`:
+
+```java
+McpTransport transport = new HttpMcpTransport.Builder()
+    .sseUrl("http://localhost:3001/sse")
+    .postUrl("http://localhost:3001/message")
```
langchain4j, line 38: nit: is there a reson to have both `sseUrl` and `postUrl`? Can they be different? Perhaps it could be simplified to a single `baseUrl` (for example, "http://localhost:3001") and then `/sse` and `/message` parts are added internally?

docs/docs/tutorials/mcp.md
```diff
@@ -0,0 +1,80 @@
...
+```java
+ToolProvider toolProvider = McpToolProvider.builder()
+    .mcpClients(List.of(mcpClient))
+    .build();
+```
+
+Note that one MCP tool provider can use multiple clients at the same time.
+If you make use of this, you can also specify the behavior of the tool provider
+in cases when retrieving tools from a particular server fails - this is done
+by the `builder.failIfOneServerFails(boolean)` method. The default is `false`,
```
langchain4j, line 67: minor: it feels like it should be an enum that will define what to do on failure. Boolean has only 2 options, with enum we could add more

langchain4j-core/src/test/java/dev/langchain4j/model/chat/mock/ChatModelMock.java
```diff
@@ -46,6 +58,28 @@ public Response<AiMessage> generate(List<ChatMessage> messages) {
...
+    @Override
+    public Response<AiMessage> generate(List<ChatMessage> messages,
+                                        List<ToolSpecification> toolSpecifications) {
+        requests.add(new ArrayList<>(messages));
+
+        if (exception != null) {
+            throw exception;
+        }
+
+        if (toolExecutionRequest == null) {
```
langchain4j:
```suggestion
        if (toolExecutionRequest != null) {
```
I guess it was supposed to be `!= null`?

langchain4j-core/src/test/java/dev/langchain4j/model/chat/mock/ChatModelMock.java
```diff
@@ -46,6 +58,28 @@ public Response<AiMessage> generate(List<ChatMessage> messages) {
         return Response.from(AiMessage.from(staticResponse));
```
langchain4j: nit: `Response<AiMessage> generate(List<ChatMessage> messages)` could now call the new method to avoid duplication

langchain4j-core/src/test/java/dev/langchain4j/model/chat/mock/ChatModelMock.java
```diff
@@ -64,6 +98,10 @@ public String userMessageText() {
         return message.text();
     }
 
+    public static ChatModelMock thatRequestsATool(String toolName) {
+        return new ChatModelMock("This is a placeholder response for tool: " + toolName);
```
langchain4j: This seems wrong, it is also not used

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -0,0 +1,87 @@
...
+        ToolProviderResult.Builder builder = ToolProviderResult.builder();
+        for (McpClient mcpClient : mcpClients) {
+            try {
+                List<ToolSpecification> toolSpecifications = mcpClient.listTools();
+                for (ToolSpecification toolSpecification : toolSpecifications) {
+                    builder.add(
+                            toolSpecification, (executionRequest, memoryId) -> mcpClient.executeTool(executionRequest));
+                }
+            } catch (Exception e) {
+                log.warn("Failed to retrieve tools from MCP server", e);
```
langchain4j: nit: could you please put `log.warn(...)` into the `else` block? Otherwise the error will be duplicated in logs

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,149 @@
...
+    private static final Logger log = LoggerFactory.getLogger(DefaultMcpClient.class);
+    private final AtomicLong idGenerator = new AtomicLong(0);
+    private final McpTransport transport;
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private final String clientName;
+    private final String clientVersion;
+    private final String protocolVersion;
+
+    public DefaultMcpClient(Builder builder) {
+        transport = ensureNotNull(builder.transport, "transport must not be null");
```
langchain4j:
```suggestion
        transport = ensureNotNull(builder.transport, "transport");
```
nit, "cannot be null" will be added automatically

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -0,0 +1,149 @@
...
+        clientName = getOrDefault(builder.clientName, "langchain4j");
+        clientVersion = getOrDefault(builder.clientVersion, "1.0");
```
langchain4j, line 39: minor: perhaps this should be the current LC4j version? We could get it from maven?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Obtains a list of tools from the MCP server.
+     */
+    List<ToolSpecification> listTools();
+
+    /**
+     * Executes a tool on the MCP server and returns the result as a String.
+     * Currently, this expects a tool execution to only contain text-based results.
+     */
+    String executeTool(ToolExecutionRequest executionRequest);
```
langchain4j, line 22: minor: it would be good to future-proof and return an object (currently with a single `String` inside), but I doubt anyone will implement a custom one, so we can do it later

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
```diff
@@ -0,0 +1,25 @@
...
+     * Extracts a single string as a result from the 'content' element of
+     * a CallToolResult message. Only supports String contents, if it finds any
+     * other content types, it throws an exception.
+     */
+    static String extractResult(ArrayNode contents) {
+        StringBuilder result = new StringBuilder();
+        for (JsonNode content : contents) {
+            String contentType = content.get("type").asText();
+            if (contentType.equals("text")) {
+                result.append(content.get("text").asText()).append("\n");
```
langchain4j: nit: last `\n` could be removed

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
langchain4j, line 22: API discussion/question: what if LLM responds with multiple `ToolExecutionRequest`?

From the protocol it seems that [only a single tool can be called at a time](https://spec.modelcontextprotocol.io/specification/server/tools/#calling-tools), is it correct?
Do servers support concurrent requests?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/ProcessIOHandler.java
```diff
@@ -0,0 +1,64 @@
+package dev.langchain4j.mcp.client.transport.stdin;
```
langchain4j:
```suggestion
package dev.langchain4j.mcp.client.transport.stdio;
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
```diff
@@ -0,0 +1,132 @@
...
+
+    public StdioMcpTransport(Builder builder) {
+        this.command = builder.command;
+        this.environment = builder.environment;
+        this.logEvents = builder.logEvents;
+    }
+
+    @Override
+    public void start() {
+        log.info("Starting process: {}", command);
```
langchain4j:
```suggestion
        log.debug("Starting process: {}", command);
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/ProcessStderrHandler.java
```diff
@@ -0,0 +1,30 @@
...
+    public ProcessStderrHandler(final Process process) {
+        this.process = process;
+    }
+
+    @Override
+    public void run() {
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getErrorStream()))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                log.info("[ERROR] {}", line);
```
langchain4j:
```suggestion
                log.debug("[ERROR] {}", line);
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
langchain4j: Agreed, we can do that separately, although I would at least hardcode some reasonable (e.g., 60 seconds) timeout to make sure it won't hang forever in corner cases. But then adding a user-configured value would also be quite easy, so maybe let's do it right away?

And the same for `HttpMcpTransport`

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/transport/http/McpHttpTransportTest.java
```diff
@@ -0,0 +1,101 @@
...
+                McpToolProvider.builder().mcpClients(List.of(mcpClient)).build();
+        ToolProviderResult toolProviderResult = toolProvider.provideTools(null);
+
+        // verify the list of tools
+        assertThat(toolProviderResult.tools().size()).isEqualTo(5);
+        Set<String> toolNames = toolProviderResult.tools().keySet().stream()
+                .map(ToolSpecification::name)
+                .collect(Collectors.toSet());
+        assertThatIterable(toolNames)
+                .containsExactlyInAnyOrder("add", "echo", "longRunningOperation", "sampleLLM", "getTinyImage");
```
langchain4j, line 78: Could you please add assertions for `ToolSpcecification` internals?

langchain4j-core/src/test/java/dev/langchain4j/model/chat/mock/ChatModelMock.java
```diff
@@ -22,15 +25,24 @@
 public class ChatModelMock implements ChatLanguageModel {
 
     private final String staticResponse;
+    private final ToolExecutionRequest toolExecutionRequest;
     private final RuntimeException exception;
     private final List<List<ChatMessage>> requests = synchronizedList(new ArrayList<>());
 
+    public ChatModelMock(ToolExecutionRequest toolExecutionRequest) {
```
langchain4j: this is not used, should we remove all the changes from this class?

ghost (commented):
@jmartisk thanks a lot!

docs/docs/tutorials/mcp.md
jmartisk: The problem is that the specification doesn't describe this, so I think we need to assume it may differ depending on the server

docs/docs/tutorials/mcp.md
jmartisk: Maybe, but what third option would there be? If we come up with something reasonable, then sure, but nothing is coming to my mind, otherwise it feels unnecessary

docs/docs/tutorials/mcp.md
langchain4j: Do you mean the base url might differ (sounds unlikely) or the path (/sse /message)?

langchain4j-core/src/test/java/dev/langchain4j/model/chat/mock/ChatModelMock.java
jmartisk: Uhm, you're right, will remove

jmartisk: Actually all my changes in this class are unnecessary, I forgot to undo  it when I removed usage of chat models from my test :)

docs/docs/tutorials/mcp.md
langchain4j: A few option from the top of my head:
- fail on the first failure (exists)
- ignore failure, continue with other servers (exists)
- fail if N servers failed
- retry on failure, fail if unsuccessful
- retry on failure, continue with other servers if unsuccessful
- etc etc

But perhaps this is not required at all since servers will notify about changes, so there will be no need in this proactive requesting?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: That would probably require reading it from a manifest file or something, but what if somebody removes it for some reason (slimming down an app, whatever. Also native mode might be a problem)... It feels brittle. Also I think this should generally refer to the version of the application, not the underlying framework

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
jmartisk: Yeah it seems the protocol can handle only one tool at a time. But the framework generally handles this case by executing tools sequentially, so it should be fine, no?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
jmartisk: It's not that trivial because we should also send a cancellation message to the server that it can stop execution of this tool (the protocol supports this). Let's keep it as a follow-up

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
langchain4j: Current implementation (sequential tool execution) in AI Services is not ideal as tool calls are independent in many cases, so they can be parallelized.
From the protocol perspective, I could imagine that having a single HTTP call with multiple tools is more efficient than multiple sequential round trips or multiple HTTP calls in parallel

docs/docs/tutorials/mcp.md
jmartisk: The paths

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
langchain4j: What will happen if we don't send the cancellation message? Do servers support "rollback" functionality?

docs/docs/tutorials/mcp.md
jmartisk: If we want to support retries, than enum won't be very suitable either... That would need a full-blown Strategy class

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdin/StdioMcpTransport.java
jmartisk: I think nothing will really happen, except that the server will keep processing something that we don't need (or may be deadlocked), but I'd like to implement the cancellation properly

ghost (approved):
@jmartisk thanks a lot, great job!

PR #1989, opened 2024-10-25
New ChatLanguageModel API

ghost:
## Issue
Closes #837
Closes #961
Fixes #2120
Fixes #2289

## Motivation

The existing `ChatLanguageModel` and `StreamingChatLanguageModel` APIs:
- are not flexible enough to support certain use cases (e.g., structured outputs, specifying chat model parameters dynamically etc.)
- limit the potential to add new features going forward

## Change

ChatModelAPI2 drawio

## New `ChatLanguageModel` API

The `ChatLanguageModel` has new methods now, which should replace the existing `generate()` methods:
- `ChatResponse chat(ChatRequest chatRequest)`
- `String chat(String userMessage)`
- `ChatRequestParameters defaultRequestParameters()`
- `Set<Capability> supportedCapabilities()`

The `chat(ChatRequest)` method is now the main API to interact with chat models. `ChatRequest` and `ChatResponse` will allow to add new request parameters and response metadata in order to support more LLM features now and in the future.
The `chat(ChatRequest)` method has a default implementation (temporarily) that allows existing `ChatLanguageModel` integrations to gradually transition from the old to the new API.

The `chat(String)` method is a convenience method which is useful for sending a single text user message to the model and getting a single text response from it.

The existing `generate()` methods will be deprecated and will be removed before 1.0 release.

### `ChatRequest`

The `ChatRequest` now consists of:
- `List<ChatMessage> messages`
- `ChatRequestParameters parameters`

## `ChatRequestParameters`

The `ChatRequestParameters` interface was introduced to represent all common chat request parameters that are supported by most LLM APIs:
- `modelName`
- `temperature`
- `topP`
- `topK`
- `frequencyPenalty`
- `presencePenalty`
- `maxOutputTokens`
- `stopSequences`
- `toolSpecifications`
- `toolChoice`
- `responseFormat`

The `ChatRequestParameters` can be specified:
- when creating the `ChatLanguageModel` object, in this case these parameters are considered "default"
- when calling the `ChatLanguageModel.chat(ChatRequest)` method, in this case they override/merge with the "default" parameters (if default parameters are configured)

Here is an example of specifying default parameters when creating the `ChatLanguageModel` object:
```java
ChatRequestParameters parameters = ChatRequestParameters.builder()
        .modelName("gpt-4o-mini")
        .temperature(0.7)
        .build();

OpenAiChatModel model = OpenAiChatModel.builder()
        .apiKey(System.getenv("OPENAI_API_KEY"))
[...]

PR #2242, opened 2024-12-05
Add ModelAware interface

alesj:
Would something like this make sense?
If yes, it can be added on all model impls that support this ...

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/ModelAware.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/ModelAware.java
@@ -1,0 +1,9 @@
+package dev.langchain4j.model;
+
+/**
+ * A simple interface marking the model implementation being model aware; e.g. name.
+ * This exposes some of the implementation details in more type-safe way.
+ */
+public interface ModelAware {
+    String modelName();
+}
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
@@ -9,6 +9,7 @@ import dev.ai4j.openai4j.chat.ResponseFormatType;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.model.ModelAware;
 import dev.langchain4j.model.Tokenizer;
 import dev.langchain4j.model.chat.Capability;
 import dev.langchain4j.model.chat.ChatLanguageModel;
@@ -62,7 +63,7 @@ import static java.util.Collections.singletonList;
  * Represents an OpenAI language model with a chat completion interface, such as gpt-3.5-turbo and gpt-4.
  * You can find description of parameters <a href="https://platform.openai.com/docs/api-reference/chat/create">here</a>.
  */
-public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
+public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator, ModelAware {
 
     private static final Logger log = org.slf4j.LoggerFactory.getLogger(OpenAiChatModel.class);
     private final OpenAiClient client;


alesj:
@geoand ^ ?

geoand:
I think it makes sense, WDYT @langchain4j ?

langchain4j:
Totally makes sense, hovewer I was considering exposing not only model name, but also other parameters (e.g., `temperature`, `maxOutputTokens`, etc. All of parameters listed [here](https://github.com/langchain4j/langchain4j/pull/1989/files#diff-0eccc7b5695d9afff5fe62c8faba2e7994cdb4f6f9750d88877799827c6308f0R21) in `ChatRequest` class) that can be configured when `ChatLanguageModel` is created. WDYT?

glaforge:
I agree we should also standardize on some other parameters, to have consistent naming conventions for the builders instantiating our models. 
Should we have one interface, or maybe even one for each parameter? Say, interfaces like `HasModelName`, `HasTopP`, etc?

langchain4j:
Yes, maybe something like:
```java
public interface ChatParameters {

    String modelName();
    Double temperature();
    ...
}

public interface ChatLanguageModel {

    ...
    
    ChatParameters parameters();
}

public interface ChatRequest {

    ...
    
    ChatParameters parameters();
}
```

langchain4j:
These model parameters will be required (as a group) in multiple cases:
- Default model parameters:
  - specify default model parameters when creating `ChatModel` instance (what @glaforge is suggesting, to standartize it)
  - specify default model parameters when creating `ChatModel` with one's custom model classes (discussed today with @mariofusco)
  - get default model parameters from `ChatModel` instance (addressed by this PR)
- Request-specific model parameters:
  - specify request-specific model parameters when creating `ChatRequest` with builder
  - specify request-specific model parameters when creating `ChatRequest` with one's custom model classes (discussed today with @mariofusco)
  - get request-specific model parameters from `ChatRequest` instance

This is also required for implementing `ObservableChatModel` (to have a common observability logic for all models).

@alesj I hope this PR is not urgent? I would like to take a bit of time (a few days) to play with this idea in scope of https://github.com/langchain4j/langchain4j/pull/1989 and see how to better implement it so that there is a uniform API for all those cases mentioned above.

Thanks for the very timely PR BTW :)

alesj:
@langchain4j nah no rush.
I just wanted to get this rolling, as I think it was useful.

e.g. what about also (Chat)Response (or any type of response, as there are already a few) having this as well?

As this would ease things with something like
* https://github.com/quarkiverse/quarkus-langchain4j/pull/1138
(this is where the need for this PR came from ...)

langchain4j:
@alesj could you please clarify regarding response/token usage?
Right now `ChatResponse` returns `TokenUsage` which has unified input/output tokens.

Regarding `Response` classes:
- `Response<T>` will be removed soon, each type of model (e.g., chat, embedding, image, etc) will return it's own response type.
- `ChatModelResponse` will be removed as well, `ChatModelListener` will operate with `ChatResponse`.

So at the end there will be a single response class for each type of model: `ChatResponse`, `EmbeddingResponse`, `ImageResponse`, etc.

alesj:
> @alesj could you please clarify regarding response/token usage?

I would also need a model type in the response -- which is only available atm in `ChatModelResponse` (all other response types currently don't have it).
(other params - from the discussion above - might also come in handy sometime in the future, you never know :-)

langchain4j:
@alesj 

> I would also need a model type in the response -- which is only available atm in ChatModelResponse (all other response types currently don't have it).
> (other params - from the discussion above - might also come in handy sometime in the future, you never know :-)

By "model type in the response" do you mean model name/id (e.g., `gpt-4o-mini-2024-07-18`)?
`ChatResponse` will have `modelName` after https://github.com/langchain4j/langchain4j/pull/1989 is merged. Every `ChatLanguageModel` implementation will eventually provide it if it is available in the API response.

alesj:
> By "model type in the response" do you mean model name/id (e.g., gpt-4o-mini-2024-07-18)?

Yes.
Something I can match from their pricing website. :-)

langchain4j:
@alesj please check the description of https://github.com/langchain4j/langchain4j/pull/1989, everything that you've described should be available in the new API.
Thus I am closing this PR. Thank you for opening this discussion, it helped a lot!

PR #2248, opened 2024-12-06
draft: fix OpenAiEmbeddingModel request text array overflow

### LINKED ISSUE #2247: [BUG] OpenAiEmbeddingModel text array limit
**Describe the bug**
When attempting to embed more than 2048 segments using OpenAI's text-embedding-3-large embedding model, the process fails because OpenAI's API can only handle up to 2048 tokens per request.

**Log and Stack trace**
I don't have one, but it throws default openai bad request with url to read api docs

**To Reproduce**

1. Use LangChain4j with OpenAI text-embedding-3-large
2. Attempt to embed more than 2048 segments at once
3. Observe the error due to the text array limit

**Expected behavior**

1. Splitting the input into chunks of 2048 texts or less
2. Processing each chunk separately.
3. Combining the results as needed.

- LangChain4j version: Langchain 0.35.0
- LLM(s) used: OpenAI text-embedding-3-large
- Java version: 21
- Spring Boot version (if applicable): 3.3.5 

**Additional context**
I am going to make pull request for that

Empatixx:
## Issue
Fixes #2247

## General checklist


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
```diff
@@ -102,7 +104,39 @@ public Response<List<Embedding>> embedAll(List<TextSegment> textSegments) {
                 .map(TextSegment::text)
                 .toList();
 
-        return embedTexts(texts);
+        List<List<String>> textBatches = splitList(texts, REQUEST_TEXT_MAX_ARRAY_SIZE);
+
+        return embedBatchedTexts(textBatches);
+    }
+
+    public List<List<String>> splitList(final List<String> inputList, final int size) {
```
omarmahamid: do you really need this as public method ?

Empatixx: Thank you, I didn't notice, updated to private.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
```diff
@@ -32,10 +34,10 @@
  * Represents an OpenAI embedding model, such as text-embedding-ada-002.
  */
 public class OpenAiEmbeddingModel extends DimensionAwareEmbeddingModel implements TokenCountEstimator {
-
+    private static final int REQUEST_TEXT_MAX_ARRAY_SIZE = 2048;
     private final OpenAiClient client;
     private final String modelName;
-    private final Integer dimensions;
+    private final Integer dimensions;;
```
omarmahamid: Remove duplicate semicolon

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
```diff
@@ -32,7 +34,7 @@
  * Represents an OpenAI embedding model, such as text-embedding-ada-002.
  */
 public class OpenAiEmbeddingModel extends DimensionAwareEmbeddingModel implements TokenCountEstimator {
-
+    private static final int REQUEST_TEXT_MAX_ARRAY_SIZE = 2048;
```
langchain4j: It would also be great to make it configurable, so that user could override this default in the builder:

```java
OpenAiEmbeddingModel.builder()
    ...
    .maxSegmentsPerBatch(1000)
    .build();
```

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
```diff
@@ -102,7 +104,39 @@ public Response<List<Embedding>> embedAll(List<TextSegment> textSegments) {
...
 
-        return embedTexts(texts);
+        List<List<String>> textBatches = splitList(texts, REQUEST_TEXT_MAX_ARRAY_SIZE);
+
+        return embedBatchedTexts(textBatches);
+    }
+
+    private List<List<String>> splitList(final List<String> inputList, final int size) {
+        if (size <= 0) {
+            throw new IllegalArgumentException("Size must be greater than 0");
```
langchain4j: If this value will be configurable, it would be nice to do this validation in the constructor, to fail fast:

`ensureGreaterThanZero(maxSegmentsPerBatch, "maxSegmentsPerBatch")`

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
```diff
@@ -102,7 +104,39 @@ public Response<List<Embedding>> embedAll(List<TextSegment> textSegments) {
                 .map(TextSegment::text)
                 .toList();
 
-        return embedTexts(texts);
+        List<List<String>> textBatches = splitList(texts, REQUEST_TEXT_MAX_ARRAY_SIZE);
+
+        return embedBatchedTexts(textBatches);
+    }
+
+    private List<List<String>> splitList(final List<String> inputList, final int size) {
```
langchain4j:
```suggestion
    private List<List<String>> partition(final List<String> inputList, final int size) {
```
nit

ghost (commented):
@Empatixx thanks a lot!

Could you please add a few tests to the `OpenAiEmbeddingModelIT`?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
Empatixx: I added a way to configure the segment range. The range is set to <1, 2048> based on OpenAI's documentation, which states: "Cannot be an empty string, and any array must be 2048 dimensions or less."
[OpenAI API Reference](https://platform.openai.com/docs/api-reference/embeddings/create)
And also I added missing test

langchain4j: @Empatixx I am not sure we should restrict users to 2048 max. What happens if OpenAI lifts this requirement? What happens if users will use `OpenAiEmbeddingModel` to connect to other OpenAI-compatible embedding APIs that support higher number of parallel embeddings?
I guess checking that the number is greater than 0 is enough

langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiEmbeddingModelIT.java
```diff
@@ -71,6 +73,32 @@ void should_embed_multiple_segments() {
         assertThat(response.finishReason()).isNull();
     }
 
+    @Test
+    void should_embed_multiple_batch_segments() {
+
+        // given
+        List<TextSegment> segments = Stream.generate(() -> TextSegment.from("hello world"))
+                .limit(50)
```
langchain4j: nit: Could you please create a specific `OpenAiEmbeddingModel` inside this test, otherwise it is not obvious how 50 relates to 10

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiEmbeddingModel.java
Empatixx: Understood, updated accordingly.

ghost (approved):
@Empatixx thanks a lot!

Could you please update the [Spring Boot starter](https://github.com/langchain4j/langchain4j-spring/blob/457bca4cd5e9c33df1dd323ce027818866a48139/langchain4j-open-ai-spring-boot-starter/src/main/java/dev/langchain4j/openai/spring/AutoConfig.java#L117) and [documentation](https://docs.langchain4j.dev/integrations/embedding-models/open-ai) as well? 🙏

PR #2250, opened 2024-12-07
Ollama Structured Outputs

### LINKED ISSUE #2246: Ollama: implement structured outputs
## Requirements
`OllamaChatModel` should implement `ChatResponse chat(ChatRequest)` method and map `ChatRequest.responseFormat.jsonSchema` into Ollama specific API.
`JsonSchemaElementHelper.toMap(JsonSchemaElement)` utility might be useful for that.

## Testing
An extensive set of tests already exists (`AiServicesWithJsonSchemaIT`).
An `OllamaAiServicesWithJsonSchemaIT` should be added and extend `AiServicesWithJsonSchemaIT`.
See `OpenAiAiServicesWithJsonSchemaIT` and `GoogleAiGeminiAiServicesWithJsonSchemaIT` as examples.

## More info
- https://ollama.com/blog/structured-outputs
- https://docs.langchain4j.dev/tutorials/structured-outputs#json-schema

bidek:
## Issue
Closes #2246

## Change
[Structured Outputs](https://ollama.com/blog/structured-outputs) are now supported by:
- `OllamaChatModel`
- `OllamaStreamingChatModel`
- `OllamaLanguageModel`
- `OllamaStreamingLanguageModel`

Read more [here](https://docs.langchain4j.dev/tutorials/structured-outputs#json-schema) and [here](https://docs.langchain4j.dev/integrations/language-models/ollama).

## General checklist


langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaChatModel.java
```diff
@@ -212,7 +243,7 @@ public OllamaChatModelBuilder stop(List<String> stop) {
             return this;
         }
 
-        public OllamaChatModelBuilder format(String format) {
+        public OllamaChatModelBuilder format(ResponseFormat format) {
             this.format = format;
```
langchain4j, line 262: This is a breaking change, is there a way to avoid it?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaStreamingChatModel.java
```diff
@@ -168,7 +170,7 @@ public OllamaStreamingChatModelBuilder stop(List<String> stop) {
             return this;
         }
 
-        public OllamaStreamingChatModelBuilder format(String format) {
+        public OllamaStreamingChatModelBuilder format(ResponseFormat format) {
```
langchain4j: Same here

ghost (commented):
@bidek thanks a lot, that was fast! :muscle:

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaChatModel.java
langchain4j: Can we still support method that accepts string and make it deprecated?

bidek: Okay, so I brought back the `format` field as a String (marked as deprecated) and introduced a new field, `responseFormat`.
I made the following assumption: when both fields are used with the builder, I throw an `IllegalStateException`, as only one can be used at a time. This assumption simplifies determining what and when a field is used.

Given this, the old APIs like `generate(messages)` can utilize either the deprecated `format` field or the new `responseFormat` field. However, the new ChatRequest API will always use the `responseFormat` field.
What are your thoughts?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaStreamingChatModel.java
bidek: Basically, the same answer as above, without the ChatRequest API part, as the streaming API does not have one for now.

langchain4j:
@bidek I've updated the PR description to correspond to the current state, could you please check?

Could you please disable failing tests in `OllamaAiServicesWithJsonSchemaIT` and/or try to adapat tests when possible (e.g., I guess "orange" case could be easily fixed).

Failing tests could be disabled ofr example like this:

```java
    @Override
    @Test
    @Disabled("llama 3.1 is cannot do it properly")
    void should_extract_pojo_with_array_of_primitives() {
    }
```

You can see more examples in `OllamaToolChatModelIT`.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaChatModel.java
```diff
@@ -131,18 +139,18 @@ public dev.langchain4j.model.chat.response.ChatResponse chat(final dev.langchain
     @Override
     public Set<Capability> supportedCapabilities() {
         Set<Capability> capabilities = new HashSet<>();
-        if (this.format != null && ResponseFormatType.JSON.equals(this.format.type())) {
+        if (this.responseFormat != null && ResponseFormatType.JSON.equals(this.responseFormat.type())) {
             capabilities.add(RESPONSE_FORMAT_JSON_SCHEMA);
```
langchain4j: I guess we should change the logic here and in streaming ollama model (as well as in `OpenAiChatModel` and other models supporting JSON schema, but out of scope of this PR of course) to rely on explicit user configuration:

```java
OllamaChatModel.builder()
    ...
    .supportedCapabilities(Set.of(RESPONSE_FORMAT_JSON_SCHEMA))
    .build();
```

So if user wants AI Service to use/generate JSON schemas for structured outputs, they will need to explicitly enable this capability.

On the other hand, if user wants to specify the response format (e.g., a concrete JSON schema to use), they will specify it like this:
```java
OllamaChatModel.builder()
    ...
    .responseFormat(ResponseFormat.builder().type(JSON).jsonSchema(...).build()) // to use structured outputs
    .responseFormat(ResponseFormat.JSON) // to use json mode
    .format("json") // this should work the same way as .responseFormat(ResponseFormat.JSON) to keep backwards compatibility
    .build();
```

Just in case, response format specified in the `ChatRequest` should always override the one specified in the constructor/builder of the model.

WDYT?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaChatModel.java
langchain4j: @bidek sounds reasonable, thank you!

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaStreamingChatModel.java
```diff
@@ -168,11 +178,19 @@ public OllamaStreamingChatModelBuilder stop(List<String> stop) {
             return this;
         }
 
+        /**
+         * @deprecated use responseFormat instead
```
langchain4j:
```suggestion
         * @deprecated Please use {@link #responseFormat(ResponseFormat)} instead.
         * For example: {@code responseFormat(ResponseFormat.JSON)}.
         * <br>
         * Instead of using JSON mode, consider using structured outputs with JSON schema instead,
         * see more info <a href="https://docs.langchain4j.dev/tutorials/structured-outputs#json-schema">here</a>.
```

langchain4j-ollama/src/test/java/dev/langchain4j/model/ollama/OllamaStructuredOutputIT.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    @Test
+    void should_generate_structured_output_streaming_old_format_way() throws ExecutionException, InterruptedException, TimeoutException {
+        // given
+        StreamingChatLanguageModel streamingOllamaChatModelWithOldFormatWayJsonSchema = OllamaStreamingChatModel.builder()
+                .baseUrl(ollamaBaseUrl())
+                .modelName(TOOL_MODEL)
+                .temperature(0.0)
+                .format(OllamaJsonUtils.toJson(
```
langchain4j: this should be .format("json") instead

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaChatModel.java
```diff
@@ -212,11 +252,20 @@ public OllamaChatModelBuilder stop(List<String> stop) {
             return this;
         }
 
+        /**
+         * @deprecated use responseFormat instead
```
langchain4j:
```suggestion
         * @deprecated Please use {@link #responseFormat(ResponseFormat)} instead.
         * For example: {@code responseFormat(ResponseFormat.JSON)}.
         * <br>
         * Instead of using JSON mode, consider using structured outputs with JSON schema instead,
         * see more info <a href="https://docs.langchain4j.dev/tutorials/structured-outputs#json-schema">here</a>.
```

langchain4j-ollama/src/test/java/dev/langchain4j/model/ollama/OllamaStructuredOutputIT.java
bidek: To clarify, the `format(String format)` method should only be used to set up JSON mode, not to pass a JSON Schema as a string. Does this mean the user can set up the JSON Schema only through the `responseFormat` method?

langchain4j: yes, correct. This is a deprecation, so we should support only what was supported so far. For the new feature they need ot use the new API

bidek: Okay, so I’ll delete that test/example since it doesn’t make any sense.

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaChatModel.java
bidek: Sounds great.
I will change the `supportedCapabilities` setup implementation to the one you suggested.

bidek:
> @bidek I've updated the PR description to correspond to the current state, could you please check?
> 
> Could you please disable failing tests in `OllamaAiServicesWithJsonSchemaIT` and/or try to adapat tests when possible (e.g., I guess "orange" case could be easily fixed).
> 
> Failing tests could be disabled ofr example like this:
> 
> ```java
>     @Override
>     @Test
>     @Disabled("llama 3.1 is cannot do it properly")
>     void should_extract_pojo_with_array_of_primitives() {
>     }
> ```
> 
> You can see more examples in `OllamaToolChatModelIT`.


I modified `AiServicesWithJsonSchemaIT` to allow disabling tests and added support for `ResponseFormat` in `OllamaLanguageModel` and `OllamaStreamingLanguageModel`.

ghost (dismissed):
@bidek thanks a lot, great job!

langchain4j:
@bidek could you please update https://docs.langchain4j.dev/integrations/language-models/ollama page and add a section for structured outputs?

Please describe:
- How to set the schema in the low-level API in:
  - constructor
  - request
- How to enable JSON schema functionality (via supported capabilities) to be able to use Ollama with AI Services

It would also be great to update existing Ollama examples in https://github.com/langchain4j/langchain4j-examples/blob/main/ollama-examples/src/main/java/OllamaChatModelTest.java and add new ones for json schema.

Thanks a lot in advance!

langchain4j:
@bidek while you are on it, could you please check something?

There is a theory that providing JSON schema twice (one time in the prompt and one time in the API that is used for constrained sampling) will give better results as LLM will "know" in advance what to output.

Since there are a few failing tests that you have disabled, could you please try to insert the schema into the prompt and see it it helps? You could try a few approaches:
- describe the schema in plain text
- use pseudo-schema like we do in AI Services now
- use real JSON schema in the prompt

This will help us a lot to set the default behaviour in AI service in order to improve the quality of outputs. Thanks a lot in advance!

bidek:
@langchain4j 
I conducted some tests as you requested—not all variants, but mainly two of them.

First of all, the last commit should be reverted. I only committed it so you could review it and perhaps suggest what to do next.

Before this commit, four tests were disabled.
After applying the changes:

- 3 out of 4 tests pass when appending only the serialized JSON schema to the prompt.
- 4 out of 4 tests pass when appending the text "Use the schema to extract data." This is strange.

In one test, I changed the `name` field from `name` to `firstName` because LLaMA always hallucinates a full name. This change was enough for the test to pass.

I did not conduct tests with a described JSON schema in the prompt or with pseudo-schema like in AIServices.

langchain4j:
@bidek thanks a lot for digging into it!

> 4 out of 4 tests pass when appending the text "Use the schema to extract data." This is strange.

Hmm, interesting. Perhaps not very strange regarding explicitly specifying the instruction to use the schema and extract the data. After all LLMs are instruction-tuned, so it makes sense that explicit request can make a difference.
But if LLM itself does not see the schema when generating tokens, then it is quite strange. Unless LLM interprets this instruction as "respond in JSON format" and thus it works better.

Do you know if Ollama automatically appends the schema into the prompt? I guess this should be visible in the logs somewhere

bidek:
@langchain4j 

Ollama does not append a schema to the prompt. Instead, it uses something called "Output Constraints." As far as I know, this instructs the model to generate answers constrained to a specified grammar using GBNF/BNF. You can read more about it https://github.com/ggerganov/llama.cpp/blob/master/grammars/README.md.

To run Ollama with more verbose logs, you need to set an environment variable: OLLAMA_DEBUG=1.

Here is an example of a prompt sent from Ollama to the llama.cpp backend:
 phi3.5:
 ```
 level=DEBUG source=routes.go:1466 msg="chat request" images=0 prompt="<|user|>\nKlaus has 2 pets: Peanut and Muffin<|end|>\n<|assistant|>\n"
 ```
 
 llama3.1:
 ```
 level=DEBUG source=routes.go:1466 msg="chat request" images=0 prompt="<|start_header_id|>user<|end_header_id|>\n\nKlaus has 2 pets: Peanut and Muffin<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"
 ```

langchain4j:
@bidek 

> Ollama does not append a schema to the prompt.

Are you sure about this?

> Instead, it uses something called "Output Constraints." As far as I know, this instructs the model to generate answers constrained to a specified grammar using GBNF/BNF. You can read more about it https://github.com/ggerganov/llama.cpp/blob/master/grammars/README.md.

As far as I understand, this approach restricts the space of valid tokens during _sampling_, but it does not `instruct` the model anyhow, so the model itself does not have the schema in the context when generating the output. Simply put, the model is not aware of the need to answer in a specific way, it responds as usual, but the respriction is applied during sampling.

langchain4j:
@bidek some LLMs add request parameters into the system prompt, for example with tools, so that LLM is aware of tools when generating response.
Same could(or even should?) happen with structured outputs. Ollama could use the provided schema not only to constrain the outputs during sampling, but also to automatically add the schema to the system message to make LLM aware of the need to answer using the provided schema. According to your tests, LLM behaved better when it knew about the schema. So it would totally make sense to do that by Ollama automatically. I am wonderign if they do it already or if they plan to. I would appreciate a lot if you could research this 🙏

bidek:
> @bidek
> 
> > Ollama does not append a schema to the prompt.
> 
> Are you sure about this?

@langchain4j 
I’m quite sure. The logs attached in the previous comment prove this, I think.
I conducted an experiment and added a SystemMessage through annotation to the `should_extract_pojo_with_set_of_pojos` test case, like this:
```
        @SystemMessage("Use schema to extract data")
        Person extractPersonFrom(String text);
```

And log line from Ollama server look like this now:
```
level=DEBUG source=routes.go:1466 msg="chat request" images=0 prompt="<|start_header_id|>system<|end_header_id|>\n\nUse schema to extract data<|eot_id|><|start_header_id|>user<|end_header_id|>\n\nKlaus has 2 pets: Peanut and Muffin<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"
```
So system message is attached to prompt, but no JsonSchema.


> 
> As far as I understand, this approach restricts the space of valid tokens during _sampling_, but it does not `instruct` the model anyhow, so the model itself does not have the schema in the context when generating the output. Simply put, the model is not aware of the need to answer in a specific way, it responds as usual, but the respriction is applied during sampling.


You are right, of course. My English is not very good, so I chose the wrong word ('instruct').


>
> some LLMs add request parameters into the system prompt, for example with tools, so that LLM is aware of tools when generating response.
Same could(or even should?) happen with structured outputs. Ollama could use the provided schema not only to constrain the outputs during sampling, but also to automatically add the schema to the system message to make LLM aware of the need to answer using the provided schema. According to your tests, LLM behaved better when it knew about the schema. So it would totally make sense to do that by Ollama automatically. I am wonderign if they do it already or if they plan to. I would appreciate a lot if you could research this 🙏


I conduct another test (modified should_execute_a_tool_then_answer from OllamaToolChatIT)
And when tools are specified, prompt looks like this:


```
level=DEBUG source=routes.go:1466 msg="chat request" images=0 prompt="<|start_header_id|>system<|end_header_id|>\n\nCutting Knowledge Date: December 2023\n\nWhen you receive a tool call response, use the output to format an answer to the orginal user question.\n\nYou are a helpful assistant with tool calling capabilities.<|eot_id|><|start_header_id|>user<|end_header_id|>\n\nGiven the following functions, please respond with a JSON for a function call with its proper arguments that best answers the given prompt.\n\nRespond in the format {\"name\": function name, \"parameters\": dictionary of argument name and its value}. Do not use variables.\n\n{\"type\":\"function\",\"function\":{\"name\":\"get_current_weather\",\"description\":\"Get the current weather for a location\",\"parameters\":{\"type\":\"object\",\"required\":[\"format\",\"location\"],\"properties\":{\"format\":{\"type\":\"string\",\"description\":\"The format to return the weather in, e.g. 'celsius' or 'fahrenheit'\",\"enum\":[\"celsius\",\"fahrenheit\"]},\"location\":{\"type\":\"string\",\"description\":\"The location to get the weather for, e.g. San Francisco, CA\"}}}}}\n\nQuestion: What is the weather today in Paris?<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"
```

Without tools:

```
level=DEBUG source=routes.go:1466 msg="chat request" images=0 prompt="<|start_header_id|>user<|end_header_id|>\n\nWhat is the weather today in Paris?<|eot_id|><|start_header_id|>assistant<|end_header_id|>\n\n"
```

I checked some templates for models tagged with the 'tools' tag, and it seems that they all include some tool awareness in their templates. 
For example, this template (https://ollama.com/library/llama3.1/blobs/948af2743fc7) was applied to the test mentioned above.

I will try to find some information about potentially using JSON Schema in those templates. Maybe I’ll try reaching out to an Ollama developer to confirm or deny this.

ghost (approved):
@bidek thanks a lot, great job!

I will merge this PR to include it in the upcoming release this week.

Could you please update the documentation/examples in a separate PR?  
It would also be great if you could update Spring Boot starter to allow configuring supported capabilities and check if it is necessary to update anything else. Thanks a lot in advance! 🙏

langchain4j:
@bidek 

> I’m quite sure. The logs attached in the previous comment prove this, I think.

My bad, I thought it was just an example, not an actual output.

> You are right, of course. My English is not very good, so I chose the wrong word ('instruct').

No problem at all! I am very happy to have this discussion :) The picture becomes much clearer when discussing with someone and verbalizing the thoughts!


> I will try to find some information about potentially using JSON Schema in those templates. Maybe I’ll try reaching out to an Ollama developer to confirm or deny this.

Thanks a lot! It would really be great if Ollama automatically added JSON schema to the system prompt, similarly to how they do for tools.
This would lift the need from us to implement this feature (adding JSON schema both in the prompt and in the `response` API parameter) on our side. And hopefully benefit a lot of people, as adding the schema seems to make the model output more consistently.

langchain4j:
Also found this, might be helpful: https://blog.danielclayton.co.uk/posts/ollama-structured-outputs/

P.S.  Idid not read if fully yet, might be irrelevant

PR #2252, opened 2024-12-09
 [BUG] unable to provide a @Tool with java.lang.UUID in parameters

### LINKED ISSUE #2251: [BUG] unable to provide a @Tool with java.lang.UUID in parameters
**Describe the bug**

When annotating with @Tool on a method having a parameter of type java.util.UUID (or if a parameter is a collection or a class containing a java.util.UUID), then a property of type JsonArraySchema will be added with mostSigBits and leastSigBits, instead of a property of a JsonStringSchema.

Later when the model tries to call the tool, the property will fail to be de-serialized because Gson expects a String, not a JSON array.

Even if we register a new TypeAdapter to parse this {mostSigBits, leastSigBits} object, it will never be converted to the original UUID due to the lack of precision in the Long values sent by the model.

**Log and Stack trace**

```
Exception in thread "main" com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 1 column 2 path $
	at com.google.gson.Gson.fromJson(Gson.java:1238)
	at com.google.gson.Gson.fromJson(Gson.java:1137)
	at com.google.gson.Gson.fromJson(Gson.java:1047)
	at com.google.gson.Gson.fromJson(Gson.java:982)
	at dev.langchain4j.internal.GsonJsonCodec.fromJson(GsonJsonCodec.java:103)
	at dev.langchain4j.internal.Json.fromJson(Json.java:99)
	at dev.langchain4j.service.tool.DefaultToolExecutor.coerceArgument(DefaultToolExecutor.java:218)
	at dev.langchain4j.service.tool.DefaultToolExecutor.prepareArguments(DefaultToolExecutor.java:120)
	at dev.langchain4j.service.tool.DefaultToolExecutor.execute(DefaultToolExecutor.java:59)
	at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:260)
	at uuidbug.$Proxy3.chat(Unknown Source)
	at uuidbug.Main.main(Main.java:58)
Caused by: java.lang.IllegalStateException: Expected a string but was BEGIN_OBJECT at line 1 column 2 path $
	at com.google.gson.stream.JsonReader.nextString(JsonReader.java:836)
	at com.google.gson.internal.bind.TypeAdapters$24.read(TypeAdapters.java:588)
	at com.google.gson.internal.bind.TypeAdapters$24.read(TypeAdapters.java:581)
	at com.google.gson.Gson.fromJson(Gson.java:1227)
	... 11 more

```

**To Reproduce**

```java
package uuidbug;

import dev.langchain4j.agent.tool.Tool;
import dev.langchain4j.memory.chat.MessageWindowChatMemory;
import dev.langchain4j.model.chat.ChatLanguageModel;
import dev.langchain4j.model.openai.OpenAiChatModel;
import dev.langchain4j.service.AiServices;

import java.util.List;
import java.util.Map;
import java.util.UUID;

import static dev.langchain4j.model.openai.OpenAiChatModelName.GPT_4_O_MINI;


public class Main {

    static class TaskRepository {

        private final Map<UUID, String> tasks = Map.of(
                UUID.randomUUID(), "Submit an issue on langchain4j",
                UUID.randomUUID(), "Refresh my resume",
                UUID.randomUUID(), "Find a job"
        );

        @Tool
        public String getTaskById(UUID id) {
            return tasks.get(id);
        }

        public List<UUID> getTasks() {
            return tasks.keySet().stream().toList();
        }
    }

    interface Assistant {
    }

        String OPENAI_API_KEY = "demo";

        ChatLanguageModel model = OpenAiChatModel.builder()
                .apiKey(OPENAI_API_KEY)
                .modelName(GPT_4_O_MINI)
                .strictTools(true)
                .build();

        TaskRepository taskRepository = new TaskRepository();
                .chatLanguageModel(model)
                .tools(taskRepository)
                .chatMemory(MessageWindowChatMemory.withMaxMessages(10))
                .build();

        String taskId = taskRepository.getTasks().get(0).toString();
        String question = "What contains the task with id " + taskId + "?";
        String answer = assistant.chat(question);
        System.out.println(answer);
    }

}
```
**Expected behavior**
The tool should be called with a proper UUID.

- LangChain4j version: 0.36.2
- LLM(s) used: OpenAI
- Java version: 22

**Additional context**

Fix can be done in langchain4j-core > dev.langchain4j.internal.TypeUtils by adding the type java.lang.UUID in isJsonString():

```java
    public static boolean isJsonString(Class<?> type) {
        return type == UUID.class || type == String.class || type == char.class || type == Character.class
                || CharSequence.class.isAssignableFrom(type);
    }

```

martinsaison:
Fixes #2251

## Issue
Closes #2251

## Change
Include java.lang.UUID to langchain4j-core > dev.langchain4j.internal.TypeUtils.isJsonString(), so the ToolSpecification will contain the correct format when the model needs to call a tool.

## General checklist


langchain4j:
@martinsaison thanks a lot!

Could you please extend `AiServicesWithNewToolsIT` and `AiServicesWithJsonSchemaIT` with an extra test that will include UUID field?

martinsaison:
@langchain4j done but it failed for Java 21, probably a network issue during the build

[ERROR] error reading /home/runner/.m2/repository/com/fasterxml/jackson/core/jackson-databind/2.16.1/jackson-databind-2.16.1.jar; zip file is empty

mariofusco (dismissed):

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithJsonSchemaIT.java
```diff
@@ -757,67 +788,179 @@ void should_extract_pojo_with_local_date_time_fields() {
...
+
+            verify(model)
+                    .chat(ChatRequest.builder()
+                            .messages(singletonList(userMessage(text)))
+                            .responseFormat(ResponseFormat.builder()
+                                    .type(JSON)
+                                    .jsonSchema(JsonSchema.builder()
+                                            .name("Person")
+                                            .rootElement(JsonObjectSchema.builder()
+                                                    .addStringProperty("id")
```
dliubarskyi: I am wondering if we should include a description (e.g., "UUID" or "String in a UUID format") for `UUID` type by default?
In case user defined their own `@Description`, we can skip it. But if there is no `@Description`, LLM will not know in which format an ID should be, I am sure this can be ambiguous for the LLM.
WDYT?

dliubarskyi (commented):
@martinsaison thank you!

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithJsonSchemaIT.java
martinsaison: @dliubarskyi you are right, in my app I am using the tool annotation @P(value = "somethingId (uuid format)"), otherwise the LLM will get confused. I had to reinstall my system and don't have the IDE and git on it yet, so I'm not sure what to change right now to set a default description, but I can look at it next week if you want

dliubarskyi: No rush, it can wait till next week. Thanks!

martinsaison: @dliubarskyi I pushed a change but the merge with main broke everything, I will rework and squash it :-)

martinsaison: @dliubarskyi I pushed the change (I reapplied the same modifications on the last main and added the default description)

langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonSchemaElementHelper.java
```diff
@@ -13,51 +21,41 @@
...
-                                                          Type type,
-                                                          String fieldDescription,
-                                                          Map<Class<?>, VisitedClassMetadata> visited) {
-        if (isJsonString(clazz)) {
+    public static JsonSchemaElement jsonSchemaElementFrom(
+            Class<?> clazz, Type type, String fieldDescription, Map<Class<?>, VisitedClassMetadata> visited) {
+        if (isJsonUUID(clazz)) {
             return JsonStringSchema.builder()
-                    .description(fieldDescription)
+                    .description(fieldDescription == null ? "String in a UUID format" : fieldDescription)
```
dliubarskyi: Hmm, I am not sure it is the right place for this logic. Are there alternatives?

martinsaison: I pushed a new commit to move the logic to descriptionFrom() and added the call inside the block "if (isJsonString(clazz)) {}". Let me know if it sounds better or if you have other suggestions.

dliubarskyi: Yeah, looks better, thank you!

dliubarskyi (approved):
@martinsaison thank you!

PR #2265, opened 2024-12-12
Configurable timeout for tool executions over MCP

### LINKED ISSUE #2262: [FEATURE] MCP: Configurable tool execution timeout and canceling
The MCP protocol supports canceling a running tool execution. We could allow specifying a timeout for tool executions, and if the timeout is reached, send the server a cancellation message.

Canceling running locally executed tools will be a bit tricky, but some kind of mechanism may be possible too (at least a way to let the code know that it can stop executing?).

jmartisk:
## Issue

Closes #2262

## General checklist

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -11,7 +11,6 @@ import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
-import dev.langchain4j.mcp.client.protocol.CancellationNotification;
 import dev.langchain4j.mcp.client.protocol.InitializeParams;
 import dev.langchain4j.mcp.client.protocol.McpCallToolRequest;
 import dev.langchain4j.mcp.client.protocol.McpInitializeRequest;
@@ -38,6 +37,7 @@ public class DefaultMcpClient implements McpClient {
     private final String protocolVersion;
     private final Duration toolExecutionTimeout;
     private final JsonNode RESULT_TIMEOUT;
+    private final String toolExecutionTimeoutErrorMessage;
 
     public DefaultMcpClient(Builder builder) {
         transport = ensureNotNull(builder.transport, "transport");
@@ -45,13 +45,15 @@ public class DefaultMcpClient implements McpClient {
         clientVersion = getOrDefault(builder.clientVersion, "1.0");
         protocolVersion = getOrDefault(builder.protocolVersion, "2024-11-05");
         toolExecutionTimeout = getOrDefault(builder.toolExecutionTimeout, Duration.ofSeconds(60));
+        toolExecutionTimeoutErrorMessage =
+                getOrDefault(builder.toolExecutionTimeoutErrorMessage, "There was a timeout executing the tool");
         RESULT_TIMEOUT = JsonNodeFactory.instance.objectNode();
         ((ObjectNode) RESULT_TIMEOUT)
                 .putObject("result")
                 .putArray("content")
                 .addObject()
                 .put("type", "text")
-                .put("text", "There was a timeout executing the tool");
+                .put("text", toolExecutionTimeoutErrorMessage);
 
         // Initialize the client...
         transport.start();
@@ -100,8 +102,7 @@ public class DefaultMcpClient implements McpClient {
                 new McpCallToolRequest(idGenerator.getAndIncrement(), executionRequest.name(), arguments);
         JsonNode executionResult = null;
         try {
-            executionResult = transport.executeTool(
-                    operation, toolExecutionTimeout, () -> new CancellationNotification(operation.getId(), "Timeout"));
+            executionResult = transport.executeTool(operation, toolExecutionTimeout);
         } catch (TimeoutException timeout) {
             executionResult = RESULT_TIMEOUT;
         }
@@ -120,6 +121,7 @@ public class DefaultMcpClient implements McpClient {
 
     public static class Builder {
 
+        public String toolExecutionTimeoutErrorMessage;
         private McpTransport transport;
         private String clientName;
         private String clientVersion;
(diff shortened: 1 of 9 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -84,9 +98,15 @@ public String executeTool(ToolExecutionRequest executionRequest) {
         }
         McpCallToolRequest operation =
                 new McpCallToolRequest(idGenerator.getAndIncrement(), executionRequest.name(), arguments);
-        JsonNode jsonNode = transport.executeTool(operation);
+        JsonNode executionResult = null;
+        try {
+            executionResult = transport.executeTool(
+                    operation, toolExecutionTimeout, () -> new CancellationNotification(operation.getId(), "Timeout"));
```
langchain4j: Is there a reason to have this `CancellationNotification` be propagated from here rather than being created inside the transport?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -32,12 +36,22 @@ public class DefaultMcpClient implements McpClient {
     private final String clientName;
     private final String clientVersion;
     private final String protocolVersion;
+    private final Duration toolExecutionTimeout;
+    private final JsonNode RESULT_TIMEOUT;
```
langchain4j, line 39: Was it supposed to be a constant?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -32,12 +36,22 @@ public class DefaultMcpClient implements McpClient {
...
         clientVersion = getOrDefault(builder.clientVersion, "1.0");
         protocolVersion = getOrDefault(builder.protocolVersion, "2024-11-05");
+        toolExecutionTimeout = getOrDefault(builder.toolExecutionTimeout, Duration.ofSeconds(60));
+        RESULT_TIMEOUT = JsonNodeFactory.instance.objectNode();
+        ((ObjectNode) RESULT_TIMEOUT)
+                .putObject("result")
+                .putArray("content")
+                .addObject()
+                .put("type", "text")
+                .put("text", "There was a timeout executing the tool");
```
langchain4j: Could you make the text of the timeout error configurable please? This is what LLM will see, so it might need some tweaking/"prompt engineering"

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -111,6 +122,47 @@ private JsonNode executeAndWait(Request request, Long id) {
             return future.get();
         } catch (Exception e) {
             throw new RuntimeException(e);
+        } finally {
+            pendingOperations.remove(id);
+        }
+    }
+
+    private JsonNode executeAndWait(
```
langchain4j: nit: there seems to be some duplication between both transports, it might make sense to consider extracting common logic, but not critical for sure

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -111,6 +122,47 @@ private JsonNode executeAndWait(Request request, Long id) {
...
+        } catch (TimeoutException timeoutException) {
+            if (cancellationNotificationSupplier != null) {
+                CancellationNotification cancellationNotification = cancellationNotificationSupplier.get();
+                try {
+                    Request cancellationRequest = new Request.Builder()
+                            .url(postUrl)
+                            .header("Content-Type", "application/json")
+                            .post(RequestBody.create(OBJECT_MAPPER.writeValueAsBytes(cancellationNotification)))
+                            .build();
+                    try (Response response = client.newCall(cancellationRequest).execute()) {}
```
langchain4j: `response` is unused, is it expected?

ghost (commented):
@jmartisk thank you!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: Yea, but I initialize it in the constructor because I couldn't figure out a way with the fluent API to build such complex object without the last statement returning the whole object, not one nested field in it

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: This handling will be the same for all transports, so I wanted to avoid code duplication

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: Will do

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
langchain4j: I see, pehaps a good place to put it is in `AbstractMcpTransport`?

jmartisk: But yeah it doesn't save much code after all I guess. I'll move it to the transport then

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: I'm not sure what we should do with it, this is a fire-and-forget operation. Maybe I'll add a warning log in case the response code is unexpected

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: After the latest commit amendment, at least my IDE no longer suggests there is duplication :)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: > I see, pehaps a good place to put it is in `AbstractMcpTransport`?

I don't think it's worth it, there wouldn't be too much code there, and just complicate things. I moved the CancellationNotification to the transports, and my IDE actually stopped suggesting that there is code duplication, so that was a good idea

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -100,10 +121,12 @@ public void close() {
 
     public static class Builder {
 
+        public String toolExecutionTimeoutErrorMessage;
```
langchain4j:
```suggestion
        private String toolExecutionTimeoutErrorMessage;
```

ghost (dismissed):
@jmartisk thanks a lot!

PR #2267, opened 2024-12-12
Properly receive the MCP POST URL from the MCP server

### LINKED ISSUE #2266: [BUG] MCP: obtain the POST url from the server
Currently, we have the POST url configurable, but it's actually something that the server gives you during the initialization, and it includes the proper session ID.

jmartisk:
Closes #2266

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
@@ -1,5 +1,6 @@
 package dev.langchain4j.mcp.client.transport.http;
 
+import static dev.langchain4j.internal.Utils.getOrDefault;
 import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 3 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -116,18 +122,32 @@ private JsonNode executeAndWait(Request request, Long id) {
...
-        CompletableFuture<Void> initializationFinished = new CompletableFuture<>();
+        CompletableFuture<String> initializationFinished = new CompletableFuture<>();
         SseEventListener listener = new SseEventListener(pendingOperations, logResponses, initializationFinished);
         EventSource eventSource = EventSources.createFactory(client).newEventSource(request, listener);
-        // wait for the SSE channel to be created, throw an exception if that failed
+        // wait for the SSE channel to be created, receive the POST url from the server, throw an exception if that
+        // failed
         try {
-            initializationFinished.get();
+            int timeout = client.callTimeoutMillis() > 0 ? client.callTimeoutMillis() : Integer.MAX_VALUE;
```
langchain4j, line 131: Is it a good idea to have `MAX_VALUE` as a timeout?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -116,18 +122,32 @@ private JsonNode executeAndWait(Request request, Long id) {
...
+            log.debug("Received the server's POST URL: " + postUrl);
         } catch (Exception e) {
             throw new RuntimeException(e);
         }
         return eventSource;
     }
 
+    private String buildAbsolutePostUrl(String relativePostUrl) {
+        try {
+            URL sseUrl = new URL(this.sseUrl);
```
langchain4j: Perharps we should use `URI` here instead?

ghost (commented):
@jmartisk thanks a lot!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: I could instead create an `if` statement and call the `get()` method without a timeout, the code will look a bit more complicated, but the behavior will basically be the same.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: Like this?
```
URI sseUrl = new URI(this.sseUrl);
URL postUrl = new URL(sseUrl.toURL(), relativePostUrl);
return postUrl.toString();
```

or what else do you have in mind?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
langchain4j: I meant that maybe we should specify a bit more meaningful default value for the timeout, perhaps 60-90 seconds? `MAX_VALUE` seems like "wait for it forever"

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
langchain4j: I have heard bad things about `URL` (how it initiates a connection in `equals()`, etc.).
IIUC, the general advice is to use `URI` instead.
So if this is true, I assume we should use URI everywhere?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: The default is 60 seconds (this is just to handle the case that somebody explicitly specifies the timeout to be 0)

jmartisk: Ah, hold on, you may actually be right that we don't have a default. Ok I'll add a default as well, but we have to keep handling the case that somebody specifies it to be zero explicitly (which should mean "no timeout", but calling `completableFuture.get(0)` would throw an exception immediately)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: Ok, I've replaced it with `URI.create(this.sseUrl).resolve(relativePostUrl).toString()`

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: I've added a 60 second timeout, right now it's just one value for callTimeout, connectTimeout, readTimeout, writeTimeout. Do you prefer having four separate values?

langchain4j: @jmartisk thank you! I think it is ok to have a single value for now

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
langchain4j: Thanks!

ghost (approved):
@jmartisk thank you!

PR #2271, opened 2024-12-13
fix #1052: Handle hallucinated tool name.

### LINKED ISSUE #1052: [BUG] Handle hallucinated tool name
**Describe the bug**

Sometimes I get NPE because toolExecutor is NULL.

**Log and Stack trace**

```
java.lang.NullPointerException: Cannot invoke "dev.langchain4j.agent.tool.ToolExecutor.execute(dev.langchain4j.agent.tool.ToolExecutionRequest, Object)" because "toolExecutor" is null
	at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:155)
```

**To Reproduce**

I asked my chatbot directly like: "Calculate document link using the documentId".
I see (in the debugger) that it knows documentId value from the context window.

**Expected behavior**

The corresponding function gets called

**Current behaviour**
NPE in the `DefaultAIServices` class.

- LangChain4j version: 0.30.0
- LLM(s) used: Azure OpenAI (with gpt-3.5-turbo)
- Java version: 21
- Spring Boot version (if applicable): 3.2.5

**Additional context**

This is where it breaks:
image

**Please NOTE** that the `toolExecutionRequest` contains a strange name "python". At the same time, it knows the `documentId`.
image

My tool is defined as follows:
```
    @Component
    static class Tools {

        @Tool("Returns document link")
        String getDocumentLink(@P("documentId") String documentId) {
            return "https://<private-address>/%s".formatted(documentId);
        }
    }
```

ehsavoie:
## Issue

Closes # 1052

## Change
If the tool name doesn't exist, LangChain4J shouldn't fail with a NPE. It should just ignore it.

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -65,6 +65,12 @@ class DefaultAiServices<T> extends AiServices<T> {
 
     private final ServiceOutputParser serviceOutputParser = new ServiceOutputParser();
     private final Collection<TokenStreamAdapter> tokenStreamAdapters = loadFactories(TokenStreamAdapter.class);
+    private final ToolExecutor hallucinatedToolExecutor =
+            (ToolExecutionRequest toolExecutionRequest, Object memoryId) -> {
+                throw runtime(
+                        "Something is wrong, the tool %s was called but it is not a part of the available tools",
+                        toolExecutionRequest.name());
+            };
 
     DefaultAiServices(AiServiceContext context) {
         super(context);
@@ -267,7 +273,8 @@ class DefaultAiServices<T> extends AiServices<T> {
                             }
 
                             for (ToolExecutionRequest toolExecutionRequest : aiMessage.toolExecutionRequests()) {
-                                ToolExecutor toolExecutor = toolExecutors.get(toolExecutionRequest.name());
+                                ToolExecutor toolExecutor = toolExecutors.getOrDefault(
+                                        toolExecutionRequest.name(), hallucinatedToolExecutor);
                                 String toolExecutionResult = toolExecutor.execute(toolExecutionRequest, memoryId);
                                 toolExecutions.add(ToolExecution.builder()
                                         .request(toolExecutionRequest)
--- a/langchain4j-core/src/test/java/dev/langchain4j/model/chat/mock/ChatModelMock.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/model/chat/mock/ChatModelMock.java
@@ -1,5 +1,11 @@
 package dev.langchain4j.model.chat.mock;
 
+import static dev.langchain4j.internal.Exceptions.runtime;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotBlank;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+import static java.util.Collections.synchronizedList;
+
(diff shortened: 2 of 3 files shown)


langchain4j:
Hi @ehsavoie, thank you!

While ignoring the hallucinated tool might be a valid strategy for some use cases, I would argue that it should not be the default one. Ideally, we should apply some retry/re prompt strategy, or fail explicitly.

Also please consider that ChatMemory should be left in a valid state, so if there is an AiMessage with tool calls, it should be followed by tool result messages.

Tests are also missing.

ehsavoie:
@langchain4j sorry I missed your comment on the issue. I'll try to think of something better

MarlonCarvalhoUO:
Hey! How are you? I hope so! Do you intend to create AI agents for specific tasks like crewAi and Langchain?

langchain4j:
@MarlonCarvalhoUO please take a look at:
- https://github.com/bsorrentino/langgraph4j
- https://github.com/czelabueno/langchain4j-workflow

ehsavoie:
I've updated the code o that it will throw a RuntimeException in this case. This behavior can be changed using a ServiceLoader to provide for example an error message to send back to the LLM.

langchain4j/src/main/java/dev/langchain4j/service/tool/HallucinatedToolExecutor.java
```diff
@@ -0,0 +1,3 @@
+package dev.langchain4j.service.tool;
+
+public interface HallucinatedToolExecutor extends ToolExecutor {}
```
langchain4j: why is new interface needed?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -50,25 +62,26 @@
...
 
     private final ServiceOutputParser serviceOutputParser = new ServiceOutputParser();
     private final Collection<TokenStreamAdapter> tokenStreamAdapters = loadFactories(TokenStreamAdapter.class);
+    private final HallucinatedToolExecutor hallucinatedToolExecutor;
 
     DefaultAiServices(AiServiceContext context) {
         super(context);
+        HallucinatedToolExecutor loadedHallucinatedToolExecutor = null;
+        for (HallucinatedToolExecutorFactory factory :
+                ServiceHelper.loadFactories(HallucinatedToolExecutorFactory.class)) {
```
langchain4j: I am not sure about using SPI for this. What if one AI service needs one behavior in case of hallucination, and another AI service - another behavior? It is not possible to support this with SPI.

IIRC @cescoffier was working on something similar with guardrails, so it would make sense to consolidate our efforts

langchain4j/src/main/java/dev/langchain4j/service/tool/HallucinatedToolExecutor.java
ehsavoie: Removing it

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
ehsavoie: Ok, since guardrails should protect against that I'll keep the code simple.

ehsavoie:
@langchain4j I've simplified the code to throw an exception and even added a test ;)

dliubarskyi:
Hi @ehsavoie, thanks a lot for the initial implementation!

@mariofusco has completed it in the https://github.com/langchain4j/langchain4j/pull/2460, so I am closing this issue

PR #2278, opened 2024-12-16
Mistral: map toolCallId for tool messages

ghost:
## Change
Mistral tests are currently fail because all of a sudden Mistral decided to make `tool_call_id` parameter in `tool` messages mandatory:
`Runtime java.lang.RuntimeException: status code: 400; body: {"object":"error","message":"Tool call id has to be defined.","type":"invalid_request_error","param":null,"code":null}`.

This probably affects all Mistral users using tools.

cc @geoand 

## General checklist

### DIFF
--- a/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/api/MistralAiChatMessage.java
+++ b/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/api/MistralAiChatMessage.java
@@ -25,4 +25,5 @@ public class MistralAiChatMessage {
     private String content;
     private String name;
     private List<MistralAiToolCall> toolCalls;
+    private String toolCallId;
 }
--- a/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java
+++ b/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java
@@ -90,6 +90,7 @@ public class MistralAiMapper {
         if (message instanceof ToolExecutionResultMessage) {
             return MistralAiChatMessage.builder()
                     .role(MistralAiRole.TOOL)
+                    .toolCallId(((ToolExecutionResultMessage) message).id())
                     .name(((ToolExecutionResultMessage) message).toolName())
                     .content(((ToolExecutionResultMessage) message).text())
                     .build();


langchain4j:
cc @katcharov, this is an example of an issue that would not be caught with mocked LLM responses

geoand:
Thanks a lot for the heads up!

czelabueno:
hey @langchain4j your changes are enough to fix the issue, however this change for tool messages have introduced also a new `MistralAiToolChoiceName` enum called `required` which is pending to add. This new enum is not strictly related to the tool messages requests but would be great to complete with all changes involved in tool messages
We would need change `ANY` by `REQUIRED` here:

https://github.com/langchain4j/langchain4j/blob/9b63e2ccdf76b8014699498af39ca8bcf81cc009/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiChatModel.java#L163

and here:

https://github.com/langchain4j/langchain4j/blob/9b63e2ccdf76b8014699498af39ca8bcf81cc009/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiStreamingChatModel.java#L156

We can handle it in other PR if you prefer it like this. WDYT?

langchain4j:
@czelabueno oh, good to know! I guess tool choice logic can be added after https://github.com/langchain4j/langchain4j/pull/1989 is merged (I plan to merge tomorrow). https://github.com/langchain4j/langchain4j/pull/1989 introduces model-agnostic `ToolChoice`, so I guess it makes sense to wait till then?

langchain4j:
BTW did Mistral announce this **_breaking_** change? I did not get any email notification

czelabueno:
@langchain4j about below

> @czelabueno oh, good to know! I guess tool choice logic can be added after https://github.com/langchain4j/langchain4j/pull/1989 is merged (I plan to merge tomorrow). https://github.com/langchain4j/langchain4j/pull/1989 introduces model-agnostic ToolChoice, so I guess it makes sense to wait till then?

ok make perfect sense to me, in fact the new `ChatParameters` fit well with the current MistralAI API specs, if it will be abstracted would be great!

czelabueno:
@langchain4j 

> BTW did Mistral announce this breaking change? I did not get any email notification

in fact it was changed from May 24 😅, but I think that due to recent API changes, the API validation is well working now, for their official SDK it's mandatory as well.
This is my bad because I was away for a few months but I'm already active again to update all the changes 🙂
 
image

langchain4j:
@czelabueno ah, I see! But in any case I would expect them to send the "final warning" email before actually making it mandatory. I can imagine there are still a lot of applicaitons that do not send the id and were broken by this change

PR #2323, opened 2024-12-21
Adds the possibility to encrypt vectors and text segments in the DB

### LINKED ISSUE #2296: [FEATURE] Encryption of segments

Encrypting segments (and even calculated vectors) before they are stored into the storage

I'd love to have an interface at this point https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java#L205 so it receives the EmbeddingVectors and the segments and they can both manipulate, either for encrypting the segments, encrypting the vectors, or both.

For now, I am reimplementing some of the classes to support it.

**Additional context**

I can provide a PR.

cc @mariofusco

lordofthejars:
## Issue

Closes #2296 

## Change

Adds the possibility to encrypt vectors and text segments in the DB

## General checklist

- [ ~]  There are no breaking changes *If the user uses Builder in EmbeddingStoreIngestor* then no break as I only changed the constructor arguments. But in the constructor now you need to set two new optional parameters. To avoid this situation we can duplicate the constructor. In fact no test needed to be rewrittten.

No IT test required

NA

cc/ @mariofusco

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/data/embedding/EmbeddingTransformer.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/data/embedding/EmbeddingTransformer.java
@@ -1,0 +1,37 @@
+package dev.langchain4j.data.embedding;
+
+import dev.langchain4j.data.segment.TextSegment;
+
+import java.util.List;
+import java.util.Objects;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Defines the interface for transforming a {@link Embedding}.
+ * Implementations can perform a variety of tasks such as encrypting, salting, etc.
+ */
+public interface EmbeddingTransformer {
+
+    /**
+     * Transforms a provided embedding.
+     *
+     * @param embedding The embedding to be transformed.
+     * @return The transformed embedding, or null if the segment should be filtered out.
+     */
+    Embedding transform(Embedding embedding);
+
+    /**
+     * Transforms all the provided embeddings.
+     *
+     * @param embeddings A list of embeddings to be transformed.
+     * @return A list of transformed embeddings. The length of this list may be shorter or longer than the original list. Returns an empty list if all segments were filtered out.
+     */
+    default List<Embedding> transformAll(List<Embedding> embeddings) {
+        return embeddings.stream()
+                .map(this::transform)
+                .filter(Objects::nonNull)
+                .collect(toList());
+    }
+
+}
--- a/langchain4j-core/src/main/java/dev/langchain4j/data/segment/TextSegmentStorageTransformer.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/data/segment/TextSegmentStorageTransformer.java
@@ -1,0 +1,36 @@
+package dev.langchain4j.data.segment;
+
+import java.util.List;
+import java.util.Objects;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Defines the interface for transforming a {@link TextSegment} before it is inserted into the database.
+ * Implementations can perform a variety of tasks such as transforming, filtering, encrypting, signing, etc.
+ * This is executed after the vector is calculated and before data is inserted into the storage.
+ */
+public interface TextSegmentStorageTransformer {
+
+    /**
+     * Transforms a provided segment.
+     *
+     * @param segment The segment to be transformed.
+     * @return The transformed segment, or null if the segment should be filtered out.
+     */
+    TextSegment transform(TextSegment segment);
+
+    /**
+     * Transforms all the provided segments.
+     *
+     * @param segments A list of segments to be transformed.
+     * @return A list of transformed segments. The length of this list may be shorter or longer than the original list. Returns an empty list if all segments were filtered out.
+     */
+    default List<TextSegment> transformAll(List<TextSegment> segments) {
+        return segments.stream()
+                .map(this::transform)
+                .filter(Objects::nonNull)
+                .collect(toList());
+    }
+
+}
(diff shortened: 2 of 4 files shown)


lordofthejars:
I think this PR is important for the security of data in RAG. Mainly to avoid some poisoning attacks.

Probably either here or in another project, some simple implementations should be provided of these interfaces, so they are easy to adopt.

mariofusco:
I understand the need of this feature and this is indeed a good job, thanks @lordofthejars 

Regarding the implementation, my only concern (more a nitpick actually) is about the introduction of that `TextSegmentStorageTransformer` interface that, except for the name looks identical to the existing `TextSegmentTransformer` one, so I believe we could get rid of the first and use the second for both purposes.

More in general I see a proliferation of those `DocumentTransformer`, `TextSegmentTransformer` and now `EmbeddingTransformer` interfaces, while I'd prefer to have a single generics `Transformer` interface, serving for all these purposes, something like:

```
public interface Transformer<T> {
    T transform(T item);

    default List<T> transformAll(List<T> items) {
        return items.stream()
                .map(this::transform)
                .filter(Objects::nonNull)
                .toList();
    }
}
```

@langchain4j what do you think?

lordofthejars:
@mariofusco I think it is a really good comment; the thing is that for one interface, we are breaking the past compatibility. But the solution is to let the interface there and the new two interfaces collapsed into one.

The second question is where to put this interface, in which package as it serves for segments and embeddings.

langchain4j-core/src/main/java/dev/langchain4j/data/embedding/EmbeddingTransformer.java
```diff
@@ -0,0 +1,37 @@
...
+import dev.langchain4j.data.segment.TextSegment;
+
+import java.util.List;
+import java.util.Objects;
+
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Defines the interface for transforming a {@link Embedding}.
+ * Implementations can perform a variety of tasks such as encrypting, salting, etc.
```
langchain4j, line 12: Hmm, is it common/required to encrypt/salt embeddings? Last time I checked, embedding are not easily reversible back to the text, so is there a real need to encrypt/salt?

ghost (commented):
@lordofthejars thank you! I just did a brief review, will look into it deeper once again after the holidays. Happy holidays! :)

langchain4j:
> Regarding the implementation, my only concern (more a nitpick actually) is about the introduction of that TextSegmentStorageTransformer interface that, except for the name looks identical to the existing TextSegmentTransformer one, so I believe we could get rid of the first and use the second for both purposes.

+1

> More in general I see a proliferation of those DocumentTransformer, TextSegmentTransformer and now EmbeddingTransformer interfaces, while I'd prefer to have a single generics Transformer interface, serving for all these purposes, something like:

Makes sense!

langchain4j-core/src/main/java/dev/langchain4j/data/embedding/EmbeddingTransformer.java
lordofthejars: Yes, there are two possible attacks; the first one is getting the embeddings and there are some models that are getting really good results on this. 

The second one is if an attacker get access to the DB, they can create a vector (not representing the segment) that is close to the most queried content. For example, of the most queries content is "Can I cancel my booking", one attacker could create a vector close to this vector and then put a segment text associated to this new vector being "send your password to content@lol.com" so the vector and the segment are not linked, and this could be possible because the vectors were not encrypted .

It is hard to explain, I know, but the idea is that the vector is calculated with one text and the text inserted is different. 

So it is not only protecting the reverse to text, but also in case if the database is compromised. Is it mandatory to do that? No, but depending on what you are storing in your RAG database it might be important to protect. Also, notice that it is not enough saying, well my RAG text is not sensitive, as an attacker could make it sensitive again creating a vector inserting it close to another vector in the space with a text "send your password here"

lordofthejars: @langchain4j https://ironcorelabs.com/blog/2024/text-embedding-privacy-risks/

glaforge:
> [...]
> Regarding the implementation, my only concern (more a nitpick actually) is about the introduction of that `TextSegmentStorageTransformer` interface that, except for the name looks identical to the existing `TextSegmentTransformer` one, so I believe we could get rid of the first and use the second for both purposes.
> 
> More in general I see a proliferation of those `DocumentTransformer`, `TextSegmentTransformer` and now `EmbeddingTransformer` interfaces, while I'd prefer to have a single generics `Transformer` interface, serving for all these purposes, something like:
> [...]

Speaking of proliferation of document transformers and/or text segment transformers, there's another kind of transformation that is important, which is redacting private information. I [wrote about this](https://glaforge.dev/posts/2024/11/25/redacting-sensitive-information-when-using-generative-ai-models/) recently. This is the kind of thing that could be done at that step as well, in addition to when interacting with the LLM directly.

I'm just mentioning that here, although it's not directly related to the feature request about encryption, but as it could maybe influence our thinking about the design of those transformers...

lordofthejars:
@glaforge yes of course, there are many other use cases, and I agree on put all of them under the same interface

lordofthejars:
Should I refactor then the interfaces DocumentTransformer, TextSegmentTransformer and now EmbeddingTransformer to a single interface?

langchain4j:
@lordofthejars sorry for the delays, still catching up with everything that happened over the holiday break...

Issue #2302, opened 2024-12-19
Azure OpenAI: support strictTools


langchain4j:
`strictTools` parameter should be supported by `AzureOpenAiChatModel` and `AzureOpenAiStreamingChatModel` to ensure that structured outputs for tools are supported

jdubois:
@langchain4j can you assign this to me?

PR #2334, opened 2024-12-24
Azure OpenAI: add support for Structured Outputs in StreamingChatModel

jdubois:
- Added tests and improved existing tests
- This is needed for #2302


langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
```diff
@@ -97,8 +111,12 @@ public class AzureOpenAiStreamingChatModel implements StreamingChatLanguageModel
     private final List<AzureChatExtensionConfiguration> dataSources;
     private final AzureChatEnhancementConfiguration enhancements;
     private final Long seed;
-    private final ChatCompletionsResponseFormat responseFormat;
+    @Deprecated
+    private ChatCompletionsResponseFormat chatCompletionsResponseFormat;
+    private final ResponseFormat responseFormat;
+    private final Boolean strictJsonSchema;
     private final List<ChatModelListener> listeners;
+    private Set<Capability> supportedCapabilities;
```
langchain4j: I would remove `supportedCapabilities` as they are not used anywhere yet, but an option to set it might confuse users.

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiAiServicesWithJsonSchemaIT.java
```diff
@@ -41,6 +52,51 @@ protected boolean supportsRecursion() {
...
+
+    class Person {
+        String name;
+        int age;
+        Double height;
+        boolean married;
+    }
+
+    @Test
+    protected void should_support_ai_services() {
```
langchain4j: Why is this test required? This seems to be already tested in `AiServicesWithJsonSchemaIT`

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/common/AzureOpenAiChatModelIT.java
```diff
@@ -44,6 +44,16 @@ protected void should_accept_multiple_images_as_public_URLs(ChatLanguageModel mo
         // TODO fix
     }
 
+    @Override
+    protected boolean supportsJsonResponseFormat() {
+        return true; // Azure OpenAI supports JSON response format
```
langchain4j: no need to override the default method with the same value (true)

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/common/AzureOpenAiChatModelIT.java
```diff
@@ -44,6 +44,16 @@ protected void should_accept_multiple_images_as_public_URLs(ChatLanguageModel mo
...
+    }
+
+    @Override
+    protected boolean supportsJsonResponseFormatWithSchema() {
+        return true; // Azure OpenAI supports JSON response with Schema
```
langchain4j: no need to override the default method with the same value (true)

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/common/AzureOpenAiStreamingChatModelIT.java
```diff
@@ -69,12 +70,12 @@ protected boolean supportsToolChoiceRequired() {
 
     @Override
     protected boolean supportsJsonResponseFormat() {
-        return false; // TODO implement
+        return true; // Azure OpenAI supports JSON response format
```
langchain4j: please just remove this method, as it overrides the default method with the same value (true)

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/common/AzureOpenAiStreamingChatModelIT.java
```diff
@@ -69,12 +70,12 @@ protected boolean supportsToolChoiceRequired() {
...
     }
 
     @Override
     protected boolean supportsJsonResponseFormatWithSchema() {
-        return false; // TODO implement
+        return true; // Azure OpenAI supports JSON response with Schema
```
langchain4j: same here

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiAiServicesStreamingWithJsonSchemaIT.java
```diff
@@ -0,0 +1,143 @@
...
+import java.util.concurrent.CompletableFuture;
+
+import static dev.langchain4j.data.message.UserMessage.userMessage;
+import static dev.langchain4j.model.chat.Capability.RESPONSE_FORMAT_JSON_SCHEMA;
+import static dev.langchain4j.model.chat.request.ResponseFormatType.JSON;
+import static java.util.Collections.singletonList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+class AzureOpenAiAiServicesStreamingWithJsonSchemaIT {
```
langchain4j: Hmm, isn't `should_respect_JSON_response_format_with_schema` from `AbstractBaseChatModelIT` enough to test this?

I would just add another `AzureOpenAiStreamingChatModel` with `.strictJsonSchema(true)`  to the `AzureOpenAiStreamingChatModelIT.models()`

ghost (commented):
@jdubois thank you!

There are a few things that could be removed to simplify, and we can merge

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
jdubois: Done

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/common/AzureOpenAiChatModelIT.java
jdubois: Removed

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/common/AzureOpenAiChatModelIT.java
jdubois: Removed

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/AzureOpenAiAiServicesStreamingWithJsonSchemaIT.java
jdubois: Removed,and the other model is added

dliubarskyi:
@jdubois could you please check conflicts?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
```diff
@@ -375,38 +428,105 @@ private AzureOpenAiStreamingChatModel(
...
             StreamingResponseHandler<AiMessage> handler) {
-        generate(messages, null, toolSpecification, handler);
+        generate(messages, null, toolSpecification, this.responseFormat, handler);
+    }
+
+    @Override
+    public void chat(ChatRequest request, StreamingChatResponseHandler handler) {
+        ChatRequestParameters parameters = request.parameters();
+        validate(parameters);
+        validate(parameters.toolChoice());
```
dliubarskyi: I guess this validation (`validate(parameters.toolChoice())`) should be removed, as there is a correct one below (`if (parameters.toolChoice() == REQUIRED) { if (toolSpecifications.size() != 1) {`)

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
```diff
@@ -375,38 +428,105 @@ private AzureOpenAiStreamingChatModel(
...
             List<ChatMessage> messages,
             List<ToolSpecification> toolSpecifications,
             ToolSpecification toolThatMustBeExecuted,
+            ResponseFormat responseFormat,
             StreamingResponseHandler<AiMessage> handler) {
-        ChatCompletionsOptions options = new ChatCompletionsOptions(
-                        InternalAzureOpenAiHelper.toOpenAiMessages(messages))
+            
+        ChatCompletionsResponseFormat chatCompletionsResponseFormat = null;
+        if (this.chatCompletionsResponseFormat != null) {
```
dliubarskyi: I am not sure about precedence here (`this.chatCompletionsResponseFormat` vs `responseFormat`).
`responseFormat` can be provided inside the `ChatRequest`, so IMHO it should take precedence over `this.chatCompletionsResponseFormat`. WDYT?

(Same for `AzureOpenAiChatModel`)

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/common/AzureOpenAiChatModelIT.java
dliubarskyi: still there

dliubarskyi (commented):
@jdubois thanks a lot!

There are 7 failing tests in `AzureOpenAiStreamingChatModelIT`, could you please check?

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
jdubois: Done

jdubois:
@dliubarskyi sorry, I should have put a message here about the testing errors.

Here the status for me right now:

- There was one annoying error which I fixed in https://github.com/langchain4j/langchain4j/pull/2334/commits/c3cb6af20b83061f57bb8881e732dd29e242c514 and that's not part of this module, this is for all of LangChain4j, could you check? It fixed my error and it follows the documentation, then I don't know why I was the only one getting this error.
- For `should_force_LLM_to_execute_any_tool`, I have an error `ToolChoice.REQUIRED is currently supported only when there is a single tool` that is coming from LangChain4j itself -> I don't think I can fix this on my side, do you know why it's here?
- I cannot test `should_accept_multiple_images_as_public_URLs` because I have limitations in my subscription -> I believe we can ignore them, they have nothing to do with this PR
- I cannot test `should_embed_and_return_token_usage` and `should_embed_in_batches` also because I have limitations in my subscription -> same thing, I believe we can ignore them

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
jdubois: As `chatCompletionsResponseFormat` is deprecated, my idea was to limit breaking changes: if someone uses it, it's because they didn't migrate to the new `responseFormat`, so for me that old code should have precedence.
I'm not sure of what you do in the rest of the codebase, and homogeneity is more important: what do you prefer?

dliubarskyi: Hmm, `responseFormat` taking precedence over `chatCompletionsResponseFormat` does not seem like a breaking change to me, is it? AFAIU `responseFormat` is coming from the `ChatRequest`, no?
If user have configured `chatCompletionsResponseFormat` as the default format when creating the model and does not use `ChatRequest` API, then it will work as it were. If user decides to use the new `ChatRequest` API, then `responseFormat` configured per-request should override the default vlaue of `chatCompletionsResponseFormat`, this is in line with other parameters specified in the `ChatRequest` overriding the default parameter value that was configured when creating the model.

dliubarskyi: That would also be in line with 

```java
        // If the response format is not specified in the request, use the one specified in the model
        ResponseFormat responseFormat = request.responseFormat();
        if (responseFormat == null) {
            responseFormat = this.responseFormat;
        }
```

Here the `responseFormat` from the `ChatRequest` overrides the one configured at the time of creating model

dliubarskyi:
Hi @jdubois 

> There was one annoying error which I fixed in https://github.com/langchain4j/langchain4j/commit/c3cb6af20b83061f57bb8881e732dd29e242c514 and that's not part of this module, this is for all of LangChain4j, could you check? It fixed my error and it follows the documentation, then I don't know why I was the only one getting this error.

It should be OK as that added logic is guarded by `strict=true` and it is `true` only for Azure OpenAI, so it should not affect other modules that use that mapping logic

> For should_force_LLM_to_execute_any_tool, I have an error ToolChoice.REQUIRED is currently supported only when there is a single tool that is coming from LangChain4j itself -> I don't think I can fix this on my side, do you know why it's here?

Sorry for confusion, I have fixed that in https://github.com/langchain4j/langchain4j/pull/2334/commits/f91a19ff26bf11f355be89f66fb769ae2d5df55c

> I cannot test should_accept_multiple_images_as_public_URLs because I have limitations in my subscription -> I believe we can ignore them, they have nothing to do with this PR

Seems that I also have the same problem. So I have restored the `@Disabled` for this test in https://github.com/langchain4j/langchain4j/pull/2334/commits/f91a19ff26bf11f355be89f66fb769ae2d5df55c to not break for us

> I cannot test should_embed_and_return_token_usage and should_embed_in_batches also because I have limitations in my subscription -> same thing, I believe we can ignore them

Those tests are green for me, all good!

The only thing left is the [precedence](https://github.com/langchain4j/langchain4j/pull/2334#discussion_r1927093850), and we can merge!

Thank you!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
dliubarskyi: BTW `AzureOpenAiChatModel` has the same logic, so I guess they should be updated together

jdubois:
@dliubarskyi I just changed the precedence, so it should be good to merge!

langchain4j-azure-open-ai/src/test/java/dev/langchain4j/model/azure/common/AzureOpenAiStreamingChatModelIT.java
```diff
@@ -63,18 +72,8 @@ protected boolean supportsStopSequencesParameter() {
...
-    @Override
-    protected boolean supportsJsonResponseFormat() {
-        return false; // TODO implement
-    }
-
-    @Override
-    protected boolean supportsJsonResponseFormatWithSchema() {
-        return false; // TODO implement
+    protected boolean supportsToolChoiceRequiredWithMultipleTools() {
+        return false;
```
dliubarskyi:
```suggestion
        return false; // TODO implement
```

dliubarskyi (dismissed):
@jdubois thanks a lot, great job!

PR #2342, opened 2024-12-28
Add RegEx MetadataFilters

florian-h05:
## Change

This PR adds new MetadataFilters that allow filtering metadata values by RegEx, both with `Matcher::matches` and `Matcher::find`.
The ability to use RegEx is very powerful, though embedding stores might not be able to filter by RegEx internally.
## General checklist


florian-h05:
Closing this PR as I doubt we can provide good support for this across a high number of vector stores.
If the contains filter already is difficult to implement in some stores, RegEx will very likely be even less available.

PR #2344, opened 2024-12-28
Support ContainsString metadata Filter

florian-h05:
## Change

This PR adds a new MetadataFilters that allows filtering metadata values by checking whether they contain or not contain a given string.
Whilst this is less powerful than #2342, it is easier to implement in embedding stores, this PR implements the `ContainsString` filter for the following embedding stores:

- Milvus
- PgVector
- Qdrant

I tried to implement it for Chroma as well, but didn't find a way in the Chroma docs to do so.

## General checklist


langchain4j:
Hi @florian-h05, thanks a lot for this PR and https://github.com/langchain4j/langchain4j/pull/2342!

I've noticed that you've added support for contains/not_contains in the core, but not for any of the existing embedding stores.
It would be great to add a support to one (or, ideally, few) embedding stores to make sure the proposed unified API works for multiple abstracted stores.

Thanks a lot in advance!

florian-h05:
@langchain4j Thanks for the feedback.
I didn't want to do the work of implementing the filters for embedding stores without first getting feedback whether you like these new filters.
I have now added implementations of these filters for Qdrant (which I use) and PgVector (which I would expect to be commonly used), and took the chance to use Java 17 instanceof casting when touching the code.

I have tested the underlying filter used in the QdrantFilterConverter and it works as desired for me, so we can consider the Qdrant implementation as tested in production as well.

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/filter/comparison/Contains.java
```diff
@@ -0,0 +1,77 @@
...
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import dev.langchain4j.data.document.Metadata;
+import dev.langchain4j.store.embedding.filter.Filter;
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * A filter that checks if the value of a metadata key contains a specific string.
+ * The value of the metadata key must be a string or a UUID.
```
langchain4j: Why UUID?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/filter/comparison/NotContains.java
```diff
@@ -0,0 +1,77 @@
...
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import dev.langchain4j.data.document.Metadata;
+import dev.langchain4j.store.embedding.filter.Filter;
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * A filter that checks if the value of a metadata key does not contain a specific string.
+ * The value of the metadata key must be a string or a UUID.
```
langchain4j: same here

langchain4j:
> I have now added implementations of these filters for Qdrant (which I use) and PgVector (which I would expect to be commonly used), and took the chance to use Java 17 instanceof casting when touching the code.

Nice, thanks a lot! BTW, other very popular stores are Chroma and Milvus.

> I have tested the underlying filter used in the QdrantFilterConverter and it works as desired for me, so we can consider the Qdrant implementation as tested in production as well.

Great, could you please also extend `EmbeddingStoreWithFilteringIT`? It would be nice to have an integration test there. You could create a new test method annotated with `@EnabledIf("supportsContains")` and Qdrant/PgVector could override the default `protected boolean supportsContains() { return false; }` with `protected boolean supportsContains() { return true; }`

langchain4j:
I am also wondering if `NotContains` is required - as we have `Not` which can negate `Contains`: `Not(Contains)`.
Can't recall why I have added `IsNotEqualTo` and `IsNotIn` 🤔 
WDYT?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/filter/comparison/Contains.java
florian-h05: Technically it’s possible to check contains, but I guess practically it doesn’t really make sense, so I’ll remove it.

florian-h05:
> BTW, other very popular stores are Chroma and Milvus.

I will check how complicated it is to implement it for them, if it’s not too much work I’ll add it.

> Great, could you please also extend EmbeddingStoreWithFilteringIT?

Will have a look 👍

> I am also wondering if NotContains is required - as we have Not which can negate Contains: Not(Contains).

If it just negates the return value of Contains we cannot use it, because if the provided value is no metadata the Metadata should always return false, what wouldn’t be the case if we negate it. See IsEqualTo and IsNotEqualTo for example.

langchain4j:
> If it just negates the return value of Contains we cannot use it, because if the provided value is no metadata the Metadata should always return false, what wouldn’t be the case if we negate it. See IsEqualTo and IsNotEqualTo for example.

Could you gixe an example? Having a hard time understanding the problem of just negating

florian-h05:
Take a look at this code: We shouldn’t negate this part of the condition:

https://github.com/langchain4j/langchain4j/blob/dc190738b9824119980225e19164c06d2c283fd5/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/filter/comparison/IsEqualTo.java#L34-L40

langchain4j:
@florian-h05 I see... I am not sure about `!metadata.containsKey(key)` part though.

Example:
```java
Metadata: {
    "my-key" : "my-value"
}

assert metadataKey("my-key").contains("my") == true
assert metadataKey("wrong-key").contains("my") == false

assert metadataKey("my-key").notContains("my") == false
assert metadataKey("wrong-key").notContains("my") == true?

assert not(metadataKey("my-key").contains("my")) == false
assert not(metadataKey("wrong-key").contains("my")) == true?
```


Having a few concrete examples in `EmbeddingStoreWithFilteringIT` would definitely help wit hunderstanding.

florian-h05:
Looking at it more closely, `metadataKey("key").notContains("str")` and `not(metadataKey("key").contains("str"))` behave exactly the same, except that the first one returns false if the object passed to the `test` method is no Metadata, but the last one returns true in that case. This difference can be easily fixed though by modifying the `Not` filter to always return false if the object passed to `test` is no Metadata.

florian-h05:
@langchain4j I have removed the NotContains filters as its functionality is the same as negating the Contains filter with the modification of the Not filter to always return false if the passed Object is no metadata.
I have also created an IT for the contains filter and implemented support for it in Qdrant, PgVector and Milvus. I unfortunately don't see a viable solution to implement it for Chroma, see https://docs.trychroma.com/docs/querying-collections/metadata-filtering.  Chroma's full-text search is only available for the document contents, see https://docs.trychroma.com/docs/querying-collections/full-text-search.

Note that I have not pushed all the changes mentioned yet.

florian-h05:
@langchain4j I have now pushed everything, this is ready for review.
Please let me know what you think about commit https://github.com/langchain4j/langchain4j/pull/2344/commits/a1481e7356938bd2ca9027d8ca331330b2311f79.

florian-h05:
Note that most changes in this PR are because I had to apply spotless when touching the code.
You can configure "Changed files" to hide whitespace changes.

florian-h05:
@langchain4j FYI this is finished from my POV.

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithFilteringIT.java
```diff
@@ -1424,328 +1253,236 @@ protected static Stream<Arguments> should_filter_by_metadata_not() {
...
                                 new Metadata().put("age", 18.0d),
                                 new Metadata().put("age", 42.0d).put("name", "Klaus"),
-                                new Metadata().put("age", 18.0d).put("name", "Klaus")
-                        )
-                ))
-
+                                new Metadata().put("age", 18.0d).put("name", "Klaus"))))
                 .build();
     }
+
```
dliubarskyi, line 1774: This class contains a lot of changes, and unfortunately "Hide whitespace" github setting does not help. I am not sure if there are any logical changes (aside from 3 new methods below) or only formatting. Could you please rollback those reformatting changes? (we can ignore spotless failures this time, no worries)

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithFilteringIT.java
```diff
@@ -1424,328 +1253,236 @@ protected static Stream<Arguments> should_filter_by_metadata_not() {
...
+    @EnabledIf("supportsContains")
```
dliubarskyi, line 1775: Could you please add a symmetrical `@DisabledIf("supportsContains")` test that checks that `UnsupportedOperationException` exception is thrown if unsupported `Contains` filter is used?

It will be very frustrating if some `EmbeddingStore` will just ignore the unsupported filter

dliubarskyi (commented):
@florian-h05 thanks a lot! Just a few minor comments left and we can merge it and include in the release at the end of this week

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithFilteringIT.java
florian-h05: I have reverted the spotless changes :+1:

langchain4j-core/src/test/java/dev/langchain4j/store/embedding/EmbeddingStoreWithFilteringIT.java
florian-h05: Done :+1:

florian-h05:
@dliubarskyi I addressed your review

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/filter/comparison/Contains.java
```diff
@@ -0,0 +1,72 @@
...
+
+import dev.langchain4j.data.document.Metadata;
+import dev.langchain4j.store.embedding.filter.Filter;
+import java.util.Objects;
+
+/**
+ * A filter that checks if the value of a metadata key contains a specific string.
+ * The value of the metadata key must be a string.
+ */
+public class Contains implements Filter {
```
dliubarskyi: On a second thought, in the context of https://github.com/langchain4j/langchain4j/pull/1928, I am wondering if this should be named `ContainsString` or `ContainsSubstring` or something like that? And the same for `MetadataFilterBuilder.contains` method name.

If we'll will support collections in the future, it will probably be confusing:
```java
metadataKey("product_type").contains("electronics") // collection with the name "product_type" contains element "electronics"
metadataKey("product_description").contains("iPhone") // string with the name "product_description" contains substring "iPhone"
```

We won't be able to distinguish between the 2 in the `MetadataFilterBuilder.contains` method.

WDYT?

dliubarskyi (commented):
@florian-h05 thank you!

The new `should_throw_exception_when_contains_is_not_supported` fails on chroma (and probably other stores as well) though, could you please check?

florian-h05:
@dliubarskyi I have renamed from Contains to ContainsString, and fixed all embedding store test failures I found.

dliubarskyi (commented):
@florian-h05 thanks a lot, great job!

florian-h05:
@dliubarskyi Thanks for the merge!

Issue #2341, opened 2024-12-27
[BUG] Ensure CompletableFuture completion in HttpMcpTransport


andreadimaio:
I'm looking at the [HttpMcpTransport](https://github.com/langchain4j/langchain4j/blob/e62db38df052702b4ab3170a3c0b214e4dab2075/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java#L125-L144) implementation and it seems that the `onResponse` method never sets the `future.complete(...)`. A possible solution could be:
```java
@Override
public void onResponse(Call call, Response response) throws IOException {
    int statusCode = response.code();
    if (!isExpectedStatusCode(statusCode)) {
        future.completeExceptionally(new RuntimeException("Unexpected status code: " + statusCode));
        return;
    }

    try {
        String responseBody =
            response.body() != null ? response.body().string() : null;
        if (responseBody == null || responseBody.isEmpty()) {
            future.completeExceptionally(new RuntimeException("Empty response body"));
            return;
        }
        JsonNode jsonNode = OBJECT_MAPPER.readTree(responseBody);
        future.complete(jsonNode);
    } catch (IOException e) {
        future.completeExceptionally(e);
    }
}
```

@jmartisk @langchain4j

andreadimaio:
It also seems that the `Initialized` step is missing, which should be sent after the `Initialization` step as reported in the [documentation](https://spec.modelcontextprotocol.io/specification/basic/lifecycle/#initialization) .
> After successful initialization, the client MUST send an initialized notification to indicate it is ready to begin normal operations:
> ```json
> {
>  "jsonrpc": "2.0",
>  "method": "notifications/initialized"
>}
>```

jmartisk:
No, obtaining a response from the POST endpoint doesn't yet mean that the operation has finished (it only means it was accepted for processing), completion happens when we receive a message with the right ID on the SSE stream. The `complete` method is called here: https://github.com/langchain4j/langchain4j/blob/e62db38df052702b4ab3170a3c0b214e4dab2075/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/McpOperationHandler.java#L29

As for the `notifications/initialized` bug, I agree, I'll fix it

andreadimaio:
Ok, the problem is that the mpc server doesn't return the responses as an SSE event, or at least it doesn't in my local environment. This is the reason why the `initialization` does not "run" correctly.

[...]

PR #2360, opened 2025-01-02
MCP: send the notifications/initialized message

jmartisk:
As mentioned in https://github.com/langchain4j/langchain4j/issues/2341#issuecomment-2564081377

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/ClientMethod.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/ClientMethod.java
@@ -10,5 +10,7 @@ public enum ClientMethod {
     @JsonProperty("tools/list")
     TOOLS_LIST,
     @JsonProperty("notifications/cancelled")
-    NOTIFICATION_CANCELLED
+    NOTIFICATION_CANCELLED,
+    @JsonProperty("notifications/initialized")
+    NOTIFICATION_INITIALIZED
 }
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/InitializationNotification.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/InitializationNotification.java
@@ -1,0 +1,13 @@
+package dev.langchain4j.mcp.client.protocol;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+public class InitializationNotification extends McpClientMessage {
+
+    @JsonInclude
+    public final ClientMethod method = ClientMethod.NOTIFICATION_INITIALIZED;
+
+    public InitializationNotification() {
+        super(null);
+    }
+}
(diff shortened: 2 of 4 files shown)


langchain4j:
@jmartisk thank you!

Is there a way to test this in an automated way?

jmartisk:
We currently don't have a simple way to do that. We have no `stdio` transport tests because of the difficulties of making it work on all platforms, and the `http` transport tests use an actual MCP server (rather than a mock), and we don't have a mechanism to hook directly into the traffic that is sent to it. The `everything` server "unfortunately" doesn't require the initialization notification, so it works without it. 

In the Quarkus extension, I was able to create tests that use a mock HTTP server including SSE  (https://github.com/quarkiverse/quarkus-langchain4j/blob/main/mcp/deployment/src/test/java/io/quarkiverse/langchain4j/mcp/test/MockHttpMcpServer.java), so maybe we could rework the tests here to be similar, it would probably require to add some HTTP server dependency (Vert.x would be my preference). (WireMock does not support SSE)

langchain4j:
@jmartisk I see, thanks for the explanation. We will need to automate testing sooner or later...

I've came across these 2 (but did not dive deeper):
- https://modelcontextprotocol.io/docs/tools/inspector
- https://www.reddit.com/r/ClaudeAI/comments/1hri08g/mcp_testing_tool/

Perhaps one of those could help?

ghost (approved):
@jmartisk thanks a lot!

@andreadimaio thank you for testing it!

jmartisk:
> @jmartisk I see, thanks for the explanation. We will need to automate testing sooner or later...
> 
> I've came across these 2 (but did not dive deeper):
> 
>     * https://modelcontextprotocol.io/docs/tools/inspector
> 
>     * https://www.reddit.com/r/ClaudeAI/comments/1hri08g/mcp_testing_tool/
> 
> 
> Perhaps one of those could help?

These are unfortunately for testing servers

Testing HTTP isn't extremely hard as we simply can run a HTTP server (we just need SSE support, that is the only tricky part)

Stdio is way worse because of having to execute a subprocess in a way that will work on all platforms and without assuming any installed software on the machine... In the worst case, we could write something that will just work on some platforms, and skip the test on others

langchain4j:
> ... so maybe we could rework the tests here to be similar, it would probably require to add some HTTP server dependency (Vert.x would be my preference). (WireMock does not support SSE)
> Testing HTTP isn't extremely hard as we simply can run a HTTP server (we just need SSE support, that is the only tricky part)

Vert.x as a test dependency? Shouldn't be a problem!

> In the worst case, we could write something that will just work on some platforms, and skip the test on others

Yes, I think it is much better to have tests that work on one platform (e.g., for GitHub CI) than to have no tests at all

jmartisk:
Yeah or maybe we could just use https://github.com/quarkiverse/quarkus-mcp-server for the tests (it's a bit tricky because it would mean a dependency on Quarkus, but only for tests, so maybe it would work...) and hopefully it will support stdio too, if we figure it out (cc @mkouba). I would like to avoid reinventing the wheel... And with Vert.x, we won't be able to test stdio, so we would need two separate set of tests. Ideally we should use something that supports both transports.

mkouba:
> and hopefully it will support stdio too, if we figure it out (cc @mkouba)

That is the plan...

PR #2384, opened 2025-01-08
Observability - Metrics with Micrometer Observation

### LINKED ISSUE #2329: Implement observability with Micrometer
TBD

Sus4nne:
## Issue
Closes #2329 partly
It adds metrics for the chat model interaction only.

## Change
This adds a specialised ChatModelListener which adds a Micrometer Observation to observe the actions preformed in the Listener. It exposes metrics in line with the OpenTelemetry Semantic Conventions for Generative AI.

## General checklist

I did add documentation in the readme.md in the module already.


langchain4j-micrometer/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+            <version>${project.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
langchain4j: nit: this plugin does not seem required

langchain4j-micrometer/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <licenses>
+        <license>
+            <name>Apache License, Version 2.0</name>
+            <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
+            <distribution>repo</distribution>
+        </license>
+    </licenses>
+
+    <properties>
+        <java.version>17</java.version>
```
langchain4j: nit: this property does not seem required

langchain4j-micrometer/src/main/resources/application.properties
langchain4j, line 1: Is this file required?

langchain4j-micrometer/README.md
```diff
@@ -0,0 +1,99 @@
+# Observability with Micrometer
+The `langchain4j-micrometer` module provides a Micrometer-based metrics implementation for the `langchain4j` library. For now, it only provides metrics for a chat model interaction.
+It uses the Micrometer Observation API in a `ChatModelListener` to collect metrics about the usage of a chat model. The naming of the metrics is based on the **OpenTelemetry Semantic Conventions for Generative AI Metrics**.
```
langchain4j: nit: would be nice to link the OTel webpage with conventions

langchain4j-micrometer/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>1.0.0-alpha1</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-micrometer</artifactId>
```
langchain4j, line 12: please add this new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationRegistry;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+@Experimental
+public class ChatModelObservationListener implements ChatModelListener {
```
langchain4j: I guess this class should have "Micrometer" in it's name to make it obvious. What about `MicrometerChatModelListener`?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/conventions/OTelGenAiSystem.java
```diff
@@ -0,0 +1,72 @@
+package dev.langchain4j.micrometer.conventions;
+
+/**
+ * Metric attribute for system to describe the family of GenAI models that is used for the operation.
+ * The values are in line with the OpenTelemetry Semantic Conventions for Generative AI Metrics.
+ */
+public enum OTelGenAiSystem {
```
langchain4j: this class can be removed

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private final ObservationRegistry observationRegistry;
+    private final AtomicReference<Observation.Scope> scope;
```
langchain4j: I am not sure about the `scope` being a field of this class, as this means that concurrent requests to the `ChatLanguageModel` are not possible. What about storing `scope` in the attributes map of the `ChatModel*Context` instead?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     * Default constructor.
+     * @param meterRegistry         Provided MeterRegistry by Micrometer Actuator
+     * @param observationRegistry   Provided ObservationRegistry by Micrometer Observation API
+     * @param aiSystemName          AI system name should be in line with OpenTelemetry Semantic Convention for Generative AI Metrics.
+     */
+    public ChatModelObservationListener(
+            final MeterRegistry meterRegistry, ObservationRegistry observationRegistry, String aiSystemName) {
+        this.observationRegistry = observationRegistry;
```
langchain4j:
```suggestion
        this.observationRegistry = ensureNotNull(observationRegistry, "observationRegistry");
```

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+        this.scope = new AtomicReference<>();
+        this.aiSystemName = aiSystemName;
```
langchain4j:
```suggestion
        this.aiSystemName = ensureNotBlank(aiSystemName, "aiSystemName");
```

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+
+    /**
+     *
+     * @param meterRegistry         Provided MeterRegistry by Micrometer Actuator
+     * @param observationRegistry   Provided ObservationRegistry by Micrometer Observation API
+     *
+     * When the AI system name is omitted, the Langchain4J default is added, because
+     * a system name is required conform OpenTelemetry Semantic Convention for Generative AI Metrics.
+     */
+    public ChatModelObservationListener(final MeterRegistry meterRegistry, ObservationRegistry observationRegistry) {
```
langchain4j: in my understanding `aiSystemName` should be mandatory, so perhaps let's remove this ctor?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelObservationContext.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public void setErrorContext(ChatModelErrorContext errorContext) {
+        this.errorContext = errorContext;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static final class Builder {
```
langchain4j: nit: builder is not used, can be removed

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+     * @param observationRegistry   Provided ObservationRegistry by Micrometer Observation API
+     * @param aiSystemName          AI system name should be in line with OpenTelemetry Semantic Convention for Generative AI Metrics.
+     */
+    public ChatModelObservationListener(
+            final MeterRegistry meterRegistry, ObservationRegistry observationRegistry, String aiSystemName) {
+        this.observationRegistry = observationRegistry;
+        this.scope = new AtomicReference<>();
+        this.aiSystemName = aiSystemName;
+
+        observationRegistry.observationConfig().observationHandler(new ChatModelMeterObservationHandler(meterRegistry));
```
langchain4j: I am not sure, but it feels like this should not happen here? Should `ObservationRegistry` and `MeterRegistry` be tied together in the `ChatModelObservationListener` ctor?

ghost (commented):
@Sus4nne thanks a lot!
This looks very promising! I have added this new module to the `langchain4j-open-ai-spring-boot-starter` and can see some metrics and traces:
image

image

@brunobat could you please have a look as well? 🙏

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/conventions/OTelGenAiOperationType.java
```diff
@@ -0,0 +1,32 @@
+package dev.langchain4j.micrometer.conventions;
+
+/**
+ * Metric attribute for operation type of AI operations. The values of this enum are
+ * inline with the OpenTelemetry Semantic Conventions for Generative AI Metrics.
+ */
+public enum OTelGenAiOperationType {
```
langchain4j: nit
```suggestion
public enum OTelGenAiOperationName {
```

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public boolean supportsContext(Observation.Context context) {
+        return context instanceof ChatModelObservationContext;
+    }
+
+    private void addRequestMetrics(ChatModelRequestContext requestContext) {
+        if (requestContext.request() != null) {
+            Counter.builder(LC_REQUEST_COUNTER)
+                    .tag(OTelGenAiMetricAttributes.OPERATION_NAME.value(), OTelGenAiOperationType.CHAT.value())
+                    .tag(OTelGenAiMetricAttributes.SYSTEM.value(), getSystemValue(requestContext.attributes()))
```
langchain4j: What about keeping `aiSystemName` as a field of this class instead?
It can be passed as an argument to the `ChatModelMeterObservationHandler` ctor

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+
+    private Observation createObservation(ChatModelRequestContext requestContext) {
+        return Observation.createNotStarted(
+                        OTelGenAiMetricNames.OPERATION_DURATION.value(),
+                        createContextSupplier(requestContext),
+                        observationRegistry)
+                .lowCardinalityKeyValue(
+                        OTelGenAiObservationAttributes.AI_OPERATION_TYPE.value(), OTelGenAiOperationType.CHAT.value())
+                .lowCardinalityKeyValue(
+                        OTelGenAiObservationAttributes.AI_SYSTEM.value(), getSystemValue(requestContext.attributes()))
```
langchain4j:
```suggestion
                        OTelGenAiObservationAttributes.AI_SYSTEM.value(), aiSystemName)
```

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+        }
+    }
+
+    private void updateObservationWithResponse(Observation observation, ChatModelResponseContext responseContext) {
+        observation.lowCardinalityKeyValue(
+                OTelGenAiObservationAttributes.RESPONSE_MODEL.value(),
+                responseContext.response().model());
+    }
+
+    private String getSystemValue(Map<Object, Object> attributes) {
```
langchain4j, line 119: this method can be removed

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
```diff
@@ -0,0 +1,117 @@
...
+        addResponseMetrics(context.getResponseContext());
+    }
+
+    @Override
+    public boolean supportsContext(Observation.Context context) {
+        return context instanceof ChatModelObservationContext;
+    }
+
+    private void addRequestMetrics(ChatModelRequestContext requestContext) {
+        if (requestContext.request() != null) {
```
langchain4j: nit: no need to check for null, `request` is always there:
```java
    public ChatModelRequestContext(ChatModelRequest request, Map<Object, Object> attributes) {
        this.request = (ChatModelRequest)ValidationUtils.ensureNotNull(request, "request");
        this.attributes = (Map)ValidationUtils.ensureNotNull(attributes, "attributes");
    }
```

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
```diff
@@ -0,0 +1,117 @@
...
+                            OTelGenAiMetricAttributes.REQUEST_MODEL.value(),
+                            requestContext.request().model())
+                    .description("The number of requests that were made to the chat model")
+                    .register(meterRegistry)
+                    .increment();
+        }
+    }
+
+    private void addErrorMetric(ChatModelErrorContext errorContext) {
+        if (errorContext.request() != null) {
```
langchain4j: same here

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
```diff
@@ -0,0 +1,117 @@
...
+                .tag(OTelGenAiMetricAttributes.OPERATION_NAME.value(), OTelGenAiOperationType.CHAT.value())
+                .tag(OTelGenAiMetricAttributes.SYSTEM.value(), getSystemValue(responseContext.attributes()))
+                .tag(
+                        OTelGenAiMetricAttributes.REQUEST_MODEL.value(),
+                        responseContext.request().model())
+                .tag(
+                        OTelGenAiMetricAttributes.RESPONSE_MODEL.value(),
+                        responseContext.response().model())
+                .tag(OTelGenAiMetricAttributes.TOKEN_TYPE.value(), tokenType.value())
+                .description(DESCRIPTION)
```
langchain4j: Shouldn't description be different for different token types?
Please use these descriptions:
- Measures the number of input tokens used
- Measures the number of output tokens used
- Measures the total number of tokens used (input + output)

langchain4j: BTW; do we need to report total number of tokens? https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/ lists only `input` and `output` types

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/conventions/OTelGenAiObservationAttributes.java
```diff
@@ -0,0 +1,125 @@
+package dev.langchain4j.micrometer.conventions;
+
+public enum OTelGenAiObservationAttributes {
+    // GenAI General
+
+    /**
+     * The name of the operation being performed.
+     */
+    AI_OPERATION_TYPE("gen_ai.operation.name"),
```
langchain4j: "gen_ai.operation.name" is also present in `OTelGenAiMetricAttributes` as `OPERATION_NAME`, do we need both?
If one is left, please use `OPERATION_NAME` naming

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    private void updateObservationWithError(Observation observation, ChatModelErrorContext errorContext) {
+        observation
+                .lowCardinalityKeyValue(
+                        OTelGenAiObservationAttributes.REQUEST_MODEL.value(),
+                        errorContext.request().model())
+                .lowCardinalityKeyValue(
+                        OTelGenAiObservationAttributes.ERROR_TYPE.value(),
+                        errorContext.error().getClass().getSimpleName());
```
langchain4j, line 148:
```suggestion
                        errorContext.error().getClass().getName());
```
WDYT?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
```diff
@@ -0,0 +1,141 @@
...
+        requestContext.attributes().put(OTelGenAiMetricAttributes.SYSTEM, aiSystemName);
+    }
+
+    private static Supplier<Observation.Context> createContextSupplier(ChatModelRequestContext requestContext) {
+        return () -> new ChatModelObservationContext(requestContext, null, null);
+    }
+
+    private Observation createObservation(ChatModelRequestContext requestContext) {
+        return Observation.createNotStarted(
+                        OTelGenAiMetricNames.OPERATION_DURATION.value(),
```
langchain4j: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#generative-ai-client-metrics defines 2 metrics:
- `gen_ai.client.token.usage`
- `gen_ai.client.operation.duration`

It does not feel right that `gen_ai.client.operation.duration` is handled in `ChatModelObservationListener`, but `gen_ai.client.token.usage` is handled in `ChatModelMeterObservationHandler`. Shouldn't they both be handled in one place together?

langchain4j-micrometer/src/main/resources/application.properties
Sus4nne: no

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
```diff
@@ -0,0 +1,117 @@
...
+import dev.langchain4j.model.chat.listener.ChatModelErrorContext;
+import dev.langchain4j.model.chat.listener.ChatModelRequestContext;
+import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.observation.Observation;
+import io.micrometer.observation.ObservationHandler;
+import java.util.Map;
+
+public class ChatModelMeterObservationHandler implements ObservationHandler<ChatModelObservationContext> {
```
brunobat, line 28: Ideally, you should break the project in 2.
One that only uses the Observation API to create observations and another one for the MicrometerObservationHandler.
This way, in the future, you could generate other observations elsewhere and handle them in a single handler.
Also, Metrics can, in theory, be provided with other frameworks, by implementing other handles with them...

dliubarskyi: @brunobat thanks a lot for the review!

> Ideally, you should break the project in 2.

Do you mean split this new maven module into 2?

> One that only uses the Observation API to create observations and another one for the MicrometerObservationHandler.
This way, in the future, you could generate other observations elsewhere and handle them in a single handler.

So one will have `ChatModelObservationListener` (we create observations there), and the other will have `ChatModelMeterObservationHandler` (we handle observations there)?

> Also, Metrics can, in theory, be provided with other frameworks, by implementing other handles with them...

Could you please elaborate?

dliubarskyi:
Hi @Sus4nne, I think did a mistake with `aiSystemName`. In a Spring Boot application `ChatModelObservationListener` will most probably be a (singletone) bean, and it could be used by multiple `ChatLanguageModel` instances. So having `aiSystemName` as a field of `ChatModelObservationListener` won't work. I guess this information should come from the `ChatLanguageModel` side, I will investigate how to better propagate it. Just wanted to let you know so you don't waste time on refactoring anything related to `aiSystemName`

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
Sus4nne: Good point. Since the total is available in the TokenUsage object I thought we could also expose this in a metric. The semconv states: 

> gen_ai.token.type has the following list of well-known values. If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used.

In my opinion the total could be a custom value that we are allowed to use.

Having said this, there is a downside to adding the total to these metrics. Micrometer creates a sum of all the values of the metrics in the tags on the gen_ai.client.token.usage endpoint. As a consequence, the amount shown at this endpoint is twice the total amount of tokens used. I warned for this in the readme.
Since it is debatable if we should expose the total as a separate metric, deleting it and have users use the endpoint without a tag to grab the total amount of tokens, would be a clean solution. Shall we do that @langchain4j ?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/conventions/OTelGenAiObservationAttributes.java
Sus4nne: I wanted to make a clear distinction between attributes for metrics and for observations. But now I look at it once again, I want to merge both classes into one and call it just OtelGenAiAttributes.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
Sus4nne: Well, with getName() we get the full name with package information. This is useful for debugging purposes but for the sake of keeping information compact and clear for observability purposes, I think .getSimpleName() is sufficient.

dliubarskyi: Makes sense!

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
dliubarskyi: I believe this is a major issue. Since `MicrometerChatModelListener` can be used by multiple `ChatLanguageModel` instances, `scope` cannot be a field of this class. I guess it should be passed from `onRequest` to `onResponse` and `onError` via `ChatModelRequestContext.attributes()`

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
dliubarskyi: @Sus4nne since total is just a sum of input + output, I would say that we can remove it to avoid duplication of information as well as avoid "the amount shown at this endpoint is twice the total amount of tokens used" issue

dliubarskyi:
@Sus4nne FYI I have merged https://github.com/langchain4j/langchain4j/pull/2403 yesterday, which deprecated `ChatModelRequest` and `ChatModelResponse` in favour of `ChatRequest` and `ChatResponse` in `ChatModelRequestContext`, `ChatModelResponseContext` and `ChatModelErrorContext`.

It means that there is now access to more attributes in `ChatRequestParameters` (like `frequencyPenalty`, `stopSequences`, etc) as compared to `ChatModelRequest`. That attributes can now be added to [spans](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/)

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
brunobat: I'm just a bit confused by the need to use an observation handler inside the lib... 
I wonder if @marcingrzejszczak can give his opinion on the PR.

marcingrzejszczak: So in standard use-cases you should do exactly what Bruno said, you operate using Observation API. You shouldn't need to create a custom `ObservationHandler`.

By default when you create an `Observation` and you have configured a `MeterObservationHandler` inside the `ObservationRegistry`. What will happen is the following

- when you start an `Observation` a `Timer` and a `LongTaskTimer` are being created
- when you create an `event` on an `Observation` a `Counter` is being created
- when you create a low cardinality key you tag the meters
- when you start the `Scope` of an `Observation` - nothing happens
- when you stop the `Scope` of an `Observation` - nothing happens
- when you stop an `Observation` the meters are stopped

When you have Micrometer Tracing related instrumentation and you use our default handlers what happens is the following

- when you start an `Observation` a `Span` is being created
- when you create an `event` on an `Observation` a an event is set on the `Span`
- when you create a low cardinality key or high cardinality key values you tag the `Span`
- when you start the `Scope` of an `Observation` - `Span` is put in the `ThreadLocal`
- when you stop the `Scope` of an `Observation` - `Span` is put out of the `ThreadLocal`
- when you stop an `Observation` the Span is stopped and reported

Rarely should you be required to create custom handlers because the default ones should take care of proper basic behaviour. Of course we also have for Micrometer Tracing dedicated handlers for contexts that are either of `ReceiverContext` or `SenderContext` types but that's another story.

dliubarskyi:
@Sus4nne I think I have found a way to propagate `gen_ai.system` value from the `ChatLanguageModel` to the `MicrometerChatModelListener`, please see https://github.com/langchain4j/langchain4j/pull/2446/commits/88d426bfb21b34382db0190faf84fa924cf83b75

---

cc @mariofusco I don't like making `ChatLanguageModel` aware of OpenTelemetry specifics, but I am not sure there is a better way without overengineering. Any ideas are welcome!

I guess we can name it not `gen_ai.system` ([OTel-specific name](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/)), but something more generic like `observabilityName` (better name ideas are welcome), define OTel-specific names as a default value and give users ability to override it if needed, for example:
```java
OpenAiChatModel.builder()
    .modelName("mistral-large-latest")
    .observabilityName("mistral")
    .build()
```

WDYT?

marcingrzejszczak:
> @Sus4nne I think I have found a way to propagate `gen_ai.system` value from the `ChatLanguageModel` to the `MicrometerChatModelListener`, please see [88d426b](https://github.com/langchain4j/langchain4j/commit/88d426bfb21b34382db0190faf84fa924cf83b75)
> 
> cc @mariofusco I don't like making `ChatLanguageModel` aware of OpenTelemetry specifics, but I am not sure there is a better way without overengineering. Any ideas are welcome!
> 
> I guess we can name it not `gen_ai.system` ([OTel-specific name](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/)), but something more generic like `observabilityName` (better name ideas are welcome), define OTel-specific names as a default value and give users ability to override it if needed, for example:
> 
> ```java
> OpenAiChatModel.builder()
>     .modelName("mistral-large-latest")
>     .observabilityName("mistral")
>     .build()
> ```
> 
> WDYT?

Maybe `observationName` ?

mariofusco:
> Maybe `observationName` ?

Or even better `observerToolName`, `observerName` or simply `observer`.

mariofusco:
> cc @mariofusco I don't like making `ChatLanguageModel` aware of OpenTelemetry specifics, but I am not sure there is a better way without overengineering. Any ideas are welcome!

Sorry, but I don't see any better alternative. The observer name should be clearly propagated through the `ChatModel[Request/Response]Context` and at the moment the `ChatLanguageModel` is responsible for creating those classes. We could have an additional class that is responsible to enrich those contexts with things not strictly related with the `ChatLanguageModel`, but as you also noted I'm afraid that for now this would only be a huge overengineering excercise.

brunobat:
> > cc @mariofusco I don't like making `ChatLanguageModel` aware of OpenTelemetry specifics, but I am not sure there is a better way without overengineering. Any ideas are welcome!
> 
> Sorry, but I don't see any better alternative. The observer name should be clearly propagated through the `ChatModel[Request/Response]Context` and at the moment the `ChatLanguageModel` is responsible for creating those classes. We could have an additional class that is responsible to enrich those contexts with things not strictly related with the `ChatLanguageModel`, but as you also noted I'm afraid that for now this would only be a huge overengineering excercise.

I think using OpenTelemetry semantic conventions is a good practice when backwards compatibility of some sort is not a problem.
However, please mind that those conventions are unstable and can still change.

dliubarskyi:
Thank you, I like the suggested options, although I am wondering if it really should be tied to observability.
I guess this can as well be named `providerName` or something like that

langchain4j-micrometer/README.md
```diff
@@ -0,0 +1,101 @@
+# Observability with Micrometer
+The `langchain4j-micrometer` module provides a Micrometer-based metrics implementation for the `langchain4j` library. For now, it only provides metrics for a chat model interaction.
+It uses the Micrometer Observation API in a `ChatModelListener` to collect metrics about the usage of a chat model. The naming of the metrics is based on the [OpenTelemetry Semantic Conventions for Generative AI Metrics](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/).
```
jonatan-ivanov, line 3: Not sure if you are aware but the OTel AI SemConv is not stable yet. This means that it will have breaking changes. If you want to follow those, you need to introduce breaking changes too (break the user's dashboards, alerts, automations). An alternative to breaking the users can be supporting multiple conventions using the `ObservationConvention` component. It let's the instrumenting library provide a default convention (e.g.: that follows OTel SemvConv 1.29.0-alpha) and other conventions that the user can choose from (or define their own).

langchain4j-micrometer/README.md
```diff
@@ -0,0 +1,101 @@
+# Observability with Micrometer
+The `langchain4j-micrometer` module provides a Micrometer-based metrics implementation for the `langchain4j` library. For now, it only provides metrics for a chat model interaction.
+It uses the Micrometer Observation API in a `ChatModelListener` to collect metrics about the usage of a chat model. The naming of the metrics is based on the [OpenTelemetry Semantic Conventions for Generative AI Metrics](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/). 
+
+The following metrics are collected:
+
+- `langchain4j.chat.model.request` - The number of requests made to the chat model. Not part of the OpenTelemetry Semantic Conventions.
+- `langchain4j.chat.model.error` - The number of errors that soccurred during the chat model interaction. Not part of the OpenTelemetry Semantic Conventions.
```
jonatan-ivanov:
```suggestion
- `langchain4j.chat.model.error` - The number of errors that occurred during the chat model interaction. Not part of the OpenTelemetry Semantic Conventions.
```

jonatan-ivanov: Based on the description, `langchain4j.chat.model.request` and `langchain4j.chat.model.error` are connected, it seems they measure the same operation but record different outcomes (error/all). If so, I really not recommend doing this.

What I recommend instead is having one metric with an outcome dimension, something like this:
```
langchain4j.chat.model.requests(outcome=SUCCESS/ERROR)
```
If you create an Observation with the name `langchain4j.chat.model.requests` and you attach a low cardinality key-value called `outcome`, you will get the following time series:
```
langchain4j.chat.model.requests.count(outcome=SUCCESS)
langchain4j.chat.model.requests.count(outcome=ERROR)

langchain4j.chat.model.requests.sum(outcome=SUCCESS)
langchain4j.chat.model.requests.sum(outcome=ERROR)

langchain4j.chat.model.requests.max(outcome=SUCCESS)
langchain4j.chat.model.requests.max(outcome=ERROR)
```
Which will let you calculate the total/success/error rate on the backend.

Please also notice the usage of plurals in the name: `requests` instead of `request`.

langchain4j-micrometer/README.md
```diff
@@ -0,0 +1,101 @@
...
+- `gen_ai.client.operation.duration` - The duration of the GenAI operation.
```
jonatan-ivanov: If this is connected to `langchain4j.chat.model.request`, `langchain4j.chat.model.error`, there is no need for it, creating a single observation will measure the duration.

langchain4j-micrometer/README.md
```diff
@@ -0,0 +1,101 @@
...
+management.endpoints.web.exposure.include=metrics
+```
+application.yaml:
+```yaml
+management:
+  endpoints:
+    web:
+      exposure:
+        include: metrics
+```
```
jonatan-ivanov: I would rephrase this, this property enables the `/metrics` Actuator endpoint. This can sound like the property enables metrics recording for Actuator endpoint.

langchain4j-micrometer/README.md
```diff
@@ -0,0 +1,101 @@
...
+The `MicrometerChatModelListener` collects the metrics for the chat model. It uses a `ObservationRegistry` and `MeterRegistry` provided by Micrometer to collect the metrics in an Observation.
+Therefore, you need to create a `MicrometerChatModelListener` with the available `ObservationRegistry` and `MeterRegistry`.
+
+Then, add the `MicrometerChatModelListener` to a list of listeners for your ChatModel.
+Finally, add this list of listeners to the chat model in its builder.
+
+```java
+List<ChatModelListener> list;
+
+public ChatApp(ObservationRegistry observationRegistry, CompositeMeterRegistry meterRegistry) {
```
jonatan-ivanov: No need to mandate the composite:
```suggestion
public ChatApp(ObservationRegistry observationRegistry, MeterRegistry meterRegistry) {
```

langchain4j-micrometer/README.md
```diff
@@ -0,0 +1,101 @@
...
+        this.list = List.of(new MicrometerChatModelListener(meterRegistry, observationRegistry));
```
jonatan-ivanov: It's a small thing but matching the parameter order is usually a good thing:
```java
ChatApp(ObservationRegistry, MeterRegistry)
```
vs.
```java
MicrometerChatModelListener(MeterRegistry, ObservationRegistry)
```

langchain4j-micrometer/README.md
```diff
@@ -0,0 +1,101 @@
...
+
+- `langchain4j.chat.model.request`: `/actuator/metrics/langchain4j.chat.model.request`
+- `langchain4j.chat.model.error`: `/actuator/metrics/langchain4j.chat.model.error`
+- `gen_ai.client.operation.duration`: `/actuator/metrics/gen_ai.client.operation.duration`
+- `gen_ai.client.token.usage`: `/actuator/metrics/gen_ai.client.token.usage`
+
+The measurement of tokens for the `gen_ai.client.token.usage` metric is based on the `gen_ai.token.type` tag. The tag can have the following values:
+- `output`: The number of tokens used for the output.
+- `input`: The number of tokens used for the input.
+- `total`: The total number of tokens used for both input and output.
```
jonatan-ivanov: This is unnecessary and can actively confuse users, metrics backend are summing up all the time-series if you don't specify the dimension, so users expect this:
```
gen_ai.client.token.usage(type=input)  # input
gen_ai.client.token.usage(type=output) # output
gen_ai.client.token.usage()            # total 
```
If you add a `total` value, when users query `gen_ai.client.token.usage()`, they will get total twice: total + input + output which is unexpected.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelObservationContext.java
```diff
@@ -0,0 +1,44 @@
+package dev.langchain4j.micrometer.observation;
+
+import dev.langchain4j.model.chat.listener.ChatModelErrorContext;
+import dev.langchain4j.model.chat.listener.ChatModelRequestContext;
+import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
+import io.micrometer.observation.Observation;
+
+public class ChatModelObservationContext extends Observation.Context {
```
jonatan-ivanov, line 8: Not sure if this is the scope of this PR but not using server/receiver context for network communication (i.e.: http requests to the backend) will result context propagation (tracing information traveling on the wire) not working.

I guess this is not a problem for commercial products since what will do with your traceId/spanId but if the backend is user-hosted, tracing information might be useful.

This can be added later but it might be a potential breaking change.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelObservationContext.java
```diff
@@ -0,0 +1,44 @@
...
+
+    ChatModelRequestContext requestContext;
+
+    ChatModelResponseContext responseContext;
+
+    ChatModelErrorContext errorContext;
```
jonatan-ivanov:
```suggestion
    private final ChatModelRequestContext requestContext;

    private ChatModelResponseContext responseContext;

    private ChatModelErrorContext errorContext;
```

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Supplier;
+
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+@Experimental
+public class MicrometerChatModelListener implements ChatModelListener {
+
+    private final ObservationRegistry observationRegistry;
+    private final AtomicReference<Observation.Scope> scope;
```
jonatan-ivanov: I'm not sure what is happening here, usually there is no need to keep a reference to the current scope since the `ObservationRegistry` does this for you, see: https://docs.micrometer.io/micrometer/reference/observation/instrumenting.html

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
```diff
@@ -0,0 +1,129 @@
...
+     * @param observationRegistry   Provided ObservationRegistry by Micrometer Observation API
+     * @param aiSystemName          AI system name should be in line with OpenTelemetry Semantic Convention for Generative AI Metrics.
+     */
+    public MicrometerChatModelListener(
+            final MeterRegistry meterRegistry, ObservationRegistry observationRegistry, String aiSystemName) {
+        this.observationRegistry = ensureNotNull(observationRegistry, "observationRegistry");
+        this.scope = new AtomicReference<>();
+        this.aiSystemName = ensureNotNull(aiSystemName, "aiSystemName");
+
+        observationRegistry.observationConfig().observationHandler(new ChatModelMeterObservationHandler(meterRegistry));
```
jonatan-ivanov: Other than having a custom handler is not a good idea if you don't have a custom output, this will register the handler as many times you create this listener, is this intentional?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
jonatan-ivanov: Since this `ObservationHandler` does not create a different output, I would first check if this could be deleted instead of moving it (see Marcin's comment).

I think the request/error counter is definitely something that the Observation API will do for you out of the box so you don't need it here. Call `observation.error(...)` if there was an error and don't call it if there wasn't, this will result in an `error` tag on your metrics with value `none` if there was not any errors and the error name if there was. If you want an `outcome` tag on top of this, se my comment on the readme, you can do `observation.lowCardinalityKeyValue("outcome", "ERROR")` when there was an error and `observation.lowCardinalityKeyValue("outcome", "SUCCESS")` otherwise, there is no need for extra counters.

For the token usage, I think you might need this handler and if so, I would keep it in the same module, but let's talk about that after the rest is resolved.

Sus4nne:
> Thank you, I like the suggested options, although I am wondering if it really should be tied to observability. I guess this can as well be named `providerName` or something like that

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
Sus4nne: I'll remove the total metric.

Sus4nne:
> Thank you, I like the suggested options, although I am wondering if it really should be tied to observability. I guess this can as well be named `providerName` or something like that

@dliubarskyi Yes, the idea behind this in the OTel Sem Conv is that it says something about the system/provider that is used. It is tied to the kind of API system that is used for the conversation. 

From Otel Sem Conv: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclienttokenusage
> The actual GenAI product may differ from the one identified by the client. Multiple systems, including Azure OpenAI and Gemini, are accessible by OpenAI client libraries. In such cases, the gen_ai.system is set to openai based on the instrumentation’s best knowledge, instead of the actual system.

The providerName or systemName gives information about to what provider/system the request was made with the chat model. It seems fitting to me that this name is set in the specialized chat model. I went back and forth with the name for this thing: provider or system. I think provider is more clear. It should hold what the system **provider** is. `providerName` would settle this.

Sus4nne:
> @Sus4nne FYI I have merged #2403 yesterday, which deprecated `ChatModelRequest` and `ChatModelResponse` in favour of `ChatRequest` and `ChatResponse` in `ChatModelRequestContext`, `ChatModelResponseContext` and `ChatModelErrorContext`.
> 
> It means that there is now access to more attributes in `ChatRequestParameters` (like `frequencyPenalty`, `stopSequences`, etc) as compared to `ChatModelRequest`. That attributes can now be added to [spans](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/)

@dliubarskyi I see. I'll update the code to use the ChatRequest and ChatResponse objects in the Context objects.

langchain4j-micrometer/README.md
Sus4nne: Yes thank you, I noticed it. I deleted the total counter.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
Sus4nne: I appreciate your reviews, thank you for your time. To be honest, I am confused by your comments. I studied the Micrometer documentation and another project implementing this, I thought this was what was meant by using the Observation API.  I hope you can help me understand this.

My plan is to start with the genAI metrics, then add genAI information to the traces/spans of the observation.

@jonatan-ivanov I don't understand what you mean with "this `ObservationHandler` does not create a different output". With the addition of the metrics I believe it does add output. What do you mean with output in this case?
Do you mean with the request/error counter the custom counters `langchain4j.chat.model.error` and `langchain4j.chat.model.request`? These are not OTel Sem Conv metrics, but I thought it was a nice addition. How does the Observation API give us this out of the box?

@brunobat @marcingrzejszczak @jonatan-ivanov What exactly did I do wrong in this set up?
setupMicrometerChatModelListener drawio

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
Sus4nne: @jonatan-ivanov I see, how should we make sure the ChatModelMeterObservationHandler 'works' on the observations of this listener?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelObservationContext.java
dliubarskyi: Hi @jonatan-ivanov, thank you for raising this concern!

There is another [PR](https://github.com/langchain4j/langchain4j/pull/2413) that makes it possible to use Spring's `RestClient` for calling the LLM provider API. In my understanding, as long as we will use the injected `RestClient.Builder`, tracing information should will be passed correctly?

jonatan-ivanov: Yes, that's correct. But will langchain4j always use `RestClient` under the hood?
The Observation API can be used in any situation with or without Spring.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
jonatan-ivanov: Not sure I understand the question.
Are you asking how to make sure the handler only processes observations from the listener (answer: `supportsContext` method of the handler) or how to make sure the handler is registered? If the latter: I'm not familiar how users configure lang4j, but that mechanism should register the handler too (once).

Sus4nne: @jonatan-ivanov I have the supportsContext method in the handler and it seems to work as intended at the moment. My question is about how to register the handler once. How would you do it?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
Sus4nne: @jonatan-ivanov I tried to do it without this and rely on the ObservationRegistry. In the onResponse and onError methods the currentObservation and the currentObservationScope are always null. While debugging I see the observation in the registery, so I can see it is there. Is the way the Observation API works not compatible with the  ChatModelListener way to record a request? Should I tap into it somewhere else to make sure the observation remains available as intended by Micrometer. What would you advise?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
dliubarskyi: @Sus4nne as you have suggested in the discord chat, the registration can be done in the `langchain4j-micrometer-spring-boot-starter` module, so you can simply remove this line

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelObservationContext.java
dliubarskyi: > But will langchain4j always use RestClient under the hood?

When using Spring Boot starters (like `langchain4j-open-ai-spring-boot-starter` and `langchain4j-ollama-spring-boot-starter`), `RestClient` is always used, unless user wants to customize it, but then it is probably out of our hands to make sure eveything is propagated?

> The Observation API can be used in any situation with or without Spring.

In a situation when Spring is not used, we will use JDK's `HttpClient` dy default. Then we will need to manually set HTTP headers with trace ID?

> This can be added later but it might be a potential breaking change.

TBH I would prefer to postpone it to the time when/if it will be required. The only thing I am worried about is why this is potentially a breaking change?

jonatan-ivanov: This can be a breaking change since if you want to do this on the "level" of `ChatModelObservation`, you will need a sender/receiver context (right now it is just context). But based on what you said, this is probably not an issue since you want to do this on the level of the (http) client instead (which I would also recommend) so chat model won't be affected.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
marcingrzejszczak: `onRequest` and `onResponse` and `onError` - are these being called in the same thread?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/observation/ChatModelMeterObservationHandler.java
marcingrzejszczak: OK I checked the code again. If you indeed want to have some custom additional metrics that are also tagged then the way you do it does make sense. Technically you could go with calling `event()` when the observation gets started, stopped and on error however you cannot put low / high cardinality key values on Events so far.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
dliubarskyi: This depends on the type of the model. For sync model - yes. For streaming model - no, `onResponse` and `onError` is called on another thread than `onRequest`.

FWIW, any state can be passed from `onRequest` to `onResponse`/`onError` via `ChatModelRequestContext.attributes` map.

LizeRaes:
Hi everyone, it would be great if we could finish this PR to have observability avilable also for Spring users. @SandraAhlgrimm @jdubois it would be greatly appreciated if you could have a look. @Sus4nne said she's happy to hop on a call about it too if it helps :) Thank you all so much!

jdubois:
1. I want to stress that this is an important feature, from a Microsoft point of view: lack of observability is the top issue customers report on LangChain4j, and the vast majority of them use Spring Boot.

2. We've discussed with @Sus4nne and @SandraAhlgrimm just did a PR that should fix most of the smallest issues, now what should remain is the top comment https://github.com/langchain4j/langchain4j/pull/2384#pullrequestreview-2547155676 from @brunobat about splitting this module in 2: in the end, I don't think you reached a consensus. I believe this is a fairly small module, and it can be later refactored if we need an abstraction layer. That's currently not the case in Spring Ai, and I don't believe we should do better than them on a first release.

3. Now my concern is that this uses a `ChatModelListener`:
  a. This only works for ChatModels, and we should have this for all kind of models (Embeddings, Images, etc). This can totally be added later, and ChatModels are by far the most important models to observe. The only other kind of models that is widely used are Embeddings, and they are very cheap and fast, and frequently used in batches, so they are less important. As a result: I don't think this should be a blocker, let's first release this, and expand its usage later while it's proven to be successful.
    b. Relying on a listener is weird from an API usage point of view. Instead of doing this on line 7:
    
    ```java
        AzureOpenAiChatModel chatModel = AzureOpenAiChatModel.builder()
                .deploymentName(deploymentName)
                .temperature(0.7)
                .listeners(List.of(listener))
                .build();
    ```
    
    I would rather have a fluent method that clearly states that we are adding observability, like:
    
    ```java
        AzureOpenAiChatModel chatModel = AzureOpenAiChatModel.builder()
                .deploymentName(deploymentName)
                .temperature(0.7)
                .observe(MicrometerObservationHandler.defaultSettings())
                .build();
    ```

  -> unlike with the `ChatModelListener`, in this case we would indeed need an interface from the start, otherwise people will be tied to Micrometer for observability.
  -> again, I believe we should do this in a second time: the beauty of the listener approach is that it's non intrusive: it works everywhere without modifying the existing code. Adding a fluent method like `.observe(...)` would have a bigger impact, and this can be done later (it could use the same approach, and add the listener under the hood, so there's no big implementation change). I'd rather have this released fast, and get feedback from users.

dliubarskyi:
@jdubois 

> I want to stress that this is an important feature, from a Microsoft point of view: lack of observability is the top issue customers report on LangChain4j, and the vast majority of them use Spring Boot.

Agreed, I will prioritize it

> We've discussed with @Sus4nne and @SandraAhlgrimm just did a PR that should fix most of the smallest issues

Nice! I will try to review it today or tomorrow. Although the PR has 2500+ files changes, I will wait untill it is sorted out

> This only works for ChatModels, and we should have this for all kind of models (Embeddings, Images, etc). This can totally be added later, and ChatModels are by far the most important models to observe. The only other kind of models that is widely used are Embeddings, and they are very cheap and fast, and frequently used in batches, so they are less important. As a result: I don't think this should be a blocker, let's first release this, and expand its usage later while it's proven to be successful.

Yes, [embedding model listener](https://github.com/langchain4j/langchain4j/pull/4349) is coming soon, and it should be pretty easy to integrate, similarly to `ChatModelListener`.

> Relying on a listener is weird from an API usage point of view. Instead of doing this on line 7:
> 
>     AzureOpenAiChatModel chatModel = AzureOpenAiChatModel.builder()
>             .endpoint(System.getenv("AZURE_OPENAI_ENDPOINT"))
>             .apiKey(System.getenv("AZURE_OPENAI_KEY"))
>             .deploymentName(deploymentName)
>             .temperature(0.7)
>             .logRequestsAndResponses(true)
>             .listeners(List.of(listener))
>             .build();
> I would rather have a fluent method that clearly states that we are adding observability, like:
> 
>     AzureOpenAiChatModel chatModel = AzureOpenAiChatModel.builder()
>             .endpoint(System.getenv("AZURE_OPENAI_ENDPOINT"))
>             .apiKey(System.getenv("AZURE_OPENAI_KEY"))
>             .deploymentName(deploymentName)
>             .temperature(0.7)
>             .logRequestsAndResponses(true)
>             .observe(MicrometerObservationHandler.defaultSettings())
>             .build();

This can be configured programmatically (your example above) or [automatically](https://docs.langchain4j.dev/tutorials/spring-boot-integration#observability), I would expect most of the users to rely on automatic `ChatModel` configuration, in this case just adding `langchain4j-micrometer` dependency would hook in the listener into all models.

___

I will be on PTO next week, so replies can be delayed. If the PR will be ready before 28 Jan, it can be included in the next release

brunobat:
@dliubarskyi will take a look tomorrow.

brunobat:
@jdubois, I promised @Sus4nne  that I would take a look at this, but I must highlight that since the initial discussion there were no changes.
Please notice that the creators of the Observation API, @jonatan-ivanov and @marcingrzejszczak commented above rising the same concerns about the module structure.

This PR is incomplete and needs to be improved in a way that addresses the following concerns:

- The project is named Micrometer, however it is using Micrometer's Observation API, something quite independent on it's own and larger in scope that the Micrometer Metrics. Not the same thing and the naming of the project could be improved.
- LangChain4J has already a tracing API with a single implementation for OpenTelemetry. Do we want to allow the instrumentation implemented with that API to be available in the Observation API... I'd say yes, meaning we are missing a bridge.
- Do we want to allow LangChan4J to instrument with the Observation API directly, @dliubarskyi ? That would make those instrumentation **unavailable** to OpenTelemetry, unless a bridge is implemented in the future (should be out of scope for this PR)
- The current implementation leaks to the Handlers things that should only be in the `Observation`, `Convention` and `Documentation` level. Most likely, we don't need these handlers, in their current form.
- If we allow LangChain4J modules to implement directly with the Observation API, modules can just depend on the Observation API directly and define their own ` Convention`/`Documentation` files, unless we want to  centralize that, in which case we would need a `langchain4j-observation-api` to wrap things.
- Frameworks using the Observation API will have their own `Handlers` in place but it's totally fine to implement a  `langchain4j-observation-<micrometer, brave, opentelemetry, ...>` if we want to use this is standalone mode... Not needed for now.
Thoughts?

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
Sus4nne: I removed the scope here and let it be handled as an attribute of the ChatModelRequestContext.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/ChatModelObservationListener.java
Sus4nne: Yes. I removed it.

langchain4j-micrometer/README.md
Sus4nne: Yes, added a warning in the Readme.

langchain4j-micrometer/README.md
Sus4nne: Changes made:
1. added outcome tags to the observation in the listener.
2. removed separate request/error counters from the handler.
3. updated readme
4. updated tests

langchain4j-micrometer/README.md
Sus4nne: Thank you. Removed it.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
Sus4nne: By storing the scope in the attributes map (which is thread-safe and passed between contexts), the scope can be retrieved on a different thread in the streaming case.

langchain4j-micrometer/src/main/java/dev/langchain4j/micrometer/listeners/MicrometerChatModelListener.java
Sus4nne: Removed this line.

brunobat:
I now realise that the listeners implementation for OTel happens in the integration libraries.

Examples: https://github.com/wildfly-extras/wildfly-ai-feature-pack/blob/5d87e6684d674bac1dd6107eb68270c88205412b/wildfly-ai/injection/src/main/java/org/wildfly/extension/ai/injection/observability/OpenTelemetryTracesChatModelListener.java#L35
https://github.com/quarkiverse/quarkus-langchain4j/blob/main/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/listeners/SpanChatModelListener.java

Probably we should move these implementations to a `langchain4j-otel` library to allow code re-usability. The only tricky part is to provide an API to resolve the correct OTel SDK used by each framework using this new lib.
CC @mariofusco @geoand @ehsavoie 

This would make a `langchain4j-observation-***` natural.

geoand:
TBH, I don't like having all OTel integrations in one module. 

Moreover, if I were working on observability features in anything more than a side-quest, I would have dispelled with that pattern in `quarkus-micrometer` and `quarkus-opentelemetry` ASAP :)

ehsavoie:
@brunobat iirc the wildfly one is relying on https://github.com/langchain4j/langchain4j-cdi/tree/main/langchain4j-cdi-mp/langchain4j-cdi-telemetry
Hum, it looks like it is not, I should change that.

brunobat:
> TBH, I don't like having all OTel integrations in one module.
> 
> Moreover, if I were working on observability features in anything more than a side-quest, I would have dispelled with that pattern in `quarkus-micrometer` and `quarkus-opentelemetry` ASAP :)

Telemetry is probably on the right place now because those extensions only include telemetry from core extensions.

In projects like Vert.x using similar listeners, there are many implementations of those listeners... All potentially producing slightly different telemetry. It's inefficient but Maybe that's fine because it provides the opportunity to enrich each metric or span with framework specific data. 

Transporting this idea to the Observation API and noting that this API doesn't define a concrete implementation either, there is a place for a `langchain4j-observation` library translating the listeners to the Observation API. Concrete implementation for this lib must be in the frameworks, meaning Spring, Wildfly, Quarkus, etc...

jeanbisutti:
>In projects like Vert.x using similar listeners, there are many implementations of those listeners... All potentially producing slightly different telemetry. It's inefficient but Maybe that's fine because it provides the opportunity to enrich each metric or span with framework specific data.

For LangChain4j users or frameworks that use the OpenTelemetry APIs, it would make sense to include LangChain4j listeners with an OpenTelemetry implementation in the LangChain4j repository  (`OpenTelemetryChatModelListener` for example) .

If frameworks want to add specific framework data, then they could create additional listeners.

brunobat:
@Sus4nne, We should move on with this.
I've done an in-depth review. The PR as is can work but I need 2 changes:
- The name of the project should be `langchain4j-micrometer-observation`
- You should have a SpringBoot IT test to make sure this will work with the framework.
Is this OK?

dliubarskyi:
@brunobat there is a new PR: https://github.com/langchain4j/langchain4j/pull/4556

brunobat:
> @brunobat there is a new PR: #4556

We should close this one.

dliubarskyi:
Closing in favour of https://github.com/langchain4j/langchain4j/pull/4556

PR #2394, opened 2025-01-10
MCP: ping responses and simplification of McpTransport

jmartisk:
- Makes the `McpTransport` interface more generic by squashing most operations into `executeOperationWithResponse` and `executeOperationWithoutResponse` instead of having a separate method for each operation
- Implements responding to server-initiated pings

Test coverage for pings I currently only have for the Quarkus side https://github.com/jmartisk/quarkus-langchain4j/commit/5cdcf4298e0168b3e5ba42d5f8780710ac4d9c93 because only there I have a mock MCP server where I am able to hook into pings and confirm to the client that the server received a pong


ghost (approved):
@jmartisk thanks a lot!

PR #2399, opened 2025-01-10
Add the ability to specify a message with the "context" role when using Ollama

### LINKED ISSUE #2391: [FEATURE] Add the ability to specify messages with a "context" role when using Ollama

The recent [Granite Guardian models](https://www.ibm.com/granite/docs/models/guardian/) provide the ability to assess the context for RAG for groundedness (whether the response is consistent with the retrieved context) and relevance (whether the context is relevant for answering the user's question).

When using the Granite Guardian models with Ollama, we need the ability to pass messages with the "context" role in addition to messages with the usual "system", "user", and "assistant" roles. 

Here's an example of how to format messages sent to the granite3-guardian:8b model to assess RAG groundedness when using Ollama:

```
messages = [
    {
        "role": "system",
        "content": "groundedness"
    }, 
    {
        "role": "context",
        "content": context_text
    }, 
    {
        "role": "assistant",
        "content": response_text
    }
]
```

Notice the "context" role that's specified in the above example.

More details about this example can be found in section 3 "Risks related to RAG - Triad" [here](https://github.com/ibm-granite/granite-guardian/blob/main/cookbooks/granite-guardian-3.0/detailed_guide_ollama.ipynb).

Currently, the only supported `ChatMessages` that can be specified when using an `OllamaChatModel` are `AiMessage`, `SystemMessage`, `UserMessage`, and `ToolExecutionResultMessage`. 

It would be good to add the ability to specify a `ContextMessage` as well when using Ollama.

I haven't been able to find any other alternatives for specifying a message with the "context" role to successfully make use of the Granite Guardian models when using Ollama with langchain4j.

**Additional context**
I can look into adding this support if you agree that it makes sense to add.

fjuma:
## Issue

Closes https://github.com/langchain4j/langchain4j/issues/2391

## Change
It's now possible to pass messages with the "context" role in addition to messages with the usual "system", "user", etc. roles when using Ollama.

This change is needed in order to be able to make use of the recent [Granite Guardian models](https://www.ibm.com/granite/docs/models/guardian/) with Ollama. More details about this can be found in the [issue description](https://github.com/langchain4j/langchain4j/issues/2391).

## General checklist


fjuma:
@dliubarskyi @mariofusco Just to let you know, I've now updated this PR with the changes discussed in https://github.com/langchain4j/langchain4j/issues/2391. Thanks!

docker/ollama/granite3-guardian/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM --platform=$BUILDPLATFORM ollama/ollama:latest
```
dliubarskyi: We do not maintain these docker images any more and they are not deployed to the dockerhub any more, so there is no need to add these files.

We need to remove `docker` directory completely (outside of this PR)

langchain4j-core/src/main/java/dev/langchain4j/data/message/CustomRoleMessage.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * Represents a message with a custom role.
+ */
+public class CustomRoleMessage implements ChatMessage {
+
+    private static final String TEXT_ATTRIBUTE = "text";
+    private final String role;
```
dliubarskyi: On a second thought: in the context of making this class as flexible as possible for various future needs, should `role` have a dedicated field in this class or should it be just one of the `attributes`?

langchain4j-core/src/main/java/dev/langchain4j/data/message/CustomRoleMessage.java
```diff
@@ -0,0 +1,100 @@
...
+        return CUSTOM_ROLE;
+    }
+
+    /**
+     * Returns the message text if present.
+     * @return the message text if present.
+     */
+    public String text() {
+        if (attributes != null) {
+            return (String) attributes.get(TEXT_ATTRIBUTE);
```
dliubarskyi: I am not sure we should dedicate a special meaning to the `text` attribute in this class, I guess it should be just one of the `attributes`. We cannot guarantee that every custom message will have/require a `text` attribute in the final message sent to the LLM.

I was also wondering if in Ollama and other modules that will support this message type we should simply pass-through all the `attributes` into the specific message (e.g., `dev.langchain4j.model.ollama.Message`) as-is (without calling any specific getters to get specific data fields), to not make modules dependent on any model-specific logic or fields for Granite or any other model, WDYT?

We could add a `Map<String, Object> additionalFields` into `dev.langchain4j.model.ollama.Message` and un-nest it when serializing:

(Claude)
```java
@JsonIgnoreProperties(ignoreUnknown = true)
@JsonInclude(NON_NULL)
@JsonNaming(SnakeCaseStrategy.class)
class Message {

    private Role role;
    private String content;
    private List<String> images;
    private List<ToolCall> toolCalls;
    
    // Additional fields map
    private Map<String, Object> additionalFields = new HashMap<>();

    // Regular getters and setters for explicit fields...

    @JsonAnyGetter
    public Map<String, Object> getAdditionalFields() {
        return additionalFields;
    }

    @JsonAnySetter
    public void setAdditionalField(String name, Object value) {
        additionalFields.put(name, value);
    }
}
```

Also, `text` method is deprecated and will be removed soon, so I would rather throw an exception if it is called.

langchain4j-ollama/src/test/java/dev/langchain4j/model/ollama/OllamaCustomRoleMessageChatModelIT.java
```diff
@@ -0,0 +1,108 @@
...
+    void should_generate_response_with_custom_role_message() {
+
+        // given
+        String retrievedContext =
+                "One significant part of treaty making is that signing a treaty implies recognition that the other side is a sovereign state and that the agreement being considered is enforceable under international law. Hence, nations can be very careful about terming an agreement to be a treaty. For example, within the United States, agreements between states are compacts and agreements between states and the federal government or between agencies of the government are memoranda of understanding.";
+
+        List<ChatMessage> messages = List.of(
+                SystemMessage.from("context_relevance"),
+                UserMessage.from("What is the history of treaty making?"),
+                CustomRoleMessage.from("context", Map.of("text", retrievedContext)));
```
dliubarskyi: As a user, I would probably expect to use "content" instead of "text", as outlined in [your example](https://github.com/langchain4j/langchain4j/issues/2391):
`CustomRoleMessage.from("context", Map.of("content", retrievedContext)`

langchain4j-ollama/src/test/java/dev/langchain4j/model/ollama/OllamaCustomRoleMessageChatModelIT.java
```diff
@@ -0,0 +1,108 @@
...
+
+        // when
+        ChatResponse chatResponse =
+                ollamaChatModel.chat(ChatRequest.builder().messages(messages).build());
```
dliubarskyi: // offtopic

@mariofusco I am wondering if we should add a `chat(List<ChatMessage>)` and/or `chat(ChatMessage...)` convenience methods to the `ChatLanguageModel`.
Currently there are `generate(List<ChatMessage>)` and `generate(ChatMessage...)` methods in the old API, so perhaps we should have the same in the new API. There as some user already using those convenient methods and it will be not nice to force them (and new users) to do
`chat(ChatRequest.builder().messages(messages).build())` every time they want to pass some messages without adding specific parameters.

WDYT?

dliubarskyi (commented):
@fjuma thanks a lot!

langchain4j-core/src/main/java/dev/langchain4j/data/message/CustomRoleMessage.java
dliubarskyi: (We cannot guarantee that every custom message will have/require a `role` attribute in the final message sent to the LLM)

docker/ollama/granite3-guardian/Dockerfile
fjuma: Makes sense, I've removed these files.

langchain4j-core/src/main/java/dev/langchain4j/data/message/CustomRoleMessage.java
fjuma: That's a good point, to make this as flexible as possible for the future, it makes sense for the "role" to just be part of the `attributes`. I've updated this and also renamed `CustomRoleMessage` to just `CustomMessage`.

langchain4j-core/src/main/java/dev/langchain4j/data/message/CustomRoleMessage.java
fjuma: Yeah I like the idea of just passing the `attributes` as is through into the message. This lets us get rid of the "context" specific handling in the Ollama module and simplifies the logic. I've made this change and also updated the `text` method to just throw an exception.

fjuma:
@dliubarskyi Thanks very much for the feedback, I've now updated this. It does seem that this approach is much more flexible now since we are just passing the `CustomMessage` attributes through into the message with no model specific logic needed in the modules that support `CustomMessage`s.

dliubarskyi (approved):
@fjuma thanks a lot, great job!

Issue #2331, opened 2024-12-23
Implement observability with OpenTelemetry


langchain4j:
Implement observability with OpenTelemetry following [Gen AI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/).

This should help with integration, among other things, with Langfuse and LangSmith, both of which support OTel.

Implement it as a separate module (e.g., `langchain4j-opentelemetry`), as an implementatrion of a `ChatModelListener` interface (e.g., `OpenTelemetryChatModelListener`).

langchain4j:
Related: https://github.com/wildfly-extras/wildfly-ai-feature-pack/blob/5d87e6684d674bac1dd6107eb68270c88205412b/wildfly-ai/injection/src/main/java/org/wildfly/extension/ai/injection/observability/OpenTelemetryTracesChatModelListener.java#L35

langchain4j:
Related: https://github.com/smallrye/smallrye-llm/blob/main/smallrye-llm-langchain4j-telemetry/src/main/java/io/smallrye/llm/langchain4j/telemetry/SpanChatModelListener.java

salaboy:
Folks any progress on this? @mariofusco, @dliubarskyi  , do you know where to check for this?

dliubarskyi:
@salaboy please check https://docs.langchain4j.dev/tutorials/observability/#opentelemetry-genai-instrumentation

Issue #2328, opened 2024-12-23
Langfuse: implement tracing


langchain4j:
Integrate [Langfuse tracing](https://langfuse.com/docs/tracing) into LangChain4j.

Investigate how to better map to the [Langfuse tracing data model](https://langfuse.com/docs/tracing-data-model) to make it easy to analyze/visualize a single AI Service call that can span multiple LLM calls (e.g., when tools/functions are executed/called), tools, RAG, etc.

Implement as a separate module (e.g., `langchain4j-langfuse`).

langchain4j:
closing in favour of https://github.com/langchain4j/langchain4j/issues/2331

yunmaoQu:
I want give it a try

yunmaoQu:
I want give it a try

langchain4j:
@yunmaoQu awesome, thanks a lot in advance!

yunmaoQu:
Hi @dliubarskyi,

I've started implementing the Langfuse tracing integration and would appreciate your early feedback before proceeding further. Here's what I've done so far:

1. Initial Implementation:
   - Created langchain4j-langfuse module
   - Integrated with Langfuse's tracing data model
   - Added basic tracing functionality for AI Service calls

2. Key Design Decisions:
   - Implemented tracing through AiServices builder configuration
   - Capturing LLM calls, tool executions, and RAG operations
   - Handling both sync and streaming responses

3. Next Steps (pending your feedback):
   - Adding comprehensive test coverage
   - Implementing error handling and edge cases
   - Adding documentation and examples

Would appreciate your thoughts on:
- Is this the right architectural approach?
- Any specific scenarios I should focus on for testing?
- Any potential issues or improvements you see?

Thank you for your time!

RobWin:
Hello,

my understanding is that Langfuse is now supporting OpenTelementry: https://langfuse.com/docs/opentelemetry/get-started

Couldn't we use Micrometer Tracing instead of creating custom Tracing classes? https://docs.micrometer.io/tracing/reference/

And then an integration (reporter) into Langfuse like https://docs.micrometer.io/tracing/reference/reporters.html

wdyt?

geoand:
It would be great if @marcklingen could provide input on whether the above suggestion about using the OpenTelemetry support of LangFuse makes sense.

It seems to me that https://github.com/langchain4j/langchain4j/pull/2499 goes further than what a simple OTel integration would.

RobWin:
[...]

Issue #2145, opened 2024-11-19
Spring Boot: a way to configure ChatModelListener(s) in application properties


ian-k:
**Is your feature request related to a problem? Please describe.**
In Spring Boot, it appears that there is no built-in way to create an OpenAiChatModel with ChatModelListener(s). The current design seems to make it impossible, requiring workarounds. The AutoConfig class always initializes ChatModel instances without attaching any listeners.

**Describe the solution you'd like**

AI properties should support keys like:
langchain4j.open-ai.chat-model.listeners=com.org.xxx.MyListener1,...,
enabling the AutoConfig class to utilize these properties for creating the model with the specified listeners.

**Describe alternatives you've considered**

Manual creation of ChatModel 

**Additional context**

langchain4j:
Hi @ian-k, what do you plan to do in your custom listeners?

myrosia:
I would find it useful as well, we do custom listeners for observability, which is what langchain4j recommends (https://docs.langchain4j.dev/tutorials/observability/) but this means that we have to abandon configuration via properties, which is really unfortunate.

But actually I would prefer to just be able to provide some beans via annotation in the application, similar to how tools are annotated, that would be my first suggestion. Or, alternatively, have a bean providing a list of listeners to add to the model. If it exists, then add the listeners, if not, assume they are an empty list, in exactly the same way as wiring happens for chat models right now.

dliubarskyi:
@ian-k thanks for the suggestion, it makes total sense to provide an option to hook up listeners.

But I am not sure about the `langchain4j.open-ai.chat-model.listeners=com.org.xxx.MyListener1,...,` way of configuring it, as it will mean that we will need to instantiate those classes with reflection and there is no way to hook up dependencies of those classes.
Unless it is something like `langchain4j.open-ai.chat-model.listeners=[customeModelListener1, customeModelListener2]` where those are the names of beans?

Hi @myrosia!

> But actually I would prefer to just be able to provide some beans via annotation in the application, similar to how tools are annotated, that would be my first suggestion.

Do you mean something like this?
```java
@Bean
ChatModelListener customeModelListener1() {
    return new CustomeModelListener1();
}

@Bean
ChatModelListener customeModelListener2() {
    return new CustomeModelListener2();
}
```
In this case both listeners will be hooked up to every `ChatLanguageModel` bean in the application. The ordering will not be guaranteed.

[...]

Issue #1963, opened 2024-10-20
[FEATURE] Enhance Listener functionality with ordering and execution control


ddobrin:
**Is your feature request related to a problem? Please describe.**
Listeners provide a flexible solution to  intercept, modify, and enhance requests, responses and errors. 

Abstracting reusable patterns into individual listeners can provide flexibility and portability across multiple models.
Ex.: request/response observability, PII data masking, implementing security guardrails for harmful content, enhancing request/response with data from a custom datasource.

The following two items would provide significant benefits to production applications

1. Ordering: retaining the ability to assign an order of execution key to each listener would provide the ability to set the desired order of execution of listeners. 
Ex.: a request followed by an enhancement listener to the request would want to add and run safety guardrails as the last listener before making the LLM call, and conversely validate the response as the first operation in a list of listeners upon receiving a response

2. Current ChatModelListener interface allows the override of onRequest(), onResponse() and onError(), intercepting the model calls before, after and on error.
It would be beneficial to allow an aroundModelCall() model, which would allow the developer to control the execution of the LLM call. 
Ex.: In case the safety guardrail fails, the code in aroundModelCall() (or a more suitable name) would allow the developer to make the decision of not proceeding with the LLM call.

**Describe the solution you'd like**
1. add a getOrder() method to the ChatModelListener, with an integral value allowing an evaluation of [0..N] in increasing order
2. add an aroundModelCall(), suggested in a ChatAroundModelListener interface, and allowing the ChatModelListener to remain unchanged

langchain4j:
@ddobrin thank you!

Please note that listeners were not designed to modify requests, they were primarily added to enable observability and/or custom logging.

Guardrails and the ability to modify requests will most probably be added as a separate feature.

ddobrin:
Agreed, @langchain4j 

A clean separation of listeners for observability/logging (Read/only) from cross-cutting concerns for PII/Safety/etc (Read/Modify) would work well, and ordering would be beneficial for the latter.

PR #2403, opened 2025-01-12
Centralise observability/listener logic for chat models

### LINKED ISSUE #2397: Centralise listener logic for chat models

dliubarskyi:
## Issue
Closes #2397

## Motivation
Currently there are a few problems with existing observability support for chat models:
- not all models support it (see "Observability" column [here](https://docs.langchain4j.dev/integrations/language-models/))
- the code to implement it is duplicated in each `ChatLanguageModel` implementation
- there are redundant `ChatModelRequest` and `ChatModelResponse` types
- there is no way to propagate LLM-provider-specific chat request parameters and chat response metadata

## Change
- centralised all the logic related to observability/listeners in `ChatLanguageModel` and `StreamingChatLanguageModel`
- deprecated `ChatModelRequest` and `ChatModelResponse` in favour of `ChatRequest` and `ChatResponse` in `ChatModelRequestContext`, `ChatModelResponseContext` and `ChatModelErrorContext`
- migrated OpenAI integration (as a pilot) to inherit observability logic from `ChatLanguageModel` and `StreamingChatLanguageModel`

## Breaking Changes

1. `OpenAiStreamingChatModel`: in case of errors that happen in the middle of streaming, partial response (what was streamed before the error occured) is not propagated into the `ChatModelListener.onError()` any more in order to simplify the design and implementation. Please reach out if you have any concerns about it.

2. `OpenAiChatModel` and `OpenAiStreamingChatModel`: if the `maxCompletionTokens` property is set, it's value is now not propagated into the `ChatModelListener` via `ChatModelRequest.maxTokens()`.
`ChatModelRequest.maxTokens()` will now return only the value of the `maxTokens` parameter set on the `OpenAiChatModel` or `OpenAiStreamingChatModel`.
Now you can get the value of the `maxCompletionTokens` parameter like this:
```java
        ChatModelListener listener = new ChatModelListener() {
            @Override
            public void onRequest(ChatModelRequestContext requestContext) {
                ChatRequestParameters parameters = requestContext.chatRequest().parameters();
                if (parameters instanceof OpenAiChatRequestParameters openAiParameters) {
                    Integer maxCompletionTokens = openAiParameters.maxCompletionTokens();
                }
            }
        };
```

## Next Steps (next PRs)
- Migrate all existing `(Streaming)ChatLanguageModel` implementations from their own (duplicated) observability functionality to inherit it from `(Streaming)ChatLanguageModel`

## General checklist
- [X] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)


langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
```diff
@@ -0,0 +1,108 @@
...
+ * listener callback.
+ * <p>
+ * All listener callbacks are executed in a fail-safe manner - exceptions in listeners are logged but won't
+ * disrupt the main request flow.
+ *
+ * @see ChatModelListener
+ * @see ChatRequest
+ * @see ChatResponse
+ */
+public interface ListenableChatModel extends ChatLanguageModel {
```
mariofusco: Could `ObservableChatModel` be a better name?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
```diff
@@ -0,0 +1,108 @@
...
+
+    Logger log = LoggerFactory.getLogger(ListenableChatModel.class);
+
+    @Override
+    default ChatResponse chat(ChatRequest chatRequest) {
```
mariofusco: Shouldn't this method invoke the `super` implementation at some point? It seems to me that the default implementation in the parent interface has an additional and different behaviour related with tooling, so overriding that method in this way probably introduce a potential bug, other than violating the Liskov substitution principle. Moreover, since you also mentioned that this acts as a decorator for the `ChatLanguageModel` in the JavaDoc, I'm wondering if we couldn't actually implement this using the decorator pattern instead of extending the other interface. This would allow greater flexibility (not sure if this could be necessary though) and allow to compose other decorators as well.

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
dliubarskyi: Agreed. Actually it was named `ObservableChatModel` before, then I renamed it to allign better with the "listener" in the `ChatModelListener`.
But I guess it is better to rename `ChatModelListener` into `ChatModelObserver` instead.

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
dliubarskyi: > Shouldn't this method invoke the super implementation at some point?

It invokes `doChat(ChatRequest)` instead of `super.chat(ChatRequest)`. The reason for this is that this is not a "real" decodator (more on that below).

> It seems to me that the default implementation in the parent interface has an additional and different behaviour related with tooling, so overriding that method in this way probably introduce a potential bug, other than violating the Liskov substitution principle.

The default implementation of `ChatLanguageModel.chat(ChatRequest)` is there only temporarily, untill all models will implement the new `chat(ChatRequest)` API. The plan is that models will first implement the new API, then extend `ListenableChatModel`, so there should be no issue.

If any model will extend `ListenableChatModel`, it will be forced to implement `doChat(ChatRequest)`, so basically implementing the new API, so the default temporary logic from `ChatLanguageModel.chat(ChatRequest)` is not applicable any more in this case.

> Moreover, since you also mentioned that this acts as a decorator for the ChatLanguageModel in the JavaDoc, I'm wondering if we couldn't actually implement this using the decorator pattern instead of extending the other interface. This would allow greater flexibility (not sure if this could be necessary though) and allow to compose other decorators as well.

Perhaps the "decorator" was not the best term to describe it. It is not a real decorator in the sense that one can dynamically wrap `ChatLanguageModel` in the runtime and get extra features.
I am not sure having a real decorator will bring a lot of benefits, as users will then have to manually wrap any model into the `ObservableChatModel` to enable observability.

mariofusco: Thanks for all your explanations, now it makes much more sense.

> I am not sure having a real decorator will bring a lot of benefits, as users will then have to manually wrap any model into the `ObservableChatModel` to enable observability.

That's not exactly what I had in mind. The use of the decorator pattern would be an internal implementation detail, not something that should leak to end-users, and (if we want to keep models immutability at runtime) also not something that could be added dynamically once the model has been already created and in use. That's just an idea, I will play around it a bit and see how it comes along.

mariofusco: I gave a quick attempt to this, but you were totally right and that comment in the JavaDocs (we need to reword it) brought me off-track: this is not a decorator in the sense that you're not decorating the `doChat` method but only having callbacks before and and after its invocation (and in case of an error) so the current implementation is the one that best fits the requirement.

However this failed attempt gave me a chance to give a deeper look into the code and now I have a further remark: do we really need a distinction between a "plain" ChatModel and a listenable/observable one? Or in other words are we planning to have `ChatModel`s that structurally won't be observable? I don't think so and however even in that case it would be enough to always return an empty collection from the `listeners()` method. In essence I'm suggesting to remove those `ListenableChatModel` interfaces, squashing its implementation into the `ChatModel`s assuming that all of them are indeed observable.

langchain4j-core/src/main/java/dev/langchain4j/model/chat/listener/ChatModelResponseContext.java
```diff
@@ -1,42 +1,74 @@
...
     private final Map<Object, Object> attributes;
 
+    public ChatModelResponseContext(ChatResponse chatResponse,
+                                    ChatRequest chatRequest,
+                                    Map<Object, Object> attributes) {
+        this.chatResponse = ensureNotNull(chatResponse, "chatResponse");
+        this.response = ChatModelResponse.fromChatResponse(chatResponse);
+        this.chatRequest = ensureNotNull(chatRequest, "chatRequest");
+        this.request = ChatModelRequest.fromChatRequest(chatRequest);
+        this.attributes = ensureNotNull(attributes, "attributes");
```
kpavlov, line 34: Here and in other contexts where Map of attributes are used: Are we worring about immutable/thread safe attribute map here? Wouldn't it be more safe to wrap it into Collections.unmodifiableMap()?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
```diff
@@ -0,0 +1,108 @@
...
+            }
+        });
+    }
+
+    static void onResponse(ChatResponse chatResponse,
+                           ChatRequest chatRequest,
+                           Map<Object, Object> attributes,
+                           List<ChatModelListener> listeners) {
+        ChatModelResponseContext responseContext = new ChatModelResponseContext(chatResponse, chatRequest, attributes);
+        listeners.forEach(listener -> {
```
kpavlov: For response processing, shouldn't the litseners be called in reverse order?

`Listener1.onRequest -> Listener2.onRequest -> Chat ->  Listener2.onResponse/onError -> Listener1.onResponse/onError`

mariofusco: I don't see why and more in general I believe that the order of listeners shouldn't have any semantic, so they could actually be notified in any order. Am I missing something? Is there any use case where the order of listeners matter?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
dliubarskyi: > I gave a quick attempt to this, but you were totally right and that comment in the JavaDocs (we need to reword it) brought me off-track: this is not a decorator in the sense that you're not decorating the doChat method but only having callbacks before and and after its invocation (and in case of an error) so the current implementation is the one that best fits the requirement.

Makes sense, I will reword it. Confession: it was generated by Claude and it seemed reasonable to me, so I accepted it 🤦 

> However this failed attempt gave me a chance to give a deeper look into the code and now I have a further remark: do we really need a distinction between a "plain" ChatModel and a listenable/observable one? Or in other words are we planning to have ChatModels that structurally won't be observable? I don't think so and however even in that case it would be enough to always return an empty collection from the listeners() method. In essence I'm suggesting to remove those ListenableChatModel interfaces, squashing its implementation into the ChatModels assuming that all of them are indeed observable.

Yes, the plan is that all chat models should be listenable/observable in the end.

I did not want to mess up with the default (and temporary) backwards-compatible implementation of `chat()` in the `ChatLanguageModel` class, so I created this new interface (to start from the blank page, so to say).

You are right that if all models will be observable, having this extra interface seems unnecessary.
On the other hand, having the observability logic isolated in one place is also not the worst thing to have. Do you see any major downsides to having this extra interface?

But it probably won't be superfluous to see how squashed interfaces look like, so please feel free to try it out

langchain4j-core/src/main/java/dev/langchain4j/model/chat/listener/ChatModelResponseContext.java
dliubarskyi: The purpose of this map is to pass data from one `ChatModelListener` method to another (e.g., propagate observability spans/contexts/etc), or even between multiple `ChatModelListener`s, so having it immutable defeats the purpose

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
kpavlov: It reminds me Servlet Filter Chain and Netty Handlers. This is also the way how spring-ai Advisors are [called](https://docs.spring.io/spring-ai/reference/api/advisors.html#_advisor_order).
There might be a use-case, when order of execution does matter. E.g. setting and cleaning MDC context.
Regarding Listeners semantic - I agree. But do we want just listeners when we can build a filter chain?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/listener/ChatModelResponseContext.java
kpavlov: Ah, I missed the ConcurrentHashMap [here](https://github.com/langchain4j/langchain4j/pull/2403/files#diff-22b8a66ca5ba3d0504b390f7616d2dbc74834ad70955cec6723c185cb1a3994cR52), so it is effectively thread-safe

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
kpavlov: There is a relevant feature request with explicit request for the ordering #1963

dliubarskyi: Regarding ordering: ok, let's consider calling it in reverse. Can you provide a few extra examples aside from MDC where reverse order is required? Regarding MDC, what exactly do you have in mind?
For special cases when users need more control over the execution way/order, they could define a single listener and then do whatever they want there (delegate a call to multiple listeners in any order they wish).

Regarding listener vs filter: sure, we can build anything, but should we? (considering we have a lot of other features to address) This feature is meant for observability, not for modifying requests/responses. There is a higher-level feature ([guardrails](https://docs.quarkiverse.io/quarkus-langchain4j/dev/guardrails.html)) for more complicated logic that perhaps will be backported to LC4j from Quarkus-LC4j extension

dliubarskyi: BTW I do not expect listeners to be used extensively by the end users (to hook some business logic).
This feature is targeted mainly for frameworks (e.g., Quarkus, Spring Boot, MP, etc) to hook up observability functionality.

dliubarskyi: The plan is to have a few OOTB `ChatModelListener` implementations (e.g., for [Micrometer](https://github.com/langchain4j/langchain4j/pull/2384), [OpenTelemetry](https://github.com/langchain4j/langchain4j/issues/2331), [Langfuse](https://github.com/langchain4j/langchain4j/issues/2328), etc.) as separate modules (e.g., `langchain4j-micrometer`, `langchain4j-opentelemetry`, etc).

For example, for Spring Boot integration, we could provide a `langchain4j-micrometer-spring-boot-starter` module that, when imported, will provide a `MicrometerChatModelListener` bean. It will be then automatically injected (along with all other available `ChatModelListener` beans, either from other lc4j observability starters (e.g., langfuse) or user-defined) in every `ChatLanguageModel` bean. In case more control is needed, user could specify which listeners should be configured on the model like this: `langchain4j.open-ai.chat-model.listeners=[micrometerChatModelListener, myCustomChatModelListener]`

Related: https://github.com/langchain4j/langchain4j/issues/2145

WDYT?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/listener/ChatModelResponseContext.java
dliubarskyi: Hmm, I am not sure we actually need a `ConcurrentHashMap` as listeners are executed sequentially and in the same thread

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
dliubarskyi: I have added a detailed documentation on how listeners are called [here](https://github.com/langchain4j/langchain4j/pull/2403/files#diff-07739c8f7b3d8fb79599a7e8a2cdf143b55c9385ba5dad9c41db94c65dc98b4aR126)

kpavlov: Yes, the observability is a good usecase I had in mind. Another usecase is PII reduction before sending data to LLM. The "obfiscation listener" might need to be at the specific position in the listener chain

kpavlov: For spring usecase, the listener might be marked with `@Ordered` annotation, but in general, it might have a Priority field (int) for predictable ordering

dliubarskyi: > Another usecase is PII reduction before sending data to LLM

listeners cannot modify the request, only listen/observe

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ChatLanguageModel.java
```diff
@@ -38,6 +38,17 @@ public interface ChatLanguageModel {
...
+     * This is the main API to interact with the chat model.
+     * All the existing generate(...) methods (see below) will be deprecated and removed before 1.0.0 release.
+     * <p>
+     * A temporary default implementation of this method is necessary
+     * until all {@link ChatLanguageModel} implementations adopt it. It should be removed once that occurs.
+     *
+     * @param chatRequest a {@link ChatRequest}, containing all the inputs to the LLM
+     * @return a {@link ChatResponse}, containing all the outputs from the LLM
+     */
+    @Experimental
```
kpavlov, line 49: Should it be still experimental?

dliubarskyi: Yes, I would keep it for at least one more release

langchain4j-core/src/main/java/dev/langchain4j/model/chat/ListenableChatModel.java
TheEliteGentleman: We've already implemented the same feature on Microprofile using Microprofile Telemetry API.

dliubarskyi: @TheEliteGentleman which exactly feature, observability for `ChatLanguageModel`? Could you please provide a link?
Is it a `ChatModelListener` implementation that uses Microprofile Telemetry API?

TheEliteGentleman: Yes, for SmallRye-LLM project (https://github.com/smallrye/smallrye-llm/tree/main/smallrye-llm-langchain4j-telemetry/src/main/java/io/smallrye/llm/langchain4j/telemetry).

dliubarskyi: @TheEliteGentleman great! With this PR merged and released (next week) you can change your listeners to use `ChatRequest` and `ChatResponse` and get much more (common) data for observability, and also model-specific `ChatRequestParameters` and `ChatResponseMetadata`, see example [here](https://github.com/langchain4j/langchain4j/pull/2403/files#diff-07739c8f7b3d8fb79599a7e8a2cdf143b55c9385ba5dad9c41db94c65dc98b4aR40)

docs/docs/tutorials/observability.md
```diff
@@ -66,4 +126,32 @@ ChatLanguageModel model = OpenAiChatModel.builder()
...
+  `ChatModelListener.onRequest()` will **_not_** be called for every retry.
+- The `ChatModelListener.onResponse()` method is called only once,
+  immediately after receiving a successful response from the LLM provider.
+- The `ChatModelListener.onError()` method is called only once.
+  If an error occurs while calling the LLM provider API and a retry happens,
+  `ChatModelListener.onError()` will **_not_** be called for every retry.
+- If an exception is thrown from one of the `ChatModelListener` methods,
+  it will be logged at the `WARN` level. The execution of subsequent listeners will continue as usual.
+- The `ChatRequest` provided via `ChatModelRequestContext`, `ChatModelResponseContext`, and `ChatModelErrorContext`
+  is the final request, containing both the default `ChatRequestParameters`
```
dliubarskyi:
```suggestion
  is the final request, containing both the default `ChatRequestParameters` configured on the `ChatLanguageModel`
```

PR #2132, opened 2024-11-17
[feat] Add Bedrock Converse API support

herbert-beckman:
## Issue

Closes #1789

## Change

This PR add support to Bedrock Converse API.

## General checklist

- [x] There are no breaking changes
- [x] I have added unit and integration tests for my change
- [ ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new maven module

- [ ] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #2404, opened 2025-01-12
[feat] Add Bedrock Converse API support

### LINKED ISSUE #1789: [FEATURE] Use Converse API instead of InvokeModel API for Amazon Bedrock integrations
Currently, Langchain4j’s Amazon Bedrock integration utilizes the InvokeModel API. Amazon has introduced the Converse API, which provides a consistent interface that works with all models that support messages. https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_Converse.html

### LINKED ISSUE #2233: AWS Bedrock: Support Nova

tenpigs267:
## Issue
Closes #1789
Closes #2233

## Change
This PR add support to Bedrock Converse API which enable to use any of text genAI models in AWS Bedrock catalog (including Nova models)
A new BedrockChatModel is added in the converse package.
- Tool use is implemented
- Image content and PDF content is implemented

## General checklist

This PR follow the PR #2132. I proposed my help to @herbert-beckman on #2132 , he gives a thumbs up so I started working on it. I opened a PR on his (https://github.com/herbert-beckman/langchain4j/pull/1) but I never had any reply since, so, as I can't push on the #2132 here I am opening a new PR.


tenpigs267:
- stopSequence have a strange behavior with Amazon Nova models, the stopSequence is included in the result. It seems all other LLM exclude the stopsequence from the response. I had to override should_respect_stopSequences_in_chat_request AbstractBaseChatModelIT test
- should_respect_system_message from AbstractBaseChatModelIT can't be disabled and titan-text-express doesn't support system prompt so I had to override this test also
- nova models failed to describe the second image in should_accept_multiple_images_as_base64_encoded_strings test. For that one I modified the user prompt from "What do you see?" to "What do you see on these images?" the test has been green since and it doesn't modify the spirit of the test

My personal AWS account is limited to 1 request per minute for claude 3.5 models so I can't check the "I have manually run all the unit and integration tests in the module I have added/changed, and they are all green" because I had some "too many requests errors. But as far as I can tell tests are green. Also I tried to add LLama 3.2 to the list of models for IT tests but my request to have access was denied (maybe because I'm in Europe or for rate limiting... same thing with LLama 3.3 70B, I did'nt try to ask access to older LLama)

I had to extends three times AbstractChatModelIT because of the diffrentes capabilities of the bedrock model catalog. Maybe the List<ChatLanguageModel> could be a List<ChatLanguageModelCapabilities> with ChatLanguageModelCapabilities as a class with a model instance and booleans showing its capabilities (tool use, image content, ...)

langchain4j-parent/pom.xml
```diff
@@ -84,7 +84,7 @@
         <elastic.version>8.15.3</elastic.version>
         <jackson.version>2.18.2</jackson.version>
         <jedis.version>5.0.0</jedis.version>
-        <aws.java.sdk.version>2.21.44</aws.java.sdk.version>
+        <aws.java.sdk.version>2.29.12</aws.java.sdk.version>
```
dliubarskyi, line 87: Just to double check: converse API is available via the same SDK, right? So there is no need for a separate module like `langchain4j-bedrock-converse`?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
+package dev.langchain4j.model.bedrock.converse;
```
dliubarskyi: Should it be here or in `dev.langchain4j.model.bedrock`?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsDocumentConverter.java
```diff
@@ -0,0 +1,169 @@
...
+import dev.langchain4j.model.chat.request.json.JsonStringSchema;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import software.amazon.awssdk.core.document.Document;
+import software.amazon.awssdk.core.document.internal.MapDocument;
+
+public class AwsDocumentConverter {
```
dliubarskyi: nit: could be package-private

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsDocumentConverter.java
```diff
@@ -0,0 +1,169 @@
...
+                list.add(getDocument(node));
+            }
+            doc = Document.fromList(list);
+        } else {
+            doc = Document.fromString(value.asText());
+        }
+        return doc;
+    }
+
+    public static Document convertJsonObjectSchemaToDocument(ToolSpecification toolSpecification) {
```
dliubarskyi: Worth checking if `JsonSchemaElementHelper.toMap()` can be (re)used somehow to help with some of the conversions in this class

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+import software.amazon.awssdk.services.bedrockruntime.model.ToolResultBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.ToolResultContentBlock;
+import software.amazon.awssdk.services.bedrockruntime.model.ToolUseBlock;
+
+public class BedrockChatModel implements ChatLanguageModel {
+
+    private final Logger logger = LoggerFactory.getLogger(BedrockChatModel.class);
+
+    // based on input modalities from https://docs.aws.amazon.com/bedrock/latest/userguide/models-supported.html
+    private static final Pattern IMAGE_SUPPORTED_PATTERN = Pattern.compile("(amazon\\.nova-lite|amazon\\.nova-pro|"
```
dliubarskyi: I am not sure it is a good idea... new models are added quite often, other models are deprecated. We will not be able to maintain this list up-to-date. This will mean that users will not be able to use newer models untill we update htis list and make a new release

dliubarskyi: If it is done to satisfy tests, perhaps we need to think about how to adapt tests

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+    private final BedrockRuntimeClient client;
+    private final ChatRequestParameters defaultRequestParameters;
+
+    public BedrockChatModel(String modelId) {
+        this(
+                Region.US_EAST_1,
+                DefaultCredentialsProvider.create(),
+                modelId,
+                null,
+                0.7d,
```
dliubarskyi: are all those parameters mandatory? I would not introduce any default values if they are not mandatory

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+                null,
+                null,
+                5,
+                Duration.ofMinutes(1),
+                null,
+                ChatRequestParameters.builder().build());
+    }
+
+    public BedrockChatModel(
```
dliubarskyi: I would accept a `Builder` in the ctor instead. once new params will be added in the future, it will be easier to propagate

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+            Double temperature,
+            Double topP,
+            List<String> stopSequences,
+            Integer maxRetries,
+            Duration timeout,
+            BedrockRuntimeClient client,
+            ChatRequestParameters defaultRequestParameters) {
+        this.region = getOrDefault(region, Region.US_EAST_1);
+        this.credentialsProvider = getOrDefault(credentialsProvider, DefaultCredentialsProvider.create());
+        this.modelId = getOrDefault(modelId, "us.amazon.nova-micro-v1:0");
```
dliubarskyi: please do not include default for the model, user should explicitly specify it

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+
+    @Override
+    public Response<AiMessage> generate(List<ChatMessage> messages, ToolSpecification toolSpecification) {
+        return generate(messages, List.of(toolSpecification));
+    }
+
+    @Override
+    public Response<AiMessage> generate(List<ChatMessage> messages, List<ToolSpecification> toolSpecifications) {
+        ConverseRequest request = buildConverseRequest(messages, toolSpecifications, null);
+        logger.debug("REQUEST : {}", request);
```
dliubarskyi: Please introduce `logRequests` and `logResponses` parameters that activate logging (see other model impls for example)

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+                .stopSequences(getOrDefault(
+                        stopSequences,
+                        nonNull(defaultRequestParameters) ? defaultRequestParameters.stopSequences() : null))
+                .build();
+
+        validate(this.defaultRequestParameters, modelId);
+    }
+
+    @Override
+    public Response<AiMessage> generate(final List<ChatMessage> messages) {
```
dliubarskyi: nit: are all these default methods need to be re-implemented?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+                tokenUsageFrom(response.usage()),
+                finishReasonFrom(response.stopReason()),
+                Map.of("id", response.responseMetadata().requestId()));
+    }
+
+    @Override
+    public ChatResponse chat(ChatRequest request) {
+        ConverseRequest convRequest = buildConverseRequest(
+                request.messages(), request.parameters().toolSpecifications(), request.parameters());
+        logger.debug("REQUEST : {}", convRequest);
```
dliubarskyi: same here

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+    private AiMessage aiMessageFrom(ConverseResponse converseResponse) {
+        ArrayList<ToolExecutionRequest> toolExecRequests = new ArrayList<>();
+        String textAnswer = "";
+        for (ContentBlock cBlock : converseResponse.output().message().content()) {
+            if (cBlock.type() == ContentBlock.Type.TOOL_USE) {
+                toolExecRequests.add(ToolExecutionRequest.builder()
+                        .name(cBlock.toolUse().name())
+                        .id(cBlock.toolUse().toolUseId())
+                        .arguments(documentToJson(cBlock.toolUse().input())
+                                .replace("\r", "")
```
dliubarskyi: why is this required?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+        }
+    }
+
+    public static Float dblToFloat(Double d) {
+        if (Objects.isNull(d)) {
+            return null;
+        } else return d.floatValue();
+    }
+
+    public static class BedrockChatModelBuilder {
```
dliubarskyi:
```suggestion
    public static class Builder {
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -0,0 +1,717 @@
...
+            return this;
+        }
+
+        public BedrockChatModelBuilder topP(Double topP) {
+            this.topP = topP;
+            return this;
+        }
+
+        public BedrockChatModelBuilder stopSequences(List<String> stopSequences) {
+            if (nonNull(stopSequences)) this.stopSequences = new ArrayList<>(stopSequences);
```
dliubarskyi: We usually do this in the constructor with a `copyIfNotNull()` helper

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
```diff
@@ -0,0 +1,94 @@
...
+        return List.of(AWS_NOVA_LITE, AWS_NOVA_PRO);
+    }
+
+    @Override
+    protected String customModelName() {
+        return "cohere.command-r-v1:0";
+    }
+
+    @Override
+    protected ChatRequestParameters createIntegrationSpecificParameters(int maxOutputTokens) {
```
dliubarskyi: The idea is to have a `BedrockChatRequestParameters` (similar to `OpenAiChatRequestParameters`) that will have Bedrock-specific parameters. Could you please investigate if adding such a class is possible?

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
```diff
@@ -0,0 +1,94 @@
...
+        return BedrockChatModel.builder()
+                .defaultRequestParameters(parameters)
+                // force a working model with stopSequence parameter for @Tests
+                .modelId("cohere.command-r-v1:0")
+                .build();
+    }
+
+    @Override
+    protected boolean supportsDefaultRequestParameters() {
+        return false;
```
dliubarskyi: but you set `defaultRequestParameters` on the lne 38, so they are supported

dliubarskyi:
> My personal AWS account is limited to 1 request per minute for claude 3.5 models so I can't check the "I have manually run all the unit and integration tests in the module I have added/changed, and they are all green" because I had some "too many requests errors. But as far as I can tell tests are green. 

For such cases we usually add this:
```java
@AfterEach
    void afterEach() throws InterruptedException {
        String ciDelaySeconds = System.getenv("CI_DELAY_SECONDS_BEDROCK");
        if (ciDelaySeconds != null) {
            Thread.sleep(Integer.parseInt(ciDelaySeconds) * 1000L);
        }
    }
```
You can then set `CI_DELAY_SECONDS_BEDROCK` env variable to a preferable setting

> Also I tried to add LLama 3.2 to the list of models for IT tests but my request to have access was denied (maybe because I'm in Europe or for rate limiting... same thing with LLama 3.3 70B, I did'nt try to ask access to older LLama)

Same for me :(


> I had to extends three times AbstractChatModelIT because of the diffrentes capabilities of the bedrock model catalog. Maybe the List could be a List with ChatLanguageModelCapabilities as a class with a model instance and booleans showing its capabilities (tool use, image content, ...)

Could you please give an example how would htis look like?

dliubarskyi:
All thests are green on my side, well done!

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    protected boolean supportsToolChoiceRequired() {
+        return false;
```
dliubarskyi: isn't it supported?

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    @Override
+    protected boolean supportsJsonResponseFormat() {
+        return false;
```
dliubarskyi: same here... could you please leave a comment (in the code) for each case when `supportsXyz()` returns `false`? It will be much easier to understand why it is not supported now and for the future updates 🙏

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelWithConverseAPIIT.java
```diff
@@ -0,0 +1,296 @@
...
+
+@EnabledIfEnvironmentVariable(named = "AWS_SECRET_ACCESS_KEY", matches = ".+")
+class BedrockChatModelWithConverseAPIIT {
+
+    private static final String CAT_IMAGE_URL =
+            "https://upload.wikimedia.org/wikipedia/commons/e/e9/Felis_silvestris_silvestris_small_gradual_decrease_of_quality.png";
+
+    //    @BeforeEach
+    //    void pauseBeforeTest() throws InterruptedException {
+    //        Thread.sleep(60000); // AWS bedrock has a default 1 request per minute quota for claude 3.5 sonnet
```
dliubarskyi: Please use `CI_DELAY_SECONDS_BEDROCK`

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockAiServicesIT.java
```diff
@@ -0,0 +1,20 @@
...
+import static dev.langchain4j.model.bedrock.converse.TestedModels.AWS_NOVA_PRO;
+import static dev.langchain4j.model.bedrock.converse.TestedModels.CLAUDE_3_HAIKU;
+import static dev.langchain4j.model.bedrock.converse.TestedModels.COHERE_COMMAND_R_PLUS;
+import static dev.langchain4j.model.bedrock.converse.TestedModels.MISTRAL_LARGE;
+
+import dev.langchain4j.model.chat.ChatLanguageModel;
+import dev.langchain4j.service.common.AbstractAiServiceIT;
+import java.util.List;
+
+public class BedrockAiServicesIT extends AbstractAiServiceIT {
```
dliubarskyi: Please add `@EnabledIfEnvironmentVariable(named = "AWS_SECRET_ACCESS_KEY", matches = ".+")` to this and all other new tests

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
```diff
@@ -0,0 +1,94 @@
...
+                .parameters(parameters)
+                .build();
+
+        // when
+        ChatResponse chatResponse = chat(model, chatRequest).chatResponse();
+
+        // then
+        AiMessage aiMessage = chatResponse.aiMessage();
+        assertThat(aiMessage.text()).containsIgnoringCase("Hello");
+        assertThat(aiMessage.text()).containsIgnoringCase("World");
```
dliubarskyi: I am not sure this tests actually tests it. Perhaps stop word should be hello? Then hello sohuld be present, world should be not

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelWithConverseAPIIT.java
```diff
@@ -0,0 +1,296 @@
...
+    //    }
+
+    @Test
+    void should_generate_with_default_config() {
```
dliubarskyi: Are all these tests in this class required? There seems to be a big overlap with inherited tests.
My concern is to not have duplicated tests as it slows down CI and makes it more expensive. But if tests are actually required, then it is ok

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelWithVisionIT.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    protected String customModelName() {
+        return "cohere.command-r-v1:0";
+    }
+
+    @Override
+    protected ChatRequestParameters createIntegrationSpecificParameters(int maxOutputTokens) {
+        return ChatRequestParameters.builder().maxOutputTokens(maxOutputTokens).build();
```
dliubarskyi: same here

dliubarskyi (commented):
@tenpigs267 thanks a lot, great job!

Do you see a better way to slice abstract tests to make it easier for integrations (Bedrock in this case) inherit those tests and override them?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
tenpigs267: Yes, it is done to satisfy tests but also to provide consistent error feedback to developers. The ChatLanguageModel raises an UnsupportedFeatureException when a feature is not supported, instead of various exceptions depending on the implementation. And maybe, UnsupportedFeatureException could be a checked exception.
The modelId (String) determine the capabilities of the model so regexp (using model name but no versions) were the more flexible way I could think of to detect unsupported features. Maybe using regexp to detect models which don't support a feature would be more future proof -> it would detect known limitations and let aws sdk raise exception in case of unknown model?

langchain4j-parent/pom.xml
tenpigs267: Yes they are in bedrockruntime

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
tenpigs267: It's a broad question. I moved it into the `converse` package to differentiate from implementations using the `InvokeAPI`. The `ConverseAPI` is a higher level of abstraction than the `InvokeAPI` since it allows access to all models in the AWS Bedrock catalog with a unified API (though some settings are not accessible at the primary level). To simplify maintenance efforts at the `langchain4j` level, it might be interesting to maintain only the `ConverseAPI` and mark the `ChatLanguageModel` using the `InvokeAPI` as `@deprecated`. However, I haven't documented myself enough on the roadmap or the differences between the two APIs to confirm that this is the right path to take (for example, according to LangChain: *"the converse API does not yet support custom Bedrock models"* [[source](https://python.langchain.com/api_reference/aws/chat_models/langchain_aws.chat_models.bedrock_converse.ChatBedrockConverse.html)](https://python.langchain.com/api_reference/aws/chat_models/langchain_aws.chat_models.bedrock_converse.ChatBedrockConverse.html)). 

Anyway, maybe I can the `BedrockChatModel` back into the parent folder and have two packages, `converse` and `invoke`, to differentiate the utilities/abstractions of each API?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
tenpigs267: because of AbstractBaseChatModelIT line 598 which compare ignoring white spaces but is sensitive to lines : assertThat(toolExecutionRequest.arguments()).isEqualToIgnoringWhitespace("{\"city\":\"Munich\"}");

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
tenpigs267: Only by Claude 3 and Mistral Large : https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolChoice.html

tenpigs267:
> > I had to extends three times AbstractChatModelIT because of the diffrentes capabilities of the bedrock model catalog. Maybe the List could be a List with ChatLanguageModelCapabilities as a class with a model instance and booleans showing its capabilities (tool use, image content, ...)
> 
> Could you please give an example how would htis look like?

```java
public record ChatLanguageModelCapabilities(
    ChatLanguageModel chatLanguageModel,
    boolean supportsModelNameParameter,
    boolean supportsMaxOutputTokensParameter,
    boolean supportsStopSequencesParameter,
    boolean supportsTemperatureParameter,
    boolean supportsTopPParameter,
    boolean ...
) {}
```

The idea would be to have tests like:

```java
@ParameterizedTest
@MethodSource("models")
protected void shouldHandleStopSequencesBasedOnModelCapabilities(ChatLanguageModelCapabilities modelCapabilities) {
    if (modelCapabilities.supportsStopSequencesParameter()) {
        // should support stop sequences
    } else {
        // should fail
    }
}
```

The current combination:

```java
@ParameterizedTest
@MethodSource("models")
@DisabledIf("supportsStopSequencesParameter")
protected void should_fail_if_stopSequences_parameter_is_not_supported(M model) { ... }

@ParameterizedTest
@MethodSource("models")
@EnabledIf("supportsStopSequencesParameter")
protected void should_respect_stopSequences_in_chat_request(M model) { ... }
```

is not as flexible. The combination of `@EnabledIf` and `@DisabledIf` means either you test models that support the feature or those that don't, but you can't test in the same integration test both models that should fail and those that should pass the test.  

Otherwise, you would need to create a separate `MethodSource` for each test and not rely on the `supportsXxx()` flags:  

```java
@ParameterizedTest
@MethodSource("modelsSupportingStopSequencesParam")
protected void should_fail_if_stopSequences_parameter_is_not_supported(M model) { ... }

@ParameterizedTest
@MethodSource("modelsNotSupportingStopSequencesParam")
protected void should_respect_stopSequences_in_chat_request(M model) { ... }
```

I think ChatLanguageModelCapabilities would be more self explanatory than having multiple source methods for List<ChatLanguageModel>

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
tenpigs267: Response<AiMessage> generate(final List<ChatMessage> messages) has no default and public Response<AiMessage> generate(List<ChatMessage> messages, ToolSpecification toolSpecification) default raise an exception

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
tenpigs267: oups

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
tenpigs267: OpenAiChatRequestParameters propose more parameters than DefaultChatRequestParameters but Bedrock converseAPI out of the box support less parameters thant DefaultChatRequestParameters. Not sure what to do with it. I will come back to this point later

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelWithConverseAPIIT.java
tenpigs267: keeping a generate() call test, a multiple functions call, a function without arguments call and a PDF content.
Except the generate() test, others are legitimate to add in AbstractBaseChatModelIT

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsDocumentConverter.java
tenpigs267: I'm not a big fan of ToolSpecification to json string and json string to Document  instead of ToolSpecification to Document but it remove 30 lines so okay.
I made Claude generate unit tests for AwsDocumentConverter.

tenpigs267:
@dliubarskyi I fixed or answered your comments, can you take a look at it? Thanks

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
dliubarskyi: Do you know what is the AWS's long-term plan for these APIs? If converse will eventuall yreplace invoke, I guess we can put `BedrockChatModel` into the parent package and gradually deprecate older model classes. In the long term, it would probably make no sense to keep it int `.converse` package if it will be the only implementation.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
dliubarskyi: > Yes, it is done to satisfy tests

I think we need to adapt tests to allign with the implementation instead of adapting implementation to allign with tests

>  but also to provide consistent error feedback to developers.

If we'll delegate this validation to Bedrock, it will also be consistent, no?

> And maybe, UnsupportedFeatureException could be a checked exception.

I would avoid bothering our users with handling `UnsupportedFeatureException` every time they want to call a model

> The modelId (String) determine the capabilities of the model so regexp (using model name but no versions) were the more flexible way I could think of to detect unsupported features

Imagine at some point Bedrock will release a new model family, the most capable one and with more supported modalities (as it usually happens). This regex logic will not handle it well. Users will not be able to switch to the new model right away, they will have to wait for 1+ months untill we release a new version with the updated regex.

> Maybe using regexp to detect models which don't support a feature would be more future proof -> it would detect known limitations and let aws sdk raise exception in case of unknown model?

Can't older model families get new capabilities?

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
dliubarskyi: Oh, interesting. Agreed, we can postpone creation of `BedrockChatRequestParameters` untill Bedrock will accept more params

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsDocumentConverter.java
```diff
@@ -58,25 +40,46 @@ public static String documentToJson(Document document) {
         }
     }
 
+    private static Object documentToObject(Document doc) {
```
dliubarskyi: In order to extensively test this mapping logic, could you please also extend `AiServicesWithNewToolsIT`, the same way we do for other models supporting tools?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsLoggingInterceptor.java
```diff
@@ -0,0 +1,95 @@
...
+import software.amazon.awssdk.core.interceptor.ExecutionAttributes;
+import software.amazon.awssdk.core.interceptor.ExecutionInterceptor;
+import software.amazon.awssdk.http.ContentStreamProvider;
+import software.amazon.awssdk.http.SdkHttpFullRequest;
+import software.amazon.awssdk.http.SdkHttpMethod;
+import software.amazon.awssdk.http.SdkHttpRequest;
+import software.amazon.awssdk.http.SdkHttpResponse;
+import software.amazon.awssdk.utils.IoUtils;
+
+public class AwsLoggingInterceptor implements ExecutionInterceptor {
```
dliubarskyi:
```suggestion
class AwsLoggingInterceptor implements ExecutionInterceptor {
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsLoggingInterceptor.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(AwsLoggingInterceptor.class);
+
+    private final boolean logRequests;
+    private final boolean logResponses;
+
+    public AwsLoggingInterceptor(final boolean logRequests, final boolean logResponses) {
+        super();
```
dliubarskyi:
```suggestion
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsLoggingInterceptor.java
```diff
@@ -0,0 +1,95 @@
...
+                } catch (IOException e) {
+                    logger.warn("Unable to obtain request body", e);
+                }
+            }
+            logger.debug(
+                    "Request:\n- method: {}\n- url: {}\n- headers: {}\n- body: {}{}",
+                    request.method(),
+                    request.getUri(),
+                    request.headers(),
+                    request.rawQueryParameters(),
```
dliubarskyi: nit: perhaps swap query params and body? Otherwite it looks like this (I guess it the most cases query params are empty, so there is no need for user to see empty brackets first?)

`- body: {}{"messages":[{"role":"user","content":[{"text":"I live in Munich"}]}],"system":[],"inferenceConfig":{"temperature":0.1},"toolConfig":{"tools":[{"toolSpec":{"name":"getWeather","inputSchema":{"json":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}}},{"toolSpec":{"name":"add_two_numbers","inputSchema":{"json":{"type":"object","properties":{"a":{"type":"integer"},"b":{"type":"integer"}},"required":["a","b"]}}}}],"toolChoice":{"any":{}}}}`

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -147,50 +148,40 @@ public class BedrockChatModel implements ChatLanguageModel {
...
-        this.maxRetries = getOrDefault(maxRetries, 3);
-        this.timeout = getOrDefault(timeout, Duration.ofMinutes(1));
-        this.client = isNull(client) ? createClient() : client;
+        this(builder().modelId(modelId));
+    }
+
+    public BedrockChatModel(Builder builder) {
+        this.region = getOrDefault(builder.region, Region.US_EAST_1);
+        this.credentialsProvider = getOrDefault(builder.credentialsProvider, DefaultCredentialsProvider.create());
+        this.modelId = builder.modelId;
```
dliubarskyi:
```suggestion
        this.modelId = ensureNotBlank(getOrDefault(builder.modelId, builder.defaultRequestParameters.modelName()), "modelId");
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -147,50 +148,40 @@ public class BedrockChatModel implements ChatLanguageModel {
...
+        this.maxRetries = getOrDefault(builder.maxRetries, 3);
+        this.timeout = getOrDefault(builder.timeout, Duration.ofMinutes(1));
+        this.client = isNull(builder.client) ? createClient(builder.logRequests, builder.logResponses) : builder.client;
         this.defaultRequestParameters = ChatRequestParameters.builder()
                 .modelName(getOrDefault(
-                        modelId, nonNull(defaultRequestParameters) ? defaultRequestParameters.modelName() : null))
+                        modelId,
```
dliubarskyi: What about keeping only `defaultRequestParameters(...)` in the model builder and remove all the `temperature(...)` and `maxTokens(...)` and similar methods? This way there will be less confusion for the user and less complicated implementation.

I had to keep both `defaultRequestParameters(...)` and all other specific setters in the OpenAI module for the backward compatibility, but this is entirely new implementation, so we do not have to support both.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -642,76 +636,76 @@ public static class BedrockChatModelBuilder {
...
         }
 
-        public BedrockChatModelBuilder client(BedrockRuntimeClient client) {
+        public Builder client(BedrockRuntimeClient client) {
             this.client = client;
             return this;
         }
 
-        public BedrockChatModelBuilder defaultRequestParameters(ChatRequestParameters defaultRequestParameters) {
+        public Builder logRequests(boolean logRequests) {
```
dliubarskyi:
```suggestion
        public Builder logRequests(Boolean logRequests) {
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
```diff
@@ -642,76 +636,76 @@ public static class BedrockChatModelBuilder {
...
+            this.logRequests = logRequests;
+            return this;
+        }
+
+        public Builder logResponses(boolean logResponses) {
```
dliubarskyi:
```suggestion
        public Builder logResponses(Boolean logResponses) {
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
dliubarskyi: We usually use `Boolean` in the builder to allow user to pass `null`s form their configuration and then do `getOrDefault(logRequests, false)` in the ctor

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/converse/BedrockChatModelNovaWithVisionIT.java
dliubarskyi: Could you please leave a comment in the code with this link?

---

Also, there are multiple 
```java
    @Override
    protected boolean supportsXyz() {
        return false;
    }
```
left in this test package without any explanation, could you please add comments?

There are so many supported LLM providers and LLMs, it is hard to keep track of what is supported and what is not and why. Having these comments should help to understand

dliubarskyi (commented):
@tenpigs267 thanks a lot!

dliubarskyi:
> > > I had to extends three times AbstractChatModelIT because of the diffrentes capabilities of the bedrock model catalog. Maybe the List could be a List with ChatLanguageModelCapabilities as a class with a model instance and booleans showing its capabilities (tool use, image content, ...)
> > 
> > 
> > Could you please give an example how would htis look like?
> 
> ```java
> public record ChatLanguageModelCapabilities(
>     ChatLanguageModel chatLanguageModel,
>     boolean supportsModelNameParameter,
>     boolean supportsMaxOutputTokensParameter,
>     boolean supportsStopSequencesParameter,
>     boolean supportsTemperatureParameter,
>     boolean supportsTopPParameter,
>     boolean ...
> ) {}
> ```
> 
> The idea would be to have tests like:
> 
> ```java
> @ParameterizedTest
> @MethodSource("models")
> protected void shouldHandleStopSequencesBasedOnModelCapabilities(ChatLanguageModelCapabilities modelCapabilities) {
>     if (modelCapabilities.supportsStopSequencesParameter()) {
>         // should support stop sequences
>     } else {
>         // should fail
>     }
> }
> ```
> 
> The current combination:
> 
> ```java
> @ParameterizedTest
> @MethodSource("models")
> @DisabledIf("supportsStopSequencesParameter")
> protected void should_fail_if_stopSequences_parameter_is_not_supported(M model) { ... }
> 
> @ParameterizedTest
> @MethodSource("models")
> @EnabledIf("supportsStopSequencesParameter")
> protected void should_respect_stopSequences_in_chat_request(M model) { ... }
> ```
> 
> is not as flexible. The combination of `@EnabledIf` and `@DisabledIf` means either you test models that support the feature or those that don't, but you can't test in the same integration test both models that should fail and those that should pass the test.
> 
> Otherwise, you would need to create a separate `MethodSource` for each test and not rely on the `supportsXxx()` flags:
> 
> ```java
> @ParameterizedTest
> @MethodSource("modelsSupportingStopSequencesParam")
> protected void should_fail_if_stopSequences_parameter_is_not_supported(M model) { ... }
> 
> @ParameterizedTest
> @MethodSource("modelsNotSupportingStopSequencesParam")
> protected void should_respect_stopSequences_in_chat_request(M model) { ... }
> ```
> 
> I think ChatLanguageModelCapabilities would be more self explanatory than having multiple source methods for List

I really like the idea!
Although this change can be quite big for this PR. There are multiple modules already relying on that stracture. How do you prefer to handle it?

dliubarskyi:
Just co double check, so if the `ChatLanguageModelCapabilities` idea will be implemented, we can merge `BedrockChatModelNovaWithVisionIT`, `BedrockChatModelWithoutVisionIT` and `BedrockChatModelWithVisionIT`, right?

tenpigs267:
> I really like the idea! Although this change can be quite big for this PR. There are multiple modules already relying on that stracture. How do you prefer to handle it?

Yes another PR, we will keep this one focused on bedrock.
And yes with a ChatLanguageModelCapabilities we will be able to simplify to one BedrockAbstractChatModelIT instead of the three BedrockChatModelNovaWithVisionIT, BedrockChatModelWithoutVisionIT and BedrockChatModelWithVisionIT.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
tenpigs267: > > but also to provide consistent error feedback to developers.
> 
> If we'll delegate this validation to Bedrock, it will also be consistent, no?
> 
Bedrock throws a software.amazon.awssdk.services.bedrockruntime.model.ValidationException so if developer want to catch this he ties his code to the aws sdk instead of a langchain4j UnsupportedFeatureException (which could remains valid when he changes the model he uses in his app), that was my point.

I hadn't been attentive enough; assertExceptionType() allows to get rid of the check on the type of exception thrown and the content of the message. I think this flag should be set to false by default, let me know if you want me to do it in this PR.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
tenpigs267: I have no clue about AWS's plan. On Spring AI, they deprecated individual model implementation in favor of the one that use ConverseAPI saying it is "Following Amazon Bedrock's recommandation" but I can't find any official mention of this in AWS Bedrock documentation or blog...

tenpigs267: Oh here : https://docs.aws.amazon.com/bedrock/latest/userguide/conversation-inference.html : "we recommend using the Converse API as it provides consistent API ..."

I just moved the BedrockChatModel back in the bedrock package and added a Javadoc comment stating that "BedrockChatModel uses the Bedrock ConverseAPI" and a link to the AWS documentation about ConverseAPI.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
dliubarskyi: > Bedrock throws a software.amazon.awssdk.services.bedrockruntime.model.ValidationException so if developer want to catch this he ties his code to the aws sdk instead of a langchain4j UnsupportedFeatureException (which could remains valid when he changes the model he uses in his app), that was my point.

Driving the logic of the application with exceptions is a well-known antipattern. IMHO zser should not be catching neither `ValidationException` nor `UnsupportedFeatureException`, as both are indicating that user does something wrong.

> I hadn't been attentive enough; assertExceptionType() allows to get rid of the check on the type of exception thrown and the content of the message. I think this flag should be set to false by default, let me know if you want me to do it in this PR.

All the flags are set to `true` for a good reason - we want to have as uniform behaviour between models as possible. So if some model is somehow different, it sohuld be an exception situation that will force the developer to override the `supportsXyz()` method and explain the reason why this model does not adhere to the common behaviour. Having `false` by default will basically not test anything

tenpigs267: **Upstream detection** seemed to me the simplest approach to implement. Looking at what is the modelId being used. Its capabilities are documented, so I can validate the inputs and raise an `UnsupportedFeatureException`. Using a regex allowed me to avoid listing all the model ids + their versions explicitly, while being slightly less sensitive to catalog evolution.

Another possible approach would be **reacting downstream**: letting Bedrock return an exception, which we would catch and then throw an `UnsupportedFeatureException`. The issue here is that at this level, we’re only working with a string (the exception message), but I might be missing more structured error reporting information due to my lack of familiarity. This makes it hard to understand the issue and determine whether it’s an unsupported feature or a misconfiguration, incorrect usage, or a bug in our langchain4j bedrock code. 

Moreover, at this level, it seems to me impossible to guarantee that the returned message would contain specific patterns like `"ToolChoice.REQUIRED"` or even `"not support"` as asserted in the should_fail_XXX tests

If another approach has been used in any of the `llangchain4j` implementations, I’d be curious to see it and take inspiration from it.

Thanks!

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsDocumentConverter.java
dliubarskyi: please move it to the `dev.langchain4j.model.bedrock` package and make package-private

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/AwsLoggingInterceptor.java
dliubarskyi: please move it to the `dev.langchain4j.model.bedrock` package and make package-private

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -0,0 +1,598 @@
...
+    private final Duration timeout;
+    private final BedrockRuntimeClient client;
+    private final ChatRequestParameters defaultRequestParameters;
+
+    public BedrockChatModel(String modelId) {
+        this(builder().modelId(modelId));
+    }
+
+    public BedrockChatModel(Builder builder) {
+        this.region = getOrDefault(builder.region, Region.US_EAST_1);
```
dliubarskyi, line 93: nit: do we really need to have `US_EAST_1` as a default value?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -0,0 +1,598 @@
...
+        this.credentialsProvider = getOrDefault(builder.credentialsProvider, DefaultCredentialsProvider.create());
+        this.modelId = ensureNotBlank(
+                getOrDefault(
+                        builder.modelId,
+                        nonNull(builder.defaultRequestParameters)
+                                ? builder.defaultRequestParameters.modelName()
+                                : null),
+                "modelId");
+        this.maxRetries = getOrDefault(builder.maxRetries, 3);
+        this.timeout = getOrDefault(builder.timeout, Duration.ofMinutes(1));
```
dliubarskyi, line 102: `timeout` value will be used only if user did not specify the client, same for `credentialsProvider`. Do you think we should allow configuring these values in the builder? Perhaps if user wants to customize the `client`, they should provide the `client` instance?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -0,0 +1,598 @@
...
+        this.client = isNull(builder.client)
+                ? createClient(getOrDefault(builder.logRequests, false), getOrDefault(builder.logResponses, false))
+                : builder.client;
+        this.defaultRequestParameters = ChatRequestParameters.builder()
```
dliubarskyi: I guess this can be simplified to something like:

```java
        if (builder.defaultRequestParameters != null) {
            this.defaultRequestParameters = builder.defaultRequestParameters;
        } else {
            this.defaultRequestParameters = ChatRequestParameters.builder()
                    .modelName(this.modelId)
                    .maxOutputTokens(builder.maxTokens)
                    .build();
        }
```

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -0,0 +1,598 @@
...
+        // The name can only contain the following characters:
+        // Alphanumeric characters, Whitespace characters (no more than one in a row), Hyphens, Parentheses, Square,
+        // brackets
+        try {
+            final String filename = Paths.get(uri).getFileName().toString();
+            int dotIndex = filename.lastIndexOf('.');
+            String filenameWithoutExtension = dotIndex > 0 ? filename.substring(0, dotIndex) : filename;
+            return filenameWithoutExtension.replace(".", "-");
+        } catch (Exception e) {
+            return "document";
```
dliubarskyi: This does not seem right, could you please explain?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -0,0 +1,598 @@
...
+                : extractURIFileExtension(imageContent.image().url());
+        return ContentBlock.builder()
+                .image(ImageBlock.builder()
+                        .format(imgFormat)
+                        .source(ImageSource.builder().bytes(bytes).build())
+                        .build())
+                .build();
+    }
+
+    private String extractURIFileExtension(URI url) {
```
dliubarskyi: This seems to be error prone. What if url is something like `http://test.com/image.jpg?some-key=some-value`?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -0,0 +1,598 @@
...
+        // Extract the file extension
+        String extension = "";
+        int lastDotIndex = path.lastIndexOf('.');
+        if (lastDotIndex != -1 && lastDotIndex < path.length() - 1) {
+            extension = path.substring(lastDotIndex + 1);
+        }
+        return extension;
+    }
+
+    private String extractImageFormat(String mimeType) {
```
dliubarskyi: Please add using tests to this method

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -0,0 +1,598 @@
...
+                                    pdfFileContent.pdfFile().url()))
+                            .build())
+                    .build();
+        } else if (content instanceof ImageContent image) {
+            return createImageBlock(image);
+        }
+        throw new IllegalArgumentException("Unsupported content type: " + content.getClass());
+    }
+
+    private static String extractFilenameWithoutExtensionFromUri(URI uri) {
```
dliubarskyi, line 343: Please add using tests for this method

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
dliubarskyi: Please add using tests for this method

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/BedrockChatModelWithConverseAPIIT.java
```diff
@@ -0,0 +1,146 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
+@EnabledIfEnvironmentVariable(named = "AWS_SECRET_ACCESS_KEY", matches = ".+")
+class BedrockChatModelWithConverseAPIIT {
```
dliubarskyi, line 34: Does this test work for you?

`should_call_multiple_functions` and `should_generate_with_default_config` are constantly failing for me with `Status Code: 429`, any idea?

For `should_call_multiple_functions`, perhaps the delay should be added before calling the model second time?

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -0,0 +1,598 @@
...
+            this.region = region;
+            return this;
+        }
+
+        public Builder credentialsProvider(AwsCredentialsProvider credentialsProvider) {
+            this.credentialsProvider = credentialsProvider;
+            return this;
+        }
+
+        public Builder modelId(String modelId) {
```
dliubarskyi, line 507: I am wondering if we really need these specific builder methods for `modelId`, `maxTokens`. Perhaps users should provide those values via `defaultRequestParameters`, so we could simplify the logic a bit, WDYT?

dliubarskyi (commented):
@tenpigs267 thanks a lot! Looks good, only a few minor things left and we can merge it and include in the release at the end of this week.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/converse/BedrockChatModel.java
dliubarskyi: Usually particular feature (e.g., tools, image inputs, etc.) is either supported or not, so there was no need previously...

> Upstream detection seemed to me the simplest approach to implement. Looking at what is the modelId being used. Its capabilities are documented, so I can validate the inputs and raise an UnsupportedFeatureException. Using a regex allowed me to avoid listing all the model ids + their versions explicitly, while being slightly less sensitive to catalog evolution

I completely understand, but there is a risk that such logic will prohibit use of some features for some models due to our logic being too strict and/or not up-to-date.

> Another possible approach would be reacting downstream: letting Bedrock return an exception, which we would catch and then throw an UnsupportedFeatureException. The issue here is that at this level, we’re only working with a string (the exception message), but I might be missing more structured error reporting information due to my lack of familiarity. This makes it hard to understand the issue and determine whether it’s an unsupported feature or a misconfiguration, incorrect usage, or a bug in our langchain4j bedrock code.
> 
> Moreover, at this level, it seems to me impossible to guarantee that the returned message would contain specific patterns like "ToolChoice.REQUIRED" or even "not support" as asserted in the should_fail_XXX tests

I guess we could catch a `software.amazon.awssdk.services.bedrockruntime.model.ValidationException` and perhaps check if it's message contains "doesn't support" text. If it contains it, map it into the `UnsupportedFeatureException`. Otherwise just re-throw it as-is.
I know, it is an ugly solution, but I do not see any better alternatives right now. Even if the message will change, it will just throw an original exception, which is not the end of the world. Since we have tests for that, it is pretty safe IMHO.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
tenpigs267: I still think it's nice for developper to be able to do just 
BedrockChatModel bedrockChatModel = new BedrockChatModel("anthropic.claude-3-5-sonnet-20240620-v1:0");
But that's your call

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
tenpigs267: timeout seems to me a first level parameter that a dev should be able to modify without the hassle to instantiate a complete client.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
tenpigs267: It feels strange to me to instantiate a `BedrockChatModel` with its default request parameters stating that this is a Claude Sonnet model client, instead of directly instantiating a `BedrockChatModel` with the model name. Providing the model name directly seems more intuitive.

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
dliubarskyi: ok, makes sense

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
```diff
@@ -107,27 +103,15 @@ public BedrockChatModel(Builder builder) {
...
+                ? ChatRequestParameters.builder().modelName(this.modelId).build()
+                : ChatRequestParameters.builder()
+                        .modelName(this.modelId)
+                        .temperature(builder.defaultRequestParameters.temperature())
+                        .maxOutputTokens(builder.defaultRequestParameters.maxOutputTokens())
+                        .topP(builder.defaultRequestParameters.topP())
+                        .stopSequences(copyIfNotNull(builder.defaultRequestParameters.stopSequences()))
+                        .build();
 
         validate(this.defaultRequestParameters);
```
dliubarskyi: this will validate paramets after copying only supported ones from `builder.defaultRequestParameters`, so it means that if some unsupported parameter is passed in the `builder.defaultRequestParameters`, it will be silently ignored

langchain4j-bedrock/src/test/java/dev/langchain4j/model/bedrock/BedrockChatModelWithConverseAPIIT.java
tenpigs267: 429 (too many request) is the quota with 1 request per minute for Claude models. You are right I added the delay before the second model call

langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatModel.java
tenpigs267: Yes, I forgot about these weak code parts (PDF and image).
I made some refactor for both, it should be more reliable.

https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_DocumentBlock.html
DocumentBlocks must have a name for bedrock so I try to get a name from the URI or I generate a random one to avoid name collision (which is not allowed so "document" was a bad idea)

dliubarskyi (dismissed):
@tenpigs267 thanks a lot, great job!

Could you please update the [documentation](https://docs.langchain4j.dev/integrations/language-models/amazon-bedrock) and [examples](https://github.com/langchain4j/langchain4j-examples/tree/main/bedrock-examples/src/main/java) ?

dliubarskyi:
@tenpigs267 if you prefer to update the documentation in a separate PR - please let me know, I will then merge this one

tenpigs267:
I just have added some documentation (keeping the previous content in a subsection "ChatLanguageModel using InvokeAPI"). 
In the index.md table I duplicated the Amazon bedrock line to reflect the differences between the two implementations. I hope I set links the good way.
Let me know if it's not okay.

I am going to add a PR in the example repo

Thank you for your patience with the reviews

docs/docs/integrations/language-models/index.md
```diff
@@ -4,26 +4,27 @@ hide_title: false
...
+|-----------------------------------------------------------------------------------------------------------------------|--------------------------------------------|---------------------------------------------------------------------|--------------------------------------------------------------------|-----------------------------------------------|--------------------------------|-------------------------------------------|---------------------------------------------------|-----------------------|-----------------------------|
+| [Amazon Bedrock(using ConverseAPI)](/integrations/language-models/amazon-bedrock#chatlanguagemodel-using-converseapi) |                                            | ✅                                                                   |                                                                    |                                               | text, image, PDF               | ✅                                         |                                                   |                       |                             |
```
dliubarskyi: Is streaming supported by Converse API at all? If yes, do you plan to implement it in the future?
Would be nice to put some indication (e.g., 🆘 or ❌)

Same regarding tools in a streaming mode.

Observability does not seem to be implemneted for Converse API, so ✅there is wrong

dliubarskyi (commented):
@tenpigs267 thank you!

docs/docs/integrations/language-models/index.md
tenpigs267: Streaming is supported by ConverseAPI https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html. but I don’t plan to implement it in the near future. My current use cases are not chat-related. I have created a new issue : https://github.com/langchain4j/langchain4j/issues/2470 and set it with the SOS flag in the table.
And you are right Observability is not implemented.

dliubarskyi: @tenpigs267 great, thanks!

dliubarskyi (approved):
@tenpigs267 well done, thank you!

PR #2408, opened 2025-01-13
fix: Add check for zero norm in normalize() method to handle zero vector

minglu7:
## Issue

Closes #

## Change
- Added a check to return early if the norm is zero to prevent division by zero.
- Ensures that zero vectors are handled gracefully without causing runtime errors.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/data/embedding/Embedding.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/data/embedding/Embedding.java
@@ -55,7 +55,6 @@ public class Embedding {
         }
         norm = Math.sqrt(norm);
         if (Math.abs(norm) < 1e-10) {
-            Arrays.fill(vector, 0.0f);
             return;
         }
         for (int i = 0; i < vector.length; i++) {
--- a/langchain4j-core/src/main/java/dev/langchain4j/data/embedding/Embedding.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/data/embedding/Embedding.java
@@ -54,7 +54,8 @@ public class Embedding {
             norm += f * f;
         }
         norm = Math.sqrt(norm);
-        if (norm == 0.0) {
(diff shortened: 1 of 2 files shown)


minglu7:
When I was looking at the source code I accidentally discovered a small problem with this code, so I tried to optimize it

langchain4j-core/src/main/java/dev/langchain4j/data/embedding/Embedding.java
```diff
@@ -54,9 +54,11 @@ public void normalize() {
             norm += f * f;
         }
         norm = Math.sqrt(norm);
-
+        if (norm == 0.0) {
```
dliubarskyi: is it the right way to check for zero?

dliubarskyi (commented):
@minglu7 thank you!

langchain4j-core/src/main/java/dev/langchain4j/data/embedding/Embedding.java
minglu7: @dliubarskyi  Thanks for the reminder, I have corrected this method

langchain4j-core/src/main/java/dev/langchain4j/data/embedding/Embedding.java
```diff
@@ -54,7 +54,8 @@ public void normalize() {
             norm += f * f;
         }
         norm = Math.sqrt(norm);
-        if (norm == 0.0) {
+        if (Math.abs(norm) < 1e-10) {
+            Arrays.fill(vector, 0.0f);
```
dliubarskyi: do we àctually have to modify the array in this case?

minglu7: > do we àctually have to modify the array in this case?

may be not,  so you can delete this line if need

dliubarskyi (approved):
@minglu7 thank you!

PR #2413, opened 2025-01-14
Customizable HTTP clients

### LINKED ISSUE #1044: Decouple modules from HTTP client (currently okhttp/retrofit)

### LINKED ISSUE #2365: [BUG] When using Ollama and a specified model name does not exist, a java.lang.NullPointerException is thrown.
**Describe the bug**

When using Ollama and a specified model name does not exist, a java.lang.NullPointerException is thrown.

image
`java.lang.NullPointerException: Cannot invoke "okhttp3.ResponseBody.byteStream()" because the return value of "retrofit2.Response.body()" is null
	at dev.langchain4j.model.ollama.OllamaClient$2.onResponse(OllamaClient.java:164) ~[langchain4j-ollama-1.0.0-alpha1.jar:na]
	at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:161) ~[retrofit-2.9.0.jar:na]
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519) ~[okhttp-4.12.0.jar:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:842) ~[na:na]`
**To Reproduce**

**Expected behavior**

throw IllegalConfigurationException

- LangChain4j version:
- LLM(s) used: 
- Java version: 

**Additional context**

### LINKED ISSUE #2497: HttpRequest: make sure header names with different casing work as expected
```java
HttpRequest.builder()
    .addHeader("accept", "application/json")
    .addHeader("Accept", "text/plain")
```

### LINKED ISSUE #2498: HttpRequest.Builder.url(baseUrl, path): make sure baseUrl and path are properly concatenated
See https://blog.lunatech.com/posts/2009-02-03-what-every-web-developer-must-know-about-url-encoding

dliubarskyi:
## Issue
Closes #1044
Fixes #2365

## Change
This is the first step towards decoupling modules from (ok)http client.
This PR introduces an abstraction layer (in `langchain4j-http-client` module) over various HTTP clients.
Main components:
- `HttpClient`
- `HttpClientBuilder`
- `HttpRequest`
- `SuccessfulHttpResponse`
- `HttpException`
- `ServerSentEvent`
- `ServerSentEventListener`

This PR also provides a single implementation of `HttpClient`: `JdkHttpClient` in the `langchain4j-http-client-jdk` module.

Another implementation (`SpringRestClient` in the `langchain4j-http-client-spring-restclient` module, which uses `RestClient` from Spring) is introduced in a [related PR](https://github.com/langchain4j/langchain4j-spring/pull/104).

The first module to migrate is `langchain4j-ollama`, this PR decouples it from OkHttp.
Users are able to provide a `HttpClientBuilder` when building/creating `Ollama*Model`. If not provided, default (`JdkHttpClient`) will be loaded via SPI.

With this change, request/response logging is also unified and moved to the `LoggingHttpClient` decorator.

## Related PRs:
- https://github.com/langchain4j/langchain4j-spring/pull/104

## Questions
- 
  - Q: Use `WebClient` for streaming HTTP calls in `langchain4j-http-client-spring-restclient`?
  - A: `RestClient` + `TaskExecutor` should do the job. Pros: Less dependencies, no need to configure/customize both `RestClient` and `WebClient`. Cons: need to manage `TaskExecutor`, make sure context is propagated. In case current approach will not  be sufficient in the future and `WebClient`-based implementation will be required, it could be added to the `langchain4j-http-client-spring-restclient` or implemented as a separate `langchain4j-http-client-spring-webclient` module.

## Next steps (out of scope of this PR)
- Get feedback from users
- Get feedback from @ehsavoie
- Consider migrating Quarkus-LC4j Ollama module to re-use LC4j Ollama module with Quarkus-specific HTTP client (cc @geoand)
- Migrate [`langchain4j-open-ai`](https://github.com/langchain4j/langchain4j/issues/2468), [`langchain4j-anthropic`](https://github.com/langchain4j/langchain4j/issues/2469), `langchain4j-mistral-ai` and `langchain4j-hugging-face` modules to use `HttpClient` (`JdkHttpClient` by default) instead of okhttp
- Fix https://github.com/langchain4j/langchain4j/issues/2497
- Fix https://github.com/langchain4j/langchain4j/issues/2498

## General checklist


kpavlov:
@dliubarskyi , I am feeling, that implementing one more abstraction on top of http clients might not be the right thing, because it will . Was it considered to use existing higher-level http client api. I am thinking about Ktor, because it is also cross platform...

dliubarskyi:
@kpavlov 

> I am feeling, that implementing one more abstraction on top of http clients might not be the right thing, because it will . 

Because it will what?

> Was it considered to use existing higher-level http client api. I am thinking about Ktor, because it is also cross platform...

We need something lightweight, without tonns of dependencies and supporting HTTP clients that we need: JDK's, Quarkus's rest client, Spring's RestClient, RESTEasy, etc. Do you know any such existing higher-level http client api?

http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Override
+    public SuccessfulHttpResponse execute(HttpRequest request) throws HttpException {
+        try {
+            java.net.http.HttpRequest jdkRequest = toJdkRequest(request);
+
+            java.net.http.HttpResponse<String> jdkResponse = delegate.send(jdkRequest, BodyHandlers.ofString());
+
+            if (!isSuccessful(jdkResponse)) {
+                throw new HttpException(jdkResponse.statusCode(), jdkResponse.body());
```
mariofusco, line 47: Just a doubt, I'm not strongly opinionated on this: do we always want to propagate an unsuccessful request as an `HttpException` and only return the response when it is successful? Couldn't we return the response regardless if it is successful or not? Is there any reason or constraint driving this decision? (in general I'm not a big fan of exception and I'm just trying to weight pros and cons)

dliubarskyi: Yes, I had this doubt as well. If we'll always return `HttpResponse` regardless of the status, we will need to check if `HttpResponse` is successful and throw an exception in all the places that call `HttpClient.execute(HttpRequest)`. This means having this logic duplicated in many places. For example, `OllamaClient` calls it 9 times.

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpException.java
```diff
@@ -0,0 +1,18 @@
+package dev.langchain4j.http.client;
+
+import dev.langchain4j.Experimental;
+
+@Experimental
+public class HttpException extends RuntimeException {
```
mariofusco: In general I prefer `RuntimeException`s over checked ones, but in this specific case, and since we decided to propagate all unsuccessful responses as `HttpException`s, I wonder if we shouldn't oblige client code to catch and manage this situation.

dliubarskyi: Hmm, I am not sure how e.g., `*ChatModel` should handle those exceptions? In my understanding we should propagate them to the caller. So if `HttpException` will be a checked exception, in all places that call `HttpClient`, we will have to wrap that call into try-catch block and wrap `HttpException` into `RuntimeException`

Consider 4XX and 5XX exceptions.
For 4XX, user is wrong, so they should see the exception and fix it.
For 5XX case, we could retry (what we already do), but eventually fail and propagate it to the user.

kpavlov:
> @kpavlov
> 
> > I am feeling, that implementing one more abstraction on top of http clients might not be the right thing, because it will .
> 
> Because it will what?
🤦🏻 sorry
... because it will look like reimplementing the wheel. Maybe it should be out of LC4j scope 

> 
> > Was it considered to use existing higher-level http client api. I am thinking about Ktor, because it is also cross platform...
> 
> We need something lightweight, without tonns of dependencies and supporting HTTP clients that we need: JDK's, Quarkus's rest client, Spring's RestClient, RESTEasy, etc. Do you know any such existing higher-level http client api?

To my knowledge, it seems that Ktor client is acually very close to it. It has API and low-level [Engines](https://ktor.io/docs/client-engines.html#platforms). Not completely the set of low-level clients to wrap, but maybe it's enough. I witnessed teams implementing such wrappers - it tends to have higher-than-originally-expected maintenance cost. Devil is in the details...

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
```diff
@@ -0,0 +1,69 @@
...
+
+import static dev.langchain4j.internal.ValidationUtils.ensureBetween;
+import static java.util.Collections.emptyMap;
+
+@Experimental
+public class SuccessfulHttpResponse {
+
+    private final int statusCode;
+    private final Map<String, List<String>> headers;
+    private final String body;
```
kpavlov, line 17: What is the idea of supporting streaming models? Should it be a separate HttpClient api or SuccessfulStreamingHttpResponse?

dliubarskyi:
> To my knowledge, it seems that Ktor client is acually very close to it. It has API and low-level [Engines](https://ktor.io/docs/client-engines.html#platforms). Not completely the set of low-level clients to wrap, but maybe it's enough.

As far as I see from the table, there is no support for Quarkus's rest client, Spring's RestClient, RESTEasy.
We do want support for those, so it is not very close to it.

> I witnessed teams implementing such wrappers - it tends to have higher-than-originally-expected maintenance cost. Devil is in the details...

Can you share a bit more details? What exactly was implemented? What kind of issues were found out later on? How many times did this happen?

I agree that it feels like reinventing the wheel, but I see no better alternative at this point. Time will tell.

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
dliubarskyi: For streaming, `body` will be `null` in `ServerSentEventListener.onOpen()`, actual body will be delivered via events in `onEvent()`.
I am not sure it deserves it's own `SuccessfulStreamingHttpResponse` just because `body` is `null`.

Regarding the separate API, I was considering to split `HttpClient` into 2: one sync, one for streaming. This can be done, but I am not sure this complication "for pureness" is justified: most models support either sync or sync+streaming. So splitting `HttpClient` into 2 will just means twice as many objects to juggle.

mariofusco:
> > I witnessed teams implementing such wrappers - it tends to have higher-than-originally-expected maintenance cost. Devil is in the details...
> 
> Can you share a bit more details? What exactly was implemented? What kind of issues were found out later on? How many times did this happen?
> 
> I agree that it feels like reinventing the wheel, but I see no better alternative at this point. Time will tell.

I agree that this feels a bit like reinventing the wheel, but also I don't see a better wheel to replace this one

My understanding is that we want to remove okhttp, adding one more abstraction layer, in order to allow the interoperability with other clients like RestClient for Spring Boot. It seems to me that introducing any other existing client, without this abstraction layer, will reintroduce the same problem that we're trying to solve by removing okhttp, or am I missing something?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpClient.java
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * TODO
+     *
+     * @param request
+     * @return
+     * @throws HttpException
+     * @throws RuntimeException
+     */
+    SuccessfulHttpResponse execute(HttpRequest request) throws HttpException, RuntimeException;
```
cescoffier, line 24: I believe it would be too big for this version, but at some point we need to think about async / non-blocking clients.

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpClientBuilder.java
```diff
@@ -0,0 +1,19 @@
+package dev.langchain4j.http.client;
+
+import dev.langchain4j.Experimental;
+
+import java.time.Duration;
+
+@Experimental
+public interface HttpClientBuilder {
+
+    Duration connectTimeout();
```
cescoffier, line 10: How would a user extend these? Typically, what about SSL, redirect handling...?

cescoffier: I'm actually wondering if it's really the goal of langchain4j to provide a HTTP client configuration object - as every client is slightly different.

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
```diff
@@ -0,0 +1,121 @@
...
+            ensureNotBlank(baseUrl, "baseUrl");
+            ensureNotBlank(path, "path");
+
+            if (baseUrl.endsWith("/")) {
+                baseUrl = baseUrl.substring(0, baseUrl.length() - 1);
+            }
+            if (path.startsWith("/")) {
+                path = path.substring(1);
+            }
+            this.url = baseUrl + "/" + path;
```
cescoffier, line 81: What about encoding?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        public Builder addHeader(String name, String... values) {
+            ensureNotBlank(name, "name");
+            ensureNotEmpty(values, "values");
+
+            if (this.headers == null) {
+                this.headers = new HashMap<>();
+            }
+            this.headers.put(name, asList(values));
```
cescoffier, line 92: HTTP headers are case-insensitive.

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpClientBuilder.java
dliubarskyi: I have included only timeouts here as it is common for users to configure those and every http client has this configuration. If more specific configuations are needed, users can do that on the concrete http client

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
dliubarskyi: Could you please elaborate?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
dliubarskyi: Good catch, thank you!

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
cescoffier: You cannot really append a path like that (only if this is safe).

See https://blog.lunatech.com/posts/2009-02-03-what-every-web-developer-must-know-about-url-encoding.

dliubarskyi: Ok, thank you!

PR #2418, opened 2025-01-14
Add splitter for Markdown, taking document hierarchy into account

### LINKED ISSUE #2417: [FEATURE] Add splitter for Markdown
For our project doing Graph RAG, we need to parse Markdown documents and break those into sections, before splitting those sections.

A splitter that splits markdown documents by header. Metadata in the resulting text segments to be able to determine where in the document structure an entry came from.

None :-)

**Additional context**
@ dliubarskyi suggested I contribute this splitter. I'll link a PR shortly

kabir:
## Issue

Closes #2417

## Change

I added a MarkdownSectionSplitter to split Markdown documents by headers.

The header goes into the metadata of the resulting TextSegments, while the body of the section goes into the TextSegments.

The Markdown must be well-formed, we don't try to do any validation beyond what the user gives us.

The list of TextSegments list the sections in order, while the metadata of each segment allows us to find where in the document the segment came from.

Headers appearing in code blocks are ignored (some languages use '#' for comments). 

A user may optionally specify a splitter to further split the sections.

Finally, there is a hook to override the creation of the Document  used to further split each section. This allows the caller to add of more metadata for each section.

## General checklist

- [C] The tests cover both positive and negative cases

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/data/document/splitter/MarkdownSectionSplitter.java
+++ b/langchain4j/src/main/java/dev/langchain4j/data/document/splitter/MarkdownSectionSplitter.java
@@ -1,0 +1,610 @@
+package dev.langchain4j.data.document.splitter;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import dev.langchain4j.data.document.Document;
+import dev.langchain4j.data.document.DocumentSplitter;
+import dev.langchain4j.data.document.Metadata;
+import dev.langchain4j.data.segment.TextSegment;
+import dev.langchain4j.internal.ValidationUtils;
+import org.commonmark.Extension;
+import org.commonmark.ext.front.matter.YamlFrontMatterExtension;
+import org.commonmark.ext.front.matter.YamlFrontMatterVisitor;
+import org.commonmark.ext.gfm.tables.TablesExtension;
+import org.commonmark.node.FencedCodeBlock;
+import org.commonmark.node.Heading;
+import org.commonmark.node.Image;
+import org.commonmark.node.IndentedCodeBlock;
+import org.commonmark.node.Link;
+import org.commonmark.node.Node;
+import org.commonmark.node.Text;
+import org.commonmark.parser.Parser;
+import org.commonmark.renderer.NodeRenderer;
+import org.commonmark.renderer.markdown.CoreMarkdownNodeRenderer;
+import org.commonmark.renderer.markdown.MarkdownNodeRendererContext;
+import org.commonmark.renderer.markdown.MarkdownNodeRendererFactory;
+import org.commonmark.renderer.markdown.MarkdownRenderer;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.ListIterator;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * A {@link DocumentSplitter} that takes a Markdown as input.
+ *
+ * <p>The class is instantiated via the {@link Builder} returned from the {@link #builder()} method.</p>
+ *
+ * <p>Internally it splits the document into sections, with metadata entries to identify the location of each section
+ * in the document. It then optionally splits each section with the {@link DocumentSplitter} passed in to the Builder.
+ *
+ */
+public class MarkdownSectionSplitter implements DocumentSplitter {
+
+    private static final Header NO_HEADER = new Header(null, 1);
+    private static final Set<Extension> EXTENSIONS = Set.of(TablesExtension.create());
+
+
+    public static final String SECTION_LEVEL = "md_section_level";
+    public static final String SECTION_HEADER = "md_section_header";
+    public static final String SECTION_INDEX_WITHIN_PARENT = "md_section_index_in_parent";
+    public static final String SECTION_PARENT_HEADER = "md_parent_header";
+    public static final String SEGMENT_LINKS = "segmen_links";
+    private static final DocumentSplitter NO_SPLIT = document -> Collections.singletonList(document.toTextSegment());
+
+    private final DocumentSplitter sectionSplitter;
+
+    private final String documentTitle;
+
+    private final String emptySectionPlaceholderText;
+
+    private final LinkHandling linkHandling;
+
+    private final DocumentAdjuster documentAdjuster;
+    private final TextSegmentsAdjuster textSegmentsAdjuster;
+    private final YamlFrontMatterConsumer yamlFrontMatterConsumer;
+
+    protected MarkdownSectionSplitter(Builder builder) {
+        ValidationUtils.ensureNotNull(builder.getSectionSplitter(), "sectionSplitter");
+        ValidationUtils.ensureNotNull(builder.getLinkHandling(), "linkHandling");
+        ValidationUtils.ensureNotNull(builder.getSectionSplitter(), "sectionSplitter");
+        ValidationUtils.ensureNotNull(builder.getLinkHandling(), "linkHandling");
+        ValidationUtils.ensureNotNull(builder.getDocumentAdjuster(), "documentAdjuster");
+        ValidationUtils.ensureNotNull(builder.getTextSegmentAdjuster(), "textSegmentAdjuster");
+        this.sectionSplitter = builder.getSectionSplitter();
+        this.documentTitle = builder.getDocumentTitle();
+        this.emptySectionPlaceholderText = builder.getEmptySectionPlaceholderText();
+        this.linkHandling = builder.getLinkHandling();
+        this.documentAdjuster = builder.getDocumentAdjuster();
+        this.textSegmentsAdjuster = builder.getTextSegmentAdjuster();
+        this.yamlFrontMatterConsumer = builder.getYamlFrontMatterConsumer();
+    }
+
+    /**
+     * Creates a new Builder used to instantiate this class.
+     *
+     * @return the builder
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @Override
+    public List<TextSegment> split(Document document) {
+        Set<Extension> parserExtensions = new HashSet<>(EXTENSIONS);
+        if (yamlFrontMatterConsumer != null) {
+            parserExtensions.add(YamlFrontMatterExtension.create());
+        }
+        Node node = Parser.builder().extensions(parserExtensions).build().parse(document.text());
+        if (yamlFrontMatterConsumer != null) {
+            YamlFrontMatterVisitor visitor = new YamlFrontMatterVisitor();
+            node.accept(visitor);
+            yamlFrontMatterConsumer.consumeFrontMatter(visitor.getData());
+        }
+
+        MarkdownSplitterContext context = new MarkdownSplitterContext(document.metadata());
+        MarkdownRenderer renderer = MarkdownRenderer.builder()
+                .nodeRendererFactory(new MarkdownSectionSplitterNodeRendererFactory(context))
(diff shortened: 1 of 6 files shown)


langchain4j/src/test/java/dev/langchain4j/data/document/splitter/MarkdownSplitterTest.java
```diff
@@ -0,0 +1,250 @@
...
+        Assert.assertEquals(12, segments.size());
+        checkTextSegment(source, segments.get(0), "Title", null, 0, 0, ".");
+        checkTextSegment(source, segments.get(1), "Section 1", "Title", 1, 0, "section 1");
+        checkTextSegment(source, segments.get(2), "Section 2", "Title", 1, 1, "section 2");
+        checkTextSegment(source, segments.get(3), "Section 2.1", "Section 2", 2, 0, "section 2.1");
+        checkTextSegment(source, segments.get(4), "Section 2.1.1", "Section 2.1", 3, 0, "section 2.1.1");
+        checkTextSegment(source, segments.get(5), "Section 2.1.2", "Section 2.1", 3, 1, "section 2.1.2\nmore");
+        checkTextSegment(source, segments.get(6), "Section 2.2", "Section 2", 2, 1, ".");
+        checkTextSegment(source, segments.get(7), "Section 2.2.1", "Section 2.2", 3, 0, ".");
+        checkTextSegment(source, segments.get(8), "Section 3", "Title", 1, 2, "section 3");
```
kabir: I have one doubt here. Here the level is taken to be 2, inferred from the number of '#' in the header.

However, in this case there is no level 1 parent since I knowingly 'broke' the structure of the document a bit. So the path to this section is:

\# Title
\### Section 3

elsewhere in the tests I am following the more standard structure, e.g:

\# Title
\## Section 2
\### Section 2.1

So my question is whether 2 for the level here is ok, or if I should use the 'effective' level which is 1.

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/MarkdownSectionSplitter.java
```diff
@@ -0,0 +1,339 @@
...
+ * Then it optionally splits each section with the {@link DocumentSplitter} passed in to the Builder.
+ *
+ */
+public class MarkdownSectionSplitter implements DocumentSplitter {
+
+    public static final String SECTION_LEVEL = "md-section-level";
+    public static final String SECTION_HEADER = "md-section-header";
+    public static final String SECTION_INDEX_WITHIN_PARENT = "md-section-index-in-parent";
+    public static final String SECTION_PARENT_HEADER = "md-parent-header";
+    private static final Pattern HEADER_PATTERN = Pattern.compile("^#+ .*");
```
glaforge: In Markdown, sometimes people also use headers underlined with 3+ `=` signes and 3+ `-` signs.
It's maybe a bit harder to parse with a regex though, but maybe we should not forget about those headers too.

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/MarkdownSectionSplitter.java
```diff
@@ -0,0 +1,339 @@
...
+     * Creates a new Builder used to instantiate this class.
+     *
+     * @return the builder
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @Override
+    public List<TextSegment> split(Document document) {
```
glaforge, line 100: General comment, but the hook could be added here, probably:

There are lots of Markdown documents that contain some "front-matter" preamble, with various metadata information. It's usually written as yaml, between two triple dashes lines.
It would be interesting to have such metadata added to the metadata of the document.

glaforge (commented):
Comment about alternative syntax for headers, and the treatment of front-matter metadata.

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/MarkdownSectionSplitter.java
kabir: Thanks @glaforge 

Do you have some examples of the front-matter preamble please?

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/MarkdownSectionSplitter.java
kabir: I gave this a go. 
3+ = seems to result in h1, no matter how many '=' I put.
3+ - seems to result in h2, no matter how many '-' I put.

Perhaps an option is if I document in the splitter the expectations for the headers, and then they could use a DocumentTransformer before splitting to massage the markdown into the expected format?

Here is my example, in a code block first, and then the same markdown rendered below

```
Normal:
# h1
## h2
### h3
### h4


3+ === seems to result in h1

h1
===
h1
====
h1
=====
h1
======
h1
=======


3+ ---- seems to result in h2

h2
---
h2
----
h2
-----
h2
------
```

Normal:
# h1
## h2
### h3
### h4


3+ === seems to result in h1

h1
===
h1
====
h1
=====
h1
======
h1
=======


3+ ---- seems to result in h2

h2
---
h2
----
h2
-----
h2
------

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/MarkdownSectionSplitter.java
kabir: Never mind, I can see how it works.
If the first line is minimum three dashes, it is the pre-amble, and is hidden. If I don't close it, everything in the document becomes 'hidden'. 

```
------
hello
---
Start doc
```

If there is any content before the pre-amble it is no longer hidden.

```
<!-- test -->
------
hello
-----
Start doc
```

So it seems I should check if the document starts with a line which only contains  dashes, and there are at least three. And then a similar line further down closes this. 

I'll look into adding something so the user can parse this.

kabir: I see this also has the same effect and hides the 'hello':

```
+++
hello
+++
Start doc
```
I also tried with *, and a few other characters.

Again, I think it is reasonable to specify the format we expect, so the user can fix up any weirdness with a transformer?

glaforge: Here's a [pointer](https://www.markdownguide.org/basic-syntax/#alternate-syntax) that shows the alternative header syntax. Triple stars, or triple dashes, can be used standalone as [horizontal bars](https://www.markdownguide.org/basic-syntax/#horizontal-rules).

For the [front-matter](https://daily-dev-tips.com/posts/what-exactly-is-frontmatter/), I think it's not part of the Markdown but I've seen that frequently used in Markdown files. For example, the static blog generator I use ([Hugo](https://gohugo.io/content-management/front-matter/)) uses that for specifying tags, and other meta attributes for the document.

> Again, I think it is reasonable to specify the format we expect, so the user can fix up any weirdness with a transformer?

Unfortunately, I have to disagree here. Developers will send Markdown coming from various places, from libraries that convert HTML to Markdown, or other kind of documents, and they don't have control over how they write their Markdown. So we can't really mandate the exact format the Markdown should have, and we have to make it work with whatever they send us.

Also, I'm wondering if we shouldn't use a Markdown parsing library that would understand all the subtleties of Markdown, like [flexmark](https://github.com/vsch/flexmark-java) or [commonmark-java](https://github.com/commonmark/commonmark-java). I'm curious if you've looked at those? 

I also had a question (or maybe a feature request) regarding the composability. 
Somehow, the other splitters we have can be composed hierarchically, so I'm wondering if you're considering doing that for your splitter too: ie. I may want to first split by Markdown headers, but then do some sentence splitting for example.

Another aspect is the configuration and depth of parsing: for example, I may want to chunk to get sections under 2000 characters and/or tokens, so if a section is under that size, no need to split at a smaller header level. I don't think it's possible right now to do that, but was curious if you were considering this too.

kabir: I vaguely looked at flexmark and commonmark, but that was for our conversion from html to markdown. Flexmark seems quite unmaintained, and generated very bad markdown from the html, so I ended up using Docling and forgot all about those libraries.. 

But it could be a good idea to use commonmark to do the parsing. I'll investigate that in the project I am working on, which is what inspired contributing the splitter.

Another thing I realised I don't like in this PR is the builder. It would probably be easier to just do:

```
MarkdownSectionSplitter splitter = new MarkdownSectionSplitter();
```
and then if something extra needs setting to just do
```
splitter.setX(x);
```

Regarding the composability question, I don't totally get what you are asking :-) I am passing in another splitter which further splits each section once we've extracted the headers. I am doing that in `testSectionSplitter()`. Do you mean something else?

Regarding the last question, that is something I have not thought about. Our use-case is currently to extract all sub-sections, so that is what this does. So, I don't think that is something I'll add at this stage

glaforge: Regarding builders, yes, if there's just one or two parameters, it can be overkill (and verbose).

For the composability, sorry that I wasn't very clear.
I'm thinking of recursive splitters.
For example, we have `DocumentSplitters.recursive(...)` which does this: it chains different splitters, like paragraph splitter, then if the chunk is still too big, it chunks by sentence, then by words, and characters. So the splitting goes deeper if the split still gives too big chunks.

So instead of manually splitting once with the Markdown header splitter, then splitting again by sentence or something else, you just chain them together.

dliubarskyi:
@kabir thank you! I did not look into the implementation yet, but I see there are many discussions about the formatting and all the variations. Just wondering if we should use some library to parse markdown into AST. It seems there are a few java libs for that. WDYT?

langchain4j/src/main/java/dev/langchain4j/data/document/splitter/MarkdownSectionSplitter.java
dliubarskyi: I see @glaforge already mentioned markdown parsing libraries, I completely agree, we should use them if possible. 

In general, it would be very helpful to have one or few examples of markdown files with multiple sections and subsections and demonstrate how you expect it to be split (given a few `maxSegmentSize` settings).

dliubarskyi: Regarding the builder: we keep our objects immutable unless there is a good reason not to. So builder is more preferable than setters

kabir:
Thanks, I'll try to look into all this, although it might take some time!

kabir:
Hi @glaforge and @dliubarskyi I have updated the pull request to use Commonmark to do the parsing.

Originally, I tried to implement my own visitor, but ran into lots of problems trying to output the original markdown from the parts I am not doing any special handling for. So I found the rendererer I am using now, which made it a lot easier. Unfortunately, the renderer has a amall bug which I am trying to fix in https://github.com/commonmark/commonmark-java/pull/361. I've not heard back on the pull request, so I have added a `TempMarkdownRenderer` which includes the fix, and otherwise delegates to Commonmark as usual.

This splitter now supports our use-case. Assuming, what I have put in is ok, may I suggest that for any improvements you do this as follow-up work?

@glaforge suggested something about chained splitters, which I **think** I am doing, at least for what we want to achieve. I feel he has something different in mind, which isn't what we want to do. But I may well be misunderstanding something.

minglu7:
How is it going now

fabriziofortino:
@kabir any update on this PR? Happy to help!

kabir:
@fabriziofortino Hi, my focus has shifted elsewhere since February. Feel free to take it over, I still think it can be useful to others :-)

PR #2422, opened 2025-01-15
Remove observable models and squash observability into chat models

mariofusco:
This is the interfaces squashing that I suggested here https://github.com/langchain4j/langchain4j/pull/2403#discussion_r1914648102

If we decide to bring in this change in I expect that, after this transitionary phase, the `doChat` method in the `ChatLanguageModel` interface will become an abstract one so that we will force all implementors to provide their own implementation.


dliubarskyi (approved):
@mariofusco makes sense, thank you!

PR #2424, opened 2025-01-16
Improvements for error handling with MCP

### LINKED ISSUE #2421: [BUG] MCP response JSON parsing issues for errors
When MCP servers return errors, like 'invalid_request', `DefaultMcpClient` will try to parse this as JSON and throw an error.
Since the LLM makes the request, invalid requests to the MCP server are bound to happen often, would it make sense to add some resiliency?
For example, if it cannot be processed as JSON, wrap whatever message that was returned and send it back to the model? Then it can make a new attempt based on the error message.

Can be reproduced by prompting the bot to pass something outrageous instead of a clean file path as argument in `[McpToolsExampleOverStdio.java](https://github.com/langchain4j/langchain4j-examples/tree/main/mcp-example/src/main/java/dev/langchain4j/example/mcp)` (but it also happens in normal circumstances, just not reliably).

jmartisk:
Fixes https://github.com/langchain4j/langchain4j/issues/2421

Test coverage for this is coming soon, as I'm currently revamping the whole MCP test suite


dliubarskyi (approved):
@jmartisk thank you!

kpavlov:
@jmartisk, I assume the tests will arrive in the follow-up PRs ;)

jmartisk:
@kpavlov, they're in https://github.com/langchain4j/langchain4j/pull/2425

PR #2430, opened 2025-01-20
Post code suggestions for Spotless violations

timtebeek:
## Change
This extends the work done in
- https://github.com/langchain4j/langchain4j/pull/2153

To integrate it with the work done in 
- https://github.com/langchain4j/langchain4j/pull/673

This way, any folks that might forget to apply spotless on their PRs, should see code suggestion comments pop up with the fixes.

PR #2437, opened 2025-01-22
Apply AssertJ best practices missed since October

timtebeek:
## Change
This follows on from
- https://github.com/langchain4j/langchain4j/pull/2430
- https://github.com/langchain4j/langchain4j/pull/673

The integration had not been running, and as such some small improvements were not pointed out on PRs. This fixes that.

### DIFF
--- a/.github/workflows/receive-pr.yml
+++ b/.github/workflows/receive-pr.yml
@@ -43,9 +43,7 @@ jobs:
       # Execute recipes
       - name: Apply OpenRewrite recipes
         run: |
-          mvn --activate-profiles openrewrite \
-            -Dspotless.check.skip=true \
-            org.openrewrite.maven:rewrite-maven-plugin:run
+          mvn --activate-profiles openrewrite -Dspotless.check.skip=true rewrite:run
 
       # Apply spotless
       - name: Apply Spotless
--- a/pom.xml
+++ b/pom.xml
@@ -191,13 +191,13 @@
         </profile>
         <profile>
             <id>openrewrite</id>
-            <!-- `mvn -P openrewrite org.openrewrite.maven:rewrite-maven-plugin:run` -->
+            <!-- `mvn -P openrewrite rewrite:run` -->
             <build>
                 <plugins>
                     <plugin>
                         <groupId>org.openrewrite.maven</groupId>
                         <artifactId>rewrite-maven-plugin</artifactId>
-                        <version>5.42.2</version>
+                        <version>5.43.4</version>
                         <configuration>
                             <activeRecipes>
                                 <!--<recipe>org.openrewrite.java.OrderImports</recipe>-->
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -191,13 +191,13 @@
...
             <id>openrewrite</id>
-            <!-- `mvn -P openrewrite org.openrewrite.maven:rewrite-maven-plugin:run` -->
+            <!-- `mvn -P openrewrite rewrite:run` -->
             <build>
                 <plugins>
                     <plugin>
                         <groupId>org.openrewrite.maven</groupId>
                         <artifactId>rewrite-maven-plugin</artifactId>
-                        <version>5.42.2</version>
+                        <version>5.43.4</version>
```
timtebeek, line 200: For now I've intentionally not bumped this higher yet, as that might also pull in a newer version of `rewrite-testing-frameworks`, which [as of v2.21.0 also includes JUnit 5 best practices](https://github.com/openrewrite/rewrite-testing-frameworks/commit/8fb11c8f0f116a87883fc30008b008eeba6568cc), which gives a much larger diff that is perhaps best agreed upon first.

[JUnit 5 best practices](https://docs.openrewrite.org/recipes/java/testing/junit5/junit5bestpractices) includes for instance 
- [Remove test prefix from JUnit 5 tests](https://docs.openrewrite.org/recipes/java/testing/cleanup/removetestprefix)
- [Remove public visibility of JUnit 5 tests](https://docs.openrewrite.org/recipes/java/testing/cleanup/testsshouldnotbepublic)

We feel those are helpful to apply, but do let me know if you'd rather not see that change in a subsequent (or this) pull request.

.github/workflows/receive-pr.yml
```diff
@@ -43,9 +43,7 @@ jobs:
       # Execute recipes
       - name: Apply OpenRewrite recipes
         run: |
-          mvn --activate-profiles openrewrite \
-            -Dspotless.check.skip=true \
-            org.openrewrite.maven:rewrite-maven-plugin:run
+          mvn --activate-profiles openrewrite -Dspotless.check.skip=true rewrite:run
```
kpavlov, line 46: 🙏🏻

kpavlov (approved):
Thank you, @timtebeek

pom.xml
dliubarskyi: Those 2 best practices seem reasonable

dliubarskyi (approved):
@timtebeek thank you!

pom.xml
timtebeek: Thanks! I've applied that on this preview PR; let me know if that's to your liking and I can open it here instead:
- https://github.com/timtebeek/langchain4j/pull/1

The diff ends up a little larger than I'd have liked due to the spotless ratchet, but at least that should only be once.

dliubarskyi: Thank you! There are many PRs in the pipeline for the release next week. Applying this change now would probably mean lots of conflicts, so I would postpone it till the week after. 🙏

timtebeek: All good! I'd indeed intentionally captured the commands I used to create those changes; I'll keep an eye out for the release and apply again after. Thanks for the quick feedback!

PR #2443, opened 2025-01-23
Replace properties() with addProperties in the JsonObjectSchema builder

### LINKED ISSUE #2419: [BUG] JsonObjectSchema.Builder: Properties Added via addStringProperty() Are Overwritten by properties() Method
**Describe the bug**
Properties added to JsonObjectSchema.Builder using `addStringProperty()`, `addNumberProperty()`, etc. are lost when `properties()` is subsequently called. This makes it impossible to combine direct property addition with nested properties, as the nested properties must be added via the properties map.

**Log and Stack trace**
No exception is thrown, but properties silently disappear. This can be demonstrated with the following test output:
```
Expected: Map containing "field1"
     but: was Map{nested=JsonObjectSchema{...}}
```

**To Reproduce**
```java
@Test
void demonstratePropertyLoss() {
    JsonSchema schema = JsonSchema.builder()
        .name("Example")
        .rootElement(JsonObjectSchema.builder()
            // This property will be lost
            .addStringProperty("field1", "First field")
            .properties(Map.of(
                "nested", JsonObjectSchema.builder()
                    .addStringProperty("subfield", "Nested field")
                    .build()
            ))
            .build())
        .build();

    JsonObjectSchema rootElement = (JsonObjectSchema) schema.rootElement();
    Map<String, JsonSchemaElement> properties = rootElement.properties();
    
    // field1 is not present in the final schema
    assert !properties.containsKey("field1");
    // only the nested property remains
    assert properties.containsKey("nested");
}
```

**Expected behavior**
Properties added via `addStringProperty()` and similar methods should be preserved when `properties()` is called. The final schema should contain both directly added properties and those provided via the properties map.

Expected properties map should contain:
```java
{
    "field1": JsonStringSchema{...},
    "nested": JsonObjectSchema{...}
}
```

- LangChain4j version: 0.36.2
- LLM(s) used: OpenAI GPT-4
- Java version: 21
- Spring Boot version: 3.3.0

**Additional context**
This issue appears to be caused by the `properties()` method overwriting the internal properties map rather than merging with existing properties. The current workaround is to collect all properties in a mutable map first:

```java
Map<String, JsonSchemaElement> properties = new HashMap<>();

// Add direct property
properties.put("field1", JsonStringSchema.builder()
    .description("First field")
    .build());

// Add nested property
properties.put("nested", JsonObjectSchema.builder()
    .addStringProperty("subfield", "Nested field")
    .build());

// Set all properties at once
JsonSchema schema = JsonSchema.builder()
    .name("Example")
    .rootElement(JsonObjectSchema.builder()
        .properties(properties)
        .build())
    .build();
```

This workaround is functional but less intuitive than the builder pattern's usual fluent interface and requires maintaining a separate mutable map.

mariofusco:
Fix https://github.com/langchain4j/langchain4j/issues/2419

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonObjectSchema.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonObjectSchema.java
@@ -70,7 +70,7 @@ public class JsonObjectSchema implements JsonSchemaElement {
         }
 
         /**
-         * Sets the properties of this JSON object.
+         * Adds all properties in the parameter Map to the properties of this JSON object.
          * Please note that {@link #required(List)} should be set explicitly if you want the properties to be mandatory.
          *
          * @see #addProperty(String, JsonSchemaElement)
@@ -85,8 +85,8 @@ public class JsonObjectSchema implements JsonSchemaElement {
          * @see #addEnumProperty(String, List)
          * @see #addEnumProperty(String, List, String)
          */
-        public Builder properties(Map<String, JsonSchemaElement> properties) {
-            this.properties = properties;
+        public Builder addProperties(Map<String, JsonSchemaElement> properties) {
+            this.properties.putAll(properties);
             return this;
         }
 
@@ -94,7 +94,7 @@ public class JsonObjectSchema implements JsonSchemaElement {
          * Adds a single property to the properties of this JSON object.
          * Please note that {@link #required(List)} should be set explicitly if you want the properties to be mandatory.
          *
-         * @see #properties(Map)
+         * @see #addProperties(Map)
          * @see #addStringProperty(String)
          * @see #addStringProperty(String, String)
          * @see #addIntegerProperty(String)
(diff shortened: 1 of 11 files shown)


langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonObjectSchema.java
```diff
@@ -85,16 +85,16 @@ public Builder description(String description) {
          * @see #addEnumProperty(String, List)
          * @see #addEnumProperty(String, List, String)
          */
-        public Builder properties(Map<String, JsonSchemaElement> properties) {
```
dliubarskyi, line 88: Can we make it in a non-breaking way? We can keep the old `properties()` method that is currently used by some users and mark it `@Deprecated`

dliubarskyi (commented):
@mariofusco thank you!

langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonObjectSchema.java
mariofusco: Done.

I was tempted to make it call the new `addProperties()` method, but since we are keeping this only for backward compatibility and will remove it soon I though to also leave it with the old behaviour.

dliubarskyi (approved):
@mariofusco thanks a lot!

PR #2485, opened 2025-01-31
Rewrite the PgVectorEmbeddingStore search query to enable index

### LINKED ISSUE #2484: [BUG] pgvector embedding store doesn't apply index for search
## Problem 
The search SELECT query in the `PgVectorEmbeddingStore` implementation is written that way that Postgres doesn't apply the IVFflat index and does sequential scan, which will cause performance degradation in big tables.

## Details
Here is the search query example from the [PgVectorEmbeddingStore.search](https://github.com/langchain4j/langchain4j/blob/48681d97b50861f02b2364b0057792cacc42e953/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java#L287) method:

```
WITH temp AS (
    SELECT (2 - (embedding <=> '[...]')) / 2 AS score, embedding_id, embedding, text, test 
    FROM document_embeddings
)
SELECT * FROM temp 
WHERE score >= 0.5 ORDER BY score desc LIMIT 10;
```
The IVFflat index for this table exists.
Now, if we run the execution plan for this query, it gives the following result:

```
EXPLAIN ANALYSE WITH temp AS (
    SELECT (2 - (embedding <=> '[...]')) / 2 AS score, embedding_id, embedding, text, test FROM document_embeddings
) SELECT * FROM temp WHERE score >= 0.5 ORDER BY score desc LIMIT 10;
```
The query plan output:
```
Limit  (cost=1548.49..1548.51 rows=10 width=55) (actual time=348.879..348.883 rows=10 loops=1)
  ->  Sort  (cost=1548.49..1581.44 rows=13180 width=55) (actual time=348.878..348.880 rows=10 loops=1)
"        Sort Key: ((('2'::double precision - (document_embeddings.embedding <=> '[...]'::vector)) / '2'::double precision)) DESC"
        Sort Method: top-N heapsort  Memory: 27kB
        ->  Seq Scan on document_embeddings  (cost=0.00..1263.67 rows=13180 width=55) (actual time=0.081..340.563 rows=39541 loops=1)
"              Filter: ((('2'::double precision - (embedding <=> '[...]'::vector)) / '2'::double precision) >= '0.5'::double precision)"
Planning Time: 0.456 ms
Execution Time: 348.937 ms
```
**The output shows 'Seq Scan', the index is not applied.**

## Explanation

The pgvector [documentation says](https://github.com/pgvector/pgvector?tab=readme-ov-file#why-isnt-a-query-using-an-index):
> The query needs to have an ORDER BY and LIMIT, and the ORDER BY must be the **result of a distance operator (not an expression)** in **ascending** order.

whereas our query violates both of these rules: the distance operator is an expression and the order is descending.

## Solution

According to the documentation mentioned above, the query should be rewritten to use distance operator in the ORDER BY directly and use the ascending order. 
Here is the expected query:

```
SELECT (embedding <=> '[...]') AS score, embedding_id, embedding, text
FROM document_embeddings
WHERE (embedding <=> '[...]') <= 0.5
ORDER BY embedding <=> '[...]' LIMIT 10;
```
And here is the execution plan:
```
Limit  (cost=411.07..412.29 rows=10 width=47) (actual time=5.074..5.183 rows=10 loops=1)
  ->  Index Scan using document_embeddings_ivfflat_index on document_embeddings  (cost=411.07..2014.16 rows=13180 width=47) (actual time=5.073..5.180 rows=10 loops=1)
"        Order By: (embedding <=> '[...]'::vector)"
"        Filter: ((embedding <=> '[...]'::vector) <= '0.5'::double precision)"
Planning Time: 0.095 ms
Execution Time: 5.225 ms
```
Now it's index scan with our ivfflat index. The costs and execution time are reduced a lot.

zmaks:
Closes #2484

## Change
Now the query looks like this and it corresponds the pgvector documentation so it uses the index (not just seq scan) which will improve the performance for large tables.

```
SELECT (embedding <=> '[...]') AS score, embedding_id, embedding, text
FROM document_embeddings
WHERE (embedding <=> '[...]') <= 0.5
ORDER BY embedding <=> '[...]' LIMIT 10;
```

And here is the execution plan:
```
Limit  (cost=411.07..412.29 rows=10 width=47) (actual time=5.074..5.183 rows=10 loops=1)
  ->  Index Scan using document_embeddings_ivfflat_index on document_embeddings  (cost=411.07..2014.16 rows=13180 width=47) (actual time=5.073..5.180 rows=10 loops=1)
"        Order By: (embedding <=> '[...]'::vector)"
"        Filter: ((embedding <=> '[...]'::vector) <= '0.5'::double precision)"
Planning Time: 0.095 ms
Execution Time: 5.225 ms
```

Note: I haven't added new tests since I didn't add or remove any functionality. This PR is just a search performance fix.

## General checklist


dliubarskyi:
Hi @zmaks, thanks a lot!

There are 5 failing tests, could you please check?

image

All of them have the same error message:

> Expected size: 2 but was: 1 in:
> [EmbeddingMatch ...

zmaks:
Hi @dliubarskyi, you're welcome!
I've updated the branch, run the tests again and all of them are green... So I don't see these failures and can't investigate them. What am I missing?
image

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -276,22 +276,24 @@ public void removeAll() {
     public EmbeddingSearchResult<TextSegment> search(EmbeddingSearchRequest request) {
         Embedding referenceEmbedding = request.queryEmbedding();
         int maxResults = request.maxResults();
-        double minScore = request.minScore();
+        double maxScore = 2 - 2 * request.minScore();
```
```suggestion
        double maxCosineDistance = 2 - 2 * request.minScore();
```

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -276,22 +276,24 @@ public void removeAll() {
...
         try (Connection connection = getConnection()) {
             String referenceVector = Arrays.toString(referenceEmbedding.vector());
             String whereClause = (filter == null) ? "" : metadataHandler.whereClause(filter);
-            whereClause = (whereClause.isEmpty()) ? "" : "WHERE " + whereClause;
+            whereClause = (whereClause.isEmpty()) ? "" : "AND " + whereClause;
             String query = String.format(
-                    "WITH temp AS (SELECT (2 - (embedding <=> '%s')) / 2 AS score, embedding_id, embedding, text, " +
-                            "%s FROM %s %s) SELECT * FROM temp WHERE score >= %s ORDER BY score desc LIMIT %s;",
-                    referenceVector, join(",", metadataHandler.columnsNames()), table, whereClause, minScore, maxResults);
+                    "SELECT (embedding <=> '%s') AS score, embedding_id, embedding, text, %s FROM %s " +
```
```suggestion
                    "SELECT (embedding <=> '%s') AS cosine_distance, embedding_id, embedding, text, %s FROM %s " +
```

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -276,22 +276,24 @@ public void removeAll() {
...
+                    "WHERE (embedding <=> '%s') <= %s %s ORDER BY embedding <=> '%s' LIMIT %s;",
+                    referenceVector, join(",", metadataHandler.columnsNames()), table, referenceVector, maxScore,
+                    whereClause, referenceVector, maxResults
+            );
             try (PreparedStatement selectStmt = connection.prepareStatement(query)) {
                 try (ResultSet resultSet = selectStmt.executeQuery()) {
                     while (resultSet.next()) {
-                        double score = resultSet.getDouble("score");
+                        double score = (2 - resultSet.getDouble("score")) / 2;
```
```suggestion
                        double score = (2 - resultSet.getDouble("cosine_distance")) / 2;
```

dliubarskyi:
@zmaks strange, I am not sure what is wrong. I have explicitly checked out your latest commit and I am still getting test failiures.
Then I checked out `main` and there are no test failures.

image

zmaks:
I assume the problem was in score calculation on the jvm side which might loose the precision due to `double` rounding. This reproduces specifically for the case where you try to find a record by the cosine distance equality, `<=` or `>=` - that's what the failing test checked.
I moved these calculations to the Postgres side and add rounding to the filtering part so it won't take into account more than 8 digits after the dot. @dliubarskyi could you please try if now the tests pass?

dliubarskyi:
@zmaks thanks a lot, the tests are passing now! Does this new query still uses the ivfflat index?

zmaks:
> Does this new query still uses the ivfflat index?

@dliubarskyi Yes. The key thing is ascending ordering by cosine distance directly without any calculations and transformations. The ivfflat/hnsw index won't be applied to the filtering by `where` clause anyhow, not matter if we compare cosine distance or some calculated from the distance value.

Here is the new execution plan, the index is applied:
```
Limit  (cost=411.07..412.39 rows=10 width=55) (actual time=4.881..5.037 rows=10 loops=1)
  ->  Index Scan using document_embeddings_ivfflat_index on document_embeddings  (cost=411.07..2150.09 rows=13180 width=55) (actual time=4.880..5.034 rows=10 loops=1)
"        Order By: (embedding <=> '[...]'::vector)"
"        Filter: ((text ~~ 't%'::text) AND (round(((embedding <=> '[...]'::vector))::numeric, 8) <= 1))"
Planning Time: 0.095 ms
Execution Time: 5.202 ms
```

dliubarskyi (approved):
@zmaks thanks a lot, great job!

PR #2501, opened 2025-02-04
#2471 Added explicit setting of socket timeout to bedrock http client…

### LINKED ISSUE #2471: [BUG] Socket timeout is not configured in Bedrock client init
when sending a large amount of data to Bedrock AI models the request times out in ~2 min regardless of how big timeout is specified in Bedrock model. The reason is described in this article: https://repost.aws/articles/AR_A_6aqweQYOR5FCgvM2vdA/solving-read-timed-out-error-and-high-latencies-in-amazon-bedrock-with-aws-java-sdk-client
In the current langchain version (0.36) that I am using, the method `BedrockRuntimeClient getClient()` is public so I can override it and configure the socket timeout correctly, but a) this override is somewhat tricky (and hardly recommended) and b) it may change in the future versions.
A solution is to add http client socket timeout configuration in the initiation of the Bedrock client, as it is suggested by AWS.

sashaepstein:
… to override default 30 sec timeout

## Issue
Fixes  #2471 

## Change
Set socket timeout on Apache HTTP client to be either the same as the specified timeout on the Bedrock model, or the default suggested by AWS in https://repost.aws/articles/AR_A_6aqweQYOR5FCgvM2vdA/solving-read-timed-out-error-and-high-latencies-in-amazon-bedrock-with-aws-java-sdk-client

## General checklist
- [X ] There are no breaking changes
- The existing tests should suffice as there is no functional changes.
- [X ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- I tested this fix with my company product that uses langchain for both, Bedrock and OpenAI models.

There is no changes in the use of Bedrock model.

### DIFF
--- a/langchain4j-bedrock/pom.xml
+++ b/langchain4j-bedrock/pom.xml
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>dev.langchain4j</groupId>
@@ -110,15 +108,10 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>apache-client</artifactId>
+        </dependency>
     </dependencies>
-
-    <licenses>
-        <license>
-            <name>Apache-2.0</name>
-            <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
-            <distribution>repo</distribution>
-            <comments>A business-friendly OSS license</comments>
-        </license>
-    </licenses>
 
 </project>
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/internal/AbstractBedrockChatModel.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/internal/AbstractBedrockChatModel.java
@@ -11,6 +11,7 @@ import dev.langchain4j.model.chat.listener.ChatModelResponse;
 import dev.langchain4j.model.chat.listener.ChatModelResponseContext;
 import dev.langchain4j.model.output.Response;
 import java.nio.charset.Charset;
+import java.time.Duration;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -20,6 +21,7 @@ import lombok.Getter;
 import lombok.experimental.SuperBuilder;
 import lombok.extern.slf4j.Slf4j;
 import software.amazon.awssdk.core.SdkBytes;
+import software.amazon.awssdk.http.apache.ApacheHttpClient;
 import software.amazon.awssdk.services.bedrockruntime.BedrockRuntimeClient;
 import software.amazon.awssdk.services.bedrockruntime.model.InvokeModelRequest;
 import software.amazon.awssdk.services.bedrockruntime.model.InvokeModelResponse;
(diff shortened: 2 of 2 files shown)


langchain4j-bedrock/pom.xml
```diff
@@ -110,15 +108,10 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>apache-client</artifactId>
```
dliubarskyi, line 113: AWS SDK seems to have an SPI for HTTP clients, and supports multiple HTTP client implementations. I am not sure it is correct to import a specific client by default and explicitly use it as this will prohibit using other clients

dliubarskyi (commented):
@sashaepstein thank you!

langchain4j-bedrock/pom.xml
sashaepstein: Well, this is what AWS recommended in that article. I had a live meeting with their solution engineer, and he pointed me to that solution.
Not sure I understand your concern. To use another http client you will need to configure it in the same method `BedrockRuntimeClient initClient()` and use a corresponding HTTP client builder. The existing implementation of that method just uses the default HTPP client, and the suggested change overrides the default with Apache's one.

dliubarskyi: My concern is that now this module is tightly coupled to the `apache-client` dependency, and users will not be able to exclude this dependency if they use another http client because of `import software.amazon.awssdk.http.apache.ApacheHttpClient;`

sashaepstein: Sorry, I guess I am missing something here. The only way I can see for users to switch to a different HTTP client is to override `BedrockRuntimeClient initClient()` in internal langchain class, which is, as I mentioned before, is not simple and not recommended. However, even if they decide to do that, how the dependency on Apache client and the import in the internal langhchain class can hurt them, except bringing Apache client in the classpath? Can you clarify it for me please?

dliubarskyi: Hey, sorry for the delayed response!

> is not simple and not recommended

Why is it not simple and not recommended?

> except bringing Apache client in the classpath

That is exactly the issue - many people care what is in their classpath for various reasons

sashaepstein: > Why is it not simple and not recommended?

AbstractBedrockChatModel is an internal class in the class hierarchy of BedrockAnthropicMessageChatModel class that you would need to instantiate to use Bedrock model. Hence to override AbstractBedrockChatModel.initClient() you would have to override BedrockAnthropicMessageChatModel class. It just took me some time to figure out that all I need to do is to override the BedrockAnthropicMessageChatModel internal constructor that is not directly exposed in the class source but generated at the build time. 
If providing our own http client for Bedrock SDK was thought to be an option in the langchain Bedrock implementation, then it would have to be done in a much more explicit manner, rather than being hidden in the internal class, and would have to be documented correspondingly.

> That is exactly the issue

I don't see the issue. ApacheHttpClient is the default http client that AWS SDK uses, and correspondingly the client that langchain indirectly deploys, hence even without the suggested (by AWS) override of that client's default configuration, the AWS SDK's implementation of ApacheHttpClient is already in the langchain  classpath.

Finally, maybe I was not explicit enough in the beginning: the current implementation of lancgchain Bedrock model doesn't work reliably when the data sent to AI contains up to ~ 30-35% of the max tokens supported by Claude Sonnet 3.5 model, and doesn't work at all if the data sent to AI contains more than 40% of the max tokens. The reason is the 30 sec default socket timeout in the AWS SDK's implementation of ApacheHttpClient, which is less than the time the Claude Sonnet 3.5 model takes to process data with the size of 35% + of the max tokens. As the result of the socket timeout, the AWS SDK BedrockRuntimeClient retries the request just to get it timeout again and again.

dliubarskyi: Hey, sorry for the late reply.

> AbstractBedrockChatModel is an internal class in the class hierarchy of BedrockAnthropicMessageChatModel class that you would need to instantiate to use Bedrock model. Hence to override AbstractBedrockChatModel.initClient() you would have to override BedrockAnthropicMessageChatModel class. It just took me some time to figure out that all I need to do is to override the BedrockAnthropicMessageChatModel internal constructor that is not directly exposed in the class source but generated at the build time.
> If providing our own http client for Bedrock SDK was thought to be an option in the langchain Bedrock implementation, then it would have to be done in a much more explicit manner, rather than being hidden in the internal class, and would have to be documented correspondingly.

My assumption is that user should be able to override/specify underlying client. If it is not possible, then it is a problem we need to fix. But it does not seems to be the case as you can do: `BedrockAnthropicMessageChatModel.builder().client(...).build()`. Yes, documentation should be improved for sure.

> I don't see the issue. ApacheHttpClient is the default http client that AWS SDK uses, and correspondingly the client that langchain indirectly deploys, hence even without the suggested (by AWS) override of that client's default configuration, the AWS SDK's implementation of ApacheHttpClient is already in the langchain classpath.

It is in the classpath, but it can be excluded from it and overriden with another client. But when there is an explicit reference to `ApacheHttpClient`, it cannot be excluded.

> Finally, maybe I was not explicit enough in the beginning: the current implementation of lancgchain Bedrock model doesn't work reliably when the data sent to AI contains up to ~ 30-35% of the max tokens supported by Claude Sonnet 3.5 model, and doesn't work at all if the data sent to AI contains more than 40% of the max tokens. The reason is the 30 sec default socket timeout in the AWS SDK's implementation of ApacheHttpClient, which is less than the time the Claude Sonnet 3.5 model takes to process data with the size of 35% + of the max tokens. As the result of the socket timeout, the AWS SDK BedrockRuntimeClient retries the request just to get it timeout again and again

I have [deprecated](https://github.com/langchain4j/langchain4j/pull/2675) models that use old "invoke" API. Now there is `BedrockChatModel` which uses new "converse" API. Could you please check if the problem still there?

sashaepstein: With the addition of a new BedrockChatModel, it became easy to supply your own Bedrock Client via the model builder. However, the default Bedrock Client that langchain supplies, still suffers from the short socket timeout in the Bedrock SDK default http client, as I explained earlier.

PR #2503, opened 2025-02-04
add setMetadata into AzureAiSearchContentRetriever.add(List<TextSegment>) function ([BUG] Metadata loss with AzureAiSearchContentRetriever.add(Document) (#2502)

### LINKED ISSUE #2502: [BUG] Metadata loss with AzureAiSearchContentRetriever.add(Document)
**Describe the bug**

After creating a langchain4j Document with its Metadata, I use the "add" method of the AzureAiSearchContentRetriever class, and then I can't see the Metadata I expect when I search for that document in Azure, resulting in metadata: null.
Searching the source code of langchain4j, I noticed that firstly the Document is converted in a SingletonList of TextSegment, and its metadata is properly copied in the TextSegment, but then the AzureAiSearchContentRetriever.add(List<TextSegment>) copies only the id and the content, it does not set the metadata when it's creating the List<Document>.
I've not found a mention of this problem yet, please let me know if I'm missing something.

**Log and Stack trace**

There's no log or stack trace for this bug.
**To Reproduce**

You need an Azure AI Search instance from Microsoft Azure, then set up a Easy Rag implementation from the langchain4j [guide](https://docs.langchain4j.dev/tutorials/rag/#easy-rag) with the difference that you need to set a AzureAiSearchContentRetriever as the Assistant's ContentRetriever.
Anyway, the assistant is not needed to reproduce this, this happens during the ingestion of any Document  with the add method of your AzureAiSearchContentRetriever instance.

**Expected behavior**

I would have expected the set of metadata alongside its content.

- LangChain4j version: 1.0.0-alpha1
- LLM(s) used: text-embedding-3-large
- Java version: 21
- Spring Boot version (if applicable): 3.2.0

**Additional context**

The document retrieval from the Azure Search Explorer inside the Indexes tab:
Image

The screenshot of the AzureAiSearchContentRetriever implementation, I think that before line 154 a document.setMetadata(segment.metadata()) is missing:

Image

JJonahJson:
## Issue

Closes #2502

## Change

Add document.setMetadata to AzureAiSearchContentRetriever.add(List<TextSegment>) to avoid metadata loss

## General checklist


langchain4j-azure-ai-search/src/main/java/dev/langchain4j/rag/content/retriever/azure/search/AzureAiSearchContentRetriever.java
```diff
@@ -151,6 +151,7 @@ public void add(List<TextSegment> segments) {
             Document document = new Document();
             document.setId(randomUUID());
             document.setContent(segment.text());
+            document.setMetadata(segment.metadata());
```
dliubarskyi: this does not compile

dliubarskyi (commented):
@JJonahJson thank you! It would be great to add at least a single test that verified that metadata is persisted and retrieved

JJonahJson:
I didn't realize that TextSegment's Metadata and Document's Metadata where two different classes, I'll fix that and add a test

JJonahJson:
I believe everything is now in order.
I've introduced a new setAttributes method that could potentially be used in the AbstractAzureAiSearchEmbeddingStore.addAll() function, I've noticed there's a piece of code that manages the conversion of Metadata, but since it's already functioning well, I thought not to make any additional changes to it.
@dliubarskyi, assuming everything is working correctly, do you have any idea of when we might expect a new version to be released?

dliubarskyi (approved):
@JJonahJson thanks a lot!

PR #2535, opened 2025-02-11
Introduce pluggable invocation policy

mariofusco:
This pull request allows to optionally configure a `ChatLanguageModel` with an `InvocationPolicy`. Both the `RetryPolicy` and the `ExceptionMapper` (used to map HttpException to specific exeception based on the http code, still to be implemented) are  `InvocationPolicy`. `InvocationPolicy` are composable, through decoration, so an `InvocationPolicy` can define both the retry policy and the exception mapping, plus other strategies that we may want to add in future.

@dliubars Please give a look and let me know if the idea makes sense for you. If so I will also implement a proper execption mapping based on http error codes as discussed.

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/chat/policy/ExceptionMapper.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/chat/policy/ExceptionMapper.java
@@ -1,0 +1,29 @@
+package dev.langchain4j.model.chat.policy;
+
+import java.util.concurrent.Callable;
+
+public interface ExceptionMapper extends InvocationPolicy {
+
+    ExceptionMapper DEFAULT = new DefaultExceptionMapper();
+
+    @Override
+    default Callable<?> apply(final Callable<?> action) {
+        return () -> {
+            try {
+                return action.call();
+            } catch (Exception e) {
+                throw mapException(e);
+            }
+        };
+    }
+
+    RuntimeException mapException(Exception e);
+
+    class DefaultExceptionMapper implements ExceptionMapper {
+        @Override
+        public RuntimeException mapException(Exception e) {
+            System.out.println(e.getMessage());
+            return e instanceof RuntimeException re ? re : new RuntimeException(e);
+        }
+    }
+}
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
@@ -3,8 +3,10 @@ package dev.langchain4j.model.openai;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
-import dev.langchain4j.exception.HttpException;
 import dev.langchain4j.http.client.HttpClientBuilder;
+import dev.langchain4j.model.chat.policy.ExceptionMapper;
+import dev.langchain4j.model.chat.policy.InvocationPolicy;
+import dev.langchain4j.model.chat.policy.RetryUtils;
 import dev.langchain4j.model.Tokenizer;
 import dev.langchain4j.model.chat.Capability;
 import dev.langchain4j.model.chat.ChatLanguageModel;
@@ -26,8 +28,11 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.StringJoiner;
 
-import static dev.langchain4j.internal.RetryUtils.withRetry;
+import static dev.langchain4j.model.chat.policy.PolicyUtil.invokePolicy;
+import static dev.langchain4j.model.chat.policy.RetryUtils.DEFAULT_RETRY_POLICY;
+import static dev.langchain4j.model.chat.policy.RetryUtils.retryPolicyBuilder;
 import static dev.langchain4j.internal.Utils.copyIfNotNull;
 import static dev.langchain4j.internal.Utils.getOrDefault;
 import static dev.langchain4j.model.chat.Capability.RESPONSE_FORMAT_JSON_SCHEMA;
@@ -51,7 +56,7 @@ import static java.util.Collections.emptyList;
 public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
 
     private final OpenAiClient client;
-    private final Integer maxRetries;
+    private final InvocationPolicy invocationPolicy;
 
     private final OpenAiChatRequestParameters defaultRequestParameters;
     private final String responseFormat;
@@ -63,7 +68,6 @@ public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
     private final List<ChatModelListener> listeners;
 
     public OpenAiChatModel(OpenAiChatModelBuilder builder) {
-
         if ("demo".equals(builder.apiKey)) {
             // TODO remove before releasing 1.0.0
             throw new RuntimeException("""
@@ -85,7 +89,8 @@ public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
                 .userAgent(DEFAULT_USER_AGENT)
                 .customHeaders(builder.customHeaders)
                 .build();
-        this.maxRetries = getOrDefault(builder.maxRetries, 3);
+
+        this.invocationPolicy = builder.invocationPolicy();
 
         ChatRequestParameters commonParameters;
         if (builder.defaultRequestParameters != null) {
@@ -196,31 +201,24 @@ public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
         ChatCompletionRequest openAiRequest =
                 toOpenAiChatRequest(chatRequest, parameters, strictTools, strictJsonSchema).build();
 
-        try {
-            ChatCompletionResponse openAiResponse = withRetry(() ->
-                    client.chatCompletion(openAiRequest).execute(), maxRetries);
+        // TODO [MF]: check how this works with the streaming API
+        ChatCompletionResponse openAiResponse = invokePolicy(invocationPolicy,
+                () -> client.chatCompletion(openAiRequest).execute());
 
-            OpenAiChatResponseMetadata responseMetadata = OpenAiChatResponseMetadata.builder()
-                    .id(openAiResponse.id())
-                    .modelName(openAiResponse.model())
-                    .tokenUsage(tokenUsageFrom(openAiResponse.usage()))
-                    .finishReason(finishReasonFrom(openAiResponse.choices().get(0).finishReason()))
-                    .created(openAiResponse.created().longValue())
-                    .serviceTier(openAiResponse.serviceTier())
-                    .systemFingerprint(openAiResponse.systemFingerprint())
-                    .build();
+        OpenAiChatResponseMetadata responseMetadata = OpenAiChatResponseMetadata.builder()
+                .id(openAiResponse.id())
+                .modelName(openAiResponse.model())
+                .tokenUsage(tokenUsageFrom(openAiResponse.usage()))
+                .finishReason(finishReasonFrom(openAiResponse.choices().get(0).finishReason()))
+                .created(openAiResponse.created().longValue())
+                .serviceTier(openAiResponse.serviceTier())
+                .systemFingerprint(openAiResponse.systemFingerprint())
+                .build();
 
-            return ChatResponse.builder()
-                    .aiMessage(aiMessageFrom(openAiResponse))
-                    .metadata(responseMetadata)
-                    .build();
-        } catch (RuntimeException e) {
-            if (e.getCause() instanceof HttpException httpException) {
-                throw httpException;
-            } else {
-                throw e;
-            }
-        }
+        return ChatResponse.builder()
+                .aiMessage(aiMessageFrom(openAiResponse))
+                .metadata(responseMetadata)
+                .build();
     }
 
     @Override
@@ -278,12 +276,14 @@ public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
         private Map<String, String> metadata;
         private String serviceTier;
         private Duration timeout;
-        private Integer maxRetries;
         private Boolean logRequests;
         private Boolean logResponses;
         private Tokenizer tokenizer;
         private Map<String, String> customHeaders;
         private List<ChatModelListener> listeners;
+
+        private RetryUtils.RetryPolicy retryPolicy = DEFAULT_RETRY_POLICY;
+        private ExceptionMapper exceptionMapper = OpenAiExceptionMapper.INSTANCE;
 
         public OpenAiChatModelBuilder() {
             // This is public so it can be extended
@@ -426,7 +426,12 @@ public class OpenAiChatModel implements ChatLanguageModel, TokenCountEstimator {
         }
 
         public OpenAiChatModelBuilder maxRetries(Integer maxRetries) {
-            this.maxRetries = maxRetries;
+            retryPolicy = retryPolicyBuilder()
+                    .maxAttempts(maxRetries)
+                    .delayMillis(500)
+                    .jitterScale(0.2)
+                    .backoffExp(1.5)
+                    .build();
             return this;
         }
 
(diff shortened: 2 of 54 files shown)


langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,29 @@
...
+        };
+    }
+
+    RuntimeException mapException(Exception e);
+
+    class DefaultExceptionMapper implements ExceptionMapper {
+        @Override
+        public RuntimeException mapException(Exception e) {
+            System.out.println(e.getMessage());
+            return e instanceof RuntimeException re ? re : new RuntimeException(e);
```
mariofusco: Default exception mapping implementation should go there. Of course we could implement a different mapping for each language model.

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
```diff
@@ -0,0 +1,47 @@
...
+
+import static dev.langchain4j.internal.RetryUtils.DEFAULT_RETRY_POLICY;
+
+public interface InvocationPolicy {
+
+    InvocationPolicy DEFAULT = ExceptionMapper.DEFAULT.andThen(RetryUtils.DEFAULT_RETRY_POLICY);
+
+    Callable<?> apply(Callable<?> action);
+
+    default InvocationPolicy andThen(InvocationPolicy invocationPolicy) {
```
mariofusco: This allows to compose multiple `InvocationPolicy`.

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
```diff
@@ -0,0 +1,47 @@
+package dev.langchain4j.internal;
+
+import java.util.concurrent.Callable;
+
+import static dev.langchain4j.internal.RetryUtils.DEFAULT_RETRY_POLICY;
+
+public interface InvocationPolicy {
+
+    InvocationPolicy DEFAULT = ExceptionMapper.DEFAULT.andThen(RetryUtils.DEFAULT_RETRY_POLICY);
```
mariofusco: The default invocation policy is a composition of the default retry policy and exception mapping.

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
```diff
@@ -28,7 +28,7 @@ public static RetryPolicy.Builder retryPolicyBuilder() {
     /**
      * This class encapsulates a retry policy.
      */
-    public static final class RetryPolicy {
+    public static final class RetryPolicy implements InvocationPolicy {
```
mariofusco, line 32: The `RetryPolicy` is now an `InvocationPolicy`.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
```diff
@@ -448,7 +455,12 @@ public OpenAiChatModelBuilder timeout(Duration timeout) {
         }
 
         public OpenAiChatModelBuilder maxRetries(Integer maxRetries) {
-            this.maxRetries = maxRetries;
+            retryPolicy = retryPolicyBuilder()
```
mariofusco, line 429: I did this for backward compatibility. Not sure if we want to keep this though.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
```diff
@@ -229,8 +234,8 @@ public ChatResponse doChat(ChatRequest chatRequest) {
                 toOpenAiChatRequest(chatRequest, parameters, strictTools, strictJsonSchema).build();
 
         try {
-            ChatCompletionResponse openAiResponse = withRetry(() ->
-                    client.chatCompletion(openAiRequest).execute(), maxRetries);
+            ChatCompletionResponse openAiResponse = invocationPolicy.execute(() ->
```
mariofusco: The LLM client is now invoked through the invocation policy, encapsulating both the retry policy and the exception mapping (plus eventually any other strategy that we would need to add in future).

dliubarskyi:
@mariofusco thanks a lot! I will review it ASAP, most probably tomorrow

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
```diff
@@ -0,0 +1,47 @@
...
+
+    Callable<?> apply(Callable<?> action);
```
dliubarskyi: What is the difference between `apply` and `execute`? It is a bit confusing

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
```diff
@@ -0,0 +1,47 @@
...
+
+    default InvocationPolicy andThen(InvocationPolicy invocationPolicy) {
+        return action -> invocationPolicy.apply(this.apply(action));
+    }
+
+    default <T> T execute(Callable<T> callable) {
+        try {
+            return (T) apply(callable).call();
+        } catch (Exception e) {
+            throw e instanceof RuntimeException re ? re : new RuntimeException(e);
```
dliubarskyi: Do we need this here? It feels like this should be a part of an exception mapper?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
dliubarskyi: I would keep it, as it seems it is not expensive/harmful?

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
dliubarskyi: Nice! Shouldn't it be the other way around though? First we retry, and only if retries were unsuccessful we map the exception before finally throwing it to the user?
I am not sure if it is the implementation or the naming, but I would probably expect to see something like 
`RetryUtils.DEFAULT_RETRY_POLICY.andThen(ExceptionMapper.DEFAULT)`?

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
```diff
@@ -0,0 +1,47 @@
+package dev.langchain4j.internal;
+
+import java.util.concurrent.Callable;
+
+import static dev.langchain4j.internal.RetryUtils.DEFAULT_RETRY_POLICY;
+
+public interface InvocationPolicy {
```
dliubarskyi: naming question: is it common to name things like this a "policy"? Would "strategy" or some other name be more suitable?

langchain4j/src/test/java/dev/langchain4j/service/AiServiceThrowingExceptionIT.java
```diff
@@ -0,0 +1,63 @@
...
+        }).withInvocationPolicy(InvocationPolicy.DEFAULT);
+
+        ChatMemory chatMemory = MessageWindowChatMemory.withMaxMessages(10);
+
+        ThrowingService assistant = AiServices.builder(ThrowingService.class)
+                .chatLanguageModel(chatLanguageModel)
+                .chatMemory(chatMemory)
+                .build();
+
+        assertThrows(RuntimeException.class, () -> assistant.chat("hi"));
```
dliubarskyi: Perhaps out of scope of this PR, but since we have a common `HttpException` now, I guess we also need to update the default retry strategy and/or exception mapper to not wrap `HttpException` into the `RuntimeException`

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,29 @@
+package dev.langchain4j.internal;
```
dliubarskyi: Do we want to keep it in the `internal` package? Aren't users supposed to be able to customize it?
(Same for `InvocationPolicy`)

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
dliubarskyi: Great! Will this work for the streaming models?

dliubarskyi: BTW existing exception mapping logic is still there (line 255), I guess we should remove it?

dliubarskyi (commented):
@mariofusco thanks a lot, I like the idea!
Do you have other use cases in mind aside from exception mapping and retries?

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
mariofusco: My understanding was that first we map the exception and then based on the kind of mapped exception we decide if we should retry or not, but it could also be the other way around indeed. Or with this design we could also have 2 levels of exception mapping one before the retry and the other after all the retry attempts have failed.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
mariofusco: Ah, I didn't notice it. Yes we should move it in the execeptions mapper.

mariofusco:
> @mariofusco thanks a lot, I like the idea! Do you have other use cases in mind aside from exception mapping and retries?

Any cross-cutting concern that we may want to apply around the LLM invocation. For instance we could allow users to plug their own logging immediately after and before that invocation. And if you think about it I believe that also guardrails could be implemented in this way.

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
mariofusco: Yes, once we have figured out the API that we want to expose.

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
mariofusco: I used "policy" only because the first thing that I implemented was the retry policy. I'm not strongly opinionated on naming, and yes, probably strategy would be a better name. I will change it.

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
mariofusco: The call method throws a checked exception, so I'm obliged to either catch it or declare it in the method signature which is something that I wanted to avoid, but I could review this decision.

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
mariofusco: The `apply` method (maybe the name is misleading?) applies this invocation policy to the given `Callable` transforming it into another Callable that, when invoked through the `execute` method, will "decorate" its execution with this policy. In essence the logic of the specific `InvocationPolicy`  has to be implemented in this `apply` method while the `execute` one is to actually trigger the invocation and should never be changed (even though I cannot declare `final` an interface method). Since the `execute` method is never supposed to be called by users, but only by the LLM maybe it would be better to bring this out of this interface? Do you have any better naming suggestion that `apply`?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
mariofusco: > Great! Will this work for the streaming models?

I didn't give this a try yet, but I'm planning to do so during my flight back home from Las Vegas :)

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
dliubarskyi: Oh, you are totally right! I had a brain fart, sorry 😅

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
dliubarskyi: I see, thanks for the explanation! I was confused by similar names.

> Since the execute method is never supposed to be called by users, but only by the LLM maybe it would be better to bring this out of this interface?

Yes, sounds good

> Do you have any better naming suggestion that apply?

I am not sure, perhaps "decorate" or something like this? I am not sure about any of those names TBH, as the interface is quite generic but it is called in a very specific point of time, around the actual call to the LLM (usually over HTTP). Do we want to reflect that in naming somehow?

dliubarskyi:
cc @TheEliteGentleman, we were discussing an option to plug in other retry mechanisms, will this PR help you with this?

TheEliteGentleman:
In Microprofile, Helidon or Quarkus, Fault Tolerance are triggered through annotations, e.g, `@Retry` that calls a CDI Fault tolerance interceptor, via CDI. Hence, the best way of triggering fault tolerance would be on a AI Service that interface with a client directly, with all invocation turned off and let the CDI container manage it (including Asynchronous call).

langchain4j-core/src/main/java/dev/langchain4j/internal/InvocationPolicy.java
edeandrea: Should probably mark this with `@FunctionalInterface`?

mariofusco: Done, thanks.

mariofusco:
@dliubarskyi I reworked this a bit (please give a look), but still there 2 main open issues/doubts:

1. **Streaming**: does this API make sense at all in streaming mode? I noticed that we hadn't the retry there and indeed I don't see how to fit it in how it works. Also the exception mapping is problematic: provided that we want it, it should be applied only to the `onError` callback, correct? Or could the `chatCompletion` method also throw a runtime exception and we want to map it?
2. **Guardrails**: the current `Callable`-based implementation, is too low level and the Callable itself is a black box that wouldn't allow to implement guardrails in a meaningful way. I could probably further refactor and change the signature of the `InvocationPolicy` function, but I wonder if we want to go in that direction. Moreover also here it is not clear to make this fit the streaming mode, but I'm about to give a look at the Quarkus integration to check how it is integrated there.

edeandrea:
> @dliubarskyi I reworked this a bit (please give a look), but still there 2 main open issues/doubts:
> 
> 1. **Streaming**: does this API make sense at all in streaming mode? I noticed that we hadn't the retry there and indeed I don't see how to fit it in how it works. Also the exception mapping is problematic: provided that we want it, it should be applied only to the `onError` callback, correct? Or could the `chatCompletion` method also throw a runtime exception and we want to map it?

In the Quarkus extension we support retry in streaming mode. Granted I think we support it outside of LC4j using the fault tolerance stuff, which is integrated with Mutiny, so I'm not sure how that would play here 

> 2. **Guardrails**: the current `Callable`-based implementation, is too low level and the Callable itself is a black box that wouldn't allow to implement guardrails in a meaningful way. I could probably further refactor and change the signature of the `InvocationPolicy` function, but I wonder if we want to go in that direction. Moreover also here it is not clear to make this fit the streaming mode, but I'm about to give a look at the Quarkus integration to check how it is integrated there.

I haven't gotten too far in my implementation yet, but what I'm aiming for is something that is common to all the model implementations so that we don't have to touch/modify each and every implementation. We'll see how that goes. 

I'm not sure when I will finish the guardrails stuff. I'm on PTO most of next week and then at a conference most of the week after.

dliubarskyi:
@mariofusco thanks a lot!

> Streaming: does this API make sense at all in streaming mode? I noticed that we hadn't the retry there and indeed I don't see how to fit it in how it works.

I generally neglected retries for streaming models, but there are some models that support retries (e.g., `GitHubModelsStreamingChatModel`), so it is worth checking how they do it there.

> Also the exception mapping is problematic: provided that we want it, it should be applied only to the onError callback, correct? Or could the chatCompletion method also throw a runtime exception and we want to map it?

`onError` for sure, but perhaps also for `chatCompletion` (this needs to be investigated, I do not have a ready answer, sorry)

> Guardrails: the current Callable-based implementation, is too low level and the Callable itself is a black box that wouldn't allow to implement guardrails in a meaningful way. I could probably further refactor and change the signature of the InvocationPolicy function, but I wonder if we want to go in that direction. Moreover also here it is not clear to make this fit the streaming mode, but I'm about to give a look at the Quarkus integration to check how it is integrated there.

I would not use this mechanism for guardrails. At this (very low) level we do not have access to the chat memory, all the tools, etc.
I agree with @edeandrea that ideally guardrails should not require duplicating logic for each model implementation.

I also have a concern about using the same mechanism for retries and exception mapping (and chaining them). I might be wrong but I assume that users will usually not do (custom) exception mapping and will use what we setup as a default exception mapper. But custom retry policy will most probably be specified much more often. So ideally there should be a way to set retry policy without overriding/dropping default exception mapper. Otherwise we will need to ask users to specify both retry policy and exception mapping if they want to override one or the other. WDYT?

langchain4j-core/src/main/java/dev/langchain4j/model/chat/policy/ExceptionMapper.java
```diff
@@ -0,0 +1,29 @@
...
+            }
+        };
+    }
+
+    RuntimeException mapException(Exception e);
+
+    class DefaultExceptionMapper implements ExceptionMapper {
+        @Override
+        public RuntimeException mapException(Exception e) {
+            System.out.println(e.getMessage());
```
dliubarskyi, line 25: please don't forget to remove this line

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
dliubarskyi: How default implementation would look like?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
```diff
@@ -458,5 +463,43 @@ public OpenAiChatModelBuilder listeners(List<ChatModelListener> listeners) {
         public OpenAiChatModel build() {
             return new OpenAiChatModel(this);
         }
+
+        private InvocationPolicy invocationPolicy() {
+            return this.exceptionMapper.andThen(this.retryPolicy);
+        }
+
+        @Override
+        public String toString() {
```
dliubarskyi, line 472: please remove `toString`, it was just removed recently

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiExceptionMapper.java
```diff
@@ -0,0 +1,20 @@
...
+
+import dev.langchain4j.exception.HttpException;
+import dev.langchain4j.model.chat.policy.ExceptionMapper;
+
+class OpenAiExceptionMapper implements ExceptionMapper {
+
+    static final OpenAiExceptionMapper INSTANCE = new OpenAiExceptionMapper();
+
+    @Override
+    public RuntimeException mapException(Exception e) {
```
dliubarskyi, line 11: It seems that this logic could be a good **_default_** exception mapper, WDYT? There is nothing OpenAI specific here

langchain4j/src/test/java/dev/langchain4j/service/AiServiceThrowingExceptionIT.java
```diff
@@ -0,0 +1,63 @@
...
```
dliubarskyi:
```suggestion
        assertThatThrownBy(() -> assistant.chat("hi"))
                .isExactlyInstanceOf(HttpException.class)
                .hasMessage("Insufficient quota");
```

mariofusco:
Replaced by https://github.com/langchain4j/langchain4j/pull/2607

PR #2537, opened 2025-02-12
Router support DeepSeek

dzy888:
## Issue

Closes #
I found that routing the DeepSeek model in LanguageModelQueryRouter fails because the DeepSeek model responds with a thinking tag<think></think>
## Change

We filtered the results returned by calling the model
one add langchain4j-core/src/main/java/dev/langchain4j/rag/query/router/FilterRouter.java
two add langchain4j-core/src/main/java/dev/langchain4j/rag/query/router/FilterRouterChain.java
three add langchain4j-core/src/main/java/dev/langchain4j/rag/query/router/DeepSeekFilter.java
four in langchain4j-core/src/main/java/dev/langchain4j/rag/query/router/LanguageModelQueryRouter.java
add field filterRouterChain and in method route()
test langchain4j-core/src/test/java/dev/langchain4j/rag/query/router/LanguageModelQueryRouterTest.java
method should_support_Deepseek_in_ollama_remove_thinking_tags_success() and 
should_support_Deepseek_in_ollama_remove_thinking_tags_failed()

## General checklist

- [ X] There are no breaking changes
- [ X] I have added unit and/or integration tests for my change
- [X ] The tests cover both positive and negative cases
- [ X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green


minglu7:
Actually, the deepseek r1 large model places reasoning_content and content at the same level. Your version, which puts the  content of CoT inside the content, should be a third-party modified version.    

image
image

dzy888:
The situation I encountered was Deepseek r1 in Olama

dzy888:
I didn't express myself clearly before

dliubarskyi (commented):
@dzy888 thank you, but I am not sure this problem (BTW it might also be temporary) requires introducing new funcitonality. You can extend `LanguageModelQueryRouter` and override the `parse` method where you can do the filtering logic easily.

dzy888:
> @dzy888谢谢，但我不确定这个问题（顺便说一句，它也可能是暂时的）是否需要引入新功能。您可以扩展`LanguageModelQueryRouter`和覆盖`parse`可以轻松执行过滤逻辑的方法。
Thank you for your response. As a user, I think adding filters would be more convenient than rewriting method 'parse'

PR #2545, opened 2025-02-13
Implementation of returnDirect in @Tool annotation

### LINKED ISSUE #2542: [FEATURE] Implement returnDirect for LangChain4j @Tool annotation

Currently, when using tools in LangChain4j, there isn't a straightforward way for a tool to immediately return a result and halt the agent loop's progression.  The agent always continues its loop after a tool is executed, even if the tool's result is conclusive and further steps are unnecessary. This can lead to unnecessary processing and potentially less efficient agent behavior.

Adding a returnDirect parameter to the @Tool annotation, that when set to true would allow a tool to return its result directly to the user, bypassing the subsequent steps in the agent loop.  This would mirror the functionality of the return_direct parameter available in the Python version of Langchain's @tool decorator.  The default value for this parameter would be false, maintaining the current behavior unless explicitly enabled.  

I haven't considered any alternative solution as the returnDirect parameter in the Python version seems to be the most direct and intuitive way to address this.  Without this feature, achieving similar behavior might require more complex workarounds within the agent logic, potentially making the agent implementation less clean and more cumbersome

**Additional context**

Here's the documentation for the [Python implementation](https://api.python.langchain.com/en/latest/tools/langchain_core.tools.tool.html). Implementing this parameter would bring feature parity with the Python version and provide a valuable tool for controlling agent behavior and optimizing efficiency

chrischriscris:
## Issue

Closes #2542 

## Change

Implemented the returnDirect parameter for the @ Tool annotation in LangChain4j.

This allows tool methods to immediately return a result and halt the agent loop, improving efficiency in scenarios where a tool's response is conclusive.

The following changes were made:

    Added a returnDirect element (boolean type with default false) to the @ Tool annotation.
    Modified ToolSpecification to include the returnDirect property.
    Updated ToolService.java to read the returnDirect value from ToolSpecification within the executeInferenceAndToolsLoop method.
    Adjusted the agent loop logic in ToolService.java to terminate the loop and return the tool's result directly when returnDirect is true.
    Adjusted test_toString in ToolSpecificationTest.
    (Will) add unit tests to verify the returnDirect functionality, covering both true and false cases.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
@@ -1,10 +1,10 @@
 package dev.langchain4j.agent.tool;
 
-import java.lang.annotation.Retention;
-import java.lang.annotation.Target;
-
 import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
 
 /**
  * Java methods annotated with {@code @Tool} are considered tools/functions that language model can execute/call.
@@ -36,4 +36,10 @@ public @interface Tool {
      * @return description of the tool.
      */
     String[] value() default "";
+
+    /**
+     * Whether to return directly from the tool rather than continuing the agent loop.
+     * Defaults to false.
+     */
+    boolean returnDirect() default false;
 }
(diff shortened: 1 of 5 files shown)


dliubarskyi:
Hi @chrischriscris, thanks a lot for your PR!
There is https://github.com/langchain4j/langchain4j/pull/2291 already, so I will close your PR to not duplicate the effort.
Thanks for understanding!

Issue #2547, opened 2025-02-13
[FEATURE] Update the following listeners to subscribe to Telemetry observability


TheEliteGentleman:
To adhere to the following requirements for Telemetry observabilty:
- Spans: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/
- Metrics: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/

We need the following to be included on Langchain4J:

1. Add a Listener to EmbeddingModel (and the modelName on the Parameters will be `embeddings`), so similar to `ChatRequest`, `ChatResponse`, we might need `EmbeddingRequest`, `EmbeddingResponse`.
2. Add the following fields to `ChatRequestParameters`.
	-  aiSystem - to determine which AI system we're interfacing (openai, deepseek, gemini, etc).
	- server address/port (if any) - of the AI.
	- seed, from ChatLanguageModel request.

Thanks

dliubarskyi:
Hi @TheEliteGentleman,

1. That is in the plans, but I am not sure we will have capacity for this before 1.0 release, so perhaps it will come a bit later
2. That's in the plans as well, there is a [prototype](https://github.com/langchain4j/langchain4j/commit/88d426bfb21b34382db0190faf84fa924cf83b75) that was triggered by https://github.com/langchain4j/langchain4j/pull/2384, but I guess we can do that without waiting for #2384 to be merged. Not sure about server/port though. How critical is server/port?

TheEliteGentleman:
@dliubarskyi the spec states server add/port as `RECOMMENDED`. I've also included `seed` as per the attribute requirement.
I guess we're trying to capture as many fields as mentioned on the OT attribute requirements.

dliubarskyi:
@TheEliteGentleman I see. Eventually we will add recommended server/port, but I can't promise this will happen very soon, as this will require proper design of common connection parameters (similar to the chat request parameters). Also, we will probably have to parse the `baseUrl` to separate the `server` from the `port`, which is not nice...

dliubarskyi:
@TheEliteGentleman BTW you should have access to the server/port from your side. If user have configured `baseUrl` on the model, then you can get that data from it. If not, then the `baseUrl` is the default one (which does not notmally change, so you can hardcode it). I understand it would be nice for LC4j to provide this into OOTB, but there is nothing blocking you at this point to get that data yourself.

PR #2556, opened 2025-02-14
Propagate ModelProvider (OpenTelemetry's gen_ai.system) into ChatModelListener

dliubarskyi:
## Issue
Partially addresses #2547

## Change

Introduced `ModelProvider` enum that is currently exposed by `ChatLanguageModel` and `StreamingChatLanguageModel`, as well as `ChatModelRequestContext`, `ChatModelResponseContext` and `ChatModelErrorContext`.

`ModelProvider` can be mapped to the OpenTelemetry's `gen_ai.system`. For example: `ModelProvider.GOOGLE_GEMINI` -> `gemini`

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/model/ModelProvider.java
```diff
@@ -0,0 +1,14 @@
+package dev.langchain4j.model;
+
+public enum ModelProvider {
+
+    ANTHROPIC,
+    AMAZON_BEDROCK,
+    AZURE_AI_INFERENCE,
+    AZURE_OPEN_AI,
+    GOOGLE_GEMINI, // TODO
```
dliubarskyi: cc @glaforge

Could you please check if I did mapping correctly?

https://opentelemetry.io/docs/specs/semconv/attributes-registry/gen-ai/ defines `gemini` and `vertex_ai` values.

I have mapped both `GoogleAiGeminiChatModel` and `VertexAiGeminiChatModel` to the `GOOGLE_GEMINI` enum value (which can be mapped to OTel's `gemini` value in another OTel-specific module).

Since `VertexAiChatModel` does not have observability (do we plan to add it at some point?), I did not add  a value for `vertex_ai` in the `ModelProvider` enum.

Do you think it makes sense or do we need separate  enum values (e.g., GOOGLE_AI_GEMINI, `GOOGLE_VERTEX_AI_GEMINI`)?

dliubarskyi: bump :)

dliubarskyi: Eventually I went with separate `GOOGLE_AI_GEMINI` and `GOOGLE_VERTEX_AI_GEMINI` as it is more flexible.
Please LMK if you prefer other options

PR #2582, opened 2025-02-19
Replacing "List<Model> models" with "List<ChatModelCapabilities<Model>>" in AbstractBaseChatModelIT

tenpigs267:
## Issue

Closes #

## Change
As discussed in #2404 models tested by an AbstractBaseChatModelIT inheritor can have differents capabilities (often the ligths version doesn't support multimodality : example GPT3.5 does'nt support images).
The idea would be to have
```java
public record ChatLanguageModelCapabilities(
    ChatLanguageModel chatLanguageModel,
    boolean supportsModelNameParameter,
    boolean supportsMaxOutputTokensParameter,
    boolean supportsStopSequencesParameter,
    boolean supportsTemperatureParameter,
    boolean supportsTopPParameter,
    boolean ...
) {}
```

and tests like:

```java
@ParameterizedTest
@MethodSource("models")
protected void shouldHandleStopSequencesBasedOnModelCapabilities(ChatLanguageModelCapabilities modelCapabilities) {
    if (modelCapabilities.supportsStopSequencesParameter()) {
        // should support stop sequences
    } else {
        // should fail
    }
}
```

The current combination:

```java
@ParameterizedTest
@MethodSource("models")
@DisabledIf("supportsStopSequencesParameter")
protected void should_fail_if_stopSequences_parameter_is_not_supported(M model) { ... }

@ParameterizedTest
@MethodSource("models")
@EnabledIf("supportsStopSequencesParameter")
protected void should_respect_stopSequences_in_chat_request(M model) { ... }
```

is not as flexible. The combination of `@EnabledIf` and `@DisabledIf` means either you test models that support the feature or those that don't, but you can't test in the same integration test both models that should fail and those that should pass the test.  

## General checklist


tenpigs267:
Hello @dliubarskyi,
I think I will set it "ready for review" tomorrow (quota stopped me from getting all the modified tests to run).

In GoogleAiGeminiChatModelIT, should_handle_execute_a_tool_then_answer fail, I tried with old and new implementation and it fails in both scenarii. If i modify the model used from gemini-1.5-flash-8b to gemini-1.5-flash then the test doesn't fail. Do you have the same issue on the main CI?

I'm having trouble with VertexAIChatModelIT, maybe you can help me. I have been able to launch ollama (with patience), openAI, Azure OpenAI and Gemini ITs but VertexAIChatModelIT still resists. I can see here https://cloud.google.com/vertex-ai/generative-ai/docs/deprecations/palm that newly created project can't use it. So the modified IT should be okay but I can't be sure.
Thanks

tenpigs267:
Also I introduced a DISABLED value for chatModelCapabilities, Ollama IT was too complicated with things like image in b64 silently ignored (so should_fail tests could not be green). It is not ideal but it is expressive and can also be useful when developping as you can enable/disable tests but DISABLED value should not pass the review process.

PR #2508, opened 2025-02-05
Support the official OpenAI Java SDK

jdubois:
Fix #2373 

This support the Chat, Embedding and Image models, with full testing.

PR #2590, opened 2025-02-21
Update azure-ai-inference from 1.0.0-beta.2 to 1.0.0-beta.3

jdubois:
This is used by the langchain4j-github-models module


jdubois:
@dliubarskyi once #2508 is merged, we'll also need to have a look at deprecating this module... I just tested and the OpenAI Java SDK does the same, so that would reduce the maintenance.

jdubois:
Closing as 1.0.0-beta.4 is released

PR #2593, opened 2025-02-23
Add method addAll to ChatMemory (closes #2479)

### LINKED ISSUE #2479: [FEATURE] Add Method to ChatMemory to Insert Multiple ChatMessages
### **Feature Request: Add Method to Insert Multiple ChatMessages**

Currently, the `ChatMemory` interface only provides a method to add a single `ChatMessage` via:  

```java
void add(ChatMessage message);
```

This means that adding multiple messages requires calling this method repeatedly in a loop, which is inefficient for implementations that persist messages to a database or other storage.  

Additionally, there is no way to indicate that a set of messages belongs to the same batch or context. Each message is added independently, making it difficult for implementations to optimize storage, retrieval, or batch processing.  

Introduce a new method in the `ChatMemory` interface that allows adding multiple messages at once:  

```java
void addAll(List<ChatMessage> messages);
```

This would allow for more efficient batch inserts in database-backed implementations, reduce I/O overhead, and explicitly indicate that the messages in the list are related.  

- **Keeping the current approach**: Calling `add(ChatMessage message)` multiple times, but this is inefficient and does not convey any batch relationships.  

**Additional context**  

Adding this method would align with best practices for working with databases or other storage systems that support batch operations. It would enable more efficient handling of chat histories and improve performance for implementations that need to persist messages.  

**Example Use Case**  

In the `DefaultCommittableChatMemory` class from the [`quarkus-langchain4j`](https://github.com/quarkiverse/quarkus-langchain4j/blob/21603945ce91fd59452eb68459adb9c247434563/core/runtime/src/main/java/io/quarkiverse/langchain4j/runtime/aiservice/DefaultCommittableChatMemory.java#L66) project, there is a scenario where multiple `ChatMessage` instances are added to the memory. Currently, this is done by iterating over a list and adding each message individually:  

```java
for (ChatMessage message : messages) {
    delegate.add(message);
}
```

With the proposed `addAll` method, this could be simplified and optimized as follows:  

```java
delegate.addAll(messages);
```

This change would reduce the overhead associated with multiple method calls and allow the underlying implementation to handle batch inserts more efficiently.  

Would love to hear feedback from maintainers on this! 🚀

ungerts:
## Issue

Closes #2479 

## Change

Added the addAll method to ChatMemory and implemented it in MessageWindowChatMemory and TokenWindowChatMemory. Introduced the abstract class AbstractWindowChatMemory as a foundation for window-based chat memory implementations.

## General checklist

### DIFF
--- a/langchain4j-cassandra/src/test/java/dev/langchain4j/store/memory/chat/cassandra/CassandraChatMemoryStoreTestSupport.java
+++ b/langchain4j-cassandra/src/test/java/dev/langchain4j/store/memory/chat/cassandra/CassandraChatMemoryStoreTestSupport.java
@@ -1,11 +1,15 @@
 package dev.langchain4j.store.memory.chat.cassandra;
 
+import static dev.langchain4j.data.message.AiMessage.aiMessage;
+import static dev.langchain4j.data.message.UserMessage.userMessage;
+import static org.assertj.core.api.Assertions.assertThat;
+
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.UserMessage;
 import dev.langchain4j.memory.ChatMemory;
 import dev.langchain4j.memory.chat.MessageWindowChatMemory;
-import dev.langchain4j.memory.chat.TokenWindowChatMemory;
-import dev.langchain4j.model.openai.OpenAiTokenizer;
+import java.util.List;
+import java.util.UUID;
 import lombok.extern.slf4j.Slf4j;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.MethodOrderer;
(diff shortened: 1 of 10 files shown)


langchain4j-cassandra/src/test/java/dev/langchain4j/store/memory/chat/cassandra/CassandraChatMemoryStoreTestSupport.java
```diff
@@ -75,10 +73,8 @@ void shouldInsertItems() {
 
         // When
         UserMessage userMessage = userMessage("I will ask you a few question about ff4j.");
-        chatMemory.add(userMessage);
-
         AiMessage aiMessage = aiMessage("Sure, go ahead!");
-        chatMemory.add(aiMessage);
+        chatMemory.addAll(List.of(userMessage, aiMessage));
```
kpavlov, line 77: would be nice to have also a signature with vararg parameters

bitsofinfo:
please add this!

PR #2607, opened 2025-02-25
Introduce ExceptionMapper

mariofusco:
Considering what discussed [here](https://github.com/langchain4j/langchain4j/pull/2535#issuecomment-2664972164) I created a second proposal, much simpler and less invasive, for exception mapping, keeping it completely separated from the retry policy, even taking count that, differently from the retry policy, it will be unlikely that we want to make this exception mapping mechanism configurable and pluggable.

@dliubarskyi @edeandrea Any feedback on this is welcome.


langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,55 @@
...
+            throw mapException(e);
+        }
+    }
+
+    RuntimeException mapException(Exception e);
+
+    class DefaultImpl implements ExceptionMapper {
+
+        @Override
+        public RuntimeException mapException(Exception e) {
```
mariofusco, line 38: I sketched the default exception mapping here. Not sure if we could do something much better than this (especially if we want to keep this generic), but we could introduce specific implementation if required by some models.

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
```diff
@@ -190,9 +191,11 @@ public <T> T withRetry(Callable<T> action, int maxAttempts) {
             while (true) {
                 try {
                     return action.call();
+                } catch (UnrecoverableException e) {
```
mariofusco: When the `ExceptionMapper` generates an `UnrecoverableException`, then the retry policy simply gives up and rethrow it as it is.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
```diff
@@ -196,31 +197,23 @@ public ChatResponse doChat(ChatRequest chatRequest) {
...
-                    .created(openAiResponse.created().longValue())
-                    .serviceTier(openAiResponse.serviceTier())
-                    .systemFingerprint(openAiResponse.systemFingerprint())
-                    .build();
-
-            return ChatResponse.builder()
-                    .aiMessage(aiMessageFrom(openAiResponse))
-                    .metadata(responseMetadata)
-                    .build();
-        } catch (RuntimeException e) {
```
mariofusco, line 184: This is now managed by the `ExceptionMapper`.

langchain4j-core/src/main/java/dev/langchain4j/exception/RateLimitException.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.exception;
+
+public class RateLimitException extends RuntimeException {
```
dliubarskyi: Does it make sense to create some "base LC4j exception" that will be a parent for both `RateLimitException` and `UnrecoverableException`?
Might be even one per API/model type, like `BaseChatModelException`? WDYT?

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,55 @@
...
+            Throwable rootCause = findRoot(e);
```
dliubarskyi, line 39: Could you please explain why we need to search for root?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
```diff
@@ -196,31 +197,23 @@ public ChatResponse doChat(ChatRequest chatRequest) {
...
-            if (e.getCause() instanceof HttpException httpException) {
-                throw httpException;
-            } else {
-                throw e;
-            }
-        }
+        ChatCompletionResponse openAiResponse = withRetry(() ->
+                mappingException(() -> client.chatCompletion(openAiRequest).execute()), maxRetries);
```
dliubarskyi: Would it make sense to call `mappingException` from inside the `withRetry` method? So that all users of `withRetry` will get this behaviour without the need to update them (90+ usages of `withRetry`)?
Maybe some `try-catch` blocs will have to be removed (as with `OpenAiChatModel` case), but it shouldn't be so big

langchain4j/src/test/java/dev/langchain4j/service/common/openai/OpenAiListenableChatModelIT.java
```diff
@@ -133,7 +134,7 @@ void should_listen_error() {
 
         // then
         Throwable exceptionReportedToListener = listener.errorContextReference.get().error();
-        assertThat(exceptionReportedToListener).isExactlyInstanceOf(HttpException.class);
+        assertThat(exceptionReportedToListener).isExactlyInstanceOf(UnrecoverableException.class);
```
dliubarskyi: I am wondering if we need to create a more granular exception hierarchy. Catching `UnrecoverableException` will not help the user much. The only piece of information user is getting is that there is no way to recover, so retries make no sense. But retries are handled by chat model, so this piece of information is kind of useless I guess? Not sure, jsut talking out loud...

dliubarskyi:
```suggestion
        assertThat(exceptionReportedToListener).isExactlyInstanceOf(UnrecoverableException.class).hasRootCauseExactlyInstanceOf(HttpException.class);
```

langchain4j-core/src/main/java/dev/langchain4j/exception/UnrecoverableException.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.exception;
+
+public class UnrecoverableException extends RuntimeException {
```
dliubarskyi: Should we have a `RecoverableException` as well? `RateLimitException` could extend it

dliubarskyi (commented):
@mariofusco thanks a lot! Looks good to me, just a few questions/suggestions

langchain4j-core/src/main/java/dev/langchain4j/exception/RateLimitException.java
mariofusco: +1

langchain4j-core/src/main/java/dev/langchain4j/exception/UnrecoverableException.java
mariofusco: +1

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
mariofusco: Internal `withRetry` could invoke `mappingException` in order to avoid duplicating everywhrere. Check what other modules do.

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,55 @@
...
+
+            if (rootCause instanceof HttpException httpException) {
+                return switch (httpException.statusCode()) {
+                    case 401, 404 -> new UnrecoverableException(rootCause);
```
mariofusco: Check Jlama integration (it doesn't go through http).

mariofusco:
@dliubarskyi I think I have addressed all what we discussed, please give a second look.

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,69 @@
...
+            if (rootCause instanceof UnresolvedAddressException) {
+                return new UnresolvedModelServerException(rootCause);
+            }
+
+            return e instanceof RuntimeException re ? re : new RuntimeException(e);
+        }
+
+        protected RuntimeException mapHttpStatusCode(Exception rootException, int httpStatusCode) {
+            return switch (httpStatusCode) {
+                case 400 -> new InvalidRequestException(rootException);
```
dliubarskyi: nit: perhaps we should map all the rest of 4xx to something (`InvalidRequestException`?) and all the rest of 5xx (to `InternalServerException`?)? We could add `is4xx()` and `is5xx()` to the `HttpException` to simplify the code here

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,69 @@
...
+        try {
+            return action.call();
+        } catch (Exception e) {
+            throw mapException(e);
+        }
+    }
+
+    RuntimeException mapException(Exception e);
+
+    class DefaultImpl implements ExceptionMapper {
```
dliubarskyi:
```suggestion
    class DefaultExceptionMapper implements ExceptionMapper {
```
nit: to be consistent with naming of all our other default implementations

langchain4j-core/src/main/java/dev/langchain4j/exception/BaseChatModelException.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.exception;
+
+public abstract class BaseChatModelException extends RuntimeException {
```
dliubarskyi: Not entirely sure, but I guess that it can be renamed into `BaseModelException` as the same type of errors will probably apply to all other model types like embedding model

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
```diff
@@ -231,14 +248,46 @@ public static <T> T withRetry(Callable<T> action, int maxAttempts) {
...
+     * If the action fails, the Exception causing the failure will be mapped with the default {@link ExceptionMapper}.
+     *
+     * @param action      The action to be executed.
+     * @param maxAttempts The maximum number of attempts to execute the action.
+     * @param <T> The type of the result of the action.
+     *
+     * @return The result of the action if it is successful.
+     * @throws RuntimeException if the action fails on all attempts.
+     */
+    public static <T> T withRetryMappingExceptions(Callable<T> action, int maxAttempts) {
```
dliubarskyi, line 275: Aside from various model classes, this method is also called by `GitHubDocumentLoader` and `GoogleAiGeminiTokenizer`. They are not really models, so it does not seem to make sense to use the default exception mapper which is currently focused on chat models, WDYT? Or is `ExceptionMapper` supposed to be used not only by models, but also by vector stores, docuemnt loaders, etc?

langchain4j-core/src/main/java/dev/langchain4j/exception/RecoverableException.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.exception;
+
+public abstract class RecoverableException extends BaseChatModelException {
```
dliubarskyi: In the future we will probably have to add a similar support for provider-agnostic exceptions to embedding stores and perhaps other APIs. Should all subtypes of `BaseChatModelException`/`BaseModelException` contain "ChatModel"/"Model" in it's name? Or should `BaseModelException` be renamed into `BaseException` or `BaseLangChain4jException` so that it could be later re-used with embedding stores?

dliubarskyi (commented):
@mariofusco thanks a lot!

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
mariofusco: You're right, I just applied the change extensively without looking at what the actual code is doing, I will revert those parts.

langchain4j-core/src/main/java/dev/langchain4j/exception/RecoverableException.java
mariofusco: At this point I would also get rid of the `Base`, call it simply `LangChain4jException` and make all our exceptions extend from it. I will do so.

mariofusco:
@dliubarskyi please give it another look.

langchain4j-core/src/main/java/dev/langchain4j/exception/AuthenticationException.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.exception;
+
+public class AuthenticationException extends UnrecoverableChatException {
```
dliubarskyi: I guess we either need to add "Chat" to all the subclasses or remove "Chat" from `UnrecoverableChatException` and `RecoverableChatException` for consistency (`AuthenticationException` sounds like a generic exception, `UnrecoverableChatException` sounds like a chat-focused exception), WDYT?

langchain4j-core/src/main/java/dev/langchain4j/exception/InternalServerException.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.exception;
+
+public class InternalServerException extends UnrecoverableChatException {
```
dliubarskyi: Shouldn't it be recoverable?

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,78 @@
...
+
+            if (rootCause instanceof HttpException httpException) {
+                return mapHttpStatusCode(httpException, httpException.statusCode());
+            }
+
+            if (rootCause instanceof UnresolvedAddressException) {
+                return new UnresolvedModelServerException(rootCause);
+            }
+
+            return e instanceof RuntimeException re ? re : new RuntimeException(e);
```
dliubarskyi: I am wondering if we should wrap `e` into `LangChain4jException` instead of `RuntimeException` here

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -0,0 +1,78 @@
...
+            if (httpStatusCode == 408) {
+                return new TimeoutException(rootException);
+            }
+            if (httpStatusCode == 429) {
+                return new RateLimitException(rootException);
+            }
+            if (httpStatusCode >= 400 && httpStatusCode < 500) {
+                return new InvalidRequestException(rootException);
+            }
+            return rootException instanceof RuntimeException re ? re : new RuntimeException(rootException);
```
dliubarskyi: If we know the error code, should we wrap into `HttpException` instead of `RuntimeException`? (applicable to Jlama)

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
```diff
@@ -190,9 +192,11 @@ public <T> T withRetry(Callable<T> action, int maxAttempts) {
             while (true) {
                 try {
                     return action.call();
+                } catch (UnrecoverableChatException e) {
+                    throw e;
                 } catch (Exception e) {
                     if (attempt >= maxAttempts) {
-                        throw new RuntimeException(e);
+                        throw e instanceof RuntimeException re ? re : new RuntimeException(e);
```
dliubarskyi: same here (LC4jException instead of Runtime)

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/VertexAiGeminiChatModelListenerIT.java
```diff
@@ -37,8 +37,10 @@ protected boolean assertResponseId() {
     @Override
     protected ChatLanguageModel createFailingModel(ChatModelListener listener) {
         return VertexAiGeminiChatModel.builder()
-                .project(System.getenv("GCP_PROJECT_ID"))
-                .location(System.getenv("GCP_LOCATION"))
+                .project("GCP_PROJECT_ID")
```
dliubarskyi: please revert

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
```diff
@@ -190,9 +192,11 @@ public <T> T withRetry(Callable<T> action, int maxAttempts) {
             while (true) {
                 try {
                     return action.call();
+                } catch (UnrecoverableChatException e) {
```
dliubarskyi: BTW, question on the naming: AFAIU "recoverable" means that we can recover from this exception by retrying the operation without the changes to the request? If **_retrying_** is the main reason why we separate exceptions in 2 types (recov and non-recov), should it be then called `RetriableException`/`NonRetriableException` or something like this?

langchain4j/src/test/java/dev/langchain4j/service/AiServiceThrowingExceptionIT.java
```diff
@@ -0,0 +1,160 @@
...
+                .logResponses(true)
+                .build();
+
+        ThrowingService assistant = AiServices.builder(ThrowingService.class)
+                .chatLanguageModel(chatLanguageModel)
+                .build();
+
+        assertThatThrownBy(() -> assistant.chat("hi"))
+                .isExactlyInstanceOf(ModelNotFoundException.class)
+                .hasMessageContaining("The model `gpt-0` does not exist or you do not have access to it.");
```
dliubarskyi, line 158: Nice tests!

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
mariofusco: Agreed.

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
mariofusco: I'm wrapping it into `LangChain4jException`.

mariofusco:
@dliubarskyi thanks for very insightful feedback, please check it again :)

dliubarskyi (approved):
@mariofusco great job, thanks a lot!

PR #2615, opened 2025-02-26
Fix for #2613, MCP Tool execution error handling is not correct

### LINKED ISSUE #2613: [BUG] MCP Tool execution error handling is  not correct
Tool execution errors are not properly returned.

In the class ToolExecutionHelper, there is 2 issues:
* If the reply contains an `error` element, the message and code are swallowed and not returned.
* If the reply contains a `content` element, there is no check done for `isError`. This should be exposed in the reply.

**Describe the bug**
Tool execution errors are not properly returned.

**Log and Stack trace**

**To Reproduce**
Call a tool that reply with the 2 kind of errors.

**Expected behavior**

* If there is an `error` element to have the following returned value: `There was an error executing the tool. Message: actual message, code: actual code`

* If there is an `isError` field to have the following reurned value: `There was an error executing the tool. content text`

- LangChain4j version: 1.0.0.beta1
- LLM(s) used:  Any
- Java version: JDK21

**Additional context**

jfdenise:
## Issue

Fixes #2613

## Change
Handle the error cases. Add new integration test. Update existing ones.

NOTE: I have failing tests in the langchain4j module caused by missing API KEY, I am pretty sure those failures are not caused by my chages.

## General checklist

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
@@ -20,7 +20,15 @@ class ToolExecutionHelper {
         if (result.has("result")) {
             JsonNode resultNode = result.get("result");
             if (resultNode.has("content")) {
-                return extractSuccessfulResult((ArrayNode) resultNode.get("content"));
+                String content = extractSuccessfulResult((ArrayNode) resultNode.get("content"));
+                boolean isError = false;
+                if (resultNode.has("isError")) {
+                    isError = resultNode.get("isError").asBoolean();
+                }
+                if (isError) {
+                    content = String.format(EXECUTION_ERROR_MESSAGE + ". The tool returned: %s", content);
+                }
+                return content;
             } else {
                 log.warn("Result does not contain 'content' element: {}", result);
                 return EXECUTION_ERROR_MESSAGE;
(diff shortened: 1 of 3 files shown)


dliubarskyi:
cc @jmartisk

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
```diff
@@ -20,7 +20,15 @@ static String extractResult(JsonNode result) {
...
             JsonNode resultNode = result.get("result");
             if (resultNode.has("content")) {
-                return extractSuccessfulResult((ArrayNode) resultNode.get("content"));
+                String content = extractSuccessfulResult((ArrayNode) resultNode.get("content"));
+                boolean isError = false;
+                if (resultNode.has("isError")) {
+                    isError = resultNode.get("isError").asBoolean();
+                }
+                if (isError) {
+                    content = String.format(EXECUTION_ERROR_MESSAGE + ". %s", content);
```
jmartisk: I think it would be clearer if we used `:` instead of `.` between these two strings. Or maybe, even better, something like `EXECUTION_ERROR_MESSAGE + ". The tool returned: %s"` 

Maybe we should make this whole thing customizable (add an SPI for applications to plug in a Supplier of error messages) but that will be a separate enhancement I guess

jfdenise: @jmartisk , thank-you. I updated as you suggested.

jmartisk (approved):
Nice!

PR #2628, opened 2025-03-01
update AnthropicTokenUsage.toString to include the cache-related fields

jbellis:
## Change
update AnthropicTokenUsage.toString to include the cache-related fields


dliubarskyi (approved):
@jbellis thank you!

PR #2649, opened 2025-03-06
pgvector store - CREATE EXTENSION can be disabled to avoid permission error in managed PostgreSQL service

kkovarik:
## Issue
We are using this library in Azure ecosystem with Azure PostgreSQL Flexible Server service. We do enable pgvector extension using automation. Since some update, we cannot use this library anymore, as it fails on permissions, basically trying to re-enable extension, which is already installed and enabled.

**ERROR: Only members of "azure_pg_admin" are allowed to use CREATE EXTENSION.**

We do not give this level of permissions to our java applications. Not sure if it is issue in PostgreSQL general, or specific to Azure service, I guess it does not matter much.

## Change
I was considering just removing the CREATE EXTENSION altogether, and just write it into preconditions for this store. If you want to use it, you have to take care of your postgresql extensions.

However, did opt for this "hopefully backward compatible" way, that will be enough for our use case - managed PostgreSQL service in Azure.

## General checklist

- [] I have added unit and/or integration tests for my change - not relevant
- [ ] The tests cover both positive and negative cases - not relevant
- [X] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and

- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features) - not relevant
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable) - not relevant
j/langchain4j-spring) (if applicable)

## Note:
My first PR here, seriously hope I did not break any best practices :) I won't be offended if someone will solve this different way, if it is already planned in some foreseenable future, etc.


kkovarik:
Anyone for review? :) Is there an process set for this?

dliubarskyi:
Oh, I am sorry that this PR got buried!
Rediscovered it when merged https://github.com/langchain4j/langchain4j/pull/4576

PR #2652, opened 2025-03-07
docs: add a note on the required kotlin compiler arguments

SerCeMan:
## Issue

This PR adds a note on the required kotlin compiler configuration for defining tools using [`@Tool`](https://github.com/langchain4j/langchain4j/blob/078c5fbc8912297b7bf67240423f1042d8fa009e/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java#L23). 

**When this parameter is missing:**
Screenshot 2025-03-07 at 2 26 05 pm

**When `javaParameters` is set to true:**
image

I couldn't find any relevant documentation, so I decided to add it. Let me know if this is already mentioned somewhere, or if it should be moved somewhere else.


kpavlov (approved):

dliubarskyi:
@SerCeMan thanks a lot!

PR #2620, opened 2025-02-27
Add ConverseStream support for Amazon Bedrock integrations

tenpigs267:
## Issue
Closes #2470 #2700

## Change
- add a BedrockStreamingChatModel
- implement Observability for BedrockChatModel
- allow adding additionalModelRequestFields (so that reasoning for Claude Sonnet 3.7 can be enabled) through BedrockChatRequestParameters

An AbstractBedrockChatModel allows to reduce code duplication between Streaming and non streaming chat model

## General checklist

- [x] There are no breaking changes
- [x] I have added unit and/or integration tests for my change
- [ ] The tests cover both positive and negative cases
- [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

Issue #2238, opened 2024-12-04
[FEATURE] Add httpClient to bedrock runtime client builder


rndtavares:
I need to use AWS Bedrock runtime client with a proxy setted on the httpClient, it is not possible right now because the BedrockRuntimeClient creation is in a private method, as a workaround and I am extending it on my company's project to add this httpClient. 

**Describe the solution you'd like**
A greater thing can be a possibility to pass the BedrockRuntimeClient itself to AbstractBedrockChatModel, without this, every time that I need to set a new property that is not on the AbstractSharedBedrockChatModel fields, I will need a new PR on langchain4j, by doing this if on some new version AWS put a new property it will automatically be ready for use on langchain4j, but I am not sure how can I pass a BedrockRuntimeClient.

Extending AbstractSharedBedrockChatModel to add a httpClient
<img width="824" alt="image" src="https://github.com/user-attachments/assets/b66b83e0-c9cb-4863-a607-b8c563354434">

So I can use it when creating a BedrockAnthropicMessageChatModel 
<img width="914" alt="image" src="https://github.com/user-attachments/assets/901ec8c8-5399-4c82-a8fd-824cf8e0ad8c">

langchain4j:
@rndtavares thank you, good point!

Would you like to implement this? I can't advise right now with technicalities though.
BTW, there is a work in progress to integrate newer Converse API [here](https://github.com/langchain4j/langchain4j/pull/2132). Perharps the same should be done there?

rndtavares:
Thanks @langchain4j, I will talk to my friend @herbert-beckman

Issue #1704, opened 2024-09-04
[FEATURE] Add timeout support for Amazon Bedrock Anthropic API


JSON-OBJECT:
**Is your feature request related to a problem? Please describe.**
The **BedrockAnthropicMessageChatModel** in LangChain4j currently lacks a **timeout()** method, which is available in other model implementations like **AzureOpenAiChatModel**. This inconsistency makes it difficult to set timeouts for API calls when using the Amazon Bedrock Anthropic API, potentially leading to long-running requests that could impact application performance.

**Describe the solution you'd like**
I propose adding a **timeout()** method to the **BedrockAnthropicMessageChatModel** builder, similar to the one available in **AzureOpenAiChatModel**. This would allow users to set a timeout duration for API calls, enhancing control over request handling. The implementation should be consistent with other LangChain4j model implementations. For example:

```java
BedrockAnthropicMessageChatModel.builder()
    .region(Region.US_EAST_1)
    .model("anthropic.claude-3-5-sonnet-20240620-v1:0")
    .temperature(0.3)
    .topP(0.3f)
    .timeout(Duration.ofSeconds(60))  // New timeout method
    .build()
```

**Describe alternatives you've considered**
1. Using the AWS SDK v2 directly, which provides timeout functionality through the **ApacheHttpClient** configuration. However, this approach requires more boilerplate code and deviates from the simplified LangChain4j API.

2. Implementing a wrapper class around **BedrockAnthropicMessageChatModel** to add timeout functionality. This solution, while feasible, would not be as clean or consistent with other LangChain4j implementations.

**Additional context**
The AWS CLI and SDK v2 both support setting timeouts for Bedrock API calls. Implementing this feature in LangChain4j would bring the library's functionality in line with these official AWS tools, providing a more consistent developer experience across the ecosystem.

For reference, here's how timeouts are set in the AWS CLI:

```bash
aws bedrock-runtime invoke-model \
    --model-id anthropic.claude-3-5-sonnet-20240620-v1:0 \
    --region us-east-1 \
    --body "$(echo '{
        "anthropic_version": "bedrock-2023-05-31",
        "max_tokens": 200000,
        "messages": [
            {
                "role": "user",
                "content": "Hello!"
            }
        ],
        "temperature": 0.3,
        "top_p": 0.3
    }' | base64 -w 0)" \
    --content-type "application/json" \
    --cli-read-timeout 60 \
    --cli-connect-timeout 60 \
    output.json
```

[...]

PR #2675, opened 2025-03-12
Amazon Bedrock: deprecate old models using invoke API

### LINKED ISSUE #2606: Amazon Bedrock: deprecate old models using invoke API

dliubarskyi:
## Issue
Closes #2606

## Change
Deprecated old Amazon Bedrock models that used invoke API:
- `BedrockAI21LabsChatModel`
- `BedrockAnthropicCompletionChatModel`
- `BedrockAnthropicMessageChatModel`
- `BedrockCohereChatModel`
- `BedrockLlamaChatModel`
- `BedrockMistralAiChatModel`
- `BedrockStabilityAIChatModel`
- `BedrockTitanChatModel`

Please use `BedrockChatModel` instead.

## General checklist

### DIFF
--- a/docs/docs/integrations/language-models/amazon-bedrock.md
+++ b/docs/docs/integrations/language-models/amazon-bedrock.md
@@ -55,11 +55,11 @@ ChatLanguageModel model = BedrockChatModel.builder()
 ## ChatLanguageModel using InvokeAPI
 
 ### AI21 Models
-- `BedrockAI21LabsChatModel`
+- `BedrockAI21LabsChatModel` (deprecated, please use `BedrockChatModel`)
 
 ### Anthropic Models
-- `BedrockAnthropicMessageChatModel`: supports new Messages API
-- `BedrockAnthropicCompletionChatModel`: supports old Text Completions API
+- `BedrockAnthropicMessageChatModel`: (deprecated, please use `BedrockChatModel`) supports new Messages API
+- `BedrockAnthropicCompletionChatModel`: (deprecated, please use `BedrockChatModel`) supports old Text Completions API
 - `BedrockAnthropicStreamingChatModel`
 
 Example:
@@ -70,19 +70,19 @@ ChatLanguageModel model = BedrockAnthropicMessageChatModel.builder()
 ```
 
 ### Cohere Models
-- `BedrockCohereChatModel`
+- `BedrockCohereChatModel` (deprecated, please use `BedrockChatModel`)
 
 ### Meta Llama Models
-- `BedrockLlamaChatModel`
+- `BedrockLlamaChatModel` (deprecated, please use `BedrockChatModel`)
 
 ### Mistral Models
-- `BedrockMistralAiChatModel`
+- `BedrockMistralAiChatModel` (deprecated, please use `BedrockChatModel`)
 
 ### Titan Models
-- `BedrockTitanChatModel`
+- `BedrockTitanChatModel` (deprecated, please use `BedrockChatModel`)
 - `BedrockTitanEmbeddingModel`
 
 ### Examples
 
-- [BedrockChatModelExample](https://github.com/langchain4j/langchain4j-examples/blob/main/bedrock-examples/src/main/java/invoke/BedrockChatModelExample.java)
+- [BedrockChatModelExample](https://github.com/langchain4j/langchain4j-examples/blob/main/bedrock-examples/src/main/java/converse/BedrockChatModelExample.java)
 - [BedrockStreamingChatModelExample](https://github.com/langchain4j/langchain4j-examples/blob/main/bedrock-examples/src/main/java/invoke/BedrockStreamingChatModelExample.java)
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockAI21LabsChatModel.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockAI21LabsChatModel.java
@@ -8,6 +8,10 @@ import lombok.experimental.SuperBuilder;
 import java.util.HashMap;
 import java.util.Map;
 
+/**
+ * @deprecated please use {@link BedrockChatModel} instead
+ */
+@Deprecated(forRemoval = true, since = "1.0.0-beta2")
 @Getter
 @SuperBuilder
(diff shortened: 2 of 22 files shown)


JensMander:
Hi @dliubarskyi,

did you kill the Listeners from BedrockChatModel in beta2 or how do I have to pass them since now? Lately I created the model like this:
`
BedrockAnthropicMessageChatModel 
.builder().temperature(0.20f).maxTokens(1000).region(Region.EU_CENTRAL_1).model(model).maxRetries(1).listeners(listeners).credentialsProvider(awsCredentialsProvider).build();
`
Same code for beta2:
`
BedrockChatModel.builder().client(BedrockRuntimeClient.builder().region(Region.EU_CENTRAL_1).credentialsProvider(awsCredentialsProvider).build()).defaultRequestParameters(ChatRequestParameters.builder().temperature(0.20).maxOutputTokens(4000).build()).region(Region.EU_CENTRAL_1).modelId(model).maxRetries(1).build();
`
Much some stuff to do but that's not my point. It seems to me like there is no way to pass the listeners to the model anymore. In beta1 (and before) there was a class attribute in AbstractSharedBedrockChatModel holding the listeners but this class doesn't seem to be in the class hierarchy of BedrockChatModel so there is no implementation for that feature. How to deal with it in beta2 (and later down the road)?

Best regards

Sascha

dliubarskyi:
@JensMander thank you for the heads up, apparently listeners are not yet supported in `BedrockChatModel`. We will implement them before removing deprecated `BedrockAnthropicMessageChatModel`. cc @tenpigs267 could you please take this over?

dliubarskyi:
created https://github.com/langchain4j/langchain4j/issues/2700

JensMander:
Great to hear :) Thanks guys

tenpigs267:
Hi,
Observability is already implemented in #2620. This PR add a BedrockStreamingChatModel with observability, so while refactoring it took just a little code to add observability to BedrockChatModel also.

pguedes17:
Hi @dliubarskyi. complementing the @JensMander listeners request,  I noticed that new API does not support to customize credentials provider .. any plans to add this behavior before remove the  `BedrockAnthropicMessageChatModel`?

E.g:
`BedrockAnthropicMessageChatModel myModel = BedrockAnthropicMessageChatModel
            .builder()
            .credentialsProvider(StaticCredentialsProvider.create(AwsBasicCredentials.create(credentials.getAccessKey(), credentials.getSecretKey())))
.build()`

tenpigs267:
Following ticket #1704 (socket timeout), the possibility to provide a custom BedrockRuntimeClient was added and this was then mimicked in the BedrockChatModel that use the converseAPI.

Since providing both a custom BedrockRuntimeClient and a CredentialProvider could lead to conflicts, the decision was made to allow customization only via the BedrockRuntimeClient (see [discussion here](https://github.com/langchain4j/langchain4j/pull/2404#discussion_r1930522615)).

In [this PR](https://github.com/langchain4j/langchain4j/pull/2501), the issue of embedding apache-client lib is discussed (and the same concern would apply to NettyClient for streaming).

If I summarize, it seems there are two main reasons to provide a BedrockRuntimeClient:
1. To override the CredentialProvider.
2. To increase the socket timeout of the default client.

So I see two possible approaches:
- Either we allow developers to provide their own BedrockRuntimeClient, and it’s up to them to manage its instantiation, including the CredentialProvider and socket timeout.
- Or we allow to provide a CredentialProvider, and we embed the apache-client and NettyClient libraries directly in the langchain4j-bedrock JAR to instantiate the HttpClient in the constructors of Bedrock(Streaming)ChatModel, cause that will allow to align the socket timeout with the timeout duration passed to the constructor. In this case, we ?may want? to remove the ability to provide a custom BedrockRuntimeClient to avoid potential conflicts.

Am I missing anything in this analysis? Do you see any other reason to allow to provide a custom BedrockRuntimeClient?

I’m leaning towards the second option: allowing a CredentialProvider, bundling the ApacheHttpClient and NettyClient lib directly in the langchain4j-bedrock JAR

What about you ?

pguedes17:
@tenpigs267 I agree with you. I also noticed that while it's currently possible to customize the BedrockRuntimeClient, it's not possible to customize the CredentialProvider directly.

I'm also leaning toward the second option — allowing the use of a CredentialProvider and bundling the ApacheHttpClient and NettyClient libs directly in the langchain4j-bedrock JAR. I believe this approach would cover the majority of use cases and help avoid potential conflicts.

tenpigs267:
#2238 needs a custom bedrockClient because of a proxy.

tenpigs267:
a third approach would be to mimic AWS SDK design, decoupling the BedrockRuntimeClient (and consequently the HttpClient) instanciation from the core BedrockChatModel code : 
- BedrockChatModelBuilder can take a custom BedrockRuntimeClient or a BedrockClientConfig containing the region, the timeout and the credentialProvidder
- if no BedrockRuntimeClient is provided we would have : 

```
public interface BedrockHttpClientFactory {
    BedrockRuntimeClient createClient(BedrockClientConfig config);
}

public class BedrockChatModel {
    
    public BedrockChatModel(Builder builder) {
        ...

        if (isNull(builder.client){
            BedrockClientFactory factory = ServiceLoader.load(BedrockClientFactory.class).findFirst()
                .orElseThrow(() -> new RuntimeException("No BedrockClientFactory implementation found"));
            this.client = factory.createClient(builder.bedrockClientConfig());
        }
    }
}
```
And we add a langchain4j-bedrock-runtime-client module (or two to separate streaming and non streaming usage) that embed apache-client (and netty for streaming usage) and have the BedrockClientFactory like this : 
```
public class DefaultBedrockClientFactory implements BedrockClientFactory {
    @Override
    public BedrockRuntimeClient createClient(BedrockClientConfig config) {
        // Use apache or netty client based on config
        return BedrockRuntimeClient.builder()
            .credentialsProvider(config.getCredentialProvider())
            .httpClientBuilder(ApacheHttpClient.builder()
                .socketTimeout(config.getSocketTimeout())
                // other configs
            )
            .build();
    }
}
```
and that solution preserve the classpath of langchainj user while adding some complexity.
@dliubarskyi WDYT?

tenpigs267:
> a third approach would be to mimic AWS SDK design, decoupling the BedrockRuntimeClient (and consequently the HttpClient) instanciation from the core BedrockChatModel code :
> ...

implemented here : https://github.com/langchain4j/langchain4j/commit/1ef03de28c88b95e8549c4ea9a233801bb39d902

Issue #2447, opened 2025-01-24
[BUG] LanguageModelQueryRouter : Failed to route query due to NumberFormatException


Yayotron:
`d.l.r.q.router.LanguageModelQueryRouter  : Failed to route query` due to `NumberFormatException`

Similar to closed issue https://github.com/langchain4j/langchain4j/issues/588

**Describe the bug**
I have an agent configured with two `ContentRetriever`: `EmbeddingStoreContentRetriever` which fetches data from a PSQL database, and `WebSearchEngine` which uses `GoogleCustomWebSearchEngine` implementation to search online.

My `RetrievalAugmentor` is configured in the following way:

```java
                return DefaultRetrievalAugmentor.builder()
                .queryTransformer(new CompressingQueryTransformer(
                        ollamaChatModel, new PromptTemplate(transformerTemplate)))
                .queryRouter(new LanguageModelQueryRouter(
                        ollamaChatModel,
                        contentRetrieverUsage))
                .build();
```

Most of the time, whenever I ask my agent for a complex query it 

**Log and Stack trace**
```
WARN 33472 --- [presenter] [           main] d.l.r.q.router.LanguageModelQueryRouter  : Failed to route query '

Query: Find resources that explain how to implement data-driven decision making in project management using machine learning techniques.

In this query, the user wants to find resources (likely articles or tutorials) that detail the process of implementing a data-driven approach to project management using machine learning. The user is likely looking for information on specific steps, best practices, and case studies related to this topic.

Data needed: Resources (articles, tutorials, etc.) explaining the implementation of data-driven decision making in project management using machine learning techniques.

Data available: None specified, but it can be assumed that any resources indexed by search engines could potentially be relevant.

Missing data: Specific sources or websites that contain the desired information are not provided.

Possible ways to improve the data: Adding keywords such as "step-by-step guide", "tutorial", "case study", and specific project management methodologies (Agile, Waterfall, Scrum, etc.) could help refine the search and find more relevant results.'

java.lang.NumberFormatException: For input string: "2
Explanation: For this query"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67) ~[na:na]
	at java.base/java.lang.Integer.parseInt(Integer.java:662) ~[na:na]
	at java.base/java.lang.Integer.parseInt(Integer.java:778) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197) ~[na:na]
[...]

PR #2676, opened 2025-03-12
Handle bug while processing response from LLM in LanguageModelQueryRouter

Yayotron:
## Issue
#2447

## Change
* Changed prompt used in LanguageModelQueryRouter to answer in a more code-friendly language that we can parse.
* Added parsing to handle thinking process coming from reasoning models like deepseek-r1 and gemini.
* Fixed compiler warnings

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/rag/query/router/LanguageModelQueryRouter.java
```diff
@@ -1,5 +1,8 @@
 package dev.langchain4j.rag.query.router;
 
+import com.google.gson.Gson;
```
dliubarskyi: please do not use Gson, use Jackson instead

dliubarskyi (commented):
@Yayotron thanks a lot, but I am not sure about this specific solution. I was considering using structured outputs (which is supported by many LLMs nowadays) as a default choice, before falling back to json mode or pure prompting

PR #2691, opened 2025-03-13
MCP prompt support + some MCP resource handling improvements

### LINKED ISSUE #2304: [FEATURE] MCP: support resources and prompts
see  https://spec.modelcontextprotocol.io/specification/server/prompts/ and https://spec.modelcontextprotocol.io/specification/server/resources/

jmartisk:
## Issue

Closes #2304 

## Change
Support for retrieving prompts from an MCP server. Plus some improvements to MCP resource handling (using Jackson for parsing where applicable, and improved error handling)

## General checklist


docs/docs/tutorials/mcp.md
```diff
@@ -110,6 +110,17 @@ This returns a list of `ResourceContents` objects (there may be more resource co
...
 string. For a binary blob, use `resourceContents.asBlob()` to access the actual data, for text, use `resourceContents.asText()`.
 
+## Prompts
+
+To obtain a list of [MCP prompts](https://modelcontextprotocol.io/docs/concepts/prompts)
+from the server, use `client.listPrompts()`. This method returns a List of `PromptRef`. A `PromptRef`
+contains information about the name and arguments of the prompt.
+
+To render the actual contents of a prompt, use `client.getPrompt(name, arguments)`. A rendered prompt can contain one to many
+messages and these are represented as `PromptMessage` objects. Each `PromptMessage` contains the role of the message (`user`, `assistant`,...)
```
dliubarskyi: Can we please avoid `PromptMessage` and map directly to `ChatMessage`? Otherwise users will need to map it manually to use this with LC4j APIs. For example, `UserMessage` has a list of `Content` which can be one of `TextContent`, `ImageContent`, etc

dliubarskyi (commented):
@jmartisk thanks a lot!

docs/docs/tutorials/mcp.md
jmartisk: I've added `PromptMessage.toChatMessage()` and mentioned its limitations in the documentation.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/EmbeddedResource.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+public record EmbeddedResource(ResourceContents resource) implements PromptContent {
+    @Override
+    public Type type() {
+        return Type.RESOURCE;
+    }
+
+    @Override
+    public PromptTextContent asText() {
+        throw new IllegalArgumentException("Not textual content");
```
dliubarskyi: What is the pro of having `asText`, `asResource` and `asImage` and throw an exception in case it is of wrong type?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/EmbeddedResource.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public PromptImageContent asImage() {
+        throw new IllegalArgumentException("Not an image content");
+    }
+
+    @Override
+    public Content toContent() {
+        if (resource.type().equals(ResourceContents.Type.TEXT)) {
+            return TextFileContent.from(
+                    resource.asText().text(), resource.asText().mimeType());
```
dliubarskyi: Do you know what can are the possible values of `mimeType`? I am wondering if it makes sense to use `TextFileContent` instead of `TextContent`.

BTW, is `resource.asText().text() ` a base64-encoded text?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpErrorHelper.java
```diff
@@ -0,0 +1,14 @@
+package dev.langchain4j.mcp.client;
+
+import com.fasterxml.jackson.databind.JsonNode;
+
+public class McpErrorHelper {
```
dliubarskyi:
```suggestion
class McpErrorHelper {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/PromptArg.java
```diff
@@ -0,0 +1,6 @@
+package dev.langchain4j.mcp.client;
+
+/**
+ * The 'PromptArgument' object from the MCP protocol schema.
+ */
+public record PromptArg(String name, String description, boolean required) {}
```
dliubarskyi:
```suggestion
public record PromptArgument(String name, String description, boolean required) {}
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpException.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.mcp.client;
+
+public class McpException extends RuntimeException {
+
+    private final McpError mcpError;
```
dliubarskyi: Is there a plan to use `McpError` outside of `McpException`? This extra layer of nesting seems to not be required, so perhaps McpException can contain `code` and `message` fields?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/PromptContent.java
```diff
@@ -0,0 +1,39 @@
...
+@JsonSubTypes({
+    @JsonSubTypes.Type(value = PromptTextContent.class, name = "text"),
+    @JsonSubTypes.Type(value = EmbeddedResource.class, name = "resource"),
+    @JsonSubTypes.Type(value = PromptImageContent.class, name = "image")
+})
+public sealed interface PromptContent permits PromptTextContent, EmbeddedResource, PromptImageContent {
+
+    @JsonProperty("type")
+    default String getType() {
+        return type().toString().toLowerCase();
```
dliubarskyi:
```suggestion
        return type().toString().toLowerCase(Locale.ROOT);
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/PromptRef.java
```diff
@@ -0,0 +1,10 @@
+package dev.langchain4j.mcp.client;
+
+import java.util.List;
+
+/**
+ * The 'Prompt' object from the MCP protocol schema.
```
dliubarskyi: I have noticed that sometimes name of the class is different from the naming in MCP, is it done on purpose? What about keeping 1:1 mapping for consistency?

dliubarskyi (commented):
@jmartisk thanks a lot!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/EmbeddedResource.java
jmartisk: I think I copied this style from the MCP server extension (https://github.com/quarkiverse/quarkus-mcp-server/blob/main/core/runtime/src/main/java/io/quarkiverse/mcp/server/Content.java) but right, it's not how langchain4j generally does it, so I'll remove it.

jmartisk: Hm actually we already have the same for `dev.langchain4j.mcp.client.ResourceContents`, so, shall I remove it too? Or keep it? I kinda like this style as it allows you to easily "cast" the result when you're pretty sure what type it is. But the rest of langchain4j doesn't use this style.. So I guess it's your call, either I keep both, or remove both...

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/EmbeddedResource.java
jmartisk: Yeah I don't know what the exact difference is, but given that it's a file, I thought `TextFileContent` might be the right choice. 

The `mimeType` can be anything AFAIK (https://github.com/modelcontextprotocol/specification/blob/246e5ce6680b592fb0111b43ba238115a54ceff0/schema/draft/schema.json#L1918)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/PromptArg.java
jmartisk: Yeah good point, I copied the name from the MCP server extension (https://github.com/quarkiverse/quarkus-mcp-server/blob/main/core/runtime/src/main/java/io/quarkiverse/mcp/server/PromptArg.java) but that is actually an annotation, so probably not much sense to have the name synchronized.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpException.java
jmartisk: I was kinda thinking that maybe the spec will, in the future, allow returning multiple business errors (which are represented as `McpError`), and then we may need this nesting... 

And I tend to think that representing a remote business error as a regular class is generally a good idea. I followed a similar principle that I have in the GraphQL client library, where a `GraphQLError` is a regular object and a `GraphQLClientException` may contain one or more of them. 

But of course, depends on you.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/PromptRef.java
jmartisk: In this case, I think just naming the class `Prompt` would be confusing - that is a very general term. `PromptRef` tells the reader that it's just a reference, not the actual contents.

jmartisk: But of course, if you prefer, we can keep the names exactly as they are in the JSON schema

jmartisk: `PromptTextContent` also seems a better class name than `TextContent`, which is a class that already exists in the core API. Hence I added the javadoc comments to make it clear what it maps to...

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/EmbeddedResource.java
dliubarskyi: I have no strong preference, was just curious about the pros and cons

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/EmbeddedResource.java
dliubarskyi: In my understanding the difference is that `TextContent` is just a string but `TextFileContent` is **_base64-encoded_** string + mime type. The only LLM that consumes `TextFileContent` is Gemini and if it is not in base64 it will fail. From the specification it seems like `TextResourceContents` is not encoded, so perhaps `TextContent` is more appropriate

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpException.java
dliubarskyi: I would simplify it TBH, we can always create `McpError` if it will be required

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/PromptRef.java
dliubarskyi: What about following the naming of MCP but add an `Mcp` prefix to not ? `McpPrompt`, `McpTextContent`, etc

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/EmbeddedResource.java
jmartisk: Ok I'll remove it, we can decide later and apply a similar style  consistently across the whole API

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/EmbeddedResource.java
jmartisk: Ok, I'm changing it

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/PromptRef.java
jmartisk: Ok that sounds quite reasonable

jmartisk:
I've hopefully resolved all your current comments now

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -184,7 +189,33 @@ public ResourceResponse readResource(String uri) {
...
-    public List<ResourceTemplateRef> listResourceTemplates() {
+    public List<McpPrompt> listPrompts() {
+        if (promptRefs.get() == null) {
+            obtainPromptList();
+        }
+        return promptRefs.get();
+    }
+
+    @Override
+    public MgpGetPromptResult getPrompt(String name, Map<String, Object> arguments) {
```
dliubarskyi, line 200:
```suggestion
    public McpGetPromptResult getPrompt(String name, Map<String, Object> arguments) {
```

dliubarskyi (approved):
@jmartisk thanks a lot!

PR #2435, opened 2025-01-22
Avoid caching chat memories in AiServiceContext

mariofusco:
Since chat memories are only stateless decorators around a `ChatStore` it is useless to cache them in the `AiServiceContext` and can be recreated at each invocation. 

The only point of attention is when the chat memories doesn't have an explicitly store configured and then implicitly use the `InMemoryChatMemoryStore`. In this case, before this change, each and every chat memory created its own store, in a wasteful way since this store used a concurrent map only to hold one single entry. If we don't cache anymore the chat memories, but they are recreated at each iteration, this was of course not only wasteful but also wrong since also the store is recreated empty every time. This change makes evident that the lifecycle of the `InMemoryChatMemoryStore` is tightly related with the one of the `AiServiceContext` from where the chat memory are created. For this reason the `AiServiceContext` has to inject the store into each newly created memory and use of the builder pattern for them makes this implementation a bit more convoluted.

PR #2692, opened 2025-03-13
Fix memory leak due to ChatMemory caching

### LINKED ISSUE #1889: [FEATURE] When using a ChatMemoryProvider, allow a way to delegate chatMemory() to always call the provider.
Currently, when a ChatMemoryProvider is configured, the AiServiceContext caches the ChatMemory in it's internal map. But in a multi-user app we want to evict unused memories after a certain interval. A simple solution would be to provide a way so that the AiServicesContext always ask the provider.

Code in AiServicesContext

```
public ChatMemory chatMemory(Object memoryId) {
        return chatMemories.computeIfAbsent(memoryId, ignored -> chatMemoryProvider.get(memoryId));
 }

```

I'm guessing the current code was probably done in case the provider loads the chat from a store like a database. So caching it after it's loaded makes sense.

mariofusco:
Closes https://github.com/langchain4j/langchain4j/issues/1889

This fixes the memory leak reported [here](https://github.com/langchain4j/langchain4j/issues/1889) in a much better, less invasive way than the [other fix](https://github.com/langchain4j/langchain4j/pull/2435) that I implemented. Also it has the advantage of not breaking backward compatibility in any way.


langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/SingleSlotChatMemoryStore.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+/**
+ * Implementation of {@link ChatMemoryStore} that stores state of {@link dev.langchain4j.memory.ChatMemory} (chat messages) in-memory.
+ * <p>
+ * This storage mechanism is transient and does not persist data across application restarts.
+ */
+public class SingleSlotChatMemoryStore implements ChatMemoryStore {
```
mariofusco: As also noted in the former proposed fix, for in memory use cases this store doesn't need to hold a map of ChatMemories. The former `InMemoryChatStore` had that map that was always filled with at most one single entry.

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
```diff
@@ -0,0 +1,40 @@
+package dev.langchain4j.memory.chat;
+
+import dev.langchain4j.memory.ChatMemory;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+public class ChatMemories {
```
mariofusco: I encapsulated the chat memory related concern into this class, in a similar way as I did for the `ToolService` one introduced when I refactored the tools management.

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
```diff
@@ -0,0 +1,40 @@
...
+        if (memoryId == DEFAULT) {
+            return defaultChatMemories;
+        }
+        return chatMemories.computeIfAbsent(memoryId, ignored -> createChatMemory(memoryId));
+    }
+
+    private ChatMemory createChatMemory(Object memoryId) {
+        ChatMemory chatMemory = chatMemoryProvider.get(memoryId);
+        if (chatMemory instanceof RemovableChatMemory rcm) {
+            rcm.onChatMemoryRemove(chatMemories::remove);
```
mariofusco: If a `ChatMemory` implements `RemovableChatMemory` it register a function to also remove the memory from the `chatMemories` map.

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
```diff
@@ -0,0 +1,40 @@
...
+public class ChatMemories {
+
+    public static final String DEFAULT = "default";
+
+    private ChatMemory defaultChatMemories;
+    private Map<Object, ChatMemory> chatMemories;
+    private ChatMemoryProvider chatMemoryProvider;
+
+    public ChatMemories(ChatMemoryProvider chatMemoryProvider) {
+        this.chatMemories = Collections.synchronizedMap(new WeakHashMap<>());
```
mariofusco: Uses a `WeakHashMap` to avoid retaining chat memories when their id is no longer reachable. I'm not entirely sure this is a good idea and I'm open to discuss this and change it into a plain `ConcurrentHashMap` if this could be safer.

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
```diff
@@ -112,17 +115,25 @@ private static void ensureCapacity(List<ChatMessage> messages, int maxMessages)
     @Override
     public void clear() {
```
mariofusco: This method represents the only issue that in my opinion is still open. If you call it, this `ChatMemory` will also remove itself from the map holding it in the `ChatMemories` class, but nobody seems to call it at the moment and I don't see how to expose the possibility to call to users.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
```diff
@@ -335,4 +335,66 @@ public void deleteMessages(Object memoryId) {
...
+        verify(chatLanguageModel)
+                .chat(ChatRequest.builder()
+                        .messages(
+                                userMessage(firstMessageFromSecondUser),
+                                aiMessage(firstAiResponseToSecondUser),
+                                userMessage(secondMessageFromSecondUser))
+                        .build());
+
+        verify(chatLanguageModel, times(4)).supportedCapabilities();
+
```
mariofusco, line 387: Here I would like to have a way to get an handle to one of the 2 `MessageWindowChatMemory` created so far and call `clear()` on it, to check that it has been really removed from the `ChatMemories`. Alternatively the AI service should not only have a method to chat with it passing a given memory id, but also a second method to allow the removal of the memory with that id, that internally would call `ChatMemory.clear()`. None of these things seem to be possible at the moment. See https://github.com/langchain4j/langchain4j/pull/2692/files#r1993807755

@dliubarskyi do you have any suggestion for this?

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/InMemoryChatMemoryStore.java
dliubarskyi, line 1: > Also it has the advantage of not breaking backward compatibility in any way.

Removing `InMemoryChatMemoryStore` breaks backward compatibility. Or was it removed by mistake?

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
dliubarskyi: Nice! Perhaps name it `ChatMemoryService` to keep the parity with `ToolService`?

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/SingleSlotChatMemoryStore.java
dliubarskyi: I am not sure about the name. I doubt user cares about the internal impl (e.g., whethere it is single- or multiple- slot), I feel like the "in-memory" part of the name is more important, WDYT?

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
dliubarskyi: If it is going to be used by Quarkus extension, perhaps leave a comment in this class to not forget aobut it when making changes in the future? Otherwise it would be better to make it package-private

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.WeakHashMap;
+
+public class ChatMemories {
+
+    public static final String DEFAULT = "default";
+
+    private ChatMemory defaultChatMemories;
```
dliubarskyi:
```suggestion
    private ChatMemory defaultChatMemory;
```
nit

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
```diff
@@ -0,0 +1,40 @@
...
+    private Map<Object, ChatMemory> chatMemories;
+    private ChatMemoryProvider chatMemoryProvider;
+
+    public ChatMemories(ChatMemoryProvider chatMemoryProvider) {
+        this.chatMemories = Collections.synchronizedMap(new WeakHashMap<>());
+        this.chatMemoryProvider = chatMemoryProvider;
```
dliubarskyi:
```suggestion
        this.chatMemoryProvider = ensureNotNull(chatMemoryProvider, "chatMemoryProvider");
```
Otherwise for `AiServices.builder(...).chatMemoryProvider(null)` case it will create `ChatMemories` instance and behave as if AI Service has chat memory (see `AiServiceContext.hasChatMemory()`)

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    public ChatMemories(ChatMemory chatMemory) {
+        defaultChatMemories = chatMemory;
```
dliubarskyi:
```suggestion
        defaultChatMemories = ensureNotNull(chatMemory, "chatMemory");
```

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
```diff
@@ -33,20 +34,22 @@
...
-public class MessageWindowChatMemory implements ChatMemory {
+public class MessageWindowChatMemory implements ChatMemory, RemovableChatMemory {
 
     private static final Logger log = LoggerFactory.getLogger(MessageWindowChatMemory.class);
 
     private final Object id;
     private final Integer maxMessages;
     private final ChatMemoryStore store;
 
+    private Consumer<Object> memoryCleaner;
```
dliubarskyi:
```suggestion
    private Consumer<Object> memoryRemover;
```

To keep consistency with `RemovableChatMemory`, WDYT?

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
dliubarskyi: Could you please elaborate? What is the issue exactly? That it is not called but it should (to make your solution work)?

langchain4j/src/main/java/dev/langchain4j/memory/chat/TokenWindowChatMemory.java
```diff
@@ -46,11 +47,13 @@ public class TokenWindowChatMemory implements ChatMemory {
     private final Tokenizer tokenizer;
     private final ChatMemoryStore store;
 
+    private Consumer<Object> memoryCleaner;
```
dliubarskyi:
```suggestion
    private Consumer<Object> memoryRemover;
```

langchain4j/src/main/java/dev/langchain4j/memory/chat/RemovableChatMemory.java
```diff
@@ -0,0 +1,8 @@
+package dev.langchain4j.memory.chat;
+
+import java.util.function.Consumer;
+
+interface RemovableChatMemory {
```
dliubarskyi:
```suggestion
interface RemovalAwareChatMemory {
```
WDYT?

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
dliubarskyi: As far as I understand, using `WeakHashMap` is the whole point of this PR? So using `ConcurrentHashMap` is not an option? Or am I misunderstanding?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
dliubarskyi: With the current implementation, `ChatMemoryProvider` above can store all created memories, so you can then iterate over them and call `clear()`. Would that work?
But I am not sure `clear()` should remove the memory from the AI service though. `clear()` is intended to clear the messages from the memory, not the memory itself.

I do not have a clear answer yet (no pun intended), but I have a feeling that instead of AI Service owning and caching memories on it's side it should instead ask for a `ChatMemory` instance from `ChatMemoryProvider` **_each time_** (instead of once currently) AI service needs it. This way user will have full control of what to do with memories (cache them or not, when to remove them, etc). WDYT?

dliubarskyi (commented):
@mariofusco thanks a lot!

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
dliubarskyi: After second thought I am afraid that removing the caching logic from AI Service might be a tricky change for users to adapt. So maybe this is something for version 2.

dliubarskyi: Getting back to the [original issue](https://github.com/langchain4j/langchain4j/issues/1889).
User wants to "evict unused memories after a certain interval".

They can (partially) do it with the following setup:
```java
    ChatMemoryStore store = new ChatMemoryStore() {

           private final Map<Object, List<ChatMessage>> messagesByMemoryId = new ConcurrentHashMap<>();

           @Override
           public List<ChatMessage> getMessages(Object memoryId) {
               return this.messagesByMemoryId.computeIfAbsent(memoryId, ignored -> new ArrayList<>());
           }

           @Override
           public void updateMessages(Object memoryId, List<ChatMessage> messages) {
               this.messagesByMemoryId.put(memoryId, messages);
           }

           @Override
           public void deleteMessages(Object memoryId) {
               this.messagesByMemoryId.remove(memoryId);
           }
       };

        ChatMemoryProvider chatMemoryProvider = memoryId -> MessageWindowChatMemory.builder()
                .id(memoryId)
                .maxMessages(10)
                .chatMemoryStore(store)
                .build();

        ChatWithSeparateMemoryForEachUser chatWithMemory = AiServices.builder(ChatWithSeparateMemoryForEachUser.class)
                .chatMemoryProvider(chatMemoryProvider)
                .build();
```

This way they have full access to `messagesByMemoryId`, so they can evict messages when required.

Yes, `AiServiceContext.chatMemories` will still reference the `ChatMemory` instance (that will reference the `ChatMemoryStore` that does not hold messages for this `ChatMemory` any more), but this should not take up much space (I assume that `ChatMessage`s take up most of the space). So perhaps converting `AiServiceContext.chatMemories` into a `WeakHashMap` (this PR) or exposing a way to explicitly evict `ChatMemory` from `AiServiceContext.chatMemories` could be a solution to this. WDYT?

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/SingleSlotChatMemoryStore.java
mariofusco: Maybe, I can change the name, but anyway I believe this should be part only of the internal implementation and not used directly by end users. This is simply the default store when no other store implementation is configured. I need to check if it is possible, but probably we should move this class and make it package private?

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
mariofusco: Ok, I will rename this, and yes I expect that this should be used by the quarkus extension.

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
mariofusco: The issue is that this should be called by the user when this chat memory, with a given id, is no longer necessary (e.g.: the corresponding http session is closed or times out), but at the moment they cannot do so, because the `ChatMemory` are unreachable, not exposed by any public API. 

In other words, the `ChatMemory` with the given id is created implicitly when a user chats with the LLM also passing a memory id, but that the `ChatMemory` object with that id is totally unaccessible for users, so it also impossible to call this `clear` method on it. Am I making sense?

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/InMemoryChatMemoryStore.java
mariofusco: It has been removed intentionally. It this true that this breaks backward compatibility, so I could reintroducing it, possibly flagging it as deprecated for removal, but in reality this shouldn't be used directly by users for the same reason that I wrote here https://github.com/langchain4j/langchain4j/pull/2692#discussion_r1999346885

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemories.java
mariofusco: Not completely. We are obliged to use the `WeakHashMap` only because the current API doesn't allow the user to say "remove the memory with id 12, I don't need it anymore". In essence yes, we need to keep the `WeakHashMap` if we don't want to (or we cannot) let users to control the `ChatMemory` life cycle.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
mariofusco: I'd personally prefer to expose a way to explicitly evict `ChatMemory` from `AiServiceContext.chatMemories` that in my understanding is what that `clear()` method on the `ChatMemory` itself is supposed to do. If the intention of the `clear()` is keeping the memory but empty it, we could add an `evict()` method instead, but to be honest I don't see why we should do it. I mean I don't see in which circumstances you may want to clear the `ChatMemory` but keeping it, how this could be useful? 

That said I still don't see how to expose the `ChatMemory` (and then its clear/evict method) to end users. How a user can retrieve from the AI service the `ChatMemory` instance with a given id?

Using the `WeakHashMap`, as it does now, is the plan B in case we don't want to (or we cannot) let users to control the `ChatMemory` life cycle.

mariofusco:
@dliubarskyi I brought in all your suggestions and added back the `InMemoryChatMemoryStore`, but as I said this is only to keep backward compatibility and I flagged it as deprecated for removal. I don't expect that there isn't any valid circumstance where users may want to use it. For the same reason I refactored the new `SingleSlotChatMemoryStore` and made it package private: it is simply the default in memory `ChatMemoryStore` and I don't expect users to use that class directly.

I still have some doubts on giving the possibility to user to explicitly evict a `ChatMemory`. It is not clear to me if we want to allow this or not, but I believe that maybe this pull request could be merged as it is and we could eventually review it at a later stage. Let me know if this is ok for you or you have any other suggestions.

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/SingleSlotChatMemoryStore.java
dliubarskyi: Perhaps, but as you can see it is used directly quite a lot: https://github.com/search?q=InMemoryChatMemoryStore&type=code

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
dliubarskyi: > I'd personally prefer to expose a way to explicitly evict ChatMemory from AiServiceContext.chatMemories

Agreed

> That said I still don't see how to expose the ChatMemory (and then its clear/evict method) to end users. How a user can retrieve from the AI service the ChatMemory instance with a given id?

What about something like:

```java
interface ChatMemoryAware { // this is LC4j's interface

    Map<Object, ChatMemory> chatMemories();

    void removeChatMemory(Object id);

    // the methods above is just an example, we need to think again which one(s) will be useful
}

interface Assistant extends ChatMemoryAware { // if AI Service extends ChatMemoryAware, we can detect it in `DefaultAiServices.build()` and implement it via proxy

}

Assistant assistant = ...
assistant.removeChatMemory("123")
```

So if user wants to have access to chat memories, they could just add `extends ChatMemoryAware` to their AI Service definition. WDYT?

---

Maybe it should be named not `ChatMemoryAware` but something more generic to accomodate for future additions related to other aspects like tools or RAG, etc.

dliubarskyi: cc @rwankar would `ChatMemoryAware` option above work for you?

rwankar: I haven't read through the whole thing but I suppose the idea is to let the user provide the implementation of the ChatMemory where it's up to the end user to clear the messages/evict the whole memory etc? e.g I could write implement this with a wrapper around Guava's Cache here. Wouldn't you need a getMemory(id) method in the interface as well?

rwankar: Also, how will WeakHashMap help if we use GUID strings as keys? Imagine a multi-user AI chat environment. We create a GUID on the client (browser) and pass it with the user prompt. On the backend we want to retain the chat memory per GUID. So on each POST request we will set the memoryId to the GUID. The GUID string is not referenced from anywhere and we don't want it evicted just because the user's chat is complete. We would like to retain it up to say 30 min of inactivity.

We could however keep the GUIDs referenced in a Cache with a timeout and achieve the same effect.

rwankar: BTW, while we're on the topic of ChatMemory, I'm currently evaluating Spring AI and it has a pretty flexible memory management mechanism via the Advisor interface. This allows you flexibility like the ability to change the messages in a request without putting them into the ChatMemory. e.g we use this to pass Hints for tools. So it would be good for LC4j to offer as much flexibility as possible.

mariofusco: > Also, how will WeakHashMap help if we use GUID strings as keys? Imagine a multi-user AI chat environment. We create a GUID on the client (browser) and pass it with the user prompt. On the backend we want to retain the chat memory per GUID. So on each POST request we will set the memoryId to the GUID. The GUID string is not referenced from anywhere and we don't want it evicted just because the user's chat is complete. We would like to retain it up to say 30 min of inactivity.
> 
> We could however keep the GUIDs referenced in a Cache with a timeout and achieve the same effect.

That's an interesting point, I thought that UID should be retained on the server side in order to be reachable and usable but indeed this may not be always the case. The timeout is another possibility, but in that case it will add even more complexity and we should make it configurable. 

For now I will try to implement the `ChatMemoryAware` solution suggested by @dliubarskyi. Giving users the possiblity to decide when to evict memories, if possible, is the best option in my opinion.

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryService.java
```diff
@@ -17,19 +18,20 @@ public class ChatMemoryService {
     private ChatMemoryProvider chatMemoryProvider;
 
     public ChatMemoryService(ChatMemoryProvider chatMemoryProvider) {
-        this.chatMemories = Collections.synchronizedMap(new WeakHashMap<>());
+        this.chatMemories = new ConcurrentHashMap<>();
```
mariofusco, line 21: Since the user is now allowed to explicitly clear the `ChatMemory`, I turned this back into a `ConcurrentHashMap`.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
```diff
@@ -383,6 +384,16 @@ void should_keep_separate_chat_memory_for_each_user_with_default_store() {
 
         verify(chatLanguageModel, times(4)).supportedCapabilities();
 
+        chatWithMemory.getChatMemory(secondMemoryId).clear();
```
mariofusco: The `clear()` method also evicts the `ChatMemory`. 

@dliubarskyi Do we want an explicit evict method instead? Is there a situation where we may want to clear the memory but keep it there empty instead of evicting it?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
```diff
@@ -242,7 +243,7 @@ void should_keep_chat_memory_and_add_new_system_message() {
     }
 
 
-    interface ChatWithSeparateMemoryForEachUser {
+    interface ChatWithSeparateMemoryForEachUser extends ChatMemoryAccess {
```
mariofusco, line 247: Extending the `ChatMemoryAccess` interface (I'm not strongly opinionated on the name) allows to access the `ChatMemory` of the AI service.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
```diff
@@ -383,6 +384,16 @@ void should_keep_separate_chat_memory_for_each_user_with_default_store() {
 
         verify(chatLanguageModel, times(4)).supportedCapabilities();
 
+        chatWithMemory.getChatMemory(secondMemoryId).clear();
+
+        // clear removes the chat memory, so now it is null
+        assertThat(chatWithMemory.getChatMemory(secondMemoryId)).isNull();
```
mariofusco, line 392: The memory with the given id has been evicted so a second attempt to access it returns `null`. Is that ok?

mariofusco:
@dliubarskyi I added the `ChatMemoryAccess` interface and tried to explain how it works, please review and let me know if you have any comment or suggestion.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
dliubarskyi: > I haven't read through the whole thing but I suppose the idea is to let the user provide the implementation of the ChatMemory where it's up to the end user to clear the messages/evict the whole memory etc? e.g I could write implement this with a wrapper around Guava's Cache here. 

It is currently (already) possible for the user to implement `ChatMemory` in any way they want.
The problem here is that AI Service caches `ChatMemory` instances and user has no access to this "cache". Over time this "cache" will grow and will lead to the memory leak. So `ChatMemoryAware` should allow user to get access to the cache and evict `ChatMemory` instances which are not needed any more. Would that work for you?

> Wouldn't you need a getMemory(id) method in the interface as well?

This can be added, but the focus here was on removing `ChatMemory` from the `Map<Object, ChatMemory>`

> This allows you flexibility like the ability to change the messages in a request without putting them into the ChatMemory. e.g we use this to pass Hints for tools. So it would be good for LC4j to offer as much flexibility as possible.

OK, noted. Although this can already be achieved like so:

```java
public class ChatModelWrapper implements ChatLanguageModel {

    private final ChatLanguageModel delegate;

    public ChatModelWrapper(ChatLanguageModel delegate) {
        this.delegate = delegate;
    }

    @Override
    public ChatResponse chat(ChatRequest chatRequest) {
        ChatRequest modifiedChatRequest = ChatRequest.builder()
                .messages(chatRequest.messages()) // messages can be manipulated here
                .parameters(chatRequest.parameters())
                .build();
        return delegate.chat(modifiedChatRequest);
    }
}
```

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/InMemoryChatMemoryStore.java
```diff
@@ -12,6 +12,7 @@
  * <p>
  * This storage mechanism is transient and does not persist data across application restarts.
  */
+@Deprecated(forRemoval = true)
```
dliubarskyi: I am not sure we should deprecate or remove it - it is still a valid choice if user wants to keep chat memories in-memory in one place. A single `InMemoryChatMemoryStore` can be created and shared for all `ChatMemory` instances

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryAccess.java
```diff
@@ -0,0 +1,17 @@
...
+ */
+public interface ChatMemoryAccess {
+
+    /**
+     * Returns the {@link ChatMemory} with the given id for this AI service or null if such memory doesn't exist.
+     *
+     * @param memoryId The ID of the chat memory.
+     * @return The {@link ChatMemory} with the given memoryId or null if such memory doesn't exist.
+     */
+    ChatMemory getChatMemory(Object memoryId);
```
dliubarskyi, line 16: Don't we want to have an `evictChatMemory(id)` method as well?

langchain4j/src/main/java/dev/langchain4j/memory/chat/SingleSlotChatMemoryStore.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public void deleteMessages(Object memoryId) {
+        checkMemoryId(memoryId);
+        this.messages = new ArrayList<>();
+    }
+
+    private void checkMemoryId(Object memoryId) {
+        if (this.memoryId != memoryId) {
+            throw new IllegalStateException();
```
dliubarskyi: Please add a suitable error message

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
dliubarskyi: I would not tie "clear" and "evict" operations together, as it limits the flexibility. Or do you see any benefit?

Imagine the situation when `ChatMemory` should be removed from the cache (to reduce memory consumption) but messages should be kept inside the (persistent) `ChatMemory` to continue the conversation in the future.

dliubarskyi: Now that there is a `ChatMemoryAccess`, user can easily trigger eviction of the `ChatMemory` from the "cache" map, so `RemovalAwareChatMemory` seems to not be required any more?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
dliubarskyi: Seems to be correct

dliubarskyi (commented):
@mariofusco thank you!

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/InMemoryChatMemoryStore.java
mariofusco: Ok for me. I still don't see in which circumstances somebody may want to use this explicitly and anyway for how this is wired you still will have an instance of this for each memory, meaning a map with no more than an entry. I may missing some use cases though, I will remove the deprecation.

dliubarskyi: > for how this is wired you still will have an instance of this for each memory, meaning a map with no more than an entry

only if you use the default store. One can create a single `InMemoryChatMemoryStore` and set is as a store when creating each `ChatMemory`

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryAccess.java
mariofusco: I have now split the `clear()` and `evict()` methods. In my opinion the way how it should work is that you get the `ChatMemory` with this method and then you do what you want on it, like reading messages, clearing or evicting it, or do anything else it will be possible to do on the `ChatMemory` in future. I'm not strongly opinionated on this but at the moment I don't see a valid reason to add method on this interface to do something that you could do directly on the `ChatMemory` instance.

langchain4j/src/main/java/dev/langchain4j/memory/chat/SingleSlotChatMemoryStore.java
mariofusco: Done.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
mariofusco: The `RemovalAwareChatMemory` interface has package private access and it is for internal use only. It is used to register a callback on the `ChatMemory` to remove it from the map when you call evict() on it.

mariofusco:
@dliubarskyi I added a proper `evict()` method, separated from the `clear()` one and addressed other suggestions (or replied to them). Please give it one more look, thanks.

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryAccess.java
dliubarskyi: This way `ChatMemory` becomes aware of AI Service and eviction, I am not sure it should be a concern of chat memory to be evicted from anywhere, WDYT?

mariofusco: Ok for me, I'm not strongly opinionated on this, I will remove the `evict()` method from the `ChatMemory` and and it to this interface. This could actually simplify a bit the code.

dliubarskyi: Yes this should simplify quite a bit

mariofusco: @dliubarskyi I did what you suggested and indeed I'm happier with the design now because it allowed me to also simplify the code and remove the `RemovalAwareChatMemory` interface. Thank you and please give it one more look.

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/ChatMemoryStore.java
```diff
@@ -12,7 +12,7 @@
  * Allows for flexibility in terms of where and how chat memory is stored.
  * <br>
  * <br>
- * Currently, the only implementation available is {@link InMemoryChatMemoryStore}.
+ * Currently, the only implementation available is {@link SingleSlotChatMemoryStore}.
```
dliubarskyi:
```suggestion
 * Currently, the only implementations available are {@link InMemoryChatMemoryStore} and {@link SingleSlotChatMemoryStore}.
```

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryAccess.java
```diff
@@ -0,0 +1,25 @@
...
+     * @param memoryId The id of the chat memory.
+     * @return The {@link ChatMemory} with the given memoryId or null if such memory doesn't exist.
+     */
+    ChatMemory getChatMemory(Object memoryId);
+
+    /**
+     * Evicts the {@link ChatMemory} with the given id.
+     *
+     * @param memoryId The id of the chat memory to be evicted.
+     * @return true if it existed a {@link ChatMemory} with the given id and it was sucessfully evicted, false otherwise.
```
dliubarskyi:
```suggestion
     * @return true if {@link ChatMemory} with the given id existed, and it was successfully evicted, false otherwise.
```

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryAccess.java
```diff
@@ -0,0 +1,25 @@
+package dev.langchain4j.memory.chat;
+
+import dev.langchain4j.memory.ChatMemory;
+
+/**
+ * Allow to access the {@link ChatMemory} of any AI service extending it.
+ */
+public interface ChatMemoryAccess {
```
dliubarskyi, line 8: Could you please document this feature in https://docs.langchain4j.dev/tutorials/ai-services#chat-memory section?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
```diff
@@ -335,4 +336,78 @@ public void deleteMessages(Object memoryId) {
...
+        verify(chatLanguageModel, times(4)).supportedCapabilities();
+
+        assertThat(chatWithMemory.getChatMemory(secondMemoryId)).isNotNull();
+        assertThat(chatWithMemory.evictChatMemory(secondMemoryId)).isTrue();
+
+        // clear removes the chat memory, so now it is null
+        assertThat(chatWithMemory.getChatMemory(secondMemoryId)).isNull();
+        assertThat(chatWithMemory.evictChatMemory(secondMemoryId)).isFalse();
+
+        // the memory has been cleared so it cannot remember the name of the user
```
dliubarskyi:
```suggestion
        // the memory has been evicted (and also lost, as it is not persistent), so it cannot remember the name of the user
```

dliubarskyi: It would be great to have a test where `InMemoryChatMemoryStore` (or a persistent one) is used and verify that after evicting the memory and chatting again (`secondMessageFromSecondUser`) the memory still exists(`assertThat(responseToSecondUserAfterMemoryClean).contains("Francine");`)

dliubarskyi (commented):
@mariofusco thanks a lot, looks good!
Just a few nitpicks left and documentation 🙏

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryAccess.java
mariofusco: Done, please give a look.

langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/ChatMemoryStore.java
```diff
@@ -12,7 +12,7 @@
  * Allows for flexibility in terms of where and how chat memory is stored.
  * <br>
  * <br>
- * Currently, the only implementation available is {@link InMemoryChatMemoryStore}.
+ * Currently, the only implementations available are {@link InMemoryChatMemoryStore} and {@link SingleSlotChatMemoryStore}.
```
dliubarskyi:
```suggestion
 * Currently, the only implementation available is {@link InMemoryChatMemoryStore}.
```
This does not compile as `SingleSlotChatMemoryStore` is in anoter module. Since `SingleSlotChatMemoryStore` is "internals", I guess it is ok to not mention it.

langchain4j/src/main/java/dev/langchain4j/memory/chat/SingleSlotChatMemoryStore.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public void deleteMessages(Object memoryId) {
+        checkMemoryId(memoryId);
+        this.messages = new ArrayList<>();
+    }
+
+    private void checkMemoryId(Object memoryId) {
+        if (this.memoryId != memoryId) {
```
dliubarskyi: In theory `Object memoryId` can be anything (also a complex object). I guess it should check for equqlity using `.equals()`?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -124,18 +126,25 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
                             return method.invoke(this, args);
                         }
 
+                        if (method.getDeclaringClass() == ChatMemoryAccess.class) {
+                            return switch (method.getName()) {
+                                case "getChatMemory" -> context.hasChatMemory() ? context.chatMemoryService.getChatMemory(args[0]) : null;
```
dliubarskyi, line 137: nit: Shouldn't we throw `IllegalStateException` or something when `ChatMemoryAccess` method is called but there is no memory? Perhaps we should validate it earlier, when creating AI Service (e.g., fail If AI service extends `ChatMemoryAccess` but there is no memory configured).

langchain4j/src/test/java/dev/langchain4j/memory/chat/SingleSlotChatMemoryStoreTest.java
```diff
@@ -1,15 +1,15 @@
-package dev.langchain4j.store.memory.chat;
+package dev.langchain4j.memory.chat;
 
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.UserMessage;
 import java.util.Arrays;
 import org.assertj.core.api.WithAssertions;
 import org.junit.jupiter.api.Test;
 
-class InMemoryChatMemoryStoreTest implements WithAssertions {
```
dliubarskyi: It would be nice to keep the `InMemoryChatMemoryStoreTest` and add a separate `SingleSlotChatMemoryStoreTest`

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithChatMemoryIT.java
dliubarskyi: @rwankar

> Currently, when a ChatMemoryProvider is configured, the AiServiceContext caches the ChatMemory in it's internal map. But in a multi-user app we want to evict unused memories after a certain interval.

I hope that current solution (this PR) will help to resolve the issue. Please check https://github.com/langchain4j/langchain4j/pull/2692/files#diff-442b5e004db50e406f349445bf1dc2fcd11f192febf367564a33e5c06da28a98

dliubarskyi (commented):
@mariofusco thanks a lot!

langchain4j/src/test/java/dev/langchain4j/memory/chat/SingleSlotChatMemoryStoreTest.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: Good point, done.

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryAccess.java
```diff
@@ -0,0 +1,25 @@
+package dev.langchain4j.memory.chat;
```
dliubarskyi, line 1: I guess this should be moved to the `dev.langchain4j.service` package as `ChatMemoryAccess` can be used only with AI Service

langchain4j/src/main/java/dev/langchain4j/memory/chat/ChatMemoryService.java
```diff
@@ -0,0 +1,40 @@
+package dev.langchain4j.memory.chat;
```
dliubarskyi, line 1: I guess this should be moved to the `dev.langchain4j.service.memory` package as `ChatMemoryService` is supposed to be used only with AI Service (similarly to `ToolService`)

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: nit: now that we have a validation, there is no need in checking `context.hasChatMemory()` (for both methods), it is always true at this point

dliubarskyi (approved):
@mariofusco thanks a lot, looks great! There are 3 tiny nitpicks which I can do myself if you don't mind

PR #2693, opened 2025-03-13
Add OpenAI Official module to the BOM

jdubois:
I totally missed that the new langchain4j-openai-official module needed to be added to the BOM.


dliubarskyi (approved):
@jdubois thanks!

PR #2769, opened 2025-03-26
Health checks for MCP clients

### LINKED ISSUE #2766: [FEATURE] [MCP] Way to check MCP client health
We could add a `mcpClient.isHealthy()` method or something like that to enable health checks (check that the MCP servers are reachable)

jmartisk:
## Issue

Closes https://github.com/langchain4j/langchain4j/issues/2766

## Change
Adds the `McpClient.checkHealth()` method that throws an exception if the server isn't reachable or isn't responding to pings within a reasonable timeout. I didn't include a dependency on MicroProfile Health as this probably isn't suitable here, but any runtime that integrates it (Quarkus) may want to wrap an invocation of this method into a built-in MicroProfile Health check.

## General checklist

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
@@ -47,4 +47,11 @@ public interface McpClient extends AutoCloseable {
      * Render the contents of a prompt.
      */
     McpGetPromptResult getPrompt(String name, Map<String, Object> arguments);
+
+    /**
+     * Performs a health check that returns normally if the MCP server is reachable and
+     * properly responding to ping requests. If this method throws an exception,
+     * the health of this MCP client is considered degraded.
+     */
+    void checkHealth();
 }
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -22,6 +22,7 @@ import dev.langchain4j.mcp.client.protocol.McpListPromptsRequest;
 import dev.langchain4j.mcp.client.protocol.McpListResourceTemplatesRequest;
 import dev.langchain4j.mcp.client.protocol.McpListResourcesRequest;
 import dev.langchain4j.mcp.client.protocol.McpListToolsRequest;
+import dev.langchain4j.mcp.client.protocol.McpPingRequest;
 import dev.langchain4j.mcp.client.protocol.McpReadResourceRequest;
 import dev.langchain4j.mcp.client.transport.McpOperationHandler;
 import dev.langchain4j.mcp.client.transport.McpTransport;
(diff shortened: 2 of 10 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/PingRequest.java
```diff
@@ -0,0 +1,13 @@
+package dev.langchain4j.mcp.client.protocol;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+public class PingRequest extends McpClientMessage {
```
dliubarskyi:
```suggestion
public class McpPingRequest extends McpClientMessage {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
```diff
@@ -47,4 +47,10 @@ public interface McpClient extends AutoCloseable {
      * Render the contents of a prompt.
      */
     McpGetPromptResult getPrompt(String name, Map<String, Object> arguments);
+
+    /**
+     * Performs a health check that returns true if the MCP server is reachable and
```
dliubarskyi: Javadoc does not correspond to the implementation.

How do you think this method will be used? Perhaps it should indeed return `true/false` instead of throwing exception? Exception-driven logic does not feel right.

dliubarskyi (commented):
@jmartisk thank you!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
jmartisk: We also need to pass an error message to say what went wrong. Just `true/false` isn't enough information. 

Ideally we could return the full `HealthCheckResponse` (https://github.com/microprofile/microprofile-health/blob/main/api/src/main/java/org/eclipse/microprofile/health/HealthCheckResponse.java) that carries all the proper information, but that would mean introducing a MP Health dependency - not sure if you want this.

So the approach I thought of was that we let it throw an exception and then the wrapping framework (Quarkus, or something that the user writes himself) would transform its message into a `HealthCheckResponse` (or something else if using something other than MP Health).

But if you don't mind being bound to a specific Health framework, I can redo this by adding MP Health and letting this method return a `HealthCheckResponse`.

jmartisk: A full integration with MP Health would require setting up an HTTP endpoint though, so I don't think we should do it (rather Quarkus, or whatever other wrapping runtime would do it). We could, however, introduce our own slimmed down version of `HealthCheckResponse` that would get transformed to the regular one by Quarkus?! That way we wouldn't depend on MP Health here. WDYT?

dliubarskyi: Both "our own slimmed down version of HealthCheckResponse" and exception seem fine, whatever you prefer

jmartisk:
I've now fixed the javadoc as requested, and added a configurable ping timeout.

dliubarskyi (approved):
@jmartisk thanks a lot!

PR #2772, opened 2025-03-26
Allow tools to directly return the raw result

mariofusco:
This is a review, rebase and reimplementation of https://github.com/langchain4j/langchain4j/pull/2291

I'm putting it here, in a different pull request because it changes a bit the logic of the original implementation, but in my opinion this is more correct. In fact it executes all the tools needed by the AI service and returns the tool result instead of the LLM one if and only if all those tools executions used tools annotated with raw return.

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
@@ -46,5 +46,5 @@ public @interface Tool {
      *
      * @return whether to return the result directly
      */
-    boolean returnDirectly() default false;
+    boolean directReturn() default false;
 }
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
@@ -1,10 +1,10 @@
 package dev.langchain4j.agent.tool;
 
-import static java.lang.annotation.ElementType.METHOD;
-import static java.lang.annotation.RetentionPolicy.RUNTIME;
-
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
  * Java methods annotated with {@code @Tool} are considered tools/functions that language model can execute/call.
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
@@ -1,10 +1,10 @@
 package dev.langchain4j.agent.tool;
 
-import java.lang.annotation.Retention;
-import java.lang.annotation.Target;
-
 import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
 
 /**
  * Java methods annotated with {@code @Tool} are considered tools/functions that language model can execute/call.
@@ -36,4 +36,15 @@ public @interface Tool {
      * @return description of the tool.
      */
     String[] value() default "";
+
+    /**
+     * Indicates whether the tool's String return value should be returned directly to the user.
+     * Only applicable if used within an AiService and if the return type is 'Result<String>'
+     * If true, the result will be returned directly without further LLM processing.
+     * If false or if return type is not Result<String>, normal processing continues.
+     * If true, the tool cannot be chained with other tools as the result will not be returned to the llm.
+     *
+     * @return whether to return the result directly
+     */
+    boolean returnDirectly() default false;
 }
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
@@ -46,5 +46,5 @@ public @interface Tool {
      *
      * @return whether to return the result directly
      */
-    boolean directReturn() default false;
+    boolean rawReturn() default false;
 }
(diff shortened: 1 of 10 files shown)


langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithNewToolsIT.java
```diff
@@ -732,6 +731,244 @@ protected void should_execute_tool_with_list_of_POJOs_parameter() {
...
+                verify(model, times(3)).chat(chatRequestCaptor.capture());
+
+                var toolSpecifications = chatRequestCaptor.getValue().toolSpecifications();
+                assertThat(toolSpecifications).hasSize(2);
+            }
+        }
+    }
+
+    @Test
+    protected void should_execute_tools_all_return_raw_true_multiple_called_at_a_time() {
```
mariofusco, line 886: This test fails with simpler model server like ollama, but only because the model is not smart enough to execute the tools invocation sequence as expected. Is it ok to ignore this test for specific providers?

langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
```diff
@@ -40,6 +41,12 @@ public DefaultToolExecutor(Object object, ToolExecutionRequest toolExecutionRequ
         this.methodToInvoke = this.originalMethod;
     }
 
+    @Override
+    public boolean returnRaw() {
```
dliubarskyi, line 45: Copying my [comment](https://github.com/langchain4j/langchain4j/pull/2291#discussion_r1928654856) from the original PR:

> I doubt ToolExecutor shoud know about what to do with the result returned from it.
> AiServiceContext seems like a better place to keep this information. Please see AiServices.tools(List<Object> objectsWithTools), there we have access to the @Tool annotation.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
```diff
@@ -36,4 +36,15 @@
      * @return description of the tool.
      */
     String[] value() default "";
+
+    /**
+     * Indicates whether the tool's return value should be returned directly to the user.
+     * Only applicable if used within an AiService and if the return type is 'Result<String>'.
+     * If true, the result will be returned directly without further LLM processing.
+     * If false or if return type is not Result<String>, normal processing continues.
```
dliubarskyi, line 44: Copying my [comment](https://github.com/langchain4j/langchain4j/pull/2291#discussion_r1895519015) from the original PR:

> Could you please add an explicit validation that checks that if there is at lest a single tool with directReturn=true, then return type of such AI Service method should be Result<String>? If it is not, we should fail

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithNewToolsIT.java
```diff
@@ -732,6 +731,244 @@ protected void should_execute_tool_with_list_of_POJOs_parameter() {
...
+                    .tools(firstTool, secondTool)
+                    .build();
+
+            var text = "First add 2 and 3, then multiply the result by 4";
+
+            // when
+            var response = assistant.chat(text);
+
+            // then
+            assertThat(response.content()).isEqualTo("20");
```
dliubarskyi, line 865: Since both tools have `returnRaw = true`, shouldn't we stop the loop and return `Result` with `content=null` and 2 `ToolExecution`s?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithNewToolsIT.java
```diff
@@ -732,6 +731,244 @@ protected void should_execute_tool_with_list_of_POJOs_parameter() {
...
+
+        static JsonSchemaElement EXPECTED_SCHEMA = JsonObjectSchema.builder()
+                .addIntegerProperty("arg0")
+                .addIntegerProperty("arg1")
+                .required("arg0", "arg1")
+                .build();
+    }
+
+    @Test
+    protected void should_execute_tool_with_primitive_parameters_return_raw() {
```
dliubarskyi, line 756: In every new test here, could you please add one more invocation of AI service at the end (of the test) to make sure that memory is not corrupted and AI Service can be used further?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithNewToolsIT.java
dliubarskyi: Yes, feel free to `@Override` it in specific providers and annotate with `@Disabled` with a comment why it is disabled

dliubarskyi (commented):
@mariofusco thank you!

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
dliubarskyi: Just in case: if we go with keeping `Result.content=null` in case of all called tools are `returnRaw=true`, we do not necessarily have to enforce `Result<String>`, we can accept `Result<?>`

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithNewToolsIT.java
dliubarskyi: Sorry, I meant 1 `ToolExecution` (first).

This is a very complicated feature...

mariofusco: No, this is the change in the logic that I mentioned in the comment to this PR: the LLM invokes all the tools it needs to fulfill the request and if and only if **all** the tools it used are flagged with raw return then the service returns the value provided by the latest tool in the chain instead of letting the LLM to further manipulate it.

I believe that this behaviour is more correct, or at least this is the one that I'd expect, but I understand that this is very questionable, so I'm open to discuss this.

mariofusco:
Replaced by https://github.com/langchain4j/langchain4j/pull/3069

PR #2780, opened 2025-03-28
Prevent NPE when parsing a McpLogMessage without logger

### LINKED ISSUE #2778: [BUG] [MCP] NPE when server sends a log message without the 'logger' parameter
https://github.com/modelcontextprotocol/specification/blob/2025-03-26/schema/2025-03-26/schema.json#L1146

The 'logger' parameter is optional, but the client fails if the server doesn't provide it.

mariofusco:
Fixes https://github.com/langchain4j/langchain4j/issues/2778


dliubarskyi (approved):
@mariofusco thanks a lot!

PR #2783, opened 2025-03-31
fix: add moderation validation to AiService using StreamingLanguageModel and add AiService moderation test

### LINKED ISSUE #2779: [BUG] AiServices' moderation models isn't taken into account when using a streaming chat model
When using `AiServices` with a `moderationModel()`, the moderation happens without a problem:

```java
ModeratedChat moderatedChat = AiServices.builder(ModeratedChat.class)
    .chatLanguageModel(
        VertexAiGeminiChatModel.builder()
            .project(System.getenv("GCP_PROJECT_ID"))
            .modelName("gemini-2.0-flash-001")
            .build())
    .moderationModel(new ModerationModel() {
        @Override
        public Response<Moderation> moderate(final String text) {
            return Response.from(Moderation.flagged(text));
        }

        @Override
        public Response<Moderation> moderate(final List<ChatMessage> messages) {
            return moderate(messages.getLast());
        }
    })
    .build();

moderatedChat.chat("Hello world!");
```

However, when using a streaming chat model:

```java
ModeratedChat moderatedChat = AiServices.builder(ModeratedChat.class)
    .streamingChatLanguageModel(
        VertexAiGeminiStreamingChatModel.builder()
            .project(System.getenv("GCP_PROJECT_ID"))
            .modelName("gemini-2.0-flash-001")
            .build())
    .moderationModel(new ModerationModel() {
        @Override
        public Response<Moderation> moderate(final String text) {
            return Response.from(Moderation.flagged(text));
        }

        @Override
        public Response<Moderation> moderate(final List<ChatMessage> messages) {
            return moderate(messages.getLast());
        }
    })
    .build();

moderatedChat.chat("Hello world!");
```

Then the moderation isn't taken into account.

Gucvii:
## Issue

Closes #2779

## Change

Add moderation validation to AiService using StreamingLanguageModel.
Add AiService moderation test

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -179,6 +179,8 @@ class DefaultAiServices<T> extends AiServices<T> {
                                 context.toolService.executionContext(memoryId, userMessage);
 
                         if (streaming) {
+                            verifyModerationIfNeeded(moderationFuture);
+
                             TokenStream tokenStream = new AiServiceTokenStream(
                                     messages,
                                     toolExecutionContext.toolSpecifications(),
@@ -186,7 +188,7 @@ class DefaultAiServices<T> extends AiServices<T> {
                                     augmentationResult != null ? augmentationResult.contents() : null,
                                     context,
                                     memoryId);
-                            // TODO moderation
+
                             if (returnType == TokenStream.class) {
                                 return tokenStream;
                             } else {
--- a/langchain4j/src/test/java/dev/langchain4j/service/AiServicesModerationModelTest.java
+++ b/langchain4j/src/test/java/dev/langchain4j/service/AiServicesModerationModelTest.java
@@ -1,0 +1,102 @@
+package dev.langchain4j.service;
+
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.UserMessage;
+import dev.langchain4j.model.chat.ChatLanguageModel;
+import dev.langchain4j.model.chat.StreamingChatLanguageModel;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.response.ChatResponse;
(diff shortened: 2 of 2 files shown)


langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -179,14 +179,16 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
                                 context.toolService.executionContext(memoryId, userMessage);
 
                         if (streaming) {
+                            verifyModerationIfNeeded(moderationFuture);
```
dliubarskyi, line 182: This will block untill moderation is complete, but for sync model there is no blocking

dliubarskyi (commented):
@Gucvii thanks!

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
Gucvii: Dear author, thank you for your response and guidance. I understand it will block the process. However, I have a few questions regarding the design of the moderation feature. Specifically, I'm curious whether it's intended to moderate both UserMessage and AiMessage, or solely UserMessage.

I've observed that the current implementation primarily focuses on moderating UserMessage. Consequently, I didn't incorporate asynchronous design in my approach. In the ChatLanguageModel, the verifyModerationIfNeeded function is executed after the Response is generated, which seems like an excellent choice for performance and compliance. If the verification fails, the user won't receive the Response.

However, in the StreamingChatLanguageModel, if verification isn't checked before the start() method, the user might receive non-compliant content. This raises concerns about potential risks.

Could you please clarify the intended scope of the moderation and provide insights on handling moderation in the streaming model? Thank you.

Gucvii: Or will the result look like 1. Generate continuously, interrupt generation and report an error if moderation completes and determine if the message needs to be collapsed, or 2. Wait until all generation is complete, add a validation near onCompleteResponse?

mr-cloud:
why this PR closed without merge or clarification?

PR #2807, opened 2025-04-03
Quarkus extension fixes

mariofusco:


langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
```diff
@@ -71,14 +72,21 @@ public static Map<String, Object> toRecord(Neo4jEmbeddingStore store, int idx, L
...
         }
 
         row.put(EMBEDDINGS_ROW_KEY, Values.value(embedding.vector()));
         row.put(PROPS, properties);
         return row;
     }
 
+    private static Object toNeo4jValue(Object obj) {
+        if (obj instanceof UUID uuid) {
+            obj = uuid.toString();
```
dliubarskyi, line 85: what is the reason for this special handling?

dliubarskyi (commented):
@mariofusco thanks!

langchain4j-neo4j/src/main/java/dev/langchain4j/store/embedding/neo4j/Neo4jEmbeddingUtils.java
mariofusco: We removed `Metadata::toMap` that returned a `Map<String, String>` and now we have only `asMap` that instead returns `Map<String, Object>`. This means that the metadata values were always converted into a String before and now this isn't happening anymore. 

There is a test in the Quarkus extension that specifically has an `UUID` as value in the metadata, but this is not a type supported by Neo4j `Values::value` transformation so that method throws an exception. I understand that this fix is very specific for this use case, and agree that probably we should do something more general, but I couldn't find a better way. Do you have any suggestion?

dliubarskyi: Oh, thanks for the heads up! My bad, I missed this during the migration.
We have an `EmbeddingStoreIT`.`should_add_embedding_with_segment_with_metadata` that tests for all types, including the `UUID`. But for some reason Neo4j IT does not extend it like all other stores do...

dliubarskyi (approved):

PR #2819, opened 2025-04-07
Allow to mix static and dynamic tools

mariofusco:


langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -88,26 +72,19 @@ public void tools(Collection<Object> objectsWithTools) {
...
-            return new ToolExecutionContext(this.toolSpecifications, this.toolExecutors);
+            return this.toolSpecifications.isEmpty() ?
+                    new ToolExecutionContext(null, null) :
+                    new ToolExecutionContext(this.toolSpecifications, this.toolExecutors);
         }
 
-        List<ToolSpecification> toolsSpecs = new ArrayList<>();
-        Map<String, ToolExecutor> toolExecs = new HashMap<>();
+        List<ToolSpecification> toolsSpecs = new ArrayList<>(this.toolSpecifications);
+        Map<String, ToolExecutor> toolExecs = new HashMap<>(this.toolExecutors);
```
dliubarskyi, line 87: Could you please add a validation that there are no duplicates in tool names between static and dynamic tools?

dliubarskyi (commented):
@mariofusco thank you!

Could you please update the documentation? I guess https://docs.langchain4j.dev/tutorials/tools#specifying-tools-dynamically section scould specify the behaviour when both static and dynamic tools are configured.

It would also be great to keep the checklist from the PR template as it helps to not forget/oversee inportant points.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: Good point, but what should we do in case of a duplication? I'd say that the static one should take precedence, but I agree that this is a bit arbitrary. Any other suggestion?

mariofusco:
I added a check for duplicated tools, always giving precedence to statically defined ones, and added another test to verify this behavior.  I also added a sentence in the docs to clarify this.

@dliubarskyi please give this another look.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
dliubarskyi: When there is a conflict between 2 static tools, we just fail with exception (see `ToolService.tools`). I guess we should do the same here?

mariofusco: > When there is a conflict between 2 static tools, we just fail with exception (see `ToolService.tools`). I guess we should do the same here?

Done. Please check again.

dliubarskyi (approved):
@mariofusco thank you!

PR #2822, opened 2025-04-08
Allow to configure custom Azure HttpClientProvider (#2799)

### LINKED ISSUE #2799: [FEATURE] AzureOpenAIChatModel: Make it possible to override the default (Netty) HTTP client with something else
Hi there,

we are facing a problem with the Azure OpenAI chat model, that it throws a "channel not registered to an event loop" error when receiving a 429 response status.

The underlying issue is actually caused by a bug in the Azure SDK itself ([link](https://github.com/Azure/azure-sdk-for-java/issues/43583)) in combination with its Netty transport. Hence we would like to swap it out against something else, i.e. the JDK or OkHttp transports.

However, right now this is not possible because:
a) the AzureOpenAiChatModel.builder does not provide a property for the Azure Core HTTP client
b) the HTTP client is hard wired to Netty here: https://github.com/langchain4j/langchain4j/blob/917834994b475da28771c453c86c7d68bf78124c/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java#L123

It would be great if you could provide a setter for this in order to work-around the Azure issue. Thanks in advance!

PS: If it helps, I would also be willing to provide a PR for this change.

philippn:
## Issue
Fixes #2799

## Change
This PR basically adds the Azure `HttpClientProvider` as a argument to `dev.langchain4j.model.azure.InternalAzureOpenAiHelper#setupAsyncClient` and `dev.langchain4j.model.azure.InternalAzureOpenAiHelper#setupSyncClient`

Then, all the models were changed to add this property to the builder that corresponds to each model.

After these changes, the Spotless.Java Maven-Plugin came in and I had to run spotless:apply to get the build working. This caused some formatting changes in the files affected by the changes. If this is a problem, please let me know how to proceed with this.

## General checklist


jdubois (approved):
Thank you for adding support for the HttpClientProvider!
The spotless reformatting is super annoying, but I guess this is mandatory & not because of anything you did, so let's keep it.

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -120,13 +171,15 @@ private static OpenAIClientBuilder setupOpenAIClientBuilder(String endpoint, Str
...
-        HttpClient httpClient = new NettyAsyncHttpClientProvider().createInstance(clientOptions);
+        httpClientProvider = getOrDefault(httpClientProvider, NettyAsyncHttpClientProvider::new);
+        HttpClient httpClient = httpClientProvider.createInstance(clientOptions);
 
         HttpLogOptions httpLogOptions = new HttpLogOptions();
         if (logRequestsAndResponses) {
             httpLogOptions.setLogLevel(HttpLogDetailLevel.BODY_AND_HEADERS);
         }
 
+        maxRetries = getOrDefault(maxRetries, 3);
```
dliubarskyi:
```suggestion
        maxRetries = getOrDefault(maxRetries, 2);
```

dliubarskyi (dismissed):
@philippn thank you!

langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
```diff
@@ -1,5 +1,19 @@
 package dev.langchain4j.model.azure;
 
+import static dev.langchain4j.data.message.AiMessage.aiMessage;
+import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.Utils.isNullOrBlank;
+import static dev.langchain4j.internal.Utils.isNullOrEmpty;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotBlank;
+import static dev.langchain4j.model.chat.request.json.JsonSchemaElementHelper.toMap;
```
dliubarskyi:
```suggestion
```

dliubarskyi (dismissed):

PR #2829, opened 2025-04-09
Add script to check split packages

mariofusco:
## Change
Adding script to check if the project has split packages. I also put it in a github action in case we want to enable this check automatically.

### DIFF
--- a/.github/workflows/split-package-detection.yaml
+++ b/.github/workflows/split-package-detection.yaml
@@ -25,85 +25,11 @@ jobs:
       - name: Build with Maven
         run: mvn package -Dspotless.check.skip=true -Dmaven.javadoc.skip=true -DskipTests -DskipITs
 
-      - name: Create split package detection script
-        run: |
-          cat > check-split-packages.sh << 'EOF'
-          #!/bin/bash
-          # Usage: ./check-split-packages.sh /path/to/your/root/dir
-          ROOT_DIR="${1:-.}"  # default to current dir if no argument passed
-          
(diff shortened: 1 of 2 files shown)


.github/workflows/split-package-detection.yaml
```diff
@@ -0,0 +1,115 @@
...
+          EOF
+          chmod +x check-split-packages.sh
+
+      - name: Run split package detection
+        run: |
+          ./check-split-packages.sh 
+
+      - name: Upload results as artifact if failure
+        if: failure()
+        uses: actions/upload-artifact@v3
```
dliubarskyi:
```suggestion
        uses: actions/upload-artifact@v4
```
If I recall correctly v3 is not supported any more, and it fails

dliubarskyi (commented):
@mariofusco thanks a lot, this is awesome!

.github/workflows/split-package-detection.yaml
mariofusco: Right, I bumped all v3 occurrences to v4.

mariofusco:
@dliubarskyi I realized that since I put the script in the root folder of the project I could simply invoke it as it is instead of embedding it into the github action. Give me a minute to fix this.

mariofusco:
> @dliubarskyi I realized that since I put the script in the root folder of the project I could simply invoke it as it is instead of embedding it into the github action. Give me a minute to fix this.

Done, please check again.

dliubarskyi (approved):
@mariofusco great stuff, thank you!

PR #2571, opened 2025-02-18
Input and output guardrails

edeandrea:
Guardrails support.

I also introduced a new `langchain4j-test` module which can be a future place to put in some testing utilities. For now I have several assertj-based assertion classes that can be used for asserting guardrails. (see https://docs.quarkiverse.io/quarkus-langchain4j/dev/guardrails.html#_unit_testing for current details).

Additionally, I've introduced a new `integration-tests` where we can build various sets of test suites (think things that use/need `ServiceLoader`s). This way we can test various implementations of things for frameworks that may extend LangChain4j (Quarkus, Spring, etc).

I've tried to include as much "genericism" as I can into this so that downstream frameworks can re-use this without having to re-implement.

I'll have @geoand / @cescoffier review this too before merging.

## To-do list
- [x] Initial scaffolding for being able to define guardrails
- [x] Guardrail testing utilities
- [x] Wire guardrails into `AiService`s via annotations
- [x] Wire guardrails into `AiService`s using the builder
- [x] Handle output guardrails on streaming responses
- [x] Lots of tests
- [x] Documentation

Fixes #2549

PR #2844, opened 2025-04-10
Refactor AiServiceTokenStream constructor with pilot way for encapsulating parameters

edeandrea:
This is somewhat of a POC for how we may in the future encapsulate classes with large numbers of parameters in their constructors.

This would allow it to be backwards-compatible and non-breaking whenever parameters to a constructor on public-facing objects are created.


dliubarskyi (approved):
@edeandrea thank you!

edeandrea:
This stemmed from a conversation that @mariofusco and I were having re: https://github.com/langchain4j/langchain4j/pull/2571/files#r2007963974

A way to be able to modify constructor parameters in a backwards-compatible way. I know you don't want to merge #2571 before the 1.0 release, but with this specific change it'll make it so that #2571 is completely backwards compatible without any breaking changes.

PR #2869, opened 2025-04-15
Move MCP traffic logging to a logger named 'MCP'

### LINKED ISSUE #2852: [FEATURE] [MCP] Move MCP traffic logging to loggers with distinct names
Use loggers with some distinct names (one for `in`, one for `out`) for logging MCP client-server traffic so that they can be more easily customized using the logging framework that is in use. Right now they are mixed with other log messages in the same loggers.

jmartisk:
Fixes https://github.com/langchain4j/langchain4j/issues/2852


dliubarskyi (approved):
@jmartisk thank you!

PR #2899, opened 2025-04-22
MCP: HTTP: Added capability to inject authorization headers into request.

### LINKED ISSUE #2898: [FEATURE] Capability to handle authorization with MCP

Fresh update to the standard assumes authorization, which is missing at the moment.
https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization#2-6-access-token-usage

Let inject interceptors for okhttp3 so tokens and actual token authorization can happen well before the actual invocation. tokens refreshed on time as needed, and other twists of authorizations.

- Putting token lambda function instead of complete interceptor.
   Slightly more more to implement

**Additional context**

MCP server may use token to authorize use of selected tools based of the token.

kirmorozov:
## Issue

Closes https://github.com/langchain4j/langchain4j/issues/2898

## Change

Extended builder to add modifiers, so there may be more ways to inject authorizing headers, or anything we really want.
There may be other ways to authorize, starting from custom header.
Added Indirect exposure of entire okhttp3 builder so any alteration are possible for the transport. 
Tests assumes using injections of lambdas, so no explicit dependency is necessary.

## General checklist


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -201,6 +207,11 @@ public Builder sseUrl(String sseUrl) {
             return this;
         }
 
+        public Builder interceptors(List<Interceptor> interceptors) {
```
dliubarskyi: I would not expose OkHttp APIs as we would probably want to be able to [use other HTTP clients](https://docs.langchain4j.dev/tutorials/customizable-http-client) in the future

dliubarskyi (commented):
@kirmorozov thank you!

cc @jmartisk

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: +1, I think we should rather accept a `Supplier<Map<String, String>>`  (instead of `List<Interceptor>`) that supplies a map where the key would be a header name, and value would be a header value, thus allowing to inject (potentially multiple) headers. That will make rewriting to the HTTP client SPI  easier (and possible without breaking changes).

jmartisk:
I think @sberyozkin should have a look at this too because we should make sure this can be re-used for OIDC integration

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
kirmorozov: What if we want to handle retry in case of 401 with fresh/regenerated/requested token?

kirmorozov: Looks like I found what I really wanted to inject:
https://github.com/square/okhttp/blob/d0436da9692dc8b7273b60ad176ebd54df2a803f/samples/guide/src/main/java/okhttp3/recipes/Authenticate.java#L29

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
```diff
@@ -14,10 +14,12 @@
 import java.io.IOException;
 import java.net.URI;
 import java.time.Duration;
+import java.util.List;
```
sberyozkin, line 18: I agree with @jmartisk , IMHO we should be dealing with the map of extra headers.

I believe things like a token refresh is a higher level concern for the consumers of this API.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
jmartisk: > What if we want to handle retry in case of 401 with fresh/regenerated/requested token?

That is why I am suggesting a `Supplier` of map so that the consumer of this API can provide dynamic updates to the map (put the new token in it after it needs a refresh), WDYT?

> Looks like I found what I really wanted to inject: https://github.com/square/okhttp/blob/d0436da9692dc8b7273b60ad176ebd54df2a803f/samples/guide/src/main/java/okhttp3/recipes/Authenticate.java#L29

This looks specific to authentication, I would make this mechanism more general to allow plugging any kind of header

yashthattebito:
AI Code Review is in progress (usually takes 3 to 15 minutes unless it's a very large PR).

yashthattebito:
AI Code Review is in progress (usually takes 3 to 15 minutes unless it's a very large PR).

yashthattebito:
<div>
<h2>Changelist by <a href="https://bito.ai">Bito</a></h2>
<p>This pull request implements the following key changes.</p>

<div>
<table width="100%">
<tbody>
  <tr>
    <th>Key Change</th>
    <th>Files Impacted</th>
  </tr>

<tr>
<td width="30%">
    <b>New Feature</b> - Enhanced Interceptor Injection Capability
</td>
<td width="70%">


<div>

-&nbsp;[`HttpMcpTransport.java`](https://github.com/langchain4j/langchain4j/pull/2899/files#diff-0fc89a9ff441d5fd061b55b508312ecc5bf7a33365d290ad0d0f5517b6429dba) - Introduced support for a list-based interceptor configuration by adding necessary imports, default interceptor initialization, and a new builder method to enable flexible injection of authorization headers.
</div>
</td>
</tr>

</tbody>
</table>
</div>
</div>

langchain4j-mcp/src/test/resources/logging_secured_mcp_server.java
```diff
@@ -0,0 +1,51 @@
+///usr/bin/env jbang "$0" "$@" ; exit $?
+//JAVA 21
```
kirmorozov, line 2: Need som guidance here to spin up secured map server

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
kirmorozov: Why would we limit and try to hide okhttp? it is in the bundle anyway.
Giving implicit interface to inject modifiers of transport does not look as wrong.
Using builders as argument for lambdas is all over aws api, so nothing new in terms of indirect exposure of the builder. Retries, refreshes, possibly constructing JWT with some encrypted components, do anything with the builder.

sberyozkin: @kirmorozov Bringing a direct okhttp dependency can have side effects for langchain4j extensions. 

As far as the token refresh and the like.
Let's be specific. 

The token can only be refreshed if the refresh token is available, client id and secret are available. 

As far as access token propagation to remote LLMs is concerned, these are incoming bearer tokens which can not be refreshed.  If it was an authorisation code flow, then the token refresh of the access tokens is out of scope for langchain4j either.

Can you actually please describe a concrete flow you have in mind ?
Thanks

jmartisk: > Why would we limit and try to hide okhttp? it is in the bundle anyway.

For example, the Quarkus extension, which consumes langchain4j, isn't using okhttp. It reimplements the HTTP/SSE transport with its own http client library (vert.x). Over time, we want to migrate away from directly using okhttp to use the general SPI to save us from having to maintain two separate implementations of the transport.

sberyozkin: As far as the interfaces are concerned, at the Quarkus LangChain4j level, https://github.com/quarkiverse/quarkus-langchain4j/blob/main/core/runtime/src/main/java/io/quarkiverse/langchain4j/auth/ModelAuthProvider.java is used.

Perhaps LangChain4j can have its own `ModelAuthProvider` ? We'd then update Quarkus LangChain4j just to extend the upstream one and eventually migrate to the upstream one.

IMHO the ability to add a Map of headers without having to create interface implementations is still useful, both options can complement each other

kirmorozov: I want to pass-thru token, which has reference to a customer.
I wish to separate out mcp that will keep same access level.

Customer gets its token via common way it gets regulate api. 
Objective is to have openai-like api, that passes thru that token, so tools within MCP actually use proper credentials.
I want MCP to extract Customer ID from JWT which significantly alters return from a tool.
As of now using tools directly with memory that uses customer id.

sberyozkin: @kirmorozov Sure, so having an option to pass a map of headers should do ? 
Having an interface is also important, for example, we use the interface I linked to [here](https://github.com/langchain4j/langchain4j/issues/2944) to have MCP clients propagate tokens to MCP servers. (I made a mistake with linking to the ModelAuthProvider above).

How about you look at adding Map of headers while we will look at resolving #2944 with `McpClientAuthProvider` ?

langchain4j-mcp/src/test/resources/logging_secured_mcp_server.java
jmartisk: You mean how to run this jbang script? If you have jbang installed, it should be just `jbang /path/to/script.java`

kirmorozov: Make secured quarkus mcp, with Bearer token123 as validation.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
kirmorozov: So here is what I am currently doing without MCP, but would prefer to do with MCP:
- User comes in with a token, and session id(using it as Memory ID), 
- Token gets stored within a (concurrent) hash map, with the session id.
- Model is set up with tools, tools are using REST API with a token base do memory id reference (Thanks for tools with memory id by the way it helped)

And in along run I would love to have it operational with ADK/A2A. 
So access credentials are being transparently passed. as agent performs activities on user behalf, with respective level of access.

stantonk:
Is there an ETA on when this will get merged? Also blocked on this

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
sberyozkin: @kirmorozov IMHO using either a map of headers or the custom `McpClientAuthProvider` should take care of it

SR-Waivgen:
Hello , We want API Key authentication atleast to be working , Is it done ? I see that its dev done but not yet merged ? When it will be done ?

ls-rein-martha:
Just wondering whether this is going to be merged, or if there is any other plan in place?

jmartisk:
Turns out that custom headers for the legacy HTTP/SSE transport were added via https://github.com/langchain4j/langchain4j/pull/3570 in the end (Uhm, I wasn't quite happy with that PR either, but it was merged anyway). Note that the transport is deprecated anyway and might go away soon (once most MCP servers have adopted Streamable HTTP). Thanks!

PR #2904, opened 2025-04-22
Create a security policy

jdubois:
This will create a security policy at https://github.com/langchain4j/langchain4j/security so we can manage CVEs properly before they are released


dliubarskyi (approved):
@jdubois thank you!

PR #1518, opened 2024-07-25
Fix DefaultPromptTemplateFactory RegExp for Android.

marble94:
## Issue

Closes #1075 #1212

## Change

Adjusted the regex according to the discussion in both issues. Tested the change with a sample repository using an instrumentation test which uses the DefaultPromptTemplateFactory. The test can be found [here](https://github.com/marble94/langchain4j-android-regex-fix/blob/5640a9743e33f7359305596fccf0fe4cb816a54a/app/src/androidTest/java/com/example/myapplication/ExampleInstrumentedTest.kt#L20).

If a test is required in this repo please advice me on possible steps, I'm not sure how to test it without setting up an Android environment.

## General checklist

- [x] There are no breaking changes
- [ ] I have added unit and integration tests for my change
  - [x] I have created an example Android repository verifying that the change fixes the runtime error in Android
- [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new model integration

- [ ] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #2946, opened 2025-04-30
Kotlin: Add support for converting TokenStream to coroutines Flow

kpavlov:
## Change

Introduce utility methods to convert `TokenStream` into a `Flow` of `String` or `StreamingChatModelReply` for Kotlin—added adapters (TokenStreamToStringFlowAdapter, TokenStreamToReplyFlowAdapter) for seamless integration. 
Unit tests were included to validate behavior and edge cases. Set DEFAULT_BUFFER_CAPACITY to Kotlin Flow to 32768.

Porting and enhancing original [PR](https://github.com/kpavlov/langchain4j-kotlin/pull/116) by  @nomisRev 

Kotlin compiler settings will be updated in #2957 

## General checklist

## Checklist for adding a new Maven module

## Checklist for changing the existing embedding store integration


langchain4j-kotlin/src/main/kotlin/dev/langchain4j/kotlin/service/TokenStreamExtensions.kt
```diff
@@ -0,0 +1,87 @@
...
+                    trySend(text)
+                }
+            }
+            close()
+        }
+        onError { close(it) }
+        start()
+        awaitClose()
+    }.buffer(
+        capacity = bufferCapacity, onBufferOverflow = onBufferOverflow
```
nomisRev, line 52: `val DEFAULT_BUFFER_CAPACITY: Int = 16384` this might start dropping tokens if the LLM streams more than `DEFAULT_BUFFER_CAPACITY` tokens.

Project Reactor uses UNLIMITED, this is the safest option. Since this is not protected by an outer `flow` wrapper, prefer to leave it to `Unlimited` by default and it can be overridden by the user by chaining an additional `buffer`.

langchain4j-kotlin/src/main/kotlin/dev/langchain4j/kotlin/service/TokenStreamExtensions.kt
```diff
@@ -0,0 +1,87 @@
...
+ *         responses, complete responses, or errors in the order they are received.
+ */
+@JvmOverloads
+public fun TokenStream.asReplyFlow(
+    bufferCapacity: Int = DEFAULT_BUFFER_CAPACITY,
+    onBufferOverflow: BufferOverflow = BufferOverflow.SUSPEND
+): Flow<StreamingChatModelReply> =
+    callbackFlow {
+        onPartialResponse { token ->
+            trySend(StreamingChatModelReply.PartialResponse(token))
```
nomisRev, line 78: How many cases does `StreamingChatModelReply` have?

kpavlov: Partial and completed responses, also error

langchain4j-kotlin/src/main/kotlin/dev/langchain4j/kotlin/service/TokenStreamExtensions.kt
kpavlov: That is true. A balance between stability and functionality. I am sorry that project reactor has prioritized functionality over stability.
There is still an option for the user to switch to UNLIMITED buffers, if they wish so.

nomisRev: > I am sorry that project reactor has prioritized functionality over stability.

We are in strong disagreement on that 😅 I agree with Project Reactor, and crated the fix PR to langchain4j-kotlin before even checking the Project Reactor. I don't believe they prioritise functionality of stability, they prioritise correctness since that cannot be fixed from user land.

> There is still an option for the user to switch to UNLIMITED buffers, if they wish so.

Still, the chance of silent errors here is huge and I cannot even think of a single use-case to reduce buffer here especially since back pressuring the LLM/HttpClient isn't possible.

langchain4j-kotlin/src/main/kotlin/dev/langchain4j/kotlin/service/TokenStreamExtensions.kt
nomisRev: This return type should only have partial and completed since the error is being signalled through the exception channel.

langchain4j-core/src/main/java/dev/langchain4j/model/input/DefaultPromptTemplateFactory.java
```diff
@@ -22,8 +21,15 @@ public DefaultTemplate create(PromptTemplateFactory.Input input) {
...
-        private static final Pattern VARIABLE_PATTERN = Pattern.compile("\\{\\{(.+?)\\}\\}");
+        /**
+         * A regular expression pattern for identifying variable placeholders within double curly braces in a template string.
+         * Variables are denoted as <code>{{variable_name}}</code> or <code>{{ variable_name }}</code>,
+         * where spaces around the variable name are allowed.
+         * <p>
+         * This pattern is used to match and extract variables from a template string for further processing,
+         * such as replacing these placeholders with their corresponding values.
+         */
+        private static final Pattern VARIABLE_PATTERN = Pattern.compile("\\{\\{\\s*(.+?)\\s*}}");
```
dliubarskyi: Could you please check that this change does not bring the problem fixed in #1518 back

dliubarskyi (commented):
@kpavlov thank you! I have reviewed/commented only on the changes in the java part

langchain4j-kotlin/src/main/kotlin/dev/langchain4j/kotlin/service/TokenStreamExtensions.kt
kpavlov: let' me change it in next PR

langchain4j-core/src/main/java/dev/langchain4j/model/input/DefaultPromptTemplateFactory.java
kpavlov: Thank you, reverted the regexp back

kpavlov:
@nomisRev , yes, we are in strong disagreement on the buffering strategy. I'm not comfortable with a setting that could cause a production incident at scale.
The API allows users to opt in for UNLIMITED buffers if they want to prioritize correctness over stability.
On the other hand, if the last message delivers the [complete response](https://github.com/langchain4j/langchain4j/pull/2946/files#diff-66818b8b09069007df4913a4d8636164b4e66620359b526d78a6fab74e8cc7f9R36), then correctness is ultimately preserved.
DEFAULT_BUFFER_SIZE is set to 32K, which is higher than the typical output tokens value. 

> For developers who prefer rewriting entire files, we’ve increased output token limits for GPT‑4.1 to 32,768 tokens (up from 16,384 tokens for GPT‑4o).
https://openai.com/index/gpt-4-1/

dliubarskyi (approved):
@kpavlov thanks a lot!

Please don't forget to update the documentation in a separate PR

PR #2951, opened 2025-05-02
Reflection variable resolver

kpavlov:
## Issue
Cleaning up to prepare for implementing async AI Services

## Change
- This PR refactors and improves how method argument values are extracted and mapped for prompt templates in AI service methods. It moves variable resolution logic out of the main service implementation and into a dedicated utility class, which enhances code clarity, modularity, and reusability. **The new utility also safely handles null argument arrays, preventing possible NPE errors.** Minor code formatting and organization tweaks were also included for better readability and consistency. No feature or behavioral changes were introduced.
- Enabled OpenAI tests to conditionally run based on the presence of the `OPENAI_API_KEY` environment variable. This ensures tests are only executed when the necessary API key is available, improving reliability in controlled environments.
- Add unit tests for ReflectionVariableResolver

## General checklist


langchain4j/src/main/java/dev/langchain4j/service/ReflectionVariableResolver.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Map;
+
+/**
+ * Utility class responsible for resolving variable names and values for prompt templates
+ * by leveraging method parameters and their annotations.
+ * <p>
+ * This class is intended for internal use only and is designed to extract and map
+ * parameter values to template variables in methods defined within AI services.
+ */
+public class ReflectionVariableResolver {
```
mariofusco: Could this be package private?

mariofusco (dismissed):
Nice refactor, thank you. Let's just try to keep all classes and methods with the most restricting visibility possible. In general we need to reduce the surface of all our public APIs.

langchain4j/src/main/java/dev/langchain4j/service/ReflectionVariableResolver.java
kpavlov: I love weaker access modifiers, but want to make it open for reuse by extensions, e.g. in [my AsyncAiServices ](https://github.com/kpavlov/langchain4j-kotlin/pull/128). It's shaping up

kpavlov:
> Nice refactor, thank you. Let's just try to keep all classes and methods with the most restricting visibility possible. In general we need to reduce the surface of all our public APIs.

+1, but I want to use it Kotlin AIServices, so, I think it should be public.

langchain4j/src/main/java/dev/langchain4j/service/ReflectionVariableResolver.java
dliubarskyi: Then let's annotate it `@Internal` and rename into something like `InternalReflectionVariableResolver`

dliubarskyi (commented):
@kpavlov thank you, nice refactoring

dliubarskyi (approved):
@kpavlov thank you!

PR #2960, opened 2025-05-05
Use info as the HTTP logging level

geoand:
Otherwise, users you have configured to log
requests and/or responses see nothing by default

### DIFF
--- a/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/log/HttpResponseLogger.java
+++ b/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/log/HttpResponseLogger.java
@@ -1,23 +1,26 @@
 package dev.langchain4j.http.client.log;
 
+import static dev.langchain4j.http.client.log.HttpRequestLogger.format;
+
 import dev.langchain4j.Internal;
 import dev.langchain4j.http.client.SuccessfulHttpResponse;
 import org.slf4j.Logger;
-
-import static dev.langchain4j.http.client.log.HttpRequestLogger.format;
 
 @Internal
 class HttpResponseLogger {
 
     static void log(Logger log, SuccessfulHttpResponse response) {
         try {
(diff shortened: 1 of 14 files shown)


langchain4j-http-client/src/main/java/dev/langchain4j/http/client/log/HttpResponseLogger.java
```diff
@@ -1,23 +1,26 @@
...
 
-import static dev.langchain4j.http.client.log.HttpRequestLogger.format;
-
 @Internal
 class HttpResponseLogger {
 
     static void log(Logger log, SuccessfulHttpResponse response) {
         try {
-            log.debug("""
+            log.debug(
```
dliubarskyi:
```suggestion
            log.info(
```

dliubarskyi: There is also `log.debug("{}", event);` in `LoggingHttpClient` which should probably be `log.info` as well?

dliubarskyi (commented):
@geoand thank you! Do you think it is a breaking change in the sense that it will require release of `1.0.0-rc2` before `1.0.0`?

geoand:
> @geoand thank you! Do you think it is a breaking change in the sense that it will require release of `1.0.0-rc2` before `1.0.0`?

I personally wouldn't do an `1.0.0-rc2` just for this.

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/log/HttpResponseLogger.java
geoand: Definitely!

dliubarskyi (approved):
@geoand thanks!

geoand:
🙏🏽

PR #2986, opened 2025-05-09
Fix: OpenAI: map exceptions for streaming and moderation models

dliubarskyi:
## Issue
`OpenAiStreamingChatModel`, `OpenAiStreamingLanguageModel` and `OpenAiModerationModel` are not mapping low-level `HttpException` to common exceptions (e.g., `AuthenticationException` and `RateLimitException`) like all other `OpenAi*Model` do.

## Change
Map exceptions for `OpenAiStreamingChatModel`, `OpenAiStreamingLanguageModel` and `OpenAiModerationModel`

## General checklist


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
```diff
@@ -135,7 +136,13 @@ public void doChat(ChatRequest chatRequest, StreamingChatResponseHandler handler
                     ChatResponse chatResponse = openAiResponseBuilder.build();
                     handler.onCompleteResponse(chatResponse);
                 })
-                .onError(handler::onError)
+                .onError(throwable -> {
+                    if (throwable instanceof Exception exception) {
+                        handler.onError(ExceptionMapper.DEFAULT.mapException(exception));
```
mariofusco: That's ok for now, but thinking twice maybe the `ExceptionMapper` should actually map a `Throwable` instead of an `Exception` so you won't need to do this distinction between the 2 here?

mariofusco (dismissed):

kpavlov:
See also #2989

kpavlov:
It make saense to create a test like [this](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiChatModelErrorsTest.java) for streaming model. There is a [mocking utility](https://mokksy.dev/docs/ai-mocks/openai/#stream-responses)

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
dliubarskyi: Good idea, let me change that

PR #2989, opened 2025-05-10
Exception mapper betterments

kpavlov:
## Issue
Improve the ExceptionMapper extensibility and handling timeouts.
This is required for #2990 

## Change
- **Breaking change: SocketTimeoutException root cause is now translated to TimeoutException**
- Added a utility method 'with' to apply consumer actions on objects, enabling inline object modifications in a functional style. Included comprehensive test coverage to ensure correctness, handling null and non-null cases.
- Mark `repeat` method as deprecated and refactor it to use `String.repeat`. Add `@Nullable` annotations to clarify nullability in the `with` method. Suppress warnings for HTTP URL usage in the `readBytes` method for cleaner code.
- Introduced `findRoot` and `findCause` methods to traverse the causal chain of throwables. These utilities help identify the root cause or specific cause type of an exception, enhancing error-handling capabilities.
- Introduced a new constructor in the HttpException class that allows wrapping a throwable and a status code.
- Introduced comprehensive unit tests for ExceptionMapper to validate exception mapping behavior.
- Enhanced DefaultExceptionMapper to support custom and default HTTP status code extraction, improving flexibility and handling of various edge cases.

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/internal/Exceptions.java
```diff
@@ -1,6 +1,7 @@
 package dev.langchain4j.internal;
 
 import dev.langchain4j.Internal;
+import org.jspecify.annotations.Nullable;
```
mariofusco, line 4: I know that Intellij (annoyingly) insert these kind of annotations when refactoring code, but please let's avoid them.

langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
```diff
@@ -19,32 +23,30 @@
...
+import java.util.function.Consumer;
 import java.util.function.Supplier;
-
-import static java.net.HttpURLConnection.HTTP_OK;
-import static java.nio.charset.StandardCharsets.UTF_8;
-import static java.util.Collections.unmodifiableList;
-import static java.util.Collections.unmodifiableMap;
-import static java.util.Collections.unmodifiableSet;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
```
mariofusco, line 29: Same as above.

langchain4j-core/src/main/java/dev/langchain4j/internal/Exceptions.java
kpavlov: I don't agree, @mariofusco. This is not a noise, but a workaround for missing null-safety in Java. Without this annotations Intyellij and Kotlin compiler can't make any reasoning about field and method result nullability. It actually pays off very well by highlighting possible NPEs and failing Kotlin compilation when NPE will be thrown.
I recommend annotating all Java code instead, especially for public APIs, also for documentation reasons.

https://jspecify.dev/docs/whether/

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -36,16 +40,35 @@ default <T> T withExceptionMapper(Callable<T> action) {
...
+            final var statusCode = httpStatusCodeExtractor.apply(e);
+            if (statusCode != null) {
+                return mapHttpStatusCode(e, statusCode);
             }
 
-            if (rootCause instanceof UnresolvedAddressException) {
+            Throwable rootCause = findRoot(e);
+            if (!(e instanceof UnresolvedModelServerException) && rootCause instanceof UnresolvedAddressException) {
                 return new UnresolvedModelServerException(rootCause);
+            } else if (!(e instanceof TimeoutException) && rootCause instanceof SocketTimeoutException) {
```
dliubarskyi: why is `!(e instanceof TimeoutException)` check required?

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
```diff
@@ -36,16 +40,35 @@ default <T> T withExceptionMapper(Callable<T> action) {
 
     class DefaultExceptionMapper implements ExceptionMapper {
 
+        private final Function<Throwable, Integer> httpStatusCodeExtractor;
+
+        public DefaultExceptionMapper(Function<Throwable, Integer> httpStatusCodeExtractor) {
+            this.httpStatusCodeExtractor = httpStatusCodeExtractor;
```
dliubarskyi, line 46: Why is this needed? Can't find any use of a custom extractor in this PR or in [Gemini PR](https://github.com/langchain4j/langchain4j/pull/2990/files)

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
kpavlov: I wanted to avoid loops, but might be it's better to check if TimeoutExceptuon is somewhere in the cause chain

langchain4j-core/src/main/java/dev/langchain4j/internal/ExceptionMapper.java
kpavlov: Right now it's not needed. 

I ~~was lazy and~~ wrapped custom Gemini http exception with lc4j HttpException, so default mapping works. 

Not sure if it's a better approach, than creating specific lc4j exceptions directly, without HttpException in the cause chain. For later approach, the custom status code extractor would be needed.
@dliubarskyi , do we want all client-native status code exceptions wrapped into HttpException or not?

PR #3060, opened 2025-05-19
Dynamic registration and deregistration of MCP clients

### LINKED ISSUE #2983: Dynamic registration and deregistration of MCP clients in McpToolProvider
Our application lets users connect or disconnect MCP servers at runtime. After `McpToolProvider` has been built, we need a way to add or remove `McpClient` instances, but the current implementation keeps the list immutable.

Expose `addMcpClient(McpClient)` and `removeMcpClient(McpClient)` on the existing provider. These methods only modify the internal list; all other behaviour stays unchanged.

1. Re‑create the entire `AiService` each time a user changes servers – this adds redundant boilerplate and possible short downtime.  
2. Create a custom `ToolProvider` that duplicates `McpToolProvider` but with a mutable list – it introduces duplication and also brings no value for other developers that might also find this feature useful.

**Additional context**  
A potential problem of the dynamic change of the MCP clients is the concurrency problems. To avoid it we might use `CopyOnWriteArrayList` data structure. The suggested code change is then:

```java
public class McpToolProvider implements ToolProvider {

    private final List<McpClient> mcpClients = new CopyOnWriteArrayList<>();

    public void addMcpClient(McpClient client) {
        Objects.requireNonNull(client);
        mcpClients.add(client);
    }

    public boolean removeMcpClient(McpClient client) {
        return mcpClients.remove(client);
    }

    //…existing code unchanged…
}
```
We're open to discussion.

mariofusco:
## Issue

Closes https://github.com/langchain4j/langchain4j/issues/2983

## Change

Allows MCP clients and filter to be dynamically changed on a `McpToolProvider` in a thread-safe way.

## General checklist

### DIFF
--- a/docs/docs/tutorials/mcp.md
+++ b/docs/docs/tutorials/mcp.md
@@ -59,7 +59,7 @@ if there are multiple MCP clients, and it is necessary to disambiguate among the
 Finally, you create an MCP tool provider from the client:
 
 ```java
-ToolProvider toolProvider = McpToolProvider.builder()
+McpToolProvider toolProvider = McpToolProvider.builder()
     .mcpClients(mcpClient)
     .build();
 ```
@@ -78,7 +78,7 @@ reduce the possibility of hallucinations. The `McpToolProvider` allows to filter
 these tools by name as it follows:
 
 ```java
-ToolProvider toolProvider = McpToolProvider.builder()
+McpToolProvider toolProvider = McpToolProvider.builder()
     .mcpClients(mcpClient)
     .filterToolNames("get_issue", "get_issue_comments", "list_issues")
     .build();
@@ -94,7 +94,7 @@ but since they both have a tool named `echoInteger`, it takes only the one from
 the MCP client with key `numeric-mcp`:
 
 ```java
-ToolProvider toolProvider = McpToolProvider.builder()
+McpToolProvider toolProvider = McpToolProvider.builder()
     .mcpClients(mcpClient1, mcpClient2)
     .filter((mcpClient, tool) ->
             !tool.name().startsWith("echoInteger") || 
@@ -102,7 +102,12 @@ ToolProvider toolProvider = McpToolProvider.builder()
     .build();
 ```
 
-Note that it is possible to configure multiple filters for the same `McpToolProvider`.
+Note that calling the `filter` method multiple time on the same `McpToolProvider`
+builder will result in a conjunction (AND) of all those filters.
+
+In order to allow applications to connect or disconnect from MCP servers at 
+runtime, it is also possible to dynamically add and remove tools and filters 
+to an existing `McpToolProvider` instance.
 
 To bind a tool provider to an AI service, simply use the `toolProvider` method
 of an AI service builder:
--- a/docs/docs/tutorials/mcp.md
+++ b/docs/docs/tutorials/mcp.md
@@ -106,7 +106,7 @@ Note that calling the `filter` method multiple time on the same `McpToolProvider
 builder will result in a conjunction (AND) of all those filters.
 
 In order to allow applications to connect or disconnect from MCP servers at 
-runtime, it is also possible to dynamically add and remove tools and filters 
+runtime, it is also possible to dynamically add and remove clients and filters 
 to an existing `McpToolProvider` instance.
 
 To bind a tool provider to an AI service, simply use the `toolProvider` method
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
@@ -7,9 +7,11 @@ import dev.langchain4j.service.IllegalConfigurationException;
 import dev.langchain4j.service.tool.ToolProvider;
 import dev.langchain4j.service.tool.ToolProviderRequest;
 import dev.langchain4j.service.tool.ToolProviderResult;
-import java.util.ArrayList;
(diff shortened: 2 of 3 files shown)


mariofusco:
@dliubarskyi @jmartisk I flagged this as draft because I haven't updated the documentation yet. Please let me know if this change is ok for you and I will also change the documentation accordingly before merging.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -20,26 +22,44 @@
...
+    public void addMcpClient(McpClient client) {
+        Objects.requireNonNull(client);
+        mcpClients.add(client);
+    }
+
+    public boolean removeMcpClient(McpClient client) {
+        return mcpClients.remove(client);
+    }
+
+    public void addFilter(BiPredicate<McpClient, ToolSpecification> filter) {
```
jmartisk, line 64: This deserves a javadoc comment explaining that it creates a conjunction (AND) with the current filter

Also, I guess we should have a `setFilter` method that simply replaces the current filter with a new one?

mariofusco: The [documentation](https://github.com/langchain4j/langchain4j/blob/main/docs/docs/tutorials/mcp.md#mcp-tool-provider) already says 

> Note that it is possible to configure multiple filters for the same McpToolProvider.

of course implicitly meaning that the filters are in AND, but maybe it is better to make this crystal clear. I will add Javadocs also for all other methods actually. 

As for adding a `setFilter` method I don't have any strong opinion and I'm open to add it. Probably it will less confusing and error-prone to keep to number of methods to a minimal amount. Related to this another doubt I have is if all those new methods should have a fluent interface, i.e. returning the `McpToolProvider` itself instead of `void`. In this way it would be possible, for example, to add a new client and a filter in a single statement or perform the "setFilter" logic with something like `toolProvider.resetFilters().addFilter(myFliter);`.

jmartisk: Ah yeah, the 

> Note that it is possible to configure multiple filters for the same McpToolProvider.

could be a bit clearer too - say that calling `Builder.filter()` multiple times will conjoin the filters

> toolProvider.resetFilters().addFilter(myFliter);

A problem with this is that this is not necessarily thread-safe (atomic), something else might have added a filter between `resetFilters` and `addFilter`. An atomic `setFilter` would be safer..

mariofusco:
I implemented @jmartisk 's suggestions and added a bit of documentation and Javadocs, please review.

jmartisk (approved):
Nice!

docs/docs/tutorials/mcp.md
```diff
@@ -94,15 +94,20 @@ but since they both have a tool named `echoInteger`, it takes only the one from
...
             mcpClient.key().equals("numeric-mcp"))
     .build();
 ```
 
-Note that it is possible to configure multiple filters for the same `McpToolProvider`.
+Note that calling the `filter` method multiple time on the same `McpToolProvider`
+builder will result in a conjunction (AND) of all those filters.
+
+In order to allow applications to connect or disconnect from MCP servers at 
+runtime, it is also possible to dynamically add and remove tools and filters
```
```suggestion
runtime, it is also possible to dynamically add and remove clients and filters 
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
dliubarskyi: TBH I would also prefer to have `setFilter` instead of `addFilter` (also in the builder). It is hard for me to imagine a situation where calling `addFilter` multiple times is 1) required and 2) is much more convenient than specifying an `AND` condition manually. But having this `AND` logic might be confusing for users and lead to frustrating bugs that can be found only by debugging

dliubarskyi (commented):
@mariofusco thank you!

mariofusco:
> TBH I would also prefer to have setFilter instead of addFilter (also in the builder). It is hard for me to imagine a situation where calling addFilter multiple times is 1) required and 2) is much more convenient than specifying an AND condition manually. But having this AND logic might be confusing for users and lead to frustrating bugs that can be found only by debugging

I see your point, this will break backward compatibility, but I guess this won't be a problem for the MCP module, right?

In reality there is a quite common situation where you want to combine 2 filters to have a filter on tool names, plus another one to disambiguate a tool names conflict like here https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/MultipleMcpToolsIT.java#L115-L116

The alternatives that I see for that are:

1. Removing the `filterToolNames` feature and leave only the generic predicate-based filter. I'm quite against this because that API is quite handy and I believe that in 90% of cases is the only kind of filter that users need.
2. Treating the `filterToolNames` as a special, separated kind of filter without internally combining it in AND with the other filters eventually set by users. This will be a little less elegant, and add some verbosity internally, but it will be something that we will manage internally and not exposed to end-users.

In essence I would prefer to keep the AND-logic (it is not something different to what it happens when you use multiple time the `filter` method of the Stream API, so hopefully it is not so confusing and alien for users) but if you want to get rid of it I would prefer option `2.` over `1.`. Please let me know.

dliubarskyi:
@mariofusco good point, I did not realize that it was indeed similar to the Stream API.

dliubarskyi (approved):
@mariofusco thank you!

PR #3069, opened 2025-05-20
AI Service: support immediate/direct return from tools

### LINKED ISSUE #124: [FEATURE] Option to directly return from tool call

mariofusco:
## Issue
Closes https://github.com/langchain4j/langchain4j/issues/124

## Change
This is a rework from scratch of what has been done in https://github.com/langchain4j/langchain4j/pull/2291 and https://github.com/langchain4j/langchain4j/pull/2772 also based on the comments made during the review of those pull requests.

As I described in the Javadocs, here I added a `ReturnBehavior` enum (names are very questionable and I'm open to change them if you have better proposals) implementing 3 different behaviors for tools:

```
    /**
     * Return behavior of the tool.
     * - If {@link ReturnBehavior#TO_LLM} is used (default), the value returned by the tool is sent back to the LLM for further processing.
     *   The LLM invocation will directly return the value returned by the last invoked tool, instead of the one further manipulated by the LLM,
     *   if and only if all the tools invoked by the LLM to fulfill the request are marked with {@link ReturnBehavior#DIRECT}.
     * - If {@link ReturnBehavior#IMMEDIATE} is used, returns immediately to the caller the value returned by the tool without
     *   allowing the LLM to further processing it.
     *
     * @return return behavior of the tool.
     */
    ReturnBehavior returnBehavior() default ReturnBehavior.TO_LLM;
```

The tests that I added should also help to clarify how this works.

Regarding the actual need for this feature, what has been discussed in [this article](https://quarkus.io/blog/agentic-ai-with-quarkus-p3/#comparing-the-workflow-and-agentic-approaches) provides another example of a situation where this could be useful.

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
```diff
@@ -36,4 +36,17 @@
      * @return description of the tool.
      */
     String[] value() default "";
+
+    /**
+     * Return behavior of the tool.
+     * - If {@link ReturnBehavior#TO_LLM} is used (default), the value returned by the tool is sent back to the LLM for further processing.
+     * - If {@link ReturnBehavior#DIRECT} is used, the value returned by the tool is sent to the LLM to allow it to invoke more tools if necessary.
```
geoand: I find the name `DIRECT` confusing in this case, but I have to admit that I can't think of anything better...

mariofusco:
@dliubarskyi I rebased this pull request against the main branch. Now that guardrails have been merged I believe that we could restart the discussion on this, please let me know if you have any feedback.

tokalak:
Hey Mario, 
thanks for working on this very important feature!

I must say, that the distinction between `TO_LLM` and `DIRECT` is not clear to me. I mean, in the end it's the LLM, who decides if another tool call is necessary or not. 

Thanks again,
Ahmet

mariofusco:
> I must say, that the distinction between `TO_LLM` and `DIRECT` is not clear to me. I mean, in the end it's the LLM, who decides if another tool call is necessary or not.

Correct, the difference is in what you return to the end user. In the `TO_LLM` case you return the LLM outcome, as it does now, while with `DIRECT` you return the response of the latest tool invoked as it is (provided that its return type is compatible with the one of the expected response), discarding the manipulation that the LLM did of it.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnBehavior.java
```diff
@@ -0,0 +1,5 @@
+package dev.langchain4j.agent.tool;
+
+public enum ReturnBehavior {
+    TO_LLM, DIRECT, IMMEDIATE;
```
dliubarskyi: nit: worth copy-pasting javadoc here as well

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
```diff
@@ -16,6 +16,7 @@ public class ToolSpecification {
     private final String name;
     private final String description;
     private final JsonObjectSchema parameters;
+    private final ReturnBehavior returnBehavior;
```
dliubarskyi: `ToolSpecification` is not the best place for this. `ToolSpecification` is the "schema" of the tool that is sent to the LLM, and LLM surely does not need to know about tool return behaviour. I guess `ReturnBehaviour` should be a concern of the `ToolService`, WDYT?

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +809,213 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+                var toolSpecifications = chatRequestCaptor.getValue().toolSpecifications();
+                assertThat(toolSpecifications).hasSize(2);
+            }
+
+            // perform one more invocation of the AI service to make sure that memory is not corrupted and the service can be used further
+            var text2 = "How much is 37 plus 87?";
+            var response2 = assistant.chat(text2);
+            assertThat(response2.content()).contains("124");
+        }
+    }
```
dliubarskyi, line 1070: The concerns regarding the `ChatMemory` state I have raised in the original PR and also in https://github.com/langchain4j/langchain4j/pull/2772 seem to be ignored in this PR as well. Could you please explicitly test the state of the memory and that it is not corrupted and conversation can be continued?

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnBehavior.java
```diff
@@ -0,0 +1,5 @@
+package dev.langchain4j.agent.tool;
+
+public enum ReturnBehavior {
```
dliubarskyi, line 9: Do we want to mark it `@Experimental` for the moment?

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
```diff
@@ -36,4 +36,17 @@
...
+     *   The LLM invocation will directly return the value returned by the last invoked tool, instead of the one further manipulated by the LLM,
+     *   if and only if all the tools invoked by the LLM to fulfill the request are marked with {@link ReturnBehavior#DIRECT}.
+     * - If {@link ReturnBehavior#IMMEDIATE} is used, returns immediately to the caller the value returned by the tool without
+     *   allowing the LLM to further processing it.
+     *
+     * @return return behavior of the tool.
+     */
+    ReturnBehavior returnBehavior() default ReturnBehavior.TO_LLM;
```
dliubarskyi, line 51: Do we want to mark it `@Experimental` for the moment?

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +809,213 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
             assertThat(toolSpecifications.get(0)).isEqualTo(ToolWithUUIDParameter.EXPECTED_SPECIFICATION);
         }
     }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    protected void should_execute_normal_tool_with_primitive_parameters() {
```
dliubarskyi: ``
```suggestion
    protected void should_execute_normal_tool_with_primitive_parameters(ChatModel chatModel) {
```
Because of `@MethodSource`.

Same for other tests

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +809,213 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+            }
+            case IMMEDIATE: {
+                toolInstance = new ImmediateToolWithPrimitiveParameters();
+                // 1 time, the model is called only once returning immediately the tool result
+                chatInvocations = 1;
+                break;
+            }
+        }
+
+        for (var model : models()) {
```
dliubarskyi: no need to iterate over models due to `@MethodSource`

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +809,213 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+
+            var addTool = spy(toolInstance);
+            var multiplyTool = spy(new MultiplyDirectTool());
+
+            var assistant = AiServices.builder(Assistant.class)
+                    .chatModel(model)
+                    .tools(addTool, multiplyTool)
+                    .build();
+
+            var text = "First add 2 and 3, then multiply the result by 4";
```
dliubarskyi: could you please add another set of tests where LLM calls multiple tools in parallel? e.g.: "what is 2 plus 3? what is 4 plus 5?"

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
dliubarskyi: TBH I doubt the necessity of the `DIRECT` option. It does not seem to bring any benefit, but it adds to the complexity and confusion... In my understanding, users requested the option to skip the call to the LLM after the tool is called, so I would focus on this request in-depth first

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
dliubarskyi, line 1: It would be great to test the same feature for the streaming use case (`TokenStream` return type)

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -199,4 +211,22 @@ public Map<String, ToolExecutor> toolExecutors() {
     public ToolProvider toolProvider() {
         return toolProvider;
     }
+
+    private static ToolServiceResult rawToolExecutionResult(ChatResponse chatResponse, List<ToolExecution> toolExecutions) {
+        AiMessage toolMessage = AiMessage.builder()
+                .text(toolExecutions.get(toolExecutions.size()-1).result())
```
dliubarskyi: This seems very confusing... A few concerns:
- what if there are multiple tools that were called?
- there seems to be no tests in regards to different return types (`Result<T>`, `T`, `String`, etc). IIRC, at some point there was a suggetion to only allow `DIRECT` if AI Service returns `Result<?>`, because then `result.content()` can be `null` (benefit: no need to override `AiMessage`), and all tool executions will be accessible in `result.toolExecutions()`

dliubarskyi (commented):
@mariofusco thanks a lot!

patriot1burke:
I stumbled into this PR.  I was looking for a way to use natural language to execute a function locally, and THAT'S IT.  Trying to build a natural language CLI for commands I can do in my application.  So, I'm excited about this PR!

One possible suggestion, instead of adding ReturnBehavior to @Tool, wouldn't it be better to create a @ReturnBehavior annotation that can be applied to an AI Service?

`
interface Assistant {
   @ReturnBehavior(IMMEDIATE)
  Result<String>  answer(@UserMessage message);
}
`

Return behavior seems orthogonal to the tool, especially if you're calling into a library.  How would somebody who has created a library of tools know how the developer will want to execute the tool?

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
mariofusco: I removed the `DIRECT` option. I left the field as an `enum` to allow for extensibility with other behaviors in future, but if we do not foresee any other possible alternative I could turn it into a `boolean`. WDYT?

dliubarskyi: Thank you, I would keep it as enum to allow for flexibility to add more options in the future

dliubarskyi: BTW perhaps now `DIRECT` is the better name for `IMMEDIATE`?

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
mariofusco: Done.

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
mariofusco: Done.

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
mariofusco: I added a test checking this.

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
mariofusco: I added a test and implemented support for this feature also in streaming mode.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: I simplified this method a bit, but in reality what this it doing is creating a `ChatResponse` containing only the response obtained from the tool invocation. Also the fact that I removed the `DIRECT` option further helped to simplify the logic. Not sure if it can be better improved, but now it should be extensively tested, covering all use cases that you suggested.

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -118,6 +118,10 @@ public void onCompleteResponse(ChatResponse completeResponse) {
                             .build();
                     toolExecutionHandler.accept(toolExecution);
                 }
+
+                if (context.toolService.isImmediateTool(toolName)) {
```
dliubarskyi: There seems to be 2 problems:
- What if LLM requests to execute multiple tools?
  - As a user I would expect all requested tools to be executed before returning their results
- What if one of them is not "immediate"?
  - Not entirely sure what would be the best approach here, but we could follow some simple rule like "if one of the requested tools is not `IMMEDIATE`, behave as if `TO_LLM`"

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -199,4 +213,21 @@ public Map<String, ToolExecutor> toolExecutors() {
...
+
+    private static ChatResponse rawToolExecutionResponse(ChatResponse chatResponse, ToolExecutionResultMessage toolExecutionResultMessage) {
+        ChatResponseMetadata toolResponseMetadata = ChatResponseMetadata.builder()
+                .id(chatResponse.metadata().id())
+                .modelName(chatResponse.metadata().modelName())
+                .tokenUsage(chatResponse.metadata().tokenUsage())
+                .finishReason(FinishReason.STOP)
+                .build();
+        return ChatResponse.builder()
+                .aiMessage(new AiMessage(toolExecutionResultMessage.text()))
```
dliubarskyi: Same here

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -118,6 +118,10 @@ public void onCompleteResponse(ChatResponse completeResponse) {
...
+                    completeResponseHandler.accept(finalResponse(completeResponse, new AiMessage(toolExecutionResultMessage.text())));
```
dliubarskyi: It does not seem right to return tool return value inside the `AiMessage`, as `AiMessage` represents something generated by the model, not something generated by the application.

In this case it is probably more correct to return the original `AiMessage` in the `onCompleteResponse()` callback, as it is the actual complete response from the LLM. Tool results are anyway acessible via `onToolExecuted()` callback.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -18,22 +19,26 @@
...
 import java.util.Map;
-import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
 
 @Internal
 public class ToolService {
 
-    private final List<ToolSpecification> toolSpecifications = new ArrayList<>();
+    private final Map<String, ToolSpecification> toolSpecifications = new HashMap<>();
```
dliubarskyi: nit: why is this change required? It does not seem like the key of the map is used anywhere

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -157,6 +167,10 @@ public ToolServiceResult executeInferenceAndToolsLoop(
                 } else {
                     messages.add(toolExecutionResultMessage);
                 }
+
+                if (isImmediateTool(toolExecutionRequest.name())) {
```
dliubarskyi: same here: what if there are multiple tools to be executed?

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +807,276 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+    protected void should_keep_memory_consistent_using_return_immediate(ChatModel model) {
+        ChatMemory chatMemory = MessageWindowChatMemory.withMaxMessages(10);
+
+        var assistant = AiServices.builder(Assistant.class)
+                .chatModel(model)
+                .tools(new ImmediateToolWithPrimitiveParameters())
+                .chatMemory(chatMemory)
+                .build();
+
+        var response = assistant.chat("How much is 37 plus 87?");
```
dliubarskyi, line 1083: Could you please add assertions on the state of the `chatMemory` (which messages, in which order, etc) after each call to assistant?

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractStreamingAiServiceIT.java
```diff
@@ -174,4 +179,42 @@ protected boolean assertTokenUsage() {
...
+        // when
+        assistant.chat("How much is 37 plus 87?")
+                .onPartialResponse(ignored -> {})
+                .onToolExecuted(toolExecutions::add)
+                .onCompleteResponse(response -> future.complete(response))
+                .onError(future::completeExceptionally)
+                .start();
+
+        assertThat(future.join().aiMessage().text()).isEqualTo("124");
+
```
dliubarskyi, line 215: same here (`chatMemory` assertions after each call to assistant)

dliubarskyi (commented):
@mariofusco thanks a lot!

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: You're right, I reverted this.

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
mariofusco: Done. I think that now the sequence of messages is exactly what I'd expect:
1. the user request
2. the ai message with the tool execution resquest
3. the tool response
4. the ai message carrying the exact same response of the tool since this is also the response returned to user

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractStreamingAiServiceIT.java
mariofusco: Done.

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +811,306 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+        // Check that the memory is not corrupted and conversation can continue
+        response = assistant.chat("Now add 47 to the previous result");
+        assertThat(response.content()).isEqualTo("171");
+
+        messages = chatMemory.messages();
+        assertThat(messages).hasSize(8);
+        checkMemoryWithImmediateTool(messages.subList(4, 8), "Now add 47 to the previous result", "171");
+    }
+
+    static void checkMemoryWithImmediateTool(List<ChatMessage> messages, String request, String response) {
```
mariofusco, line 1101: Check the memory state after an immediate tool return.

mariofusco:
If I'm understanding correctly there are 2 outstanding concerns:
1. What should it happen when there is the request of 2 or more tools and only one of them has immediate return?
2. What should contain the final `AiMessage` when returning immediately from a tool?

My personal opinion about them (but of course I'm open to discuss this) is the following:
1. Immediate means immediate, so the first response of the first invoked tool flagged as such should be returned immediately as it is.
2. The main goal of an immediately returning tool is avoiding wasting time and resources letting the LLM generating a response that the user doesn't want/need, so I may not have at all the actual LLM-generated `AiMessage`. What I'm returning to user is the tool response and the chat memory simply reflect this situation, see https://github.com/langchain4j/langchain4j/pull/3069/files#r2212940760

dliubarskyi:
@mariofusco 

> If I'm understanding correctly there are 2 outstanding concerns:
> 
> What should it happen when there is the request of 2 or more tools and only one of them has immediate return?

Yes, but also what should happen if there are multiple immediate tool calls.

> What should contain the final AiMessage when returning immediately from a tool?
> My personal opinion about them (but of course I'm open to discuss this) is the following:
> 
> Immediate means immediate, so the first response of the first invoked tool flagged as such should be returned immediately as it is.

If LLM requests execution of multiple immediate tools, why can't we execute all of them before returning? If LLM requested execution of multiple tools, there is probably reason for it (e.g., maybe user explicitly asked for it). I am also concerned that 1) user will not be able to force execution of ignored tool after AI Service returns and 2) some stricter LLMs will fail if tool execution request is not followed by tool result message.

The original concern/issue was redundant LLM call after tool executions, not tool executions themselves, so I think it will be more correct to execute all requested tools before returning. WDYT?

> The main goal of an immediately returning tool is avoiding wasting time and resources letting the LLM generating a response that the user doesn't want/need, so I may not have at all the actual LLM-generated AiMessage.

We do have original AiMessage that contains tool execution requests, we could return it?

Personally, I would allow using immediate tools only when return type is `Result<?>`. In case tools are called `Result.content()` can be `null` and `Result.toolExecutions()` can contain tool results. WDYT?

mariofusco:
> Personally, I would allow using immediate tools only when return type is `Result<?>`. In case tools are called `Result.content()` can be `null` and `Result.toolExecutions()` can contain tool results. WDYT?

I agree with all you wrote, but this last sentence. I find this limitation (unnecessarily?) restrictive even considering the fact that in my opinion very few users (and only for specific situations) use `Result<?>` as return type for a AI service.

mariofusco:
> > Personally, I would allow using immediate tools only when return type is `Result<?>`. In case tools are called `Result.content()` can be `null` and `Result.toolExecutions()` can contain tool results. WDYT?
> 
> I agree with all you wrote, but this last sentence. I find this limitation (unnecessarily?) restrictive even considering the fact that in my opinion very few users (and only for specific situations) use `Result<?>` as return type for a AI service.

Ok, I take this back and now I see why this is unavoidable: when 2 tools with immediate return are executed in parallel it should return both results of their invocations and this is possible only putting them in the `Result<?>`

mariofusco:
@dliubarskyi I should have implemented all your suggestions now, thanks for patience :) 
Please give it one more look.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -189,9 +200,10 @@ public ToolServiceResult executeInferenceAndToolsLoop(
             ChatModel chatModel,
             ChatMemory chatMemory,
             Object memoryId,
-            Map<String, ToolExecutor> toolExecutors) {
-
+            Map<String, ToolExecutor> toolExecutors,
+            boolean isResultType) {
         TokenUsage aggregateTokenUsage = chatResponse.metadata().tokenUsage();
+        int executionsLeft = maxSequentialToolsInvocations;
```
dliubarskyi:
```suggestion
```

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -250,6 +254,15 @@ public void onCompleteResponse(ChatResponse chatResponse) {
...
 
+    private ChatResponse finalResponse(ChatResponse completeResponse, AiMessage aiMessage) {
+        return ChatResponse.builder()
+                .aiMessage(aiMessage)
+                .metadata(completeResponse.metadata().toBuilder()
+                        .tokenUsage(tokenUsage.add(
+                                completeResponse.metadata().tokenUsage()))
+                        .build())
+                .build();
+
```
dliubarskyi, line 268:
```suggestion
    }

```

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -118,6 +120,13 @@ public void onCompleteResponse(ChatResponse completeResponse) {
                             .build();
                     toolExecutionHandler.accept(toolExecution);
                 }
+
+                immediateToolReturn = immediateToolReturn && context.toolService.isImmediateTool(toolExecutionRequest.name());
+            }
+
+            if (immediateToolReturn) {
+                completeResponseHandler.accept(finalResponse(completeResponse, null));
```
dliubarskyi:
```suggestion
                completeResponseHandler.accept(finalResponse(chatResponse, null));
```

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -143,13 +152,7 @@ public void onCompleteResponse(ChatResponse completeResponse) {
...
         } else {
             if (completeResponseHandler != null) {
-                ChatResponse finalChatResponse = ChatResponse.builder()
-                        .aiMessage(aiMessage)
-                        .metadata(completeResponse.metadata().toBuilder()
-                                .tokenUsage(tokenUsage.add(
-                                        completeResponse.metadata().tokenUsage()))
-                                .build())
-                        .build();
+                ChatResponse finalChatResponse = finalResponse(completeResponse, aiMessage);
```
dliubarskyi:
```suggestion
                ChatResponse finalChatResponse = finalResponse(chatResponse, aiMessage);
```

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -34,6 +34,7 @@
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.function.Consumer;
+import dev.langchain4j.service.tool.ToolServiceResult;
```
dliubarskyi:
```suggestion
import dev.langchain4j.model.output.FinishReason;
import dev.langchain4j.service.tool.ToolServiceResult;
```

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -34,6 +34,8 @@
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.function.Consumer;
+import dev.langchain4j.model.output.FinishReason;
```
dliubarskyi:
```suggestion
```

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -273,17 +273,30 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
...
 
+                        if (chatResponse == null) {
+                            return Result.builder()
+                                    .sources(augmentationResult == null ? null : augmentationResult.contents())
+                                    .finishReason(FinishReason.STOP)
+                                    .toolExecutions(toolServiceResult.toolExecutions())
+                                    .build();
+                        }
+
+                        FinishReason finishReason = chatResponse.metadata().finishReason();
```
```suggestion
```

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -273,17 +273,30 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Exceptio
...
-                                toolServiceContext.toolExecutors());
+                                toolServiceContext.toolExecutors(),
+                                isResultType);
 
                         ChatResponse aggregateResponse = toolServiceResult.aggregateResponse();
 
+                        if (chatResponse == null) {
+                            return Result.builder()
+                                    .sources(augmentationResult == null ? null : augmentationResult.contents())
+                                    .finishReason(FinishReason.STOP)
```
```suggestion
                                    .finishReason(dev.langchain4j.model.output.FinishReason.STOP)
```

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
```diff
@@ -36,4 +37,15 @@
      * @return description of the tool.
      */
     String[] value() default "";
+
+    /**
+     * Return behavior of the tool.
+     * - If {@link ReturnBehavior#TO_LLM} is used (default), the value returned by the tool is sent back to the LLM for further processing.
+     * - If {@link ReturnBehavior#IMMEDIATE} is used, returns immediately to the caller the value returned by the tool without
+     *   allowing the LLM to further processing it.
```
dliubarskyi: Worth adding a note that return type of AI Service method should be `Result<T>`

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnBehavior.java
```diff
@@ -0,0 +1,22 @@
...
+
+    /**
+     * The value returned by the tool is sent back to the LLM for further processing.
+     * This is the default behavior.
+     */
+    TO_LLM,
+
+    /**
+     * Returns immediately to the caller the value returned by the tool without allowing the LLM
```
dliubarskyi, line 18: Worth adding a note that return type of AI Service method should be `Result<T>`

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -92,9 +98,14 @@ private void processToolMethod(Object object, Method method) {
             throw new IllegalConfigurationException(
                     "Duplicated definition for tool: " + toolSpecification.name());
         }
+        toolSpecifications.add(toolSpecification);
+
         ToolExecutor toolExecutor = createToolExecutor(object, method);
         toolExecutors.put(toolSpecification.name(), toolExecutor);
-        toolSpecifications.add(toolSpecificationFrom(method));
+
+        if (method.getAnnotation(Tool.class).returnBehavior() == ReturnBehavior.IMMEDIATE) {
```
dliubarskyi, line 106: Will this work if it is a proxy? Perhaps we should use a similar approach like in `Utils.getAnnotatedMethod()`?

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -785,4 +810,338 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+        assertThat(((ToolExecutionResultMessage) messages.get(2)).text()).isEqualTo(response);
+    }
+
+    interface StringAssistant {
+        String chat(String userMessage);
+    }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    protected void should_ignore_immediate_tool_when_not_returning_Result(ChatModel model) {
```
dliubarskyi: I would rather throw an exception as soon as possible if AI service without `Result<T>` return type is configured with at least one immediate tool, WDYT? Otherwise user might be confused for a long time why it does not work

mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: The `method` accessed here is already retrieved using the `Utils.getAnnotatedMethod()`  approach at line 90.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnBehavior.java
mariofusco: Done.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/Tool.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
dliubarskyi: Oh, my bad!

docs/docs/tutorials/tools.md
```diff
@@ -676,6 +676,46 @@ Assistant assistant = AiServices.builder(Assistant.class)
 
 It is possible for an AI service to use both programmatically and dynamically specified tools in the same invocation.
 
+### Returning immediately the result of a tool execution request
+
+By default, the result of a tool execution request is sent back to the LLM that uses this result and further reprocesses it. However, in some circumstance, the result produced by that tool execution request already represents the expected result of the AI service invocation. In this case it is possible to configure the tool to immediately return its result, skipping a wasteful and resource consuming reprocessing by the LLM. This can be done by configuring the `returnBehavior` field of the `@Tool` annotation as in the following example:
```
dliubarskyi:
```suggestion
By default, the result of a tool execution request is sent back to the LLM that uses this result and further reprocesses it. However, in some circumstances, the result produced by that tool execution request already represents the expected result of the AI service invocation. In this case it is possible to configure the tool to immediately/directly return its result, skipping a wasteful and resource consuming reprocessing by the LLM. This can be done by configuring the `returnBehavior` field of the `@Tool` annotation as in the following example:
```

adding "directly" for making text search easier (this feature is called "return direct" in Python LC)

docs/docs/tutorials/tools.md
```diff
@@ -676,6 +676,46 @@ Assistant assistant = AiServices.builder(Assistant.class)
...
+class CalculatorWithImmediateReturn {
+    
+    @Tool(returnBehavior = ReturnBehavior.IMMEDIATE)
+    double add(int a, int b) {
+        return a + b;
+    }
+}
+```
+
+Note that this feature is supported only on AI Services having a `Result` as returned type, while attempting to use it on a service with a different return type will produce a `RuntimeException`.
```
dliubarskyi:
```suggestion
:::note
This feature is supported only on AI Services having a `Result<T>` return type. Attempting to use it on AI Service with a different return type will produce an exception. See [Return Types](/tutorials/ai-services#return-types) for more information about `Result<T>`.
:::
```

docs/docs/tutorials/tools.md
```diff
@@ -676,6 +676,46 @@ Assistant assistant = AiServices.builder(Assistant.class)
...
+Assistant assistant = AiServices.builder(Assistant.class)
+        .chatModel(model)
+        .tools(new CalculatorWithImmediateReturn())
+        .build();
+```
+
+will return a response with content exactly equal to `124` from an invocation like
+
+```java
+var response = assistant.chat("How much is 37 plus 87?");
```
dliubarskyi:
```suggestion
Result<String> result = assistant.chat("How much is 37 plus 87?");
```

for clarity

docs/docs/tutorials/tools.md
```diff
@@ -676,6 +676,46 @@ Assistant assistant = AiServices.builder(Assistant.class)
...
+```
+
+instead of letting the LLM reprocess the result of the `add` tool execution request, thus returning a response like: `The result of adding 37 and 87 is 124.`
```
dliubarskyi: Please mention that `result.content()` will be `null` in this case and user should check `result.toolExecutions()` instead.

docs/docs/tutorials/tools.md
```diff
@@ -676,6 +676,46 @@ Assistant assistant = AiServices.builder(Assistant.class)
...
```
dliubarskyi, line 723: It is also worth mentioning that if LLM calls multiple tools and at least one of them is not immediate, then reprocessing will happen.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -236,6 +248,29 @@ public ToolServiceResult executeInferenceAndToolsLoop(
                 } else {
                     messages.add(toolExecutionResultMessage);
                 }
+
+                if (immediateToolReturn) {
+                    if (isImmediateTool(toolExecutionRequest.name())) {
+                        if (!isReturnTypeResult) {
+                            throw runtime(
```
dliubarskyi:
```suggestion
                            throw illegalConfiguration(
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -236,6 +248,29 @@ public ToolServiceResult executeInferenceAndToolsLoop(
...
+                                    "Tool '%s' with immediate return is not allowed on a AI service not returning Result.",
+                                    toolExecutionRequest.name());
```
dliubarskyi, line 257: We can (and probably should) check this much earlier, when building AI service. At that point we have all the tools and return type. If `ToolService.immediateReturnTools` is not empty and return type is not `Result`, then fail.

But this can be done in a separate PR.

mariofusco: I was thinking the same, but the drawback of that approach is that the exception will be thrown even if the LLM doesn't actually need to use the tool at all.

dliubarskyi: True, but if user configures such a tool, then there is a high chance of LLM using it at some point. The problem with the current approach is that the issue might be spotted too late in the development process, so I would opt for a fail-fast approach to save user a bit of back and forth

docs/docs/tutorials/tools.md
mariofusco: Done.

docs/docs/tutorials/tools.md
mariofusco: Done.

dliubarskyi (approved):
@mariofusco thanks a lot, great job!

cultcargo:
Hey! Thank you for this pr. Is there any way to use this feature for the dynamically specified tools?

mariofusco:
> Hey! Thank you for this pr. Is there any way to use this feature for the dynamically specified tools?

Not at the moment, I will try to add support to it.

Issue #3016, opened 2025-05-14
How to disable HTTP2.0


JonSmall:
How to disable HTTP2.0
vllm serve not support the header ->Upgrade: h2c
JDK->21
Map<String, String> headerMap = new HashMap<>();
        headerMap.put("Upgrade", "111");
        ChatModel chatModel = OpenAiChatModel.builder()
                .customHeaders(headerMap)
                ...
java.lang.IllegalArgumentException: restricted header name: "Upgrade"
	at java.net.http/jdk.internal.net.http.common.Utils.newIAE(Utils.java:326) ~[java.net.http:na]

Can this be removed:
dev.langchain4j.http.client.jdk.JdkHttpClient -> toJdkRequest
...
        request.headers().forEach((name, values) -> {
            if (values != null) {
                values.forEach(value -> builder.header(name, value));
            }
        });
...

dliubarskyi:
Hi @JonSmall, I guess you are looking for this: https://github.com/langchain4j/langchain4j/issues/2758#issuecomment-2749439972

Please reopen the issue if required

Issue #2882, opened 2025-04-16
[BUG] LLM Studio OpenAI Compatibility API does not work with langchain4j open-ai because stream: false is added


aronsemle:
**Describe the bug**
Using LLM Studio OpenAI Compatibility API mode (https://lmstudio.ai/docs/app/api/endpoints/openai) doesn't work with langchain4j-open-ai because it's injecting stream: false into the JSON payload.

This was introduced in the https://github.com/ai-for-java/openai4j/pull/47 where stream(null) was changed to stream(false) in the builder.

**To Reproduce**
1. Host LLM Studio and enable the local API
2. Use the open-ai builder and set the baseURL

OpenAiChatModel.builder()
         .baseUrl('http://127.0.0.1:1234/v1');
            
3. Send a message and use wire shark. The payload looks like this. Note stream false is included

{
    "model": "meta-llama-3.1-8b-instruct",
    "messages": [{"role":"user", "content":"what is fav color?"}],
    "stream": false
}

4. LLM Studio closes the socket with no response
5. I can reproduce this in postman, and removing stream: false resolves the issue
6. Using the OpenAI SDK directly and overriding the baseURL, stream: false is not included

**Expected behavior**
LLM Studio should respond. I should have a way via the SDK to control stream: ?; either remove it from the payload, set it to true, or set it to false.

**Please complete the following information:**
- LangChain4j version: 1.0.0-beta3
- LLM(s) used: meta-llama-3.1-8b-instruct with LLM Studio
- Java version: 21

**Additional context**
Verified this with wireshark and postman.

dliubarskyi:
Hi @aronsemle, this feels like an issue on the LM Studio side. `"stream": false` is a correct property in OpenAI API, so it should be interpreted by LM Studio without issue if they claim OpenAI API compatibility.

aronsemle:
That's fair. OpenAI can handle stream: false or no stream attribute. The only other point I'd make is if you use OpenAI SDK directly they are not putting stream:false in the payload. That said, I understand if you think this is an LLM Studio issue. I agree they should fix it regardless.

dliubarskyi:
@aronsemle yes, I would first try to raise this issue on the LM Strudio side, it would be great to fix it there.
I will close this issue for now. In case you will not get a fix/help from LM Studio side, feel free to reopen it!

arminio:
For those facing this issue, this is not a problem with streaming, it's the http version.

[...]

PR #3087, opened 2025-05-22
feat : add the HTTP version attribute for requests/responses

codingmiao:
## Related Issues

See discussions: [#3016](https://github.com/langchain4j/langchain4j/issues/3016), [#2882](https://github.com/langchain4j/langchain4j/issues/2882)

## Changes

This PR introduces the following changes:

Adds a `java.net.http.HttpClient.Version version` property to:

* `dev.langchain4j.http.client.HttpRequest`
* `dev.langchain4j.http.client.SuccessfulHttpResponse`

Details:

* Accepts and stores the `version` in constructors.
* Includes the version in HTTP request/response logs.
* Updates the `dev.langchain4j.http.client.jdk` package to:

  * Allow `SuccessfulHttpResponse` to carry and expose the HTTP version.
  * Ensure the version is settable and retrievable from responses.

**Backward compatibility is preserved** — if no version is explicitly set, it defaults to `HttpClient.Version.HTTP_2`, consistent with the current behavior.

---

### Background

When calling a locally deployed vLLM service, the following code was used to create an `OpenAiStreamingChatModel`:

```java
java.net.http.HttpClient.Builder httpClientBuilder = java.net.http.HttpClient.newBuilder();
// httpClientBuilder.version(HttpClient.Version.HTTP_1_1);
JdkHttpClientBuilder jdkHttpClientBuilder = new JdkHttpClientBuilder().httpClientBuilder(httpClientBuilder);
OpenAiStreamingChatModel streamingChatModel = OpenAiStreamingChatModel.builder()
        .baseUrl("http://localhost:9999/v1")
        .apiKey("xxx")
        .modelName("Qwen3-32B")
        .timeout(Duration.ofSeconds(120))
        .logRequests(true)
        .logResponses(true)
        .httpClientBuilder(jdkHttpClientBuilder)
        .build();
streamingChatModel.chat("hello", ...);
```

Since the HTTP version was not explicitly set, Java defaulted to HTTP/2, which caused the following error when calling vLLM:

```bash
dev.langchain4j.http.client.log.LoggingHttpClient 2025-05-22 15:34:40 [DEBUG] - HTTP request:
- method: POST
- url: http://localhost:9999/v1/chat/completions
- headers: [Authorization: Beare...r ], [User-Agent: langchain4j-openai], [Content-Type: application/json]
- body: '{
  "model" : "Qwen3-32B",
  "messages" : [{
    "role" : "user",
    "content" : "hello"
  }],
  "stream" : true
}'

...

dev.langchain4j.exception.HttpException: {"object":"error","message":"[{'type': 'missing', 'loc': ('body',), 'msg': 'Field required', 'input': None}]","type":"BadRequestError","param":null,"code":400}
	at dev.langchain4j.http.client.jdk.JdkHttpClient.lambda$execute$0(JdkHttpClient.java:68)
```

This error message was not very informative, and debugging took a considerable amount of time. Eventually, it was determined that vLLM cannot properly parse HTTP/2 requests — only when explicitly setting the version to `HTTP_1_1` did the request succeed.

---

### Benefits of the Change

1. **Improved Log Transparency**: Users can now clearly see which HTTP version is being used, which helps diagnose similar issues more quickly.
2. **Greater Flexibility and Extensibility**:

   * Version can be configured per-request without requiring a fully custom `HttpClient`.
   * Responses now carry version metadata, enabling version-based conditional logic.
   * Aligns with JDK `HttpRequest`/`HttpResponse` behavior.

---

## Checklist

* [x] Non-breaking change with full backward compatibility
* [x] Added tests covering positive and negative scenarios
* [x] All related module tests pass locally
* [x] All `core` and `main` module tests pass locally


langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
```diff
@@ -1,28 +1,31 @@
...
-import java.util.List;
-import java.util.Map;
-
 public class HttpRequest {
 
     private final HttpMethod method;
     private final String url;
     private final Map<String, List<String>> headers;
     private final String body;
+    private final HttpClient.Version version;
```
dliubarskyi: this is a field specific only for JDK's HttpClient implementation, so it should not be in the generic `HttpRequest` API

codingmiao: Thanks for the feedback!

You're absolutely right — `HttpClient.Version` shouldn't be exposed in the generic `HttpRequest` or `HttpResponse` APIs.

To address this, I followed the pattern used by `dev.langchain4j.http.client.HttpMethod` and introduced a new enum `dev.langchain4j.http.client.HttpVersion`. This abstracts away the direct dependency on the JDK `HttpClient`, making the API more implementation-agnostic.

Please take another look when you have a chance.

http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
```diff
@@ -98,6 +99,16 @@ private java.net.http.HttpRequest toJdkRequest(HttpRequest request) {
         java.net.http.HttpRequest.Builder builder = java.net.http.HttpRequest.newBuilder()
                 .uri(URI.create(request.url()));
 
+        if (request.version() != null) {
+            if (request.version() == HttpVersion.HTTP_1_1) {
+                builder.version(java.net.http.HttpClient.Version.HTTP_1_1);
+            } else if (request.version() == HttpVersion.HTTP_2) {
+                builder.version(java.net.http.HttpClient.Version.HTTP_2);
+            }else {
```
dliubarskyi:
```suggestion
            } else {
```

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpVersion.java
```diff
@@ -0,0 +1,13 @@
+package dev.langchain4j.http.client;
+
+public enum HttpVersion {
```
dliubarskyi, line 3:
```suggestion
/**
 * @since 1.5.0
 */
public enum HttpVersion {
```

dliubarskyi:
> Benefits of the Change
> Improved Log Transparency: Users can now clearly see which HTTP version is being used, which helps diagnose similar issues more quickly.

But this can be done by just adding version information to the logs

> Version can be configured per-request without requiring a fully custom HttpClient.

Why would anyone need this? Moreover, some HTTP clients in Java cannot change the version dynamically for each reqest, so this abstraction is a leaky one

> Responses now carry version metadata, enabling version-based conditional logic.

When can this be helpful?

dliubarskyi:
@codingmiao thank you for your PR, but TBH I am not sure how this change improves things. Version is not propagated from anywhere. `HttpClient` was not designed to be used separately and directly, so having this logic inside `HttpRequest` and response is not really useful as it is now

dliubarskyi:
Closing for now as there is no response. Feel free to reopen if still required

PR #3090, opened 2025-05-22
Allowing caller to pass in class loader for classpath document loading

edeandrea:
## Change

This is an enhancement to the classpath document loading to allow the caller to pass in which class loader to use rather than dependending on the default one. Quarkus in particular needs to do this.

## General checklist

- [ X ] There are no breaking changes
- [ X ] The tests cover both positive and negative cases
- [ X ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ X ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green


dliubarskyi (approved):
@edeandrea thank you!

Issue #1122, opened 2024-05-17
[FEATURE] Option to pass extra parameters to RAG/Tool/XyzProvider via AI service


langchain4j:
TODO
Currently only memory id can be passed around.
It would be great to be able to pass custom parameters.
For example, to be able to set `Filter` expression.
Consider passing `ChatMemory` or `List<ChatMessage>` as well, so that i cna be accessed from tool method.

Yellow--:
how about this?

- add annotation```UserData```which will be used by AiService Interface
- add annotation```ToolUserData```which indicated that relatived parameter in tools is userData
- the userData passed in the AiService method will by bypass to tools and ```MetaData``` of RAG

### Example

```java
    // the definition of tool
    public static class SendMailTools {
        public void sendEmail(String content, @ToolUserData Map<String, Object> userData) {
            String subject = (String) userData.getOrDefault("subject", "default subject");
            log.info("send email with content: {} userData: {}", content, subject);
        }
    }
  
    //the definition of AiService
    public interface ChatService {
        @UserMessage("summary text ```{{text}}``` and send summary by email")
        String summaryAndSendEmail(@V("text") String text, @UserData Map<String, Object> userData);
    }

    //example
    public static void main(String[] args) {
        ChatService chatService = AiServices.builder(ChatService.class)
                .tools(new SendMailTools())
                .build();
        Map<String, Object> userData = Collections.singletonMap("subject", "customizeSubject");
        chatService.summaryAndSendEmail("a very long text", userData);
    }
```

### Things may need to discuss

- does the annotation name understandable?may be need another name?
- does strict the ```UserData``` to Map<String, Object> is good?

@langchain4j

langchain4j:
@Yellow-- Sounds good!

> does the annotation name understandable?may be need another name?

We should consider other names, because this data might not be really a **_user_** data, but something related to the tenant or the whole app

> does strict the UserData to Map<String, Object> is good?

Ideally we shouldn't restrict the type (as it is done for memory id)

Yellow--:
@langchain4j 
with you advices
the final look may be like this?

```java
    // the definition of tool
    public static class SendMailTools {
        public void sendEmail(String content, @ToolTransparentData Map<String, Object> emailContext) {
            String subject = (String) userData.getOrDefault("subject", "default subject");
            log.info("send email with content: {} userData: {}", content, subject);
        }
    }
  
    //the definition of AiService
[...]

PR #3095, opened 2025-05-23
Allow to use AI services as tools for another AI service

mariofusco:
## Issue

At the moment it is not possible to use an AI service as tool for another AI service because the proxy method doesn't carry the annotation of the interface method.

## Change

Improve discovery of the `@Tool` annotation also looking for it on the interfaces implemented by the proxy class.

## General checklist

### DIFF
--- a/docs/docs/tutorials/tools.md
+++ b/docs/docs/tutorials/tools.md
@@ -300,6 +300,85 @@ If the method has a `String` return type, the returned value is sent to the LLM 
 
 For other return types, the returned value is converted into a JSON string before being sent to the LLM.
 
+### AI services as tools for other AI services
+
+AI services can also be used as tools for other AI services. This can be useful in many agentic use cases, where one AI service can ask the help of another, more specialized, AI service to perform a specific task. For instance, having defined the following AI services:
+
+```java
+    interface RouterAgent {
+
+        @dev.langchain4j.service.UserMessage("""
+            Analyze the following user request and categorize it as 'legal', 'medical' or 'technical',
+            then forward the request as it is to the corresponding expert provided as a tool.
+            Finally return the answer that you received from the expert without any modification.
+
+            The user request is: '{{it}}'.
+            """)
+        String askToExpert(String request);
+    }
+
+    interface MedicalExpert {
+
+        @dev.langchain4j.service.UserMessage("""
+            You are a medical expert.
+            Analyze the following user request under a medical point of view and provide the best possible answer.
+            The user request is {{it}}.
+            """)
+        @Tool("A medical expert")
+        String medicalRequest(String request);
+    }
+
+    interface LegalExpert {
+
+        @dev.langchain4j.service.UserMessage("""
+            You are a legal expert.
+            Analyze the following user request under a legal point of view and provide the best possible answer.
+            The user request is {{it}}.
+            """)
+        @Tool("A legal expert")
+        String legalRequest(String request);
+    }
+
+    interface TechnicalExpert {
+
+        @dev.langchain4j.service.UserMessage("""
+            You are a technical expert.
+            Analyze the following user request under a technical point of view and provide the best possible answer.
+            The user request is {{it}}.
+            """)
+        @Tool("A technical expert")
+        String technicalRequest(String request);
+    }
+```
+
+The `RouterAgent` can be configured to use as tools the 3 other AI services, experts in specific fields, routing the user request to one of them.
+
+```java
+MedicalExpert medicalExpert = AiServices.builder(MedicalExpert.class)
+        .chatModel(model)
+        .build();
+LegalExpert legalExpert = AiServices.builder(LegalExpert.class)
+        .chatModel(model)
+        .build();
+TechnicalExpert technicalExpert = AiServices.builder(TechnicalExpert.class)
+        .chatModel(model)
+        .build();
+
+RouterAgent routerAgent = AiServices.builder(RouterAgent.class)
+        .chatModel(model)
+        .tools(medicalExpert, legalExpert, technicalExpert)
+        .build();
+
+routerAgent.askToExpert("I broke my leg what should I do");
+```
+
+:::note
+Using AI services as tools for other AI services is a powerful feature that enables to build complex agentic systems. However, this approach also comes with a few relevant drawbacks that are important to be aware of:
+- This implementation requires the LLM to copy-paste the user request without modifications as a tool call and this could be an error-prone operation.
+- The LLM calling the other LLM as a tool has to reprocess its response, as it happens for any other tool invocation, and this could be a wasteful computation in terms of both time and consumed tokens.
+- The agent-tool, being a totally separated AI service, has no access to the chat memory of the agent calling it, so it cannot use the chat memory to provide a more informed answer.
+:::
+
 ### Exception Handling
 If a method annotated with `@Tool` throws an `Exception`,
 the message of the `Exception` (`e.getMessage()`) will be sent to the LLM as the result of tool's execution.
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
@@ -4,6 +4,9 @@ import dev.langchain4j.Internal;
 
 import java.io.ByteArrayOutputStream;
 import java.io.InputStream;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URL;
@@ -17,6 +20,7 @@ import java.util.HexFormat;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.function.Supplier;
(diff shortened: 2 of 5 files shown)


glaforge:
Not a formal PR review but just a comment to say, YES!, I love the idea and wanted to have something like this.
It's a nice composition of AiServices to implement multi-agent systems!

glaforge:
One remark: it's also a pattern we see in some agent frameworks.
For example, in ADK (Agent Development Kit), you can create an "agent tool".
For (type-safe) composability of AiServices, that'll be handy, and make LangChain4j even better for multi-agent scenarios.

LizeRaes:
This looks great, thanks a lot and like @glaforge says, a frequently used pattern!
I have a question on context-tracking: when building generic or dynamic AiServices I struggle with tracking which AiService called which other AiService or Tool in which conversation. The tool call ID is triggered by the LLM and passes out of sight of the user. How do you deal with this? Is there a way to add tool calls to the AiService memory for example?

glaforge:
Excellent question @LizeRaes!

In ADK, there are different notions of [contexts](https://google.github.io/adk-docs/context/#the-different-types-of-context).
There's a tool context for when a tool is invoked, an invocation context when an agent is called, etc.
Furthermore, ADK has a concept of session, which also holds a state (a map, basically), and (sub)agents can share this state.
Additionally, there are various before/after callbacks: when an LLM is called, when a tool is called, etc.
So it's easier to track what's going on.

Now, to take an example, let's say my LLM calls a tool, in the signature of the tool method, I can add an extra (optional) parameter which is a `ToolContext`.

So maybe the definitions of our `AiService` and `Tool` could/should also add extra parameters to convey the ongoing context?

dliubarskyi:
Related: https://github.com/langchain4j/langchain4j/issues/1122

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +787,72 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+            Analyze the following user request under a technical point of view and provide the best possible answer.
+            The user request is {{it}}.
+            """)
+        @Tool("A technical expert")
+        String technicalRequest(String request);
+    }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    void tools_as_agents_tests(ChatModel model) {
```
dliubarskyi: Please move it to the `AiServicesWithToolsIT`, as it tests not integrations but a general capability of LC4j

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +787,72 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+        TechnicalExpert technicalExpert = spy(AiServices.builder(TechnicalExpert.class)
+                .chatModel(model)
+                .build());
+
+        RouterAgent routerAgent = AiServices.builder(RouterAgent.class)
+                .chatModel(model)
+                .tools(medicalExpert, legalExpert, technicalExpert)
+                .build();
+
+        System.out.println(routerAgent.askToExpert("I broke my leg what should I do"));
```
dliubarskyi: nit: please avoid `System.out.println`

langchain4j/src/test/java/dev/langchain4j/service/common/AbstractAiServiceWithToolsIT.java
```diff
@@ -786,4 +787,72 @@ protected void should_execute_tool_with_uuid_parameter(ChatModel model) {
...
+
+        verify(medicalExpert).medicalRequest("I broke my leg what should I do");
```
dliubarskyi: nit: would also be nice to verify:
- that other experts were not called
- configure `ChatMemory` for `routerAgent` and verify the state of it after the call

dliubarskyi (commented):
@mariofusco good stuff, thank you! It would be great to have this documented 🙏

mariofusco:
> So maybe the definitions of our `AiService` and `Tool` could/should also add extra parameters to convey the ongoing context?

I think that the idea of a shared context between an `AiService` and the `Tool` it uses is very useful, especially if also the tools are `AiService` themselves. Specifically in this last case, in some circumstances, they could also have a shared chat memory.

For sure I will experiment with these ideas, but regarding this pull request I believe that it could be merged as it is and then the other proposed improvements could be built on top of it.

dliubarskyi:
Hi @mariofusco 

> For sure I will experiment with these ideas

If you do, please check discussions in https://github.com/langchain4j/langchain4j/issues/1122, there might be some insights on this topic

> but regarding this pull request I believe that it could be merged as it is and then the other proposed improvements could be built on top of it.

We could merge it as-is, but "undocumented feature is a feature which noone will use"... Another concern is that CI will become slower and more expensive because the new test will run for 13 times instead of 1

mariofusco:
> We could merge it as-is, but "undocumented feature is a feature which noone will use"... Another concern is that CI will become slower and more expensive because the new test will run for 13 times instead of 1

Sorry for misunderstanding, I implicitly meant "merge it as it is AFTER that I've added docs and addressed all your other remarks" :) Will do this asap.

mariofusco:
@dliubarskyi All done, please give it a second look.

docs/docs/tutorials/tools.md
```diff
@@ -300,6 +300,78 @@ If the method has a `String` return type, the returned value is sent to the LLM
...
+            You are a technical expert.
+            Analyze the following user request under a technical point of view and provide the best possible answer.
+            The user request is {{it}}.
+            """)
+        @Tool("A technical expert")
+        String technicalRequest(String request);
+    }
+```
+
+the `RouterAgent` can be configured to use as tools the 3 other AI services, experts in specific fields, routing the user request to one of them.
```
dliubarskyi: nit:
```suggestion
The `RouterAgent` can be configured to use as tools the 3 other AI services, experts in specific fields, routing the user request to one of them.
```

docs/docs/tutorials/tools.md
```diff
@@ -300,6 +300,78 @@ If the method has a `String` return type, the returned value is sent to the LLM
...
+
+AI services can also be used as tools for other AI services. This can be useful in many agentic use cases, where one AI service can ask the help of another, more specialized, AI service to perform a specific task. For instance, having defined the following AI services:
+
+```java
+    interface RouterAgent {
+
+        @dev.langchain4j.service.UserMessage("""
+            Analyze the following user request and categorize it as 'legal', 'medical' or 'technical',
+            then forward the request as it is to the corresponding expert provided as a tool.
+            Finally return the answer that you received from the expert without any modification.
```
dliubarskyi, line 313: I am not sure this (routing) is a good example to include in the documentation, as users might take it as a "go-to recipe", while IMHO it has some big drawbacks.

This implementation requires the LLM to copy-paste the user request (can be very long) without modifications as a tool call, then copy-paste the expert's response (can be very long) without modifications  to the user, so it:
1. consumes way more tokens than it could (latency + money)
2. error-prone because of copy-pasting

Perhaps you have other ideas what can be a good example?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
```diff
@@ -862,4 +862,82 @@ LocalDate currentDate() {
...
+        ChatMemory chatMemory = MessageWindowChatMemory.withMaxMessages(10);
+        RouterAgent routerAgent = AiServices.builder(RouterAgent.class)
+                .chatModel(model)
+                .chatMemory(chatMemory)
+                .tools(medicalExpert, legalExpert, technicalExpert)
+                .build();
+
+        routerAgent.askToExpert("I broke my leg what should I do");
+        verify(medicalExpert).medicalRequest("I broke my leg what should I do");
+        verifyNoMoreInteractions(legalExpert, technicalExpert);
```
dliubarskyi:
```suggestion
        verifyNoInteractions(legalExpert, technicalExpert);
```

dliubarskyi (commented):
@mariofusco thanks a lot!

docs/docs/tutorials/tools.md
mariofusco: I see your point and this is exactly what I discussed here https://quarkus.io/blog/agentic-ai-with-quarkus-p3/#comparing-the-workflow-and-agentic-approaches

However this is quite unavoidable at the moment and I cannot think of any other example that wouldn't suffer of this problem. The tool result is always reprocessed by the LLM who called it, regardless if also the tool is an LLM. The only structural solution to this that I see will be merging that other pull request allowing to directly return the tool's response.

dliubarskyi: I see... returning tool directly can help, but only partially, as input should be copy-pasted. Unless the agent-tool accepts processed input (e.g., only a relevant part).

Another concern is that follow-up questions to the agent-tool will lack the memory, unless we explicitly instruct the main agent to include memory, but it is brittle.

Perhaps worth adding a note/arning into the documentation about the limitations of this approach?

mariofusco: Ok, good point. I will clarify all those limitations in docs.

mariofusco: I added to docs a note reporting the issues that we discussed, please review again.

docs/docs/tutorials/tools.md
```diff
@@ -300,6 +300,85 @@ If the method has a `String` return type, the returned value is sent to the LLM
...
+
+            The user request is: '{{it}}'.
+            """)
+        String askToExpert(String request);
+    }
+
+    interface MedicalExpert {
+
+        @dev.langchain4j.service.UserMessage("""
+            You are a medical expert.
```
kpavlov, line 323: Nit: I would extract system prompt part to separate annotation. In a separate PR

langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
```diff
@@ -396,4 +400,34 @@ public static Map<String, String> toStringValueMap(Map<String, Object> map) {
...
+     * Returns the method eventually annotated with the given annotation.
+     * It could be the method itself or, if the method belongs to a proxy,
+     * a method from one of the interfaces implemented by the proxy.
+     *
+     * @param method The method to check for the annotation.
+     * @param annotation The annotation to look for.
+     * @return An {@link Optional} containing the method having the given annotation,
+     *         or an empty {@link Optional} if there isn't any.
+     */
+    public static Optional<Method> getAnnotatedMethod(Method method, Class<? extends Annotation> annotation) {
```
kpavlov, line 414: nit: would be great to update UtilsTest

kpavlov (dismissed):
Thank you, @mariofusco ! Very useful feature. Make sense to merge as it is

langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
mariofusco: Good suggestion, done.

dliubarskyi (approved):
@mariofusco thanks a lot!

PR #3127, opened 2025-05-29
Fix spotless plugin run on CI and in Makefile

kpavlov:
## Issue
Spotless plugin run is failing on CI and the Makefile was updated after the change: https://github.com/langchain4j/langchain4j/commit/35d8bee5db5e30fbd969a867a5fd29131ad6bf7e

## Change
- Set git clone depth to 0 for r`atchetFrom=origin/main` to work.
- Use spotless maven profile in Makefile

## General checklist


.github/workflows/main.yaml
```diff
@@ -139,6 +139,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
kpavlov: this fixes "No such reference 'origin/main'"

dliubarskyi (approved):
@kpavlov thank you!

PR #3134, opened 2025-05-31
Ollama - Add support for tracking and retrieving capabilities in Ollama models

bidek:
## Change
Add new enum `OllamaCapability` describing model capabilities, for example Vision, Tools ...
Updated `OllamaModelCard` DTO
Add new test

## General checklist


langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaCapability.java
```diff
@@ -0,0 +1,17 @@
...
+    COMPLETION,
+    TOOLS,
+    INSERT,
+    VISION,
+    EMBEDDING,
+    THINKING;
+
+    @JsonCreator
+    public static OllamaCapability fromString(String key) {
+        return key == null ? null : OllamaCapability.valueOf(key.toUpperCase());
```
dliubarskyi: If/when Ollama will add new capabilities, this will fail

dliubarskyi (commented):
@bidek thank you!

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaCapability.java
bidek: Do you have any sugestion how to fix this? Maybe another enum value, for example UNKNOWN_CAP ?

dliubarskyi: I would just use `String` instead of enum

bidek: ok, i change it to list of strings

dliubarskyi (approved):
@bidek thank you!

PR #3162, opened 2025-06-05
Update Metadata Documentation to Include UUID Support

harikrishna553:
### Summary

This PR updates the documentation for the `Metadata` key-value map to include `UUID` as one of the supported value types.

### Details

The current documentation lists the supported value types as:
- String
- Integer
- Long
- Float
- Double

However, `UUID` is already supported in the implementation but missing from the docs.

This update adds `UUID` to the list of supported types in the documentation to improve clarity and accuracy.

This PR does not change any code logic, only updates the documentation.


dliubarskyi (approved):
@harikrishna553 thank you!

PR #3165, opened 2025-06-05
Add blocking utility methods for streaming response handling

### LINKED ISSUE #3159: [FEATURE] Add blocking variants to LambdaStreamingResponseHandler for easier synchronization

Yes. The current `LambdaStreamingResponseHandler` class provides convenient lambda-based methods for handling streaming responses, but the `onCompleteResponse` method is empty with no way to customize it. This makes it difficult to coordinate with streaming operations in scenarios where you need to wait for completion, such as:

- Unit tests that need to verify the complete streaming response
- Simple command-line applications 
- Scripts where the main thread would exit before streaming completes

Current problematic usage:
```java
import static dev.langchain4j.model.LambdaStreamingResponseHandler.onPartialResponse;

// This starts streaming but provides no way to wait for completion
model.chat("Tell me a joke", onPartialResponse(System.out::print));
// Main thread may exit before streaming finishes
```

Add blocking variants to `LambdaStreamingResponseHandler` that handle the complete streaming lifecycle and only return after the response is fully processed:

```java
public static void onPartialResponseBlocking(
        StreamingChatModel model,
        String message,
        Consumer<String> onPartialResponse) throws InterruptedException

public static void onPartialResponseAndErrorBlocking(
        StreamingChatModel model,
        String message,
        Consumer<String> onPartialResponse,
        Consumer<Throwable> onError) throws InterruptedException
```

Usage would be:

```java
import static dev.langchain4j.model.LambdaStreamingResponseHandler.onPartialResponseBlocking;

// This blocks until streaming completes
onPartialResponseBlocking(model, "Tell me a joke", System.out::print);
// Execution continues only after streaming is done
```

1. Adding methods that return a wrapper object with an `awaitCompletion()` method - but this requires users to remember to call the additional method, which doesn't fully solve the coordination problem.
2. Using `CompletableFuture`-based APIs - but this would be inconsistent with the existing lambda-focused design.
3. Expecting users to implement their own `CountDownLatch` coordination - but this defeats the purpose of the convenience methods.

**Additional context**

The implementation would use `CountDownLatch` internally to coordinate completion, calling `countDown()` in both `onCompleteResponse` and `onError` handlers, then blocking on `await()` before returning. This maintains backward compatibility (no changes to existing methods) while preserving the clean lambda-based API design that LangChain4j is known for.

I'm happy to implement this enhancement (with comprehensive unit tests) if the maintainers think it would be valuable for the project.

kousen:
- Introduced `onPartialResponseBlocking` and `onPartialResponseAndErrorBlocking` methods in `LambdaStreamingResponseHandler`.
- Added corresponding tests to ensure proper handling of streaming completion, errors, and interruptions.

## Issue

Closes #3159

## Change

Added blocking variants to the `LambdaStreamingResponseHandler` class to address the coordination problem where applications need to wait for streaming completion. The existing methods provide no way to block until streaming is complete, causing issues in tests and simple applications where the main thread might exit before streaming finishes.

**New methods added:**
- `onPartialResponseBlocking(StreamingChatModel model, String message, Consumer<String> onPartialResponse)` - Blocks until streaming completes, using default error handling
- `onPartialResponseAndErrorBlocking(StreamingChatModel model, String message, Consumer<String> onPartialResponse, Consumer<Throwable> onError)` - Blocks until streaming completes with custom error handling

**Implementation details:**
- Uses `CountDownLatch` internally to coordinate completion
- Calls `countDown()` in both `onCompleteResponse` and `onError` handlers
- Blocks on `await()` before returning
- Maintains backward compatibility (no changes to existing methods)
- Preserves the clean lambda-based API design

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/model/LambdaStreamingResponseHandler.java
```diff
@@ -22,6 +24,11 @@
  *       onPartialResponse(System.out::println);
  * model.chat("Why is the sky blue?",
  *       onPartialResponseAndError(System.out::println, Throwable::printStackTrace));
+ *
+ * // Blocking (new methods)
```
glaforge: Maybe remove the `(new methods)` part of this line, as it won't age well

langchain4j-core/src/test/java/dev/langchain4j/model/LambdaStreamingResponseHandlerTest.java
```diff
@@ -5,12 +5,14 @@
...
+import org.junit.jupiter.api.Timeout;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
-import static dev.langchain4j.model.LambdaStreamingResponseHandler.onPartialResponse;
-import static dev.langchain4j.model.LambdaStreamingResponseHandler.onPartialResponseAndError;
+import static dev.langchain4j.model.LambdaStreamingResponseHandler.*;
```
glaforge: We tend to not use wildcards in the codebase, so individual static imports would be better.

glaforge (dismissed):
This is a useful improvement, Ken, thanks for this contribution!
(I have two minor nits on static imports and a JavaDoc comment, but that's just a nit)

glaforge:
Looks like spotless is unhappy. Could you please run `mvn spotless:apply`?

langchain4j-core/src/main/java/dev/langchain4j/model/LambdaStreamingResponseHandler.java
kousen: Right. I removed it.

langchain4j-core/src/test/java/dev/langchain4j/model/LambdaStreamingResponseHandlerTest.java
kousen: I just replaced the * with explicit methods for the static import.

kousen:
I tried running `mvn spotless:apply' again. Hopefully this took care of the problem. Not sure why I'm having such trouble with it, but oh well.

glaforge (approved):
Spotless is happy and builds are passing!

glaforge:
Thank you @kousen 🥇

kousen:
Sweet! That's my first contribution to LC4j, and my first open source contribution anywhere since my single contribution to Groovy sometime around 2009. I'm sure you remember that like it was yesterday, mostly because I recall I broke the build in the process :)

(It turns out my contribution was adding a JsonBuilder to a Groovlet in March, 2012. Been a while.)

glaforge:
Let's not wait for another 15 years before your next open source contribution :-)

PR #3172, opened 2025-06-06
Classpath document loading - relative path not working

edeandrea:
While working on https://github.com/quarkiverse/quarkus-langchain4j/issues/1079 I discovered the way we were referencing classpath roots wasn't correct here, so this fixes that.


dliubarskyi (approved):
@edeandrea thank you!

PR #3188, opened 2025-06-09
Gemini: Fix a test whose assertion is a bit too strict.

glaforge:


glaforge (commented):
Build broken as a test is a bit too strict.

dliubarskyi (approved):
@glaforge thank you!

PR #3202, opened 2025-06-11
Google AI Gemini: use HttpClient abstraction

dliubarskyi:
## Change

Migrated `langchain4j-google-ai-gemini` from using JDK's `HttpClient` to using `dev.langchain4j.http.client.HttpClient`.
This simplifies the implementation of the module and allows using other HTTP clients and better integrate into frameworks.
This PR also fixes the issue with streaming. Now streaming response is handled in a separate thread (previously streaming was handled in the same thread which led to issues like frozen UI).

Exceptions are now also mapped using default exception mapper.

## General checklist


langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiService.java
```diff
@@ -51,94 +62,75 @@ GoogleAiBatchEmbeddingResponse batchEmbed(String modelName, String apiKey, Googl
...
 
-    Stream<GeminiGenerateContentResponse> generateContentStream(String modelName, String apiKey, GeminiGenerateContentRequest request) {
+    void generateContentStream(String modelName,
+                               String apiKey,
+                               GeminiGenerateContentRequest request,
+                               boolean includeCodeExecutionOutput,
+                               StreamingChatResponseHandler handler) {
         String url = String.format("%s/models/%s:streamGenerateContent?alt=sse", GEMINI_AI_ENDPOINT, modelName);
-        return streamRequest(url, apiKey, request, GeminiGenerateContentResponse.class);
+        streamRequest(url, apiKey, request, includeCodeExecutionOutput, handler);
```
glaforge, line 71: Which reminds me that we've just added thinking support. Should we have an `includeThoughts` parameter?

glaforge (dismissed):
Looks great!
Using the LC4j HTTP client streamlines the code.
No particular feedback.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiService.java
dliubarskyi: Perhaps, but let's keep it out of this PR please

PR #3241, opened 2025-06-20
Add mcp server health check

### LINKED ISSUE #3212: [FEATURE] Implement Scheduled Health Check for MCP Service

Currently, the DefaultMcpClient is passive regarding connection health. It only detects a disconnection when an operation is actively performed and fails, or when the underlying transport layer explicitly reports a failure.

I propose implementing an optional, proactive health-checking mechanism within the DefaultMcpClient. This feature would provide a self-healing capability for more resilient long-lived connections.

The key aspects of the solution are:

Background Health Checks: Introduce a ScheduledExecutorService that runs checkHealth() periodically in the background.
Automatic Reconnection: If the health check (ping) fails, it automatically triggers the client's internal reconnection logic to re-establish the session.
Configurability: This feature should be fully configurable via the Builder, allowing users to:
Enable or disable the automatic health check (defaulting to disabled for backward compatibility).
Configure the interval of the health checks (e.g., every 30 seconds).

**Additional context**
Below are the primary proposed code changes to implement this feature.

1.Add new fields and logic to DefaultMcpClient:
```java
        autoHealthCheckEnabled = getOrDefault(builder.autoHealthCheckEnabled, Boolean.TRUE);
        autoHealthCheckInterval = getOrDefault(builder.autoHealthCheckInterval, Duration.ofSeconds(30));
        healthCheckScheduler = Boolean.TRUE.equals(autoHealthCheckEnabled)
                ? Executors.newSingleThreadScheduledExecutor()
                : null;
....
        transport.onFailure(() -> {
            try {
                TimeUnit.MILLISECONDS.sleep(reconnectInterval.toMillis());
            } catch (InterruptedException e) {
            }
            log.info("Trying to reconnect...");
            triggerReconnection();
        });
        initialize();
        startAutoHealthCheck();
```
2. Implement the new private helper methods in DefaultMcpClient
```java
       private void startAutoHealthCheck() {
        if (Boolean.FALSE.equals(autoHealthCheckEnabled)) {
            return;
        }
        Runnable healthCheckTask = () -> {
            try {
                checkHealth();
            } catch (Exception e) {
                log.warn("mcp server health check failed. Attempting to reconnect...", e);
                triggerReconnection();
            }
        };
        healthCheckScheduler.scheduleAtFixedRate(
                healthCheckTask,
                autoHealthCheckInterval.toMillis(),
                autoHealthCheckInterval.toMillis(),
                TimeUnit.MILLISECONDS
        );
    }

    private void triggerReconnection() {
        if (initializationLock.tryLock()) {
            try {
                initialize();
            } finally {
                initializationLock.unlock();
            }
        }

    }
```

Cooosin:
## Issue

Closes #3212

## Change

Add a scheduled task in the client to perform heartbeat detection on the MCP server

## General checklist

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -34,11 +34,14 @@ import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.concurrent.locks.ReentrantLock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -70,6 +73,10 @@ public class DefaultMcpClient implements McpClient {
     private final AtomicReference<CompletableFuture<Void>> toolListUpdateInProgress = new AtomicReference<>(null);
     private final Duration reconnectInterval;
     private volatile boolean closed = false;
+    private final Boolean autoHealthCheck;
+    private final Duration autoHealthCheckInterval;
+    private final ScheduledExecutorService healthCheckScheduler;
+    private final ReentrantLock initializationLock = new ReentrantLock();
 
     public DefaultMcpClient(Builder builder) {
         transport = ensureNotNull(builder.transport, "transport");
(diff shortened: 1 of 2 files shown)


Cooosin:
Hi @jmartisk 

Just wanted to gently follow up on this PR. Please let me know if there's any feedback or anything I can do to help move this forward.

﻿Thanks for your time!

jmartisk (approved):
I think this makes sense, thanks! Let's see what @dliubars thinks

dliubarskyi:
Thanks! This sounds reasonable, although I am wondering how this (eager) approach compares to e.g., lazy approach (when reconnect happens only if the normal request to the server fails)?

Cooosin:
@dliubarskyi  Thanks for raising this key question! It's indeed a trade-off between two strategies.

The biggest pain point of the "lazy" approach is how it handles "silent disconnect" scenarios. In this situation, a user's request is sent but then hangs, only failing after a long read timeout. This creates a poor user experience with a significant perceived stall.

The primary advantage of the "eager" (heartbeat) approach is its fast failure detection then reconnect .This ensures every user request gets a responsive experience.

Therefore, this feature is also can configurable, allowing users to select the optimal approach for their specific needs.

dliubarskyi:
@Cooosin makes total sense, thanks for clarifying!

dliubarskyi:
Hi @Cooosin, `McpAutoHealthCheckIT.shouldDetectDeadServerByDefault()` is failing on CI:

> McpAutoHealthCheckIT.shouldDetectDeadServerByDefault:77->executeAToolAndAssertSuccess:86 » Runtime java.util.concurrent.ExecutionException: java.lang.RuntimeException: Unexpected status code: 400

Could you please check what is wrong?

dliubarskyi:
Here is the failing job: https://github.com/langchain4j/langchain4j/actions/runs/16104844279/job/45438740259

Cooosin:
> Hi @Cooosin, `McpAutoHealthCheckIT.shouldDetectDeadServerByDefault()` is failing on CI:
> 
> > McpAutoHealthCheckIT.shouldDetectDeadServerByDefault:77->executeAToolAndAssertSuccess:86 » Runtime java.util.concurrent.ExecutionException: java.lang.RuntimeException: Unexpected status code: 400
> 
> Could you please check what is wrong?

I think it was because the local mcp server took longer than 500 milliseconds to create on the way back from destruction. I'll mention a PR later and increase the wait time from 500 milliseconds to 5000 milliseconds. Like here dev/LANGCHAIN4J/MCP/client/integration/mcpreconnectit. Java: 48

Issue #3262, opened 2025-06-26
[FEATURE] Using McpClient in Low Level Tool API


doukai:
The [MCP related documents](https://docs.langchain4j.dev/tutorials/mcp) show examples of using AIService to call MCP, but there is no mention of how to use MCP in the [Low Level Tool API](https://docs.langchain4j.dev/tutorials/tools#low-level-tool-api). Could you provide some tips? Thank you very much

jmartisk:
Right, I'll probably look at adding something into the docs, but it inevitably gets a bit clunky. You'll need to build a JSON document that captures the arguments.

You can use the instance of `DefaultMcpClient` and call it like this:

```
// to obtain descriptions of available tools
List<ToolSpecification> tools = mcpClient.listTools(); 

// to execute a tool named "tool1" with argument "a=b"
ToolExecutionRequest request = ToolExecutionRequest.builder()
                .name("tool1")
                .arguments("{\"a\": \"b\"}")
                .build();
String toolResult = mcpClient.executeTool(request);
```

jmartisk:
Or, if you want to, instead of calling the tool directly, add it to a `ChatRequest`, then you would simply build a `ChatRequest` like

```
ChatRequest request = ChatRequest.builder()
    .messages(UserMessage.from("What will the weather be like in London tomorrow?"))
    .toolSpecifications(toolSpecifications)
    .build();
```

where `toolsSpecifications` is a (potentially filtered) list obtained by calling `mcpClient.listTools()` from my previous comment

jmartisk:
Let me know if this is helpful, if yes, I can try to convert that into a piece of documentation

doukai:
> Or, if you want to, instead of calling the tool directly, add it to a `ChatRequest`, then you would simply build a `ChatRequest` like或者，如果您愿意，而不是直接调用该工具，而是将其添加到 `ChatRequest`，然后您只需构建一个 `ChatRequest`，例如
> 
> ```
> ChatRequest request = ChatRequest.builder()
>     .messages(UserMessage.from("What will the weather be like in London tomorrow?"))
>     .toolSpecifications(toolSpecifications)
>     .build();
> ```
> 
> where `toolsSpecifications` is a (potentially filtered) list obtained by calling `mcpClient.listTools()` from my previous comment其中 `toolsSpecifications` 是通过从我之前的评论中调用 `mcpClient.listTools（）` 获得的（可能经过筛选的）列表

This is very useful, thanks!
[...]

PR #3263, opened 2025-06-26
Add some docs about using MCP without AI services

jmartisk:
Follow-up to the discussion in https://github.com/langchain4j/langchain4j/issues/3262


dliubarskyi (approved):
@jmartisk thank you!

Issue #3211, opened 2025-06-13
Improve JSON extraction from LLM response


mariofusco:
The regexp based implementation used [here](https://github.com/langchain4j/langchain4j/blob/efde65543584c62d2f227044ff9278dda62dc833/langchain4j/src/main/java/dev/langchain4j/service/output/PojoOutputParser.java#L146) to extract a JSON from the LLM response is not fully satisfying. 

In particular it fails when you have a first spurious `{` or a `[`  in the LLM response not-related to the actual JSON that anyway can be present later in the same response, but in this case cannot be correctly recognized.

This can be made more robust and also cover this scenario. The same implementation could be used in the output guardrail used for the same purpose when it will be merged.

glaforge:
I had proposed at some point a lightweight JSON parser (I had written back in the day for Groovy) that I had tweaked to ignore non JSON starting characters (curly & square brackets), and start decoding from there. 

The spurious brace might be an issue, but it could be bypassed by finding the longest JSON sequence potentially.

mariofusco:
Implemented by https://github.com/langchain4j/langchain4j/pull/3268

PR #3268, opened 2025-06-27
Improve JSON extraction from LLM response

mariofusco:
## Issue

Closes #
https://github.com/langchain4j/langchain4j/issues/3211

## Change

The current JSON extraction from the LLM response fails when you have a first spurious { or a [ not-related to the actual JSON that anyway can be present later in the same response, but in this case cannot be correctly recognized.

This pull request makes the json extraction more robust and also cover this scenario. The same implementation is also used in the output guardrail.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
@@ -1,0 +1,74 @@
+package dev.langchain4j.internal;
+
+import dev.langchain4j.Internal;
+
+import java.util.Optional;
+
+@Internal
+public class JsonParsingUtils {
+
+    @FunctionalInterface
+    public interface ThrowingFunction<T, R> {
+        R apply(T t) throws Exception;
+    }
+
+    public record ParsedJson<T>(T value, String json) { }
+
+    public static <T> Optional<ParsedJson<T>> extractAndParseJson(String text, Class<T> type) {
+        return extractAndParseJson(text, s -> Json.fromJson(s, type));
+    }
+
+    public static <T> Optional<ParsedJson<T>> extractAndParseJson(String text, ThrowingFunction<String, T> parser) {
+        try {
+            return Optional.of(new ParsedJson<>(parser.apply(text), text));
+        } catch (Exception ignored) {
+            // Ignore parsing errors and try to find a JSON block in the text
+        }
+
+        int index = text.length();
+        while (true) {
+
+            int jsonEnd = findJsonEnd(text, index);
+            if (jsonEnd < 0) {
+                return Optional.empty();
+            }
+
+            int jsonStart = findJsonStart(text, jsonEnd, text.charAt(jsonEnd));
+            if (jsonStart < 0) {
+                return Optional.empty();
+            }
+
+            try {
+                String tentativeJson = text.substring(jsonStart, jsonEnd + 1);
+                return Optional.of(new ParsedJson<>(parser.apply(tentativeJson), tentativeJson));
+            } catch (Exception ignored) {
+                // If parsing fails, try to extract a JSON block from the text
+            }
+            index = jsonStart; // Move to the next character after the found JSON block
+        }
+    }
+
+    private static int findJsonEnd(String text, int fromIndex) {
+        int jsonMapEnd = text.lastIndexOf('}', fromIndex);
+        int jsonListEnd = text.indexOf(']', fromIndex);
+        return Math.max(jsonMapEnd, jsonListEnd);
+    }
+
+    private static int findJsonStart(String text, int jsonEnd, char closingBrace) {
+        char openingBrace = closingBrace == '}' ? '{' : '[';
+        int braceCount = 0;
+
+        for (int i = jsonEnd; i >= 0; i--) {
+            char c = text.charAt(i);
+            if (c == openingBrace) {
+                braceCount++;
+                if (braceCount == 0) {
+                    return i == 0 || text.charAt(i-1) != openingBrace ? i : -1;
+                }
+            } else if (c == closingBrace) {
+                braceCount--;
+            }
+        }
+        return -1;
+    }
+}
--- a/langchain4j-core/src/test/java/dev/langchain4j/guardrail/JsonExtractorOutputGuardrailTests.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/guardrail/JsonExtractorOutputGuardrailTests.java
@@ -2,9 +2,7 @@ package dev.langchain4j.guardrail;
 
 import static dev.langchain4j.test.guardrail.GuardrailAssertions.assertThat;
 import static org.mockito.Mockito.any;
-import static org.mockito.Mockito.anyString;
 import static org.mockito.Mockito.eq;
-import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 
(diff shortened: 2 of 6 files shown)


langchain4j-core/src/main/java/dev/langchain4j/internal/OutputParsingException.java
```diff
@@ -1,4 +1,4 @@
-package dev.langchain4j.service.output;
+package dev.langchain4j.internal;
```
dliubarskyi: this is a breaking change

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
```diff
@@ -0,0 +1,88 @@
+package dev.langchain4j.internal;
+
+import java.lang.reflect.Type;
+
+import static dev.langchain4j.internal.Utils.quoted;
+
+public class JsonParsingUtils {
```
dliubarskyi, line 8:
```suggestion
@Internal
public class JsonParsingUtils {
```

dliubarskyi (commented):
@mariofusco thanks a lot!

This change requires good testing, could you please add tests?
This will also help to document what issues with JSON are covered by the current implementation

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
```diff
@@ -0,0 +1,88 @@
...
+        try {
+            return new ParsedJson<>(parser.apply(text), text);
+        } catch (Exception ignored) {
+            parseException = ignored;
+        }
+
+        int index = 0;
+        while (true) {
+
+            int jsonStart = findJsonStart(text, index);
```
glaforge: The approach taken here will find the first JSON object or array in the text, correct?
I'm wondering if, instead, we could search for the longest JSON object or array?

I can imagine an output that says here's your {"result":...} object, and then shows a fenced markdown with the proper JSON object. So you'd find the first abbreviated valid or invalid JSON object, but not necessarily the longest one which could be the real JSON output of the model.

This is maybe a bit of a hypothetical situation, a bit of stretch, I reckon, but maybe it'd be more solid?

mariofusco: Checking length could be a possible heuristic, but it seems quite arbitrary. What I'm doing now is checking all the possible json blocks one after the other trying to deserialize it into the required type, until this deserialization succeeds. I find it very unlikely (or however it never happened to me) that you could have a json block that could be unrmashalled in the object that you expect and still not being the right one. And if when this happens I don't know how correct would be to simply disambiguate on the length.

Conversely a think that I noticed quite often, is that the required json is toward the end of the LLM response, in many cases it is the very last thing after the reasoning explaining how it get there. So probably a better heuristic would be scanning the response in reverse order and find the last json block that can be deserialized into the required type?

langchain4j-core/src/main/java/dev/langchain4j/internal/OutputParsingException.java
mariofusco: You're right I will revert this change and add a strategy to throw different exceptions.

langchain4j-core/src/test/java/dev/langchain4j/guardrail/JsonExtractorOutputGuardrailTests.java
```diff
@@ -44,15 +44,27 @@ void successfulValidation(String json, JsonExtractorOutputGuardrail<?> guardrail
...
     void successfulValidationAfterTrimming(
             String json, JsonExtractorOutputGuardrail<?> guardrail, Object expectedResult) {
-        var guardrailSpy = spy(guardrail);
         var input = "abc" + json;
+        parseJsonRequiringTrimming(json, guardrail, expectedResult, input);
+    }
+
+    @ParameterizedTest
+    @MethodSource("guardrails")
+    void successfulValidationAfterTrimmingWithInvalidJson(
```
mariofusco, line 58: This test covers the use case that was broken without this change.

mariofusco:
> This change requires good testing, could you please add tests? This will also help to document what issues with JSON are covered by the current implementation

Since the new parsing algorithm is now used also by the output guardrails extracting the json from the LLM response, I believe that testing is already quite well covered by guardrails tests. There I added one more test demonstrating this improvement at work, extracting the json from a String that before this improvement broke the parser. Do you have any more specific tests in mind that I could add?

mariofusco:
@dliubarskyi I reverted the backward incompatibility that I introduced, let me know if it is ok now.

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingException.java
```diff
@@ -0,0 +1,10 @@
+package dev.langchain4j.internal;
```
dliubarskyi:
```suggestion
package dev.langchain4j.exception;
```
Or is it meant to be used only internally?

mariofusco:
@dliubarskyi I'm now parsing the LLM response backward from the end as I suggested. This should be ready to be merged now.

langchain4j/src/main/java/dev/langchain4j/service/output/ParsingUtils.java
```diff
@@ -97,11 +97,11 @@ private static <T> T parse(String text, Function<String, T> parser, String type)
         }
     }
 
-    static OutputParsingException outputParsingException(String text, Type type) {
+    public static OutputParsingException outputParsingException(String text, Type type) {
```
dliubarskyi:
```suggestion
    static OutputParsingException outputParsingException(String text, Type type) {
```

langchain4j/src/main/java/dev/langchain4j/service/output/ParsingUtils.java
```diff
@@ -97,11 +97,11 @@ private static <T> T parse(String text, Function<String, T> parser, String type)
...
         return outputParsingException(text, type.getTypeName(), null);
     }
 
-    static OutputParsingException outputParsingException(String text, String type, Throwable cause) {
+    public static OutputParsingException outputParsingException(String text, String type, Throwable cause) {
```
dliubarskyi:
```suggestion
    static OutputParsingException outputParsingException(String text, String type, Throwable cause) {
```

langchain4j-core/src/main/java/dev/langchain4j/exception/JsonParsingException.java
```diff
@@ -0,0 +1,8 @@
+package dev.langchain4j.exception;
+
+public class JsonParsingException extends LangChain4jException {
```
dliubarskyi: Is this new exception type necessary? Seems like one could do without it

dliubarskyi (commented):
@edeandrea could you please review as well? 🙏

langchain4j-core/src/main/java/dev/langchain4j/exception/JsonParsingException.java
mariofusco: I introduced it because I couldn't reuse the `OutputParsingException` one since it is in another module. It really depends on how you want to report the fact that it wasn't possible to extract a json from the provided string. I believed that we wanted to report this with an exception, or at least this is what it did so far. The alternative would be to return an `Optional`, I'm also fine with this, do you prefer it?

dliubarskyi: In the `JsonExtractorOutputGuardrail` case, the `JsonParsingException` can be thrown, it is catched and `Optional.empty()` is returned.
In the `PojoOutputParser` case, the `JsonParsingException` can be thrown, it is catched and is converted into `OutputParsingException`.

It seems that one can remove the `JsonParsingException` class and do without it, WDYT?
```java
        try {
            return extractAndParseJson(text, type).value();
        } catch (Exception e) {
            throw new OutputParsingException(e.getMessage(), e.getCause());
        }
```

mariofusco: Agreed, I removed that exception, please review again.

dliubarskyi (approved):
@mariofusco thanks a lot!

PR #3278, opened 2025-06-30
Provide JSON serialize/deserialize to ChatMemory

loicmathieu:
Provide `fromJson()` and `serializeToJson()` for `MessageWindowCharMemory` and `TokenWindowChatMemory`

## Issue

Closes #

## Change
The `InMemoryEmbedingStore` provides convenient methods for serializing to a String and being constructed from a JSON String.
There is none in the `MessageWindowChatMemory` nor the `TokenWindowChatMemory`, when using them and wanted to persist them as JSON, I had to search across all examples to discover that you can easily serialize/deserialize chat messages using `ChatMessageDeseralizer` and `ChatMessageSerializer`.

Bringing the same `serializeToJson()` and `fromJson()` methods to `MessageWindowChatMemory` and `TokenWIndowChatMemory` would be a nice to have convenience for developers for quick prototyping.

_The current implementation lack tests and support of serializing/deserializing to files to avoid loosing time in case it would not be accepted, I would add them if we agree to proceed on it before mering._

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/spi/memory/store/InMemoryChatMemoryStoreJsonCodecFactory.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/spi/memory/store/InMemoryChatMemoryStoreJsonCodecFactory.java
@@ -1,0 +1,9 @@
+package dev.langchain4j.spi.memory.store;
+
+import dev.langchain4j.Internal;
+import dev.langchain4j.store.memory.chat.InMemoryChatMemoryStoreJsonCodec;
+
+@Internal
+public interface InMemoryChatMemoryStoreJsonCodecFactory {
+    InMemoryChatMemoryStoreJsonCodec create();
+}
--- a/langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/InMemoryChatMemoryStore.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/store/memory/chat/InMemoryChatMemoryStore.java
@@ -1,11 +1,16 @@
 package dev.langchain4j.store.memory.chat;
 
 import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.ChatMessageDeserializer;
+import dev.langchain4j.data.message.ChatMessageSerializer;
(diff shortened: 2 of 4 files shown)


kpavlov:
Thank you, @loicmathieu 
What would you think about separating persistence mechanism from chat memory itself? I see the usecase of persisting to RDBMS, files and NoSql databases like DynamoDB or MongoDb. JSON serialization would only be needed for persisting to file storage, maybe

loicmathieu:
@kpavlov of course it could be done, this was only to provide convenience methods as existed in the embedding store which have a memory impl with builtin JSON to String (or File) serialization capability.

Even with such simple JSON capability, implementing a NoSQL persistent memory would be easier as you would use it for creating the document to store.

That being said, if you think we should not mimic embedding store here and provide something more complex I'm OK with that, just need some design first to know what I would be implementing ;)

loicmathieu:
@kpavlov please advise.

ChatMemory is already a kind of persistence layer, which currently is only implemented by an InMemory ChatMemory.
Do you want me to add another layer of interface separating pure ChatMemory responsibility with persistent of the messages? In this case I would need some help to design the interfaces.

My original PR was inspired by the InMemoryEmbeddingStore which provides methods to serialize the items, it was just a way to provides ready to use methods for serialization so a user can easily craft it's own persistence mechanism.
We do plan to also implement memory stored in a persistent store like Redis later so we can work together to design this persistence layer for memory and maybe provide some implementation.

kpavlov:
ChatMemory is an interface, providing a conversation history. So, different ChatMemory implementations might implement different persistent mechanisms such as in-memory persistence, rdbms or local file persistence with json encoding.

loicmathieu:
@kpavlov it took me some time to understand what you are asking because I based my code on the old serialization implementation of InMemoryEmbeddingStore that didn't provide a factory.

I switched to a factory approach and move my code to the InMemoryChatMemoryStore, it fit there more elegantly but this class is not used by default.
I opened a discussion in Discord to better understand the current implementation and discuss possible changes.

Meanwhile, if what I do makes sense, I'm happy to finish the implementation by adding tests and documentation.
InMemoryEmbeddingStore also provides methods to serialize/deserialize to a file, should I also add them?

PR #3279, opened 2025-06-30
[minor fix] Remove unnecessary imports

vesense:
## Issue

Closes #

## Change

[minor fix] Remove unnecessary imports

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/model/output/Response.java
```diff
@@ -121,12 +122,11 @@ public int hashCode() {
 
     @Override
     public String toString() {
-        return "Response {" +
-                " content = " + content +
```
kpavlov, line 125: The style where key and value are on the same line is better: there is a single responsibility for each line

kpavlov:
Thank you, @vesense 
Why not add `@NullMarked` annotation to the package and annotate all nullable fields explicitly?

langchain4j-core/src/main/java/dev/langchain4j/model/output/Response.java
vesense: Hi @kpavlov This code is automatically formatted by  `./mvnw spotless:apply`

vesense: The `./mvnw spotless:check` check will be failed when key and value are on the same line.

> The following files had format violations:
> ```
>         return "Response {" +
>                 " content = " + content +
>                 ", tokenUsage = " + tokenUsage +
>                 ", finishReason = " + finishReason +
>                 ", metadata = " + metadata +
>                 " }";
> ```
>  Run 'mvn spotless:apply' to fix these violations.

vesense:
Hi @kpavlov , thank you for your suggestion. Updated.

vesense:
Hi @kpavlov, can you take a look?

vesense:
Hi @dliubarskyi, can you take a look at this minor fix?

dliubarskyi (commented):
@vesense thank you, it is fine to remove unnecessary imports, but let's not bring in those annotations please

kpavlov (dismissed):

vesense:
@kpavlov hi, I updated the branch and all tests passed.

kpavlov (approved):

PR #3283, opened 2025-07-01
MCP: roots support

jmartisk:


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/ClientMethod.java
```diff
@@ -24,5 +24,7 @@ public enum ClientMethod {
     @JsonProperty("prompts/list")
     PROMPTS_LIST,
     @JsonProperty("prompts/get")
-    PROMPTS_GET
+    PROMPTS_GET,
+    @JsonProperty("notifications/roots/list_changed")
+    NOTIFICATION_ROOT_LIST_CHANGED
```
dliubarskyi:
```suggestion
    NOTIFICATION_ROOTS_LIST_CHANGED
```
nit

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/RootListChangedNotification.java
```diff
@@ -0,0 +1,13 @@
+package dev.langchain4j.mcp.client.protocol;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+public class RootListChangedNotification extends McpClientMessage {
```
dliubarskyi:
```suggestion
public class McpRootsListChangedNotification extends McpClientMessage {
```
nit for consistency

dliubarskyi (commented):
@jmartisk thank you!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/protocol/RootListChangedNotification.java
jmartisk: Unfortunately we're already kinda inconsistent, some classes in `dev.langchain4j.mcp.client.protocol` don't start with `Mcp` while most others do. Shall I rename those too? (They are public, but not really meant to be used by developers)

dliubarskyi: Is `RootListChangedNotification` meant to be used by users?

Perhaps we could mark internal but public classes with `@Internal`?

jmartisk: > Is `RootListChangedNotification` meant to be used by users?

No it's not. 

> Perhaps we could mark internal but public classes with `@Internal`?

Ah, I wasn't aware of this annotation. In that case, let's put it on all classes in `dev.langchain4j.mcp.client.protocol` and rename them to start with `Mcp*`, ok?

dliubarskyi: Soulds good, one can also add `internal` to the package path so that it is clear which classes are not meant to be used by users

jmartisk: Done

dliubarskyi: Thanks!

PR #3287, opened 2025-07-01
Remove the need for Azure dependencies at compile time

jdubois:
Moving the imports to the Azure Identity library to another class removes the need to add this library as a dependency at compile time (it wasn't required at build time, but requiring at compile time is annoying for people not using Azure).

### DIFF
--- a/langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/AzureInternalOpenAiOfficialHelper.java
+++ b/langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/AzureInternalOpenAiOfficialHelper.java
@@ -1,0 +1,14 @@
+package dev.langchain4j.model.openaiofficial;
+
+import com.azure.identity.AuthenticationUtil;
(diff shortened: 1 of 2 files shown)


langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/AzureInternalOpenAiOfficialHelper.java
```diff
@@ -0,0 +1,14 @@
+package dev.langchain4j.model.openaiofficial;
+
+import com.azure.identity.AuthenticationUtil;
+import com.azure.identity.DefaultAzureCredentialBuilder;
+import com.openai.credential.BearerTokenCredential;
+import com.openai.credential.Credential;
+
+public class AzureInternalOpenAiOfficialHelper {
```
dliubarskyi:
```suggestion
class AzureInternalOpenAiOfficialHelper {
```

dliubarskyi (commented):
@jdubois thank you!

dliubarskyi (approved):

PR #3297, opened 2025-07-03
Vertex AI Gemini: fix streaming model: introduce Executor to make it async

### LINKED ISSUE #3288: [BUG] Streaming issue: Google VertexAI Gemini
**Describe the bug**

There is an issue with streaming when using Gemini VertexAI integration. Partial tokens are accumulated, and only the final response is returned, freezing the entire UI until the `onCompleteResponse` callback is invoked. I'm not sure but it seems to be the same problem that was fixed [here](https://github.com/langchain4j/langchain4j/pull/3202).

- LangChain4j version: 1.1.0
- LLM(s) used: Google Vertex AI Gemini
- Java version: 21
- Spring Boot version (if applicable): 3.4.5

@glaforge 
@dliubarskyi

douglas-DS:
## Issue
Fixes #3288

## Change
- Added an `Executor` call in the `generate` method to improve asynchronous task handling by using a thread executor.
- Update logger target class for accurate logging context.
- Remove unnecessary null check for `allowedFunctionNames` property.
- Fixed empty arguments in function call.

## General checklist


langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModel.java
```diff
@@ -52,7 +55,7 @@
  */
 public class VertexAiGeminiStreamingChatModel implements StreamingChatModel, Closeable {
 
-    private static final Logger logger = LoggerFactory.getLogger(VertexAiGeminiChatModel.class);
+    private static final Logger logger = LoggerFactory.getLogger(VertexAiGeminiStreamingChatModel.class);
```
kpavlov, line 67: 👍🏻

kpavlov (commented):
@douglas-DS , are you sure there is no native calls or synchronized blocks in the underlaying code?

Please refer to 
- https://quarkus.io/guides/virtual-threads
- https://quarkus.io/blog/virtual-thread-1/#five-things-you-need-to-know-before-using-virtual-threads-for-everything
- https://quarkus.io/blog/virtual-threads-2/

Let's add thread pinning detection (`-Djdk.tracePinnedThreads`) to the java 21 maven profile: 
https://quarkus.io/guides/virtual-threads#detect-pinned-thread-in-tests

douglas-DS:
> @douglas-DS , are you sure there is no native calls or synchronized blocks in the underlaying code?
> 
> Please refer to
> 
>     * https://quarkus.io/guides/virtual-threads
> 
>     * https://quarkus.io/blog/virtual-thread-1/#five-things-you-need-to-know-before-using-virtual-threads-for-everything
> 
>     * https://quarkus.io/blog/virtual-threads-2/
> 
> 
> Let's add thread pinning detection (`-Djdk.tracePinnedThreads`) to the java 21 maven profile: https://quarkus.io/guides/virtual-threads#detect-pinned-thread-in-tests

What do you think would be a better approach here?

I’m wondering whether it would be better to pass the executor to the builder class.

kpavlov:
Passing the executor to builder class is very good approach. Even making it a general pattern by introducing abstract builder and inheriting from it. But I am just worrying about virtual thread safety of the underlaying vertex api.
Yes, making it customizable via builder would not break things for everyone, but will allow to enable virtual threads in your application if you want.

kpavlov:
And let's add jvm flag to report thread pinning. It might help to detect such issues in other places

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModel.java
```diff
@@ -52,7 +56,9 @@
...
+
+    private static final Executor DEFAULT_EXECUTOR = Executors.newCachedThreadPool();
```
dliubarskyi: Can we create default executor lazily, only if it is not provided by the user?
Also it it probably better to have a dedicated executor per each `VertexAiGeminiStreamingChatModel` instance.

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModel.java
```diff
@@ -168,6 +174,7 @@ public VertexAiGeminiStreamingChatModel(VertexAiGeminiStreamingChatModelBuilder
         this.logRequests = getOrDefault(builder.logRequests, false);
         this.logResponses = getOrDefault(builder.logResponses, false);
         this.listeners = copy(builder.listeners);
+        this.executor = Objects.requireNonNullElse(builder.executor, DEFAULT_EXECUTOR);
```
dliubarskyi:
```suggestion
        this.executor = getOrElse(builder.executor, () -> createDefaultExecutor());
```

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModel.java
```diff
@@ -371,105 +372,106 @@ private void generate(
...
-                } catch (Exception t) {
-                    logger.warn("Exception while calling model listener (onError)", t);
-                }
-            });
+            StreamingChatResponseBuilder responseBuilder = new StreamingChatResponseBuilder();
 
-            handler.onError(exception);
-        }
+            try {
+                model.generateContentStream(instructionAndContent.contents).stream().forEach(partialResponse -> {
```
dliubarskyi: Maybe let's execute code in the executor starting from this line? All the request preparation is better to do in the caller thread (in case there will be some validation exceptions thrown)

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/FunctionCallHelper.java
```diff
@@ -25,7 +25,9 @@ static FunctionCall fromToolExecutionRequest(ToolExecutionRequest toolExecutionR
 
         Struct.Builder structBuilder = Struct.newBuilder();
         try {
-            JsonFormat.parser().merge(toolExecutionRequest.arguments(), structBuilder);
+            String toolArguments = toolExecutionRequest.arguments();
+            String arguments = toolArguments.isBlank() ? "{}" : toolArguments;
```
dliubarskyi:
```suggestion
            String arguments = isNullOrBlank(toolArguments) ? "{}" : toolArguments;
```

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModel.java
dliubarskyi: 👍🏻

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModel.java
dliubarskyi: We also need to make sure this executor shuts down by itself if not used for some (short) time, otherwise it will prevent the user application to shutdown.
See an example here: https://github.com/langchain4j/langchain4j/blob/536705ecff46d9749f3d8711da37d0ffd1131b6c/langchain4j-core/src/main/java/dev/langchain4j/rag/DefaultRetrievalAugmentor.java#L128

dliubarskyi (commented):
@douglas-DS thank you!

douglas-DS:
@dliubarskyi I've made all the requested changes. Could you please review it again?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModel.java
```diff
@@ -321,6 +324,15 @@ public VertexAiGeminiStreamingChatModel(GenerativeModel generativeModel, Generat
         this.logRequests = false;
         this.logResponses = false;
         this.listeners = Collections.emptyList();
+        this.executor = getOrDefault(executor, VertexAiGeminiStreamingChatModel::createDefaultExecutor);
+    }
+
+    private static ExecutorService createDefaultExecutor() {
+        return new ThreadPoolExecutor(
+            0, Integer.MAX_VALUE,
```
kpavlov, line 358: max pool size should not be Integer.MAX_VALUE

dliubarskyi: Why not? `Executors.newCachedThreadPool()` does exactly that

kpavlov: Unconstrained thread pool is a typically is not good for production as it can may consume too many platform threads. The JDK implementation is optimistic, although it's difficult to come up with exact value.
I am fine to leave it unconstrained, since it is possible to tune for the specific application.

douglas-DS: To be honest, I don’t see any benefit in having granular control over the default executor, since we already have the option to pass a specific executor implementation through the builder if needed.

douglas-DS:
@dliubarskyi @kpavlov  Is this ready for merge?

langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModel.java
```diff
@@ -482,6 +497,12 @@ public void close() {
         if (this.vertexAI != null) {
             this.vertexAI.close();
         }
+
+        if (this.executor != null) {
+            if (executor instanceof ExecutorService) {
```
kpavlov: can be simplified: if executor is null - it is not instance of ExecutorService. `if (this.executor != null)` is not needed

langchain4j-vertex-ai-gemini/src/test/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiStreamingChatModelIT.java
```diff
@@ -165,7 +165,7 @@ void should_allow_custom_generativeModel_and_generationConfig() {
         GenerativeModel generativeModel = new GenerativeModel(MODEL_NAME, vertexAi);
         GenerationConfig generationConfig = GenerationConfig.getDefaultInstance();
 
-        StreamingChatModel model = new VertexAiGeminiStreamingChatModel(generativeModel, generationConfig);
+        StreamingChatModel model = new VertexAiGeminiStreamingChatModel(generativeModel, generationConfig, null);
```
kpavlov: Shouldn't we create overriding method without the `executor` parameter for backwards compatibility ?

douglas-DS: I agree! I’ve brought the previous constructor back. Now we have both overloaded constructors, one with it and one without it.

douglas-DS:
@kpavlov  @dliubarskyi Could you review this again?

dliubarskyi (approved):
@douglas-DS thanks a lot!

PR #3302, opened 2025-07-04
Exposing MCP resources automatically over synthetic tools

### LINKED ISSUE #2767: [FEATURE] [MCP] Presenting prompts and resources to LLMs
Currently, we only have ways to access MCP prompts and resources programmatically. We could find a way to expose them in some more automated way to LLMs, possibly in the form of a `ToolProvider` that provides tools like:

- list prompts
- get prompt
- list resources
- get resource

These "synthetic" tools would then actually work with prompts/resources on the MCP server.

Opening this up to discussion, because somebody may have a better idea how to expose this.

jmartisk:
Fixes https://github.com/langchain4j/langchain4j/issues/2767

Wanted to get some early feedback before I proceed to creating more tests and documentation (and probably extend it to prompts too).

The idea is that the user can now, when creating a `McpToolProvider`, enable the `exposeResourcesAsTools` option. If this is enabled, then the tool provider will expose two extra 'synthetic' tools:

- `list_resources` to obtain a list of all resources available for all MCP clients that are wrapped into this tool provider
- `get_resource` to render a specific resource - it takes two arguments:
  - resource URI (I chose URI instead of the name to be able to retrieve [resource templates](https://modelcontextprotocol.io/specification/2025-06-18/server/resources#resource-templates) too in the same way)
  - name (`key`) of the MCP client that this resource belongs to

The result of `list_resources` is JSON and looks something like this:

```
[ {
  "mcpServer" : "alice",
  "uri" : "file:///info",
  "uriTemplate" : null,
  "name" : "basicInfo",
  "description" : "Basic information about Alice",
  "mimeType" : "text/plain"
}, {
  "mcpServer" : "bob",
  "uri" : "file:///info",
  "uriTemplate" : null,
  "name" : "basicInfo",
  "description" : "Basic information about Bob",
  "mimeType" : "text/plain"
} ]
```

I tested this with `gpt-4o-mini` and it understood how to find and retrieve a relevant resource (see `McpResourcesAsToolsTestBase#integrationTestWithAiService` for demonstration)

So a resource is identified by uri AND the name of the mcp connection, because it may happen that two servers will have resources on the same URI.
When it's a resource, then `uri` is specified here, if it's a template, then  `uriTemplate` is specified instead.

Note: another way would be to expose a separate synthetic tool for each resource instead of having `get_resource` over all of them. That would render it unnecessary to invent a new JSON schema for the `list_resources`, but I'm afraid that there will, in the real world, be too many resources, and exposing a tool for each separately would explode the context size.

### DIFF
--- a/docs/docs/tutorials/mcp.md
+++ b/docs/docs/tutorials/mcp.md
@@ -140,6 +140,13 @@ McpClient mcpClient = new DefaultMcpClient.Builder()
 
 ## Resources
 
+There are two ways how to work with resources. Either the application can call the MCP client's
+resource-related methods to access resources programmatically, or there is an option to expose
+resources automatically to LLM calls via synthetic tools (one tool to obtain a list of resources
+and one to obtain the contents of a resource) so that Chat Models can consult resources by themselves.
+
+### Accessing resources programmatically
+
 To obtain a list of [MCP resources](https://modelcontextprotocol.io/docs/concepts/resources) 
 on the server, use `client.listResources()`, or `client.listResourceTemplates()` in case of resource templates.
 This will return a list of `McpResource` objects (or `McpResourceTemplate` respectively). These
@@ -149,6 +156,49 @@ To obtain the actual contents of the resource, use `client.readResource(uri)`, s
 This returns a `McpReadResourceResult`, which contains a  list of `McpResourceContents` objects (there may be more resource contents on a single URI, for 
 example if the URI represents a directory). Each `McpResourceContents` object represents either a 
 binary blob (`McpBlobResourceContents`) or text (`McpTextResourceContents`).
+
+### Exposing resources automatically via synthetic tools
+
+If you set a `McpResourcesAsToolsPresenter` instance using the builder when building an `McpToolProvider`,
+the MCP tool provider will automatically add two synthetic tools to the result of its `provideTools` method,
+along the 'regular' tools that are supported by the backing MCP servers. One tool should serve for obtaining
+a list of resources and the other for obtaining a particular resource. LangChain4j provides
+a default implementation named `DefaultMcpResourcesAsToolsPresenter` that adds these two tools:
+
+**_NOTE:_** The rest of this section describes `DefaultMcpResourcesAsToolsPresenter`. You may plug in your own implementation
+that behaves differently.
+
+- `list_resources`: Lists all resources exposed by the backing MCP servers. This tool takes no arguments.
+- `get_resource`: Reads the contents of a resource. This tool takes two arguments that together identify a resource:
+MCP server name and the URI.
+
+The output of `list_resources` is a JSON array like:
+
+```json
+[ {
+  "mcpServer" : "alice",
+  "uri" : "file:///info",
+  "uriTemplate" : null,
+  "name" : "basicInfo",
+  "description" : "Basic information about Alice",
+  "mimeType" : "text/plain"
+}, {
+  "mcpServer" : "bob",
+  "uri" : "file:///info",
+  "uriTemplate" : null,
+  "name" : "basicInfo",
+  "description" : "Basic information about Bob",
+  "mimeType" : "text/plain"
+} ]
+```
+
+Each document in this array represents one resource. Each resource is identified by a combination of `uri` and `mcpServer`, 
+where `mcpServer` is the value of `key` that was assigned to the MCP client during creation (see `DefaultMcpClient.Builder#key`).
+When the Chat model invokes the `list_resources` tool, 
+it receives this list of resources and can then decide to invoke `read_resource`. The default descriptions of `list_resources`
+and `get_resource` tools should suffice under most circumstances to explain to an LLM how to use them. However, if you need
+to customize the descriptions of these tools and their arguments, you override them using the methods of
+`DefaultMcpResourcesAsToolsPresenter.Builder`.
 
 ## Prompts
 
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
@@ -4,20 +4,20 @@ import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.internal.Utils;
 import dev.langchain4j.mcp.client.McpClient;
+import dev.langchain4j.mcp.resourcesastools.McpResourcesAsToolsPresenter;
 import dev.langchain4j.service.IllegalConfigurationException;
 import dev.langchain4j.service.tool.ToolExecutor;
 import dev.langchain4j.service.tool.ToolProvider;
 import dev.langchain4j.service.tool.ToolProviderRequest;
 import dev.langchain4j.service.tool.ToolProviderResult;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiPredicate;
-
 import java.util.function.Function;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -31,20 +31,28 @@ public class McpToolProvider implements ToolProvider {
     private final AtomicReference<BiPredicate<McpClient, ToolSpecification>> mcpToolsFilter;
     private final Function<ToolExecutor, ToolExecutor> toolWrapper;
     private static final Logger log = LoggerFactory.getLogger(McpToolProvider.class);
+    private final McpResourcesAsToolsPresenter resourcesAsToolsPresenter;
 
     private McpToolProvider(Builder builder) {
-        this(builder.mcpClients, Utils.getOrDefault(builder.failIfOneServerFails, false), builder.mcpToolsFilter, builder.toolWrapper);
+        this(
+                builder.mcpClients,
+                Utils.getOrDefault(builder.failIfOneServerFails, false),
+                builder.mcpToolsFilter,
+                builder.toolWrapper,
+                builder.resourcesAsToolsPresenter);
     }
 
-    protected McpToolProvider(List<McpClient> mcpClients, boolean failIfOneServerFails, BiPredicate<McpClient, ToolSpecification> mcpToolsFilter) {
-        this(Objects.requireNonNull(mcpClients), failIfOneServerFails, mcpToolsFilter, Function.identity());
-    }
-
-    protected McpToolProvider(List<McpClient> mcpClients, boolean failIfOneServerFails, BiPredicate<McpClient, ToolSpecification> mcpToolsFilter, Function<ToolExecutor, ToolExecutor> toolWrapper) {
+    protected McpToolProvider(
+            List<McpClient> mcpClients,
+            boolean failIfOneServerFails,
+            BiPredicate<McpClient, ToolSpecification> mcpToolsFilter,
+            Function<ToolExecutor, ToolExecutor> toolWrapper,
+            McpResourcesAsToolsPresenter resourcesAsToolsPresenter) {
         this.mcpClients = new CopyOnWriteArrayList<>(mcpClients);
         this.failIfOneServerFails = failIfOneServerFails;
         this.mcpToolsFilter = new AtomicReference<>(mcpToolsFilter);
         this.toolWrapper = toolWrapper;
+        this.resourcesAsToolsPresenter = resourcesAsToolsPresenter;
     }
 
     /**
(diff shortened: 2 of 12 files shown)


jmartisk:
FYI @geoand if you'd like to review too. On the Quarkus side, we would have a `expose-resources-as-tools` config property.

geoand:
Makes sense to me!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -123,6 +133,29 @@ protected ToolProviderResult provideTools(ToolProviderRequest request, BiPredica
...
+                    .description(
+                            "Lists all available resources. The result is in JSON format. When a resource description"
+                                    + "contains uriTemplate instead of uri, then the URI is a template and placeholders (in curly brackets) need to be substituted with an actual value")
+                    .parameters(JsonObjectSchema.builder().build())
+                    .build();
+            builder.add(
+                    listResourcesToolSpec, toolWrapper.apply(new ListResourcesToolExecutor(mcpClientsUnmodifiable)));
+
+            ToolSpecification getResourceToolSpec = ToolSpecification.builder()
+                    .name("get_resource")
```
dliubarskyi: It would be nice to allow user to override default:
- tool name
- tool description
- parameter descriptions

jmartisk:
also @mariofusco

dliubarskyi (commented):
@jmartisk thank you, sounds good!

> but I'm afraid that there will, in the real world, be too many resources, and exposing a tool for each separately would explode the context size.

Agreed, the proposed approach with 2 tools seems more flexible and cheaper. In case someone will actually want the second approach, we can always add it as a separate config property in the future.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
jmartisk: Yeah I agree, I'm just worried about how this will look when `McpToolProvider.Builder` receives these methods

- `getResourceToolName`
- `getResourceToolDescription`
- `getResourceToolMcpServerParameterDescription`
- `getResourceToolUriParameterDescription`
- `listResourcesToolName`
- `listResourcesToolDescription`

It looks kinda ugly, but... yeah, got any better idea?

jmartisk: Maybe rather

- `setNameForGetResourceTool`
- `setDescriptionForGetResourceTool`
- `setDescriptionForMcpServerParameterOfGetResourceTool`
- `setDescriptionForUriParameterOfGetResourceTool`
- `setNameForListResourcesTool`
- `setDescriptionForListResourcesTool`

which is even longer, but clearer (the previous examples would be setter methods whose name actually starts with `get`/`list`, which is ultra weird). 

An idea: we could also add a specific inner class into the `McpToolProvider.Builder` that holds this configuration, to avoid 'polluting' the normal API of the builder.. so it would go like

```
McpToolProvider.ExposeResourcesAsTools config = McpToolProvider.ExposeResourcesAsTools.builder()
   .expose(true)
  .setNameForGetResourceTool("get_resources")
  .build();
ToolProvider toolProvider = McpToolProvider.builder()
                .mcpClients(mcpClientAlice, mcpClientBob)
                .exposeResourcesAsTools(config)
                .build();
```

it would hide these weirdly named 6 methods from the basic API, but it would be an extra class, so maybe not worth it...

dliubarskyi: What about `exposeResourcesAsTools(true)` (for defaults) and `exposeResourcesAsTools(true, config)` (for those who want to configure precisely)?


We could also probably merge name/description/params into one method call, so that config only has 2 methods.

But I am not sure aobut any of this either.

jmartisk: I think I'll extract the whole thing into an interface named `McpResourcesAsToolsPresenter` plus  a default configurable implementation of it. And the `McpToolProvider` will optionally contain a reference to an instance of `McpResourcesAsToolsPresenter`. That sounds quite clean and also flexible.

jmartisk:
It was a bit crazy, but I refactored it and extracted a `McpResourcesAsToolsPresenter` interface plus a default implementation for it, this is the cleanest (and most flexible) solution I can think of.

One open question is how to handle binary resources because we have to pass them to the chat model as a string... Right now it throws an UnsupportedOperationException...

docs/docs/tutorials/mcp.md
```diff
@@ -140,6 +140,13 @@ McpClient mcpClient = new DefaultMcpClient.Builder()
 
 ## Resources
 
+There are two ways how to work with resources. Either the application can call the MCP client's
+resource-related methods to access resources programmatically, or there is an option to expose
+resources automatically to LLM calls via synthetic tools (one tool to obtain a list of resources
+and one to obtain the contents of a resource) so that Chat Models can consult resources by themselves.
```
dliubarskyi, line 146:
```suggestion
and one to obtain the contents of a resource) so that LLMs can consult resources by themselves.
```

docs/docs/tutorials/mcp.md
```diff
@@ -150,6 +157,49 @@ This returns a `McpReadResourceResult`, which contains a  list of `McpResourceCo
...
+  "uri" : "file:///info",
+  "uriTemplate" : null,
+  "name" : "basicInfo",
+  "description" : "Basic information about Bob",
+  "mimeType" : "text/plain"
+} ]
+```
+
+Each document in this array represents one resource. Each resource is identified by a combination of `uri` and `mcpServer`, 
+where `mcpServer` is the value of `key` that was assigned to the MCP client during creation (see `DefaultMcpClient.Builder#key`).
```
dliubarskyi, line 196: I am not sure about this, but in most cases there will probably be a single MCP server configured.
So maybe in this case it makes sense to omit the `mcpServer` argument from the tool?
Benefits:
- LLM will be less likely to hallucinate it and fail, also less tokens consumed
- user does not need to explicitly specify `key`

WDYT?

dliubarskyi:
> One open question is how to handle binary resources because we have to pass them to the chat model as a string... Right now it throws an UnsupportedOperationException...

Maybe filter binary resources out by default? Only if user sets e.g. `includeBinaryResources(true)`, then binary resources are included. But then it is not clear for what reason, as tool can only return text content at the moment...

docs/docs/tutorials/mcp.md
jmartisk: Without it, how will we handle cases where multiple servers offer resources with the same URI?
Besides, the user does not have to specify this. The key, if unspecified, is auto-generated (https://github.com/langchain4j/langchain4j/blob/1.1.0/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java#L76) - maybe I could repeat that in the docs to avoid scaring the user that they have to specify it

jmartisk: ~Maybe I could imagine ignoring a wrong key in cases where the URI is enough to identify a resource~

Ah, no, with multiple servers, we would need to contact all of them anyway to find out where the resource is. So, we could ignore it in cases where there is only a single MCP server. 

> I am not sure about this, but in most cases there will probably be a single MCP server configured.

I don't know that much about real-world deployments yet, but I would assume that a lot of apps will use multiple MCP servers

dliubarskyi: > So, we could ignore it in cases where there is only a single MCP server.

But then what's the point in forcing the LLM to generate the argument that will be ignored anyway?

What I meant is:
When there is a single MCP client configured, get_resource would have just one argument (uri).
When there are multiple MCP clients configured, get_resource would have 2 arguments.

jmartisk:
> Maybe filter binary resources out by default? Only if user sets e.g. `includeBinaryResources(true)`, then binary resources are included. But then it is not clear for what reason, as tool can only return text content at the moment...

Problem is that you don't know for sure what type a resource is until you actually retrieve it from the server... You can guess from the MIME type, but that wouldn't be reliable. See https://modelcontextprotocol.io/specification/2025-06-18/server/resources#text-content - when you retrieve, either the `text` or the `blob` field will be present, and you don't know in advance. That's another thing I don't like about the protocol...

Would it make sense to still simply include blob resources and pass them as base64-encoded? Not sure how much LLMs have a chance of understanding it...

docs/docs/tutorials/mcp.md
jmartisk: That's a bit of a danger zone because the `McpToolProvider` allows adding and removing MCP clients dynamically, there might be one when `list_resources` is invoked and then more when `get_resource` is invoked. The signature may change within a single conversation...

I don't think requiring the LLM to pass the key is a big problem. And even if it's wrong and we can still retrieve the resource, we will retrieve it... I mean, just for leniency, to allow room for a little hallucination.

jmartisk: Thinking about it again, if we allow dynamically adding and removing clients, maybe using keys should be required, to avoid accidentally contacting a wrong server after the server list changes, which may even be a security problem

dliubarskyi:
> Would it make sense to still simply include blob resources and pass them as base64-encoded? Not sure how much LLMs have a chance of understanding it...

This is risky as base64-encoded content is usually huge in size, this will consume a lot of tokens which will most probably be wasted. I would probably return "Error: content is binary and cannot be converted to text" or something like this from the `get_resource` tool in this case, at least by default.

docs/docs/tutorials/mcp.md
dliubarskyi: > I don't think requiring the LLM to pass the key is a big problem. And even if it's wrong and we can still retrieve the resource, we will retrieve it... I mean, just for leniency, to allow room for a little hallucination.

Yeah, makes sense

jmartisk:
Pushed. I kept the `mcpServer` argument required in the end to avoid potential security problems, and made it return an error on binary resources

docs/docs/tutorials/mcp.md
dliubarskyi: > That's a bit of a danger zone because the McpToolProvider allows adding and removing MCP clients dynamically, there might be one when list_resources is invoked and then more when get_resource is invoked. The signature may change within a single conversation...

With the current implementation, this does not seem possible as you make a copy (`Collections.unmodifiableList(mcpClients)`), so there is no risk.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/resourcesastools/DefaultMcpResourcesAsToolsPresenter.java
```diff
@@ -0,0 +1,189 @@
...
+        private String descriptionOfGetResourceTool;
+        private String descriptionOfMcpServerParameterOfGetResourceTool;
+        private String descriptionOfUriParameterOfGetResourceTool;
+        private String nameOfListResourcesTool;
+        private String descriptionOfListResourcesTool;
+
+        /**
+         * Overrides the name of the `get_resource` tool.
+         */
+        public Builder nameOfGetResourceTool(String nameOfGetResourceTool) {
```
dliubarskyi, line 139: nit: from the IDE autocomplete point of view it is probably more convenient to name it `getResourceToolName` and `getResourceToolDescription`, etc

docs/docs/tutorials/mcp.md
```diff
@@ -150,6 +157,49 @@ This returns a `McpReadResourceResult`, which contains a  list of `McpResourceCo
...
+  "mcpServer" : "alice",
+  "uri" : "file:///info",
+  "uriTemplate" : null,
+  "name" : "basicInfo",
+  "description" : "Basic information about Alice",
+  "mimeType" : "text/plain"
+}, {
+  "mcpServer" : "bob",
+  "uri" : "file:///info",
+  "uriTemplate" : null,
```
dliubarskyi, line 188: Would it be possible to not serialize `null` properties in order to save tokens?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/resourcesastools/DefaultMcpResourcesAsToolsPresenter.java
jmartisk: That is what it was named before but I changed it, because I would absolutely hate for a setter method to be named starting with `get`

docs/docs/tutorials/mcp.md
jmartisk: `Collections.unmodifiableList` does not create a copy, only a view that is unmodifiable by the receiver of that object

docs/docs/tutorials/mcp.md
jmartisk: I'm not completely sure it wouldn't confuse the LLM, but hopefully not...

But, I'm not that familiar with the Json SPI, I just call `dev.langchain4j.internal.Json.toJson(...)`, is it possible to configure it to leave out nulls? I guess that would need some updates to the SPI and its implementations, right?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/resourcesastools/DefaultMcpResourcesAsToolsPresenter.java
jmartisk: `setNameOfGetResourceTool` would work, maybe

jmartisk: `setGetResourceToolName` is more "standard" but in this case also quite confusing  :laughing:

docs/docs/tutorials/mcp.md
dliubarskyi: You are right, then let's keep it simple

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/resourcesastools/DefaultMcpResourcesAsToolsPresenter.java
dliubarskyi: Yeah, that will be even more confusing :D

dliubarskyi (approved):
@jmartisk thanks a lot!

PR #3329, opened 2025-07-08
Expose invoke on ToolExecutor

### LINKED ISSUE #3328: [FEATURE] Expose invoke on ToolExecutor interface
I'd like to be able to get the results of a raw method invocation on ToolExecutor rather than the marshalled json string.

patriot1burke:
## Issue
Closes #3328 

## Change
* Added an invoke() method to ToolExecutor that returns result of raw method invocation
* Implemented invoke() invocation in DefaultToolExecutor

In my current use case, I don't want to marshal the result into json as I want to do some local processing on the tool execution.
If the methods in DefaultToolExecutor were protected instead of private, I could have extended DefaultToolExecutor.  I think adding an invoke method to ToolExecutor would be better and useful.

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
@@ -85,6 +85,19 @@ public class DefaultToolExecutor implements ToolExecutor {
         }
     }
 
+    public Object invoke(ToolExecutionRequest toolExecutionRequest, Object memoryId) throws Throwable {
+        Map<String, Object> argumentsMap = argumentsAsMap(toolExecutionRequest.arguments());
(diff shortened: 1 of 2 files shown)


dliubarskyi (commented):
Hi @patriot1burke, thank you for your PR. You have added a new method to the API, but it is not called by anyone. Or do you plan to call this method manually?

patriot1burke:
@dliubarskyi 

I plan to call it manually.  I'm invoking tools in some different ways, and it would be nice to expose the invoke() method.   Right now all that code is private in DefaultToolExecutor and I had to fork it to reuse it.

dliubarskyi:
@patriot1burke I am hesitant to add a new method to the `ToolExecutor` interface, especially if it will be not used by the library.
Let's better make some of the methods protected, but please make as minimal change as possible

PR #3378, opened 2025-07-18
Add possibility to transform a ChatRequest before it is sent to the LLM

mariofusco:
## Issue

Closes #

## Change
Add possibility to transform a `ChatRequest` before it is sent to the LLM.

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
@@ -6,6 +6,7 @@ import static dev.langchain4j.spi.ServiceHelper.loadFactories;
 import static java.util.Arrays.asList;
 import static java.util.stream.Collectors.toList;
 
+import dev.langchain4j.Internal;
 import dev.langchain4j.agent.tool.Tool;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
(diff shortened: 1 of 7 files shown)


langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -176,6 +180,11 @@ public static <T> T create(Class<T> aiService, StreamingChatModel streamingChatM
      */
     public static <T> AiServices<T> builder(Class<T> aiService) {
         AiServiceContext context = new AiServiceContext(aiService);
+        return builder(context);
+    }
+
+    @Internal
+    public static <T> AiServices<T> builder(AiServiceContext context) {
```
mariofusco, line 187: This change is totally unrelated with this feature, but I need it for my agentic framework so I added it here.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesIT.java
```diff
@@ -926,9 +929,60 @@ void should_use_content_retriever_and_return_sources_inside_result_with_pojo() {
...
+    }
+
     static ChatRequest chatRequest(String userMessage) {
         return ChatRequest.builder()
                 .messages(dev.langchain4j.data.message.UserMessage.from(userMessage))
                 .build();
     }
+
+    @FunctionalInterface
+    public interface UserMessageTransformer extends UnaryOperator<ChatRequest> {
```
mariofusco, line 959: I believe that this class could be of general usefulness, so probably it could be a good idea to promote it from this test to an actual utility class in our code base?

dliubarskyi: Hmm, not sure how universal it is in it's current implementation. There are many things that can be done differently: which user message to transfroms: first, last, specific, or all? Also, user message can contain not only text but other modalities. Etc, etc. I would keep it in the test for the moment and see if there will be a need for it, we can always add it in the future.

dliubarskyi (approved):
@mariofusco thanks a lot!

PR #3397, opened 2025-07-24
Fix bedrock streaming model hangs, when using tools.

### LINKED ISSUE #3396: [BUG] Bedrock Streaming model hangs when using tools (or any other nested chat call)
**Describe the bug**
When using AWS Bedrock streaming model, calling another chat call in the completion handler will cause the AWS SDK to fail borrowing another connection from the POOL. This is caused because the current implementation awaits for the CompletableFuture completion inside the chat method. This is reproduced easliy using Quarkus langchain4j.
I will open a PR shortly, which I confirmed fixes the issue, but would like to understand the motive for this initial implementation as I might missed something.

**To Reproduce**
Use Quarkus Langchain4J with aws bedrock model in streaming mode and add tools.

**Expected behavior**
Using Bedrock Streaming Model with tools should work properly.

- LangChain4j version: 1.2.0
- LLM(s) used:  AWS Bedrock
- Java version: 21

tomas1885:
## Issue
Closes #3396

## Change
Removed the blocking .get call in the BedrockStreamingChatModel, which prevents nesting chat calls.

## General checklist

### DIFF
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockStreamingChatModel.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockStreamingChatModel.java
@@ -105,13 +105,13 @@ public class BedrockStreamingChatModel extends AbstractBedrockChatModel implemen
                         .onMessageStop(responseBuilder::append)
                         .build())
                 .build();
+            this.client.converseStream(converseStreamRequest, converseStreamResponseHandler)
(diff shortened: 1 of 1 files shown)


geoand (approved):
Thanks a lot!

geoand:
cc @dliubarskyi

dliubarskyi:
cc @tenpigs267

dliubarskyi:
@tomas1885 thanks a lot for the fix!

PR #3412, opened 2025-07-25
Azure OpenAI - add support for token usage reporting on streaming chat model

### LINKED ISSUE #1068: [FEATURE] Azure OpenAI: Use provided token usage instead of manually calculated

ssmogos:
## Issue
Closes #1068
Azure AI Chat client did not have support to report token usage.

## Change
Azure chat completions api supports requesting token usage reporting by setting:

`"stream_options": {"include_usage":true}`

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/cognitiveservices/data-plane/AzureOpenAI/inference/preview/2025-04-01-preview/inference.json

Impl includes:
If a token estimator is set on the chat completion options, then do not request token usage.
Updated sse responses to set token usage if present in the response.
Updated integration tests to validate that token usage is set if no token count estimator is set.

### DIFF
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java
@@ -20,10 +20,18 @@ import static dev.langchain4j.model.azure.InternalAzureOpenAiHelper.validate;
 import static dev.langchain4j.spi.ServiceHelper.loadFactories;
 import static java.util.Arrays.asList;
 
+import java.time.Duration;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 import com.azure.ai.openai.OpenAIAsyncClient;
+import com.azure.ai.openai.implementation.accesshelpers.ChatCompletionsOptionsAccessHelper;
 import com.azure.ai.openai.models.AzureChatEnhancementConfiguration;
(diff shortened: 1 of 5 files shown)


dliubarskyi:
@ssmogos thanks a lot! I was not aware of `ChatCompletionsOptionsAccessHelper` and was wondering why is setter for             `setStreamOptions` private, any idea?

ssmogos:
> @ssmogos thanks a lot! I was not aware of `ChatCompletionsOptionsAccessHelper` and was wondering why is setter for `setStreamOptions` private, any idea?

@dliubarskyi , assumption is that they wanted to protect against setting it from other packages, and make sure the stream is set to true when streaming options are set... but that did not help them that much the way it's implemented.

Now looking more into it, and thanks for raising the question, maybe might have been a better approach to just update 
https://github.com/langchain4j/langchain4j/blob/main/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiStreamingChatModel.java#L227
to use the chat streaming completion options on:
```
client.getChatCompletionsStream(parameters.modelName(), options, 
                new ChatCompletionStreamOptions().setIncludeUsage(true));
```
and not include their hack-ish way of setting the streaming options through the access helper.

dliubarskyi (approved):
@ssmogos thank you!

PR #2124, opened 2024-11-15
integrate watsonx ai

omarmahamid:
## Issue

Closes #https://github.com/langchain4j/langchain4j/issues/2050

## General checklist

- [ ] There are no breaking changes
- [ ] I have added unit and integration tests for my change
- [ ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new maven module

- [ ] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #3380, opened 2025-07-18
Support thinking/reasoning

dliubarskyi:
## Issue
Closes https://github.com/langchain4j/langchain4j/issues/3013
Closes https://github.com/langchain4j/langchain4j/issues/2605
Closes https://github.com/langchain4j/langchain4j/issues/2472
Closes https://github.com/langchain4j/langchain4j/issues/3361
Closes https://github.com/langchain4j/langchain4j/issues/1224

## Change
Support thinking/reasoning feature in `ChatModel`/`StreamingChatModel` API for:
- [X] [Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/userguide/inference-reasoning.html)
- [X] [Anthropic](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking)
- [X] [DeepSeek](https://api-docs.deepseek.com/guides/reasoning_model) (via OpenAI module)
- [X] [Gemini](https://ai.google.dev/gemini-api/docs/thinking)
- [X] [Ollama](https://ollama.com/blog/thinking)

### API changes:
- `AiMessage`: added `thinking()` field to keep thinking/reasoning text
- `AiMessage`: added generic `attributes()` field to keep thinking/reasoning signatures (for Bedrock, Anthropic, Gemini)
- `StreamingChatResponseHandler`: added `onPartialThinking(PartialThinking)` callback
- `TokenStream`: added `onPartialThinking(Consumer<PartialThinking>)` callback
- `BedrockChatModel` and `BedrockStreamingChatModel`: added `returnThinking` and `sendThinking` parameters to the builder
- `AnthropicChatModel` and `AnthropicStreamingChatModel`: added `returnThinking` and `sendThinking` parameters to the builder
- `GoogleAiGeminiChatModel` and `GoogleAiGeminiStreamingChatModel`: added `returnThinking` and `sendThinking` parameters to the builder
- `OllamaChatModel` and `OllamaStreamingChatModel`: added `think` and `returnThinking` parameters to the builder
- `OpenAiChatModel` and `OpenAiStreamingChatModel`: added `returnThinking` parameter to the builder

## General checklist
- [X] There are no breaking changes
- [X] I have added unit and/or integration tests for my change
- [X] The tests cover both positive and negative cases
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
[...]

PR #136, opened 2023-08-26
Tool function

BOFA1ex:
### Feature

- `tool_function` must instantiate the impl-method by lambdaMeta(capturingClazz,implMethodSignature,etc.)
- lambda: use `reflect` as normal extraction, shadow was slow in pref, so made it last_guard_policy to extract it.
[note] idea evaluate lambda expression via `Proxy`.

- func_interface: wrap `tool_instance` or companion, only support 1-3 tool_function_argument length now.

### Followup
- `ToolFunctionExecutorTest` will be commit in later
- `AiServiceIT` could append more `tool_function` cases.

PR #3423, opened 2025-07-28
Integrate watsonx.ai module

### LINKED ISSUE #2050: [FEATURE] Integrate IBM Watsonx Model

Add integration of Watsonx ai in langchain4j

https://python.langchain.com/docs/integrations/llms/ibm_watsonx/

andreadimaio:
Closes #2050
Closes #2124

## Change
This is my first contribution to this repository, so there might be something I've missed. Let me know! :smile:

## General checklist
- [x] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (**I have to create a new PR**)


docs/docs/integrations/language-models/index.md
```diff
@@ -25,6 +25,7 @@ sidebar_position: 0
...
 | [Zhipu AI](/integrations/language-models/zhipu-ai)                               | ✅                                          | ✅/✅                                                                                                      |                                                                    |                                               |                      | text, image                                                                        | ✅                                         |                                                                 |                                                                                                                         |                       |                                                                            |
+[watsonx.ai](/integrations/language-models/watsonx) | ✅|  ✅/✅/❌/✅ | ✅ | ✅ | ✅ (Granite) | text, image | ✅ |✅ | | |
```
dliubarskyi: nit: this does not seem to be properly formated

docs/docs/integrations/language-models/watsonx.md
```diff
@@ -0,0 +1,331 @@
...
+- [watsonx.ai API Reference](https://cloud.ibm.com/apidocs/watsonx-ai)
+- [watsonx.ai Java SDK](https://github.com/IBM/watsonx-ai-java-sdk)
+
+## Maven Dependency
+
+```xml
+<dependency>
+    <groupId>dev.langchain4j</groupId>
+    <artifactId>langchain4j-watsonx</artifactId>
+    <version>1.3.0</version>
```
dliubarskyi:
```suggestion
    <version>1.4.0-beta10</version>
```

langchain4j-bom/pom.xml
```diff
@@ -195,6 +197,12 @@
                 <version>1.4.0-beta10-SNAPSHOT</version>
             </dependency>
 
+            <dependency>
+                <groupId>dev.langchain4j</groupId>
+                <artifactId>langchain4j-watsonx</artifactId>
+                <version>1.4.0-SNAPSHOT</version>
```
```suggestion
                <version>1.4.0-beta10-SNAPSHOT</version>
```

langchain4j-watsonx/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>1.4.0-beta10-SNAPSHOT</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-watsonx</artifactId>
+    <version>1.4.0-SNAPSHOT</version>
```
dliubarskyi:
```suggestion
```

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/WatsonxChatModelTest.java
```diff
@@ -0,0 +1,441 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Captor;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+public class WatsonxChatModelTest {
```
dliubarskyi, line 60: These tests are nice, but it would be great to have a full set of tests that are inherited from `AbstractChatModelIT`, `AbstractStreamingChatModelIT`, `AbstractChatModelListenerIT` and `AbstractStreamingChatModelListenerIT`. They test if the integration is "compliant".

Do you know if it is possible to setup a test account for LC4j to be able to test against the real watsonx API?

dliubarskyi (commented):
@andreadimaio thanks a lot!

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/WatsonxChatModelTest.java
andreadimaio: Yes, it should be possible to create a free account with monthly limitations. Could this be a solution?

andreadimaio: I'm working on integrating the `IT` tests into the watsonx module. In the meantime, I have a question. I don't know how the release pipeline works, but since we have a limited number of tokens, would it be useful to only launch watsonx's IT tests when a release is made?

andreadimaio:
@dliubarskyi I don't understand the purpose of the [should_propagate_user_exceptions_thrown_from_onPartialResponse](https://github.com/andreadimaio/langchain4j/blob/7c2f14632ef8de835deaf80937597c9e1f5e781e/langchain4j-core/src/test/java/dev/langchain4j/model/chat/common/AbstractStreamingChatModelIT.java#L42) test. Do you expect the `onCompleteResponse` to be called if an exception occurs during the `onPartialResponse`?

In my implementation, if an exception occurs during `onPartialResponse`, only `onError` is called.

andreadimaio:
@dliubarskyi In my opinion, the line [#136 in the test should_respect_user_message](https://github.com/andreadimaio/langchain4j/blob/7c2f14632ef8de835deaf80937597c9e1f5e781e/langchain4j-core/src/test/java/dev/langchain4j/model/chat/common/AbstractBaseChatModelIT.java#L136) should be changed from `isExactlyInstanceOf` to `isInstanceOf`.

dliubarskyi:
@andreadimaio if an exception is thrown from one of the callback methods (such as `onPartialResponse`), it should be propagated into `onError` callback. This way, if user code throws exceptions, it will not be lost.

But even if an exception occurs inside the user code, we should continue streaming and eventually call `onCompleteResponse` (unless there is a problem with the stream itself).

dliubarskyi:
> @dliubarskyi In my opinion, the line [#136 in the test should_respect_user_message](https://github.com/andreadimaio/langchain4j/blob/7c2f14632ef8de835deaf80937597c9e1f5e781e/langchain4j-core/src/test/java/dev/langchain4j/model/chat/common/AbstractBaseChatModelIT.java#L136) should be changed from isExactlyInstanceOf to isInstanceOf.

Why? If you have introduced a subtype, you can just override `chatResponseMetadataType` method

andreadimaio:
> @andreadimaio if an exception is thrown from one of the callback methods (such as `onPartialResponse`), it should be propagated into `onError` callback. This way, if user code throws exceptions, it will not be lost.
> 
> But even if an exception occurs inside the user code, we should continue streaming and eventually call `onCompleteResponse` (unless there is a problem with the stream itself).

Ok, so I have to review my implementation, because I stop the streaming at the first error.

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/WatsonxChatModelTest.java
dliubarskyi: @andreadimaio yes, this is possible. How many tokens are we talking about?

andreadimaio: From IBM Cloud I see:

```	
Service instance
Instance includes:
• 20 capacity unit-hours (CUH) per month
• 50,000 tokens/data points per month
• 100 pages per month
```

andreadimaio:
@dliubarskyi, I have also added the IT tests. When I run them from my laptop, they are all green. Please let me know if there are other changes

docs/docs/integrations/language-models/watsonx.md
```diff
@@ -0,0 +1,396 @@
...
+System.out.println(answer);
+```
+
+> **NOTE:** Ensure your selected model supports tool use.
+---
+
+## Enabling Thinking / Reasoning Output
+
+Some IBM foundation models can include "thinking" or "reasoning" steps in their responses.  
+You can capture and separate this reasoning content from the final answer by using the `thinking(...)` builder method with `ExtractionTags`.
```
dliubarskyi, line 155: TODO (for me): check whether this works well with generic thinking API introduced in https://github.com/langchain4j/langchain4j/pull/3380

Especially streaming

docs/docs/integrations/language-models/watsonx.md
```diff
@@ -0,0 +1,396 @@
...
+    .timeout(Duration.ofSeconds(30))
+    .build();
+
+ChatModel model = WatsonxChatModel.builder()
+    .service(chatService)
+    .thinking(ExtractionTags.of("think", "response"))
+    .build();
+
+ChatResponse chatResponse = chatModel.chat(
+    CustomMessage.from(Map.of("content", "thinking")),
```
dliubarskyi: What is the purpose of `CustomMessage`? Could you please add a comment?

langchain4j-watsonx/pom.xml
```diff
@@ -0,0 +1,105 @@
...
+
+    <properties>
+        <skipWatsonxITs>false</skipWatsonxITs>
+    </properties>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.fasterxml.jackson</groupId>
+                <artifactId>jackson-bom</artifactId>
```
dliubarskyi: nit: this seems redundant, `jackson-bom` is already specified in `langchain4j-parent`

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
```diff
@@ -0,0 +1,85 @@
...
+import static java.util.Objects.requireNonNullElse;
+
+import com.ibm.watsonx.ai.chat.ChatProvider;
+import com.ibm.watsonx.ai.chat.model.ExtractionTags;
+import dev.langchain4j.data.message.AiMessage;
+import dev.langchain4j.model.chat.listener.ChatModelListener;
+import dev.langchain4j.model.chat.request.ChatRequestParameters;
+import java.util.List;
+
+public abstract class WatsonxChat {
```
dliubarskyi: Does it has to be `public`? Please check all classes and make internals non-`public`

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
```diff
@@ -0,0 +1,85 @@
...
+
+    protected final ChatProvider chatProvider;
+    protected final List<ChatModelListener> listeners;
+    protected final ChatRequestParameters defaultRequestParameters;
+    protected final boolean enableJsonSchema;
+    protected final ExtractionTags tags;
+
+    protected WatsonxChat(Builder<?> builder) {
+        this.chatProvider = requireNonNull(builder.chatProvider);
+        this.listeners = requireNonNullElse(builder.listeners, List.of());
```
dliubarskyi:
```suggestion
        this.listeners = copy(builder.listeners);
```

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/util/Converter.java
```diff
@@ -0,0 +1,239 @@
...
+import dev.langchain4j.internal.JsonSchemaElementUtils;
+import dev.langchain4j.model.chat.request.ChatRequestParameters;
+import dev.langchain4j.model.chat.request.ToolChoice;
+import dev.langchain4j.model.chat.response.CompleteToolCall;
+import dev.langchain4j.model.chat.response.PartialToolCall;
+import dev.langchain4j.model.output.FinishReason;
+import dev.langchain4j.model.watsonx.WatsonxChatRequestParameters;
+import java.util.List;
+
+public class Converter {
```
dliubarskyi: this class can be moved one level up and made package-private

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
```diff
@@ -0,0 +1,85 @@
...
+        this.defaultRequestParameters = requireNonNullElse(
+                builder.defaultRequestParameters,
+                WatsonxChatRequestParameters.builder().build());
```
dliubarskyi: Merging of `builder.defaultRequestParameters` and parameters from `ChatService` seems to be missing.

If no `defaultRequestParameters` are specified, `ChatModelListener` will not get values from `ChatService` (e.g., `modelId`).

See `OpenAiChatModel:90` for example

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatRequestParameters.java
```diff
@@ -0,0 +1,160 @@
...
+import java.util.Map;
+
+public class WatsonxChatRequestParameters extends DefaultChatRequestParameters {
+
+    private final String projectId;
+    private final String spaceId;
+    private final Map<String, Integer> logitBias;
+    private final Boolean logprobs;
+    private final Integer topLogprobs;
+    private final Integer n;
```
dliubarskyi: Is `n` really required? LC4j API does not support multiple completions

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @SuppressWarnings("unchecked")
+    public abstract static class Builder<T extends Builder<T>> {
+        private ChatProvider chatProvider;
+        private List<ChatModelListener> listeners;
+        private ChatRequestParameters defaultRequestParameters;
+        private Boolean enableJsonSchema;
+        private ExtractionTags tags;
+
+        public T service(ChatProvider chatProvider) {
```
dliubarskyi: nit: is mismatch in names on purpose?

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/util/Converter.java
```diff
@@ -0,0 +1,239 @@
...
+                .id(partialToolCall.id())
+                .index(partialToolCall.index())
+                .name(partialToolCall.name())
+                .partialArguments(partialToolCall.arguments())
+                .build();
+    }
+
+    public static ChatParameters toChatParameters(ChatRequestParameters parameters) {
+        ChatParameters.Builder builder = ChatParameters.builder()
+                .modelId(getOrDefault(parameters.modelName(), parameters.modelName()))
```
dliubarskyi: Something strange is happening in this method, parameters are taken from the same source (`parameters`), calling `getOrDefault` makes no sense

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
```diff
@@ -0,0 +1,194 @@
...
+                .aiMessage(aiMessage.build())
+                .metadata(WatsonxChatResponseMetadata.builder()
+                        .created(chatResponse.getCreated())
+                        .createdAt(chatResponse.getCreatedAt())
+                        .finishReason(finishReason)
+                        .id(chatResponse.getId())
+                        .modelName(chatResponse.getModelId())
+                        .model(chatResponse.getModel())
+                        .modelVersion(chatResponse.getModelVersion())
+                        .object(chatResponse.getObject())
```
dliubarskyi: nit: `object` ("chat.completion") does not seem to be very useful

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
```diff
@@ -0,0 +1,194 @@
...
+
+        return ChatResponse.builder()
+                .aiMessage(aiMessage.build())
+                .metadata(WatsonxChatResponseMetadata.builder()
+                        .created(chatResponse.getCreated())
+                        .createdAt(chatResponse.getCreatedAt())
+                        .finishReason(finishReason)
+                        .id(chatResponse.getId())
+                        .modelName(chatResponse.getModelId())
+                        .model(chatResponse.getModel())
```
dliubarskyi: what's the difference between `model` and `modelId`?

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    @Override
+    public Set<Capability> supportedCapabilities() {
+        var capatibilities = new HashSet<Capability>();
+        if (enableJsonSchema
+                || (nonNull(defaultRequestParameters.responseFormat())
+                        && defaultRequestParameters.responseFormat().type().equals(ResponseFormatType.JSON)
+                        && nonNull(defaultRequestParameters.responseFormat().jsonSchema())))
+            capatibilities.add(Capability.RESPONSE_FORMAT_JSON_SCHEMA);
```
dliubarskyi: This implementation is different from others, is there a reason?
For other integrations, [this is how it is configured](https://docs.langchain4j.dev/tutorials/structured-outputs#using-json-schema-with-ai-services):
```java
ChatModel chatModel = OpenAiChatModel.builder()
        .apiKey(System.getenv("OPENAI_API_KEY"))
        .modelName("gpt-4o-mini")
        .supportedCapabilities(RESPONSE_FORMAT_JSON_SCHEMA)
        .build();
```

If there is no good reason, please change it from `enableJsonSchema` to `supportedCapabilities` to keep it consistent.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatResponseMetadata.java
```diff
@@ -0,0 +1,83 @@
...
+
+import dev.langchain4j.model.chat.response.ChatResponseMetadata;
+
+public class WatsonxChatResponseMetadata extends ChatResponseMetadata {
+
+    private final String object;
+    private final String model;
+    private final Long created;
+    private final String modelVersion;
+    private final String createdAt;
```
dliubarskyi: nit: please keep created and createdAt together

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatResponseMetadata.java
```diff
@@ -0,0 +1,83 @@
+package dev.langchain4j.model.watsonx;
+
+import dev.langchain4j.model.chat.response.ChatResponseMetadata;
+
+public class WatsonxChatResponseMetadata extends ChatResponseMetadata {
+
+    private final String object;
+    private final String model;
+    private final Long created;
+    private final String modelVersion;
```
dliubarskyi, line 8: nit: please keep model and modelVersion together

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatResponseMetadata.java
dliubarskyi: Same for getters, builders, etc

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxEmbeddingModel.java
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @see EmbeddingService
+ */
+public class WatsonxEmbeddingModel implements EmbeddingModel {
+
+    private final EmbeddingService embeddingService;
+
+    protected WatsonxEmbeddingModel(Builder builder) {
+        requireNonNull(builder, "builder is required");
+        this.embeddingService = builder.embeddingService;
```
dliubarskyi: nit: It does not seem to be very user friendly to ask users to create a builder in order to only set one argument. What about creating another ctor that accepts `embeddingService`?

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxScoringModel.java
```diff
@@ -0,0 +1,115 @@
...
+ *     .build();
+ * }</pre>
+ *
+ * @see RerankService
+ */
+public class WatsonxScoringModel implements ScoringModel {
+
+    private final RerankService rerankService;
+
+    protected WatsonxScoringModel(Builder builder) {
```
dliubarskyi: same here

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxStreamingChatModel.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Override
+    public Set<Capability> supportedCapabilities() {
+        var capatibilities = new HashSet<Capability>();
+        if (enableJsonSchema
+                || (nonNull(defaultRequestParameters.responseFormat())
+                        && defaultRequestParameters.responseFormat().type().equals(ResponseFormatType.JSON)
+                        && nonNull(defaultRequestParameters.responseFormat().jsonSchema())))
+            capatibilities.add(Capability.RESPONSE_FORMAT_JSON_SCHEMA);
```
dliubarskyi: same here

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxTokenCountEstimator.java
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * @see TokenizationService
+ */
+public class WatsonxTokenCountEstimator implements TokenCountEstimator {
+
+    private final TokenizationService tokenizationService;
+
+    protected WatsonxTokenCountEstimator(Builder builder) {
+        requireNonNull(builder, "builder is required");
+        this.tokenizationService = builder.tokenizationService;
```
dliubarskyi: same here

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxTokenCountEstimator.java
```diff
@@ -0,0 +1,107 @@
...
+     * @param parameters the tokenization parameters to use.
+     * @return the estimated count of tokens.
+     */
+    public int estimateTokenCountInText(String text, TokenizationParameters parameters) {
+        return tokenizationService.tokenize(text, parameters).result().tokenCount();
+    }
+
+    @Override
+    public int estimateTokenCountInMessage(ChatMessage message) {
+        throw new UnsupportedOperationException("Unimplemented method 'estimateTokenCountInMessage'");
```
dliubarskyi, line 72: Is there a reason why this was left unimplemented? This means that `TokenWindowChatMemory` cannot be used. Not a blocker though

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/util/Converter.java
```diff
@@ -0,0 +1,239 @@
...
+
+        List<ToolExecutionRequest> toolExecutionRequests = null;
+
+        if (nonNull(assistantMessage.toolCalls())) {
+            toolExecutionRequests = assistantMessage.toolCalls().stream()
+                    .map(Converter::toToolExecutionRequest)
+                    .toList();
+        }
+
+        return AiMessage.from(assistantMessage.content(), toolExecutionRequests);
```
dliubarskyi: I would also check for `assistantMessage.refusal` and throw `ContentFilteredException` in case of refusal

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/util/Converter.java
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        if (parameters instanceof WatsonxChatRequestParameters watsonxParameters) {
+            builder.projectId(watsonxParameters.projectId());
+            builder.spaceId(watsonxParameters.spaceId());
+            builder.logitBias(watsonxParameters.logitBias());
+            builder.logprobs(watsonxParameters.logprobs());
+            builder.n(watsonxParameters.n());
+            builder.seed(watsonxParameters.seed());
+            builder.timeLimit(nonNull(watsonxParameters.timeLimit()) ? watsonxParameters.timeLimit() : null);
```
dliubarskyi:
```suggestion
            builder.timeLimit(watsonxParameters.timeLimit());
```

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/util/Converter.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    private static AssistantMessage toAssistantMessage(AiMessage aiMessage) {
+        if (aiMessage.hasToolExecutionRequests()) {
+            var toolCalls = aiMessage.toolExecutionRequests().stream()
+                    .map(Converter::toToolCall)
+                    .toList();
+            return AssistantMessage.tools(toolCalls);
+        }
+        return AssistantMessage.text(aiMessage.text());
```
dliubarskyi: Not 100% sure for Watsonx, but `AiMessage` can contain both text and tools

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/WatsonxChatModelIT.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    @Override
+    public boolean supportsSingleImageInputAsPublicURL() {
+        // Watsonx does not support images as URLs, only as Base64-encoded strings
+        return false;
+    }
+
+    @Override
+    @Disabled("It seems that supportsSingleImageInputAsPublicURL doesn't work with this method")
```
dliubarskyi: What does it mean? `WatsonxChatModel` should throw `UnsupportedFeatureException` if image is specified as URL

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/WatsonxStreamingChatModelIT.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public boolean supportsSingleImageInputAsPublicURL() {
+        // Watsonx does not support images as URLs, only as Base64-encoded strings
+        return false;
+    }
+
+    @Override
+    @Disabled("It seems that supportsSingleImageInputAsPublicURL doesn't work with this method")
```
dliubarskyi: same here

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/WatsonxChatModelListenerIT.java
```diff
@@ -0,0 +1,70 @@
...
+    protected ChatModel createFailingModel(ChatModelListener listener) {
+        return WatsonxChatModel.builder()
+                .service(createChatService("invalid-model"))
+                .listeners(List.of(listener))
+                .build();
+    }
+
+    @Override
+    protected Class<? extends Exception> expectedExceptionClass() {
+        return WatsonxException.class;
```
dliubarskyi: Would it be possible to implement `ExceptionMapper` (see `AzureOpenAiExceptionMapper` as example)

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/mock/WatsonxStreamingChatModelTest.java
```diff
@@ -0,0 +1,341 @@
...
+    public void testDoChatWithTool() throws Exception {
+
+        var toolCall = new ToolCall(0, "id", "function", new FunctionCall("name", "{}"));
+        var resultMessage = new ResultMessage(AssistantMessage.ROLE, null, "refusal", List.of(toolCall));
+        var resultChoice = new ChatResponse.ResultChoice(0, resultMessage, "tool_calls");
+        chatResponse.setChoices(List.of(resultChoice));
+
+        doAnswer(invocation -> {
+                    ChatHandler handler = invocation.getArgument(1);
+                    handler.onCompleteResponse(chatResponse);
```
dliubarskyi: The ordering seems to be off: usually first goes thinking, then tool calls, then `onCompleteResponse`

dliubarskyi (commented):
@andreadimaio thanks a lot, great job! There are many comments, but nothing really critical, I think it will be a good candidate for including in the release next week!

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/WatsonxChatModelTest.java
dliubarskyi: @andreadimaio do you have a personal account? I created a personal IBM account but free quota was used pretty quickly :(

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
andreadimaio: Yes because in watsonx.ai you can use a `ChatService` or a `DeploymentService`

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: As these are two parameters returned from the API, I have returned both of them.

andreadimaio: But, if I remember correctly, they always have the same value in my use case.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: As with other parameters, it is something that the API returns, so I return it as metadata.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxTokenCountEstimator.java
andreadimaio: Watsonx.ai makes a REST API available to calculate tokens for text only. What I could do is implement the method, but it throws an exception if the ChatMessage contains anything other than text. WDYT?

docs/docs/integrations/language-models/watsonx.md
andreadimaio: `CustomMessage` is used to enable reasoning/thinking mode for models in watsonx.ai

andreadimaio: Currently, watsonx.ai only offers one model for reasoning: `ibm/granite-3-3-8b-Instruct`.

andreadimaio: The `CustomMessage` is converted to a `TextChatMessageControl`.

image

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/util/Converter.java
andreadimaio: Yes, I forgot to pass the `default` parameters too. It should be ok now.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: I have updated it. I now use the `supportedCapabilities` field.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
andreadimaio: Now the merge is done into the `Converter.toChatParameters` method.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: The `WatsonxChatResponseMetadata` now only contains the `created` field.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: The `WatsonxChatResponseMetadata` now only contains the `created` field.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatResponseMetadata.java
andreadimaio: The `WatsonxChatResponseMetadata` now only contains the `created` field.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatResponseMetadata.java
andreadimaio: The `WatsonxChatResponseMetadata` now only contains the `created` field.

docs/docs/integrations/language-models/watsonx.md
dliubarskyi: Wouldn't it be more appropriate/intuitive to use parameters (similar to `logprobs` `seed`, etc.) in order to enable/disable thinking? Using `CustomMessage` seems like a misuse of the API

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxTokenCountEstimator.java
dliubarskyi: @andreadimaio sounds good

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
```diff
@@ -0,0 +1,254 @@
...
+            com.ibm.watsonx.ai.chat.util.StreamingToolFetcher.PartialToolCall partialToolCall) {
+        return PartialToolCall.builder()
+                .id(partialToolCall.id())
+                .index(partialToolCall.index())
+                .name(partialToolCall.name())
+                .partialArguments(partialToolCall.arguments())
+                .build();
+    }
+
+    public static ChatParameters toChatParameters(
```
dliubarskyi: Please check `ChatModel.chat()` and `doChat()`. In `chat()` we do:

```java
ChatRequest finalChatRequest = ChatRequest.builder()
                .messages(chatRequest.messages())
                .parameters(defaultRequestParameters().overrideWith(chatRequest.parameters()))
                .build();
```

Then we pass `finalChatRequest` into `ChatModelListener`. If `defaultRequestParameters()` is empty (as it is currently), then it means that `ChatModelListener` will see something different from what is actually being sent to the LLM.

Given this config:

```java
WatsonxChatModel.builder()
                .service(createChatService("meta-llama/llama-4-maverick-17b-128e-instruct-fp8"))
                .build()
```

If you configure a `ChatModelListener`, you will see that when `onRequest()` is called, it does not contain many data, for example model id.

Please check how `OpenAiChatModel` is implemented. In the ctor, you need to merge provided `defaultRequestParameters` with parameters provided in `ChatService`

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
```diff
@@ -0,0 +1,254 @@
...
+                    }
+                }
+                case TEXT -> {
+                    // Do nothing.
+                }
+            }
+        }
+
+        if (parameters instanceof WatsonxChatRequestParameters watsonxParameters) {
+            builder.projectId(watsonxParameters.projectId());
```
dliubarskyi, line 121: What if `defaultParameters` are also `WatsonxChatRequestParameters`? `getOrDefault` is required here

docs/docs/integrations/language-models/watsonx.md
andreadimaio: Could you please explain that better? Is your idea to create the TextChatMessageControl automatically? For example, if ExtractionTags is not null, would you like to add the TextChatMessageControl automatically to the list of messages to be sent?

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
```diff
@@ -0,0 +1,254 @@
...
+                .name(partialToolCall.name())
+                .partialArguments(partialToolCall.arguments())
+                .build();
+    }
+
+    public static ChatParameters toChatParameters(
+            ChatRequestParameters defaultParameters, ChatRequestParameters parameters) {
+
+        ChatParameters.Builder builder = ChatParameters.builder()
+                .modelId(getOrDefault(parameters.modelName(), defaultParameters.modelName()))
```
dliubarskyi: This merge here should not be required, as it already happens in `ChatModel.chat()`:

```java
ChatRequest finalChatRequest = ChatRequest.builder()
                .messages(chatRequest.messages())
                .parameters(defaultRequestParameters().overrideWith(chatRequest.parameters()))
                .build();
```

if `WatsonxChatRequestParameters.overrideWith()` is implemented properly, then it should anlready do the job

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
dliubarskyi: Please ignore, this merge is not required (see comment above)

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
dliubarskyi: `Converter.toChatParameters` is invoked too late, after `ChatModelListener.onRequest()` was called. Listener will not get the correct values

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
```diff
@@ -0,0 +1,194 @@
...
+    public Set<Capability> supportedCapabilities() {
+
+        ResponseFormat responseFormat = defaultRequestParameters.responseFormat();
+
+        if (isNull(responseFormat)) return Set.of();
+
+        if (responseFormat.type().equals(ResponseFormatType.JSON) && nonNull(responseFormat.jsonSchema()))
+            return Set.of(RESPONSE_FORMAT_JSON_SCHEMA);
+
+        return Set.of();
```
dliubarskyi: If user configures `WatsonxChatModel.builder().supportedCapabilities(RESPONSE_FORMAT_JSON_SCHEMA)`, this method should return `RESPONSE_FORMAT_JSON_SCHEMA`

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
dliubarskyi: But why? Many of the fields seemed to be useful? Like actual model name, version, etc

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxStreamingChatModel.java
```diff
@@ -0,0 +1,215 @@
...
+    public Set<Capability> supportedCapabilities() {
+
+        ResponseFormat responseFormat = defaultRequestParameters.responseFormat();
+
+        if (isNull(responseFormat)) return Set.of();
+
+        if (responseFormat.type().equals(ResponseFormatType.JSON) && nonNull(responseFormat.jsonSchema()))
+            return Set.of(RESPONSE_FORMAT_JSON_SCHEMA);
+
+        return Set.of();
```
dliubarskyi: same here

dliubarskyi (commented):
@andreadimaio thank you!

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
andreadimaio: Now is clear, thanks.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: Oops! I also removed the `modelVersion`, my mistake. I think the useful fields in the `WatsonxChatResponseMetadata` class are the 'created' and 'version' fields.

andreadimaio:
@dliubarskyi I have done a small refactor to align the watsonx.ai integration with other LangChain4j modules.  
Now also the parameters part should be ok. Let me know what you think about these changes.

docs/docs/integrations/language-models/watsonx.md
dliubarskyi: Other LLM provider APIs have dedicated parameters to enable thinking, but here a chat message is (mis)used to enable thinking, which seems very weird. Messages are for representing a conversation, not configurations for the LLM...

But if the API is already set in stone, then creating `TextChatMessageControl` automatically under the hood (if some property is set, like `WatsonxChatRequestParameters.enableThinking`) would probably be better than asking user to append `CustomMessage` each time they want to call the LLM

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    private static AssistantMessage toAssistantMessage(AiMessage aiMessage) {
+        if (aiMessage.hasToolExecutionRequests()) {
+            var toolCalls = aiMessage.toolExecutionRequests().stream()
+                    .map(Converter::toToolCall)
+                    .toList();
+            return AssistantMessage.tools(toolCalls);
+        }
+        return AssistantMessage.text(aiMessage.text());
```
dliubarskyi: In theory `AiMessage` can contain all of:
- thinking
- tool calls
- text

But the logic in this method assumes that `AiMessage` can contain either tool calls or text

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
```diff
@@ -0,0 +1,168 @@
...
+                chatRequest.parameters().toolSpecifications(), defaultRequestParameters.toolSpecifications());
+
+        List<ChatMessage> messages =
+                chatRequest.messages().stream().map(Converter::toChatMessage).collect(toCollection(ArrayList::new));
+
+        List<Tool> tools = nonNull(toolSpecifications) && toolSpecifications.size() > 0
+                ? toolSpecifications.stream().map(Converter::toTool).toList()
+                : null;
+
+        if (isThinkingActivable(chatRequest.messages(), toolSpecifications)) messages.add(THINKING);
```
dliubarskyi, line 69: Shouldn't thinking be activated by the user explicitly enabling it?

dliubarskyi:
@andreadimaio looks good, thank you!

There are a few comments left open, please check.

I also don't have credits any more to test, so could you please run all the tests at the very end and configur that all is good an ready for merge/release? FYI I will be releasing this thursday, so it this PR will be ready, it can go into the release

docs/docs/integrations/language-models/watsonx.md
andreadimaio: Yes, this is how watsonx.ai works today. To enable the thinking part, a `TextChatMessageControl` is needed.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: The problem here (as you can see from the code) is that today, watsonx.ai has some limitations when the thinking part is enabled. First of all, today only Granite is enabled for thinking, and second, when it is enabled, no SystemMessage or Tools can be used. What happens today is that if one of these is sent together with the `TextChatMessageControl` message, the thinking part will be ignored. I have added warning logs to notify the developer. Looking at the code, maybe an exception would be better.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
andreadimaio: I could pass all the values, but in watsonx.ai, if both `text` and `tool calls` are present, only the latter is considered.

andreadimaio: In any case, if both values are passed, no errors are returned by the watsonx.ai API. If you would prefer, I can change the code.

andreadimaio:
> @andreadimaio looks good, thank you!
> 
> There are a few comments left open, please check.
> 
> I also don't have credits any more to test, so could you please run all the tests at the very end and configur that all is good an ready for merge/release? FYI I will be releasing this thursday, so it this PR will be ready, it can go into the release


All the tests are green. There is only one problem:
https://github.com/langchain4j/langchain4j/blob/64c4197e360f8af25bb6cf78b62a696e16b45597/langchain4j-core/src/test/java/dev/langchain4j/model/chat/common/AbstractBaseChatModelIT.java#L1479-L1482

The `readBytes()` method returns `403`, but this should affect all modules, not only watsonx.ai.

```
java.lang.RuntimeException: java.lang.RuntimeException: Error while reading: 403
 at dev.langchain4j.internal.Utils.readBytes([Utils.java:339](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html))
 at dev.langchain4j.model.chat.common.AbstractBaseChatModelIT.should_accept_single_image_as_base64_encoded_string([AbstractBaseChatModelIT.java:1482](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html))
```

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxTokenCountEstimator.java
andreadimaio: I will try to implement these methods before Thursday, but due to other commitments, I'm not sure I will be able to. We could leave this PR open to see if I can implement these changes. Otherwise, if it's OK with you, we can release a first version without them.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
dliubarskyi: I think it makes sense to supply all the available information and leave it to the Watsonx API to decide what to do with it

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
dliubarskyi: AFAIU current implementation always enables thinking when there is no system message and tools. But what if user does not want thinking at all? Is there a way to disable it? As a user, I would expect to have some switch to enable/disable thinking that I can set when creating `WatsonxChatModel` or when creating `ChatReqeust`

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxTokenCountEstimator.java
dliubarskyi: @andreadimaio no worries, it is not critical, feel free to address it in a separate PR

dliubarskyi:
@andreadimaio thank you! There are 2 non-critical comments left open (`AiMessage` mapping and thinking), I am not sure if you want to address them in this PR or not. Please LMK when this should be merged

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: To enable thinking, the developer must pass the `ExtractionTags` to the builder.

```java
boolean isThinkingActivable(List<ChatMessage> messages, List<ToolSpecification> tools) {
        if (isNull(tags)) return false;
```

If no tags are present, thinking will always be disabled.

andreadimaio: The switch is
```java
WatsonxChatModel.builder()
            ...
            .thinking(ExtractionTags.of("think", "response"))
            .build();
```

andreadimaio: Another point to consider is whether logging a warning is sufficient or whether it would be better to send an exception if the developer enables the thinking part and sends one or more tools or a SystemMessage.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/Converter.java
andreadimaio: Both text and tool calls are now used.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
dliubarskyi: @andreadimaio got it, thanks for the explanation!

Yes, I would rather throw an exception, otherwise it can become confusing (logs not enabled, etc)

andreadimaio: @dliubarskyi, please take a look at the latest commit.

andreadimaio:
> @andreadimaio thank you! There are 2 non-critical comments left open (`AiMessage` mapping and thinking), I am not sure if you want to address them in this PR or not. Please LMK when this should be merged

@dliubarskyi, If the latest commit is also OK, the PR is complete.
However, I would ask you to wait until tomorrow before merging it (this will give me time to try to close the tokenization part).

dliubarskyi:
@andreadimaio should I merge it now?

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
dliubarskyi: Looks good, although I would use more specific `UnsupportedFeatureException` instead of `LangChain4jException`

andreadimaio:
Yes, I didn't have the time to merge the tokenization part. I will open a new PR.

langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
andreadimaio: I will incorporate this change into the new PR.

andreadimaio:
@dliubarskyi I could work on it in the early afternoon, but I'm not sure if you need to merge it right away.

dliubarskyi:
@andreadimaio thanks a lot for the PR, great job!

Please open a separate PR for the leftover tasks

PR #3459, opened 2025-07-31
AI Service: fix implementation of methods declared in Object

dliubarskyi:
## Issue

Currently, `equals` and `toString` behave inconsistently for the (proxy) instance of the AI Service:
- `assistant.equals(assistant)` returns `false`
- `assistant.toString()` returns non-informative `dev.langchain4j.service.DefaultAiServices$1@1144a55a` instead of mentioning the interface of the AI Service

Also, default methods defined on AI Service interface do not work.

## Change
- Fixed the implementation of `equals`, `hashCode` and `toString` methods on AI Service instance.
- Allow default methods on AI Service

## General checklist


langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -149,11 +149,24 @@ public T build() {
...
-                    public Object invoke(Object proxy, Method method, Object[] args) throws Exception {
+                    public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+
+                        if (method.isDefault()) {
+                            return InvocationHandler.invokeDefault(proxy, method, args);
+                        }
 
                         if (method.getDeclaringClass() == Object.class) {
-                            // methods like equals(), hashCode() and toString() should not be handled by this proxy
-                            return method.invoke(this, args);
```
dliubarskyi, line 158: The problem here was that `this` refers to the invocation handler, not to the instance of the proxy

mariofusco (dismissed):
Your implementation is reasonable. I just wonder in which circumstances somebody may want to use the `equals` and especially the `hashCode` method. This could be useful only if the agent is used as key in a `Map` which would be pretty odd in my opinion.  In essence the PR is ok for me, but I wonder if we want to prevent AI services to be used as keys for a `Map` and then also return an `IllegalStateException` when the `hashCode` method is called.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesTest.java
```diff
@@ -0,0 +1,88 @@
...
+        // dynamic proxies don't intercept final methods
+        assertThat(assistant.getClass()).isNotEqualTo(Assistant.class);
+    }
+
+    @Test
+    void test_hashCode() throws Exception {
+        assertThat(assistant.hashCode())
+                .isNotZero()
+                .isEqualTo(assistant.hashCode())
+                .isNotEqualTo(AiServices.create(Assistant.class, chatModel).hashCode());
```
kpavlov:
```suggestion
    void test_hashCode() {
        assertThat(assistant.hashCode())
                .isNotZero()
                .isEqualTo(assistant.hashCode())
                .isEqualTo(System.identityHashCode(assistant))
                .isNotEqualTo(AiServices.create(Assistant.class, chatModel).hashCode());
```

langchain4j/src/test/java/dev/langchain4j/service/AiServicesTest.java
```diff
@@ -0,0 +1,88 @@
...
+    void test_getClass() throws Exception {
+        // dynamic proxies don't intercept final methods
+        assertThat(assistant.getClass()).isNotEqualTo(Assistant.class);
+    }
+
+    @Test
+    void test_hashCode() throws Exception {
+        assertThat(assistant.hashCode())
+                .isNotZero()
+                .isEqualTo(assistant.hashCode())
```
kpavlov, line 73:
```suggestion
                .isEqualTo(assistant.hashCode())
                .isEqualTo(System.identityHashCode(assistant))
```

langchain4j/src/test/java/dev/langchain4j/service/AiServicesTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Test
+    void test_getClass() throws Exception {
+        // dynamic proxies don't intercept final methods
+        assertThat(assistant.getClass()).isNotEqualTo(Assistant.class);
+    }
+
+    @Test
+    void test_hashCode() throws Exception {
```
kpavlov:
```suggestion
    void test_hashCode() {
```

kpavlov (dismissed):
Thank you, @dliubarskyi 
This make sense

langchain4j/src/test/java/dev/langchain4j/service/AiServicesTest.java
```diff
@@ -0,0 +1,90 @@
...
+    @Test
+    void test_default_method_throwing_exception() {
+        assertThatThrownBy(() -> assistant.getException())
+                .isExactlyInstanceOf(RuntimeException.class)
+                .hasMessage("oops");
+        verifyNoInteractions(chatModel);
+    }
+
+    @Test
+    void test_getClass() throws Exception {
```
dliubarskyi:
```suggestion
    void test_getClass() {
```

dliubarskyi:
@mariofusco not sure TBH, but I would not restrict even using AI Service as a key, it is up to the user to shoot themselves in the knee 😆. Calling `equals` and `toString` seems reasonably fine, no?

PR #3470, opened 2025-08-01
Fix bug in output guardrails when maxRetries == 0

edeandrea:
There's a bug in the output guardrail execution if maxRetries == 0. Currently it will always execute at least 1 retry if the guardrail says to retry or reprompt.


dliubarskyi (approved):
@edeandrea thank you!

PR #3512, opened 2025-08-08
AI Service: option to execute tools concurrently

### LINKED ISSUE #3290: [FEATURE] AI Service: support executing tools concurrently

This feature is not related to a problem, but an enhancement to the general API.

I hope to add a configuration parameter for parallel calling of tools on `dev.langchain4j.service.AiServices`. Because many LLMs now support the ability to call tools concurrently, such as OpenAI, Anthropic, Qwen-max, etc.

**Additional context**

Image

Image

Image

dliubarskyi:
## Issue
Closes #3290

## Change
Added an option to execute tools concurrently:

```java
    /**
     * By default, when the LLM calls multiple tools, the AI Service executes them sequentially.
     * If you enable this option, tools will be executed concurrently (with one exception - see below),
     * using the default {@link Executor}.
     * You can also specify your own {@link Executor}, see {@link #executeToolsConcurrently(Executor)}.
     * <ul>
     *     <li>When using {@link ChatModel}:</li>
     *     <ul>
     *         <li>When the LLM calls multiple tools, they are executed concurrently in separate threads
     *             using the {@link Executor}.</li>
     *         <li>When the LLM calls a single tool, it is executed in the same (caller) thread,
     *             the {@link Executor} is not used to avoid wasting resources.</li>
     *     </ul>
     *     <li>When using {@link StreamingChatModel}:</li>
     *     <ul>
     *         <li>When the LLM calls multiple tools, they are executed concurrently in separate threads
     *             using the {@link Executor}.
     *             Each tool is executed as soon as {@link StreamingChatResponseHandler#onCompleteToolCall(CompleteToolCall)}
     *             is called, without waiting for other tools or for the response streaming to complete.</li>
     *         <li>When the LLM calls a single tool, it is executed in a separate thread using the {@link Executor}.
     *             We cannot execute it in the same thread because, at that point,
     *             we do not yet know how many tools the LLM will call.</li>
     *     </ul>
     * </ul>
     *
     * @return builder
     * @see #executeToolsConcurrently(Executor)
     * @since 1.4.0
     */
    public AiServices<T> executeToolsConcurrently() {
        context.toolService.executeToolsConcurrently();
        return this;
    }

    /**
     * See {@link #executeToolsConcurrently()}'s Javadoc for more info.
     * <p>
     * If {@code null} is specified, the default {@link Executor} will be used.
     *
     * @param executor The {@link Executor} to be used to execute tools.
     * @return builder
     * @see #executeToolsConcurrently()
     * @since 1.4.0
     */
    public AiServices<T> executeToolsConcurrently(Executor executor) {
        context.toolService.executeToolsConcurrently(executor);
        return this;
    }
```

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -166,8 +166,8 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
             } else {
                 for (ToolExecutionRequest toolExecutionRequest : aiMessage.toolExecutionRequests()) {
                     handleBeforeTool(toolExecutionRequest);
-                    // TODO applyToolHallucinationStrategy
                     ToolExecutor toolExecutor = toolExecutors.get(toolExecutionRequest.name());
+                    // TODO applyToolHallucinationStrategy
                     String toolExecutionResult = toolExecutor.execute(toolExecutionRequest, memoryId);
                     handleAfterTool(toolExecutionRequest, toolExecutionResult);
                     addToMemory(ToolExecutionResultMessage.from(toolExecutionRequest, toolExecutionResult));
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -15,6 +15,7 @@ import dev.langchain4j.guardrail.OutputGuardrailRequest;
 import dev.langchain4j.memory.ChatMemory;
 import dev.langchain4j.model.chat.request.ChatRequest;
 import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.model.chat.response.CompleteToolCall;
 import dev.langchain4j.model.chat.response.PartialThinking;
 import dev.langchain4j.model.chat.response.StreamingChatResponseHandler;
 import dev.langchain4j.model.output.TokenUsage;
@@ -24,6 +25,11 @@ import dev.langchain4j.service.tool.ToolExecutor;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Queue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executor;
 import java.util.function.Consumer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -57,6 +63,9 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
 
     private final List<ToolSpecification> toolSpecifications;
     private final Map<String, ToolExecutor> toolExecutors;
+    private final Executor toolExecutor;
+    private final Queue<CompletableFuture<ToolExecutionResultMessage>> toolResultFutures = new ConcurrentLinkedQueue<>();
+
     private final List<String> responseBuffer = new ArrayList<>();
     private final boolean hasOutputGuardrails;
 
@@ -75,6 +84,7 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
             TokenUsage tokenUsage,
             List<ToolSpecification> toolSpecifications,
             Map<String, ToolExecutor> toolExecutors,
+            Executor toolExecutor,
             GuardrailRequestParams commonGuardrailParams,
             Object methodKey) {
         this.chatExecutor = ensureNotNull(chatExecutor, "chatExecutor");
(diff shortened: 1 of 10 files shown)


langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -116,6 +128,18 @@ public void onPartialThinking(PartialThinking partialThinking) {
         }
     }
 
+    @Override
+    public void onCompleteToolCall(CompleteToolCall completeToolCall) {
+        if (toolExecutor != null) {
```
mariofusco, line 133: I'm sure if I'm understanding the logic here, but if the `toolExecutor` is `null` it should mean that the tool execution should be performed sequentially and synchronously on this same thread, while here it isn't doing anything. Is this correct?

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -78,6 +87,28 @@ private void processToolMethod(Object objectWithTool, Method method) {
         toolSpecifications.add(toolSpecificationFrom(method));
     }
 
+    /**
+     * @since 1.4.0
+     */
+    public void executeToolsConcurrently() {
```
mariofusco, line 92: I believe that there should be one single default executor share among all `ToolService` instances and it could be created lazily in a more thread safe way through the lazy initialization holder class idiom.

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
dliubarskyi: Yes, because tools are executed sequentially in the `onCompleteResponse` callback, when the streaming is over. `onCompleteToolCall` is called after each tool call streaming is done, but we can't execute it synchronously here because it would block the streaming of other stuff

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -381,6 +384,55 @@ public AiServices<T> tools(Map<ToolSpecification, ToolExecutor> tools) {
...
+     * See {@link #executeToolsConcurrently()}'s Javadoc for more info.
+     * <p>
+     * If {@code null} is specified, the default {@link Executor} will be used.
+     *
+     * @param executor The {@link Executor} to be used to execute tools.
+     * @return builder
+     * @see #executeToolsConcurrently()
+     * @since 1.4.0
+     */
+    public AiServices<T> executeToolsConcurrently(Executor executor) {
```
mariofusco, line 431: Consider asking for an `ExecutorService` here instead of a pure `Executor`. The difference for users will be minimal and you will be allowed to use many of the convenient method made available by that interface.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
dliubarskyi: > I believe that there should be one single default executor share among all ToolService instances

`Executor` is configured per AI Service, each AI Service has a single `ToolService`. I am not sure it will be possible to share the same instance without confusing the user.

> it could be created lazily in a more thread safe way through the lazy initialization holder class idiom.

Do you mean create it (if not created) right before the first tool is executed?

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
dliubarskyi: When this will be integrated in Quarkus and Spring, there will be a need for context propagation, and in the case of Spring, we will need to use `AsyncTaskExecutor` that can propagate context, but it is `Executor`, not `ExecutorService`

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: > > I believe that there should be one single default executor share among all ToolService instances
> 
> `Executor` is configured per AI Service, each AI Service has a single `ToolService`. I am not sure it will be possible to share the same instance without confusing the user.

I'm speaking of the default executor, the one in use when the user says that he wants parallel tool execution without providing his own executor, that is an internal thing, not exposed to the users at all, or am I missing something?

> > it could be created lazily in a more thread safe way through the lazy initialization holder class idiom.
> 
> Do you mean create it (if not created) right before the first tool is executed?

Yes.

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
mariofusco: Ok, if so probably I should do the same also in the parallel workflow of the new agentic module and use an `Executor` instead of an `ExecutorService` also there.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
dliubarskyi: > I'm speaking of the default executor, the one in use when the user says that he wants parallel tool execution without providing his own executor, that is an internal thing, not exposed to the users at all, or am I missing something?

Good point, makes sense! Changed.

> > > it could be created lazily in a more thread safe way through the lazy initialization holder class idiom.
> > Do you mean create it (if not created) right before the first tool is executed?
> Yes.

Not sure we should wait untill the first tool is executed. If user explicitly calls `executeToolsConcurrently()`, it means that tools will be executed, so I see no issue with creating it eagerly. It is also keeps the implementation simpler

PR #3513, opened 2025-08-08
Docker MCP transport

### LINKED ISSUE #3493: [FEATURE] Docker MCP Transport
Current MCP transports are HTTP and Stdio.
Most existing MCP servers are provided as Stdio.
But Stdio uses a Process under the cover which is not secure (can access the host) and mandate to have some tool installed (node, docker, ...).

Many existing MCP servers also provide a Docker container (afaik, the Stdio transport is often used with the `docker` command). Having a Docker transport would be more secured (isolated container) and would not mandate any tool installed on the host.

Provides a Docker transport, based on the Java Docker client.

Using the Stdio transport.

**Additional context**
I'm working at Kestra which uses Langchain4J for its AI plugin.
On secured installation, the Stdio transport is excluded as we prohibit starting new processes for security reason.

I'm currently working on a prototype, I would be able to provide a PR for this feature.

loicmathieu:
PR raised to open discussion for the implementation, it's only a prototype for now.

## Issue

Closes #3493

## General checklist


dliubarskyi:
@loicmathieu thank you!

@jmartisk could you please review? 🙏

jmartisk:
Hm, I'm trying to get this to work with my local Podman installation,but I can't. My piece of code:

```
DockerMcpTransport transport = new DockerMcpTransport.Builder()
                .host("unix:///run/user/1000/podman/podman.sock")
                .image("docker.io/mcp/fetch")
                .logEvents(true)
                .build();
        DefaultMcpClient client = new DefaultMcpClient.Builder()
                .transport(transport)
                .build();

        System.out.println(client.listTools());
```

It seems that the initialization goes through successfully (the client logs the server's capabilities etc), but after sending `{"jsonrpc":"2.0","id":1,"method":"tools/list"}` to the server, there is no response. The last log is
```
10:35:50.730 [docker-java-httpclient5-hijacking-stream-1725520335] DEBUG org.apache.hc.client5.http.wire -- http-outgoing-1 >> "{"jsonrpc":"2.0","id":1,"method":"tools/list"}[\n]"
```
and the container is in `INITIALIZED` state (I would expect RUNNING)

jmartisk:
If you have a working example that you could share and later submit to https://github.com/langchain4j/langchain4j-examples that would be nice

loicmathieu:
@jmartisk I added a test that passes.
But there is still an issue, there is a timeout somewhere, then a chekcHealth, a restart, and it works after.
So calling the tool took more than 30s, I still need to find the cause of this.

langchain4j-mcp/pom.xml
```diff
@@ -34,6 +34,33 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>com.github.docker-java</groupId>
+            <artifactId>docker-java</artifactId>
```
dliubarskyi: If `DockerMcpTransport` requires extra dependencies, let's separate it into a `langchain4j-mcp-docker` or similar

dliubarskyi (commented):
@loicmathieu thanks!

langchain4j-mcp/pom.xml
loicmathieu: OK, as discussed with Jan, I also need to add an SPI for the Docker transport.
And of course, fix the issue with the timoueout leading to any calls ending in more than 30s ;)

loicmathieu: I extract it to a new module

loicmathieu:
@dliubarskyi @jmartisk I fixed the timeout issue, extracted it to a new module add test and documentation.
It's ready to go!

docs/docs/tutorials/mcp.md
```diff
@@ -62,14 +65,33 @@ McpTransport transport = new HttpMcpTransport.Builder()
...
+<dependency>
+    <groupId>dev.langchain4j</groupId>
+    <artifactId>langchain4j-mcp-docker</artifactId>
+</dependency>
+```
+
+Then you need to create a Docker transport:
+
+```java
+McpTransport transport = new StdioMcpTransport.Builder()
```
dliubarskyi: shouldn't it be `DockerMcpTransport`?

langchain4j-mcp-docker/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.github.docker-java</groupId>
+            <artifactId>docker-java-transport-httpclient5</artifactId>
+            <version>3.5.3</version>
+        </dependency>
+        <!-- Fix dependency divergence issue -->
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
```
dliubarskyi: This should probably be in the `dependencyManagement` section?

langchain4j-mcp-docker/src/test/java/dev/langchain4j/mcp/client/transport/docker/DockerMcpTransportIT.java
```diff
@@ -0,0 +1,117 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+class DockerMcpTransportIT {
+    private static McpClient mcpClient;
+
+    @BeforeAll
+    static void setup() {
+        McpTransport transport = new DockerMcpTransport.Builder()
+                .image("mcp/time")
+                .dockerHost("unix:///var/run/docker.sock")
```
dliubarskyi, line 34: How will this work on GitHub CI?

pom.xml
```diff
@@ -20,6 +20,7 @@
 
         <module>langchain4j-easy-rag</module>
         <module>langchain4j-mcp</module>
+        <module>langchain4j-mcp-docker</module>
```
dliubarskyi, line 23: Please add it to the `langchain4j-bom` as well

docs/docs/tutorials/mcp.md
```diff
@@ -14,6 +14,9 @@ The protocol specifies two types of transport, both of these are supported:
   can run an MCP server as a local subprocess and
   communicate with it directly via standard input/output.
 
+Additionally, Langchain4J supports a Docker stdio transport that can use a stdio MCP server distributed as a
```
```suggestion
Additionally, LangChain4j supports a Docker stdio transport that can use a stdio MCP server distributed as a 
```

dliubarskyi (commented):
@loicmathieu thanks!

@jmartisk could you please review as well?

langchain4j-mcp-docker/src/test/java/dev/langchain4j/mcp/client/transport/docker/DockerMcpTransportIT.java
loicmathieu: I see testcontainers is used so Docker is available in CI.
I'm not sure if it uses the standard Unix socket, if not, I'll update the test accordingly.

docs/docs/tutorials/mcp.md
loicmathieu: Yes, obviously. Fixed

langchain4j-mcp-docker/pom.xml
loicmathieu: Done

pom.xml
loicmathieu: Done

docs/docs/tutorials/mcp.md
loicmathieu: Done

jmartisk:
Sorry for the delay, going to try this today

docs/docs/tutorials/mcp.md
```diff
@@ -62,14 +65,34 @@ McpTransport transport = new HttpMcpTransport.Builder()
...
+
 ### MCP Client
 
 To create an MCP client from the transport:
 
 ```java
-McpClient mcpClient = new DefaultMcpClient.Builder()
-    .key("MyMCPClient")
-    .transport(transport)
+McpTransport transport = new DockerMcpTransport.Builder()
```
jmartisk: This change doesn't make sense to me, here you're supposed to show how to create a client from the transport, but you're showing the transport again

langchain4j-mcp-docker/src/test/java/dev/langchain4j/mcp/client/transport/docker/DockerMcpTransportIT.java
jmartisk, line 34: I guess we should make this test automatically skipped if this socket does not exist. Otherwise, it will unnecessarily fail in environments where it's not expected to work (Windows for sure). For example, on my machine with Podman, the file is `unix:///run/user/1000/podman/podman.sock` (but even after changing the path, the test doesn't work for me). Maybe the path should even be configurable using a system property??

jmartisk:
I still can't get this to work with Podman (but yes, Podman is often a source of problems...). It just gets stuck after the initialization phase and when attempting to invoke any operations, the docker client doesn't receive any response anymore, and the container is still in status=Initialized. I guess I don't have enough knowledge to debug this though, so I guess I'll consider it working if I see the test passing at least in the CI with Docker...

docs/docs/tutorials/mcp.md
loicmathieu: Sorry, bad copy/paste

langchain4j-mcp-docker/src/test/java/dev/langchain4j/mcp/client/transport/docker/DockerMcpTransportIT.java
loicmathieu: I made it conditional on the existance of the Docker socket.

loicmathieu:
@jmartisk there is a non-easy issue with how Docker send frames when you attach to a container.
I had a hard time making it work for Docker, it may not be correct for Podman.

The `DockerResultCallback` contains logic to know when to send the response to the MCP client:
- On each frame we aggregate it inside `logAggregator` until we find a new line delimiter
- If we receive an empty frame and have something in the logAggregator we consider we should also send the message

If Podman didn't send empty frames or newline, we may have issues with it.

There is also the case of not waiting for any response from the container attach command, in this case I wait for 100ms to be sure the MCP server receive and process the message. We may increase this but as far as I know there is no reliable way to be sure that it has been received (and I'm surprised we don't have the same issue with the Stdio transport)

langchain4j-mcp-docker/src/main/java/dev/langchain4j/mcp/client/transport/docker/DockerMcpTransport.java
```diff
@@ -0,0 +1,310 @@
...
+                PipedOutputStream out = new PipedOutputStream();
+                PipedInputStream in = new PipedInputStream(out)
+        ) {
+            DockerResultCallback callback = dockerClient.attachContainerCmd(containerId)
+                    .withStdOut(true)
+                    .withStdErr(true)
+                    .withFollowStream(true)
+                    .withStdIn(in)
+                    .exec(new DockerResultCallback(logEvents, messageHandler));
+            callback.awaitStarted(30, TimeUnit.SECONDS);
```
jmartisk: Make this configurable?

loicmathieu: Done.

loicmathieu:
@jmartisk I allow configuring the attach timeout and fixed a bug with STDERR handling, this bug may explain some testing issue as I previously throw an error when an STDERR frame was received but it's pretty common for Python for ex to log in STDERR instead of STDOUT.

loicmathieu:
@jmartisk I allow configuring container binds because I needed them to gather files from MCP tools that generates files but as `HostConfig` can contains tons of different config, I wonder if the best would not be to allow passing the HostConfig.
This would allow for more extensibility but would leak the Docker API inside Langchain4J.

loicmathieu:
@jmartisk silly bug! I forgot to pull the image! This can explain why you was not able to make it work.

dliubarskyi:
Hi, what's the state of this PR? Is it ready to be merged and released?

loicmathieu:
@dliubarskyi it should be reviewed a last time by @jmartisk.

jmartisk:
Unfortunately with Podman I'm still stuck. But we would probably need somebody who understands it better.
Can we at least get it working in the CI? I don't see the `DockerMcpTransportIT` running there

loicmathieu:
@jmartisk in the CI, only unit tests are running not integration test, I switch my test to a unit test if needed but I did as for the other MCP transport that only have integration test.

jmartisk (dismissed):
I guess we should be good to go

dliubarskyi:
Please take https://github.com/langchain4j/langchain4j/issues/3493#issuecomment-3235469357 into account

loicmathieu:
@dliubarskyi I'm not sure how this comment applies here.
The Docker MCP gateway is itself a Docker container that allow to use MCP server as Docker containers, I guess MCP servers that are not container based. This this is more something that can be used with this new Docker transport.

As for testcontainers support, again, it's to run an MCP server as a container in a test context so, well, it can avoid using this transport in test but we already have test for MCP server without Docker...

loicmathieu:
@dliubarskyi is there something I should still do for this PR to be merged?

langchain4j-bom/pom.xml
```diff
@@ -263,6 +263,12 @@
                 <version>1.7.0-beta13-SNAPSHOT</version>
             </dependency>
 
+            <dependency>
+                <groupId>dev.langchain4j</groupId>
+                <artifactId>langchain4j-mcp-docker</artifactId>
+                <version>1.4.0-beta10-SNAPSHOT</version>
```
```suggestion
                <version>1.7.0-beta13-SNAPSHOT</version>
```

langchain4j-mcp-docker/pom.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>1.4.0-beta10-SNAPSHOT</version>
```
dliubarskyi:
```suggestion
        <version>1.7.0-beta13-SNAPSHOT</version>
```

langchain4j-mcp-docker/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+                <version>2.19.0</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-mcp</artifactId>
+            <version>1.4.0-beta10-SNAPSHOT</version>
```
dliubarskyi:
```suggestion
            <version>1.7.0-beta13-SNAPSHOT</version>
```

langchain4j-mcp-docker/src/main/java/dev/langchain4j/mcp/client/transport/docker/DockerMcpTransport.java
```diff
@@ -0,0 +1,369 @@
...
+            return this;
+        }
+
+        public DockerMcpTransport.Builder attachTimeout(Duration attachTimeout) {
+            this.attachTimeout = attachTimeout;
+            return this;
+        }
+
+        public DockerMcpTransport build() {
+            if (dockerHost == null || dockerHost.isEmpty()) {
```
dliubarskyi, line 348: nit: these validations are better off inside the `DockerMcpTransport` ctor

langchain4j-mcp-docker/src/main/java/dev/langchain4j/mcp/client/transport/docker/DockerResultCallback.java
```diff
@@ -0,0 +1,83 @@
...
+import com.github.dockerjava.api.model.Frame;
+import com.github.dockerjava.api.model.StreamType;
+import dev.langchain4j.mcp.client.transport.McpOperationHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+
+public class DockerResultCallback extends ResultCallback.Adapter<Frame> {
```
dliubarskyi:
```suggestion
class DockerResultCallback extends ResultCallback.Adapter<Frame> {
```

dliubarskyi (dismissed):
@loicmathieu thank you!

PR #3516, opened 2025-08-08
Ignore @JsonIgnore fields.  Add superclass fields

### LINKED ISSUE #3515: [BUG] ToolSpecification schema: ignore @JsonIgnore fields. Add superclass fields
ToolSpecification schema generation for tool object params should ignore @JsonIgnore fields and add superclass fields.  I labeled this as a bug as this snuck up on me.  I assumed that it @JsonIgnore and superclass fields were included in schema generation and didn't find out until I looked at raw requests to OpenAI.

PR incoming

patriot1burke:
## Issue
Closes #3515

## Change
See issue.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
@@ -4,6 +4,7 @@ import static dev.langchain4j.internal.Utils.generateUUIDFrom;
 import static java.lang.reflect.Modifier.isStatic;
 import static java.util.Arrays.stream;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import dev.langchain4j.Internal;
 import dev.langchain4j.model.chat.request.json.JsonAnyOfSchema;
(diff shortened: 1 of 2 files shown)


sunyuhan1998 (approved):

patriot1burke:
@dliubarskyi 

Additional changes were to get by spotless checks.

dliubarskyi (commented):
@patriot1burke thanks a lot!

These are both breaking changes. What about guarding them with some feature toggles? For example, it could be configured when AI Service is built:
```java
AiServices.builder(Assistant.class)
                .chatModel(model)
                .respectJsonIgnoreAnnotations(true)
                .respectInheritedProperties(true)
                .build();
```

OR

```java
AiServices.builder(Assistant.class)
                .chatModel(model)
                .enable(STRUCTURED_OUTPUTS_RESPECT_JSON_IGNORE_ANNOTATIONS, STRUCTURED_OUTPUTS_RESPECT_INHERITED_PROPERTIES)
                .build();
```

For the 2.0 version of LC4j, we could consider enabling them automatically.

WDYT?

dliubarskyi (commented):
BTW, there is also `@JsonIgnoreProperties` annotation which can be used to specify which fields should be ignored (on the class level)

patriot1burke:
@dliubarskyi 
IMO, this PR is a **bug** fix, not "breaking changes" and backward compatibility flags are not needed.   Since you already support the use of JsonProperty, one would would assume that you are using Jackson within langchain4j and that inheritance would just work.  For @JsonIgnore, can you really see somebody having JsonIgnore on a field, but expecting it to be marshalled for tool invocations?  

In my app, I expected inheritance and JsonIgnore  to just work and it didn't.  My AI interaction was flaky and I didn't figure out why until I saw the message payload and the langchain4j codebase.

I'll add support for JsonIgnoreProperties

dliubarskyi:
@patriot1burke I understand that from your point of view it is a bug fix, but I am not convinced that it is safe to simply change the behaviour and assume it will not affect existing users (I am sure it will, at least in some corner cases). To be on the safe side, I would prefer to guard it by feature flags. In the version 2.0 we can remove them and have it as default behaviour.

PR #3525, opened 2025-08-11
Allow passing null memory or memory provider

loicmathieu:
This is not an important change so it can be dismissed.

When using the AiServices to build an AiService, most if not all methods of the builder allow passing null which is convenient when you want to have a generic way of instantiating an Ai service.
The tow ones that didn't allow that are `chatMemory` and `chatMemoryProvider` as they init the memory immediately.

Allowing to pass null would facilitate code that want to conditionnaly enable a memory as a default memory is anyway created inside all AiService is none is configured. Obviously, this PR add an if inside langchain4j instead of inside the user code so I'll totally agree if it would be dismissed.

## Issue

Closes #

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
@@ -260,7 +260,9 @@ public abstract class AiServices<T> {
      * @return builder
      */
     public AiServices<T> chatMemory(ChatMemory chatMemory) {
-        context.initChatMemories(chatMemory);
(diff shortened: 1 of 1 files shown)


loicmathieu:
Another solution would be to allow public access to the default ChatMemoryStore either by moving `SingleSlotChatMemoryStore` from package private to public or by having a `ChatMemoryStore.default()` static method that return an instance of it.

dliubarskyi (approved):
@loicmathieu thanks!

PR #3536, opened 2025-08-12
AI Service: introduce tool arguments and tool execution error handlers

### LINKED ISSUE #3363: [FEATURE] Allow for intercepting exceptions thrown by @Tool methods
Exceptions thrown inside methods annotated with `@Tool` get propagated to the LLM without any indication being given to the client.

This is bad because:
1) Error messages with sensitive information can inadvertently leak into the LLM and become exposed to the user
2) The client has no visibility into the errors without prompting the LLM
3) The client does not have the ability to return a value to the LLM that's more appropriate than the exception
4) The client is inclined to overlook error handling and have 1-3 happen until they realize the default behavior is not what they expected

The client can currently workaround 1-3 by surrounding every `@Tool` method with a try catch but that seems like an anti-pattern.

One idea is to let the client configure the default error handling behavior during `AiService` composition, also with the ability to transform the exception into some other response that the LLM will ultimately receive.

Pardon the Kotlin solution, I haven't written Java in a while.
```
return AiServices.builder(VideoCreationOrchestrator::class.java)
            .chatModel(chatModelRigidJson)
            .chatMemoryProvider(userMemoryProvider)
            .onToolError(
                { toolContext, exception ->
                    // We can log or take some other action here
                    logger.error("Error in ${toolContext.toolName}", exception)

                    // This is what the LLM receives.
                    // I suppose the client could also choose to re-throw the exception here if they want to crash.
                    return@onError "Something went wrong"
                },
            )
            .tools(
                RequirementsChecker,
                ImageAnalyzer
            )
            .build()
```

There may be a better solution that encourages the client to handle the error at the tool level versus where the `AiService` is constructed.

Try catch the entire `@Tool` method's implementation. It works but introduces boilerplate and is easy to miss.

**Additional context**

dliubarskyi:
## Issue
Closes #3363

## Change
Introduced `ToolArgumentsErrorHandler` and `ToolExecutionErrorHandler` that can be configured on the AI Service to handle errors related to tools:
- when tool arguments cannot be parsed/prepared in order to execute a tool
- when something wrong happens during tool execution

Currently, there are two ways to handle errors:
1. Throw an exception: this will stop the AI service flow.
2. Return a text message (e.g., an error description) that will be sent back to the LLM,
allowing it to respond appropriately (for example, by correcting the error and retrying).

## Example

```java
    .chatModel(chatModel)
    .tools(tools)
    .toolArgumentsErrorHandler((error, context) -> ToolErrorHandlerResult.text("Can't parse tool arguments: " + error.getMessage()))
    .toolExecutionErrorHandler((error, context) -> ToolErrorHandlerResult.text("Can't execute tool: " + error.getMessage()))
    .build();
```

## Breaking Changes
This PR requires breaking changes in `DefaultMcpClient.executeTool(ToolExecutionRequest)`: instead of returning strings with error messages, `ToolArgumentsException`, `ToolExecutionException` or `RuntimeException` will be thrown in case of errors.

## General checklist


langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -237,9 +252,26 @@ private String execute(ToolExecutionRequest toolExecutionRequest) {
...
+        } catch (Exception e) {
+            if (e instanceof ToolExecutionException) {
+                ToolErrorContext errorContext = ToolErrorContext.builder()
+                        .toolExecutionRequest(toolExecutionRequest)
+                        .memoryId(memoryId)
+                        .build();
+                ToolErrorHandlerResult errorHandlerResult = toolErrorHandler.handle((Exception) e.getCause(), errorContext);
+                return errorHandlerResult.text();
+            } else {
+                throw e; // TODO should be handled the same way? e.g. errors when with arguments
```
karolzdebel: When you say "errors when with arguments" what does this mean in practice? What would be an example scenario triggering this kind of error?

It would be interesting to be able to handle errors where the arguments failed to be serialized into JSON, whether when calling the tool or parsing the response. I would prefer to retry the tool call a few times in those situations. Not sure whether you're referring to that here though.

dliubarskyi: Pardon, I wanted to write "errors when parsing arguments".
Let's say LLM hallucinated and did not provide the right type or value of the argument (e.g., not existing enum value). In this case we cannot even call the tool.

I am not sure yet if we want to have a separate handler for this kind of errors or re-use this one.

To systematize, right now `ToolExecutor` can propagate 2 types of errors:
- Errors happening **_before_** tool call (e.g., can`t parse/prepare arguments)
- Errors happening **_during_** tool call (e.g., thrown by the tool itself)

Errors happening **_after_** the tool call does not seem to be possible, as we just serialize tool result into a JSON.

In case of `McpToolExecutor`, there might be other types of issues (e.g., can't connect to the MCP server), which we also need to consider. I guess they logically fall into "during" category.

Sorry for the brain dump 😆

karolzdebel: Got it. 

To me it feels like you would want to reuse the `ToolErrorHandler` for both cases, except with different exceptions: `ToolExecutionException`, `ToolArgumentParsingException`, etc. 

Although technically the error isn't the tool's fault, as a client I perceive the error to be closer to the tool than the LLM calling the tool, if that makes sense. `ToolErrorHandler` sounds abstract enough that it could encapsulate the pre, during, and post. I'm not strongly opinionated on this though, probably both approaches could work.

> Errors happening after the tool call does not seem to be possible, as we just serialize tool result into a JSON.

What if the JSON is malformed? i.e. it doesn't fit the object structure.

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -453,6 +455,20 @@ public AiServices<T> hallucinatedToolNameStrategy(
...
+     * // enable {@link DefaultToolExecutor.Builder#propagateToolExecutionException(Boolean)}
+     *
+     * @param toolErrorHandler TODO
+     * @return TODO
+     */
+    public AiServices<T> toolErrorHandler(ToolErrorHandler toolErrorHandler) {
+        context.toolService.errorHandler(toolErrorHandler);
+        return this;
+    }
+
```
misselvexu, line 519: In response to inconsistencies in user scenarios, simplified call scenarios can be added as appropriate.

```java
      // Add convenient methods (simplify usage)
      /**
       * Configure the tool error handler - Simplified version, suitable for most scenarios
       */
      public AiServices<T> onToolError(BiFunction<Exception, ToolExecutionRequest, String> errorHandler) {
          return toolErrorHandler((error, context) ->
              ToolErrorHandlerResult.from(errorHandler.apply(error, context.toolExecutionRequest())));
      }

      // More convenient methods can be added
      /**
       * The most simplified version - Only handle exception messages
       */
      public AiServices<T> onToolError(Function<Exception, String> errorHandler) {
          return toolErrorHandler((error, context) ->
              ToolErrorHandlerResult.from(errorHandler.apply(error)));
      }

```

misselvexu:
@dliubarskyi 
The current anomaly classification and design are relatively simple. Is it more appropriate to focus on the abnormal design of the MCP Tool ?

Here are some of my thoughts.

## 1. Design Objectives

Build a unified, intelligent, and highly resilient exception handling system that seamlessly integrates exception management for both local tool calls and distributed MCP service interactions.

## 2. Core Architecture Design

### 2.1 Unified Exception Hierarchy

```
ToolException (base)
├── ToolArgumentException        // Parameter errors (local + MCP)
├── ToolTimeoutException         // Execution timeout (local + MCP)
├── ToolSecurityException        // Auth/permission issues (local + MCP)
├── ToolUnavailableException     // Service unavailable (local + MCP)
├── ToolTransportException       // MCP transport layer failures
├── ToolProtocolException        // MCP protocol layer failures
├── ToolResourceException        // MCP resource access failures
└── ToolExecutionException       // General execution failures (local + MCP)
```

### 2.2 MCP-Specific Exception Definitions

**Transport Layer Exception**
- Covers connection establishment and communication issues with MCP services
- Key context: `transportType` (STDIO/HTTP/WebSocket), `endpoint`, `reconnectable` flag
- Scenarios: Connection loss, HTTP timeout, STDIO pipe interruption, WebSocket disconnection

**Protocol Layer Exception**
- Indicates violations of MCP communication protocol specifications
- Key context: `protocolVersion`, `messageId`, `errorCode` (MCP-specific error codes)
- Scenarios: Invalid message format, unsupported protocol version, message sequence errors

**Resource Exception**
- Related to failures in accessing specific resources on MCP servers
- Key context: `resourceUri`, `resourceType`, `operation` (GET/LIST/SUBSCRIBE)
- Scenarios: Resource not found, access denied, subscription failure, invalid content

## 3. Enhanced Exception Context

### 3.1 MCP-Aware Context Information

```java
class McpToolErrorContext extends ToolErrorContext {
    // Core MCP information
    private final String mcpServerId;              // MCP server identifier
    private final TransportType mcpTransportType;  // Transport type
    private final String mcpMessageId;             // Message tracking ID
    
    // Extended information
    private final Set<String> mcpServerCapabilities;     // Server capabilities
    private final List<McpServerInfo> availableServers;  // Available server list
    private final Map<String, HealthMetrics> transportHealth; // Transport health status
    private final Instant lastSuccessfulExecution;       // Last successful execution time
}
```

### 3.2 Exception Mapping Strategy

The system maps low-level MCP exceptions to unified tool exceptions:

- `McpConnectionException` → `ToolTransportException`
- `McpTimeoutException` → `ToolTimeoutException`
- `McpResourceNotFoundException` → `ToolResourceException`
- `McpProtocolException` → `ToolProtocolException`
- `McpAuthException` → `ToolSecurityException`

This mapping ensures consistent exception handling across both local and distributed tool executions while preserving MCP-specific diagnostic information for troubleshooting and monitoring purposes.

The enhanced context provides rich information for intelligent error handling decisions, including server health status, available alternatives, and historical execution patterns.

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
dliubarskyi: @karolzdebel 

> To me it feels like you would want to reuse the ToolErrorHandler for both cases, except with different exceptions: ToolExecutionException, ToolArgumentParsingException, etc.
> 
> Although technically the error isn't the tool's fault, as a client I perceive the error to be closer to the tool than the LLM calling the tool, if that makes sense. ToolErrorHandler sounds abstract enough that it could encapsulate the pre, during, and post. I'm not strongly opinionated on this though, probably both approaches could work.

Sounds reasonable, I will try to implement `ToolArgumentParsingException`, thanks for the suggestion!

> What if the JSON is malformed? i.e. it doesn't fit the object structure.

This is serialization from a Java `Object` (whatever the return value of the `@Tool`-annotated method) into the JSON `String`, so there should be no issues

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
dliubarskyi: Thanks for the suggestion, I will check at the very end which of theese can be added

dliubarskyi:
@misselvexu good idea regarding more rich exception types, although I am not sure this complexity is required at this point. We can always add more specific exception types without breaking changes in the future. I would start simple, with just 2 exceptions: tool execution and parsing arguments

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
dliubarskyi: BTW, the reason why I was not sure about whether this handler should also handle errors with tool arguments: we already have a [tool hallucination strategy](https://docs.langchain4j.dev/tutorials/tools#tools-hallucination-strategy), which is a separate handler. It feels that hallucinated arguments are closer to it than to the new `ToolErrorHandler`.

But perhaps `ToolErrorHandler` can handle everything, including hallucinated tool names...

dliubarskyi: @karolzdebel FYI I have implemented the `ToolArgumentParsingException`

karolzdebel: > BTW, the reason why I was not sure about whether this handler should also handle errors with tool arguments: we already have a [tool hallucination strategy](https://docs.langchain4j.dev/tutorials/tools#tools-hallucination-strategy)

Ahh ok that makes sense. I suppose it may make sense to deprecate the hallucination strategy in favor of the error handler approach in case there's conflicting behavior there (Let's say you have both an error handler and a hallucination strategy).

karolzdebel (commented):
Draft overall looks good to me as far as the client facing portion goes.

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
karolzdebel: The reason I say that is, if I implemented a `ToolErrorHandler` I would be surprised to find that some of the tool related errors still aren't being handled and that I also need this other hallucination strategy component. I suspect most people would fall into this trap.

dliubarskyi: cc @mariofusco WDYT?

mariofusco: During (and before) a tool invocation there are many things that can go potentially wrong and the `hallucinatedToolNameStrategy`, as its name clarifies, only covers one of them. It's correct to give users the possibility plug their own strategies to handle the other kinds of errors. I see 2 alternatives to implement this:

1. We define a single `ToolErrorHandler` interface with a method for any of the possible type of errors and let users to implement it in order to provide their own strategies for all those possible errors.
2. We define a set of functional interfaces letting users to plug a different strategy for each kind of error they want to handle and of which the `hallucinatedToolNameStrategy` is just one (and the first) of them.

I'm strongly in favor of this second alternative for a number of reasons:

- It will be a natural extension of the existing `hallucinatedToolNameStrategy` and will allow us to add more strategies for different errors without breaking the backward compatibility.
- It is nicer for users to have an API with multiple functional interfaces instead of a single interface requiring to implement multiple methods, because in the first case they can simply implement their handlers with lambdas.
- It also allow users to have finer grained control implementing and providing strategies only for the kind of errors for which they want a custom behavior, leaving the default implementation for others.
- It ensures a better backward and forward compatibility: if at some point you want to add one more handler you can easily do so without breaking the others, while if you have to add a new method on the existing `ToolErrorHandler` interface you either will break existing implementation or will be obliged to make that method `default` with an empty, and very likely meaningless, implementation.

Am I making sense?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
```diff
@@ -21,24 +20,24 @@ static String extractResult(JsonNode result) {
...
             if (resultNode.has("content")) {
                 String content = extractSuccessfulResult((ArrayNode) resultNode.get("content"));
-                boolean isError = false;
-                if (resultNode.has("isError")) {
-                    isError = resultNode.get("isError").asBoolean();
-                }
-                if (isError) {
-                    content = String.format(EXECUTION_ERROR_MESSAGE + ". The tool returned: %s", content);
+                if (isError(resultNode)) {
+                    throw new ToolExecutionException(content);
```
dliubarskyi, line 24: @jmartisk I am trying to make `DefaultToolExecutor` and `McpToolExecutor` to behave consistently (in this PR):
- when it cannot parse tool arguments - throw `ToolArgumentParsingException`
- when it fails to execute the tool - throw `ToolExecutionException`.

Because of this, I need to adapt the behaviour of `ToolExecutionHelper` and `DefaultMcpClient`, do you see any problems with that? It is a breaking change, but since MCP module is still in beta, I guess it should not be a big deal?

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpToolsTestBase.java
```diff
@@ -86,23 +85,25 @@ public void executeToolWithWrongArgumentType() {
                 .name("echoString")
                 .arguments("{\"input\": 1}") // wrong argument type
                 .build();
-        String toolExecutionResultString = executor.execute(toolExecutionRequest, null);
-        assertThat(toolExecutionResultString)
-                .isEqualTo("There was an error executing the tool. Message: Internal error. Code: -32603");
+        assertThatThrownBy(() -> executor.execute(toolExecutionRequest, null))
+                .isExactlyInstanceOf(ToolExecutionException.class) // TODO should be ToolArgumentParsingException
+                .hasMessage("Internal error")
+                .hasFieldOrPropertyWithValue("errorCode", -32603); // TODO should be -32602
```
dliubarskyi, line 91: Hi @mkouba, this seems to be an issue on the MCP server side? As far as I understand, it should return `-32602` error code in case of issues with arguments (wrong argument type in this case)

Sources:
- https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling
- https://www.mcpevals.io/blog/mcp-error-codes

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
karolzdebel: I think a problem with the second option is that it's difficult to understand as a client whether I'm handling **all** of the errors, or if there's some obscure strategy that exists that I may be missing or that may get introduced in a later version that I'll forget to integrate. There should be a place where the client can go to get peace of mind about their error handling being complete, without errors inadvertently leaking to the LLM.

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpToolsTestBase.java
mkouba: Hi @dliubarskyi, I'm completely out of context ;-). Could you be more specific? I mean, what's the error on the server, what's deployed to the MCP server, etc.

dliubarskyi: MCP server with tools (does not matter which/how many), sending a tool request with a wrong argument type causes error `-32603`, but it should be `-32602`.

You can check few lines above, `echoString` tool with arguments `"{\"input\": 1}"`

mkouba: I see. Yes, we currently only report `-32602` if a tool with the given name does not exist or when a required argument is missing. I agree that it would make more sense to also report `-32602` if the type does not match. However, keep in mind that the MCP spec is not exactly clear about what error codes should be returned when; and `-32602 - Invalid params` basically means that anything in the JSON-RPC method `params` is wrong, i.e. in case of a tool call like:

```json
{
  "jsonrpc": "2.0",
  "id": 2,
  "method": "tools/call",
  "params": {
    "name": "toolName",
    "arguments": {
      "foo": "bar"
    }
  }
}

```
it could be the tool name or tool arguments.

I've created https://github.com/quarkiverse/quarkus-mcp-server/issues/371 to track this issue.

dliubarskyi: @mkouba thanks! I understand that spec does not specify this, but it would be great to differentiate between issues using different error codes. For example, users of LC4j might want to implement different strategies for these 2 situations:
- wrong arguments -> ask LLM to fix the issue and try again
- tool failed -> do something else

mkouba: > @mkouba thanks! I understand that spec does not specify this, but it would be great to differentiate between issues using different error codes. For example, users of LC4j might want to implement different strategies for these 2 situations:
> 
>     * wrong arguments -> ask LLM to fix the issue and try again
> 
>     * tool failed -> do something else

Actually, when a tool fails then you should not receive a _protocol error_ at all but a tool result with `isError: true`; see also https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling.

dliubarskyi: Do you mean that if I get `{"jsonrpc":"2.0","id":2,"error":{"code":-32603,"message":"Internal error"}}`, then it is definitely not "tool execution error"?

mkouba: > Do you mean that if I get `{"jsonrpc":"2.0","id":2,"error":{"code":-32603,"message":"Internal error"}}`, then it is definitely not "tool execution error"?

Well, this could be anything in the server implementation but it should not be a business logic error, i.e. when a tool is actually called but something goes wrong on the business level (not the infra level). But like I said the MCP spec is not very clear :shrug:.

In the context of Quarkus MCP server - we have a dedicated `ToolCallException` and `@WrapBusinessError` interceptor binding to produce a failed `ToolResponse` (instead of producing protocol error like above).

dliubarskyi: Out of curiosity, why 

```java
    @Tool(description = "Throws a business error")
    public String error() throws Exception {
        throw new RuntimeException("business error");
    }
```

leads to protocol error and not to business error by default (without `@WrapBusinessError`)?

In this case "business error" message is lost and I am getting "Internal error"

mkouba: > In this case "business error" message is lost and I am getting "Internal error"

It's not lost but logged on the server. A JAX-RS resource doesn't report the error message either by default :shrug:.

Also it was the default behavior before we introduced `@WrapBusinessError`.

You can also use `ToolCallException` instead of `@WrapBusinessError`.
 
```java
@Tool(description = "Throws a business error")
public String error() throws Exception {
        throw new ToolCallException("business error");
}
```

dliubarskyi: > It's not lost but logged on the server.

But it is lost from the client point of view. Clients might not even have access to the server. In both protocol and tool execution errors there is a space for the error message, so I think that getting an actual exception message there will be more benefitial than static "Internal error".

I am just assuming that in many cases users will not use `@WrapBusinessError` (as it requires an extra step and reading documentation), so potentially valuable error message will not be propagated from the MCP server to MCP the client and further (e.g., to the LLM or to the end user).

dliubarskyi: > You can also use ToolCallException instead of @WrapBusinessError.

This will also require an extra step. Also exceptions can be thrown from a deeper layers of user code or even from code that user does not control.

mkouba: > so potentially valuable error message will not be propagated from the MCP server to MCP the client and further (e.g., to the LLM or to the end user).

Well, it's quite questionable whether `NullPointerException`, `ArrayIndexOutOfBoundsException` and similar exceptions provide a valuable error message for clients. I'd argue that these exceptions are valuable for server authors but not for clients. That's why I prefer _"explicit handling of business errors"_.

> This will also require an extra step.

Yes, and it's intentional.

dliubarskyi: > Well, it's quite questionable whether NullPointerException, ArrayIndexOutOfBoundsException and similar exceptions provide a valuable error message for clients.

Those exceptions are mostly bugs and should be handled by server developers, yes. But code that is being called from the `@Tool` method can throw many kinds of exceptions, including business ones.

mkouba: > > Well, it's quite questionable whether NullPointerException, ArrayIndexOutOfBoundsException and similar exceptions provide a valuable error message for clients.
> 
> Those exceptions are mostly bugs and should be handled by server developers, yes. But code that is being called from the `@Tool` method can throw many kinds of exceptions, including business ones.

Exactly, that's why I believe a `@Tool` author should handle them explicitly...

dliubarskyi: @mkouba but why so? Sorry, but I am genuinely curious

mkouba: > @mkouba but why so? Sorry, but I am genuinely curious

No problem. But I already tried to explain - not all exceptions thrown within a `@Tool` method bring value to clients. Quite the opposite, I believe that most of the exceptions are of no use for clients (it's mostly an implementation detail). Hence, it's up to the server developer to decide which info would be useful for clients.

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
dliubarskyi: I agree with @mariofusco, there are more benefits in second approach, thus I have implemented it.

> I think a problem with the second option is that it's difficult to understand as a client whether I'm handling all of the errors, or if there's some obscure strategy that exists that I may be missing or that may get introduced in a later version that I'll forget to integrate. There should be a place where the client can go to get peace of mind about their error handling being complete, without errors inadvertently leaking to the LLM.

I have introduced Javadoc links between all 3 tool error handlers and will properly document them in one place.
I also don't anticipate more tool error handlers introduced in the future, we already cover all types of issues related to tools with those 3.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -233,7 +235,10 @@ public String executeTool(ToolExecutionRequest executionRequest) {
         } catch (TimeoutException timeout) {
             transport.executeOperationWithoutResponse(new McpCancellationNotification(operationId, "Timeout"));
             return ToolExecutionHelper.extractResult(RESULT_TIMEOUT);
-        } catch (ExecutionException | InterruptedException e) {
+        } catch (ExecutionException e) {
+            throw new ToolExecutionException(e.getCause());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
```
mariofusco, line 241: This looks harmless, but I'm not sure if it is necessary, isn't a thread already interrupted when the `InterruptedException` is thrown?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
```diff
@@ -21,24 +20,24 @@ static String extractResult(JsonNode result) {
...
             if (resultNode.has("content")) {
                 String content = extractSuccessfulResult((ArrayNode) resultNode.get("content"));
-                boolean isError = false;
-                if (resultNode.has("isError")) {
-                    isError = resultNode.get("isError").asBoolean();
-                }
-                if (isError) {
-                    content = String.format(EXECUTION_ERROR_MESSAGE + ". The tool returned: %s", content);
+                if (isError(resultNode)) {
+                    throw new ToolExecutionException(new RuntimeException(content));
```
mariofusco: This double exception looks wasteful, can't we simply add one more constructor to the `ToolExecutionException` taking the `content` as argument?

langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
```diff
@@ -78,10 +97,40 @@ public String execute(ToolExecutionRequest toolExecutionRequest, Object memoryId
...
+            if (wrapToolArgumentsExceptions) {
+                throw new ToolArgumentsException(getActualCause(e));
+            } else {
+                throw e;
+            }
+        }
+    }
+
+    private static Throwable getActualCause(Exception e) {
+        if (e.getClass() == RuntimeException.class && e.getCause() != null) {
```
mariofusco: Why not using `instaceof` here? Do we want to unwrap the cause only if the exception is **exactly** a `RuntimeException`?

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolErrorContext.java
```diff
@@ -0,0 +1,74 @@
...
+                "toolExecutionRequest=" + toolExecutionRequest +
+                ", memoryId=" + memoryId +
+                '}';
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
```
mariofusco, line 55: Very personal opinion: using the builder pattern seems a bit overkilling here.

mariofusco (dismissed):
I added just a few very minor and optional comments, but overall this pull request looks ok for me.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
dliubarskyi: AFAIK when interruption is not fully handled, the best practice is to restore the interruption flag. But I am not 100% sure, dou you think I should remove it?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
dliubarskyi: `ToolExecutionException` will be unwrapped and error handler will see only internal `RuntimeException`. Otherwise it will be inconsistent, some exceptions will be actual exceptions, but in this case it will be a `ToolExecutionException` which is used only as an API between `ToolExecutor` and AI Service/error handler...

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: AFAIK, catching the `InterruptedException` clears the interrupt flag, so you have to re-interrupt to keep the flag, this is the generally accepted good practice

langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
dliubarskyi: Yes, this is to handle the case when checked exceptions are wrapped into `RuntimeException` for the sake of re-throwing them. In this case `RuntimeException` is not informative, we care about it's internals. If using `instanceof`, other runtime exceptions will be unwrapped and will potentially lose some valuable info

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: The flag tells the application that it's supposed to stop, so some other piece of code may check that flag on the same thread later (to find out that it's supposed to exit), that is why you should keep it set.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolErrorContext.java
dliubarskyi: Agreed, but this is for future-proofing, I already anticipate at least one more field to be added here in the near future. At some point builder will be needed and one will need to create a new ctor and deprecate the old one, so I am trying to avoid it

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
mariofusco: Ok, I see. At least you could still add another constructor to the `ToolExecutionException` and create the necessary `RuntimeException` only there, thus hiding this detail here and in all other places where you're doing the same. This will help also if in future we will decide that we don't need that internal `RuntimeException` anymore and we want to propagate the exception in a different way (like propagating the `ToolExecutionException` itself if it has no cause), but this is a very minor improvement, so feel free to ignore it.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
mariofusco: @dliubarskyi @jmartisk Thanks for clarification.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
dliubarskyi: @mariofusco good point, let me do the change

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
dliubarskyi: @jmartisk polite ping :)

jmartisk: Oh, sorry, I'd completely missed this! Feel free to prod me on Slack also whenever this happens.

> when it cannot parse tool arguments - throw ToolArgumentParsingException 

OK

> when it fails to execute the tool - throw ToolExecutionException. 

So, don't you need to also introduce a `propagateToolExecutionExceptions` field for MCP clients too? IIUC with your current code, it would always throw an exception instead of returning something to the chat model?

jmartisk: One might want to return an error string instead (in some cases, it may be expected that an MCP tool throws an error)

jmartisk: Or even better, we could allow both `DefaultToolExecutor` and `McpToolExecutor` to have a custom `ToolErrorHandler` class that dynamically decides how the error would be handled. But maybe that would be unnecessarily complex...

dliubarskyi: > Or even better, we could allow both DefaultToolExecutor and McpToolExecutor to have a custom ToolErrorHandler class that dynamically decides how the error would be handled. But maybe that would be unnecessarily complex...

That is what this PR is trying to achieve :)

The idea is that `ToolExecutor` throws an exception when error happens (instead of returning a `String`, which hides the fact that error happened), and depending on the error type, either `ToolArgumentsErrorHandler` or `ToolExecutionErrorHandler` will handle it.

> So, don't you need to also introduce a propagateToolExecutionExceptions field for MCP clients too? IIUC with your current code, it would always throw an exception instead of returning something to the chat model?

I had to introduce `propagateToolExecutionExceptions` in `DefaultToolExecutor` because that module is stable and we cannot break the default behaviour. It adds more complexity, but keeps behaviour unchanged for those who interact with `DefaultToolExecutor` directly.

`McpToolExecutor`, on the other hand, is still beta/experimental, so we could break it IMHO. Also, I am not sure if many users really call `DefaultMcpClient.executeTool()` or `McpToolExecutor.execute()` directly. Only those who call them directly will see the breaking changes, those who configure `McpToolProvider` on AI Service will not see any changes.

dliubarskyi: If you look into `ToolService`, there are `DEFAULT_TOOL_ARGUMENTS_ERROR_HANDLER` and `DEFAULT_TOOL_EXECUTION_ERROR_HANDLER` which keep the default behaviour for tool execution errors (both default and MCP) untouched, only the "API contract" between AI Service and `ToolExecutor` is changing: instead of returning an error as a `String` from `ToolExecutor`, an exception is thrown. This way we can detect in AI Service that error happened and call the appropriate error handler. Or, potentially in the future, do something else.

jmartisk: Ah ok, thanks for explaining (I was too lazy to read the whole code), it makes sense to me, thanks!

PR #3541, opened 2025-08-13
Use a pool of virtual threads as internal default executor

mariofusco:
## Issue

Closes #

## Change

The vast majority of long running operations, eventually running in parallel, are I/O tasks and for this reason it is very beneficial to execute them on virtual threads.

This pull request introduces a default executor, that should be used by all parallel operations performed by langchain4j. When possible, i.e. when running on Java 21+, this executor is a pool of virtual threads, otherwise normal cached platform threads are used.

This pull request also contains a minor change in the agentic parallel workflow, making it to use a more generic `Executor` instead of a more restrictive `ExecutorService`.

## General checklist

### DIFF
--- a/docs/docs/tutorials/agents.md
+++ b/docs/docs/tutorials/agents.md
@@ -289,7 +289,7 @@ MovieExpert movieExpert = AgenticServices
 EveningPlannerAgent eveningPlannerAgent = AgenticServices
         .parallelBuilder(EveningPlannerAgent.class)
         .subAgents(foodExpert, movieExpert)
-        .executorService(Executors.newFixedThreadPool(2))
+        .executor(Executors.newFixedThreadPool(2))
         .outputName("plans")
         .output(agenticScope -> {
             List<String> movies = agenticScope.readState("movies", List.of());
@@ -309,7 +309,7 @@ EveningPlannerAgent eveningPlannerAgent = AgenticServices
 List<EveningPlan> plans = eveningPlannerAgent.plan("romantic");
 ```
 
-Here the `output` function of the `AgenticScope` defined in the `EveningPlannerAgent` allows to assemble the outputs of the two subagents, creating a list of `EveningPlan` objects that combine a movie and a meal matching the given mood. The `output` method, even if especially relevant for parallel workflows, can be actually used in any workflow pattern to define how to combine the outputs of the subagents into a single result, instead of simply returning a value from the `AgenticScope`. The `executorService` method also allows to optionally provide an `ExecutorService` that will be used to execute the subagents in parallel, otherwise an internal cached thread pool will be used by default.
+Here the `output` function of the `AgenticScope` defined in the `EveningPlannerAgent` allows to assemble the outputs of the two subagents, creating a list of `EveningPlan` objects that combine a movie and a meal matching the given mood. The `output` method, even if especially relevant for parallel workflows, can be actually used in any workflow pattern to define how to combine the outputs of the subagents into a single result, instead of simply returning a value from the `AgenticScope`. The `executor` method also allows to optionally provide an `Executor` that will be used to execute the subagents in parallel, otherwise an internal cached thread pool will be used by default.
 
 ### Conditional workflow
 
@@ -473,8 +473,8 @@ public interface EveningPlannerAgent {
     })
     List<EveningPlan> plan(@V("mood") String mood);
 
-    @ExecutorService
-    static ExecutorService executor() {
+    @ParallelExecutor
+    static Executor executor() {
         return Executors.newFixedThreadPool(2);
     }
 
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
@@ -1,11 +1,12 @@
 package dev.langchain4j.internal;
 
 import dev.langchain4j.Internal;
-import java.lang.reflect.Method;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+
+import static dev.langchain4j.internal.VirtualThreadUtils.createVirtualThreadExecutor;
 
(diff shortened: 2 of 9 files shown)


docs/docs/tutorials/agents.md
```diff
@@ -473,8 +473,8 @@ public interface EveningPlannerAgent {
     })
     List<EveningPlan> plan(@V("mood") String mood);
 
-    @ExecutorService
-    static ExecutorService executor() {
+    @ParallelExecutor
```
mariofusco, line 476: I renamed the annotation as `ParallelExecutor` because it is used specifically for the parallel workflow and in this name I can avoid the name clash with the java `Executor` class.

langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
```diff
@@ -0,0 +1,60 @@
...
+
+                return (ExecutorService) newVirtualThreadPerTaskExecutorMethod.invoke(null);
+            } catch (Exception e) {
+                // Fallback in case of unexpected error
+                return createPlatformThreadExecutorService();
+            }
+        }
+
+        private static ExecutorService createPlatformThreadExecutorService() {
+            return Executors.newCachedThreadPool();
```
dliubarskyi: Could you please use config below instead? In this case the application will not hang for up to 60 seconds before shutting down. It was a source of complaints from users in the past

```java
return new ThreadPoolExecutor(
                0, Integer.MAX_VALUE,
                1, SECONDS,
                new SynchronousQueue<>()
        );
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -101,19 +99,7 @@ public void executeToolsConcurrently(Executor executor) {
     }
 
     private static Executor defaultExecutor() {
-        return DefaultExecutorHolder.INSTANCE;
```
dliubarskyi, line 104: Thank you for updating this as well! 🙏

langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
mariofusco: I did so. My only doubt is on that `Integer.MAX_VALUE`, shouldn't we set a more reasonable upper bound?

mariofusco:
@dliubarskyi Just for curiosity, is there any reason why you labeled this as `breaking change`? Is it only for the change in the parallel workflow method (that shouldn't be a problem since we flagged the whole module as experimental) or do you think that there could be other impacts that I'm missing?

dliubarskyi:
> @dliubarskyi Just for curiosity, is there any reason why you labeled this as breaking change? Is it only for the change in the parallel workflow method (that shouldn't be a problem since we flagged the whole module as experimental) or do you think that there could be other impacts that I'm missing?

Yes, it is experimental, but it still contains breaking changes (method name was changed, annotation was renamed, etc.). I have added a label for the future reference (e.g., for release notes to list all breaking changes)

langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
```diff
@@ -0,0 +1,60 @@
+package dev.langchain4j.internal;
+
+import dev.langchain4j.Internal;
+import java.lang.reflect.Method;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+@Internal
+public class DefaultExecutorProvider {
```
dliubarskyi, line 12: There is a `VirtualThreadUtils`, perhaps we can reuse something from there?

langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
dliubarskyi: @mariofusco if you look inside the `Executors.newCachedThreadPool()`, it also has `Integer.MAX_VALUE`. I think it is fine, considering that this is a default setting, user can always override it.

langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
mariofusco: I totally missed it. Now I'm reusing what's provided by that class, please review again.

dliubarskyi (approved):
@mariofusco thank you!

xuebin1111:
1) Not sure how langchain4j supports **Java Virtual Thread and Structured Concurrency**.
2) Seems **VirtualThreadUtils** is related.  Not sure why did not use the below standard code to create virtual threads but used refection Method.
**ExecutorService executor = Executors.newVirtualThreadPerTaskExecutor();**
3) If **DefaultExecutorProvider** is using Virtual Thread Executor then the whole high-throughput AI application (built on langchain4j) will be executing the concurrent tasks on virtual threads ?

Please help me in understanding this part, thanks!

mariofusco:
> 1. Not sure how langchain4j supports **Java Virtual Thread and Structured Concurrency**.
> 
>     2. Seems **VirtualThreadUtils** is related.  Not sure why did not use the below standard code to create virtual threads but used refection Method.
>        **ExecutorService executor = Executors.newVirtualThreadPerTaskExecutor();**
> 
>     3. If **DefaultExecutorProvider** is using Virtual Thread Executor then the whole high-throughput AI application (built on langchain4j) will be executing the concurrent tasks on virtual threads ?
> 
> 
> Please help me in understanding this part, thanks!

The use of reflection is due by the fact that we are still supporting Java 17 that doesn't have virtual threads. For this reason we cannot directly call the virtual thread API that is available only since Java 21. In case you're Java 21+ the default executor can make use of virtual thread, while on older Java versions it can use old native thread without breaking compatibility.

xuebin1111:
Thanks for your information!   
The Virtual Threads had been greatly improved in JDK24/25,  and will continue to improve Pinging and Structured Concurrency in JDK26 (March 2026).  Hence, the suggestion is to make langchain4j **support JDK26 directly** in the next year with the new code.

PR #1196, opened 2024-05-29
[FEATURE] Implement embedding removal methods for Weaviate embedding

kiratkumar47:
## Issue
fix #1171 

## Change
Implemented EmbeddingStore removal method (remove(String id), removeAll(Collection<String> ids)) and removeAll()) in WeaviateEmbeddingStore.

## General checklist
- [X] There are no breaking changes
- [ ] I have added unit and integration tests for my change
- [ ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)

PR #3555, opened 2025-08-18
add: implementation of removeAll(Filter) for WeaviateEmbeddingStore

### LINKED ISSUE #3330: [FEATURE] Weaviate: implementation of remove(String id) and removeAll(Filter filter)
In the WeaviateEmbeddingStore there is no overwrites of remove(String id) and removeAll(Filter filter).

I'd like an implementation of the missing methods.

I found these issues, but it seems like nobody completed it
- https://github.com/langchain4j/langchain4j/issues/1171
- https://github.com/langchain4j/langchain4j/pull/1196

AminAK11:
## Issue
Closes #3330

## Change
Added missing implementation of removeAll(Filter) for the WeaviateEmbeddingStore

## General checklist
- [X] I have added unit and/or integration tests for my change (already existing)
- [X] The tests cover both positive and negative cases (already existing)
- [] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j


AminAK11:
Apparently there is some format violations. When i run `mvn spotless:apply` i get no issues and no changes: 
image
Any suggestions?

dliubarskyi (commented):
@AminAK11 thank you, but both `EmbeddingStoreWithRemovalIT.should_remove_all_by_filter` and `should_remove_all` tests are failing, please check

dliubarskyi:
@AminAK11 please ignore spotless for now

AminAK11:
> @AminAK11 thank you, but both `EmbeddingStoreWithRemovalIT.should_remove_all_by_filter` and `should_remove_all` tests are failing, please check

Hi @dliubarskyi. Sorry for the last response. I have been looking at it for a while without any luck. I could use a helping hand.

dliubarskyi:
Perhaps @Jaland could help? 🙏

Jaland:
Have not touched this code in quite a while, but let me take a quick look

Jaland:
Hey played with it a little bit yesterday but just don't have a ton of time right now (and was having trouble getting my debugger up and running on my IDE).


Obviously there is definatly something wrong with this:

```
        ensureNotNull(filter, "filter");
        client.batch().objectsBatchDeleter()
                .withClassName(objectClass)
                .withWhere(WeaviateMetadataFilterMapper.map(filter))
                .run();
```

Cause it is just not filtering anything on the broken test. My AI guessed that this was an async function with eventual consistency and the issue could be temp fixed by just adding a sleep after but I don't think that is true and the sleep did not seem to fix the test.

May try to play with this more today if I get some time but just wanted to let y'all know it may be a bit before I do get time to get back around to this.

dliubarskyi:
@Jaland thanks a lot for the insights! 🙏

AminAK11:
@Jaland thank you for taking a look! Your insight that the objectBatchDeleter is async aligns with my understanding. However, i don't believe we need to handle this, notice that the unit-test uses awaitUntilAsserted which should handle this problem.

Here is some of my observations:
In the test, after `embeddingStore().add(...)` i can see in my debugger that the metadata is actually not even set. After further inspection this happens in the buildObject method of WeaviateEmbeddingStore, specefically the check: `if (metadataMap.containsKey(metadataKey))` is what ignores it. To make a temporary fix to this i added:
`.metadataKeys(List.of("id", "type"))` in LocalWeaviateEmbeddingStoreRemovalIT where the embeddingStore is created. 
Note "type" is the metadataKey we use in the testcase. 

Even with this fix, the test is not passing. This i think is because of the filtering in `WeaviateMetadataFilterMapper`. Specefically i think the problem is with the "path" field in

`WhereFilter.builder()
                .path(filter.key())
                .operator(Operator.Equal)
                .valueText(filter.comparisonValue().toString())
                .build();`

I found someone who had the same problem: https://forum.weaviate.io/t/wherefilter-not-working-in-deleting-records/2623
but i still couldn't get it to work.

AminAK11:
And it looks like this is not the first time someone have tried to implement it, they also had the exact same problems: 
https://github.com/langchain4j/langchain4j/pull/1196

AminAK11:
Hi @dliubarskyi and @Jaland 👋

Just a gentle reminder about this issue — I really hope it won’t get forgotten. I know it’s not a trivial one, but this feature would be very valuable for me (and likely others too). Thanks a lot for the effort!

dliubarskyi:
@AminAK11 could you please clarify what is the current state? Is something on LC4j side blocking you? Unfortunately I do not have a capacity to investigate in-depth right now

AminAK11:
@dliubarskyi i explained in an earlier comment exactly what the current state is, and exactly where in the code the problem is. I have tried to solve it but now i am asking for help. I'm not asking you to "investigate in-depth right now", i'm just asking for this issue to not be forgotten, and for someone to help me at some point.

AminAK11:
@dliubarskyi i finally got it to work. I found a workaround not needing to use the nested .path() in the objectsBatchDeleter. Now i just use Filter to get all ids that pass the filter, and use our existing removeAll(Collection<String> ids) to remove them.

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -180,9 +183,45 @@ public void removeAll(Collection<String> ids) {
...
+        if (!allIds.isEmpty()) {
+            removeAll(allIds);
+        }
+    }
+
+    @Override
+    public void removeAll(Filter filter) {
+        ensureNotNull(filter, "filter");
+
+        List<String> allIds = client.data().objectsGetter()
```
dliubarskyi, line 203: Shouldn't this be implemented with `objectsBatchDeleter` as shown [here](https://docs.weaviate.io/weaviate/manage-objects/delete#delete-multiple-objects)?

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
```diff
@@ -180,9 +183,45 @@ public void removeAll(Collection<String> ids) {
 
     @Override
     public void removeAll() {
-        client.batch().objectsBatchDeleter()
```
dliubarskyi, line 183: While it works, it does not seem to be optimal, shouldn't we [delete the collection](https://docs.weaviate.io/weaviate/manage-objects/delete#delete-all-objects) and create an empty one instead?

dliubarskyi (commented):
@AminAK11 thank you!

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
dliubarskyi: Reading all IDs and then filtering them on the client side seems to be suboptimal

AminAK11: You are totally right. The problem with the objectsBatchDeleter however, is that it needs a path of a property to filter by. In our case this would be subproperties of the metadata property, and it seems that the objectsBatchDeleter does not support subproperty lookups very well. I.e adding a path to metadata/property does not work (at least that is what i conclude after lots of trail and error). 

This is what i have described in previous comments, so that is why i solved it in this "suboptimal" way

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
AminAK11: Agree, but again, that is exactly how this was implemented before i changed it, but that didn't work (the testcases was excluded), right now the removeAll implementation in production doesn't even work...

langchain4j-weaviate/src/main/java/dev/langchain4j/store/embedding/weaviate/WeaviateEmbeddingStore.java
AminAK11: One way to fix it could be to remove the metadata property and store all the metadata toplevel, but i dont know?

dliubarskyi: Worth opening an issue on Weaviate github, perhaps they could fix this?

Please leave a comment in the code explaining the reason for this implementation and we can merge

PR #3557, opened 2025-08-18
Add tool execution request id for Gemini model

ChiaoGeek:
Unlike OpenAI, whose tool call responses already contain a unique [tool_call_id](https://github.com/langchain4j/langchain4j/blob/f60698986d04a52698a405a643e93526d515d53a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java#L335), the Gemini API does not provide such identifiers. Instead, according to the [[Gemini function calling documentation]](https://ai.google.dev/gemini-api/docs/function-calling?example=meeting#parallel_function_calling), the API relies on the index order of function calls to match requests with responses.

This PR updates the Gemini integration in langchain4j to embed a request ID based on the function call index whenever a ToolExecutionRequest is created. This ensures consistent and reliable mapping between Gemini function calls and their corresponding tool execution results.

One use case is that in front end, we can know when the tool execution is invoked and finished by leveraging two callback handlers (`beforeToolExecution` and `onToolExecuted`) in TokenStream  

Closes #

## Change

1. FunctionCallHelper
2. VertexAiGeminiStreamingChatModel:

## General checklist


ChiaoGeek:
Hey @dliubarskyi, could you please take a look ?

ChiaoGeek:
Hey @glaforge, could you please take a look ?

dliubarskyi (approved):
@ChiaoGeek thank you!

PR #3561, opened 2025-08-19
Allow configuring the logger for logging requests and responses

### LINKED ISSUE #3556: [FEATURE] Allow sending request/response logs to a log handler
Currently, ChatModel support configuring request/response logs using `ChatModel.logRequest()` and `ChatModel.logResponse()` those logs are then send to logback at DEBUG.

This is not very flexible as the log backend and the level are fixed.

Allow for extending the current login system for more flexibility.

One idea is to allow configuring a `LogHandler` so that instead of logging to logback, it could log to any JUL-compatible log handler.

Another solution is to allow configuring a `Consumer<String>` that would be used for logging each event, but it may be more complex.

**Additional context**
If there is an agreement, I can try to provide a PR for it.

loicmathieu:
## Issue
Part-of #3556

Support for Google Gemini, Mistral and OpenAI providers.

## Change
Allow passing a `org.slf4j.Logger` to log requests and responses instead of using the one provided by Langchain4J.
It allows user of the framework to use their own logger instead.

This is currently a quick prototype based on the discussions on #3556
The prototype only impact one provider, if it's agreed to pursue all models would be impacted.

The solution is the more basic one, if we want, we can do something more involved:
- Providing the logger using a `Producer<Logger>` instead of just a `Logger`.
- Using the `java.util.logging.LogHandler` to have a more agnostic logger
- Allow configuring also the level

The current prototype is only for openning discussions...

## General checklist


langchain4j-http-client/src/main/java/dev/langchain4j/http/client/log/LoggingHttpClient.java
```diff
@@ -17,16 +17,26 @@
...
 
-    private static final Logger log = LoggerFactory.getLogger(LoggingHttpClient.class);
+    private static final Logger DEFAULT_LOG = LoggerFactory.getLogger(LoggingHttpClient.class);
 
     private final HttpClient delegateHttpClient;
     private final boolean logRequests;
     private final boolean logResponses;
+    private final Logger log;
 
+    // FIXME will be removed if the change is accepted, to avoid too much changes only Gemini is impacted for now
```
dliubarskyi: What does this mean exactly? What will be removed? Please keep the original constructor in place, we don't want to do any breaking changes

dliubarskyi (dismissed):
@loicmathieu thanks!
Looks good in general.

> Providing the logger using a Producer<Logger> instead of just a Logger.

Why could this be required?

> Using the java.util.logging.LogHandler to have a more agnostic logger

SLF4j is a defacto standard and we use it everywhere, so I would stick to it.

> Allow configuring also the level

Any good reason for it?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/log/LoggingHttpClient.java
loicmathieu: If all existing usage of the previous constructor didn't exist anymore we could delete it.
Anyway, let's see when it's implemented everywhere.

loicmathieu:
> Providing the logger using a Producer instead of just a Logger.
> Why could this be required?

To avoid creating a logger if it is not used (of logResponses and logRequests are false). Logger can be costly to compute.
For what I need it, it's not important so I'm OK with the current implementation.

> Allow configuring also the level
> Any good reason for it?

Logging in debug is opinionated but I didn't have issue with that, it works for me.
One can always trick the logger in some way to redirect to another level if needed.
It's just that while we're at it, we may take the opportunity for it. Your call.

loicmathieu:
@dliubarskyi please give your advise on my last comments and then I'll do the change for all the providers!

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/log/LoggingHttpClient.java
dliubarskyi: `LoggingHttpClient` is a public class and can be used by anyone, not only by the code in this repository.

dliubarskyi:
> To avoid creating a logger if it is not used (of logResponses and logRequests are false). Logger can be costly to compute.
> For what I need it, it's not important so I'm OK with the current implementation.

Either option works for me, it's up to you.

> Logging in debug is opinionated but I didn't have issue with that, it works for me.
> One can always trick the logger in some way to redirect to another level if needed.
> It's just that while we're at it, we may take the opportunity for it. Your call.

I would not do it, unless there is a good reason. As you said, you can always redirect, so it is not a blocker.

dliubarskyi:
@loicmathieu 

> I'll do the change for all the providers!

Could you please split it in multiple PRs? Like 3 providers per PR?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/log/LoggingHttpClient.java
loicmathieu: OK, I'll remove the comment.

loicmathieu:
@dliubarskyi I'll stick to passing directly the logger, KISS ;)
I'll implement it it 3 providers by 3 providers then.

loicmathieu:
@dliubarskyi this PR include the first batch of implementation for Google Gemini, Mistral and OpenAI.
If you agree with it, I'll submit PRs for the other provides, 3 by 3, tomorrow.

dliubarskyi (approved):
@loicmathieu thanks a lot!

PR #3563, opened 2025-08-19
Fix a NPE that occurs if there is no parsed SSE

edeandrea:
In downstream frameworks (Quarkus) we marshal directly from String to `ChatCompletionResponse`, so this is throwing a `NullPointerException` when streaming over `String` types.


dliubarskyi (approved):
@edeandrea thank you!

PR #3567, opened 2025-08-20
fix(gemini): Fix Google Gemini API field naming to use camelCase

### LINKED ISSUE #3559: GeminiGenerationConfig uses snake_case naming but Google API expects camelCase, causing thinkingConfig to be ignored
**Description:**

**Describe the bug**
Parameters within `generationConfig`, such as `thinkingConfig`, are not being honored by the Google Gemini API when making requests.

The root cause is that the `GeminiThinkingConfig` class is annotated with `@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)`. This forces Jackson to serialize Java camelCase fields (e.g., `includeThoughts`, `thinkingBudget`) into snake_case JSON properties (e.g., `include_thoughts`, `thinking_budget`).

However, the official Google Gemini REST API documentation specifies that these parameters must be in camelCase. Because of this mismatch, the Google API silently ignores these important configuration settings, and features like "thinking" do not work.

**Log and Stack trace**
There is no exception or stack trace. The issue is a silent failure where the API does not apply the provided configuration because the field names in the JSON payload are incorrect.

**To Reproduce**
1.  Create an instance of `GoogleAiGeminiStreamingChatModel` or `GoogleAiGeminiChatModel` and configure it to use thinking.
    ```java
    StreamingChatModel model = GoogleAiGeminiStreamingChatModel.builder()
        .thinkingConfig(thinkingConfig)
        .build();
    
    model.chat();
    ```
2.  Set a breakpoint in the `dev.langchain4j.model.googleai.GeminiService` class, inside the `streamRequest` method (or `request` for non-streaming).
3.  Execute the code to trigger the model's `generate` method.
4.  When the breakpoint is hit, inspect the `jsonBody` variable after the line `String jsonBody = Json.toJson(requestBody);`.
5.  Observe that the `generationConfig` object in the JSON string contains snake_cased keys, like `"thinking_config"`, instead of the required camelCased keys.

**Expected behavior**
The JSON payload sent to the Google Gemini API should use camelCase for all fields within the `generationConfig` object to align with the official API documentation. This will ensure that features like `thinkingConfig`, `responseMimeType`, etc., are correctly recognized and applied by the model.

**Example of incorrect JSON being generated:**
```json
{
  "generationConfig": {
    "thinking_config": {
      "include_thoughts": true,
      "thinking_budget": -1
    }
  }
}
```

**Example of correct JSON that should be generated:**
```json
{
  "generationConfig": {
    "thinkingConfig": {
      "includeThoughts": true,
      "thinkingBudget": -1
    }
  }
}
```

- LangChain4j version: 1.3.0
- LLM(s) used: Google Gemini 1.5 Pro (and other Gemini models)
- Java version: [Please fill in your Java version]

**Additional context**
This issue likely affects all parameters within `GeminiGenerationConfig`. A simple fix would be to remove the `@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)` annotation from the `GeminiThinkingConfig` class.

[PLACEHOLDER FOR YOUR SCREENSHOTS: 

Image

Image

minguncle:
## Issue
Fixes #3559

## Change
Fixed Google Gemini API field naming compatibility issue by removing snake_case naming strategy and using camelCase field names as required by the official Google Gemini API specification.

**Changes made:**
- Removed `@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)` annotation from `GeminiGenerationConfig` and `GeminiThinkingConfig` classes
- Added explicit camelCase field names in `@JsonProperty` annotations for all existing fields in both classes

**Root cause:** 
The snake_case naming strategy was causing JSON fields to be serialized as `thinking_config`, `include_thoughts`, `thinking_budget`, `response_mime_type`, etc., but Google Gemini API expects `thinkingConfig`, `includeThoughts`, `thinkingBudget`, `responseMimeType`, etc. This caused important configuration parameters like thinking functionality to be silently ignored by the API.

**Impact:**
- Fixes thinking functionality not working due to incorrect field naming
- Ensures all generation config parameters are properly recognized by Google Gemini API
- No breaking changes to the public API - only internal JSON serialization is affected

## General checklist


dliubarskyi:
@minguncle TBH I am not sure there is any problem there. You can check `GoogleAiGeminiChatModelThinkingIT` and see that camel_case parameters are honored for sure.

cc @glaforge

minguncle:
> @minguncle TBH I am not sure there is any problem there. You can check `GoogleAiGeminiChatModelThinkingIT` and see that camel_case parameters are honored for sure.
> 
> cc @glaforge

Hi @dliubarskyi,

Thanks for your feedback. I've run the first test in GoogleAiGeminiChatModelThinkingIT as you suggested.

Based on the logs, the request body still uses snake_case for parameters like thinking_config, include_thoughts, and thinking_budget, rather than camelCase. However, the test itself passes successfully.

This makes me wonder if Google's API is indeed compatible with both snake_case and camelCase for these parameters.

My concern is that some API proxy services might strictly adhere to the official Google documentation, which specifies camelCase. In such scenarios, the current snake_case naming could potentially lead to issues.

Could this be a case where Google's API has some backward compatibility or flexibility, but it might not be universally true for all integrations or proxies?
image

dliubarskyi (approved):
@minguncle thank you!

PR #3569, opened 2025-08-20
AI Service: option to pass extra parameters into and between Tool/ToolProvider/RAG components

### LINKED ISSUE #1122: [FEATURE] Option to pass extra parameters to RAG/Tool/XyzProvider via AI service

### LINKED ISSUE #3545: Revisit `ToolExecutor` API
`ToolExecutor` should allow for more reach API:
- be able to accept additional context
- accept already parsed `arguments` (e.g., `Map<String, Object>`)? Both `DefaultToolExecutor` and `McpToolExecutor` have to parse arguments
- be able to return other types than `String` (e.g., image: for Anthropic)
- be able to indicate that returned value is an error? (for Anthropic)
- indicate which types of exceptions it can throw?

dliubarskyi:
## Issue
Closes #1122
Closes #3545

## Change
- Introduced `InvocationContext` class. It represents the context of a single AI Service invocation. A new instance is created each time an AI Service method is invoked, and it exists until the end of the AI Service invocation, potentially spanning multiple calls to the `ChatModel`.
- Introduced `InvocationParameters` class. It represents arbitrary parameters available during a single AI Service invocation. `InvocationParameters` can be specified when invoking the AI Service, as well as used to pass data between AI Service components (e.g., between tools, RAG components, etc.) in scope of a single AI Service invocation.
- Extended `ToolExecutor` API to be able to accept `InvocationContext` with `InvocationParameters` as well as return extra data in addition to the textual tool execution result.

See the updated documentation for more info: [1](https://github.com/langchain4j/langchain4j/pull/3569/files#diff-9ca302b5805896cdeaec9345077d7d34ca9d203a9a62e017eb70c6445e869ae0), [2](https://github.com/langchain4j/langchain4j/pull/3569/files#diff-9024e7c5e272c39c17a09b8f6240dee04df878e956ce1589d3fb651c510d451d)

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
```diff
@@ -0,0 +1,110 @@
...
+ * {@code InvocationParameters} can also be accessed within other AI Service components, such as:
+ * <pre>
+ * - ToolProvider: inside the ToolProviderRequest
+ * - ToolArgumentsErrorHandler and ToolExecutionErrorHandler: inside the ToolErrorContext
+ * - RAG components: inside the {@link Query} -> {@link Metadata}
+ * </pre>
+ * <p>
+ * Parameters are stored in a mutable, thread safe {@link Map}.
+ *
+ * @since 1.5.0
```
mariofusco:
```suggestion
 * @since 1.6.0
```

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.Objects;
+import dev.langchain4j.model.chat.ChatModel;
+
+/**
+ * Represents the context of a single AI Service invocation.
+ * A new instance is created each time an AI Service method is invoked,
+ * and it exists until the end of the AI Service invocation,
+ * potentially spanning multiple calls to the {@link ChatModel}.
+ *
+ * @since 1.5.0
```
mariofusco:
```suggestion
 * @since 1.6.0
```

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class InvocationParameters {
```
mariofusco, line 50: If I understand correctly these parameters are only intended to be consumed by tools and simply passed around and ignored by other components. If so wouldn't it be clearer to name them `ToolParameters`?

langchain4j-core/src/main/java/dev/langchain4j/rag/query/Metadata.java
```diff
@@ -50,31 +58,75 @@ public List<ChatMessage> chatMemory() {
         return chatMemory;
     }
 
+    /**
+     * @since 1.5.0
```
mariofusco:
```suggestion
     * @since 1.6.0
```

langchain4j-core/src/main/java/dev/langchain4j/rag/query/Metadata.java
```diff
@@ -50,31 +58,75 @@ public List<ChatMessage> chatMemory() {
...
+     */
+    public InvocationContext invocationContext() {
+        return invocationContext;
+    }
+
+    /**
+     * @since 1.5.0
```
mariofusco:
```suggestion
     * @since 1.6.0
```

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
dliubarskyi: Thanks!

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
dliubarskyi: No, currently they can be used not only by tools, but also by RAG components. In the future by all other components configured on AI Service, like memory provider, system message provider, etc.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -65,23 +68,41 @@ class DefaultAiServices<T> extends AiServices<T> {
...
             V v = parameter.getAnnotation(V.class);
             dev.langchain4j.service.UserMessage userMessage =
                     parameter.getAnnotation(dev.langchain4j.service.UserMessage.class);
             MemoryId memoryId = parameter.getAnnotation(MemoryId.class);
             UserName userName = parameter.getAnnotation(UserName.class);
-            if (v == null && userMessage == null && memoryId == null && userName == null) {
+
+            boolean isInvocationParameters = parameter.getType() == InvocationParameters.class;
+            if (isInvocationParameters) {
+                invocationParametersCount++;
```
mariofusco: Nitpick: you don't need a counter but only a `boolean`.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -65,23 +68,41 @@ class DefaultAiServices<T> extends AiServices<T> {
...
+                if (invocationParametersCount > 1) {
+                    throw illegalConfiguration("There can be at most one parameter of type %s",
+                            InvocationParameters.class.getName());
+                }
```
mariofusco, line 90: Nitpick: wouldn't it be better to do a `continue` here instead of checking the boolean immediately after?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -65,23 +68,41 @@ class DefaultAiServices<T> extends AiServices<T> {
...
+
         for (Parameter parameter : parameters) {
             V v = parameter.getAnnotation(V.class);
             dev.langchain4j.service.UserMessage userMessage =
                     parameter.getAnnotation(dev.langchain4j.service.UserMessage.class);
             MemoryId memoryId = parameter.getAnnotation(MemoryId.class);
             UserName userName = parameter.getAnnotation(UserName.class);
-            if (v == null && userMessage == null && memoryId == null && userName == null) {
+
+            boolean isInvocationParameters = parameter.getType() == InvocationParameters.class;
```
mariofusco: Could `InvocationParameters` be extended? If shouldn't we check with an `instaceof` here?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: not sure I understand, could you please elaborate?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: Oh right, good catch, I should test it

langchain4j/src/main/java/dev/langchain4j/service/AiServiceTokenStreamParameters.java
```diff
@@ -234,14 +235,8 @@ public Builder context(AiServiceContext context) {
             return this;
         }
 
-        /**
-         * Sets the memory ID.
-         *
-         * @param memoryId the memory ID
-         * @return this builder
-         */
-        public Builder memoryId(Object memoryId) {
```
mariofusco, line 243: This is a more general question: do we want to remove the possibility to set the `memoryId` alone here, instead of deprecating the method as you did somewhere else? And even about deprecation, I'm not sure of it: why should we deprecate or even remove a method that is convenient for users who don't need to deal with a complete `InvocationContext` (provided that we are always able to deal internally with it in a completely unambiguous way)?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: This is a very minor thing, so feel free to ignore it. Anyway, this is in a for loop and you want to skip the remaining part of the loop if `isInvocationParameters` is `true`, so why not simply add a `continue` here at the end of this block, instead of adding one more check in the (already verbose) if condition immediately later?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -351,6 +384,21 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                         }
                     }
 
+                    private Optional<InvocationParameters> findInvocationParameters(Object[] arguments, Parameter[] parameters) {
+                        if (arguments == null) {
+                            return Optional.empty();
+                        }
+                        for (int i = 0; i < parameters.length; i++) {
+                            Parameter parameter = parameters[i];
+                            if (parameter.getType().isAssignableFrom(InvocationParameters.class)) {
```
mariofusco: I always get confused about this, but I believe that this should be the opposite:

`if (InvocationParameters.class.isAssignableFrom(parameter.getType()))`

or even better maybe just use

`if (parameter.getType() instanceof InvocationParameters)`

that imo is more readable and less error-prone.

dliubarskyi: Thanks, I've just noticed it too. Changed and tested

langchain4j/src/main/java/dev/langchain4j/service/AiServiceTokenStreamParameters.java
dliubarskyi: > This is a more general question: do we want to remove the possibility to set the memoryId alone here, instead of deprecating the method as you did somewhere else?

`AiServiceTokenStreamParameters` is an internal class, so users should not manipulate it directly.

> And even about deprecation, I'm not sure of it: why should we deprecate or even remove a method that is convenient for users who don't need to deal with a complete InvocationContext (provided that we are always able to deal internally with it in a completely unambiguous way)?

`ToolErrorContext` was introduced just in the recent release, but since it is public API, I have deprecated the method instead of removing it. But it should not be used by users directly as well. If we keep both `memoryId` and `invocationContext` (which also contains `memoryId`), it becomes confusing...

mariofusco (dismissed):

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: ok, makes sense, thanks

PR #3570, opened 2025-08-20
Feature/mcp sse add header

### LINKED ISSUE #3157: [FEATURE] Authentication support for remote MCP servers

Remote MCP servers typically require authentication.
Currently there is no way (as far as I can tell) to be able to support these servers in langchain4j without spinning up a local proxy (`npx mcp-remote`). (and intercepting the output to perform a logon dance).
Spinning up local proxies is not sustainable due to
1) the requirement for OATH dance needing to know local ports etc for callbacks
2) additionally spinning up processes is a poor workaround

Storing credentials locally is not a solution for a MUA/Server based app where multiple users would have their own chat and credentials.

As an application developer I would like to be able to get a callback when authentication fails.
If the authentication is OUATH I would like to be able to direct the user to perform an OAUTH dance to then be able to provide the credentials.
~if it is a BASIC I would like to be able to have the callback provide  be able to (in my app) prompt the user for credentials) and return them.~

As Authentication may not be quick, I would like long timeouts.
Until the app is authenticated I would like the langchain4j to be either able to block on the authentication, or to be able to skip the server until it is authenticated (depending on configuration).

I am hacking around this using `StdioMcpTransport` `npx mcp-remote` and docker, where I poll the docker logs and then request the user start the authentication dance.

**Additional context**

https://blog.cloudflare.com/building-ai-agents-with-mcp-authn-authz-and-durable-objects/
https://modelcontextprotocol.io/specification/draft/basic/authorization#2-authorization-flow

This is considered distinct from #2944 which allows the injection of headers.   This issue is about being able to be informed that auth is needed so the application can handle it.

Y7000-save-me:
## Issue
Closes #3157

## Change

The addition of custom header capabilities to the HttpMcpTransport and StreamableHttpMcpTransport classes allows users to pass in their own headers, which is a common use case, such as including authentication information for MCP requests. The code has been self-tested and does not impact the existing logic.

## General checklist

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/HttpMcpTransport.java
@@ -36,11 +36,14 @@ public class HttpMcpTransport implements McpTransport {
 
     private static final Logger log = LoggerFactory.getLogger(HttpMcpTransport.class);
     private final String sseUrl;
+    private final Map<String, String> headers;
     private final OkHttpClient client;
     private final boolean logResponses;
     private final boolean logRequests;
     private EventSource mcpSseEventListener;
     private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private static final String CONTENT_TYPE = "Content-Type";
+    private static final String CONTENT_TYPE_JSON = "application/json";
     private volatile Runnable onFailure;
 
     // this is obtained from the server after initializing the SSE channel
@@ -60,6 +63,7 @@ public class HttpMcpTransport implements McpTransport {
         }
         this.logResponses = builder.logResponses;
         sseUrl = ensureNotNull(builder.sseUrl, "Missing SSE endpoint URL");
+        headers = getOrDefault(builder.headers, Map.of());
         client = httpClientBuilder.build();
     }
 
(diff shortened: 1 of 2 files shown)


jmartisk:
Thanks!
Could we please use some neutral APIs in both transports (avoid using `okhttp3.Headers` but rather a Map) so that we make it easier to later migrate the transports to use the neutral HTTP client SPI without introducing breaking changes...

Also, some headers may need to be dynamic (they will change over time, like authentication tokens). So, I think the object that the transports will receive will be a `Supplier<Map<String, String>>`, and the supplier would be called before every call to the MCP server to obtain the current headers, does that make sense?

anna-geller:
Hey @Y7000-save-me and @jmartisk 👋 just checking in — any update on the merge progress? We’d really appreciate having this PR merged soon.

dliubarskyi:
@Y7000-save-me I will merge it as it is now, but please consider adding a `Supplier<Map<String, String>>` in a separate PR

dliubarskyi (approved):
@Y7000-save-me thank you!

dliubarskyi:
@Y7000-save-me adding a documentation is also much appreciated 🙏

dliubarskyi:
@Y7000-save-me FYI I have renamed `headers` into `customHeaders` to keep it consistent with other locations where we allow setting custom headers

Y7000-save-me:
No need to thank me, I've learned from your code. If there's an opportunity, I'd like to continue contributing PRs and make a contribution to the community.

@dliubarskyi

PR #3578, opened 2025-08-21
Allow to evict or completely disable the tool list cache in DefaultMcpClient

### LINKED ISSUE #3573: [FEATURE] Options to evict DefaultMcpClient tool cache
Per issue #2263, there was tool caching with a dynamic eviction implemented for `DefaultMcpClient`. The eviction is triggered when the MCP server sends `notifications/tools/list_changed` notification. 

I have an MCP server that doesn't send this.

I make modifications to the tools provided by the MCP server, restart it, and it now hosts the new and modified tools. `DefaultMcpClient`, however, keeps the old tools still cached. It doesn't drop them because the notification never arrives.

For starters, the simplest possible option would be to allow manual cache eviction by calling a public method on `McpClient` interface or `DefaultMcpClient`.

Something more sophisticated could check if the HTTP connection was lost to the MCP server and evict the cache on this basis. A time-based cache eviction could also be an option. 

Alternatively, Langchain's internal cache could be disableable altogether.

Force internal cache eviction via reflection. Pardon my Kotlin:
```
fun listMcpToolsUncached(mcpClient: McpClient): ToolProviderResult {
    // Force Langchain's internal caching off.
    if (mcpClient is DefaultMcpClient) {
        val toolListOutOfDateField = DefaultMcpClient::class.java.getDeclaredField("toolListOutOfDate")
        toolListOutOfDateField.isAccessible = true
        val toolListOutOfDate = toolListOutOfDateField.get(mcpClient) as AtomicBoolean
        toolListOutOfDate.set(true)
    }

    return McpToolProvider
        .builder()
        .mcpClients(listOf(mcpClient))
        .build()
        .provideTools(null)
}
```

Zalewa:
## Issue
Closes #3573 

## Change

- Add `DefaultMcpClient.evictToolListCache()`
- Add `DefaultMcpClient.Builder.cacheToolList()`
- Add tests for all of that, plus tests for `DefaultMcpClient.listTools()`

## General checklist

### DIFF
--- a/docs/docs/tutorials/mcp.md
+++ b/docs/docs/tutorials/mcp.md
@@ -411,3 +411,19 @@ ToolExecutionRequest request = ToolExecutionRequest.builder()
                 .build();
 String toolResult = mcpClient.executeTool(request);
 ```
+
+## Notes about Tool Caching
+
+`DefaultMcpClient` maintains an internal cache of MCP tools. Once retrieved, the
+tool list won't be requested again from the MCP server unless the server sends a
+notification that the list has been updated. You can manually clear this cache
+by calling `DefaultMcpClient.evictToolListCache()`. If you prefer to disable
+caching entirely, configure the client as follows:
+
+```java
+McpClient mcpClient = new DefaultMcpClient.Builder()
+    .key("MyMCPClient")
+    .transport(transport)
+    .cacheToolList(false)
+    .build();
+```
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -12,6 +12,8 @@ import com.fasterxml.jackson.databind.node.JsonNodeFactory;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.exception.ToolArgumentsException;
+import dev.langchain4j.exception.ToolExecutionException;
 import dev.langchain4j.mcp.client.logging.DefaultMcpLogMessageHandler;
 import dev.langchain4j.mcp.client.logging.McpLogMessageHandler;
 import dev.langchain4j.mcp.client.protocol.McpCallToolRequest;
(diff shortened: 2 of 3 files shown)


dliubarskyi:
cc @jmartisk could you please review?

jmartisk (dismissed):

jmartisk:
LGTM, I guess this is ready, no?

jmartisk:
Maybe also just add some little note somewhere in the docs and I guess we're good to go.

Zalewa:
> add some little note somewhere in the docs

Dunno if I chose the right place, but: done & done.
(Spotless fails through no fault of my own)

docs/docs/tutorials/mcp.md
```diff
@@ -76,6 +76,20 @@ McpClient mcpClient = new DefaultMcpClient.Builder()
 Note that the client key is optional, but it is recommended to set it, especially
 if there are multiple MCP clients, and it is necessary to disambiguate among them.
 
+`DefaultMcpClient` maintains an internal cache of MCP tools. Once retrieved, the
```
jmartisk: I think the text itself is good, but let's maybe move it into a separate section at the end of the document, WDYT?  This part is a quick tutorial about how to get it running, so it doesn't belong here that well.

I would just create a section,  for example, `## Notes about tool caching`, at the end, and just move this into it.

Zalewa: Done as requested. New commit pushed to the PR.

jmartisk:
LGTM! Could you squash into a single commit and rebase? Thanks

Zalewa:
Sure. Keeping the Git history clean. I'm inclined to keep the 1st commit with basic unit tests for `listTools()` as separate because that's not directly related to issue #3573.

jmartisk (dismissed):
LGTM

jmartisk:
@Zalewa please mark as ready
@dliubarskyi I'd like to see the CI passing, but it doesn't seem to be working :( do you know what's wrong?

jmartisk:
Ah right, I guess it's because of the failing spotless. @Zalewa could you run `mvn spotless:apply` before re-committing?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -211,6 +213,19 @@ public List<ToolSpecification> listTools() {
         }
     }
 
+    private boolean isToolListRefreshNeeded() {
+        return !Boolean.TRUE.equals(cacheToolList) || toolListOutOfDate.get();
```
dliubarskyi:
```suggestion
        return Boolean.FALSE.equals(cacheToolList) || toolListOutOfDate.get();
```

Zalewa:
> Ah right, I guess it's because of the failing spotless. @Zalewa could you run `mvn spotless:apply` before re-committing?

Yes, it fails spotless because there's been changes in the meanwhile that have been pushed directly into the repo that cause it to fail. `mvn spotless:apply` also doesn't run very well; there's a failed build due to the spotless plugin being unknown. I've been running `make format` so far, which activates "spotless" profile in Maven. Running it at the current main's head modifies quite a few files, mostly in the "agentic" module.

dliubarskyi:
@Zalewa thank you!

PR #3583, opened 2025-08-25
added neo4j follow-up query docs

### LINKED ISSUE #269: [FEATURE] Neo4j follow-up query handling for EmbeddingStore searches
## Is your feature request related to a problem? Please describe.

I think the way langchain4j's `EmbeddingStore` is designed,
it's not intended for executing follow-ups unless we run a manual query e.g. leveraging the id properties of the nodes.

Here is an example of how to do it:

```
Neo4jEmbeddingStore embeddingStore = Neo4jEmbeddingStore.builder()
                .withBasicAuth(neo4jContainer.getBoltUrl(), USERNAME, ADMIN_PASSWORD)
                .dimension(384)
                .label("CustomLabel")
                .build();

// --- add embeddings.... 

// search embeddings
final List<EmbeddingMatch<TextSegment>> results = embeddingStore.search(<insert the search request with dev.langchain4j.store.embedding.EmbeddingSearchRequest instance>).matches();

// -- follow-up query part --
// retrieve the ids to execute the follow-up
List<String> nodeIds = results.stream().map(dev.langchain4j.store.embedding.EmbeddingMatch:embeddingId).toList();

// change `CustomLabel` with `Document` if label(<label>) is not set in the Neo4jEmbeddingStore builder
String cypher = """
        MATCH (d:CustomLabel)
        WHERE d.id IN $ids
        // here the follow-up query, for example
        WITH (d)-[:CONNECTED_TO]->(o:OtherLabel ) RETURN o.id
    """;

// run the follow-up query
Map<String, Object> params = Map.of("ids", nodeIds);
final List<Record> list = session.run(cypher, params).list();
// -- end of follow-up query part --
```

## Describe the solution you'd like

Understand if it is possible to do something like this in langchain4j more easily (maybe through some configuration or similar).

 Or at least improve the documentation with the above solution.

## Describe alternatives you've considered

I think the most feasible solution is to extend `EmbeddingSearchRequest` with a followUp configuration (or something more generic)

## Additional context

Example of use case: https://github.com/JMHReif/langchain4j-quarkus-goodreads

In that repo, using the Goodreads data set and embeddings are on the Review nodes, and If we want users to search for general themes or topics they like in books, the similarity search returns only Reviews (which isn’t helpful), so we have to retrieve the related books connected to those similar reviews.

vga91:
## Issue

Closes https://github.com/langchain4j/langchain4j-community/issues/269

## Change

Added documentation about how to execute a Neo4j follow-up query after an embedding search.

## General checklist

### DIFF
--- a/docs/docs/integrations/embedding-stores/neo4j.md
+++ b/docs/docs/integrations/embedding-stores/neo4j.md
@@ -302,6 +302,30 @@ final EmbeddingSearchResult<TextSegment> searchWithoutFilter = embeddingStore.se
 final List<EmbeddingMatch<TextSegment>> matchesWithoutFilter = searchWithoutFilter.matches();
 ```
 
+To execute a follow-up query for reading or writing data retrieved by the embedding search, we can leverage the nodes' `embeddingId`s.
+For example:
+```java
+// ... Neo4jEmbeddingStore instance creation ...
+// ... add embeddings.... 
+
+final List<EmbeddingMatch<TextSegment>> results = embeddingStore.search(/*dev.langchain4j.store.embedding.EmbeddingSearchRequest instance*/)
+        .matches();
+
+// retrieve the ids to execute the follow-up
+List<String> nodeIds = results.stream().map(dev.langchain4j.store.embedding.EmbeddingMatch:embeddingId).toList();
+
+String cypher = """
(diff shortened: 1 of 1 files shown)


dliubarskyi (approved):
@vga91 thank you!

PR #3589, opened 2025-08-25
#3588: Handle IllegalArgumentException during URI.create

### LINKED ISSUE #3588: [BUG] Unhandled IllegalArgumentException on GoogleCustomWebSearchEngine
**Describe the bug**
When mapping links contained in `com.google.api.services.customsearch.v1` models coming from Google PSE (https://www.googleapis.com/customsearch/v1) responses, `URI.create` is executed in several places (result links, image links, thumbnail links) without encoding the string and without capturing `IllegalArgumentException`. As a consequence, result items could be missed, and the exception interrupts the entire processing.

We should:

1. Handle the links appropriately before executing URI.create. 
2. Catch the exception for the failing item and continue processing.

**Log and Stack trace**
```
Error during web search with dev.langchain4j.web.search.google.customsearch.GoogleCustomWebSearchEngine@5cc9655b at start index 1: Illegal character in query at index 72: https://www.golfargentino.com/products/?search=taylormade&price_regular={745000,1762500} 
```

**To Reproduce**
Any resulting query that might contain unsupported characters such as { or } ends up in the error logs. For example, I performed the search query driver golf taylormade qi35 specifically on the site https://golfargentino.com. Which resulted on: `https://www.golfargentino.com/products/?search=taylormade&price_regular={745000,1762500}`
This can also be reproduced with a simple mock in unit tests.

**Expected behavior**
String links should be encoded before calling URI.create, and in case of an IllegalArgumentException, we should continue processing the remaining items without interrupting the overall process.

- LangChain4j version: 1.3.0 & 1.3.0-beta9 for langchain4j-web-search-engine-google-custom
- LLM(s) used: gemini-1.5-flash
- Java version: 21
- Spring Boot version (if applicable): 3.4.3

msampietro:
## Issue

Fixes #3588

## Change
This PR fixes URI creation safety by replacing direct URI constructors with the createUriSafely() utility method to prevent crashes from malformed URLs returned by Google's API.

Changes Made:

Safe URI creation for image search results (imageLink, contextLink, thumbnailLink)
Safe URI creation for organic search results
Added static import for createUriSafely utility (introduced in separate class)
Prevents IllegalArgumentException and applied fallback when Google API returns invalid URL strings, making the search mapping more robust.

## General checklist

### DIFF
--- a/web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleCustomWebSearchUtils.java
+++ b/web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleCustomWebSearchUtils.java
@@ -1,0 +1,99 @@
+package dev.langchain4j.web.search.google.customsearch;
+
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.BitSet;
+
+/**
+ * Utility class for Google Custom Web Search operations.
+ */
+public final class GoogleCustomWebSearchUtils {
+
+	private GoogleCustomWebSearchUtils() {}
+
+	private static final BitSet ALLOWED = new BitSet(128);
+	private static final char[] HEX = "0123456789ABCDEF".toCharArray();
+
+	static {
+		// RFC 3986: unreserved
+		for (char c = 'a'; c <= 'z'; c++) ALLOWED.set(c);
+		for (char c = 'A'; c <= 'Z'; c++) ALLOWED.set(c);
+		for (char c = '0'; c <= '9'; c++) ALLOWED.set(c);
+		for (char c : new char[]{'-','.','_','~'}) ALLOWED.set(c);
+		// gen-delims + sub-delims (keep URL syntax intact)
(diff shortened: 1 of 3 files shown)


msampietro:
I believe failing checks are unrelated to the committed changes

web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleCustomWebSearchUtils.java
```diff
@@ -0,0 +1,96 @@
+package dev.langchain4j.web.search.google.customsearch;
+
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.BitSet;
+
+/**
+ * Utility class for Google Custom Web Search operations.
+ */
+public final class GoogleCustomWebSearchUtils {
```
dliubarskyi:
```suggestion
final class GoogleCustomWebSearchUtils {
```

msampietro: Thanks @dliubarskyi for the review and commit! Anything else required for an approval?

dliubarskyi (approved):
@msampietro thank you!

PR #3605, opened 2025-08-27
Update watsonx.ai SDK and implement tokenization methods

andreadimaio:
## Issue

Update the watsonx.ai SDK and implement the tokenization methods

## General checklist

### DIFF
--- a/docs/docs/integrations/language-models/watsonx.md
+++ b/docs/docs/integrations/language-models/watsonx.md
@@ -298,3 +298,4 @@ See more details [here](https://docs.quarkiverse.io/quarkus-langchain4j/dev/wats
 - [WatsonxToolsTest](https://github.com/langchain4j/langchain4j-examples/blob/main/watsonx-ai-examples/src/main/java/WatsonxToolsTest.java)
 - [WatsonxEmbeddingModelTest](https://github.com/langchain4j/langchain4j-examples/blob/main/watsonx-ai-examples/src/main/java/WatsonxEmbeddingModelTest.java)
 - [WatsonxScoringModelTest](https://github.com/langchain4j/langchain4j-examples/blob/main/watsonx-ai-examples/src/main/java/WatsonxScoringModelTest.java)
+- [WatsonxTokenCounterEstimatorTest](https://github.com/langchain4j/langchain4j-examples/blob/main/watsonx-ai-examples/src/main/java/WatsonxTokenCounterEstimatorTest.java)
--- a/langchain4j-watsonx/pom.xml
+++ b/langchain4j-watsonx/pom.xml
@@ -27,7 +27,7 @@
         <dependency>
             <groupId>com.ibm.watsonx</groupId>
             <artifactId>watsonx-ai</artifactId>
-            <version>0.6.1</version>
+            <version>0.7.0</version>
         </dependency>
 
         <!-- test dependencies -->
--- a/langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxBuilder.java
+++ b/langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxBuilder.java
@@ -2,7 +2,6 @@ package dev.langchain4j.model.watsonx;
 
 import com.ibm.watsonx.ai.core.auth.AuthenticationProvider;
 import java.net.URI;
-import java.net.http.HttpClient;
 
 @SuppressWarnings("unchecked")
 abstract class WatsonxBuilder<T extends WatsonxBuilder<T>> {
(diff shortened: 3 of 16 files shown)


dliubarskyi (approved):
@andreadimaio thank you!

andreadimaio:
@dliubarskyi I'm curious why the test automation expects the streaming part to run on a single thread. Is this to prevent concurrency issues?

https://github.com/langchain4j/langchain4j/blob/05b20492f3cb90d9ada6266f8f74758306af285b/langchain4j-core/src/test/java/dev/langchain4j/model/chat/common/AbstractBaseChatModelIT.java#L329-L341

dliubarskyi:
@andreadimaio well, I just wanted to make sure that it is not run on the current thread. Perhaps this

```java
assertThat(threads).hasSize(1); 
assertThat(threads.iterator().next()).isNotEqualTo(Thread.currentThread()); 
```
can be changed into 
```java
`assertThat(threads).doesNotContain(Thread.currentThread());`
```
to better reflect the constraint.

But in general, all integrations are running in the same thread, at least with the default HTTP client.

Why are you asking?

andreadimaio:
Because by default watsonx sdk uses an executor sized to the number of CPU cores for streaming.
With the current assertion `(hasSize(1))`, the tests fail unless we override it with a custom, single-threaded executor (see `WatsonxExecutorProvider`). However, if we change the check to `doesNotContain(Thread.currentThread())`, I can remove the custom SPI and rely on the sdk's default behaviour.

https://github.com/langchain4j/langchain4j/blob/dc233357acb2aec5b6e06d15ee0e0459b758b931/langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxExecutorProvider.java#L12-L30

WDYT?

andreadimaio:
I had a quick look at the langChain4j http client implementation, and I think I understand the difference with the watsonx sdk (which also relies on Java’s HttpClient). By default, if no executor is provided, the client uses `ForkJoinPool.commonPool()` as its executor. This pool can create additional (compensating) threads when needed, which is why the tests succeed (the number of threads remains at one).

On the other hand, if you configure a fixed thread pool, the set of threads observed in the test will easily exceed one, even if there are no real concurrency issues, and the assertion will fail.

dliubarskyi:
@andreadimaio do you mean that a single response can be handled by multiple threads?

> By default, if no executor is provided, the client uses ForkJoinPool.commonPool() as its executor.

You might want to double-check if it is a good idea to use common pool for IO-heavy operations

> On the other hand, if you configure a fixed thread pool, the set of threads observed in the test will easily exceed one, even if there are no real concurrency issues, and the assertion will fail.

I am not sure I understand the isse, `should_respect_maxOutputTokens_in_chat_request` does a single request and gets a single response, why it should use more than one thread?

andreadimaio:
> do you mean that a single response can be handled by multiple threads?

No, each chunk is handled by one thread. However, if you have a **fixed** pool of threads, the first chunk might be handled by **thread-1**, the second by **thread-2**, and so on, always sequentially, so there are no concurrency issues.

> You might want to double-check if it is a good idea to use common pool for IO-heavy operations

If no executor is passed to the Watsonx SDK, it uses an `Executors.newFixedThreadPool(Runtime.getRuntime().availableProcessors())`.

> I am not sure I understand the isse, should_respect_maxOutputTokens_in_chat_request does a single request and gets a single response, why it should use more than one thread?

As I mentioned above, if you use a pool of threads, the set of threads can contain more than one even for a single response, because different chunks of the response may be handled by different threads sequentially.

dliubarskyi:
> No, each chunk is handled by one thread. However, if you have a fixed pool of threads, the first chunk might be handled by thread-1, the second by thread-2, and so on, always sequentially, so there are no concurrency issues.

That sounds a bit unusual, is this the desired behaviour?

andreadimaio:
If you have a thread pool handling I/O operations, chunks of a single response might be processed by different threads depending on availability, or they could all be handled by the same thread if it’s free. Do you see it differently, or did you expect something else?

dliubarskyi:
@andreadimaio I was expecting all chunks to be handled by the same thread, at least this is how all other integrations work, not sure whether this difference is critical though. Might be unexpected from the user point of view as well.

andreadimaio:
> all chunks to be handled by the same thread

From my point of view, this is true if you use a single-threaded Executor, or even a cached thread pool Executor (in the latter case, the tests pass because the Executor ends up creating only one thread, although in real scenarios, more threads could be created).

If you create a fixed thread pool with more than one thread, the behavior can change, because the JVM may choose any available thread in the pool.

From my perspective, this is not a problem, as each chunk is still processed sequentially and there are no concurrency issues, but I’m open to discussion. :smile:

PR #3633, opened 2025-09-02
Use ToolExecutionException in web search engines

loicmathieu:
## Change
Use the new `ToolExecutionException` in the web search engine tools to be able to intercept them in the new tool execution exception handler on ai services.

## General checklist

### DIFF
--- a/web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleCustomSearchApiClient.java
+++ b/web-search-engines/langchain4j-web-search-engine-google-custom/src/main/java/dev/langchain4j/web/search/google/customsearch/GoogleCustomSearchApiClient.java
@@ -7,6 +7,7 @@ import com.google.api.client.json.gson.GsonFactory;
 import com.google.api.services.customsearch.v1.CustomSearchAPI;
 import com.google.api.services.customsearch.v1.CustomSearchAPIRequest;
 import com.google.api.services.customsearch.v1.model.Search;
+import dev.langchain4j.exception.ToolExecutionException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


dliubarskyi (commented):
@loicmathieu thank you, but I am not sure why are these changes required?

`DefaultToolExecutor` already wraps any exception thrown from `@Tool`-annotated method into `ToolExecutionException`.

Moreover, web search clients definitely should not be aware of tools, those are 2 separate features

loicmathieu:
OK, I may have misread the code.

I had a situation where an exception in a tool execution didn't lead to a ToolExecutionException in my own tool so I switch to directly thows the ToolExecutionException.

Then I asked myself whether or not Langchain4J provided tools have the same issue and I noticed that the MCP client directly throws `ToolExecutionException` but not the WebSearch client that why I proposed that.

> DefaultToolExecutor already wraps any exception thrown from @Tool-annotated method into ToolExecutionException.

Are you sure?  Reading it here https://github.com/langchain4j/langchain4j/blob/80826507e77c42005a7e39b8738ba80c15f04f09/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java#L129 I didn't see any wrapping but I may not understand it correctly.

Anyway, feel free to dismiss my PR.

loicmathieu:
OK, my bad, it will throw an InvocationTargetException in case of exception and this one is correctly unwrapped to the method's exception then wrap if needed insidea ToolExecutionException.

Closing it, sorry.

dliubarskyi:
> I had a situation where an exception in a tool execution didn't lead to a ToolExecutionException in my own tool so I switch to directly thows the ToolExecutionException.

How are your tools defined? `@Tool` or programmatic?

> the MCP client directly throws ToolExecutionException but not the WebSearch client that why I proposed that.

Only in the tool-handling part (`executeTool` method)

> Are you sure? Reading it here

Yes, here: https://github.com/langchain4j/langchain4j/blob/80826507e77c42005a7e39b8738ba80c15f04f09/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java#L109

loicmathieu:
@dliubarskyi thanks for the explaination, effectively the issue was in my own tool and when I checked if all tools provided by Langchain4J wrap inside a ToolExecution I misread the default tool executor (due to forgetting that InvocationTargetException wraps all reflective invocation exception) and thought it was missing...

PR #3643, opened 2025-09-04
Introduce asynchronous agents

mariofusco:
## Issue

Closes #

## Change

I started working on this change with the idea of making the `HumanInTheLoop` non-blocking, meaning that the invocation is made on another thread and the agentic system execution blocks only when the result provided by the `HumanInTheLoop` is really necessary because it is read as an input for another agent. Then I figured out that this non-blocking feature was quite useful, so I decided to extend it to all agents. In this way, for instance, it is possible to achieve a parallel execution without using the parallel workflow, but simply putting the non-blocking agents in sequence.

## General checklist

### DIFF
--- a/docs/docs/tutorials/agents.md
+++ b/docs/docs/tutorials/agents.md
@@ -417,6 +417,52 @@ ExpertRouterAgent expertRouterAgent = AgenticServices
 String response = expertRouterAgent.ask("I broke my leg what should I do");
 ```
 
+## Non-blocking agents
+
+By default, all agents invocations are performed in the same thread that invoked the root agent of the agentic system, and therefore they are blocking, meaning that the execution of the agentic system waits for the completion of each agent before proceeding to the next one. However, in many cases this is not necessary, and it could be useful to invoke an agent in a non-blocking way, allowing the execution of the agentic system to proceed without waiting for the completion of that agent.
+
+For this reason it is possible to flag an agent as non-blocking using the `nonBlocking` method of the agent builder. When doing so, the invocation of that agent is performed in a separate thread, and the execution of the agentic system will proceed without waiting for the completion of that agent. The result of the non-blocking agent will be available in the `AgenticScope` as soon as it is completed, and the `AgenticScope` will be blocked waiting for that result only when it is required as an input for a subsequent invocation of a different agent.
+
+For instance, since they are independent of each other, flagging the `FoodExpert` and `MovieExpert` agents, discussed in the parallel workflow section, as non-blocking, will make them to be executed at the same time even when used in a sequential workflow.
+
+```java
+FoodExpert foodExpert = AgenticServices
+        .agentBuilder(FoodExpert.class)
+        .chatModel(BASE_MODEL)
+        .nonBlocking(true)
+        .outputName("meals")
+        .build();
+
+MovieExpert movieExpert = AgenticServices
+        .agentBuilder(MovieExpert.class)
+        .chatModel(BASE_MODEL)
+        .nonBlocking(true)
+        .outputName("movies")
+        .build();
+
+EveningPlannerAgent eveningPlannerAgent = AgenticServices
+        .sequenceBuilder(EveningPlannerAgent.class)
+        .subAgents(foodExpert, movieExpert)
+        .executor(Executors.newFixedThreadPool(2))
+        .outputName("plans")
+        .output(agenticScope -> {
+            List<String> movies = agenticScope.readState("movies", List.of());
+            List<String> meals = agenticScope.readState("meals", List.of());
+
+            List<EveningPlan> moviesAndMeals = new ArrayList<>();
+            for (int i = 0; i < movies.size(); i++) {
+                if (i >= meals.size()) {
+                    break;
+                }
+                moviesAndMeals.add(new EveningPlan(movies.get(i), meals.get(i)));
+            }
+            return moviesAndMeals;
+        })
+        .build();
+
+List<EveningPlan> plans = eveningPlannerAgent.plan("romantic");
+```
+
 ## Error handling
 
 In a complex agentic system, many things can go wrong, such as an agent failing to produce a result, an external tool not being available, or an unexpected error occurring during the execution of an agent.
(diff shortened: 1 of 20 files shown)


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/Agent.java
```diff
@@ -42,4 +42,11 @@
      * @return name of the output variable.
      */
     String outputName() default "";
+
+    /**
+     * If true, the agent will be invoked in a non-blocking manner, allowing the workflow to continue without waiting for the agent's result.
+     *
+     * @return true if the agent should be invoked in a non-blocking manner, false otherwise.
+     */
+    boolean nonBlocking() default false;
```
dliubarskyi:
```suggestion
    boolean blocking() default true;
```

Just a slight simplification, WDYT? If it makes sense, please rename it everywhere

docs/docs/tutorials/agents.md
```diff
@@ -417,6 +417,52 @@ ExpertRouterAgent expertRouterAgent = AgenticServices
 String response = expertRouterAgent.ask("I broke my leg what should I do");
 ```
 
+## Non-blocking agents
```
dliubarskyi: Naming nitpick (not sure if it makes sense, please feel free to ignore), it seems that "asynchronous" is a more appropriate term? https://chatgpt.com/share/68bef431-4efc-800e-880a-57e579a78c16

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/Agent.java
mariofusco: Maybe I'm overthinking this, but in my opinion it really depends on the default behavior of this feature. To be honest if I had thought to this possibility when I implemented the first release of the agentic module, I would have made all agents non-blocking by default. I'm not doing this now to avoid changing their behavior and breaking backward compatibility in a quite implicit and subtle way, even though I'm still tempted to change this.

In essence I see 2 alternatives:

1. I leave it as it is
2. I rename `nonBlocking` with `blocking` (and change the code accordingly), but also make the non-blocking behavior as the default one.

WDYT?

docs/docs/tutorials/agents.md
mariofusco: Good point. I also had doubts between the 2 terms and I like chatgpt take on this.

So also related to the former comment, I think that I will rename `nonBlocking` to `asynch` (or do you prefer the longer `asynchronous`?) and leave the default behavior as it is. Is it ok?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/Agent.java
dliubarskyi: I was referring only to the name of the `boolean` flag, not the default behaviour. AFAIK it is a good practice to avoid unnecessary negations when naming things to avoid additional mental overhead, but it is not really critical.

As for the default behaviour: if you are sure that non-blocking is the better default, I would say that it is still a good time to change it, as module is quite young and still experimental

docs/docs/tutorials/agents.md
dliubarskyi: I was thinking about `async` (instead of `nonBlocking`) as well, I guess it is a good one

mariofusco:
@dliubarskyi I renamed `nonBlocking` with `async` everywhere. Thinking twice for now I'm leaving the current synchronous behavior, because I'm afraid that the async one could be confusing for users not expecting it, but I'm not strongly opinionated on this and I may re-evaluate this decision at a later stage. Please review again.

langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/A2AClientAgentInvoker.java
```diff
@@ -12,6 +12,7 @@ public class A2AClientAgentInvoker implements AgentInvoker {
 
     private final String[] inputNames;
     private final String outputName;
+    private final boolean nonBlocking;
```
dliubarskyi:
```suggestion
    private final boolean async;
```

langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/A2AClientAgentInvoker.java
```diff
@@ -20,6 +21,7 @@ public A2AClientAgentInvoker(A2AClientSpecification a2AClientInstance, Method me
         this.agentCard = a2AClientInstance.agentCard();
         this.inputNames = inputNames(a2AClientInstance);
         this.outputName = a2AClientInstance.outputName();
+        this.nonBlocking = a2AClientInstance.async();
```
dliubarskyi:
```suggestion
        this.async = a2AClientInstance.async();
```

langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/A2AClientAgentInvoker.java
```diff
@@ -43,6 +45,11 @@ public String outputName() {
         return outputName;
     }
 
+    @Override
+    public boolean async() {
+        return nonBlocking;
```
dliubarskyi:
```suggestion
        return async;
```

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentExecutor.java
```diff
@@ -3,35 +3,62 @@
...
-                case THROW_EXCEPTION -> throw e;
-                case RETRY -> internalExecute(agenticScope, invokedAgent);
-                case RETURN_RESULT -> recoveryResult.result();
-            };
+        return execute(agenticScope, agentInvoker.async());
+    }
+
+    public Object syncExecute(DefaultAgenticScope agenticScope) {
+        if (agentInvoker.async()) {
+            LOG.info("Executing '{}' agent in a blocking way even if declared as non-blocking", agentInvoker.name());
```
dliubarskyi:
```suggestion
            LOG.info("Executing '{}' agent in a sync way even if declared as async", agentInvoker.name());
```

mariofusco:
@dliubarskyi thanks for fixing outstanding `nonBlocking` namings (I forgot to check the A2A module), is it ok now?

dliubarskyi (approved):
@mariofusco thanks a lot!

PR #3661, opened 2025-09-09
#3628: Use received description for visited json object schema

### LINKED ISSUE #3628: [BUG] Description is reused for a tool parameter from a custom class
**Describe the bug**
[`JsonSchemaElementUtils`](https://github.com/langchain4j/langchain4j/blob/e900f3f718e2196ce9aa9e6082351cc9ee850590/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java#L38) is used to define the parameters of a tool (among other functionalities). However, the method [`jsonObjectOrReferenceSchemaFrom`](https://github.com/langchain4j/langchain4j/blob/e900f3f718e2196ce9aa9e6082351cc9ee850590/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java#L98) reuses the description of a `JsonObject` already defined/visited. Thus, this can cause issues when calling the tool with the right parameters.

**To Reproduce**
Create a tool with two parameters with the same POJO (e.g., Place) and observe how the description is reused. For example, in this case, both parameters (`origin` and `destination`) will have the description `origin place`.

```
  @Tool(
      name = "getDestinations",
      value = "Get a list of recommended destinations with flight prices and images")
  public ToolResponse<List<Destination>> getDestinations(
      @P(value = "origin place") Place origin,
      @P(value = "destination place") Place destination) {
}
```

**Expected behavior**
The description should be unique and updated in the reused [JsonSchemaElement](https://github.com/langchain4j/langchain4j/blob/e900f3f718e2196ce9aa9e6082351cc9ee850590/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java#L107)

- LangChain4j version: 1.4.0
- LLM(s) used: -
- Java version: 21
- Spring Boot version (if applicable): No

jordibisbal8:
## Issue
Fixes #3628 

## Change
In this PR, we aim to return the received description instead of using the one from the visited `JsonObjectSchema`. For this, we need to build a new `JsonObjectSchema` with all properties from the visited one, but with the received description.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
@@ -24,14 +24,7 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.math.BigDecimal;
 import java.math.BigInteger;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
 import java.util.stream.Collectors;
 
 @Internal
@@ -107,6 +100,19 @@ public class JsonSchemaElementUtils {
             if (jsonSchemaElement instanceof JsonReferenceSchema) {
                 visitedClassMetadata.recursionDetected = true;
             }
+            if (jsonSchemaElement instanceof JsonObjectSchema obj) {
+                if (Objects.equals(description, obj.description())) {
+                    return obj;
+                }
+                return JsonObjectSchema.builder()
+                        .description(description)
+                        .addProperties(obj.properties())
+                        .required(obj.required())
+                        .additionalProperties(obj.additionalProperties())
+                        .definitions(obj.definitions())
+                        .build();
+            }
+
             return jsonSchemaElement;
         }
 
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
@@ -104,13 +104,7 @@ public class JsonSchemaElementUtils {
                 if (Objects.equals(description, obj.description())) {
                     return obj;
                 }
-                return JsonObjectSchema.builder()
-                        .description(description)
-                        .addProperties(obj.properties())
-                        .required(obj.required())
-                        .additionalProperties(obj.additionalProperties())
-                        .definitions(obj.definitions())
-                        .build();
+                return obj.toBuilder().description(description).build();
             }
 
             return jsonSchemaElement;
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonObjectSchema.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonObjectSchema.java
@@ -4,10 +4,7 @@ import static dev.langchain4j.internal.Utils.copy;
 import static dev.langchain4j.internal.Utils.quoted;
 import static java.util.Arrays.asList;
 
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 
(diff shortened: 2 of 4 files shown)


langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
```diff
@@ -107,6 +100,19 @@ public static JsonSchemaElement jsonObjectOrReferenceSchemaFrom(
             if (jsonSchemaElement instanceof JsonReferenceSchema) {
                 visitedClassMetadata.recursionDetected = true;
             }
+            if (jsonSchemaElement instanceof JsonObjectSchema obj) {
+                if (Objects.equals(description, obj.description())) {
+                    return obj;
+                }
+                return JsonObjectSchema.builder()
```
dliubarskyi: nit: if/when new field(s) will be added to `JsonObjectSchema` in the future, there is a high probability that we will forget to update this place. Basically, here you want to make a copy of `JsonObjectSchema`, only overriding the description. What about adding a `JsonObjectSchema.toBuilder()` method that will return builder will all fields prefilled, then you will be able to do:
`return obj.toBuilder().description(description).build();`

WDYT?

langchain4j-core/src/test/java/dev/langchain4j/internal/JsonSchemaElementUtilsTest.java
```diff
@@ -136,6 +133,37 @@ void should_use_non_null_description_for_uuid() {
                         .build());
     }
 
+    @Test
+    void givenVisitedJsonObjectSchema_whenDescriptionIsDifferent_thenReturnsNewSchemaWithUpdatedDescription() {
```
dliubarskyi, line 137: nice tests, although I would also add an integration test that defines tool with `@Tool` and verifies that all pieces work well together

dliubarskyi (commented):
@jordibisbal8 good job, thank you!

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
jordibisbal8: Sounds good, similar to the `toBuilder()` from Lombok, although I have seen that we should not use it. Let me do the changes :)

langchain4j-core/src/test/java/dev/langchain4j/internal/JsonSchemaElementUtilsTest.java
jordibisbal8: Sure! Should I add it under[ this test](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/test/java/dev/langchain4j/agent/tool/ToolSpecificationsTest.java)? I have tested it locally, but an additional IT never hurts.

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
jordibisbal8: https://github.com/langchain4j/langchain4j/pull/3661/commits/1f9cfeb5f368097b3ac7ffe30ea6804868acc67e

langchain4j-core/src/test/java/dev/langchain4j/internal/JsonSchemaElementUtilsTest.java
jordibisbal8: https://github.com/langchain4j/langchain4j/pull/3661/commits/67c2dde9145a59504c75f8c0d2603f74d87b4f97

langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/json/JsonObjectSchema.java
```diff
@@ -53,6 +50,15 @@ public static Builder builder() {
         return new Builder();
     }
 
+    public Builder toBuilder() {
+        return builder()
+                .description(this.description)
+                .addProperties(this.properties)
+                .required(this.required != null ? new ArrayList<>(this.required) : null)
```
jordibisbal8, line 57: Need to make a defensive copy, not to use the original list

dliubarskyi (approved):
@jordibisbal8 thanks a lot, great job!

PR #3672, opened 2025-09-11
Add callback allowing to configure agents created through the declarative API

mariofusco:
## Issue

Closes #

## Change

Add an optional callback to tweak configuration of agents created through the declarative API (necessary for the Quarkus extension).

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java
@@ -231,7 +231,12 @@ public class AgenticServices {
         return A2AService.get().a2aBuilder(a2aServerUrl, agentServiceClass);
     }
 
-    public record DeclarativeAgentCreationContext(Class<?> agentServiceClass, AgentBuilder<?> agentBuilder) { }
+    public interface DeclarativeAgentCreationContext<T> {
+            Class<T> agentServiceClass();
+            AgentBuilder<T> agentBuilder();
+    }
+
+    public record DefaultDeclarativeAgentCreationContext<T>(Class<T> agentServiceClass, AgentBuilder<T> agentBuilder) implements DeclarativeAgentCreationContext<T> { }
 
     /**
      * Creates an instance of an agentic system defined through the declarative API.
@@ -569,7 +574,7 @@ public class AgenticServices {
                             agentBuilder.chatModel(chatModel);
                         });
 
-        agentConfigurator.accept(new DeclarativeAgentCreationContext(agentType, agentBuilder));
+        agentConfigurator.accept(new DefaultDeclarativeAgentCreationContext(agentType, agentBuilder));
     }
 
     private static <T> T invokeStatic(Method method, Object... args) {
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java
@@ -51,6 +51,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.concurrent.Executor;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


mariofusco:
@geoand Please give a look and let me know if this is enough for you.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/DeclarativeAgentIT.java
```diff
@@ -90,6 +90,37 @@ void declarative_sequence_tests() {
...
+    void declarative_sequence_with_agent_configuration_tests() {
+        StoryCreatorWithConfigurableStyleEditor storyCreator = AgenticServices.createAgenticSystem(StoryCreatorWithConfigurableStyleEditor.class, baseModel(), ctx -> {
+            if (ctx.agentServiceClass() == StyleEditor.class) {
+                ctx.agentBuilder().outputName("styledStory");
+            }
+        });
+
+        String story = storyCreator.write("dragons and wizards", "fantasy", "young adults");
+        assertThat(story).isNotBlank();
+        System.out.println(story);
```
dliubarskyi: nit: do we really need printing it out?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java
```diff
@@ -230,19 +231,15 @@ public static <T> A2AClientBuilder<T> a2aBuilder(String a2aServerUrl, Class<T> a
         return A2AService.get().a2aBuilder(a2aServerUrl, agentServiceClass);
     }
 
+    public record DeclarativeAgentCreationContext(Class<?> agentServiceClass, AgentBuilder<?> agentBuilder) { }
```
dliubarskyi:
```suggestion
    public record DeclarativeAgentCreationContext<T>(Class<T> agentServiceClass, AgentBuilder<T> agentBuilder) { }
```

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/DeclarativeAgentIT.java
```diff
@@ -90,6 +90,37 @@ void declarative_sequence_tests() {
...
+        })
+        String write(@V("topic") String topic, @V("style") String style, @V("audience") String audience);
+    }
+
+    @Test
+    void declarative_sequence_without_agent_configuration_tests() {
+        StoryCreatorWithConfigurableStyleEditor storyCreator = AgenticServices.createAgenticSystem(StoryCreatorWithConfigurableStyleEditor.class, baseModel());
+
+        String story = storyCreator.write("dragons and wizards", "fantasy", "young adults");
+        assertThat(story).isNull();
```
dliubarskyi, line 108: Is it OK to return `null`? If I understand correctly, this is a misconfiguration case? Perhaps we should detect this when building agent and throw an exception?

dliubarskyi (dismissed):
@mariofusco thank you!

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java
geoand: The ideal way to future-proof this would be to make it an interface, not a record. This is obviously a small nitpick

geoand:
I tested this and it does exactly what I needed it do. Thanks @mariofusco!

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/DeclarativeAgentIT.java
mariofusco: I removed all printings in this test class.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/DeclarativeAgentIT.java
mariofusco: It's correct that this is a misconfiguration, but it will be very hard, if not impossible, to catch it at build time, especially because the `outputName` could be provided in many ways, including the callback that I'm adding with this pull request. The agentic system it is returning a `null` here because it is reading a state of the `AgenticScope` that no former agent has set before. It is questionable if it should return a `null` or throw a runtime exception in this case, but if any change will be required on this I would leave out from this pull request anyway.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java
mariofusco: Done.

geoand: 🙏🏽

dliubarskyi (approved):
@mariofusco thank you!

PR #3678, opened 2025-09-12
Use a pool executor of daemon threads to avoid keeping the app alive

mariofusco:
## Issue

Closes #

## Change

Using a version of Java before 21, it cannot use virtual threads so it uses an executor made of platform ones. However since these threads haven't been flagged as daemon they keep the application alive even after the application using them has completed all tasks.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/DefaultExecutorProvider.java
@@ -22,7 +22,12 @@ public class DefaultExecutorProvider {
                 createVirtualThreadExecutor(Holder::createPlatformThreadExecutorService);
 
         private static ExecutorService createPlatformThreadExecutorService() {
-            return new ThreadPoolExecutor(0, Integer.MAX_VALUE, 1, TimeUnit.SECONDS, new SynchronousQueue<>());
+            return new ThreadPoolExecutor(0, Integer.MAX_VALUE, 1, TimeUnit.SECONDS, new SynchronousQueue<>(),
+                    runnable -> {
+                        Thread thread = new Thread(runnable);
+                        thread.setDaemon(true);
+                        return thread;
+                    });
         }
     }
 }


dliubarskyi (commented):
@mariofusco but these threads expire in 1 second, why is it required to make them daemon? AFAIU, it can lead to a situation when LC4j is invoked and VM exits because it executes in daemon thread.

mariofusco:
> @mariofusco but these threads expire in 1 second, why is it required to make them daemon?

That 1 second doesn't refer to threads expiration in general. The javadocs says:

> keepAliveTime – when the number of threads is greater than the core, this is the maximum time that excess idle threads will wait for new tasks before terminating.

so if the number of threads is not greater than the number of cores, no idle thread is expired, and they are kept alive.

@LizeRaes showed me the problem this morning: she is running on Java 17 and if using the parallel workflow for agents with the default executor her test case never terminates and hangs forever.

> AFAIU, it can lead to a situation when LC4j is invoked and VM exits because it executes in daemon thread.

Normally LC4J is not invoked in a daemon thread itself (and if the user explicitly do this for any reason then it means that he expects that invocation to be terminated even before all tasks have been completed), but the worker threads internally used by it (the ones created by the thread pool) have to be daemons in order to avoid keeping the application alive even when it is terminated.

In fact another thing to consider is that virtual threads are all daemons by default, so if there would be any problem in this sense we would already have figured out and experienced it using  Java >=21.

dliubarskyi:
> @mariofusco but these threads expire in 1 second, why is it required to make them daemon?
> 
>> That 1 second doesn't refer to threads expiration in general. The javadocs says:
> 
> keepAliveTime – when the number of threads is greater than the core, this is the maximum time that excess idle threads will wait for new tasks before terminating.
> 
>> so if the number of threads is not greater than the number of cores, no idle thread is expired, and they are kept alive.

Sorry, I am not following... You can run this:

```java
        System.out.println("before");
        Holder.createPlatformThreadExecutorService().execute(() -> System.out.println("execute"));
        System.out.println("after");
    }
```

And it shuts down after 1 second:
```
C:\Users\...\.jdks\temurin-21.0.4\bin\java.exe ... dev.langchain4j.internal.DefaultExecutorProvider

before
after
execute

Process finished with exit code 0
```

The core pool size is 0, so in my understanding all idle threads shut down after maximum 1 second. Or am I missing something?

> Normally LC4J is not invoked in a daemon thread itself (and if the user explicitly do this for any reason then it means that he expects that invocation to be terminated even before all tasks have been completed), but the worker threads internally used by it (the ones created by the thread pool) have to be daemons in order to avoid keeping the application alive even when it is terminated.

What I meant is if LC4j is invoked (from a normal thread) and continues the execution in the daemon thread, then the application will just shut down.

IMHO an agent running in a parallel workflow is not a daemon at all, it is not some background supporting process, it is the "meat" of the application

mariofusco:
@dliubarskyi Sorry, you are right, I misinterpreted the test case provided by @LizeRaes.

PR #3679, opened 2025-09-12
Backport AI Service observability from Quarkus LangChain4j

### LINKED ISSUE #3627: [FEATURE] Backport Auditing features from Quarkus
I'd like to backport the [auditing capabilities](https://docs.quarkiverse.io/quarkus-langchain4j/dev/observability.html#_auditing) that are currently available in the Quarkus LangChain4j implementation.

Obviously how to use the capability here in upstream LangChain4j will be different (i.e. NOT implemented as CDI observers), but at least introduce the events and a way for applications to listen for them.

I will take on this work in the next couple of weeks.

edeandrea:
## Issue

Closes #3627

## Change

Backport [auditing capabilities](https://docs.quarkiverse.io/quarkus-langchain4j/dev/observability.html#_auditing) from Quarkus LangChain4j.

## General checklist

- [X] I have manually run all the unit tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green


edeandrea:
@dliubarskyi / @geoand / @mariofusco ^^^^

edeandrea:
I merged in `main` this morning and now tests are broken :( I'll take a look this morning.

edeandrea:
@dliubarskyi / @mariofusco The CI for this PR was passing before I merged `main` in. The failures are occurring in a module I didn't even touch (`[INFO] LangChain4j :: Integration :: OpenAI 1.5.0-SNAPSHOT  FAILURE`):

```
[ERROR] Errors: 
[ERROR]   OpenAiTokenCountEstimatorTest.should_support_all_chat_model_names:129 » IllegalArgument Model 'gpt-5' is unknown to jtokkit
[ERROR]   OpenAiTokenCountEstimatorTest.should_support_all_chat_model_names:129 » IllegalArgument Model 'gpt-5-mini' is unknown to jtokkit
[ERROR]   OpenAiTokenCountEstimatorTest.should_support_all_chat_model_names:129 » IllegalArgument Model 'gpt-5-nano' is unknown to jtokkit
```

I don't think that is at all related to anything I've done in this PR.

dliubarskyi:
@edeandrea that issue has been fixed, sorry for the inconvenience

docs/docs/tutorials/auditing.md
```diff
@@ -0,0 +1,104 @@
...
+| [`LLMInteractionStartedEvent`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionStartedEvent.java)     | Invoked when an LLM interaction has started.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```
dliubarskyi: Do you mind replacing `LLM` with `ChatModel` in all the classes here to be consistent with existing naming?

docs/docs/tutorials/auditing.md
```diff
@@ -0,0 +1,104 @@
+---
+sidebar_position: 13
+toc_max_heading_level: 5
+---
+
+# Auditing
```
dliubarskyi: We already use ["observability"/"listener"](https://docs.langchain4j.dev/tutorials/observability) naming, do you mind renaming the feature and interfaces to follow the existing naming? e.g. `AiServiceListener`

docs/docs/tutorials/auditing.md
```diff
@@ -0,0 +1,104 @@
...
+| [`OutputGuardrailExecutedEvent`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/OutputGuardrailExecutedEvent.java) | Invoked when an [output guardrail](https://docs.langchain4j.dev/tutorials/guardrails#output-guardrails) validation has been executed. One of these events will be fired for each invocation of a guardrail.<br/><br/>Contains information about the input to an individual output guardrail as well as its output (i.e. was it successful? failure? a retry? reprompt?).                                                                                                                                                                             |
+
+## Listening for an event
+
+Each of the [types of events](#types-of-events) has its own listener that can be implemented to receive the event. You can pick and choose which events you want to listen for.
```
dliubarskyi: I am wondering why each event has it's own listener? I understand the flexibility, but this can also be achieved with this approach:

```java
interface Listener {
    default void onEvent1(Event1) {}
    ...
    default void onEventN(EventN) {}
}
```

But with this approach users can also easily create listener that listens for all events at once, they don't need to search for all types of listeners and events and implement each one separately.

docs/docs/tutorials/auditing.md
```diff
@@ -0,0 +1,104 @@
...
+        Optional<Object> memoryId = interactionSource.memoryId();
+        Instant eventTimestamp = interactionSource.timestamp();
+        
+        // Do something with the data
+        
+    }
+}
+
+MyLLMInteractionCompleteEventListener myListener = new MyLLMInteractionCompleteEventListener();
+LLMInteractionEventListenerRegistrar.getInstance().register(myListener);
```
dliubarskyi: I am not sure about the approach of registering listeners globally as opposed to per-AiService...
I understand that it makes it easy to "cover everything" with just one merhod call, but it also has drawbacks (harder to test, especialy when tests are running in parallel)...

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionEvent.java
```diff
@@ -0,0 +1,52 @@
+package dev.langchain4j.audit.api.event;
+
+public interface LLMInteractionEvent {
```
dliubarskyi: I am not sure about the "LLM interaction" naming:
- in scope of a single invocation of an AI Service there can be multiple interactions with the LLM, not just one
- in scope of an AI Service invocation other actions (e.g., tools, guardrails, RAG (embedding models, retrievers), moderation models, etc) can happen which do not fit into the "LLM" naming

WDYT about "AI Service invocation"?

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/InteractionSource.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import org.jspecify.annotations.NonNull;
+
+/**
+ * Contains general information about the source of the interaction.
+ */
+public interface InteractionSource {
```
dliubarskyi: While working on https://github.com/langchain4j/langchain4j/pull/3569 I have created an [`InvocationContext`](https://github.com/langchain4j/langchain4j/pull/3569/files#diff-a5e1f0fe0fec7578b91836da69752bf2e7957989b9683556fed06151c9a8d12b) which seems to be representing the same thing as `InteractionSource`. Do you mind if I'll merge these 2 together afterwards (after our PRs are merged) to avoid duplication of entities?

langchain4j-core/src/main/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrar.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    private static class EventListeners<T extends LLMInteractionEvent> {
+        /**
+         * <strong>Implementation note:</strong> I chose {@link CopyOnWriteArraySet} here because it is thread-safe.
+         * The list will be very read heavy. The only time writes will occur is when listeners are initially registered and subsequently unregistered.
+         * I felt this was better than alternatives involving coarse-grained locking/synchronization
+         * (i.e. {@link java.util.Collections#synchronizedSet(Set)} or {@link java.util.concurrent.locks.ReadWriteLock}).
+         */
+        private final Set<@NonNull LLMInteractionEventListener<T>> listeners = new CopyOnWriteArraySet<>();
```
dliubarskyi: I am wondering if using a `List` would be beneficial here, as it would allow users to rely on the order of listeners.
As a user, when I register listener A, then listener B, I would probably expect them to be called in the same order.
This will also be consistent with the other listeners we already have.

langchain4j-core/src/main/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrar.java
```diff
@@ -0,0 +1,111 @@
...
+            this.listeners.add(ensureNotNull(listener, "listener"));
+        }
+
+        private void remove(LLMInteractionEventListener<T> listener) {
+            this.listeners.remove(ensureNotNull(listener, "listener"));
+        }
+
+        private void fireEvent(T event) {
+            ensureNotNull(event, "event");
+            this.listeners.forEach(listener -> listener.onEvent(event));
```
dliubarskyi: It would be great to wrap each `onEvent` call in a `try-catch` block to make sure if one listener fails, the following listeners are still called. In case of an exception, I would log it with a `WARN` level (same as we do elsewhere)

langchain4j-core/src/main/java/dev/langchain4j/audit/api/listener/LLMInteractionFailureEventListener.java
```diff
@@ -0,0 +1,21 @@
...
+ * that occurs when an interaction with a large language model (LLM) fails.
+ * This interface extends the generic {@link LLMInteractionEventListener},
+ * specializing it for handling failure events.
+ *
+ * Classes implementing this interface should handle scenarios where an LLM
+ * interaction encounters an error. These scenarios include capturing and
+ * processing the associated error details encapsulated within the event.
+ */
+@FunctionalInterface
+public interface LLMInteractionFailureEventListener extends LLMInteractionEventListener<LLMInteractionFailureEvent> {
```
dliubarskyi:
```suggestion
public interface LLMInteractionErrorEventListener extends LLMInteractionEventListener<LLMInteractionErrorEvent> {
```
Do you mind renaming it (failure -> error) to keep consistent with existing naming?

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionCompleteEvent.java
```diff
@@ -0,0 +1,82 @@
...
+ * with a large language model (LLM). This interface extends {@link LLMInteractionEvent}
+ * to include additional information about the result of the interaction.
+ *
+ * Classes implementing this interface are expected to provide details
+ * on what constitutes the result of an interaction, which may vary based
+ * on the specifics of the use case. The result can represent the outcome
+ * of the interaction process, a returned value, a processed value, or
+ * {@code null} if no meaningful result exists.
+ */
+public interface LLMInteractionCompleteEvent extends LLMInteractionEvent {
```
dliubarskyi:
```suggestion
public interface LLMInteractionCompletedEvent extends LLMInteractionEvent {
```
to keep consistent with other `*edEvent` classes

langchain4j-core/src/main/java/dev/langchain4j/audit/api/listener/LLMInteractionCompleteEventListener.java
```diff
@@ -0,0 +1,17 @@
...
+import dev.langchain4j.audit.api.event.LLMInteractionCompleteEvent;
+
+/**
+ * A listener for {@link LLMInteractionCompleteEvent}, which represents an event
+ * that occurs upon the completion of an interaction with a large language model (LLM).
+ * This interface extends the generic {@link LLMInteractionEventListener} interface,
+ * allowing it to specifically listen for completion events.
+ */
+@FunctionalInterface
+public interface LLMInteractionCompleteEventListener extends LLMInteractionEventListener<LLMInteractionCompleteEvent> {
```
dliubarskyi:
```suggestion
public interface LLMInteractionCompletedEventListener extends LLMInteractionEventListener<LLMInteractionCompletedEvent> {
```
Complete -> Completed

langchain4j-core/src/main/java/dev/langchain4j/audit/event/DefaultLLMInteractionCompleteEvent.java
```diff
@@ -0,0 +1,23 @@
+package dev.langchain4j.audit.event;
+
+import dev.langchain4j.audit.api.event.LLMInteractionCompleteEvent;
+import java.util.Optional;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Default implementation of {@link LLMInteractionCompleteEvent}.
+ */
+public class DefaultLLMInteractionCompleteEvent extends AbstractLLMInteractionEvent
```
dliubarskyi:
```suggestion
public class DefaultLLMInteractionCompletedEvent extends AbstractLLMInteractionEvent
```

langchain4j-core/src/test/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrarTests.java
```diff
@@ -0,0 +1,239 @@
...
+    private static final ToolExecutedEvent TOOL_EXECUTED_EVENT = ToolExecutedEvent.builder()
+            .interactionSource(DEFAULT_INTERACTION_SOURCE)
+            .request(ToolExecutionRequest.builder().build())
+            .result("Success!")
+            .build();
+
+    private static final List<LLMInteractionEvent> ALL_EVENTS = List.of(
+            LLM_RESPONSE_RECEIVED_EVENT,
+            LLM_INTERACTION_FAILURE_EVENT,
+            LLM_INTERACTION_COMPLETE_EVENT,
```
dliubarskyi:
```suggestion
            LLM_INTERACTION_COMPLETED_EVENT,
```

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -158,15 +183,19 @@ public void onCompleteResponse(ChatResponse chatResponse) {
...
                 intermediateResponseHandler.accept(chatResponse);
             }
-          
+
             boolean immediateToolReturn = true;
 
             if (toolExecutor != null) {
-                for (CompletableFuture<ToolExecutionResultMessage> toolResultFuture : toolResultFutures) {
+                for (Map.Entry<ToolExecutionRequest, CompletableFuture<ToolExecutionResultMessage>> toolResultFuture :
+                        toolResultFutures.entrySet()) {
```
dliubarskyi: Please make sure tool calls are processed in the same order they were invoked by LLM. I guess `entrySet` does not preserve order, unless it is `LinkedHashMap`

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi, line 1: It was very hard to review this file, so I have commited a slight refactoring that extracts try-catch from the main method, thus avoids additional nesting (that requires reformatting everything)

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -177,8 +184,34 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
...
 
-                        final Object memoryId = findMemoryId(method, args).orElse(ChatMemoryService.DEFAULT);
-                        final ChatMemory chatMemory = context.hasChatMemory()
+                        var methodArgs = (args != null) ? Arrays.asList(args) : Collections.emptyList();
+                        var memoryIdOpt = findMemoryId(method, args);
+                        var interactionSource = InteractionSource.builder()
+                                .interfaceName(context.aiServiceClass.getName())
+                                .methodName(method.getName())
+                                .methodArguments(methodArgs)
+                                .memoryId(memoryIdOpt.orElse(null))
```
dliubarskyi: By convention, `memoryId` is "default" in case it is not specified explicitly. Is there a reason why this behaviour is different here?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -187,11 +220,21 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                         var variables = InternalReflectionVariableResolver.findTemplateVariables(
                                 userMessageTemplate, method, args);
                         UserMessage userMessage = prepareUserMessage(method, args, userMessageTemplate, variables);
+
+                        LLMInteractionEventListenerRegistrar.getInstance()
+                                .fireEvent(LLMInteractionStartedEvent.builder()
+                                        .interactionSource(interactionSource)
+                                        .systemMessage(systemMessage)
+                                        .userMessage(userMessage)
```
dliubarskyi: user message can be modified multiple times below (RAG, guardrails, strucutred outputs). Is it intended to only record it at the "initial" state?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -323,32 +379,61 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                                     .intermediateResponses(toolServiceResult.intermediateResponses())
                                     .finalResponse(toolServiceResult.finalResponse())
                                     .build();
+
+                            // fire an interaction complete event
```
```suggestion
```

Please remove unnecessary comments everywhere (did you use an LLM to code this? 😉)

langchain4j/src/main/java/dev/langchain4j/service/guardrail/AbstractGuardrailService.java
```diff
@@ -53,18 +54,20 @@ public Class<?> aiServiceClass() {
...
     }
 
     @Override
-    public <MethodKey> OutputGuardrailResult executeOutputGuardrails(MethodKey method, OutputGuardrailRequest params) {
+    public <MethodKey> OutputGuardrailResult executeOutputGuardrails(
+            MethodKey method, OutputGuardrailRequest request, InteractionSource auditInteractionSource) {
         return Optional.ofNullable(method)
                 .map(this.outputGuardrails::get)
-                .map(executor -> executor.execute(params))
+                .map(executor -> executor.execute(request, auditInteractionSource))
```
dliubarskyi: Can't `InteractionSource` be a part of `OutputGuardrailRequest`? (same for input) This would avoid breaking the API

langchain4j/src/test/java/dev/langchain4j/service/AiServicesObservabilityTests.java
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    @AfterEach
+    void afterEach() {
+        Optional.ofNullable(this.wireMock).ifPresent(WireMockServer::stop);
+    }
+
+    @SystemMessage("You are a chat bot that answers questions")
+    interface Assistant {
+        String chat(String message);
```
dliubarskyi, line 516: Please test streaming case as well (AI Service returning `TokenStream`)

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionCompleteEvent.java
dliubarskyi: But in the end I would name it `AiServiceInvocationCompletedEvent`

langchain4j-core/src/main/java/dev/langchain4j/audit/api/listener/LLMInteractionFailureEventListener.java
dliubarskyi: BTW, `LLMInteractionFailureEvent` has an `error()` method

langchain4j/src/test/java/dev/langchain4j/service/AiServicesAuditingTests.java
```diff
@@ -0,0 +1,545 @@
...
+                        new MyOutputGuardrailExecutedEventListener(),
+                        new MyToolExecutedEventListener())
+                .collect(Collectors.toMap(LLMInteractionEventListener::getEventClass, Function.identity()));
+    }
+
+    @BeforeEach
+    void beforeEach() {
+        this.wireMock = new WireMockServer(wireMockConfig().dynamicPort());
+
+        var firstToolResponse =
```
dliubarskyi: Does this test really require raw JSON and wiremock? Perhaps `ChatModelMock` can be used?

dliubarskyi (commented):
@edeandrea thanks a lot, great job as always!

docs/docs/tutorials/auditing.md
edeandrea: I thought about that, but really its higher level than a `ChatModel`.

Maybe `AiService<EventName>Event`?

docs/docs/tutorials/auditing.md
edeandrea: I thought about this and always came back to the fact that eventually we'd like to add metrics/tracing to this as well, and that would be implemented with this infrastructure.

Each event can be delineated by AiService (and method within the service) via the `AiServiceInvocationContext`, as the `AiServiceInvocationContext` contains a reference to the AiService interface & method that triggered the event.

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionEvent.java
edeandrea: naming is hard :) I wasn't in love with the names I picked (to be fair, when I implemented this in Quarkus I didn't like the names that were picked there either, so I was trying to do better here :) ).

I'm open to naming them whatever you think is best.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
edeandrea: In a "started" event, the user message is whatever is being sent to the llm at that point in time. The interaction id would be different though, if you are aggregating/tracing by interaction id you could follow the user message changes.

I'm doing this in my POC app (https://github.com/edeandrea/non-deterministic-no-problem)

More specific events (`InputGuardrailExecutedEvent` for example) will also track the userMessage (it has a `rewrittenUserMessage()` method).

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
edeandrea: Actually I didn't :) I first went through and put in some comments where i thought I needed to do stuff. Forgot to clean them up :)

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/InteractionSource.java
edeandrea: By all means if its capturing the same information then of course.

BTW I like the `InvocationContext` name better than `InteractionSource` too :)

Maybe it should be `AiServiceInvocationContext`?

langchain4j-core/src/main/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrar.java
edeandrea: I went with a `Set` because I wanted to guarantee uniqueness, and you get that for free with a `Set`. We can switch to a `List` (or even expose it as `Collection`), but we would need to implement/guarantee uniqueness.

langchain4j-core/src/main/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrar.java
edeandrea: Good catch!

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
edeandrea: Agreed - this class is kind of a mess with lots of branches, early returns, etc. I didn't want to do a major refactoring as part of this PR.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
edeandrea: its only in the builder (so the builder supports `null`). When `builder.build()` is called, `null` is handled appropriately.

Downstream frameworks (i.e. Quarkus) don't use `default` - we use `null`, so I wanted to support `null`.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesObservabilityTests.java
edeandrea: Crap! Thanks for the reminder! Totally forgot about it. My fault.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesAuditingTests.java
edeandrea: I've found with things like this I always prefer to make a "real" call vs trying to mock behavior. You discover things that work (& don't work so well).

In this case too I really do need a scenario. I need the response to be different depending on a "state", which is very easy to do with wiremock scenarios.

And in my defense, I mostly copied/adapted the test from Quarkus where we prefer the wiremock approach over mocking :)

langchain4j/src/main/java/dev/langchain4j/service/guardrail/AbstractGuardrailService.java
edeandrea: You're right. I can probably even add it to the generic `GuardrailRequestParams`. Nice catch!

docs/docs/tutorials/auditing.md
edeandrea: I somewhat modeled this after how Spring & Quarkus works. Spring has the [`ApplicationListener`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/ApplicationListener.html#onApplicationEvent(E)) for `ApplicationEvent`s. Quarkus also has a similar capability where each event/listener is separate and the listener method name is the same, just the actual event being passed is different.

Well, actually in quarkus, there isn't an actual listener interface at all. You just register CDI observer for particular events, so I suppose I really modeled it after how Spring does it.

edeandrea:
Thanks @dliubarskyi for the thorough review! I've commented back on some of your comments. Others I'll start working on and hopefully will push changes in the next day or so.

langchain4j-core/src/main/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrar.java
edeandrea: Also, this is actually an implementation detail. This `Set` is never exposed outside of this class, and this class isn't intended to be used directly by applications. Sure its the default implementation of `LLMInteractionEventListenerRegistrar`, but its an implementation detail.

Also, I'm not sure order really matters? If we fire a `started` event, for example, does it matter in which order the listeners are registered? A downstream framework could also parallelize the execution so that all listeners are notified in parallel for example.

I'm not sure that order matters here. Uniqueness is far more important IMO.

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
edeandrea: Good catch!

langchain4j/src/main/java/dev/langchain4j/service/guardrail/AbstractGuardrailService.java
edeandrea: Done

langchain4j-core/src/main/java/dev/langchain4j/audit/api/listener/LLMInteractionFailureEventListener.java
edeandrea: Done

langchain4j-core/src/main/java/dev/langchain4j/audit/api/listener/LLMInteractionCompleteEventListener.java
edeandrea: Done

langchain4j-core/src/test/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrarTests.java
edeandrea: Done

langchain4j-core/src/main/java/dev/langchain4j/audit/event/DefaultLLMInteractionCompleteEvent.java
edeandrea: Done

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionCompleteEvent.java
edeandrea: Done

docs/docs/tutorials/auditing.md
edeandrea: Done

docs/docs/tutorials/auditing.md
edeandrea: I called everything `AiService<EventName>Event` to correlate with `AiService<EventName>EventListener`

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionEvent.java
edeandrea: I think in another comment we settled on `AiServiceInteractionEvent`, which is what I've done.

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/InteractionSource.java
edeandrea: Renamed to `AiServiceInvocationContext`

langchain4j-core/src/main/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrar.java
edeandrea: Done

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
edeandrea: Done

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
edeandrea: For RAG, maybe we should invent a new event around that (i.e. before/after a call to `RetrievalAugmentor.augment`?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
edeandrea: Done

langchain4j/src/test/java/dev/langchain4j/service/AiServicesAuditingTests.java
edeandrea: I've gotten it to work using `ChatModelMock`. I'll use that if you'd prefer.

edeandrea: Done

langchain4j/src/test/java/dev/langchain4j/service/AiServicesObservabilityTests.java
edeandrea: Working on this now

docs/docs/tutorials/auditing.md
edeandrea: Oh and I think the most important reason (and I forgot to call it out) was so that users could add their own custom events/listeners and fire them. 

Maybe spring boot wants to introduce some of its own events. Maybe Quarkus does. 

If there was only a single listener interface then that wouldn't be possible. Keeping them separate allows for extension.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesObservabilityTests.java
edeandrea: Done

edeandrea:
Hi @dliubarskyi !

I've fixed most of the review comments you had provided. These ones are still open because I felt like either there was nothing for me to do (as I replied to a question you asked), or there was still some discussion to be had:

- [x] [Why does each event have its own listener?](https://github.com/langchain4j/langchain4j/pull/3679#discussion_r2355037990)
- [x] [Registering listeners globally](https://github.com/langchain4j/langchain4j/pull/3679#discussion_r2355044183)
- [x] [Is ordering of the listener firing important?](https://github.com/langchain4j/langchain4j/pull/3679#discussion_r2355085146)
- [x] [Handling `null` memoryId](https://github.com/langchain4j/langchain4j/pull/3679#discussion_r2355363028)
- [x] [User message handling throughout event sequence](https://github.com/langchain4j/langchain4j/pull/3679#discussion_r2355368950)

langchain4j-core/src/main/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrar.java
dliubarskyi: > Also, this is actually an implementation detail.

True, but this behaviour is visible to the user. And because of it, it should be predictable.

> Also, I'm not sure order really matters? If we fire a started event, for example, does it matter in which order the listeners are registered?

Maybe it matters not for all users, but I can imagine there are users that care about the ordering because of their specific use case.

> Uniqueness is far more important IMO.

We can have both uniqueness and ordering, can't we?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: Sorry, I am not sure I understand.
If I am reading it correctly, `AiServicesAuditingTests` expects `memoryId` to be `Optional.empty()`, but it should be "default" to be consistent with all other places where we expose memory ID.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesAuditingTests.java
dliubarskyi: Thanks, but I can also see why you prefer wiremock, so feel free to revert

docs/docs/tutorials/auditing.md
dliubarskyi: Makes sense, thanks for the explanation!

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionEvent.java
dliubarskyi: We have `AiServiceInvocation*` classes and `AiServiceInteraction*` classes, I would stick with the `Invocation` everywhere to be consistent

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: Makes sense, this even can be added later if there will be a need

docs/docs/tutorials/auditing.md
dliubarskyi: > I thought about this and always came back to the fact that eventually we'd like to add metrics/tracing to this as well, and that would be implemented with this infrastructure.

Not sure I follow, how registering listeners on the AI Service instead of on the global registrar will prevent implementing metrics/tracing?

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/LLMInteractionEvent.java
edeandrea: Done

langchain4j-core/src/main/java/dev/langchain4j/audit/api/DefaultLLMInteractionEventListenerRegistrar.java
edeandrea: Done

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
edeandrea: Fixed

docs/docs/tutorials/auditing.md
edeandrea: I'm not sure why you would want to register any of these event listeners on a per-ai service basis? This should be functionality that is built in "for-free" IMO.

Each event can be delineated by AiService (and method within the service) via the `AiServiceInvocationContext`, as the `AiServiceInvocationContext` contains a reference to the AiService interface & method that triggered the event.

If a listener really does want to filter it can do so.

I feel strongly about this TBH.

mariofusco: I'm not as strongly opinionated, so FWIW I'm just adding my point of view here. Of course there are pros and cons in both approaches:  in general I'd prefer to register a listener on a single instance and I'm afraid that this global listeners design could badly affect testability, but I'm also totally understanding the reasons why Eric is doing this.  

Personally if, for any reason, I'd need to listen for an event emitted by a specific AIService I would like to register a listener directly on that service, instead of creating a global one and do something only for that service using an if condition. I don't know how common this scenario could be, but if it is a quite frequent need, as I suppose, the user will end with a single giant, hard to read and maintain, listening method filled with tons of if condition. I don't know how reasonable (or hard to implement/maintain) this could be, but ideally I would at least give to users the possibility to follow both approaches, the global one and the one specific for the single service.

edeandrea: Thank you for weighing in @mariofusco . Based on the conversation I will change the listening mechanism to be specific to an ai service.

dliubarskyi: @edeandrea thank you! Perhaps the difference in our points of view is that you approach it from the framework point of view, while I approach it from the toolbox/library point of view.

In any case, I hope that having it per-AiService will give it maximum flexibility, while also allowing it to be "built in for free" for Quarkus users

edeandrea: > Perhaps the difference in our points of view is that you approach it from the framework point of view

That might be true. Whenever I'm building something in here I always have downstream frameworks in mind. I want to build things as composible and extendable as possible so that it doesn't need to be "reimplemented" downstream.

You and @mariofusco make valid points though. I'm reluctant to do both global & per-service registration though. I think that might get too confusing for users, so I'll stick with per-service.

I'll get on this today and hopefully have it finished soon.

edeandrea: Done!

edeandrea:
I'm not sure what happened with the Java 24 build. Looks like some very strange error. Maybe the GitHub worker got messed up somehow?

docs/docs/tutorials/auditing.md
dliubarskyi: Do you mind moving the contents of this page into https://docs.langchain4j.dev/tutorials/observability under the "AI Service Observability" section? There is also many places left where the term "audit" is used

langchain4j-core/src/main/java/dev/langchain4j/audit/api/AiServiceInvocationEventListenerRegistrar.java
```diff
@@ -0,0 +1,99 @@
...
+import dev.langchain4j.audit.api.event.AiServiceInvocationEvent;
+import dev.langchain4j.audit.api.listener.AiServiceInvocationEventListener;
+import dev.langchain4j.spi.audit.AiServiceInvocationEventListenerRegistrarFactory;
+import java.util.Arrays;
+import java.util.ServiceLoader;
+
+/**
+ * A registrar for registering {@link AiServiceInvocationEventListener}s.
+ */
+public interface AiServiceInvocationEventListenerRegistrar {
```
dliubarskyi: I am wondering if we could shorten "AiServiceInvocationEventListener" into "AiServiceListener"?

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/AiServiceInvocationEvent.java
```diff
@@ -0,0 +1,55 @@
+package dev.langchain4j.audit.api.event;
+
+import dev.langchain4j.invocation.InvocationContext;
+
+public interface AiServiceInvocationEvent {
```
dliubarskyi:
```suggestion
public interface AiServiceEvent {
```
Would that still be clear?

langchain4j-core/src/main/java/dev/langchain4j/audit/api/listener/AiServiceInvocationErrorEventListener.java
```diff
@@ -0,0 +1,22 @@
...
+ * that occurs when an AI Service invocation fails.
+ * This interface extends the generic {@link AiServiceInvocationEventListener},
+ * specializing it for handling failure events.
+ * <p>
+ * Classes implementing this interface should handle scenarios where an AI Service
+ * invocation encounters an error. These scenarios include capturing and
+ * processing the associated error details encapsulated within the event.
+ */
+@FunctionalInterface
+public interface AiServiceInvocationErrorEventListener
```
dliubarskyi:
```suggestion
public interface AiServiceErrorListener
```
Seems to be clear, but much shorter, WDYT?

langchain4j-core/src/main/java/dev/langchain4j/audit/api/AiServiceInvocationEventListenerRegistrar.java
edeandrea: We can if you like.

edeandrea: if we want to go more generic, maybe we rename the package structure too (i.e. `dev.langchain4j.observability.*`)?

dliubarskyi: IMHO it is still clear, but much shorter and easier to read/understand. I would rename all `AiServiceInvocation*` classes into `AiService*` and `*EventListener` into `*Listener`, WDYT?

dliubarskyi: Yes, `observability` or `listener` works well!

edeandrea: i like `observability`. Its more generic (plus there are sub-packages called `listener`)

docs/docs/tutorials/auditing.md
edeandrea: Done

langchain4j-core/src/main/java/dev/langchain4j/audit/api/AiServiceInvocationEventListenerRegistrar.java
edeandrea: Done (package re-name + `AiServiceInvocation*` -> `AiService*` and `*EventListener` -> `*Listener`

langchain4j-core/src/main/java/dev/langchain4j/audit/api/event/AiServiceInvocationEvent.java
edeandrea: Done

langchain4j-core/src/main/java/dev/langchain4j/audit/api/listener/AiServiceInvocationErrorEventListener.java
edeandrea: Done

edeandrea:
Not sure why spotless is failing on files that weren't at all touched as part of this PR....

```
Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.44.4:check (default) on project langchain4j-core: The following files had format violations:
Error:      src/test/java/dev/langchain4j/rag/query/MetadataTest.java
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -268,6 +276,7 @@ public ToolServiceResult executeInferenceAndToolsLoop(
 
             if (immediateToolReturn) {
                 ChatResponse finalResponse = intermediateResponses.remove(intermediateResponses.size() - 1);
+                fireResponseReceivedEvent(finalResponse, invocationContext, invocationEventListenerRegistrar);
```
dliubarskyi:
```suggestion
```

Firing an event here would make it duplicate, there was an event fired already before invoking `executeInferenceAndToolsLoop` method

dliubarskyi (approved):
@edeandrea thank you so much, great job! 🙏

PR #3684, opened 2025-09-15
Initial MCP registry client implementation

jmartisk:
Just the raw implementation + basic docs + basic tests that run against the official registry (when/if we have our own MCP registry SDK, we might want to switch to using that)

The next step will probably be some tooling to attempt to automatically transform MCP server metadata into usable instances of `McpTransport` (for building an actual MCP client that talks to the server) in cases when it's possible...

### DIFF
--- a/docs/docs/tutorials/mcp.md
+++ b/docs/docs/tutorials/mcp.md
@@ -443,7 +443,7 @@ The registry client
 lives in the `dev.langchain4j.mcp.registryclient` package and can be initialized like this:
 
 ```java
-McpRegistryClient client = new DefaultMcpRegistryClient.Builder()
+McpRegistryClient client = DefaultMcpRegistryClient.builder()
         .baseUrl("URL-OF-THE-REGISTRY")
         .build();
 ```
--- a/docs/docs/tutorials/mcp.md
+++ b/docs/docs/tutorials/mcp.md
@@ -427,3 +427,29 @@ McpClient mcpClient = new DefaultMcpClient.Builder()
     .cacheToolList(false)
     .build();
 ```
+
+## MCP Registry client
+
+LangChain4j also offers a separate client implementation that can talk to 
+[MCP registries](https://registry.modelcontextprotocol.io/docs#/). Right now, only
+read-only operations are implemented (you can search for MCP servers, but managing and adding servers
+is not supported - please use the 
+[official tools](https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md) for that).
+
+**_WARNING:_** Discovering MCP servers and using them (especially running locally) may present serious security risks. Before
+running any MCP server that you locate in a public registry, make sure you can trust it.
+
+The registry client
+lives in the `dev.langchain4j.mcp.registryclient` package and can be initialized like this:
+
+```java
+McpRegistryClient client = new DefaultMcpRegistryClient.Builder()
+        .baseUrl("URL-OF-THE-REGISTRY")
+        .build();
+```
+
+If no base URL is provided, the official registry will be used as the default (https://registry.modelcontextprotocol.io).
+Then, to search for MCP servers, use the `registry.listServers(McpServerListRequest)` method. A `McpServerListRequest`
+object can be built using the `McpServerListRequest.Builder` class. The Java API in LangChain4j
+closely mirrors the REST API of MCP registries as described in the official
+[MCP Registry Reference](https://registry.modelcontextprotocol.io/docs).
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
@@ -3,6 +3,7 @@ package dev.langchain4j.internal;
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
 import static com.fasterxml.jackson.annotation.PropertyAccessor.FIELD;
 import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
+import static java.time.format.DateTimeFormatter.ISO_DATE_TIME;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE_TIME;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_TIME;
@@ -13,6 +14,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.module.SimpleModule;
@@ -24,6 +26,7 @@ import java.lang.reflect.Type;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.time.ZonedDateTime;
 import java.util.Optional;
 
 /**
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
@@ -3,7 +3,6 @@ package dev.langchain4j.internal;
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
 import static com.fasterxml.jackson.annotation.PropertyAccessor.FIELD;
 import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
-import static java.time.format.DateTimeFormatter.ISO_DATE_TIME;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE_TIME;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_TIME;
@@ -14,7 +13,6 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.SerializerProvider;
 import com.fasterxml.jackson.databind.module.SimpleModule;
@@ -26,7 +24,6 @@ import java.lang.reflect.Type;
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
-import java.time.ZonedDateTime;
 import java.util.Optional;
 
 /**
--- a/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
+++ b/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
@@ -8,10 +8,8 @@ import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 import static java.util.Arrays.asList;
 
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
 
 public class HttpRequest {
 
@@ -116,5 +114,4 @@ public class HttpRequest {
             return new HttpRequest(this);
         }
     }
-
 }
--- a/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
+++ b/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
@@ -8,8 +8,10 @@ import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 import static java.util.Arrays.asList;
 
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 public class HttpRequest {
 
@@ -114,4 +116,5 @@ public class HttpRequest {
             return new HttpRequest(this);
         }
     }
+
 }
--- a/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
+++ b/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
@@ -61,4 +61,5 @@ public class SuccessfulHttpResponse {
             return new SuccessfulHttpResponse(this);
         }
     }
+
 }
--- a/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
+++ b/langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
@@ -61,5 +61,4 @@ public class SuccessfulHttpResponse {
             return new SuccessfulHttpResponse(this);
         }
     }
-
 }
--- a/langchain4j-mcp/pom.xml
+++ b/langchain4j-mcp/pom.xml
@@ -40,6 +40,12 @@
             <version>1.5.0-SNAPSHOT</version>
         </dependency>
 
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-http-client-jdk</artifactId>
+            <version>1.5.0-SNAPSHOT</version>
+            <scope>runtime</scope>
+        </dependency>
         <!-- test dependencies -->
 
         <dependency>
--- a/langchain4j-mcp/pom.xml
+++ b/langchain4j-mcp/pom.xml
@@ -34,6 +34,12 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-http-client</artifactId>
+            <version>1.5.0-SNAPSHOT</version>
+        </dependency>
+
         <!-- test dependencies -->
 
         <dependency>
--- a/langchain4j-mcp/pom.xml
+++ b/langchain4j-mcp/pom.xml
@@ -46,6 +46,7 @@
             <version>1.5.0-SNAPSHOT</version>
             <scope>runtime</scope>
         </dependency>
+
         <!-- test dependencies -->
 
         <dependency>
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
@@ -1,0 +1,201 @@
+package dev.langchain4j.mcp.registryclient;
+
+import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
+import static com.fasterxml.jackson.annotation.PropertyAccessor.FIELD;
+import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
+import static java.time.format.DateTimeFormatter.ISO_DATE_TIME;
+
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.module.SimpleModule;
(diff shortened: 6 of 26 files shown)


jmartisk:
@dliubars this implements the `v0` version of the API. I wasn't completely sure how generic it should be to be able to handle possible future versions of the protocol, but given that we have no idea how the protocol will evolve, I didn't make it generic in the end. That's why `McpRegistryClient.listServers(McpServerListRequest)` directly uses `McpServerListRequest` which exactly mirrors the [query parameters](https://registry.modelcontextprotocol.io/docs#/operations/list-servers#Query-Parameters) from the API instead of trying to abstract it away somehow

langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
```diff
@@ -115,6 +118,14 @@ public LocalDateTime deserialize(JsonParser p, DeserializationContext ctxt) thro
             }
         });
 
+        module.addDeserializer(ZonedDateTime.class, new JsonDeserializer<>() {
```
dliubarskyi: Why is this change needed?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
```diff
@@ -114,4 +116,34 @@ public HttpRequest build() {
             return new HttpRequest(this);
         }
     }
+
+    @Override
+    public String toString() {
```
dliubarskyi: Perhaps you could reuse some logic from `HttpRequestLogger`?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
```diff
@@ -114,4 +116,34 @@ public HttpRequest build() {
...
+        return """
+            Method: %s
+            Url: %s
+            Headers:
+            %s
+            Body: %s
+            """.formatted(method, url, formatHeaders(headers), body);
+    }
+
+    private String formatHeaders(Map<String, List<String>> headers) {
```
dliubarskyi: Perhaps you could reuse `HttpRequestLogger.format()`?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
```diff
@@ -61,4 +61,12 @@ public SuccessfulHttpResponse build() {
             return new SuccessfulHttpResponse(this);
         }
     }
+
+    @Override
+    public String toString() {
```
dliubarskyi: Same here, `HttpResponseLogger`

langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
dliubarskyi: Sorry, I see why it is needed. Can it be contained in the MCP module?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        if (request.getVersion() != null) {
+            params.add("version=" + request.getVersion());
+        }
+        return params.stream().collect(Collectors.joining("&"));
+    }
+
+    private <T> T sendAndProcessResponse(HttpRequest httpRequest, Class<T> returnType) {
+        if (logRequests) {
+            log.info(httpRequest.toString());
```
dliubarskyi: Perhaps `LoggingHttpClient` can be used?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        SuccessfulHttpResponse response = httpClient.execute(httpRequest);
+        if (logResponses) {
+            log.info(response.toString());
+        }
+        if (response.statusCode() != 200) {
+            throw new McpRegistryClientException(
+                    "The server returned: " + response.statusCode() + ", body: " + response.body());
```
dliubarskyi: Not sure this is required, `SuccessfulHttpResponse` permits only `2XX` status code, otherwise an exception will be thrown from the `HttpClient.execute()`

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
```diff
@@ -0,0 +1,174 @@
...
+            this.baseUrl = baseUrl;
+            return this;
+        }
+
+        public Builder httpClient(HttpClient httpClient) {
+            this.httpClient = httpClient;
+            return this;
+        }
+
+        public Builder headers(Map<String, Supplier<String>> headers) {
```
dliubarskyi: nit: it would be great to have `headers(Map<String, String> headers)` as well for convenience

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/McpRegistryClientException.java
```diff
@@ -0,0 +1,7 @@
+package dev.langchain4j.mcp.registryclient;
+
+public class McpRegistryClientException extends RuntimeException {
```
dliubarskyi:
```suggestion
public class McpRegistryClientException extends LangChain4jException {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/model/McpEnvironmentVariable.java
```diff
@@ -0,0 +1,60 @@
+package dev.langchain4j.mcp.registryclient.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Map;
+
+public class McpEnvironmentVariable {
+    private String[] choices;
```
dliubarskyi: It would probably be more convenient to use `List<T>` instead of `T[]` here and in all other client classes

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/model/McpPackage.java
```diff
@@ -0,0 +1,70 @@
+package dev.langchain4j.mcp.registryclient.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class McpPackage {
+
+    private String file_sha256;
```
dliubarskyi: this field and it's getter are not following the same convention as other fields

langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
jmartisk: The registry (at least the official one) returns a zoned date time, so I had to add a deserializer for it

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/registryclient/McpRegistryClientTest.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+public class McpRegistryClientTest {
+
+    private static final Logger log = LoggerFactory.getLogger(McpRegistryClientTest.class);
+
+    static DefaultMcpRegistryClient client;
+
+    @BeforeAll
+    public static void prepare() {
+        client = new DefaultMcpRegistryClient.Builder()
```
dliubarskyi:
```suggestion
        client = DefaultMcpRegistryClient.builder()
```
nit: would be nice to have this convenience method

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/registryclient/McpRegistryClientTest.java
```diff
@@ -0,0 +1,108 @@
...
+                .logRequests(true)
+                .logResponses(true)
+                .build();
+    }
+
+    @Test
+    public void testListServers() {
+        McpServerList response = client.listServers(new McpServerListRequest.Builder()
```
dliubarskyi:
```suggestion
        McpServerList response = client.listServers(McpServerListRequest.builder()
```

dliubarskyi (commented):
@jmartisk good job, thanks a lot!

langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
dliubarskyi: Can you use a dedicated `ObjectMapper`?

dliubarskyi: Ideally `dev.langchain4j.internal.Json` class should not be used by MCP modules

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
jmartisk: Ah, I'd completely missed its existence. Yeah, it makes no sense to duplicate, I'll update it to reuse that class.

jmartisk: I would say that having a nice `toString()` on `HttpRequest` would also be nice too, but for that we would probably have to make the `HttpRequestLogger` methods public if that's ok.

dliubarskyi: @jmartisk yeah, sounds fine!

dliubarskyi: But in general, this sohuld not be needed as there is `LoggingHttpClient` that already takes care of logging

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
jmartisk: Ah ok. Should I remove the `McpRegistryClientException` class too? (This is currently the only usage of it. But maybe I could start throwing it when the JSON parsing fails...?)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
jmartisk: That won't work because of generic type erasure in Java :/ unless I name the other method differently. Could this one be `headerSuppliers`?? (it sounds a bit odd)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/model/McpEnvironmentVariable.java
jmartisk: That makes sense, will do

langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
jmartisk: Hmm I thought the reason the `Json` class exists is that all code should be unified to use it. But of course, I can create an `ObjectMapper` instead. Having a single set of static methods is inflexible.

jmartisk: I thought the goal was to abstract away from Jackson-specific APIs, but we have Jackson annotations on the classes anyway.

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
jmartisk: Yeah, I was thinking more for troubleshooting purposes (so that you can see a nice representation in the debugger even when logging is disabled) but.... maybe let's not change the visibility (and rework the logger class a bit) just for that

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
jmartisk: I kept `McpRegistryClientException` and throw it on JSON parsing problems

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/DefaultMcpRegistryClient.java
dliubarskyi: Oh right, forgot about erasure... I am wondering if it should be `Supplier<Map<String,String>>` instead? Then it could be named `headersSupplier`

jmartisk: Ok, I've changed it to `Supplier<Map<String, String>>` and you can pass either a supplier or a map directly now

docs/docs/tutorials/mcp.md
```diff
@@ -427,3 +427,29 @@ McpClient mcpClient = new DefaultMcpClient.Builder()
...
+[official tools](https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md) for that).
+
+**_WARNING:_** Discovering MCP servers and using them (especially running locally) may present serious security risks. Before
+running any MCP server that you locate in a public registry, make sure you can trust it.
+
+The registry client
+lives in the `dev.langchain4j.mcp.registryclient` package and can be initialized like this:
+
+```java
+McpRegistryClient client = new DefaultMcpRegistryClient.Builder()
```
```suggestion
McpRegistryClient client = DefaultMcpRegistryClient.builder()
```

langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
```diff
@@ -3,6 +3,7 @@
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
 import static com.fasterxml.jackson.annotation.PropertyAccessor.FIELD;
 import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
+import static java.time.format.DateTimeFormatter.ISO_DATE_TIME;
```
dliubarskyi:
```suggestion
```

langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
```diff
@@ -13,6 +14,7 @@
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.MapperFeature;
```
dliubarskyi:
```suggestion
```

langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
```diff
@@ -24,6 +26,7 @@
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.time.ZonedDateTime;
```
dliubarskyi:
```suggestion
```

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
```diff
@@ -8,8 +8,10 @@
 import static java.util.Arrays.asList;
 
 import java.util.HashMap;
+import java.util.Iterator;
```
dliubarskyi:
```suggestion
```

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
```diff
@@ -8,8 +8,10 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
```
dliubarskyi:
```suggestion
```

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/HttpRequest.java
```diff
@@ -114,4 +116,5 @@ public HttpRequest build() {
             return new HttpRequest(this);
         }
     }
+
```
dliubarskyi:
```suggestion
```

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
```diff
@@ -61,4 +61,5 @@ public SuccessfulHttpResponse build() {
             return new SuccessfulHttpResponse(this);
         }
     }
+
```
dliubarskyi:
```suggestion
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/model/McpRuntimeArgument.java
```diff
@@ -0,0 +1,81 @@
...
+    private boolean isSecret;
+
+    private String name;
+    private String type;
+    private String value;
+
+    @JsonProperty("value_hint")
+    private String valueHint;
+
+    private McpVariable[] variables;
```
dliubarskyi:
```suggestion
    private List<McpVariable> variables;
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/model/McpRuntimeArgument.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public String getValue() {
+        return value;
+    }
+
+    public String getValueHint() {
+        return valueHint;
+    }
+
+    public McpVariable[] getVariables() {
```
dliubarskyi:
```suggestion
    public List<McpVariable> getVariables() {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/model/McpPackage.java
```diff
@@ -0,0 +1,73 @@
...
+    private McpTransport transport;
+
+    @JsonProperty("runtime_arguments")
+    private List<McpRuntimeArgument> runtimeArguments;
+
+    @JsonProperty("package_arguments")
+    private List<McpPackageArgument> packageArguments;
+
+    @JsonProperty("environment_variables")
+    private McpEnvironmentVariable[] environmentVariables;
```
dliubarskyi:
```suggestion
    private List<McpEnvironmentVariable> environmentVariables;
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/registryclient/model/McpPackage.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public List<McpRuntimeArgument> getRuntimeArguments() {
+        return runtimeArguments;
+    }
+
+    public List<McpPackageArgument> getPackageArguments() {
+        return packageArguments;
+    }
+
+    public McpEnvironmentVariable[] getEnvironmentVariables() {
```
dliubarskyi:
```suggestion
    public List<McpEnvironmentVariable> getEnvironmentVariables() {
```

langchain4j-mcp/pom.xml
```diff
@@ -34,6 +34,12 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-http-client</artifactId>
+            <version>1.5.0-SNAPSHOT</version>
+        </dependency>
+
```
dliubarskyi, line 42:
```suggestion

        <dependency>
            <groupId>dev.langchain4j</groupId>
            <artifactId>langchain4j-http-client-jdk</artifactId>
            <version>1.5.0-SNAPSHOT</version>
            <scope>runtime</scope>
        </dependency>
```

langchain4j-mcp/pom.xml
```diff
@@ -34,6 +34,18 @@
...
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-http-client-jdk</artifactId>
+            <version>1.5.0-SNAPSHOT</version>
+            <scope>runtime</scope>
+        </dependency>
         <!-- test dependencies -->
```
dliubarskyi, line 50:
```suggestion

        <!-- test dependencies -->
```

dliubarskyi (approved):
@jmartisk thanks a lot!

dliubarskyi:
@jmartisk some tests are failing on [CI](https://github.com/langchain4j/langchain4j/actions/runs/17901075699), could you please check?

> 2025-09-22T00:31:43.8396820Z [ERROR]   McpRegistryClientTest.testGetServer:67 » Http {"title":"Not Found","status":404,"detail":"Server not found"}

It would also be great to rename (potentially unstable/flaky) integration tests from `*Test` into `*IT`, so that if one of them fails, they do not fail the whole build

jmartisk:
@dliubarskyi Oh man, give it just a few days and the content and behavior of the official registry will break the tests several times over (I've checked and there's more to it than just the 404 - they seem to have changed the datetime format; and one of the servers we were relying on has disappeared...). That's what you get for relying on an external service beyond your control. I'll come up with a quick fix and rename it to IT.

jmartisk:
We need to rework it to run against some mock server that is under our control ASAP...

jmartisk:
Right, the response format changed in a way that it's not just tests, the whole client is broken, nice.

PR #3685, opened 2025-09-15
Remove Lombok annotations from remaining code and replace with standard java checks and implementations.

### LINKED ISSUE #1636: [FEATURE] Get rid of Lombok

kushal-dixit:
## Issue

Fixes #1636 (Probably fully as I removed Lombok from wherever it was remaining). 

## Change

- Removed Lombok annotations and replaced with standard Java checks for ex. `Objects.requireNonNull` or equivalent implementation.
- Removed dependency from pom.xml as well, as there is no Lombok annotation remaining in code.

## General checklist

### DIFF
--- a/langchain4j-cassandra/src/main/java/dev/langchain4j/store/memory/chat/cassandra/CassandraChatMemoryStore.java
+++ b/langchain4j-cassandra/src/main/java/dev/langchain4j/store/memory/chat/cassandra/CassandraChatMemoryStore.java
@@ -11,12 +11,11 @@ import dev.langchain4j.data.message.ChatMessage;
 import dev.langchain4j.data.message.ChatMessageDeserializer;
 import dev.langchain4j.data.message.ChatMessageSerializer;
 import dev.langchain4j.store.memory.chat.ChatMemoryStore;
-import lombok.NonNull;
-import lombok.extern.slf4j.Slf4j;
 
 import java.net.InetSocketAddress;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.UUID;
 
 import static java.util.stream.Collectors.toList;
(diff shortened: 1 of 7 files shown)


kushal-dixit:
Note:
- Java build is currently failing due to known issues in `langchain4j-open-ai` test cases.
- Regarding Spotless, I’m hesitant to apply the suggested changes since they would introduce significant refactoring in unrelated code.

dliubarskyi (approved):
@kushal-dixit thank you!

langchain4j-cassandra/src/main/java/dev/langchain4j/store/memory/chat/cassandra/CassandraChatMemoryStore.java
```diff
@@ -127,7 +128,8 @@ public void updateMessages(@NonNull Object memoryId, @NonNull List<ChatMessage>
      * {@inheritDoc}
      */
     @Override
-    public void deleteMessages(@NonNull Object memoryId) {
```
kpavlov, line 130: I would argue against removing nullability annotations. Annotations help IDE and Kotlin compiler to understand data type and fail at compile time when needed. Let's replace it with JSpecify annotations

kpavlov (commented):
Thank @kushal-dixit for removing Lombok, but let migrate nullability annotations to JSpecify

langchain4j-cassandra/src/main/java/dev/langchain4j/store/memory/chat/cassandra/CassandraChatMemoryStore.java
kushal-dixit: Agreed, nullability annotations are super useful for IDEs and bug finders, I didn't realize we already had similar annotations in the project. 
btw I believe JSpecify is just compile time, we'd still need `Objects.requireNonNull` for runtime safety, right?

Since this PR is already merged, I will raise a follow-up PR to complements this PR changes.

kushal-dixit: PR https://github.com/langchain4j/langchain4j/pull/3694 raised for this.

PR #3687, opened 2025-09-15
Introduce AiServiceContextFactory

mariofusco:
## Issue

Closes #

## Change

Introduce an AiServiceContextFactory needed for the Quarkus extension.

## General checklist


langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -187,16 +188,17 @@ public static <T> T create(Class<T> aiService, StreamingChatModel streamingChatM
      * @return builder
      */
     public static <T> AiServices<T> builder(Class<T> aiService) {
-        AiServiceContext context = new AiServiceContext(aiService);
+        AiServiceContext context = AiServiceContext.create(aiService);
         return builder(context);
     }
 
+    private static class FactoryHolder {
+        private static final AiServicesFactory aiServicesFactory = loadFactory(AiServicesFactory.class);
```
mariofusco, line 196: This improvement is not directly related with the introduction of the new factory, but it allows to query the factory service (which is a relatively expensive operation) only the first time and not at each and every invocation.

dliubarskyi (approved):
@mariofusco thank you!

PR #3706, opened 2025-09-18
test: Enhance test coverage for core component validation and edge cases

alxkm:
This PR adds targeted test coverage for critical validation logic and edge cases across three core components:

**ContainsStringTest:**
- Added case sensitivity validation tests to ensure proper string matching behavior
- Added empty string edge case handling (empty search string, both empty)
- Added type safety tests for non-string metadata values (Double type mismatch)

**EmbeddingStoreContentRetrieverTest:**
- Added validation for dynamic configuration precedence over static values
- Added builder validation tests to ensure required fields (embedding store, embedding model) are enforced

**TypeUtilsTest:**
- Added wildcard type validation to prevent invalid service definitions with Set<?> return types
- Ensures proper error messaging for unsupported generic wildcards

These tests focus on preventing common configuration errors and ensuring robust validation of user inputs, improving the overall reliability of the framework.


dliubarskyi (approved):
@alxkm thank you!

PR #3707, opened 2025-09-18
When cancelling a tool, send the requestId as a number

### LINKED ISSUE #3689: [BUG] Cancellation feature - `requestId` can be either a string or a number
**Describe the bug**

The request ID can be either a [string or a number](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/6ca766214b99ed2241d8cdf1a570caa664306b62/schema/2025-06-18/schema.ts#L70). I think it is the case for every version of the MCP spec. 

When a tool is running for too long, langchain4j client properly timeout and send a cancellation notification to the server. The `requestId` used in this notification is a string. And langchain4j client uses always a number. So the server might mismatch the request ID and not cancel the tool. 

**Log and Stack trace**
n/a

**To Reproduce**

**Expected behavior**

The cancellation notification should have a request ID number instead of string.

- LangChain4j version: 1.4.0
- LLM(s) used: n/a
- Java version: 21

**Additional context**

jmartisk:
Fixes https://github.com/langchain4j/langchain4j/issues/3689


dliubarskyi (approved):
@jmartisk thank you!

PR #3710, opened 2025-09-18
Upgrading A2A SDK to 0.3.0.Beta1

ehsavoie:
## Issue

Upgrade the A2A support to 0.3.0.Beta1 
It it still using JSON-RPC but we can add more protocols if required.

## General checklist

### DIFF
--- a/langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
+++ b/langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
@@ -1,18 +1,24 @@
 package dev.langchain4j.agentic.a2a;
 
+import static dev.langchain4j.agentic.internal.AgentUtil.uniqueAgentName;
+
 import dev.langchain4j.agentic.UntypedAgent;
 import dev.langchain4j.agentic.internal.A2AClientBuilder;
 import dev.langchain4j.agentic.internal.AgentSpecification;
 import io.a2a.A2A;
-import io.a2a.client.A2AClient;
+import io.a2a.client.Client;
+import io.a2a.client.ClientEvent;
+import io.a2a.client.MessageEvent;
+import io.a2a.client.TaskEvent;
+import io.a2a.client.TaskUpdateEvent;
+import io.a2a.client.config.ClientConfig;
+import io.a2a.client.transport.jsonrpc.JSONRPCTransport;
+import io.a2a.client.transport.jsonrpc.JSONRPCTransportConfig;
 import io.a2a.spec.A2AClientError;
-import io.a2a.spec.A2AServerException;
+import io.a2a.spec.A2AClientException;
 import io.a2a.spec.AgentCard;
 import io.a2a.spec.Message;
-import io.a2a.spec.MessageSendParams;
 import io.a2a.spec.Part;
-import io.a2a.spec.SendMessageResponse;
-import io.a2a.spec.Task;
 import io.a2a.spec.TextPart;
 import java.lang.reflect.InvocationHandler;
 import java.lang.reflect.Method;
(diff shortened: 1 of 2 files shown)


langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
```diff
@@ -41,7 +49,15 @@ public class DefaultA2AClientBuilder<T> implements A2AClientBuilder<T> {
         this.agentCard = agentCard(a2aServerUrl);
         this.name = agentCard.name();
         this.uniqueName = uniqueAgentName(this.name);
-        this.a2aClient = new A2AClient(agentCard);
+        try {
+            this.a2aClient = Client.builder(agentCard)
+                    //Disabling streaming
+                    .clientConfig(new ClientConfig.Builder().setStreaming(false).setUseClientPreference(true).build())
```
fjuma: I don't think `setUseClientPreference` is needed here - that's used to try the transports in the order specified by the client as opposed to the order specified by the server. The client here only has one transport configured so we don't need to specify this.

langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
```diff
@@ -109,23 +125,46 @@ private Object invokeAgent(Object[] args) throws A2AServerException {
...
-
-        return ((Task)response.getResult()).getArtifacts().stream()
-                .flatMap(a -> a.parts().stream())
-                .filter(TextPart.class::isInstance)
-                .map(TextPart.class::cast)
-                .map(TextPart::getText)
-                .collect(Collectors.joining("\n"));
+        Message message =
+                new Message.Builder().role(Message.Role.USER).parts(parts).build();
+        AtomicReference<String> receivedEvent = new AtomicReference<>();
```
fjuma: We could use `CompletableFuture` here.

langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
```diff
@@ -109,23 +125,46 @@ private Object invokeAgent(Object[] args) throws A2AServerException {
...
+        AtomicBoolean wasUnexpectedEvent = new AtomicBoolean(false);
+        List<BiConsumer<ClientEvent, AgentCard>> consumers = List.of((event, agentCard) -> {
+            if (event instanceof MessageEvent messageEvent) {
+                receivedEvent.set(messageEvent.getMessage().getParts().stream()
+                        .filter(TextPart.class::isInstance)
+                        .map(TextPart.class::cast)
+                        .map(TextPart::getText)
+                        .collect(Collectors.joining("\n")));
+            } else if (event instanceof TaskEvent taskEvent) {
+                receivedEvent.set(taskEvent.getTask().getArtifacts().stream()
```
fjuma: If the client cannot send a streaming message, then the server should always send back a Message as opposed to a Task and then update events but it's ok to have this handling here.

langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
```diff
@@ -109,23 +125,46 @@ private Object invokeAgent(Object[] args) throws A2AServerException {
...
+                            .map(TextPart.class::cast)
+                            .map(TextPart::getText)
+                            .collect(Collectors.joining("\n")));
+                }
+            } else {
+                wasUnexpectedEvent.set(true);
+            }
+        });
+        AtomicReference<Throwable> receivedError = new AtomicReference<>();
+        a2aClient.sendMessage(message, consumers, (t) -> receivedError.set(t));
```
fjuma: Instead of setting `receivedError`, we could call `completeExceptionally` for the `CompletableFuture` like in this example:

https://github.com/a2aproject/a2a-java/blob/main/examples/helloworld/client/src/main/java/io/a2a/examples/helloworld/HelloWorldClient.java#L58-L84

langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
```diff
@@ -41,7 +49,15 @@ public class DefaultA2AClientBuilder<T> implements A2AClientBuilder<T> {
         this.agentCard = agentCard(a2aServerUrl);
         this.name = agentCard.name();
         this.uniqueName = uniqueAgentName(this.name);
-        this.a2aClient = new A2AClient(agentCard);
+        try {
+            this.a2aClient = Client.builder(agentCard)
+                    //Disabling streaming
+                    .clientConfig(new ClientConfig.Builder().setStreaming(false).setUseClientPreference(true).build())
+                    .withTransport(JSONRPCTransport.class, new JSONRPCTransportConfig())
```
fjuma, line 60: `addConsumers` and `streamingErrorHandler` could be specified here instead of via `sendMessage`. (The ability to specify them via `sendMessage` is meant for the case where you want to override what's currently set.)

mariofusco (approved):
@ehsavoie Great job, thanks a lot.

@dliubars Please feel free to merge this pull request.

fjuma (approved):
Thanks @ehsavoie!

Issue #3620, opened 2025-08-31
[FEATURE] Allow for listening to agent invocations


karolzdebel:
**Is your feature request related to a problem? Please describe.**
I can't find an easy way to listen to when an agent from the `agentic` module is invoked by a super agent so that I can produce logs and update the agent status that we display on our frontend UI.

**Describe the solution you'd like**
Allow for supplying an `AgentInvocationListener` (or something to that effect) in the `agentBuilder()` configuration. The listener could be invoked after the agent has, providing no guarantee that the agent completes successfully, just that the agent has been invoked at all. Although I wonder if invoking it afterwards could create race conditions, in which case it may be better to invoke it prior, blocking the invocation until the listener completes.

If we want to be prudent it may also be convenient to include an `AgentCompletionListener`. The completion listener
would be handy because it would be a convenient place to reset `AgenticScope` state if it was meant to only exist for the duration of an agents execution, versus persisting between executions.

**Describe alternatives you've considered**
An alternative I'm probably going with for the time being is doing this work inside the `contextProvider` callback but it feels hacky since its API is not intended for this.

LizeRaes:
I would have some additional requests for observability and ease of testing:
- it would be nice to have something like `List <Invocation> callTrace = AgenticScope.getInvocationTrace()` that contains a list of invocations with agent name and input and output values for each agent
- it would be nice to have access to the loop counter (on the AgenticScope maybe?) inside looping agents for both logging reasons as well as dynamically update exitCondition

A more general question about logging: Currently the model requests and responses are logged either as full HTTP calls (DEBUG) or not at all (INFO). It's never been easy to inspect those full model calls, but now I'm trying to trace what agentic systems are doing I was close to loosing my mind, so I made a PRETTY log wrapper (whether it's really pretty is open for discussion). Would it be possible to provide a similar logging level or system, optionally with some parameters, built into LangChain4j?

For inspiration, my log wrapper can be used like this on top of each class [see also this example](https://github.com/langchain4j/langchain4j-examples/pull/168/files#diff-f18155e7eaedf52ab7fef2e609203d71907416afaf115f9e2f01a5a0778981cb):
```
CustomLogging.setLevel(LogLevels.PRETTY, 100);
```
It intercepts the DEBUG logs and then parses those to display the 'latest' action wrt the model, being either:
- user's last request (incl. available tools - name only)
- model's last answer
- tool call with params emitted by the model
- our tool call answer
- any other type of log is rendered as usual

[...]

PR #3718, opened 2025-09-19
Introduce agents listeners

mariofusco:
## Issue

Closes #

https://github.com/langchain4j/langchain4j/issues/3620

## Change

This pull request introduces 2 agents listener:

1. `onAgentInvocation` invoked immediately before an agent is invoked
2. `onAgentCompletion` invoked immediatly after an agent invocation returns

This is still a draft and in particular the documentation is still missing because I would prefer to gather some feedback before writing it.

## General checklist

### DIFF
--- a/docs/docs/tutorials/agents.md
+++ b/docs/docs/tutorials/agents.md
@@ -537,6 +537,21 @@ UntypedAgent novelCreator = AgenticServices.sequenceBuilder()
         .build();
 ```
 
+## Observability
+
+Tracking and logging the agents' invocations can be crucial for debugging and understanding the aggregate behavior of the whole agentic system in which those agents participate. For this reason, the `langchain4j-agentic` module allows to register two different listeners through the `onAgentInvocation` and `onAgentCompletion` methods of the agent builders, that are notified respectively immediately before the invocation of an agent and immediately after it has completed its task and returned a result. For instance the following configuration of the `CreativeWriter` agent will log to the console when it is invoked and what is the story it generated.
+
+```java
+CreativeWriter creativeWriter = AgenticServices.agentBuilder(CreativeWriter.class)
+    .chatModel(baseModel())
+    .outputName("story")
+    .onAgentInvocation(request -> System.out.println("Invoking CreativeWriter with topic: " + request.inputs().get("topic")))
+    .onAgentCompletion(response -> System.out.println("CreativeWriter generated this story: " + response.output()))
+    .build();
+```
+
+These listeners receive as argument respectively an `AgentRequest` and an `AgentResponse` that provide useful information about the agent invocation, like its name, the inputs it received and the output it produced, together with the instance of the `AgenticScope` used for that invocation. Note that these listeners are invoked in the same thread used to also perform the agent invocation, so they are synchronous with it and should not perform long blocking operations.
+
 ## Declarative API
 
 All the workflow patterns discussed so far can be defined using a declarative API, which allows you to define workflows in a more concise and readable way. The `langchain4j-agentic` module provides a set of annotations that can be used to define agents and their workflows in a more declarative style.
@@ -603,17 +618,19 @@ public interface FoodExpert {
 }
 ```
 
-In a very similar way, annotating other `static` methods in the agent interface, it is possible to declaratively configure other aspects of the agent like its chat memory, the tools it can use, and so on. Those methods must have no arguments except for the one annotated with `@ChatMemoryProviderSupplier`. The list of annotations available to this purpose follows:
+In a very similar way, annotating other `static` methods in the agent interface, it is possible to declaratively configure other aspects of the agent like its chat memory, the tools it can use, and so on. Those methods must have no arguments unless differently specified in the following table. The list of annotations available to this purpose follows:
 
-| Annotation Name               | Description                                                                                                                                               |
-|-------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `@ChatModelSupplier`          | Returns the `ChatModel` to be used by this agent.                                                                                                         |
-| `@ChatMemorySupplier`         | Returns the `ChatMemory` to be used by this agent.                                                                                                        |
+| Annotation Name               | Description                                                                                                                                                   |
+|-------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `@ChatModelSupplier`          | Returns the `ChatModel` to be used by this agent.                                                                                                             |
+| `@ChatMemorySupplier`         | Returns the `ChatMemory` to be used by this agent.                                                                                                            |
 | `@ChatMemoryProviderSupplier` | Returns the `ChatMemoryProvider` to be used by this agent.<br/>This method requires as argument an `Object` to be used as the memoryId of the created memory. |
-| `@ContentRetrieverSupplier` | Returns the `ContentRetriever` to be used by this agent.                                                                                                  |
-| `@RetrievalAugmentorSupplier` | Returns the `RetrievalAugmentor` to be used by this agent.                                                                                                |
-| `@ToolsSupplier` | Returns the tool or set of tools to be used by this agent.<br/> It can return either a single `Object` or a `Object[]`                                         |
-| `@ToolProviderSupplier` | Returns the `ToolProvider` to be used by this agent.                                        |
+| `@ContentRetrieverSupplier`   | Returns the `ContentRetriever` to be used by this agent.                                                                                                      |
+| `@OnAgentCompletion`          | Notified when an agent invocation has been completed.<br/>This method requires as argument an `AgentResponse`.                                                |
+| `@OnAgentInvocation`          | Notified immediately before to perform an agent invocation.<br/>This method requires as argument an `AgentRequest`.                                           |
+| `@RetrievalAugmentorSupplier` | Returns the `RetrievalAugmentor` to be used by this agent.                                                                                                    |
+| `@ToolsSupplier`              | Returns the tool or set of tools to be used by this agent.<br/> It can return either a single `Object` or a `Object[]`                                        |
+| `@ToolProviderSupplier`       | Returns the `ToolProvider` to be used by this agent.                                                                                                          |
 
 To give another example of this declarative API, let's redefine through it the `ExpertsAgent` demonstrated in the conditional workflow section.
 
--- a/docs/docs/tutorials/agents.md
+++ b/docs/docs/tutorials/agents.md
@@ -539,14 +539,14 @@ UntypedAgent novelCreator = AgenticServices.sequenceBuilder()
 
 ## Observability
 
-Tracking and logging the agents' invocations can be crucial for debugging and understanding the aggregate behavior of the whole agentic system in which those agents participate. For this reason, the `langchain4j-agentic` module allows to register two different listeners through the `onAgentInvocation` and `onAgentCompletion` methods of the agent builders, that are notified respectively immediately before the invocation of an agent and immediately after it has completed its task and returned a result. For instance the following configuration of the `CreativeWriter` agent will log to the console when it is invoked and what is the story it generated.
+Tracking and logging the agents' invocations can be crucial for debugging and understanding the aggregate behavior of the whole agentic system in which those agents participate. For this reason, the `langchain4j-agentic` module allows to register two different listeners through the `beforeAgentInvocation` and `afterAgentInvocation` methods of the agent builders, that are notified respectively immediately before the invocation of an agent and immediately after it has completed its task and returned a result. For instance the following configuration of the `CreativeWriter` agent will log to the console when it is invoked and what is the story it generated.
 
 ```java
 CreativeWriter creativeWriter = AgenticServices.agentBuilder(CreativeWriter.class)
     .chatModel(baseModel())
     .outputName("story")
-    .onAgentInvocation(request -> System.out.println("Invoking CreativeWriter with topic: " + request.inputs().get("topic")))
-    .onAgentCompletion(response -> System.out.println("CreativeWriter generated this story: " + response.output()))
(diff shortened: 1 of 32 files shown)


mariofusco:
@karolzdebel I would appreciate it if you could give a look, let me know if it fits your needs and provide any other general feedback.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
```diff
@@ -20,9 +23,12 @@ public interface AgentInvoker extends AgentSpecification {
 
     Object[] toInvocationArguments(AgenticScope agenticScope) throws MissingArgumentException;
 
-    default Object invoke(Object agent, Object... args) throws AgentInvocationException {
+    default Object invoke(DefaultAgenticScope agenticScope, Object agent, Object... args) throws AgentInvocationException {
         try {
-            return method().invoke(agent, args);
+            onInvocation(new AgentRequest(agenticScope, name(), args));
+            Object result = method().invoke(agent, args);
+            onCompletion(new AgentResponse(agenticScope, name(), args, result));
```
karolzdebel: Am I understanding correctly that this will this work such that the execution of OnAgentCompletion will block the next agent from running? So that if I want to inject some state into the AgenticScope as a consequence of the completion so that the next agent can use it, I can do that safely?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
```diff
@@ -20,9 +23,12 @@ public interface AgentInvoker extends AgentSpecification {
 
     Object[] toInvocationArguments(AgenticScope agenticScope) throws MissingArgumentException;
 
-    default Object invoke(Object agent, Object... args) throws AgentInvocationException {
+    default Object invoke(DefaultAgenticScope agenticScope, Object agent, Object... args) throws AgentInvocationException {
```
karolzdebel: agenticScope is mutable right? Meaning changes can happen inside onInvocation() and will be reflected inside onCompletion?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
mariofusco: Yes, that's correct. It will work as you wrote, even though I'm not entirely sure if this could be considered a good practice. Can you please give a practical example of how you may want to use this?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
mariofusco: This is correct. Just consider that when `onInvocation` is called the arguments needed to invoke the agent have been already extracted from the `AgenticScope`, so that changing something in the  `AgenticScope` in the listener won't have any effect on the agent invocation. I believe this is the safest approach, but let me know if it makes sense for you.

karolzdebel: Sounds good.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
karolzdebel: We have the following agents
1) Requirements gatherer which runs first when the creative concept for a video is missing
2) Creative composition agent which puts together a concept based on requirements and subsequent user requests
3) Video generation agent which produces a video using a creative concept

I may want to reset the creative concept state (stemming from creative concept composition agent) after a video was generated (using the video generation agent) so that the next agent that runs is the requirements gatherer, not the creative concept composer, in anticipation of the user wanting a new video to be created next.

Although in practice there will typically be a turn of gathering input from the user before deciding whether to engage the creative composition agent or the requirements gatherer which introduces enough latency that the `OnCompletion` will likely finish so this may be a non-issue, but technically the system still wouldn't be thread safe if `OnCompletion` ran async.

karolzdebel (approved):
Draft LGTM from client perspective

karolzdebel:
@mariofusco Not sure if you noted this already but someone in Discord asked "Is there a way to notify the user about the current step or progress of a supervisor agent's invocation?". 

Wanted to mention in case supervisor agent has caveats.

mariofusco:
@karolzdebel I would appreciate it if you could give a look, let me know if it fits your needs and provide any other general feedback.

> @mariofusco Not sure if you noted this already but someone in Discord asked "Is there a way to notify the user about the current step or progress of a supervisor agent's invocation?".
> 
> Wanted to mention in case supervisor agent has caveats.

Agents participating to a supervisor architecture can be observed like any other agents, if I'm understanding the requirement correctly there shouldn't be nothing specific for the supervisor pattern in this regard.

docs/docs/tutorials/agents.md
```diff
@@ -520,6 +520,21 @@ UntypedAgent novelCreator = AgenticServices.sequenceBuilder()
         .build();
 ```
 
+## Observability
+
+Tracking and logging the agents' invocations can be crucial for debugging and understanding the aggregate behavior of the whole agentic system in which those agents participate. For this reason, the `langchain4j-agentic` module allows to register two different listeners through the `onAgentInvocation` and `onAgentCompletion` methods of the agent builders, that are notified respectively immediately before the invocation of an agent and immediately after it has completed its task and returned a result. For instance the following configuration of the `CreativeWriter` agent will log to the console when it is invoked and what is the story it generated.
```
dliubarskyi: nit: consider using `beforeAgentInvocation` and `afterAgentInvocation` names, as it seems (at least for me) a bit more clear (especially for `onAgentInvocation` - it is not entirely clear from the name alone whether it will be called right before or right after the invocation) and is symmetrical

docs/docs/tutorials/agents.md
```diff
@@ -586,17 +601,19 @@ public interface FoodExpert {
...
-| `@ContentRetrieverSupplier` | Returns the `ContentRetriever` to be used by this agent.                                                                                                  |
-| `@RetrievalAugmentorSupplier` | Returns the `RetrievalAugmentor` to be used by this agent.                                                                                                |
-| `@ToolsSupplier` | Returns the tool or set of tools to be used by this agent.<br/> It can return either a single `Object` or a `Object[]`                                         |
-| `@ToolProviderSupplier` | Returns the `ToolProvider` to be used by this agent.                                        |
+| `@ContentRetrieverSupplier`   | Returns the `ContentRetriever` to be used by this agent.                                                                                                      |
+| `@OnAgentCompletion`          | Notified when an agent invocation has been completed.<br/>This method requires as argument an `AgentResponse`.                                                |
```
dliubarskyi: nit: same here (before/after) and below

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
```diff
@@ -18,11 +21,14 @@ public interface AgentInvoker extends AgentSpecification {
...
     String toCard();
 
-    Object[] toInvocationArguments(AgenticScope agenticScope) throws MissingArgumentException;
+    AgentInvocationArguments toInvocationArguments(AgenticScope agenticScope) throws MissingArgumentException;
 
-    default Object invoke(Object agent, Object... args) throws AgentInvocationException {
+    default Object invoke(DefaultAgenticScope agenticScope, Object agent, AgentInvocationArguments args) throws AgentInvocationException {
         try {
-            return method().invoke(agent, args);
+            onInvocation(new AgentRequest(agenticScope, name(), args.namedArgs()));
```
dliubarskyi: What would/should happen if an exception will be thrown from the user-provided callback?
- fail the whole flow
- ignore the exception
- ignore the exception and log it
- somehting else?

dliubarskyi (dismissed):

docs/docs/tutorials/agents.md
mariofusco: Agreed, I will change it.

dliubarskyi: It is also worth thinking right away how to name the callback that will be called if agent invocation fails for some reason (to make sure all 3 share the same naming schema and fit well together)

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
mariofusco: I would say that an exception thrown in user's code is something that should be raised up to the user again and then make the whole flow to fail, but to be honest I'm not strongly opinionated on this. Any different point of view?

dliubarskyi: Considering that this is an observability feature (secondary priority), I am not sure whether an exception (for example, because of an observability backend is temporarily unreachable) should stop the flow (primary priority). User should definitely be notified though

docs/docs/tutorials/agents.md
mariofusco: Done.

docs/docs/tutorials/agents.md
mariofusco: Done.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
mariofusco: Agreed, done.

dliubarskyi (dismissed):
@mariofusco thanks a lot!

mariofusco:
Hi @karolzdebel, we added a few changes after your last review, please let us know if you want to review this again before we merge it, thanks.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
```diff
@@ -14,15 +19,32 @@
...
+            beforeInvocation(new AgentRequest(agenticScope, name(), args.namedArgs()));
+        } catch (Exception e) {
+            LOG.error("Before agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
+        }
+        Object result = internalInvoke(agent, args);
+        try {
+            afterInvocation(new AgentResponse(agenticScope, name(), args.namedArgs(), result));
+        } catch (Exception e) {
+            LOG.error("After agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
+        }
```
karolzdebel, line 41: I don't think you want to enable a silent failure here if the client throws an exception (whether deliberately or otherwise). This isn't the behavior I'd expect as a client and think it's generally bad practice for the system to swallow exceptions like this since it hides errors. Why not just let the system fail?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
```diff
@@ -14,15 +19,32 @@
...
+
+    default Object invoke(AgenticScope agenticScope, Object agent, AgentInvocationArguments args) throws AgentInvocationException {
+        try {
+            beforeInvocation(new AgentRequest(agenticScope, name(), args.namedArgs()));
+        } catch (Exception e) {
+            LOG.error("Before agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
+        }
+        Object result = internalInvoke(agent, args);
+        try {
+            afterInvocation(new AgentResponse(agenticScope, name(), args.namedArgs(), result));
```
karolzdebel, line 38: `afterInvocation()` to me sounds like "after the agent was invoked" and **not** "after the agent completed executing the method that was invoked", which are two different things, and I think the intent is to be able to detect the latter. Not sure what a better name is, maybe `afterExecution()` or `afterCompletion()` is better?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
mariofusco: I discussed this with @dliubarskyi and he clarified here https://github.com/langchain4j/langchain4j/pull/3718#discussion_r2368683694 the rationals behind this. Basically we believe that an observability issue shouldn't be enough to break the main flow of the application.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
mariofusco: I'm terrible at naming things, and I see what you mean. Anyway we want a pair of `beforeX` and `afterX` methods, but considering the distinction that you're making between `invocation` and `completion`, I'm not able to find a better `X` that clarifies the actual meaning of the 2 methods. Any better suggestion?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
karolzdebel: That feels like it should be the client's decision since systems have different standards around error handling and not everyone will use this only for side effect free observability. Catching the exception this way basically takes away the clients ability to let unexpected exceptions propagate, if that's their standard. If someone doesn't want to crash and instead wants to log they can easily still do that through a simple try catch in the callback. 

In our system even if the component responsible only for streaming a status update produces an unexpected exception I would not tolerate that being silently logged, if we had control over the code.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
karolzdebel: Some ideas:

`beforeRun` and `afterRun`

`beforeExecution` and `afterExecution`

But IMO it's better to have a different `X` for each method than use the current semantics since they're inaccurate, even if it's a little tricker to find the corresponding method.

PR #3724, opened 2025-09-22
Make the parsing of SSE events more robust

jmartisk:
Fixes https://github.com/langchain4j/langchain4j/issues/3697

PR #3725, opened 2025-09-22
Test for 'data:' inside a SSE event

jmartisk:
Follow-up on https://github.com/langchain4j/langchain4j/pull/3724/


dliubarskyi (approved):
@jmartisk thanks a lot!

PR #3733, opened 2025-09-23
Rework MCP client to support structured content results

### LINKED ISSUE #3714: [FEATURE] Support structured responses from MCP tools
The latest version of the MCP spec allows tools to return content in a structured way:
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content

```json
{
  "jsonrpc": "2.0",
  "id": 5,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "{\"temperature\": 22.5, \"conditions\": \"Partly cloudy\", \"humidity\": 65}"
      }
    ],
    "structuredContent": {
      "temperature": 22.5,
      "conditions": "Partly cloudy",
      "humidity": 65
    }
  }
}
```

The "content" element is recommended for backward compatibility (SHOULD), but not necessarily present.
In the latter case, LangChain4J fails to process the response because of checks performed here: 
https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java#L28

This is typically the case when consuming tools exposed by a Quarkus MCP server (ref https://github.com/quarkiverse/quarkus-mcp-server/issues/426).

Would it be possible to support this feature, or at least make response parsing more lenient?

jmartisk:
Fixes #3714 

Also bumps quarkus-mcp-server for integration tests to 1.5.3

@dliubars I'm not completely sure about changing the return type of `McpClient.executeTool`, if you prefer I can instead add a new method (`executeToolStructured`?) and keep the original one that returns String. The problem is I don't like the name `executeToolStructured` (better idea?) and it feels like it would unnecessarily clutter the API a bit.

The best long-term strategy is probably to have just `executeTool` that returns a `ToolExecutionResult`, but there's no way to achieve that without a breaking change. I would say most people use a `McpToolProvider` instead of directly working with `McpClient`, so the breakage should be relatively minor.

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -30,6 +30,7 @@ import dev.langchain4j.mcp.client.protocol.McpReadResourceRequest;
 import dev.langchain4j.mcp.client.protocol.McpRootsListChangedNotification;
 import dev.langchain4j.mcp.client.transport.McpOperationHandler;
 import dev.langchain4j.mcp.client.transport.McpTransport;
+import dev.langchain4j.service.tool.ToolExecutionResult;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
@@ -90,7 +91,7 @@ public class DefaultMcpClient implements McpClient {
             key = getOrDefault(builder.key, () -> UUID.randomUUID().toString());
             clientName = getOrDefault(builder.clientName, "langchain4j");
             clientVersion = getOrDefault(builder.clientVersion, "1.0");
-            protocolVersion = getOrDefault(builder.protocolVersion, "2024-11-05");
+            protocolVersion = getOrDefault(builder.protocolVersion, "2025-06-18");
             initializationTimeout = getOrDefault(builder.initializationTimeout, Duration.ofSeconds(30));
             toolExecutionTimeout = getOrDefault(builder.toolExecutionTimeout, Duration.ofSeconds(60));
             resourcesTimeout = getOrDefault(builder.resourcesTimeout, Duration.ofSeconds(60));
@@ -227,7 +228,7 @@ public class DefaultMcpClient implements McpClient {
     }
 
     @Override
-    public String executeTool(ToolExecutionRequest executionRequest) {
+    public ToolExecutionResult executeTool(ToolExecutionRequest executionRequest) {
         assertNotClosed();
         ObjectNode arguments = null;
         try {
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
@@ -2,6 +2,7 @@ package dev.langchain4j.mcp.client;
 
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.service.tool.ToolExecutionResult;
 import java.util.List;
 import java.util.Map;
 
@@ -22,10 +23,10 @@ public interface McpClient extends AutoCloseable {
     List<ToolSpecification> listTools();
 
     /**
-     * Executes a tool on the MCP server and returns the result as a String.
-     * Currently, this expects a tool execution to only contain text-based results.
+     * Executes a tool on the MCP server and returns the result.
+     * Currently, this expects a tool execution to only contain text-based results or JSON structured content.
      */
-    String executeTool(ToolExecutionRequest executionRequest);
+    ToolExecutionResult executeTool(ToolExecutionRequest executionRequest);
 
     /**
      * Obtains the current list of resources available on the MCP server.
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
@@ -1,12 +1,16 @@
 package dev.langchain4j.mcp.client;
 
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
-import java.util.stream.StreamSupport;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import dev.langchain4j.exception.ToolArgumentsException;
 import dev.langchain4j.exception.ToolExecutionException;
+import dev.langchain4j.service.tool.ToolExecutionResult;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 
 class ToolExecutionHelper {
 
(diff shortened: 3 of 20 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
```diff
@@ -41,6 +56,38 @@ static String extractResult(JsonNode result) {
...
+        return switch (content.getNodeType()) {
+            case BOOLEAN -> content.asBoolean();
+            case NUMBER ->
+                switch (content.numberType()) {
+                    case INT -> content.asInt();
+                    case LONG, BIG_INTEGER -> content.asLong();
+                    case FLOAT, DOUBLE, BIG_DECIMAL -> content.asDouble();
+                };
+            case STRING -> content.asText();
+            case NULL -> new Object(); // use Object as placeholder for a null value in a map
```
dliubarskyi: Why `new Object()` instead of `null`?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -90,7 +91,7 @@ public DefaultMcpClient(Builder builder) {
             key = getOrDefault(builder.key, () -> UUID.randomUUID().toString());
             clientName = getOrDefault(builder.clientName, "langchain4j");
             clientVersion = getOrDefault(builder.clientVersion, "1.0");
-            protocolVersion = getOrDefault(builder.protocolVersion, "2024-11-05");
+            protocolVersion = getOrDefault(builder.protocolVersion, "2025-06-18");
```
dliubarskyi, line 94: Is it a breaking change?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
jmartisk: It turns into a map value within `Collectors.toMap` which forbids producing null values. I didn't want to limit it only to map types that support null values.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
```diff
@@ -1,29 +1,44 @@
...
             JsonNode resultNode = result.get("result");
-            if (resultNode.has("content")) {
+            if (resultNode.has("structuredContent")
+                    && !resultNode.get("structuredContent").isNull()) {
+                JsonNode content = resultNode.get("structuredContent");
+                if (isError(resultNode)) {
+                    throw new ToolExecutionException(content.toString());
+                }
+                return ToolExecutionResult.builder()
+                        .result(toObject(content))
```
dliubarskyi, line 34: Perharps it is worth propagating `isError` here? Both `ToolExecutionResult` and [MCP spec](https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling) have it

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: Not really, this is just a pointer for the server, it will tell the server that we now support features from that protocol version (structured content is new in that version).

dliubarskyi: Nice, thanks!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
jmartisk: I see, `McpClient` should probably set `isError` if it's a business error (we can receive a response) and only throw `ToolExecutionException` if we don't get any response at all, correct?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
dliubarskyi: But it leads to the wrong data in the `ToolExecutionResult.result()`: if server returns `null` for some field, the `Map` inside the `ToolExecutionResult.result()` will contain an `Object` instead of `null`

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
dliubarskyi: Oh no, actually existing logic is right: errors should lead to exceptions

dliubarskyi: `isError` inside the `ToolExecutionResult` is a temporary (until LC4j 2.0) way to say that `String` returned by `DefaultToolExecutor` is an error. This is useful for some providers like Anthropic that have the `isError` field inside the tool result object.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolExecutionHelper.java
jmartisk: I think it's a general recommendation to avoid maps with null values in Java (it leads to a bit of a mess because `map.get("key")` returns null when there is a null value as well as no mapping at all, and many Map implementations explicitly forbid it), at least that's what I've been told many times.

We could avoid that by introducing some special placeholder for null (other than `new Object()`).

Or if you don't believe in this rule, we could change the algorithm to avoid `Collectors.toMap` and instead manually create and fill a `HashMap`, which allows null values.

dliubarskyi: What if we don't include keys with `null` values?

jmartisk: Then the application can't tell whether the MCP server returned that key or not, that's not good either.. I'm starting to lean towards just ignoring the rule this time and including a plain `null`...

dliubarskyi: Ok, sounds good

jmartisk: Done

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
```diff
@@ -22,10 +23,10 @@ public interface McpClient extends AutoCloseable {
     List<ToolSpecification> listTools();
 
     /**
-     * Executes a tool on the MCP server and returns the result as a String.
-     * Currently, this expects a tool execution to only contain text-based results.
+     * Executes a tool on the MCP server and returns the result.
+     * Currently, this expects a tool execution to only contain text-based results or JSON structured content.
      */
-    String executeTool(ToolExecutionRequest executionRequest);
+    ToolExecutionResult executeTool(ToolExecutionRequest executionRequest);
```
dliubarskyi, line 29: Perhaps we can add a new method that will accept `ToolExecutionRequest` with an additional `McpToolContext` (not sure it it sohuld be `InvocationContext`, after all this is an MCP client, far away from AI Service domain)?
Pros:
- it will allow to pass extra data (that does not belong to `ToolExecutionRequest`) to the MCP server in the future (e.g., user IDs, authorization data, etc)
- no breaking change
WDYT?

dliubarskyi: Another option would be to create a `McpToolExecutionRequest` class that contains `ToolExecutionRequest` (for now) and add a new method:
`ToolExecutionResult executeTool(McpToolExecutionRequest request)`

dliubarskyi: Second option would also allow for extension in the future, but will not require us to create `McpToolContext` right now

jmartisk: I'm not sure what the `McpToolContext` or `McpToolExecutionRequest` would contain (in the future) - authorization data is usually either baked by the developer into the headers when creating an MCP client, or passed by propagating the `Authorization` header of the current user (in which case, at least in Quarkus, the user data is stored in the CDI request scope). I'm not sure there will be a need to add any such data to a MCP context, and the MCP protocol currently doesn't allow passing any extra metadata with a tool call. So I fear it might be prematurely cluttering the API for no good reason...

If you're worried about this breaking change, I think it's not that bad since most people don't probably use `McpClient` directly... And the MCP module is still beta :)

dliubarskyi: OK, I am probably overthinking it now

dliubarskyi (approved):
@jmartisk thanks a lot!

PR #3785, opened 2025-09-29
Allow accessing AgenticScope from within @Tool-annotated method

### LINKED ISSUE #3762: [FEATURE] Allow for writing into AgenticScope from tools
There is sometimes state that's relevant to the overall agentic system that's available inside of tools that would be convenient to write into `AgenticScope` without having to go through the hurdle of returning it in the agent's response. 

For example, in our system we have an image generator that's used in our video generation agent. The image produced is relevant to other agents however the image is never returned as part of the agent's response since the response goes to the user, and the user does not need the image. The reason the image is relevant to the other agents is because in later conversation turns the user can indicate they want to save the ingredients that composed their video, and the image is one of those resources.

Allow for injecting `AgenticScope` into tool methods when using the `agentic` module. `AgenticScope` can be a parameter you optionally add that's automatically resolved by the framework.

I'm currently holding a reference to the `AgenticScope` in memory for the duration of the agent's execution and passing it around our system but this is prone to memory leaks, stale state, and requires extra effort.

Another alternative is to implement our own internal version of `AgenticScope` that we use for such state, but this will get ugly too I figure.

mariofusco:
## Issue
Closes https://github.com/langchain4j/langchain4j/issues/3762

## Change

This pull request adds the `LangChain4jManaged` marker interface, making the `AgenticScope` to extend it, in order to introduce a specific type of (internal?) parameter that can be used for both AiServices and Tools invocations.

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
@@ -38,7 +38,7 @@ public interface AgentInvoker extends AgentSpecification {
         new InvocationParameters(Map.of(AgenticScope.class.getName(), agenticScope)).setAsCurrent();
         Object result = internalInvoke(agent, args);
         try {
-            InvocationParameters.resetCurrentParameters();
+            InvocationParameters.removeCurrent();
             afterInvocation(new AgentResponse(agenticScope, name(), args.namedArgs(), result));
         } catch (Exception e) {
             LOG.error("After agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
@@ -8,11 +8,13 @@ import dev.langchain4j.agentic.agent.MissingArgumentException;
 import dev.langchain4j.agentic.scope.AgenticScope;
 import dev.langchain4j.agentic.UntypedAgent;
 import dev.langchain4j.agentic.scope.DefaultAgenticScope;
+import dev.langchain4j.invocation.InvocationParameters;
 import dev.langchain4j.service.V;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
+import java.util.Map;
 import java.util.Optional;
 
 import static dev.langchain4j.agentic.internal.AgentUtil.argumentsFromMethod;
@@ -33,8 +35,10 @@ public interface AgentInvoker extends AgentSpecification {
         } catch (Exception e) {
             LOG.error("Before agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
         }
+        new InvocationParameters(Map.of(AgenticScope.class.getName(), agenticScope)).setAsCurrent();
         Object result = internalInvoke(agent, args);
         try {
+            InvocationParameters.resetCurrentParameters();
             afterInvocation(new AgentResponse(agenticScope, name(), args.namedArgs(), result));
         } catch (Exception e) {
             LOG.error("After agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
@@ -35,7 +35,7 @@ public interface AgentInvoker extends AgentSpecification {
         } catch (Exception e) {
             LOG.error("Before agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
         }
-        new InvocationParameters(Map.of(AgenticScope.class.getName(), agenticScope)).setAsCurrent();
+        InvocationParameters.setCurrent(new InvocationParameters(Map.of(AgenticScope.class.getName(), agenticScope)));
         Object result = internalInvoke(agent, args);
         try {
             InvocationParameters.removeCurrent();
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
@@ -7,8 +7,7 @@ import dev.langchain4j.agentic.agent.AgentResponse;
 import dev.langchain4j.agentic.agent.MissingArgumentException;
 import dev.langchain4j.agentic.scope.AgenticScope;
 import dev.langchain4j.agentic.UntypedAgent;
-import dev.langchain4j.agentic.scope.DefaultAgenticScope;
-import dev.langchain4j.invocation.InvocationParameters;
+import dev.langchain4j.invocation.LangChain4jManaged;
 import dev.langchain4j.service.V;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -35,10 +34,10 @@ public interface AgentInvoker extends AgentSpecification {
         } catch (Exception e) {
             LOG.error("Before agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
         }
-        InvocationParameters.setCurrent(new InvocationParameters(Map.of(AgenticScope.class.getName(), agenticScope)));
+        LangChain4jManaged.setCurrent(Map.of(AgenticScope.class, agenticScope));
         Object result = internalInvoke(agent, args);
         try {
-            InvocationParameters.removeCurrent();
+            LangChain4jManaged.removeCurrent();
             afterInvocation(new AgentResponse(agenticScope, name(), args.namedArgs(), result));
         } catch (Exception e) {
             LOG.error("After agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
@@ -1,4 +1,4 @@
-package dev.langchain4j.service;
+package dev.langchain4j.invocation;
 
 import dev.langchain4j.Internal;
 
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
@@ -1,7 +1,0 @@
-package dev.langchain4j.invocation;
-
-import dev.langchain4j.Internal;
-
-@Internal
-public interface BuiltInParameter {
-}
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
@@ -112,7 +112,7 @@ public class InvocationParameters {
         setCurrentParameters(this);
     }
 
-    public static void setCurrentParameters(InvocationParameters invocationParameters) {
+    public static void setCurrent(InvocationParameters invocationParameters) {
         currentInvocationParameters.set(invocationParameters);
     }
 
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
@@ -116,7 +116,7 @@ public class InvocationParameters {
         currentInvocationParameters.set(invocationParameters);
     }
 
-    public static InvocationParameters getCurrentParameters() {
+    public static InvocationParameters current() {
         return currentInvocationParameters.get();
     }
 
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
@@ -109,7 +109,7 @@ public class InvocationParameters {
     }
 
     public void setAsCurrent() {
-        setCurrentParameters(this);
+        setCurrent(this);
     }
 
     public static void setCurrent(InvocationParameters invocationParameters) {
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
@@ -120,7 +120,7 @@ public class InvocationParameters {
         return currentInvocationParameters.get();
     }
 
-    public static void resetCurrentParameters() {
+    public static void removeCurrent() {
         currentInvocationParameters.remove();
     }
 }
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
@@ -51,8 +51,6 @@ public class InvocationParameters {
 
     private final ConcurrentHashMap<String, Object> map;
 
-    private static final ThreadLocal<InvocationParameters> currentInvocationParameters = new ThreadLocal<>();
-
     public InvocationParameters() {
         this.map = new ConcurrentHashMap<>();
     }
@@ -106,17 +104,5 @@ public class InvocationParameters {
 
     public static InvocationParameters from(Map<String, Object> map) {
         return new InvocationParameters(map);
-    }
-
-    public static void setCurrent(InvocationParameters invocationParameters) {
-        currentInvocationParameters.set(invocationParameters);
-    }
-
-    public static InvocationParameters current() {
-        return currentInvocationParameters.get();
-    }
-
-    public static void removeCurrent() {
-        currentInvocationParameters.remove();
     }
 }
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
@@ -51,6 +51,8 @@ public class InvocationParameters {
 
     private final ConcurrentHashMap<String, Object> map;
 
+    private static final ThreadLocal<InvocationParameters> currentInvocationParameters = new ThreadLocal<>();
+
     public InvocationParameters() {
         this.map = new ConcurrentHashMap<>();
     }
@@ -104,5 +106,21 @@ public class InvocationParameters {
 
     public static InvocationParameters from(Map<String, Object> map) {
         return new InvocationParameters(map);
+    }
+
+    public void setAsCurrent() {
+        setCurrentParameters(this);
+    }
+
+    public static void setCurrentParameters(InvocationParameters invocationParameters) {
+        currentInvocationParameters.set(invocationParameters);
+    }
+
+    public static InvocationParameters getCurrentParameters() {
+        return currentInvocationParameters.get();
+    }
+
+    public static void resetCurrentParameters() {
+        currentInvocationParameters.remove();
     }
 }
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
@@ -108,10 +108,6 @@ public class InvocationParameters {
         return new InvocationParameters(map);
     }
 
-    public void setAsCurrent() {
-        setCurrent(this);
-    }
-
     public static void setCurrent(InvocationParameters invocationParameters) {
         currentInvocationParameters.set(invocationParameters);
     }
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
@@ -6,7 +6,7 @@ import dev.langchain4j.Internal;
  * A marker interface for components that are managed by LangChain4j framework.
  * <p>
  * Implementing this interface indicates that the component is internally managed by LangChain4j,
- * and doesn't require to be instatiated or passed around by the user.
+ * and doesn't require to be instatiated or passed around by the user or LLM.
  *
  * @since 1.8.0
  */
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
@@ -1,5 +1,6 @@
 package dev.langchain4j.invocation;
 
+import java.util.Map;
 import dev.langchain4j.Internal;
 
 /**
@@ -12,4 +13,18 @@ import dev.langchain4j.Internal;
  */
 @Internal
 public interface LangChain4jManaged {
+
+    ThreadLocal<Map<Class<? extends LangChain4jManaged>, LangChain4jManaged>> CURRENT = new ThreadLocal<>();
+
+    static void setCurrent(Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> current) {
+        CURRENT.set(current);
+    }
+
+    static Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> current() {
+        return CURRENT.get();
+    }
+
+    static void removeCurrent() {
+        CURRENT.remove();
+    }
 }
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
@@ -1,0 +1,15 @@
+package dev.langchain4j.invocation;
+
+import dev.langchain4j.Internal;
+
+/**
+ * A marker interface for components that are managed by LangChain4j framework.
+ * <p>
+ * Implementing this interface indicates that the component is internally managed by LangChain4j,
+ * and doesn't require to be instatiated or passed around by the user.
+ *
+ * @since 1.8.0
+ */
+@Internal
+public interface LangChain4jManaged {
+}
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
@@ -19,6 +19,7 @@ import dev.langchain4j.guardrail.InputGuardrail;
 import dev.langchain4j.guardrail.OutputGuardrail;
 import dev.langchain4j.guardrail.config.InputGuardrailsConfig;
 import dev.langchain4j.guardrail.config.OutputGuardrailsConfig;
+import dev.langchain4j.invocation.InvocationParameters;
 import dev.langchain4j.memory.ChatMemory;
 import dev.langchain4j.memory.chat.ChatMemoryProvider;
 import dev.langchain4j.model.chat.ChatModel;
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
@@ -19,7 +19,6 @@ import dev.langchain4j.guardrail.InputGuardrail;
 import dev.langchain4j.guardrail.OutputGuardrail;
 import dev.langchain4j.guardrail.config.InputGuardrailsConfig;
 import dev.langchain4j.guardrail.config.OutputGuardrailsConfig;
-import dev.langchain4j.invocation.InvocationParameters;
 import dev.langchain4j.memory.ChatMemory;
 import dev.langchain4j.memory.chat.ChatMemoryProvider;
 import dev.langchain4j.model.chat.ChatModel;
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -64,6 +64,7 @@ import java.util.UUID;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.stream.Stream;
 
 @Internal
 class DefaultAiServices<T> extends AiServices<T> {
@@ -96,6 +97,10 @@ class DefaultAiServices<T> extends AiServices<T> {
                             "There can be at most one parameter of type %s", InvocationParameters.class.getName());
                 }
                 invocationParametersExist = true;
+                continue;
+            }
+
+            if (BuiltInParameter.class.isAssignableFrom(parameter.getType())) {
                 continue;
             }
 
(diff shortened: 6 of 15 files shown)


langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
```diff
@@ -105,4 +107,20 @@ public static InvocationParameters from(String key, Object value) {
...
+
+    public static void setCurrentParameters(InvocationParameters invocationParameters) {
+        currentInvocationParameters.set(invocationParameters);
+    }
+
+    public static InvocationParameters getCurrentParameters() {
+        return currentInvocationParameters.get();
+    }
+
+    public static void resetCurrentParameters() {
```
dliubarskyi:
```suggestion
    public static void removeCurrent() {
```

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
```diff
@@ -105,4 +107,20 @@ public static InvocationParameters from(String key, Object value) {
...
+
+    public void setAsCurrent() {
+        setCurrentParameters(this);
+    }
+
+    public static void setCurrentParameters(InvocationParameters invocationParameters) {
+        currentInvocationParameters.set(invocationParameters);
+    }
+
+    public static InvocationParameters getCurrentParameters() {
```
dliubarskyi:
```suggestion
    public static InvocationParameters current() {
```

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
```diff
@@ -33,8 +35,10 @@ default Object invoke(AgenticScope agenticScope, Object agent, AgentInvocationAr
         } catch (Exception e) {
             LOG.error("Before agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
         }
+        new InvocationParameters(Map.of(AgenticScope.class.getName(), agenticScope)).setAsCurrent();
```
dliubarskyi: can we please do `InvocationParameters.setCurrent(...)` instead? This one is a bit confusing.
It will also be in line with resetting below

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
```diff
@@ -105,4 +107,20 @@ public static InvocationParameters from(String key, Object value) {
     public static InvocationParameters from(Map<String, Object> map) {
         return new InvocationParameters(map);
     }
+
+    public void setAsCurrent() {
+        setCurrentParameters(this);
+    }
+
+    public static void setCurrentParameters(InvocationParameters invocationParameters) {
```
dliubarskyi:
```suggestion
    public static void setCurrent(InvocationParameters invocationParameters) {
```

langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
```diff
@@ -0,0 +1,7 @@
+package dev.langchain4j.service;
+
+import dev.langchain4j.Internal;
+
+@Internal
+public interface BuiltInParameter {
```
dliubarskyi: Is this new type necessary? I first thought that this interface will mark `InvocationContext` as well, so both `InvocationContext` and `AgenticScope` will be treated the same, but it does not seem the case. If there is no clear vision that it will be useful for multiple things, I guess adding this interface might be premature, WDYT?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -99,6 +99,10 @@ static void validateParameters(Class<?> aiServiceClass, Method method) {
                 continue;
             }
 
+            if (BuiltInParameter.class.isAssignableFrom(parameter.getType())) {
+                continue;
```
dliubarskyi, line 105: don't we need to do a similar thing inside `ToolSpecifications`?

dliubarskyi (commented):
@mariofusco thank you!

langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
mariofusco: At the moment I need the `BuiltInParameter` only here https://github.com/langchain4j/langchain4j/pull/3785/files#diff-165bbd5fb2d7e3529612a7c2964ff8553eb40ed8d5c31cbaf5ebe18998dbe13eR193 to understand that this specific argument value should be taken from the invocation parameters. I'm not sure how I could understand this if I remove that marker interface, any suggestion?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: I don't think so, where?

dliubarskyi: Inside the `parametersFrom` method, we have:
```java
for (Parameter parameter : parameters) {
            if (parameter.isAnnotationPresent(ToolMemoryId.class)
                    || InvocationParameters.class.isAssignableFrom(parameter.getType())
                    || parameter.getType() == InvocationContext.class) {
                continue;
            }
```

Shouldn't we ignore `AgenticContext` here as well? LLM should not know about it

langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
dliubarskyi: Oh right, sorry, my bad

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: Agreed, I'm also skipping `BuiltInParameter` there. (To do so I had to move it to the `langchain4j-core` module, but I believe this is fine and actually makes sense.)

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
```diff
@@ -51,6 +51,8 @@ public class InvocationParameters {
 
     private final ConcurrentHashMap<String, Object> map;
 
+    private static final ThreadLocal<InvocationParameters> currentInvocationParameters = new ThreadLocal<>();
```
dliubarskyi: Since this is a "hack" and is applied only there, perhaps worth keeping it inside the `AgentInvoker`?

langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
```diff
@@ -0,0 +1,7 @@
+package dev.langchain4j.invocation;
+
+import dev.langchain4j.Internal;
+
+@Internal
+public interface BuiltInParameter {
```
dliubarskyi: I have doubts about the name, parhaps you have other options?
Since we use this marker to:
- check whether LLM should know about the parameter (e.g., when creating tool spec)
- check whether to inject appropriate object (e.g., when preparing arguments for calling a tool method)
Perhaps something like "ModelAgnostic" or "LibraryManaged" or similar? e.g., `public interface AgenticScope implements ModelAgnostic`

Also perhaps worth adding a (small) javadoc explaining why is it needed

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -452,19 +458,27 @@ public Object invoke(Method method, Object[] args, InvocationContext invocationC
                         return actualResponse;
                     }
 
-                    private Optional<InvocationParameters> findInvocationParams(Object[] args, Parameter[] params) {
-                        if (args == null) {
-                            return Optional.empty();
-                        }
+                    private void setupBuiltInParameters(InvocationParameters invocationParameters, Object[] args, Parameter[] params) {
```
dliubarskyi: Is this logic still relevant? There is no reason to have `AgenticScope` as a method parameter any more, right?

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
mariofusco: I'm afraid this is not possible since the `ThreadLocal` has to be accessed by the `DefaultAiServices`, see https://github.com/langchain4j/langchain4j/pull/3785/files#diff-1d724bf189ebfdb3d76eaefc0f9f1768b01e615ae3521ade0ea4fe4c8d774bd2R481

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: Yes, you're right, I also had doubt about this, but for now this is not necessary, I will remove it.

langchain4j-core/src/main/java/dev/langchain4j/invocation/BuiltInParameter.java
mariofusco: Maybe `LangChain4jManaged`? I will rename it so and add some javadocs. Let me know if ok for you.

dliubarskyi: Sounds good, thanks!

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
dliubarskyi: Not anymore, right?

mariofusco: Why not? Yes, it's used in the `findInvocationParams` method.

langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.invocation;
+
+import dev.langchain4j.Internal;
+
+/**
+ * A marker interface for components that are managed by LangChain4j framework.
+ * <p>
+ * Implementing this interface indicates that the component is internally managed by LangChain4j,
+ * and doesn't require to be instatiated or passed around by the user.
```
dliubarskyi:
```suggestion
 * and doesn't require to be instatiated or passed around by the user or LLM.
```

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -453,18 +458,17 @@ public Object invoke(Method method, Object[] args, InvocationContext invocationC
...
+                                Parameter parameter = params[i];
+                                if (InvocationParameters.class.isAssignableFrom(parameter.getType())) {
+                                    InvocationParameters invocationParameters = (InvocationParameters) args[i];
+                                    ensureNotNull(invocationParameters, "InvocationParameters");
+                                    return Optional.of(invocationParameters);
+                                }
                             }
                         }
-                        return Optional.empty();
+                        return Optional.ofNullable(InvocationParameters.current());
```
dliubarskyi: This makes it impossible to combine both
1. passing `InvocationParameters` to the `@Tool` method (e.g., user wants to pass some custom value to the tool when calling an Agent)
2. accessing `AgenticSope` in the `@Tool` method (e.g., user wants to pass some value from one agent to another)
I am also not sure if `AgenticSope` and `InvocationParameters` should depend on one another, WDYT?

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -19,6 +19,7 @@
 import dev.langchain4j.guardrail.OutputGuardrail;
 import dev.langchain4j.guardrail.config.InputGuardrailsConfig;
 import dev.langchain4j.guardrail.config.OutputGuardrailsConfig;
+import dev.langchain4j.invocation.InvocationParameters;
```
```suggestion
```

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: Good point, I'm now combining the content of the 2 if both are present. 

About the `AgenticScope` depending on the `InvocationParameters`, I don't think this is necessary and I don't understand why it should, but maybe I'm missing something?

dliubarskyi: TBH I am not sure, but there is something off with keeping `AgenticSope` inside the `InvocationParameters` by string key.
Moreover, users can see and manipulate it (from within tools and RAG pipeline), but it is supposed to be "LC4j internals" as far as I understand.

WDYT about this approach? I have extracted LC4j managed parameters from `InvocationParameters` into `InvocationContext`. Attaching a patch because I cannot push in your repo:

```patch
Subject: [PATCH] decouple InvocationParameters and AgenticScope
---
Index: langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -7,8 +7,7 @@
 import dev.langchain4j.agentic.agent.MissingArgumentException;
 import dev.langchain4j.agentic.scope.AgenticScope;
 import dev.langchain4j.agentic.UntypedAgent;
-import dev.langchain4j.agentic.scope.DefaultAgenticScope;
-import dev.langchain4j.invocation.InvocationParameters;
+import dev.langchain4j.invocation.LangChain4jManaged;
 import dev.langchain4j.service.V;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -35,10 +34,10 @@
         } catch (Exception e) {
             LOG.error("Before agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
         }
-        InvocationParameters.setCurrent(new InvocationParameters(Map.of(AgenticScope.class.getName(), agenticScope)));
+        LangChain4jManaged.setCurrent(Map.of(AgenticScope.class, agenticScope));
         Object result = internalInvoke(agent, args);
         try {
-            InvocationParameters.removeCurrent();
+            LangChain4jManaged.removeCurrent();
             afterInvocation(new AgentResponse(agenticScope, name(), args.namedArgs(), result));
         } catch (Exception e) {
             LOG.error("After agent invocation listener for agent " + name() + " failed: " + e.getMessage(), e);
Index: langchain4j-core/src/main/java/dev/langchain4j/invocation/DefaultInvocationContext.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j-core/src/main/java/dev/langchain4j/invocation/DefaultInvocationContext.java b/langchain4j-core/src/main/java/dev/langchain4j/invocation/DefaultInvocationContext.java
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/DefaultInvocationContext.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/DefaultInvocationContext.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -3,6 +3,7 @@
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 
@@ -17,6 +18,7 @@
     private final List<Object> methodArguments = new ArrayList<>();
     private final Object chatMemoryId;
     private final InvocationParameters invocationParameters;
+    private final Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> managedParameters;
     private final Instant timestamp;
 
     public DefaultInvocationContext(InvocationContext.Builder builder) {
@@ -26,6 +28,7 @@
         this.methodArguments.addAll(builder.methodArguments());
         this.chatMemoryId = builder.chatMemoryId();
         this.invocationParameters = builder.invocationParameters();
+        this.managedParameters = builder.managedParameters();
         this.timestamp = builder.timestamp();
     }
 
@@ -59,6 +62,11 @@
         return invocationParameters;
     }
 
+    @Override
+    public Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> managedParameters() {
+        return managedParameters;
+    }
+
     @Override
     public Instant timestamp() {
         return timestamp;
@@ -75,6 +83,7 @@
                 && Objects.equals(methodArguments, that.methodArguments)
                 && Objects.equals(chatMemoryId, that.chatMemoryId)
                 && Objects.equals(invocationParameters, that.invocationParameters)
+                && Objects.equals(managedParameters, that.managedParameters)
                 && Objects.equals(timestamp, that.timestamp);
     }
 
@@ -87,18 +96,21 @@
                 methodArguments,
                 chatMemoryId,
                 invocationParameters,
+                managedParameters,
                 timestamp);
     }
 
     @Override
     public String toString() {
-        return "DefaultInvocationContext{" + "invocationId="
-                + invocationId + ", interfaceName='"
-                + interfaceName + '\'' + ", methodName='"
-                + methodName + '\'' + ", methodArguments="
-                + methodArguments + ", chatMemoryId="
-                + chatMemoryId + ", invocationParameters="
-                + invocationParameters + ", timestamp="
-                + timestamp + '}';
+        return "DefaultInvocationContext{" +
+                "invocationId=" + invocationId +
+                ", interfaceName='" + interfaceName + '\'' +
+                ", methodName='" + methodName + '\'' +
+                ", methodArguments=" + methodArguments +
+                ", chatMemoryId=" + chatMemoryId +
+                ", invocationParameters=" + invocationParameters +
+                ", managedParameters=" + managedParameters +
+                ", timestamp=" + timestamp +
+                '}';
     }
 }
Index: langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -4,6 +4,7 @@
 import java.time.Instant;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 import org.jspecify.annotations.NonNull;
 
@@ -49,6 +50,14 @@
      */
     InvocationParameters invocationParameters();
 
+    /**
+     * LangChain4j managed parameters
+     * @since 1.8.0
+     */
+    default Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> managedParameters() {
+        return Map.of();
+    }
+
     /**
      * Retrieves the point in time when the invocation occurred.
      */
@@ -81,6 +90,7 @@
         private List<@NonNull Object> methodArguments = new ArrayList<>();
         private Object chatMemoryId;
         private InvocationParameters invocationParameters;
+        private Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> managedParameters;
         private Instant timestamp;
 
         protected Builder() {}
@@ -92,6 +102,7 @@
             methodArguments(invocationContext.methodArguments());
             chatMemoryId(invocationContext.chatMemoryId());
             invocationParameters(invocationContext.invocationParameters());
+            managedParameters(invocationContext.managedParameters());
             timestamp(invocationContext.timestamp());
         }
 
@@ -159,6 +170,14 @@
             return this;
         }
 
+        /**
+         * Sets the LC4j managed parameters for the builder.
+         */
+        public Builder managedParameters(Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> managedParameters) {
+            this.managedParameters = managedParameters;
+            return this;
+        }
+
         /**
          * Sets the timestamp for the builder.
          */
@@ -207,6 +226,10 @@
             return invocationParameters;
         }
 
+        public Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> managedParameters() {
+            return managedParameters;
+        }
+
         public Instant timestamp() {
             return timestamp;
         }
Index: langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationParameters.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -51,8 +51,6 @@
 
     private final ConcurrentHashMap<String, Object> map;
 
-    private static final ThreadLocal<InvocationParameters> currentInvocationParameters = new ThreadLocal<>();
-
     public InvocationParameters() {
         this.map = new ConcurrentHashMap<>();
     }
@@ -107,16 +105,4 @@
     public static InvocationParameters from(Map<String, Object> map) {
         return new InvocationParameters(map);
     }
-
-    public static void setCurrent(InvocationParameters invocationParameters) {
-        currentInvocationParameters.set(invocationParameters);
-    }
-
-    public static InvocationParameters current() {
-        return currentInvocationParameters.get();
-    }
-
-    public static void removeCurrent() {
-        currentInvocationParameters.remove();
-    }
 }
Index: langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java b/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java
--- a/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j-core/src/main/java/dev/langchain4j/invocation/LangChain4jManaged.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -1,5 +1,6 @@
 package dev.langchain4j.invocation;
 
+import java.util.Map;
 import dev.langchain4j.Internal;
 
 /**
@@ -12,4 +13,18 @@
  */
 @Internal
 public interface LangChain4jManaged {
+
+    ThreadLocal<Map<Class<? extends LangChain4jManaged>, LangChain4jManaged>> CURRENT = new ThreadLocal<>();
+
+    static void setCurrent(Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> current) {
+        CURRENT.set(current);
+    }
+
+    static Map<Class<? extends LangChain4jManaged>, LangChain4jManaged> current() {
+        return CURRENT.get();
+    }
+
+    static void removeCurrent() {
+        CURRENT.remove();
+    }
 }
Index: langchain4j-core/src/test/java/dev/langchain4j/rag/query/MetadataTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j-core/src/test/java/dev/langchain4j/rag/query/MetadataTest.java b/langchain4j-core/src/test/java/dev/langchain4j/rag/query/MetadataTest.java
--- a/langchain4j-core/src/test/java/dev/langchain4j/rag/query/MetadataTest.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j-core/src/test/java/dev/langchain4j/rag/query/MetadataTest.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -72,7 +72,7 @@
         // then
         assertThat(toString)
                 .isEqualToIgnoringWhitespace("""
-                        Metadata { chatMessage = UserMessage { name = null contents = [TextContent { text = "user message" }] }, chatMemory = [UserMessage { name = null contents = [TextContent { text = "Hello" }] }, AiMessage { text = "Hi, how can I help you today?", thinking = null, toolExecutionRequests = [], attributes = {} }], invocationContext = DefaultInvocationContext{invocationId=null, interfaceName='null', methodName='null', methodArguments=[], chatMemoryId=42, invocationParameters=null, timestamp=null} }
+                        Metadata { chatMessage = UserMessage { name = null contents = [TextContent { text = "user message" }] }, chatMemory = [UserMessage { name = null contents = [TextContent { text = "Hello" }] }, AiMessage { text = "Hi, how can I help you today?", thinking = null, toolExecutionRequests = [], attributes = {} }], invocationContext = DefaultInvocationContext{invocationId=null, interfaceName='null', methodName='null', methodArguments=[], chatMemoryId=42, invocationParameters=null, managedParameters=null, timestamp=null} }
                         """);
     }
 
Index: langchain4j-core/src/test/java/dev/langchain4j/rag/query/QueryTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j-core/src/test/java/dev/langchain4j/rag/query/QueryTest.java b/langchain4j-core/src/test/java/dev/langchain4j/rag/query/QueryTest.java
--- a/langchain4j-core/src/test/java/dev/langchain4j/rag/query/QueryTest.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j-core/src/test/java/dev/langchain4j/rag/query/QueryTest.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -87,7 +87,7 @@
         // then
         assertThat(toString)
                 .isEqualToIgnoringWhitespace("""
-                        Query { text = "query", metadata = Metadata { chatMessage = UserMessage { name = null contents = [TextContent { text = "user message" }] }, chatMemory = [UserMessage { name = null contents = [TextContent { text = "Hello" }] }, AiMessage { text = "Hi, how can I help you today?", thinking = null, toolExecutionRequests = [], attributes = {} }], invocationContext = DefaultInvocationContext{invocationId=null, interfaceName='null', methodName='null', methodArguments=[], chatMemoryId=42, invocationParameters=null, timestamp=null} } }
+                        Query { text = "query", metadata = Metadata { chatMessage = UserMessage { name = null contents = [TextContent { text = "user message" }] }, chatMemory = [UserMessage { name = null contents = [TextContent { text = "Hello" }] }, AiMessage { text = "Hi, how can I help you today?", thinking = null, toolExecutionRequests = [], attributes = {} }], invocationContext = DefaultInvocationContext{invocationId=null, interfaceName='null', methodName='null', methodArguments=[], chatMemoryId=42, invocationParameters=null, managedParameters=null, timestamp=null} } }
                         """);
     }
 
Index: langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -228,6 +228,7 @@
                                 .methodArguments(args != null ? Arrays.asList(args) : List.of())
                                 .chatMemoryId(findMemoryId(method, args).orElse(ChatMemoryService.DEFAULT))
                                 .invocationParameters(invocationParameters)
+                                .managedParameters(LangChain4jManaged.current())
                                 .timestampNow()
                                 .build();
                         try {
@@ -459,24 +460,18 @@
                     }
 
                     private Optional<InvocationParameters> findInvocationParams(Object[] args, Parameter[] params) {
-                        InvocationParameters currentParams = InvocationParameters.current();
-                        if (args != null) {
-                            for (int i = 0; i < params.length; i++) {
-                                Parameter parameter = params[i];
-                                if (InvocationParameters.class.isAssignableFrom(parameter.getType())) {
-                                    InvocationParameters invocationParameters = (InvocationParameters) args[i];
-                                    ensureNotNull(invocationParameters, "InvocationParameters");
-                                    if (currentParams == null) {
-                                        return Optional.of(invocationParameters);
-                                    } else {
-                                        HashMap<String, Object> newParamsMap = new HashMap<>(currentParams.asMap());
-                                        newParamsMap.putAll(invocationParameters.asMap());
-                                        return Optional.of(new InvocationParameters(newParamsMap));
-                                    }
-                                }
+                        if (args == null) {
+                            return Optional.empty();
+                        }
+                        for (int i = 0; i < params.length; i++) {
+                            Parameter parameter = params[i];
+                            if (InvocationParameters.class.isAssignableFrom(parameter.getType())) {
+                                InvocationParameters invocationParameters = (InvocationParameters) args[i];
+                                ensureNotNull(invocationParameters, "InvocationParameters");
+                                return Optional.of(invocationParameters);
                             }
                         }
-                        return Optional.ofNullable(currentParams);
+                        return Optional.empty();
                     }
 
                     private boolean canAdaptTokenStreamTo(Type returnType) {
Index: langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java b/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java	(revision ba73ebaeb6d805c787255c1c2f4d7f315425df03)
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java	(revision 469dfafe9de63b9f64c8c13ffe337708e719b2a6)
@@ -191,7 +191,7 @@
             }
 
             if (LangChain4jManaged.class.isAssignableFrom(parameter.getType())) {
-                arguments[i] = context.invocationParameters().get(parameter.getType().getName());
+                arguments[i] = context.managedParameters().get(parameter.getType());
                 continue;
             }
 

```

mariofusco: I was striving to fit this new feature in the `InvocationParameters`, but yes I agree that keeping them separated is much cleaner, great job, thanks a lot :pray: 

I merged all the changes that you suggested in this pull request, please give it one more look.

dliubarskyi (approved):
@mariofusco thank you!

czyzowskibartosz:
Is it possible to access it from https://docs.langchain4j.dev/tutorials/tools/#specifying-tools-programmatically?

PR #3791, opened 2025-09-30
test: enhance test coverage for IsEqualTo filter and Tablestore utilities

alxkm:
This PR enhances test coverage for the `IsEqualTo` filter class and `Tablestore` embedding utilities, focusing on edge cases, error conditions, and type conversion scenarios.

**Changes**
Added 7 new tests to `IsEqualToTest`:

- Value comparison: Tests inequality detection between different values
- Type conversion: Verifies number type conversion (Long to Integer)
- Case sensitivity: Ensures string comparisons are case-sensitive
- Null handling: Tests behavior with null objects and null comparison values
- Special characters: Validates keys containing dots work correctly
- Floating-point: Tests decimal number comparisons

Added 4 new tests to `TablestoreEmbeddingStoreTest`:

- Empty arrays: Tests serialization/deserialization of empty embeddings
- Null validation: Ensures proper exception handling for null inputs
- Round-trip consistency: Validates empty array parsing

**Impact**
The existing tests missed critical edge cases that could lead to runtime errors. These additions ensure proper null handling, validate type conversions work as expected, and document the API's behavior for edge cases like empty arrays and special characters.

### DIFF
--- a/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/filter/comparison/IsEqualToTest.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/store/embedding/filter/comparison/IsEqualToTest.java
@@ -1,6 +1,7 @@
 package dev.langchain4j.store.embedding.filter.comparison;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@alxkm thank you!

PR #3799, opened 2025-10-02
Support tool name mapping for MCP clients

### LINKED ISSUE #3619: Tool Calls identity
How can I match a tool call request from LLM with the available tools of my MCP? The Tool execution request contains an ID field, the name of the tool and the message, but I'm not finding the field ID in the MCP client tool specifications, only the name. This ID is the ID of the request or of the tool? Because maybe if a have a loot of tools available from different MCPs, can in some situation MCPs contains tools with the same name, causing a conflict about how to decide the correct tool to call, what is the best way to handle it? Or its a situation that we need to add to the project?

jmartisk:
Fixes https://github.com/langchain4j/langchain4j/issues/3619

### DIFF
--- a/docs/docs/tutorials/mcp.md
+++ b/docs/docs/tutorials/mcp.md
@@ -183,6 +183,26 @@ Bot bot = AiServices.builder(Bot.class)
 
 More information on tool support in LangChain4j can be found [here](/tutorials/tools).
 
+### MCP Tool name mapping
+
+If you use multiple MCP servers, and they expose tools with clashing names (or you simply want to
+adjust an inappropriately chosen name), it may be useful to apply a tool name mapping function.
+This can be done by providing a `BiFunction<McpClient, ToolSpecification, String>` to the MCP tool provider.
+
+For example":
+```java
+McpToolProvider toolProvider = McpToolProvider.builder()
+        .mcpClients(mcpClient1, mcpClient2)
+        .toolNameMapper((client, toolSpec) -> {
+            // Prefix all tool names with the name of the MCP client and an underscore
+            return client.key() + "_" + toolSpec.name();
+        })
+        .build();
+```
+
+After this, the `ToolSpecification` objects returned by the tool provider will contain the mapped (logical) names,
+but the generated `ToolExecutor` objects will be fixed to pass the original (physical) names to the server when invoking the tools.
+
 ## Logging
 
 The MCP protocol also defines a way for the server to send log messages to
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiFunction;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import org.slf4j.Logger;
@@ -31,6 +32,8 @@ public class McpToolProvider implements ToolProvider {
     private final Function<ToolExecutor, ToolExecutor> toolWrapper;
     private static final Logger log = LoggerFactory.getLogger(McpToolProvider.class);
     private final McpResourcesAsToolsPresenter resourcesAsToolsPresenter;
+    private final AtomicReference<BiFunction<McpClient, ToolSpecification, String>> toolNameMapper;
+    private final AtomicReference<BiFunction<McpClient, ToolSpecification, ToolSpecification>> toolSpecificationMapper;
 
     private McpToolProvider(Builder builder) {
         this(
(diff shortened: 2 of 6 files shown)


dliubarskyi (commented):
@jmartisk thank you! Don't we want to do this on the `McpToolProvider` level? Similarly to how we can filter which tools from which MCP servers are exposed (`McpToolProvider.addFilter()`/`McpToolProvider.setFilter()`/`McpToolProvider.Builder.filter()`/`McpToolProvider.Builder.filterToolNames()`), we could transform (prefix is nice, but we could allow making any type of renaming, with the same implementation price) the name of the tool in the same place (`McpToolProvider`). WDYT?

jmartisk:
> @jmartisk thank you! Don't we want to do this on the `McpToolProvider` level? Similarly to how we can filter which tools from which MCP servers are exposed (`McpToolProvider.addFilter()`/`McpToolProvider.setFilter()`/`McpToolProvider.Builder.filter()`/`McpToolProvider.Builder.filterToolNames()`), we could transform (prefix is nice, but we could allow making any type of renaming, with the same implementation price) the name of the tool in the same place (`McpToolProvider`). WDYT?

Done - now it's on the tool provider level and allows any arbitrary mapping of the tool name, please check

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpToolsTestBase.java
```diff
@@ -121,21 +116,6 @@ public void executeToolWithWrongArgumentType() {
                 .hasFieldOrPropertyWithValue("errorCode", -32602);
     }
 
-    @Test
-    public void executeNonExistentTool() {
```
jmartisk, line 125: This test now doesn't make sense anymore because the tool executor obtained by `toolProviderResult.toolExecutorByName("echoString")` is bound to only call the `echoString` tool.

docs/docs/tutorials/mcp.md
```diff
@@ -183,6 +183,26 @@ Bot bot = AiServices.builder(Bot.class)
 
 More information on tool support in LangChain4j can be found [here](/tutorials/tools).
 
+### MCP Tool name mapping
+
+If you use multiple MCP servers, and they expose tools with clashing names (or you simply want to
+adjust an inappropriately chosen name), it may be useful to apply a tool name mapping function.
+This can be done by providing a `BiFunction<McpClient, ToolSpecification, String>` to the MCP tool provider.
```
dliubarskyi, line 190: I am wondering if it is worth having a `BiFunction<McpClient, ToolSpecification, ToolSpecification>` instead, so that not only the `name` of the tool can be adjusted, but also the `description`, and potentially some new fields in the future. WDYT?

dliubarskyi: If there are multiple tools with the same name from different MCP servers, chances are that they do very similar things and the description of those tools might need to be adjusted to clarify when/how each of them should be used.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -215,6 +242,15 @@ public McpToolProvider.Builder resourcesAsToolsPresenter(
             return this;
         }
 
+        /**
+         * Defines a mapping function to customize the tool names as they are registered in the tool provider.
+         * By default, the tool names are used as-is.
+         */
+        public McpToolProvider.Builder toolNameMapper(BiFunction<McpClient, ToolSpecification, String> toolNameMapper) {
```
dliubarskyi, line 277:
```suggestion
        public McpToolProvider.Builder toolSpecificationTransformer(BiFunction<McpClient, ToolSpecification, ToolSpecification> toolNameMapper) {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
```diff
@@ -31,27 +32,31 @@ public class McpToolProvider implements ToolProvider {
...
     }
 
     protected McpToolProvider(
             List<McpClient> mcpClients,
             boolean failIfOneServerFails,
             BiPredicate<McpClient, ToolSpecification> mcpToolsFilter,
             Function<ToolExecutor, ToolExecutor> toolWrapper,
-            McpResourcesAsToolsPresenter resourcesAsToolsPresenter) {
+            McpResourcesAsToolsPresenter resourcesAsToolsPresenter,
+            BiFunction<McpClient, ToolSpecification, String> toolNameMapper) {
```
dliubarskyi: This is a breaking change, please create a new ctor (that accepts builder) and deprecate this one

dliubarskyi: Ah, this module is still in beta, feel free to ignore

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/McpToolProvider.java
dliubarskyi: We can ofc have a separate `toolNameTransformer` for convenience, but it can use the `toolSpecificationTransformer` under the hood

dliubarskyi (commented):
@jmartisk thank you!

docs/docs/tutorials/mcp.md
jmartisk: Yeah I've been thinking about this too but I wasn't sure.
One minor issue is that if we require the user to create an instance of `ToolSpecification`, they also have to include the (unchanged) parameters from the original specification, and it feels a bit odd to specify that they have to do this (somehow obvious and seemingly unnecessary thing) and must not forget.

jmartisk: DONE - there is now a name mapper and a specification mapper

dliubarskyi (approved):
@jmartisk thank you!

PR #3803, opened 2025-10-02
Allow configuring the logger for logging MCP events

loicmathieu:
## Issue
No issue.

## Change
Like what was done for language model provides, allow configuring the logger used to log MCP server events.

## General checklist

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
@@ -30,11 +30,13 @@ import org.slf4j.LoggerFactory;
 
 public class StreamableHttpMcpTransport implements McpTransport {
 
-    private static final Logger log = LoggerFactory.getLogger(StreamableHttpMcpTransport.class);
+    private static final Logger DEFAULT_TRAFFIC_LOG = LoggerFactory.getLogger("MCP");
+    private static final Logger LOG = LoggerFactory.getLogger(StreamableHttpMcpTransport.class);
     private final String url;
     private final Map<String, String> customHeaders;
     private final boolean logResponses;
     private final boolean logRequests;
+    private final Logger trafficLog;
     static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
     private volatile McpOperationHandler operationHandler;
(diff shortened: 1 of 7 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
```diff
@@ -235,6 +238,15 @@ public StreamableHttpMcpTransport.Builder logResponses(boolean logResponses) {
             return this;
         }
 
+        /**
+         * @param logger an alternate {@link Logger} to be used instead of the default one provided by Langchain4J for logging events.
```
jmartisk: I think we should rather use the word "traffic" than events.
Yes I know, in the `stdio` transport, we use "events" to refer to incoming traffic, that wasn't a good choice from my side.

"Logging events" could be easily mistaken for handling MCP protocol-level log events (that the MCP server sends to the client over the protocol, and these are handled separately). "Traffic" more clearly refers to the actual exchange of protocol messages

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
```diff
@@ -132,7 +134,7 @@ public BodySubscriber<Void> apply(HttpResponse.ResponseInfo responseInfo) {
                             Optional<String> mcpSessionId =
                                     responseInfo.headers().firstValue("Mcp-Session-Id");
                             if (mcpSessionId.isPresent()) {
-                                log.debug("Assigned MCP session ID: " + mcpSessionId);
+                                log.debug("Assigned MCP session ID: {}", mcpSessionId);
```
jmartisk: This logging statement is not related to traffic. Perhaps, should this be a separate logger than the one that logs raw requests and responses?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
loicmathieu: Should I change it everywhere or only on that transport?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
loicmathieu: I use a separate logger for this log event.
I take the opportunity to use the MCP logger for traffic log as for other transport.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
loicmathieu: I changed it everywhere to be consistent.

jmartisk (approved):

PR #3820, opened 2025-10-06
Fix guardrails result after a successful reprompt

mariofusco:
## Issue

Closes #

## Change

If an output guardrail returns a `success()` after a first reprompt, the returned result is the original one, the one for which the reprompting was requested, and not the one actually validated and successful after the reprompt. This pull request fixes this issue returning the result generated after the reprompting.

/cc @edeandrea

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/guardrail/OutputGuardrailExecutor.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/guardrail/OutputGuardrailExecutor.java
@@ -1,5 +1,6 @@
 package dev.langchain4j.guardrail;
 
+import static dev.langchain4j.guardrail.OutputGuardrailResult.successWith;
 import static dev.langchain4j.observability.api.event.OutputGuardrailExecutedEvent.OutputGuardrailExecutedEventBuilder;
 
 import dev.langchain4j.data.message.UserMessage;
@@ -63,7 +64,7 @@ public non-sealed class OutputGuardrailExecutor
         }
 
         while (attempt < maxAttempts) {
-            result = executeGuardrails(accumulatedRequest);
+            result = rewriteResult(request, accumulatedRequest, executeGuardrails(accumulatedRequest));
 
             if (result.isSuccess()) {
                 return result;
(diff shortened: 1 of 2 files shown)


langchain4j-core/src/main/java/dev/langchain4j/guardrail/OutputGuardrailExecutor.java
```diff
@@ -106,6 +107,19 @@ public OutputGuardrailResult execute(OutputGuardrailRequest request) {
...
     }
 
+    private OutputGuardrailResult rewriteResult(OutputGuardrailRequest originalRequest, OutputGuardrailRequest validatedRequest, OutputGuardrailResult result) {
+        if (result.isSuccess() && !result.hasRewrittenResult()) {
+            String originalText = originalRequest.responseFromLLM().aiMessage().text();
+            String validatedText = validatedRequest.responseFromLLM().aiMessage().text();
+            if (!originalText.equals(validatedText)) {
+                // The text validated by the output guardrail is different form the original one because of a
+                // successful reprompt, so we need to create a new success result with the new text
+                return successWith(validatedText);
```
dliubarskyi, line 117: `AiMessage` can contain not only `text`, but also `thinking` and `attributes` (that can contain provider-specific thinking signatures)

dliubarskyi (commented):
@mariofusco thank you!

langchain4j-core/src/main/java/dev/langchain4j/guardrail/OutputGuardrailExecutor.java
mariofusco: I see your point, but in this case we also have the same problem when we generate the `AiMessage` from the rewritten response here https://github.com/langchain4j/langchain4j/blob/c9d44b336d4270acd4d0107416318d0593587eef/langchain4j-core/src/main/java/dev/langchain4j/guardrail/OutputGuardrailResult.java#L174 and in all honesty I don't know how to fix or improve this. Any suggestion?

dliubarskyi: Perhaps we could introduce a `AiMessage.toBuilder()` method that you could use to override only text and keep the rest of contents intact?

mariofusco: Yes, but it is not only that: the problem is that at the moment the `successWith()` method of the output guardrail takes a `String` while, more in general, it should probably have an overload taking an `AiMessage instead`. If you agree I could work on this improvement. If so do you prefer that I do this in this same PR or would it be better to bring this fix in as it is (I found this problem working on an example for our Devoxx workshop) and then work on this in a different PR?

dliubarskyi: @mariofusco yes, makes total sense to have `successWith(AiMessage)`! Regarding where/how to do it: it's up to you

mariofusco: I open and assigned to myself a feature request for this change https://github.com/langchain4j/langchain4j/issues/3830

If you don't have any further objection or remark please merge this PR.

dliubarskyi (approved):
@mariofusco thank you!

PR #3833, opened 2025-10-07
feat: Added support to auto reinitialize session if not found

### LINKED ISSUE #3802: [BUG] StreamableHttpMcpTransport doesn't renew a lost Mcp-Session-Id
**Describe the bug**

LangChain4j now has the `StreamableHttpMcpTransport` transport for streamable-http MCP protocol (https://github.com/langchain4j/langchain4j/pull/3534). 

It supports the `Mcp-Session-Id` header that tracks the MCP session, but it doesn't handle the case where the server loses/terminates the session.

**Log and Stack trace**

```
java.util.concurrent.ExecutionException: java.lang.RuntimeException: Unexpected status code: 404
        at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2096) ~[na:na]
        at dev.langchain4j.mcp.client.DefaultMcpClient.checkHealth(DefaultMcpClient.java:329) ~[langchain4j-mcp-1.6.0-beta12.jar:na]
        ... 82 common frames omitted
Caused by: java.lang.RuntimeException: Unexpected status code: 404
        at dev.langchain4j.mcp.client.transport.http.StreamableHttpMcpTransport$1.apply(StreamableHttpMcpTransport.java:127) ~[langchain4j-mcp-1.6.0-beta12.jar:na]
        at java.net.http/jdk.internal.net.http.Http1Exchange.createResponseSubscriber(Http1Exchange.java:460) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.Http1Exchange.readBodyAsync(Http1Exchange.java:451) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.Exchange.readBodyAsync(Exchange.java:225) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.MultiExchange.lambda$responseAsync0$4(MultiExchange.java:367) ~[java.net.http:na]
        at java.base/java.util.concurrent.CompletableFuture$UniCompose.tryFire(CompletableFuture.java:1150) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture.postFire(CompletableFuture.java:614) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:653) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:482) ~[na:na]
        at java.net.http/jdk.internal.net.http.HttpClientImpl$DelegatingExecutor.execute(HttpClientImpl.java:177) ~[java.net.http:na]
        at java.base/java.util.concurrent.CompletableFuture$UniCompletion.claim(CompletableFuture.java:572) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:642) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510) ~[na:na]
        at java.base/java.util.concurrent.CompletableFuture.complete(CompletableFuture.java:2179) ~[na:na]
        at java.net.http/jdk.internal.net.http.Http1Response$HeadersReader.handle(Http1Response.java:610) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.Http1Response$HeadersReader.handle(Http1Response.java:536) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.Http1Response$Receiver.accept(Http1Response.java:527) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.Http1Response$HeadersReader.tryAsyncReceive(Http1Response.java:583) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.Http1AsyncReceiver.flush(Http1AsyncReceiver.java:233) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.common.SequentialScheduler$LockingRestartableTask.run(SequentialScheduler.java:182) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.common.SequentialScheduler$CompleteRestartableTask.run(SequentialScheduler.java:149) ~[java.net.http:na]
        at java.net.http/jdk.internal.net.http.common.SequentialScheduler$SchedulableTask.run(SequentialScheduler.java:207) ~[java.net.http:na]
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144) ~[na:na]
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642) ~[na:na]
        at java.base/java.lang.Thread.run(Thread.java:1583) ~[na:na]
```

**To Reproduce**

1. Initialize a `DefaultMcpClient` with a `StreamableHttpMcpTransport`.
2. Use it to connect to an MCP server that handles MCP sessions (e.g. one based on Python-SDK: https://pypi.org/project/mcp/). 
3. A simple health check should suffice to spawn a new session. LangChain4j will now start sending `Mcp-Session-Id` in the headers.
4. Make the server lose the session (e.g. by restarting the mentioned server).
5. Try to contact the MCP server via the client again, e.g. run another health check.

**Current behavior**
A `RuntimeException` is thrown. It's caused by HTTP reply with an error code 4xx.

`StreamableHttpMcpTransport.java`:
```java
if (!isExpectedStatusCode(responseInfo.statusCode())) {
    future.completeExceptionally(
            new RuntimeException("Unexpected status code: " + responseInfo.statusCode()));
    return null;
}
```
```java
private boolean isExpectedStatusCode(int statusCode) {
    return statusCode >= 200 && statusCode < 300;
}
```

**Expected behavior**
Some type of handling for this case should be provided. The spec's point 2.5. [1] says:

> 3. The server **MAY** terminate the session at any time, after which it **MUST** respond to requests containing that session ID with HTTP 404 Not Found.
> 4. When a client receives HTTP 404 in response to a request containing an `Mcp-Session-Id`, it **MUST** start a new session by sending a new `InitializeRequest` without a session ID attached.

Whether this should happen automatically or with some intervention from the library's user -- I'm not sure.

Also, the lib should not just throw `RuntimeException` under any circumstance, please very much.

- LangChain4j version: 1.6.0

**Additional context:**
There's currently a separate bug in the Python SDK where it returns "400 Bad Request" whereas the spec says "404 Not Found" should be returned (https://github.com/modelcontextprotocol/python-sdk/issues/1004). I pushed a PR to fix that (https://github.com/modelcontextprotocol/python-sdk/pull/1417).

**Reference:**
1. https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management

vijaypatidar:
## Issue
Closes #3802

## Change
This PR adds support to automatically reinitialize the session with the MCP server if it is not found (e.g., after a server restart). This ensures that clients can recover gracefully without manual intervention or a full restart.

## General checklist

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
@@ -37,9 +37,10 @@ public class StreamableHttpMcpTransport implements McpTransport {
     private final boolean logRequests;
     private final Logger trafficLog;
     static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
-
+    private final AtomicReference<CompletableFuture<JsonNode>> initializeInProgress = new AtomicReference<>(null);
     private volatile McpOperationHandler operationHandler;
     private final HttpClient httpClient;
+    private McpInitializeRequest initializeRequest;
     private final AtomicReference<String> mcpSessionId = new AtomicReference<>();
 
     public StreamableHttpMcpTransport(StreamableHttpMcpTransport.Builder builder) {
@@ -63,7 +64,14 @@ public class StreamableHttpMcpTransport implements McpTransport {
 
     @Override
     public CompletableFuture<JsonNode> initialize(McpInitializeRequest operation) {
-        return execute(operation, operation.getId())
+        this.initializeRequest = operation;
+        CompletableFuture<JsonNode> completableFuture = execute(operation, operation.getId());
+        initializeInProgress.set(completableFuture);
+        return completableFuture
+                .thenCompose(originalResponse -> {
+                    initializeInProgress.set(null);
+                    return CompletableFuture.completedFuture(originalResponse);
+                })
                 .thenCompose(originalResponse -> execute(new McpInitializationNotification(), null)
                         .thenCompose(nullNode -> CompletableFuture.completedFuture(originalResponse)));
     }
@@ -76,7 +84,7 @@ public class StreamableHttpMcpTransport implements McpTransport {
         }
         final HttpRequest.Builder builder = HttpRequest.newBuilder();
         String sessionId = mcpSessionId.get();
-        if (sessionId != null) {
+        if (sessionId != null && !(message instanceof McpInitializeRequest)) {
             builder.header("Mcp-Session-Id", sessionId);
         }
         customHeaders.forEach(builder::header);
@@ -108,6 +116,14 @@ public class StreamableHttpMcpTransport implements McpTransport {
     }
 
     private CompletableFuture<JsonNode> execute(McpClientMessage message, Long id) {
+        return execute(message, id, false);
+    }
+
+    private CompletableFuture<JsonNode> execute(McpClientMessage message, Long id, boolean isRetry) {
+        CompletableFuture<JsonNode> reinitializeInProgress = this.initializeInProgress.get();
+        if (reinitializeInProgress != null) {
+            reinitializeInProgress.join();
+        }
         HttpRequest request = null;
         try {
             request = createRequest(message);
@@ -122,9 +138,27 @@ public class StreamableHttpMcpTransport implements McpTransport {
         httpClient
                 .sendAsync(request, responseInfo -> {
                     if (!isExpectedStatusCode(responseInfo.statusCode())) {
-                        future.completeExceptionally(
-                                new RuntimeException("Unexpected status code: " + responseInfo.statusCode()));
-                        return null;
+                        if (!(message instanceof McpInitializeRequest) && responseInfo.statusCode() == 404) {
+                            if (!isRetry) {
+                                initialize(StreamableHttpMcpTransport.this.initializeRequest)
+                                        .thenAccept(node -> {
+                                            execute(message, id, true)
+                                                    .thenAccept(future::complete)
+                                                    .exceptionally(t -> {
+                                                        future.completeExceptionally(t);
+                                                        return null;
+                                                    });
+                                        })
+                                        .exceptionally(t -> {
+                                            future.completeExceptionally(t);
+                                            return null;
+                                        });
+                            }
+                        } else {
+                            future.completeExceptionally(
+                                    new RuntimeException("Unexpected status code: " + responseInfo.statusCode()));
+                        }
+                        return HttpResponse.BodySubscribers.discarding();
                     } else {
                         Optional<String> contentType = responseInfo.headers().firstValue("Content-Type");
                         Optional<String> mcpSessionId = responseInfo.headers().firstValue("Mcp-Session-Id");
--- a/langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpServerHelper.java
+++ b/langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpServerHelper.java
@@ -6,16 +6,8 @@ import static org.junit.jupiter.api.Assumptions.assumeTrue;
 import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
-import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
-
-import dev.langchain4j.agent.tool.ToolSpecification;
-import dev.langchain4j.mcp.McpToolProvider;
-import dev.langchain4j.mcp.client.McpClient;
-import dev.langchain4j.service.tool.ToolExecutor;
-import dev.langchain4j.service.tool.ToolProvider;
-import dev.langchain4j.service.tool.ToolProviderResult;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import org.slf4j.Logger;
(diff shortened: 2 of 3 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
```diff
@@ -122,6 +134,9 @@ private CompletableFuture<JsonNode> execute(McpClientMessage message, Long id) {
         httpClient
                 .sendAsync(request, responseInfo -> {
                     if (!isExpectedStatusCode(responseInfo.statusCode())) {
+                        if (!(message instanceof McpInitializeRequest) && responseInfo.statusCode() == 404) {
```
jmartisk, line 141:
```suggestion
                        if (!(message instanceof McpInitializeRequest) && responseInfo.statusCode() == 404) {
                            // the server indicates that the session ID is not valid anymore
                            this.mcpSessionId.set(null);
```

jmartisk: You have to set the session ID to null, otherwise it will be added as a header to the following initialization request, and so the server will probably return 404 again.
Without this, your test is failing for that reason (when I run it)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
```diff
@@ -122,6 +134,9 @@ private CompletableFuture<JsonNode> execute(McpClientMessage message, Long id) {
...
+                            initialize(StreamableHttpMcpTransport.this.initializeRequest);
```
jmartisk:
```suggestion
                            if(!isRetry) {
                                initialize(StreamableHttpMcpTransport.this.initializeRequest).thenAccept(node -> {
                                    execute(message, id, true).thenAccept(future::complete).exceptionally(t -> {
                                        future.completeExceptionally(t);
                                        return null;
                                    });
                                }).exceptionally(t -> {
                                    future.completeExceptionally(t);
                                    return null;
                                });
                            }
                        } else {
                            future.completeExceptionally(
                                    new RuntimeException("Unexpected status code: " + responseInfo.statusCode()));
                        }
                        return HttpResponse.BodySubscribers.discarding();
```


How about this? With this piece of code, it will not only re-initialize the connection but also retry the original request - at most once (limiting retries here is important so that we don't get stuck in an infinite retry loop).
You will also have to add the `boolean isRetry` argument to the ` private CompletableFuture<JsonNode> execute` method and fill all other usages of this method with `isRetry=false`.

With these adjustments, I was able to make your test pass even after removing the part that says
```
        // First call after restart should fail due to missing session
        Assertions.assertThrows(
                Exception.class, () -> mcpClient.checkHealth(), "Expected session-related failure after restart");
```

because the next health check will already succeed

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
jmartisk: Or you know what, maybe it would be even better to move this ` this.mcpSessionId.set(null)` call into the `initialize` method , this will be more thread-safe

jmartisk (changes requested):

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
vijaypatidar: I’ve updated the code based on your suggestion — it now retries the failed operation after the session is reinitialized.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
vijaypatidar: I’ve updated the `createRequest` method so that the `Mcp-Session-Id` header is set to `null` for `McpInitializeRequest`.

jmartisk (approved):
LGTM, thanks a lot!!

jmartisk:
I've attempted to squash the commits, resolve conflicts and rebase it.. Hopefully I didn't mess it up :)

jmartisk:
@dliubars please have a look, I would've merged it but somehow it doesn't let me (maybe because the java 24 build is stuck?)

dliubarskyi:
@vijaypatidar thank you for the PR!

@jmartisk thank you for the review! Not sure why that build stuck. BTW, why do you need to squash if it is done automatically by github?

jmartisk:
There were conflicts on rebase so I guess the auto squashing wouldn't have worked.

PR #3857, opened 2025-10-11
feat: I think the Document class should be public so that users can i…

GuXiangFly:
feat: I think the Document class should be public so that users can implement the internalSearch method themselves.


dliubarskyi (approved):
@GuXiangFly thanks!

PR #3864, opened 2025-10-13
Disable McpRegistryClientIT.testListServersUpdatedSince for now

jmartisk:


jmartisk:
No, actually we can fix this test, I'll send another PR

PR #3870, opened 2025-10-13
test: Add validation and instance identity tests for ClassInstanceLoader

alxkm:
This PR adds explicit test coverage for input validation and instance creation behavior in `ClassInstanceLoader`.

### Changes

**Null input validation:**
- Added test verifying that `null` class parameter throws `NullPointerException`
- Ensures the API properly validates input at entry point

**Instance identity verification:**
- Added test confirming each call returns a new instance object
- Uses `isNotSameAs()` to verify object identity rather than equality
- Important for frameworks expecting fresh instances (not singletons)

**Cross-type validation:**
- Added test explicitly verifying type isolation between different classes
- Ensures `Class1` instances are never of type `Class2` and vice versa
- Makes type safety guarantees more explicit


dliubarskyi (approved):
@alxkm thank you!

Issue #3840, opened 2025-10-08
[FEATURE] Allow an Ai Service (and an Agent) to return void


mariofusco:
**Is your feature request related to a problem? Please describe.**

In some cases it could be useful to call an AiService not to directly produce a result, but only for its side effects like a function calling. In those cases it would be nice to have the possibility to return void.

**Describe the solution you'd like**

Allow an AiService method signature to return void and in that case also automatically add to the user message an instruction for the LLM to produce the smallest possible response (or none at all).

**Describe alternatives you've considered**

**Additional context**

patriot1burke:
FWIW: In the little app I wrote, I had all IMMEDIATE tools, but I was still interested in the case where the LLM did not match up to any of the tools, i.e. if the user typed "help" in the chat.

mariofusco:
> FWIW: In the little app I wrote, I had all IMMEDIATE tools, but I was still interested in the case where the LLM did not match up to any of the tools, i.e. if the user typed "help" in the chat.

Thanks for your feedback. Does this mean that you don't think that you could leverage the possibility or returning void in any AiService? The immediate tool return can be somewhat related but not exactly an overlapping feature. With immediate tool return you generally still want to return a result, for instance you have an LLM that calls a tool to perform a math operation and return the result of that operation as it is without having the LLM to further and unnecessarily reprocess it. With an AiService returning void you're saying that you're only interested in the eventual tools side effect (like writing in a DB) and nothing else.

mariofusco:
Implemented by https://github.com/langchain4j/langchain4j/pull/3874

PR #3874, opened 2025-10-14
Allow AiServices to return void

mariofusco:
## Issue

Closes #

https://github.com/langchain4j/langchain4j/issues/3840

## Change

Allow an AiServices to return `void` or `Result<Void>`. This doesn't imply any automatically generated change in the prompt sent to the LLM, so if users want to instruct the LLM to return the smallest possible result or no result at all, they have to put this explicitly in their own prompt. Any LLM outcome is simply discarded.

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -124,6 +124,10 @@ class DefaultAiServices<T> extends AiServices<T> {
     }
 
     private void validateMethods() {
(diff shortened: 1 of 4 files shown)


dliubarskyi (commented):
@mariofusco thank you! Please check failing tests

mariofusco:
> @mariofusco thank you! Please check failing tests

The `should_handle_non_interface_class` test was succeeding from the wrong reason, actually relying on one of the `void` returning method of the `String` class in order to throw the expected `IllegalConfigurationException`. With my last commit I added an explicit check for this.

dliubarskyi (approved):
@mariofusco thank you!

PR #3878, opened 2025-10-14
feat: agentic AgentBuilder supports streaming

### LINKED ISSUE #3519: [FEATURE] agent supports streamChatModel

1. agentic supports streamChatModel call
2. What is the difference between @Agent and @SystemMessage in agent? Or what is the necessity of @Agent?

``` java
public interface WithdrawAgent {

    @SystemMessage("""
            You are a banker that can only withdraw US dollars (USD) from a user account,
            """)
    @UserMessage("""
            Withdraw {{amount}} USD from {{user}}'s account and return the new balance.
            """)
    @Agent("A banker that withdraw USD from an account")
    String withdraw(@V("user") String user, @V("amount") Double amount);
}
```
@SystemMessage writes the same description as @Agent. At present, I do not use langchain4j-agentic, but it can also achieve an effect similar to agent.

**Additional context**

vesense:
## Issue

Closes https://github.com/langchain4j/langchain4j/issues/3519

## Change

agentic AgentBuilder supports streaming

## General checklist


mariofusco (changes requested):
The problem in using a streaming model within the agentic module, is not enabling a single agent to do so. Things are more complicated than that because an agent is not supposed to be used in isolation (you have plain AiServices for that) but to communicate and collaborate with other agents. In this context if an agent is streaming is response it is still not clear at all how other agents could read and use it. For instance, if a downstream agent has to wait regardless for the whole stream to be complete there is no point in using a streaming model at all.

If you can demonstrate how you can use a streaming agent inside an agentic system, together with other agents, please add a concrete example, otherwise I would rather put this change on hold until we won't have a solution for streaming agents coordination.

vesense:
@mariofusco Thanks for your response.

```
public class AgentStreamingWorkflowDemo {
    static ChatModel model;
    static StreamingChatModel steamingModel;

        CreativeWriter creativeWriter = AgenticServices.agentBuilder(CreativeWriter.class)
                .chatModel(model)
                .outputName("story")
                .build();

        AudienceEditor audienceEditor = AgenticServices.agentBuilder(AudienceEditor.class)
                .chatModel(model)
                .outputName("story")
                .build();

        StyleEditor styleEditor = AgenticServices.agentBuilder(StyleEditor.class)
                .streamingChatModel(steamingModel)
                .outputName("story")
                .build();

        UntypedAgent novelCreator = AgenticServices
                .sequenceBuilder()
                .subAgents(creativeWriter, audienceEditor, styleEditor)
                .outputName("story")
                .build();

        Map<String, Object> input = Map.of(
                "topic", "dragons and wizards",
                "style", "fantasy",
                "audience", "young adults");

        TokenStream tokenStream = (TokenStream) novelCreator.invoke(input);
        CompletableFuture<ChatResponse> futureResponse = new CompletableFuture<>();
        tokenStream
                .onPartialResponse(System.out::print)
                .onCompleteResponse(futureResponse::complete)
                .onError(futureResponse::completeExceptionally)
                .start();

        futureResponse.join();
    }

    public interface CreativeWriter {

        @UserMessage(
                """
            You are a creative writer.
            Generate a draft of a story no more than
            3 sentences long around the given topic.
            Return only the story and nothing else.
            Note: use chinese.
            The topic is {{topic}}.
            """)
        @Agent("Generates a story based on the given topic")
        String generateStory(@V("topic") String topic);
    }

    public interface AudienceEditor {

        @UserMessage(
                """
        You are a professional editor.
        Analyze and rewrite the following story to better align
        with the target audience of {{audience}}.
        Return only the story and nothing else.
        The story is "{{story}}".
        """)
        @Agent("Edits a story to better fit a given audience")
        String editStory(@V("story") String story, @V("audience") String audience);
    }

    public interface StyleEditor {

        @UserMessage(
                """
        You are a professional editor.
        Analyze and rewrite the following story to better fit and be more coherent with the {{style}} style.
        Return only the story and nothing else.
        Note: use chinese.
        The story is "{{story}}".
        """)
        @Agent("Edits a story to better fit a given style")
        TokenStream editStory(@V("story") String story, @V("style") String style);
    }
}

```

Here is an example of streaming agent used in the *sequence* workflow and now the streaming agent can just used in the *last* agent.

This PR maybe the first step to support streaming agent. And the feature,  streaming agents communicate and collaborate with other streaming agents, is not yet fully supported. I'm OK to hold until we have a a comprehensive solution. @mariofusco Thanks again.

ejstuart:
@vesense I think your example makes a great case for why streaming from an agent can be useful, only if used at the boundary between an orchestrator agent and a user.

@mariofusco Would you be open to a PR that enforces this explicitly (sub-agents can't be streaming agents), so that this top-level streaming can be supported as a first step?

Keeps coordination behaviour predictable while still enabling better UX for end user facing workflows?

vesense:
@ejstuart 
> I think your example makes a great case for why streaming from an agent can be useful, only if used at the boundary between an orchestrator agent and a user.

Yes,  typically the streaming agents can provide better user experience.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/Agents.java
```diff
@@ -24,7 +25,8 @@ public interface ExpertRouterAgentWithMemory extends AgenticScopeAccess {
 
     public interface CategoryRouter {
 
-        @UserMessage("""
+        @UserMessage(
```
mariofusco, line 35: Can you please rollback and avoid all the useless code reformatting in this class?

mariofusco:
> @vesense I think your example makes a great case for why streaming from an agent can be useful, only if used at the boundary between an orchestrator agent and a user.

I agree, please @vesense, add your example among tests.

> @mariofusco Would you be open to a PR that enforces this explicitly (sub-agents can't be streaming agents), so that this top-level streaming can be supported as a first step?
> 
> Keeps coordination behaviour predictable while still enabling better UX for end user facing workflows?

That would be great, I'm only very skeptical that you could clearly identify when an agent is "at the boundary" of the agentic system, except in the trivial case of the last agent in a sequence. In a loop an agent will be "at the boundary" if its output satisfies the exit condition, otherwise the loop will continue and its response won't be "at the boundary" anymore. With the other workflows you will have similar problems, not to mention the supervisor, where you literally have no idea of what will be the last agent to be invoked. 

In essence at the moment I see the value of a streaming response only for the obvious and very limited case of the last agent in a sequence, which is also the reason why I didn't implement this yet. If I'm right (but let me know if I'm missing something, I'm very open to discuss this) I would rather still leave this feature out, because we are giving users the possibility to have a streaming agent in the most trivial use case, but also a weapon to shoot themselves in a foot in the remaining 99% of the situations.

ejstuart:
> In essence at the moment I see the value of a streaming response only for the obvious and very limited case of the last agent in a sequence, which is also the reason why I didn't implement this yet. If I'm right (but let me know if I'm missing something, I'm very open to discuss this) I would rather still leave this feature out, because we are giving users the possibility to have a streaming agent in the most trivial use case, but also a weapon to shoot themselves in a foot in the remaining 99% of the situations.

It's a bit naive but in `langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java` could we just do a check if the agent has streaming model set when registering a subagent?

e.g.
```java
private static AgentExecutor createSubagent(
            SubAgent subagent, ChatModel chatModel, Consumer<DeclarativeAgentCreationContext> agentConfigurator) {
        if (isStreamingAgent(subagent)) {
            throw new IllegalArgumentException("Agent cannot be used as a sub-agent because it returns TokenStream. ");
        }
...
``` 

Essentially error when any subagent contains streaming methods?

mariofusco:
> Essentially error when any subagent contains streaming methods?

Yes, but I'm still not sure that I see the point, if we do so the example of @vesense will fail, which is the only thing that we want/can support with this improvement, am I wrong?

vesense:
@mariofusco @ejstuart 
 I updated the PR to keep coordination behavior predictable while still supporting streaming agent in workflows.

This implementation supports two scenarios of using streaming agents:
1. The **only last** agent is streaming in **sequential** workflow.
2. And **all** agents are streaming in **conditional** workflow.

There are some an examples here: [StreamingAgentsWorkflowIT.java](https://github.com/langchain4j/langchain4j/pull/3878/files#diff-11b94a7320a9c8cc2ff1bfe0404220e5afe55d9aa7dcbc8dace0dba649329a9e)

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/Agents.java
vesense: These changes are generated by ` ./mvnw spotless:apply`

mariofusco (changes requested):
> This implementation supports two scenarios of using streaming agents:
> 
>     1. The **only last** agent is streaming in **sequential** workflow.
> 
>     2. And **all** agents are streaming in **conditional** workflow.

The current implementation relies on strong assumptions that are not always true:

1. It assumes that that those sequential or conditional workflow are top level agents, but what happens if they are themselves subagent or a more complex workflow?
2. It assumes that the `outputKey` of the last agent in a sequential workflow is the same `outputKey` of the workflow itself. This is reasonable in most cases, but not always: for instance the last agent could be invoked to call a tool writing the result on a database, but not producing a result itself, or a human-in-the-loop used to ask a confirmation for all the actions taken so far before moving forward.
3. It assumes that all agents in a conditional workflow have the same `outputKey` and this `outputKey` is also the output of the conditional workflow itself. As for the former point this assumption is true in many cases but not necessarily.

I'm not sure if this is entirely possible, but it would be great to also explicitly check these assumptions.

Also it seems that this solution only works for `UntypedAgent`s? To be honest I introduced them before I thought that they could be convenient mostly as intermediate agents but for top level ones (that then define the API of the whole agentic system) I always prefer typed agent. Would it be possible to make this work for instance for a typed sequential agent returning a `TokenStream`?

vesense:
hi @mariofusco 

> I'm not sure if this is entirely possible, but it would be great to also explicitly check these assumptions.

Updated the PR to explicitly check these assumptions. Examples also updated accordingly.

> Also it seems that this solution only works for UntypedAgents? To be honest I introduced them before I thought that they could be convenient mostly as intermediate agents but for top level ones (that then define the API of the whole agentic system) I always prefer typed agent. Would it be possible to make this work for instance for a typed sequential agent returning a TokenStream?

The implementation supports typed agent, please check the example: https://github.com/langchain4j/langchain4j/pull/3878/files#diff-11b94a7320a9c8cc2ff1bfe0404220e5afe55d9aa7dcbc8dace0dba649329a9eR260

mariofusco (changes requested):
There is one last missing check: a workflow with a streaming agent must be the top level in the agentic system and not used itself as a subagent of another agentic pattern.

Also be aware that this pull request will be in contrast with a bigger rework that I'm developing here https://github.com/langchain4j/langchain4j/pull/3929 and I don't want to bring this in before that other pull request will be merged.

My suggestion is to wait until the introduction of that new planner architecture will be finalized and then also rebase and rework this change accordingly.

vesense:
> There is one last missing check: a workflow with a streaming agent must be the top level in the agentic system and not used itself as a subagent of another agentic pattern.

Hi @mariofusco 
This implementation supports streaming agent used in top level **or sub-agents**.

- Top level agent is Loop / Parallel / Supervisor, its sub-agents: 
  - [c,c,c]   // `c` means an agent using chatModel. Sub-agents cannot contain a streaming agent.
- Top level agent is Sequential, its sub-agents: 
  - Sequential[c,c,c]
  - Sequential[c,c,s]   // `s` means streaming agent. **The last streaming agent here also can be a:**
    - **Sequential[c,c,s]**
    - **Conditional[s,s,s]**
- Top level agent is Conditional, its sub-agents: 
  - Conditional[c,c,c]
  - Conditional[s,s,s]   // **The streaming agent here also can be a:**
    - **Sequential[c,c,s]**
    - **Conditional[s,s,s]**

I will provide more examples later.

> Also be aware that this pull request will be in contrast with a bigger rework that I'm developing here https://github.com/langchain4j/langchain4j/pull/3929 and I don't want to bring this in before that other pull request will be merged.
> My suggestion is to wait until the introduction of that new planner architecture will be finalized and then also rebase and rework this change accordingly.

No problem, I will update the PR accordingly after https://github.com/langchain4j/langchain4j/pull/3929 getting merged.

vesense:
@mariofusco @dliubarskyi

This PR has refactored based on the agentic planner(https://github.com/langchain4j/langchain4j/pull/3929) , and I also added some examples for streaming agent used in top level **or sub-agents**, please check the example: https://github.com/langchain4j/langchain4j/pull/3878/files#diff-11b94a7320a9c8cc2ff1bfe0404220e5afe55d9aa7dcbc8dace0dba649329a9eR140

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentUtil.java
```diff
@@ -29,8 +35,14 @@ public class AgentUtil {
     public static final String AGENTIC_SCOPE_ARG_NAME = "@AgenticScope";
     public static final String LOOP_COUNTER_ARG_NAME = "@LoopCounter";
 
+    private static final Set<String> WORKFLOW_STREAMING_AGENTS = new CopyOnWriteArraySet<>();
```
mariofusco: I really don't like this static data structure and I hope we could get rid of this and find a better way to pass around the information that an agent has a streaming chat model.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentUtil.java
```diff
@@ -194,7 +213,57 @@ public static Method validateAgentClass(Class<?> agentServiceClass, boolean fail
...
+        return true;
+    }
+
+    public static boolean isOnlyLastStreamingAgent(List<AgentInstance> agentInstances) {
+        List<AgentInstance> instances = agentInstances.subList(0, agentInstances.size() - 1);
+        return !hasStreamingAgent(instances) && isStreamingAgent(agentInstances.get(agentInstances.size() - 1));
+    }
+
+    public static boolean isStreamingAgent(AgentInstance agentInstance) {
+        return WORKFLOW_STREAMING_AGENTS.contains(agentInstance.agentId())
```
mariofusco: In fact I believe you can get rid of this `WORKFLOW_STREAMING_AGENTS`, add a `default isStreaming()` method to the `AgentInstance` interface that returns `false` and override this method in the `AgentExecutor` using the implementation below. Alternatively (which is probably even better) the AgentInvocationHandler could register in an internal `boolean` if the agent is using a streaming model and return this information [here](https://github.com/langchain4j/langchain4j/blob/320658d0592bf257b07db2c98ed7338ee48cf82e/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentInvocationHandler.java#L72).

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
```diff
@@ -90,8 +92,18 @@ private PlannerBasedInvocationHandler(AbstractServiceBuilder<?, ?> service, Supp
...
+                agentId,
+                description,
+                outputKey,
                 service.agenticMethod != null ? argumentsFromMethod(service.agenticMethod) : List.of());
+
+        checkSubAgents();
+    }
+
+    private void checkSubAgents() {
+        plannerSupplier.get().checkSubAgents(agentInstances, plannerInstance);
```
mariofusco: I'm not sure if there could be a better alternative, but I don't like the idea of having to instance a `Planner` (each and every planner actually) only to check if the subagents are streaming and if so if their configuration is valid. Also I don't like to "pollute" the generic `Planner` interface with this check. Probably you can move this check in the `Supplier` itself.

My idea is that you could have a `StreamingSubagentsChecker` interface and creating dedicated sequential and conditional suppliers implementing both `Supplier` and `StreamingSubagentsChecker`.

Then in this method you could do the following: if any of the subagent is streaming and the `plannerSupplier` is not also an instance of `StreamingSubagentsChecker` then you throw an exception, otherwise you cast to that interface and call the check on it. Let me know if I'm making sense.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/Planner.java
```diff
@@ -32,4 +34,9 @@ default Action done(Object result) {
         return new Action.DoneWithResultAction(result);
     }
 
+    default void checkSubAgents(List<AgentInstance> subAgents, AgentInstance plannerAgent) {
```
mariofusco: As I wrote, I would prefer to avoid having this method on the `Planner` interface, please see https://github.com/langchain4j/langchain4j/pull/3878/files#r2526711521

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorAgentServiceImpl.java
```diff
@@ -37,9 +37,15 @@ public T build() {
             });
         }
 
-        return build(() -> new SupervisorPlanner(chatModel, chatMemoryProvider, maxAgentsInvocations,
-                contextStrategy, responseStrategy, requestGenerator,
-                outputKey, output));
+        return build(() -> new SupervisorPlanner(
```
mariofusco, line 40: It would be great if you could revert this. Please don't change the formatting in classes were you didn't change anything else, otherwise it will be harder to review this pull request.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/ConditionalAgentServiceImpl.java
```diff
@@ -15,7 +15,8 @@
 import java.util.List;
 import java.util.function.Predicate;
 
-public class ConditionalAgentServiceImpl<T> extends AbstractServiceBuilder<T, ConditionalAgentService<T>> implements ConditionalAgentService<T> {
+public class ConditionalAgentServiceImpl<T> extends AbstractServiceBuilder<T, ConditionalAgentService<T>>
```
mariofusco, line 18: Same as https://github.com/langchain4j/langchain4j/pull/3878/files#r2526784937

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/LoopAgentServiceImpl.java
```diff
@@ -1,16 +1,17 @@
...
+import dev.langchain4j.agentic.scope.AgenticScope;
 import dev.langchain4j.agentic.workflow.LoopAgentService;
 import java.lang.reflect.Method;
 import java.util.function.BiPredicate;
 import java.util.function.Predicate;
 
-import static dev.langchain4j.agentic.internal.AgentUtil.validateAgentClass;
-
-public class LoopAgentServiceImpl<T> extends AbstractServiceBuilder<T, LoopAgentService<T>> implements LoopAgentService<T> {
+public class LoopAgentServiceImpl<T> extends AbstractServiceBuilder<T, LoopAgentService<T>>
```
mariofusco, line 13: Same as https://github.com/langchain4j/langchain4j/pull/3878/files#r2526784937

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/ParallelAgentServiceImpl.java
```diff
@@ -1,13 +1,14 @@
...
+
 import dev.langchain4j.agentic.UntypedAgent;
 import dev.langchain4j.agentic.internal.AbstractServiceBuilder;
 import dev.langchain4j.agentic.workflow.ParallelAgentService;
 import java.lang.reflect.Method;
 
-import static dev.langchain4j.agentic.internal.AgentUtil.validateAgentClass;
-
-public class ParallelAgentServiceImpl<T> extends AbstractServiceBuilder<T, ParallelAgentService<T>> implements ParallelAgentService<T> {
+public class ParallelAgentServiceImpl<T> extends AbstractServiceBuilder<T, ParallelAgentService<T>>
```
mariofusco, line 10: Same as https://github.com/langchain4j/langchain4j/pull/3878/files#r2526784937

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/SequentialAgentServiceImpl.java
```diff
@@ -1,13 +1,14 @@
...
+
 import dev.langchain4j.agentic.UntypedAgent;
 import dev.langchain4j.agentic.internal.AbstractServiceBuilder;
 import dev.langchain4j.agentic.workflow.SequentialAgentService;
 import java.lang.reflect.Method;
 
-import static dev.langchain4j.agentic.internal.AgentUtil.validateAgentClass;
-
-public class SequentialAgentServiceImpl<T> extends AbstractServiceBuilder<T, SequentialAgentService<T>> implements SequentialAgentService<T> {
+public class SequentialAgentServiceImpl<T> extends AbstractServiceBuilder<T, SequentialAgentService<T>>
```
mariofusco, line 10: Same as https://github.com/langchain4j/langchain4j/pull/3878/files#r2526784937

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/DeclarativeAgentIT.java
```diff
@@ -83,6 +90,35 @@ void declarative_single_agent_tests() {
         assertThat(story).isNotBlank();
     }
 
+    @Test
+    void declarative_streaming_agent_tests() throws Exception {
+        CreativeWriterForStreaming creativeWriter = AgenticServices.agentBuilder(CreativeWriterForStreaming.class)
+                .streamingChatModel(streamingBaseModel())
```
mariofusco: In the declarative API all configurations, including the `ChatModel`, are provided via annotations. You should add something similar also for the `StreamingChatModel`.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorAgentServiceImpl.java
vesense: This was auto-formatted by `./mvnw spotless:apply` to avoid spotless checking failure.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/ConditionalAgentServiceImpl.java
vesense: This was auto-formatted by `./mvnw spotless:apply` to avoid spotless checking failure.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/LoopAgentServiceImpl.java
vesense: This was auto-formatted by `./mvnw spotless:apply` to avoid spotless checking failure.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/ParallelAgentServiceImpl.java
vesense: This was auto-formatted by `./mvnw spotless:apply` to avoid spotless checking failure.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/SequentialAgentServiceImpl.java
vesense: This was auto-formatted by `./mvnw spotless:apply` to avoid spotless checking failure.

vesense:
@mariofusco @dliubarskyi Updated.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/AgentInstance.java
```diff
@@ -20,4 +20,10 @@ public interface AgentInstance {
     List<AgentArgument> arguments();
 
     List<AgentInstance> subagents();
+
+    default boolean isStreaming() {
+        return false;
+    }
+
+    default void setStreaming(boolean streaming) {}
```
mariofusco: This `AgentInstance` interface is exposed to users both when creating a custom `Planner` and now also as an informative model of an agent as you can see here https://github.com/langchain4j/langchain4j/blob/5774f82fca04adc77fe0b29e18f80be67fa96f7b/langchain4j-agentic/src/test/java/dev/langchain4j/agentic/WorkflowAgentsIT.java#L503

Representing the static model of an agent it shouldn't have any setter or mutable method, so please remove this.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/DefaultAgentInstance.java
```diff
@@ -3,7 +3,83 @@
 import java.lang.reflect.Type;
 import java.util.List;
 
-public record DefaultAgentInstance(Class<?> type, String name, String agentId, String description, Type outputType, String outputKey,
```
mariofusco, line 6: For what written [above](https://github.com/langchain4j/langchain4j/pull/3878/files#r2563760729) this must remain a `record`. Please revert this change.

vesense: @mariofusco Updated.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/AgentInstance.java
vesense: @mariofusco Updated.

vesense:
It's strange, the spotless check failed. I had already successfully run `./mvnw spotless:apply` before pushing.
And the  integration_test failed seems unrelated -> the localai model's ConnectException.

mariofusco (dismissed):
The implementation looks ok to me now and I don't have any other remark about it, except for the doubt weather the extra complexity that it brings in is compensated enough by the introduction of this streaming capability, that I still find of very limited usefulness and applicability in an agentic system.

@dliubarskyi I'd be curious to know your opinion about this.

@vesense Can you please add a new section in the `agents.md` documentation explaining how this works and its limitations?

vesense:
> @vesense Can you please add a new section in the agents.md documentation explaining how this works and its limitations?

OK. I will do this later, maybe in another patch.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
```diff
@@ -78,20 +80,44 @@ private PlannerBasedInvocationHandler(AbstractServiceBuilder<?, ?> service, Supp
...
                 service.agenticMethod != null ? argumentsFromMethod(service.agenticMethod) : List.of(),
-                service.agentExecutors().stream().map(AgentInstance.class::cast).toList());
+                agentInstances,
+                streaming);
         agenticSystemDataTypes(this.plannerAgent);
     }
 
+    private boolean checkSubAgents(List<AgentInstance> subAgents, String plannerAgentOutputKey) {
+        if (hasStreamingAgent(subAgents) && !(plannerSupplier instanceof StreamingSubAgentsChecker)) {
+            throw new IllegalArgumentException("Agent cannot be used as a sub-agent because it returns TokenStream.");
```
dliubarskyi, line 102: In general, there should be no problem substituting `ChatModel` with `StreamingChatModel`, as with `StreamingChatModel` you can do both 1) streaming token by token and 2) wait for the complete response. I am wondering if it makes sense to allow streaming (e.g., with `TokenStream`) in the middle of the agentic flow, also to simplify the overall logic? If a user (human) is waiting for the response, they can subscribe to the individual tokens with `TokenStream.onPartialResponse()` and be happy because they see the response as soon as possible. But if another agent is waiting for the response, it can subscribe only for the final response with the `TokenStream.onCompleteResponse()`. WDYT?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/StreamingSubAgentsChecker.java
```diff
@@ -0,0 +1,14 @@
...
+import java.util.List;
+
+public interface StreamingSubAgentsChecker {
+    /**
+     * Check the sub-agents whether is streaming or not
+     * @param subAgents the sub-agents needed to be checked
+     * @param plannerAgentOutputKey the planner agent outputKey
+     * @return `true` is streaming or `false` is not
+     */
+    boolean checkSubAgents(List<AgentInstance> subAgents, String plannerAgentOutputKey);
```
dliubarskyi:
```suggestion
    boolean isStreaming(List<AgentInstance> subAgents, String plannerAgentOutputKey);
```
According to the javadoc, this name seems more appropriate?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/AgentInstance.java
```diff
@@ -20,4 +20,8 @@ public interface AgentInstance {
     List<AgentArgument> arguments();
 
     List<AgentInstance> subagents();
+
+    default boolean isStreaming() {
+        return false;
```
dliubarskyi, line 25: Can't `outputType()` method be used to check if it is `TokenStream` or not?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
mariofusco: That's a great proposal. To recap we can let the possibility to configure a streaming model regardless and if the agentic system need the complete response before moving forward you just wait for its completion.

@vesense What do you think? Is the proposal clear? If you agree do you want to implement it in this pull request (I would prefer so) or do you prefer to bring this in as it is and work on this improvement in another PR?

jiangsier-xyz: BTW, in fact, sometimes we use streaming mode to circumvent the timeout limits of general gateways (especially after enabling thinking mode, where ChatModel response can take a particularly long time, but we cannot interfere with some gateway configurations in the network). In this case, we prefer to use StreamingChatModel, but only focus on the complete response.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/StreamingSubAgentsChecker.java
vesense: Appropriate. Updated.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/AgentInstance.java
vesense: No, `TokenStream` works well for MethodAgentInvoker, but not work for UntypedAgentInvoker, e.g. a workflow untyped agent : https://github.com/langchain4j/langchain4j/pull/3878/files#diff-3ebff31d677dfff703fd8578d1a0cc564e660e85e6852b7cb54518fd7fadf48eR87

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
vesense: >  In general, there should be no problem substituting ChatModel with StreamingChatModel, as with StreamingChatModel you can do both 1) streaming token by token and 2) wait for the complete response. I am wondering if it makes sense to allow streaming (e.g., with TokenStream) in the middle of the agentic flow, also to simplify the overall logic?

@dliubarskyi I think this is a more universal approach. In fact, I also wonder if it makes sense to allow streaming in the middle of the agentic flow. That's also why I didn't do it back then. @jiangsier-xyz 's input makes me think that maybe not a problem in some scenarios, there is its 'limited' value. (Or you may need a more intelligent gateway)

> @vesense What do you think? Is the proposal clear? If you agree do you want to implement it in this pull request (I would prefer so) or do you prefer to bring this in as it is and work on this improvement in another PR?

Yes, I'd like to do it, but I need some time to design and implement. I prefer to work on this improvement in another PR(based on this one).

 @dliubarskyi @mariofusco @jiangsier-xyz Thanks for your inputs.

dliubarskyi: > I also wonder if it makes sense to allow streaming in the middle of the agentic flow

Perhaps to unify the logic. But also: `StreamingChatModel` is better than `ChatModel` in this (maybe not groundbreaking) way: when calling tools, LLM can call multiple tools in a row, and arguments for tools can be quite long. When using `StreamingChatModel`, we could call the first tool as soon as streaming arguments for it is complete, while LLM is still generating tool calls/arguments for the next tools. So we could save a bit of time and start calling tools as soon as possible.

mariofusco:
@vesense sorry for late feedback on this.

We discussed again this improvement and we definitively want to support streaming models also in the agentic module, even because many users are asking for this. However, as I anticipated in some other comments, I believe that this solution looks quite complicated and of relatively limited applicability. We would prefer to implement what @dliubarskyi  suggested here https://github.com/langchain4j/langchain4j/pull/3878#discussion_r2580253044

@vesense let me know if you're going to work on this either rebasing and changing this pull request or restarting from scratch with a different one (which is probably the easiest alternative), otherwise I will work on it after the xmas break. Thank you.

mariofusco (changes requested):
I'm reworking this pull request following the idea proposed by @dliubarskyi and trying to allow the use of a streaming model on each and every agent, eventually waiting for the completion of the stream if an agent is not a terminal one.

mariofusco:
Alternative implementation https://github.com/langchain4j/langchain4j/pull/4350

mariofusco:
Replaced by https://github.com/langchain4j/langchain4j/pull/4350

vesense:
@mariofusco Sorry for the late reply. I've been too busy recently. Thanks for your great work.

PR #3880, opened 2025-10-14
feat(open-ai): Add support to set query parameters to an OpenAI request

### LINKED ISSUE #3879: [FEATURE] Support adding query parameters to OpenAI requests

As the OpenAI language model can be used to support different providers implementing OpenAI's API format, it can be necessary to support setting query parameters to the endpoint URL to be fully compatible with certain providers. For example, it might be necessary to set an `apiVersion` parameter to end up with an URL like `https://my-endpoint/v1/chat/completions?apiVersion=123`.

In addition to custom headers, the OpenAI integration should allow setting custom query parameters which will be appended to the endpoint URL.

The problem can be avoided by switching providers or asking providers to avoid relying on query parameters, but that might not always be possible, especially in a corporate context.

**Additional context**

Setting query params is already possible on the OpenAI official JDK, so it would probably make sense to expose the same functionality on the open-ai-official module as well in a follow-up.

maff:
## Issue

Closes #3879 

## Change

Adds a `customQueryParams` option to OpenAI requests (similar to `customHeaders`) which allows setting query parameters to be included in the URL when calling an OpenAI endpoint. As described in #3879, this can be necessary when using the OpenAI language model with model providers requiring certain query parameters (such as `?apiVersion=123`). 

## General checklist

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/DefaultOpenAiClient.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/DefaultOpenAiClient.java
@@ -1,5 +1,10 @@
 package dev.langchain4j.model.openai.internal;
 
+import static dev.langchain4j.http.client.HttpMethod.POST;
+import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotBlank;
+import static java.time.Duration.ofSeconds;
+
 import dev.langchain4j.http.client.HttpClient;
 import dev.langchain4j.http.client.HttpClientBuilder;
 import dev.langchain4j.http.client.HttpClientBuilderLoader;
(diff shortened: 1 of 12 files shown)


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/DefaultOpenAiClient.java
```diff
@@ -15,33 +20,31 @@
...
         HttpClientBuilder httpClientBuilder =
                 getOrDefault(builder.httpClientBuilder, HttpClientBuilderLoader::loadHttpClientBuilder);
 
         HttpClient httpClient = httpClientBuilder
-                .connectTimeout(getOrDefault(getOrDefault(builder.connectTimeout, httpClientBuilder.connectTimeout()), ofSeconds(15)))
-                .readTimeout(getOrDefault(getOrDefault(builder.readTimeout, httpClientBuilder.readTimeout()), ofSeconds(60)))
+                .connectTimeout(getOrDefault(
+                        getOrDefault(builder.connectTimeout, httpClientBuilder.connectTimeout()), ofSeconds(15)))
+                .readTimeout(
+                        getOrDefault(getOrDefault(builder.readTimeout, httpClientBuilder.readTimeout()), ofSeconds(60)))
```
maff, line 42: Note: these changes are unrelated to the PR, but failed the spotless check on touched files. Spotless formatting was done in a separate commit 4c6652b3bff537f01cd4a378479132cd09a8c618

dliubarskyi (approved):
@maff well done, thank you!

PR #3886, opened 2025-10-15
Retain description of anyOf elements from the MCP schema

### LINKED ISSUE #3884: [CRICRITICAL]  The parameters in ToolSpecification lost description, causing MCP invocation failed
**Observation:**

When an AI Service binds with a MCPToolProvider, MCP initializes tool information and parameters.I've identified that params JsonSchema descriptions are dropped when:

The parameter schema defines multiple types (e.g., "type": ["string", "null"], "type": ["integer", "string", "null"])

**Impact Analysis:**

Parameter descriptions typically include:

-  Parameter examples: promQL List, Like：["sum(pv)by(idc)", "increase(req_count[1m])"]

- Parameter Explanations: this parameter is core info, you must ....

- Constraints: must be positive integer 、 cant contain xxx, xxx

**Their loss directly causes:**

- LLM Tool Calling Failures: Models cannot generate valid requests without these semantics

- Degraded MCP Reliability: Incorrect parameter inference increases error rates

**Severity:**

P0(I think) - Core functionality breakage requiring fix

**How can I find this bug:**

The Mcp server define a parameter:

Image

The params uid, orgId are single type (string). LabelValues, startTime, endTime are nullable, which is multi type like [string , null]

Four params in toolProvidor:

Image
After Mcp handling, the params of tool which include multi type is wrong, the description is discarded. The single type params are normal.

As we see, all of mutilType param miss description:
Image

The generated params type by LLM is wrong as well due to this mistake:

Image
The type of param labelValues is string, but the true type is Map:

Image

jmartisk:
Fixes https://github.com/langchain4j/langchain4j/issues/3884

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
@@ -51,6 +51,9 @@ class ToolSpecificationHelper {
                     .map(ToolSpecificationHelper::jsonNodeToJsonSchemaElement)
                     .toArray(JsonSchemaElement[]::new);
             anyOf.anyOf(types);
+            if (node.has("description")) {
+                anyOf.description(node.get("description").asText());
+            }
             return anyOf.build();
         }
         JsonNode typeNode = node.get("type");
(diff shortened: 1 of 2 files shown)


vesense (approved):
LGTM

dliubarskyi (approved):
@jmartisk thanks a lot!

PR #3892, opened 2025-10-16
fix: thinking disabled should set thinking to null, toString should not expose secret

NeatGuyCoding:
## Issue

Closes #

## Change

thinking disabled should set thinking to null, toString should not expose secret

## General checklist

- [x] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [x] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

### DIFF
--- a/langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
+++ b/langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
@@ -273,7 +273,10 @@ abstract class WatsonxChat {
         }
 
         public T thinking(boolean enabled) {
-            if (enabled) return thinking(Thinking.builder().build());
+            if (enabled) {
+                return thinking(Thinking.builder().build());
+            }
+            this.thinking = null;
             return (T) this;
         }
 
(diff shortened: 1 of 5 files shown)


langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChat.java
```diff
@@ -273,7 +273,10 @@ public T defaultRequestParameters(ChatRequestParameters defaultRequestParameters
         }
 
         public T thinking(boolean enabled) {
-            if (enabled) return thinking(Thinking.builder().build());
+            if (enabled) {
```
dliubarskyi, line 276: @andreadimaio could you please review this file?

andreadimaio: Yes, this change is correct.

dliubarskyi (approved):
@NeatGuyCoding thank you!

PR #3900, opened 2025-10-16
Explicitly for the ResponseFormat instead of checking whether it's null when validating a ChatRequest for any Anthropic model

Bram--:
Claude supports ResponseFormat.TEXT but not JSON, so ensure to not throw an Exception when ResponseFormat.TEXT is explicitly set on the ChatRequestParameters. Also added some tests for the validate method.

## Issue
Closes # -

## Change
Because of how `ChatParameters` are build in our application `ResponseFormat` is always set. This results in an exception when using any Anthropic model as the InternalAnthropicHelper only checks for `null`. This PR simply checks the actualt ResponseFormat set.

## General checklist


dliubarskyi (approved):
@Bram-- thank you!

PR #3903, opened 2025-10-16
Prevent an infinite loop when using AiService with ToolChoice.REQUIRED

### LINKED ISSUE #3481: [FEATURE] Support One-Time Tool Enforcement in AiService

Yes. When using OpenAiChatModel, there's a toolChoice(REQUIRED) parameter available to enforce a tool call. However, when this model is wired into an AiService, it causes the service to enter an infinite loop of tool calls.

The issue is that toolChoice(REQUIRED) forces the model to always initiate a tool call even when the tool has already been called and a result is being returned. This makes the AiService unusable in scenarios where we only want a tool to be called once during the interaction.

I would like to see support for a new setting at the AiService level (not ChatModel) that allows tool invocation to be enforced only on the first request, and then behave normally (i.e., allow the LLM to decide whether to invoke a tool or not) on subsequent calls.

In other words:

1- On the first LLM request, tool call is mandatory.
2- On all subsequent requests, tool calls are optional and based on LLM discretion.

1- Using toolChoice(REQUIRED) on the ChatModel, but this causes an infinite loop when wired into an AiService, because every response continues to demand a tool call, even after a tool result is processed.

2- Manually implementing custom tool invocation logic outside of AiService, which defeats the purpose of using the abstraction.

**Additional context**

This behavior results from mixing ChatModel and AiService settings/parameters. It makes sense for this new tool enforcement behavior to be part of AiService, since it involves control over the flow of the conversation, not just raw model settings.

And it would be terrific if it becomes available in wiringMode = AiServiceWiringMode.EXPLICIT for Spring applications

andreadimaio:
Closes #3481

## Issue
The issue occurs when using AiService with a ChatModel that is configured with `ToolChoice.REQUIRED`. 
During message execution, AiService invoked the chat model multiple times, first to request tool execution and then again to produce the final response. Because the `ToolChoice` remained `REQUIRED` in both calls, the model repeatedly requested tool execution, leading to an infinite loop.  

The fix ensures that AiService automatically switches `ToolChoice.NONE` after the tool has been executed.

@dliubarskyi I have added a JUnit test and an IT test. However, from my understanding, watsonx's IT tests are not executed from the pipeline. It might be better to move the IT test under another model provider.

One more thing, I didn't run the IT tests for the other model providers.

## General checklist

- [x] I have manually run all the unit ~and integration tests~ in the module I have added/changed, and they are all green
- [x] I have manually run all the unit ~and integration tests~ in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

### DIFF
--- a/langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/it/WatsonxAIServiceIT.java
+++ b/langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/it/WatsonxAIServiceIT.java
@@ -1,0 +1,66 @@
+package dev.langchain4j.model.watsonx.it;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import dev.langchain4j.agent.tool.Tool;
+import dev.langchain4j.memory.chat.MessageWindowChatMemory;
+import dev.langchain4j.model.chat.Capability;
+import dev.langchain4j.model.chat.request.ToolChoice;
+import dev.langchain4j.model.watsonx.WatsonxChatModel;
+import dev.langchain4j.service.AiServices;
+import java.time.Duration;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
+@EnabledIfEnvironmentVariable(named = "WATSONX_API_KEY", matches = ".+")
+@EnabledIfEnvironmentVariable(named = "WATSONX_PROJECT_ID", matches = ".+")
+@EnabledIfEnvironmentVariable(named = "WATSONX_URL", matches = ".+")
+public class WatsonxAIServiceIT {
+
+    static final String API_KEY = System.getenv("WATSONX_API_KEY");
+    static final String PROJECT_ID = System.getenv("WATSONX_PROJECT_ID");
+    static final String URL = System.getenv("WATSONX_URL");
+
+    record Result(Double result) {}
+    ;
+
+    static class Calculator {
+        @Tool("Adds two given numbers")
+        double add(double a, double b) {
+            return a + b;
+        }
+
+        @Tool("Multiplies two given numbers")
+        String multiply(double a, double b) {
+            return String.valueOf(a * b);
+        }
+    }
+
+    interface Assistant {
+        Result chat(String userMessage);
+    }
+
+    @Test
+    public void should_not_loop_when_tool_choice_required() {
+
+        Assistant assistant = AiServices.builder(Assistant.class)
+                .chatModel(WatsonxChatModel.builder()
+                        .url(URL)
+                        .apiKey(API_KEY)
+                        .projectId(PROJECT_ID)
+                        .modelName("ibm/granite-4-h-small")
+                        .logRequests(true)
+                        .logResponses(true)
+                        .timeLimit(Duration.ofSeconds(30))
+                        .toolChoice(ToolChoice.REQUIRED)
+                        .temperature(0.5)
+                        .supportedCapabilities(Capability.RESPONSE_FORMAT_JSON_SCHEMA)
+                        .build())
+                .chatMemory(MessageWindowChatMemory.withMaxMessages(10))
+                .tools(new Calculator())
+                .build();
+
+        var result = assistant.chat("What is the result of 2 + 2?");
+        assertEquals(4, result.result());
+    }
+}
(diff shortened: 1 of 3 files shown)


langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -292,6 +294,16 @@ public ToolServiceResult executeInferenceAndToolsLoop(
                 messages = chatMemory.messages();
             }
 
+            ToolChoice toolChoice = getOrDefault(
+                    parameters.toolChoice(),
+                    chatModel.defaultRequestParameters().toolChoice());
+            // This is necessary to prevent an infinite loop.
+            if (ToolChoice.REQUIRED.equals(toolChoice)) {
+                parameters = parameters.overrideWith(ChatRequestParameters.builder()
+                        .toolChoice(ToolChoice.NONE)
```
dliubarskyi, line 303:
```suggestion
                        .toolChoice(ToolChoice.AUTO)
```
LLM might want to call another tool after the first tool is done, no need to restrict it

langchain4j/src/test/java/dev/langchain4j/service/AiServiceWithToolChoice.java
```diff
@@ -0,0 +1,83 @@
...
+            return a + b;
+        }
+    }
+
+    interface Assistant {
+        String chat(String userMessage);
+    }
+
+    @Test
+    void should_not_loop_when_tool_choice_required() {
```
dliubarskyi, line 44: please move this test into `AiServicesWithToolsIT`

langchain4j/src/test/java/dev/langchain4j/service/AiServiceWithToolChoice.java
```diff
@@ -0,0 +1,83 @@
...
+
+        when(chatModel.defaultRequestParameters())
```
dliubarskyi, line 46: I would suggest to use the real model for this test

langchain4j-watsonx/src/test/java/dev/langchain4j/model/watsonx/it/WatsonxAIServiceIT.java
```diff
@@ -0,0 +1,66 @@
...
+import dev.langchain4j.model.watsonx.WatsonxChatModel;
+import dev.langchain4j.service.AiServices;
+import java.time.Duration;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+
+@EnabledIfEnvironmentVariable(named = "WATSONX_API_KEY", matches = ".+")
+@EnabledIfEnvironmentVariable(named = "WATSONX_PROJECT_ID", matches = ".+")
+@EnabledIfEnvironmentVariable(named = "WATSONX_URL", matches = ".+")
+public class WatsonxAIServiceIT {
```
dliubarskyi, line 18: Do we need to test this for WatsonX specifically?

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
andreadimaio: In my tests, I can see that multiple calls to tools are all in the same 'JSON request'. However, your reasoning is correct: `AUTO` is much better than `NONE`. I'm going to change it.

langchain4j/src/test/java/dev/langchain4j/service/AiServiceWithToolChoice.java
andreadimaio: This is a JUnit test. Do I need to move it into an IT class?

dliubarskyi (commented):
@andreadimaio thank you for your PR!

I agree with [this comment](https://github.com/langchain4j/langchain4j/issues/3481#issuecomment-3200500187) that this should be configurable on the AI Service level. While your solution is solving the issue, I am not sure it flexible enough for future extension.

Moreover, it can break existing behaviour if users set `ToolChoice.REQUIRED` and `ReturnBehavior.IMMEDIATE`.

andreadimaio:
> I agree with https://github.com/langchain4j/langchain4j/issues/3481#issuecomment-3200500187 that this should be configurable on the AI Service level. While your solution is solving the issue, I am not sure it flexible enough for future extension.

I see your point, what is described in the comment is definitely more flexible. Mine was more of a quick fix to allow using `ToolChoice.REQUIRED` with the AI Service and the Agentic module.

> Moreover, it can break existing behaviour if users set ToolChoice.REQUIRED and ReturnBehavior.IMMEDIATE.

I’m not totally sure about this, but I might be missing some context.

At this point, if the idea is to go for a more flexible solution, we can close this PR.

PR #3906, opened 2025-10-17
feat: Added versioned docs and added doc for 1.7.1 release

### LINKED ISSUE #3895: [FEATURE] Introduce Versioned Documentation for LangChain4j
Currently, the LangChain4j documentation is unversioned. This creates a challenge for users who are working with older versions of the library, as the current docs may not accurately reflect the APIs or behavior they are using.

- Introduce versioned documentation using Docusaurus’s versioning system.
- Maintain docs for each released version (e.g., v0.1, v0.2, etc.), while keeping the next folder for upcoming changes.
- Provide a version dropdown in the navbar so users can easily switch between versions.
- Ensure sidebars and navigation are properly maintained per version.

**Additional context**

vijaypatidar:
## Issue 

Closes #3895

## Change
Enabled support for versioned documentation and added docs for the 1.7.1 release.
The 1.7.1 documentation was copied from the same Git commit hash used to create the release tag.

Additionally, removed some links from the header as it was getting cluttered with too many links. 

image

## General checklist


vijaypatidar:
@dliubarskyi I manually ran the command `npm run docusaurus docs:version 1.7.1` to create the versioned docs. We’ll need to do this for every release, though we could consider automating it with GitHub Actions in the future.

dliubarskyi:
I am wondering if it is very useful to have documentation for each minor version increase, as we do not do breaking changes usually, just add more features. Perhaps one per major version is enough? Once 2.X will be released, we can do a snapshot of 1.X. WDYT?

vijaypatidar:
> I am wondering if it is very useful to have documentation for each minor version increase, as we do not do breaking changes usually, just add more features. Perhaps one per major version is enough? Once 2.X will be released, we can do a snapshot of 1.X. WDYT?

I think version doc for each major release is good

PR #3911, opened 2025-10-17
demo possible approach to dynamic value injection into http layer

### LINKED ISSUE #2218: ChatRequest: support specifying API key, base URL, custom HTTP headers dynamically
`ChatRequest` should support specifying the following fields dynamically:
- API key
- base URL
- custom HTTP headers

This issue depends on https://github.com/langchain4j/langchain4j/issues/1044

tkroman:
Might be helpful with move towards #2218 resolution. This is definitely a breaking change but i just wanted to discuss if this would be a viable approach in which case i guess we could add overloads with defaults etc etc

## Issue
Closes #2218 (not really, just sticking to the pr template)

## Change
We add a new RequestContext record to help carry dynamic values like headers/query parameters (and potentially any other number of values that might have to be dynamically injected). Main use-case is to be able to pass these values from external systems/integrations/contexts. For instance, ADK has its own langchain4j shim, and its own internal request (`LlmRequest`) which has the capability to carry context similarly to `AgenticScope` but there is no way affect per-request behaviour dynamically. 

Actually, if we simplify this even further, the basic use-case is attaching tracing headers or any other sort of "scoped" data.

This doesn't have tests or docs updated because again, i dont know what's the maintainers' take on this approach but i would be willing to work on this further by e.g. adding useful overloads that would preserve API compatibility, adding tests/docs etc etc if we agree on the approach.

I'm in no rush to get this reviewed/discussed so even tho this PR breaks some checklists, i hope we could discuss potential solutions here at the pace that's comfortable for everyone.

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/ChatRequest.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/ChatRequest.java
@@ -7,6 +7,7 @@ import static java.util.Arrays.asList;
 
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.internal.context.RequestContext;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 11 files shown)


langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/ChatRequest.java
```diff
@@ -7,15 +7,19 @@
...
 import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.internal.context.RequestContext;
 import java.util.List;
 import java.util.Objects;
 
 public class ChatRequest {
 
     private final List<ChatMessage> messages;
     private final ChatRequestParameters parameters;
+    private final RequestContext context;
```
dliubarskyi, line 18: This is a purely HTTP data, I am not sure transport-agnostic `ChatRequest` should contain it. Perhaps we need an `HttpChatRequest` subtype

dliubarskyi (commented):
@tkroman thank you for your PR! TBH I am not sure we should move on with this now as there does not seems to be much interest in this feature.

Issue #3916, opened 2025-10-20
[FEATURE] Support Batch API where available


Bram--:
**Is your feature request related to a problem? Please describe.**
Batching can save quite a bit of money when working with Frontier models. Most major suppliers seem to support this and offer a 50% discount on requests:
 
 * [Gemini batch API](https://ai.google.dev/gemini-api/docs/batch-api)  
   * 50% discount when using batching 
   * Results can take up to 24 hours to process
   * Does not mention expiration of batch results.
   * Inline(limited to 20MB) and file(upto 2GB) API available
   * generateContent and Batching endpoints supported: `batchGenerateContent` & `asyncBatchEmbedContent`

 * [Anthropic batch API](https://docs.claude.com/en/docs/build-with-claude/batch-processing)  
   * 50% discount when using batching 
   * Most batches complete within 1 hour, results available when all messages have completed or after 24 hour whichever comes first
   * Batch results expire after 24 hours
   * Only inline(upto 256MB) API available
   * Only one endpoint supported: `messages`

 * [OpenAi batch API](https://platform.openai.com/docs/api-reference/batch) 
   * 50% discount when using batching 
   * Results can take up to 24 hours to process
   * Does not mention expiration of batch results.
   * Only file(max 50,000 requests up to 200 MB ) API available
   * Supports several endpoints: `/v1/responses`, `/v1/chat/completions`, `/v1/embeddings`, and `/v1/completions`

 * [Mistral batch API](https://docs.mistral.ai/api/#tag/batch)
   * 50% discount when using batching     
   * Results can take up to 24 hours to process (but can be configured to be lower/higher)
   * Does not mention expiration of batch results.
   * Only file(upto 512MB) API available
   * Supports several endpoints: `/v1/chat/completions`, `/v1/embeddings`, `/v1/fim/completions`, `/v1/moderations`, `/v1/chat/moderations`

 * More models/apis... ?

**Describe the solution you'd like**
I suppose it'd be nice to have concise generalized API but there appear to be many nuanced differences between the APIs. Making these APIs available in LangChain4j could really benefit the user base as there could be huge savings to be had. 

**Additional context**
The Anthropic API slightly differs from the Gemini API unfortunately (Making it harder to generalize). Where Anthropic's API has 2 different endpoints for polling and retrieving Gemini has this as a single endpoint and will always return the full batch on completion (opposed to an URL with the results). Anthropic also does not support deleting batches, but only cancelling them.

[...]

Issue #3774, opened 2025-09-28
[FEATURE] Add Remaining Features to Google AI Gemini


eljefe6a:
The Google AI Gemini features are missing some of the new API features of Google AI Gemini. These features are present in Google Vertex AI Gemini. However, Google Vertex AI Gemini is deprecated by Google.

Google AI Gemini should have feature parity with Google Vertex AI Gemini at a minimum. Ideally, the newest features in Google AI Gemini should be supported.

I have looked at using Google Vertex AI Gemini. However, I don't want to put time into already deprecated APIs.

cc: @glaforge

dliubarskyi:
Hi @eljefe6a, could you please specify the list of feature that are missing?

eljefe6a:
`VertexAiGeminiChatModel` has [`useGoogleSearch`](https://ai.google.dev/gemini-api/docs/google-search) and `GoogleAiGeminiChatModel` doesn't.

`GoogleAiGeminiChatModel` is missing from the Google Gemini API:
- https://ai.google.dev/gemini-api/docs/url-context
- https://ai.google.dev/gemini-api/docs/batch-api
- https://ai.google.dev/gemini-api/docs/caching

PR #3921, opened 2025-10-20
Improve readability/visibility of Gemini public API

Bram--:
## Issue
Preparations for: #3774 #3916

## Change
Improve the readability/visibility of the google-ai-gemini module's public API.

Move the internal-only classes into an internal package, which appears to be convention as per other modules. IMHO, this improves the parsability of the package although the internal package can now use some modularization. This PR does not change the public API.

## General checklist


dliubarskyi:
@Bram-- thanks! Not sure about this one TBH - too many changes for unclear benefit... Moving classes to another package makes them public, which is not ideal, we try to keep internals private as much as possible.

Bram--:
> @Bram-- thanks! Not sure about this one TBH - too many changes for unclear benefit... Moving classes to another package makes them public, which is not ideal, we try to keep internals private as much as possible.

Fair enough. The idea was to have the main package be a bit less overwhelming as there's currently 62 classes which makes it hard to comprehend. Perhaps a set of smaller refactors are in order to increase the simplicity of the main package. 

Will revert this for now. Thanks

dliubarskyi:
@Bram-- agreed, it is crowded, but this is the price we pay for internal classes to stay package-private so that we can freely change them without a risk of breaking someone elses code

PR #3927, opened 2025-10-21
Make GoogleAiGeminiChatModel unit-testable

Bram--:
## Issue
Preparation for: #3916 

## Change
Makes the `GoogleAiGeminiChatModel` testable by allowing the `GeminiService` to be passed in as ctor param. Other than that, I've added some basic coverage to the `GeminiService` and `GoogleAiGeminiChatModel` - a lot of defaults or edge-cases are not tested in the unit tests.

Production code changes:
 * Require the `GeminiService` to be passed into the Base client - I've opted for this opposed to using the `ServiceLoader` for simplicity let me know if you think `ServiceLoader` is a better option here
 * Changed `GeminiGenerationConfig` to be a record so that it can easily be compared in tests.

## General checklist

- [x] I have manually run all the unit and integration tests in the [core]


dliubarskyi (approved):
@Bram-- thank you!

Issue #3248, opened 2025-06-23
[FEATURE] LangChain4j Common Guardrails Library


rokon12:
**Is your feature request related to a problem? Please describe.**

While experimenting with LangChain4j's newly introduced guardrail feature, I noticed that developers need to implement common guardrails from scratch for each project. The framework provides excellent `InputGuardrail` and `OutputGuardrail` interfaces, but even basic guardrails like prompt injection detection or PII masking require custom implementation.

This leads to:
- Teams independently implementing the same common patterns (for example, prompt injection, PII detection, toxicity filtering)
- Varying implementation quality across different projects
- Development time spent on well-established patterns rather than application-specific logic
- A learning curve for developers new to AI safety best practices

Since these guardrails  are needed in almost every LLM application, providing ready-to-use implementations would be valuable.

**Describe the solution you'd like**

I propose creating a **LangChain4j Common Guardrails Library** - a collection of production-ready guardrail implementations that complement the existing framework. This library would provide:

### Core Guardrails:
- **Input Guardrails**: Prompt Injection Detection, PII Masking, Content Length Validation, Format Validation
- **Output Guardrails**: Toxicity Detection, Format Validation, Content Safety, Bias Detection

### Key Features:
- **Simple integration** with factory methods:
  ```java
  PromptInjectionGuardrail promptGuard = GuardrailFactory.promptInjection();
  PIIMaskingGuardrail piiGuard = GuardrailFactory.piiMasking();
  ```
- **Configurable implementations** via builders:
  ```java
  ContentLengthGuardrail lengthGuard = ContentLengthGuardrail.builder()
      .maxLength(280)
      .allowTruncation()
      .build();
  ```
- **Common presets** for typical scenarios:
  ```java
  Object[] chatbotSafety = GuardrailFactory.Presets.chatbotSafety();
  Object[] enterpriseSecurity = GuardrailFactory.Presets.enterpriseSecurity();
  ```
- **Production-ready** with comprehensive testing, thread-safety, and performance optimization
- **Maven/Gradle artifact** for easy dependency management

**Describe alternatives you've considered**

1. **Documentation Examples**: Provide implementation examples in the documentation
   - *Limitation*: Developers still need to copy and adapt code for each project

2. **External Libraries**: Use generic text validation libraries
   - *Limitation*: Not designed for LLM-specific concerns and don't integrate with LangChain4j's guardrail framework

3. **Third-party Services**: Use external content moderation APIs
[...]

PR #3928, opened 2025-10-21
Initializing the guardrails module

agoncal:
## Issue
Creates a new module that will gather all the out-of-the-box guardrails as discussed in https://github.com/langchain4j/langchain4j/issues/3248

## Change
* Added a new Maven module `langchain4j-guardrail` (I did not use the plural `langchain4j-guardrails` because there is no other plural repo)
* Move the `JsonExtractorOutputGuardrail` (and test) from the core to this new module (this introduces a breaking change as people have now to depend on this module, and not just core, to get this guardrail)
* Updated the documentation to point at the new location of the `JsonExtractorOutputGuardrailTests`

## General checklist


langchain4j-guardrail/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>1.8.0-beta15-SNAPSHOT</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-guardrail</artifactId>
+    <version>1.8.0-SNAPSHOT</version>
+    <name>LangChain4j :: Guardrails</name>
+    <description>Inpout and output out-of-the-box guardrails</description>
```
dliubarskyi:
```suggestion
    <description>Input and output out-of-the-box guardrails</description>
```

langchain4j-guardrail/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>1.8.0-beta15-SNAPSHOT</version>
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-guardrail</artifactId>
```
dliubarskyi:
```suggestion
    <artifactId>langchain4j-guardrails</artifactId>
```
Seems to be more natural

langchain4j-guardrails/src/main/java/dev/langchain4j/guardrails/JsonExtractorOutputGuardrail.java
dliubarskyi, line 1: Not sure about this breaking change, do we really want to do it? Although guardrail is an experimental feature, it is located in the stable module (core).
Also, making sure JSON is valid sounds like a core feature, so I would keep it in core TBH

dliubarskyi (commented):
@agoncal thank you!

langchain4j-guardrails/src/main/java/dev/langchain4j/guardrails/JsonExtractorOutputGuardrail.java
agoncal: @edeandrea what do you think? You were the one creating the `JsonExtractorOutputGuardrail`, no ? WDYT ? Is it worth moving it an introducing a breaking change ? Or leaving it in the core and use this new repo for future out-of-the-box guardrails ?

edeandrea: We could make it non-breaking if we wanted by leaving the original class in its location, deprecate it, and just having it extend the one in the guardrail repos module (and then `core` would need an `optional` dependency on the guardrails module.

I'm somewhat conflicted though. One one hand I agree with @dliubarskyi that LLM output JSON protection is a core feature, but then if we want to go down this road of having a guardrails repo, it makes sense to have it contain **ALL*** out of the box guardrails. Some of the ones we add may be more complicated than this one.

I don't have a strong opinion either way, but if we're voting I would vote to move it to the out of the box guardrails repo, and then do what I mentioned so that its backwards-compatible.

agoncal: Well, it's up to you, but this is not a major breaking change (it's just a new dependency to add to the pom as the package and class name stays the same). Having two classes, one inheriting the other, is kind of ugly. But you see. I would move it and just document the breaking change to have something clean.

edeandrea: > Having two classes, one inheriting the other, is kind of ugly

Agreed, but it gives us a path to deprecate and remove later so that anyone using it (like me :) ) is aware.

agoncal: The new Guardrail module depends on `core`. `JsonExtractorOutputGuardrail` is in `core`. So it now, `JsonExtractorOutputGuardrail` from core inherits from `JsonExtractorOutputGuardrail` in guardrails, then we have a cyclic dependency. So we need a `optional` dependency. And I'm not even talking about the fully class name that are identical. For me it's adding more hacks for something that is kind of ok. But again, you decide.

edeandrea: > we have a cyclic dependency

Ugh thats not good. It might be more of a headache than its worth. Still though I'm hesitant to just go out and break things. Maybe we maintain 2 copies & deprecate the one in core?

At the end of the day @dliubarskyi which would you prefer? Breaking something thats experimental or having some duplicate stuff in multiple artifacts for a period of time to allow us to properly deprecate stuff.

If this were the Quarkus extension we would just move it.

dliubarskyi: TBH I would prefer to introduce a new module when we have more than 1 OOTB guardrail. Right now this PR seems premature.

But if you insist to introduce this new module right now, I vote for avoiding breaking things and go with duplicating and deprecating the existing one.

agoncal: Well, it's the chicken and the egg thing. I already have an input guardrail to add, but following https://github.com/langchain4j/langchain4j/issues/3248, it looks like @rokon12 already have many. So once the module is create, we could already have a few.

Let me duplicate the `JsonExtractorOutputGuardrail`, update this PR and you'll see

agoncal: What do you think of this depreacted message https://github.com/langchain4j/langchain4j/pull/3928/commits/dc4b4aff19456a2c6cebb60f2bf7f6577527f3ef#diff-e2b5c1184c264c5a953cf05f4b2b205f9c203e0af23cde1d60817921a0120a82R23

edeandrea: I think that there are others who have guardrails they'd like to contribute...

@rokon12 @lordofthejars @sshaaf ^^^

@dliubarskyi see #3248

lordofthejars: I have a lot yes :)

sshaaf: I have a couple I can contribute..:) 

Furthermore there can be a set of standard guardrails. For example the Json one makes sense in many usecases apart from the core. However I also think there is an element where specific usecases for example branding checks (guardrail) etc are required. At that point IMHO it would make sense that people can contribute different guardrails. And it seems like this repo exactly intended for that. thanks @agoncal 

*Question*: If any developer is looking for guardrails for a specific use case, how will they navigate to it? Docs? or there is a plan to list them on the webpage etc? 
for example: https://hub.guardrailsai.com/

agoncal: @dliubarskyi looks like many people are motivated to contribute guardrails. I've turned off the draft PR. Feel free to comment or merge if you think it's ok

docs/docs/tutorials/guardrails.md
```diff
@@ -427,9 +427,9 @@ In the situation where a **_retry_** or **_reprompt_** in the chain eventually s
...
+| [`JsonExtractorOutputGuardrail`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-guardrail/src/main/java/dev/langchain4j/guardrail/JsonExtractorOutputGuardrail.java) | An output guardrail that will check whether or not a response can be successfully deserialized from JSON to an object of a certain type.<br/> - Uses a [Jackson ObjectMapper](https://github.com/FasterXML/jackson-databind) to try and deserialize an object.<br/> - The LLM is reprompted if the response can't be deserialized into the expected object type.<br/> - Can be used as-is, or can be extended and customized (there are several `protected` methods that can be overridden to customize behavior). |
```
dliubarskyi: Having the same package in 2 modules leads to "split package" problem:

🚨 Split package detected: dev.langchain4j.guardrail
    ↳ in: langchain4j-guardrails-1.8.0-SNAPSHOT.jar
    ↳ and: langchain4j-core-1.8.0-SNAPSHOT.jar

langchain4j-core/src/main/java/dev/langchain4j/guardrail/JsonExtractorOutputGuardrail.java
```diff
@@ -3,24 +3,27 @@
...
- *     defaults to {@link #DEFAULT_REPROMPT_PROMPT}.
+ * If deserialization fails, the LLM will be reprompted with {@link #getInvalidJsonReprompt(AiMessage, String)}, which
+ * defaults to {@link #DEFAULT_REPROMPT_PROMPT}.
  * </p>
  *
  * @param <T> The type of object that the class should deserialize from JSON
+ * @deprecated This class has been moved to the `langchain4j-guardrails` module and will be removed in a future release.
+ * *             Add a dependency on `dev.langchain4j:langchain4j-guardrails` and use the class from that module instead.
  */
+@Deprecated(forRemoval = true)
```
dliubarskyi:
```suggestion
@Deprecated(since = "1.9.0", forRemoval = true)
```

langchain4j-guardrails/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>1.8.0-beta15-SNAPSHOT</version>
```
dliubarskyi:
```suggestion
        <version>1.9.0-beta16-SNAPSHOT</version>
```

langchain4j-guardrails/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-guardrails</artifactId>
+    <version>1.8.0-SNAPSHOT</version>
```
dliubarskyi:
```suggestion
    <version>1.9.0-beta16-SNAPSHOT</version>
```

langchain4j-guardrails/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+    <name>LangChain4j :: Guardrails</name>
+    <description>Input and output out-of-the-box guardrails</description>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
+            <version>1.8.0-SNAPSHOT</version>
```
dliubarskyi:
```suggestion
            <version>1.9.0-SNAPSHOT</version>
```

pom.xml
```diff
@@ -103,6 +103,9 @@
         <!-- embedding store filter parsers -->
         <module>embedding-store-filter-parsers/langchain4j-embedding-store-filter-parser-sql</module>
 
+        <!-- guardrails -->
+        <module>langchain4j-guardrails</module>
```
dliubarskyi, line 107: please add it to `langchain4j-bom/pom.xml` as well

langchain4j-guardrails/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
```
dliubarskyi: seems like this dependency is not used

dliubarskyi (commented):
@agoncal thank you! Please fix the remaining comments and we can merge

docs/docs/tutorials/guardrails.md
agoncal: I've changed the package name to `guardrails` plural

docs/docs/tutorials/guardrails.md
```diff
@@ -427,9 +427,9 @@ In the situation where a **_retry_** or **_reprompt_** in the chain eventually s
...
+| [`JsonExtractorOutputGuardrail`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-guardrail/src/main/java/dev/langchain4j/guardrails/JsonExtractorOutputGuardrail.java) | An output guardrail that will check whether or not a response can be successfully deserialized from JSON to an object of a certain type.<br/> - Uses a [Jackson ObjectMapper](https://github.com/FasterXML/jackson-databind) to try and deserialize an object.<br/> - The LLM is reprompted if the response can't be deserialized into the expected object type.<br/> - Can be used as-is, or can be extended and customized (there are several `protected` methods that can be overridden to customize behavior). |
```
dliubarskyi:
```suggestion
| [`JsonExtractorOutputGuardrail`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-guardrails/src/main/java/dev/langchain4j/guardrails/JsonExtractorOutputGuardrail.java) | An output guardrail that will check whether or not a response can be successfully deserialized from JSON to an object of a certain type.<br/> - Uses a [Jackson ObjectMapper](https://github.com/FasterXML/jackson-databind) to try and deserialize an object.<br/> - The LLM is reprompted if the response can't be deserialized into the expected object type.<br/> - Can be used as-is, or can be extended and customized (there are several `protected` methods that can be overridden to customize behavior). |
```

dliubarskyi (dismissed):
@agoncal thank you!

PR #3929, opened 2025-10-21
Introduce a generic agentic planner and reimplement all agentic patterns through it

mariofusco:
## Custom agentic patterns

The `langchain4j-agentic` module provides out-of-the-box a few useful agentic pattern, but what if none of them fit the specific needs of your application? In this case it is possible to create your own custom pattern, that orchestrates the interactions among a set of subagents in a way that is tailored to your requirements.

In more details an agentic pattern is simply the specification of an execution plan for the subagents that it coordinates. This plan can be defined by implementing the following `Planner` interface:

```java
public interface Planner {

    default void init(InitPlanningContext initPlanningContext) { }

    default Action firstAction(PlanningContext planningContext) {
        return nextAction(planningContext);
    }

    Action nextAction(PlanningContext planningContext);
}
```

This interface has three methods: `init`, `firstAction`, and `nextAction`. The `init` method is called once at the beginning of the execution, and can be used to initialize any state or data structures needed by the planner. The `firstAction` method is called to determine the first action to be taken by the agentic pattern, while the `nextAction` method is called after each agent execution to determine the next action to be taken based on the current state of the `AgenticScope` and the result of the previous agent execution.

Note that the `firstAction` method has been introduced only because in many cases it is convenient to have a distinct callback defining the very first agent to be invoked by the `Planner`. However, for the situations where this distinction is not necessary, it provides a default implementation that simply forwards the call to the `nextAction` method, so it is not strictly necessary to override it.

The `Action` class returned by the `firstAction` and `nextAction` methods represents the next step to be taken by the agentic pattern, and can be one either a list of one or more subagents to be called next, or a signal that the execution has been completed. If the action specifies only one subagent invocation, then it will be executed sequentially, and in the same thread that is executing the planner itself, while if there are more than one, they will be executed in parallel using the provided `Executor` or the LangChain4j default one.

All the built-in agentic patterns are also written in terms of this `Planner` abstraction and giving a look at their implementation can clarify how this works and be a good starting point to create your own custom patterns. For instance the parallel workflow is probably the simplest of those implementation, and it is defined as follows:

```java
public class ParallelPlanner implements Planner {

    private List<AgentInstance> agents;

    @Override
    public void init(InitPlanningContext initPlanningContext) {
        this.agents = initPlanningContext.subagents();
    }

    @Override
    public Action firstAction(PlanningContext planningContext) {
        return call(agents);
    }

    @Override
    public Action nextAction(PlanningContext planningContext) {
        return done();
    }
}
```

Here the `init` method simply stores the list of subagents with which the parallel workflow has been configured, while the `firstAction` method returns an action that calls all those agents in parallel. Once this parallel execution is completed, there isn't any other action to be taken, so the `nextAction` method simply returns `done()` used to signal the termination of the execution.

The `Planner` implementing the sequential workflow is only slightly more complex, as it needs to keep track of the next subagent to be invoked using an internal cursor, and then return the appropriate action in the `nextAction` method or signal the termination of the execution when all subagents have been invoked.

```java
public class SequentialPlanner implements Planner {

    private List<AgentInstance> agents;
    private int agentCursor = 0;

    @Override
    public void init(InitPlanningContext initPlanningContext) {
        this.agents = initPlanningContext.subagents();
    }

    @Override
    public Action nextAction(PlanningContext planningContext) {
        return agentCursor >= agents.size() ? done() : call(agents.get(agentCursor++));
    }
}
```

To understand how to define an agentic system from a planner implementation, it is possible, for example, to create an instance of the formerly discussed sequential workflow generating a novel for a topic and then editing it for a specific style and audience, as it follows:

```java
UntypedAgent novelCreator = AgenticServices.plannerBuilder()
                .subAgents(creativeWriter, audienceEditor, styleEditor)
                .outputKey("story")
                .planner(SequentialPlanner::new)
                .build();
```

which is totally equivalent to use the dedicated API for sequential workflows:

```java
UntypedAgent novelCreator = AgenticServices.sequenceBuilder()
                .subAgents(creativeWriter, audienceEditor, styleEditor)
                .outputKey("story")
                .build();
```

The `plannerBuilder()` method is similar to all other agent builders, with the only difference that it requires to provide a `Supplier<Planner>` returning a new instance of the specific planner to be used by this agentic system. Of course an agentic system implementing a custom planner can be seamlessly combined with any other of the agentic pattern offered out-of-the-box by the `langchain4j-agentic` module.

Having clarified how this `Planner` abstraction works, it is now possible to create your own custom agentic patterns by implementing it. The following sections discuss the two examples of custom patterns, provided in the `langchain4j-agentic-patterns` module, that can be useful in different scenarios. Other custom patterns can be created following the same approach and may be contributed back to the LangChain4j project.

### Goal oriented agentic pattern

The workflow patterns and the supervisor agent represents the two extremes of the spectrum of possible agentic systems: the former is completely deterministic and rigid, forcing to decide in advance the sequence of agents to be invoked, while the latter is completely flexible and adaptive, but delegates the decision of the sequence of agents to be invoked to a non-deterministic LLM. However, there are cases where a middle ground between these two extremes can be more appropriate, allowing agents to work towards a specific goal in a relatively flexible way, but also determining how these agents should be invoked in an algorithmic way.

In order to put this approach in practice, not only the whole agentic system needs to define a goal, but also each subagent needs to declare its own pre and postconditions. This is necessary to calculate the sequence of agent invocations that lead to the achievement of the goal in the fastest possible way. However, all this information are implicitly already present in the agentic system, as those pre and postconditions are nothing else than the required inputs and produced outputs of each agent, and the final goal is simply the desired outputs of the whole agentic system.

Following this idea, it is possible to calculate a dependency graph of all the subagents participating in the agentic system, and then to implement a `Planner` that is capable of analyzing the initial state of the `AgenticScope`, comparing it with the desired goal, and then using that graph to determine the sequence of agent invocations that can lead to the achievement of that goal.

```java
public class GoalOrientedPlanner implements Planner {

    private String goal;

    private GoalOrientedSearchGraph graph;
    private List<AgentInstance> path;

    private int agentCursor = 0;

    @Override
    public void init(InitPlanningContext initPlanningContext) {
        this.goal = initPlanningContext.plannerAgent().outputKey();
        this.graph = new GoalOrientedSearchGraph(initPlanningContext.subagents());
    }

    @Override
    public Action firstAction(PlanningContext planningContext) {
        path = graph.search(planningContext.agenticScope().state().keySet(), goal);
        if (path.isEmpty()) {
            throw new IllegalStateException("No path found for goal: " + goal);
        }
        return call(path.get(agentCursor++));
    }

    @Override
    public Action nextAction(PlanningContext planningContext) {
        return agentCursor >= path.size() ? done() : call(path.get(agentCursor++));
    }
}
```

As anticipated, here the goal coincides with the final output of the planner-based agentic pattern itself, while the path from the initial state to the goal is calculated using a `GoalOrientedSearchGraph`, that is built analyzing the input and output keys of all subagents. The sequence of agents to be invoked is then calculated as the shortest path on that graph from the current state to the desired goal.

To give a practical example of how this works, let's try to build a goal-oriented agentic system that can extract the name and zodiac sign of a person from a prompt, generate the horoscope for that sign, look for a related story on the internet and finally create a nice writeup combining all this information. We can achieve this set of tasks by using the following 5 agents:

```java
public interface HoroscopeGenerator {
    @SystemMessage("You are an astrologist that generates horoscopes based on the user's name and zodiac sign.")
    @UserMessage("Generate the horoscope for {{person}} who is a {{sign}}.")
    @Agent("An astrologist that generates horoscopes based on the user's name and zodiac sign.")
    String horoscope(@V("person") Person person, @V("sign") Sign sign);
}

public interface PersonExtractor {

    @UserMessage("Extract a person from the following prompt: {{prompt}}")
    @Agent("Extract a person from user's prompt")
    Person extractPerson(@V("prompt") String prompt);
}

public interface SignExtractor {

    @UserMessage("Extract the zodiac sign of a person from the following prompt: {{prompt}}")
    @Agent("Extract a person from user's prompt")
    Sign extractSign(@V("prompt") String prompt);
}

public interface Writer {
    @UserMessage("""
            Create an amusing writeup for {{person}} based on the following:
            - their horoscope: {{horoscope}}
            - a current news story: {{story}}
            """)
    @Agent("Create an amusing writeup for the target person based on their horoscope and current news stories")
    String write(@V("person") Person person, @V("horoscope") String horoscope, @V("story") String story);
}

public interface StoryFinder {

    @SystemMessage("""
            You're a story finder, use the provided web search tools, calling it once and only once,
            to find a fictional and funny story on the internet about the user provided topic.
            """)
    @UserMessage("""
            Find a story on the internet for {{person}} who has the following horoscope: {{horoscope}}.
            """)
    @Agent("Find a story on the internet for a given person with a given horoscope")
    String findStory(@V("person") Person person, @V("horoscope") String horoscope);
}
```

Leveraging the `GoalOrientedPlanner` developed before, these agents can be combined in a goal-oriented agentic system as follows:

```java
HoroscopeGenerator horoscopeGenerator = AgenticServices.agentBuilder(HoroscopeGenerator.class)
        .chatModel(baseModel())
        .outputKey("horoscope")
        .build();

PersonExtractor personExtractor = AgenticServices.agentBuilder(PersonExtractor.class)
        .chatModel(baseModel())
        .outputKey("person")
        .build();

SignExtractor signExtractor = AgenticServices.agentBuilder(SignExtractor.class)
        .chatModel(baseModel())
        .outputKey("sign")
        .build();

Writer writer = AgenticServices.agentBuilder(Writer.class)
        .chatModel(baseModel())
        .outputKey("writeup")
        .build();

StoryFinder storyFinder = AgenticServices.agentBuilder(StoryFinder.class)
        .chatModel(baseModel())
        .tools(new WebSearchTool())
        .outputKey("story")
        .build();

UntypedAgent horoscopeAgent = AgenticServices.plannerBuilder()
        .subAgents(horoscopeGenerator, personExtractor, signExtractor, writer, storyFinder)
        .outputKey("writeup")
        .planner(GoalOrientedPlanner::new)
        .build();
```

As anticipated, the overall goal of this agentic system is to produce a `writeup` which is also the output key of the GOAP-based planner itself. Taking into account the inputs and outputs of all subagents, the dependency graph built by the `GoalOrientedSearchGraph` will look like this:

image

When invoking this agentic system with a prompt like "My name is Mario and my zodiac sign is pisces"

```java
Map<String, Object> input = Map.of("prompt", "My name is Mario and my zodiac sign is pisces");
String writeup = horoscopeAgent.invoke(input);
```

the `GoalOrientedPlanner` will analyze the initial state of the `AgenticScope`, that contains only the `prompt` variable, and then it will calculate the shortest path on the dependency graph from that initial state to the desired goal, which is the `writeup`, so that the resulting sequence of agent invocations will be:

```
Agents path sequence: [extractPerson, extractSign, horoscope, findStory, write]
```

Note that, as anticipated, this goal-oriented agentic pattern can be mixed and combined with any other of the existing agentic patterns. For instance this possibility can be used to overcome an evident limitation of this approach that, being optimized to reach a specific goal following the shortest possible path, structurally doesn't allow loops, so in some cases it could be useful to have a loop agentic pattern as a subagent of this goal-oriented one.

### Peer-to-peer agentic pattern

All the agentic system discussed up to this point are based on a centralized and hierarchical architecture. In fact all the workflow patterns had a well-defined top-level agent coordinating the activities of multiple sub-agents in a programmatically predetermined way. Even the supervisor pattern, which is more flexible and dynamic thanks to the presence of its LLM-based planner agent, still relies on a coordinator agent that controls the interactions among the various sub-agents. This typology of architectures are suitable for many applications and scenarios, but they can also have some limitations, especially in terms of scalability and fault tolerance. This is why we may want to offer an alternative peer-to-peer approach for multi-agent systems, that can overcome these limitations by adopting a more decentralized and distributed strategy.

In a peer-to-peer agentic systems there isn't any top level agent, and all agents are equal peers that are coordinated through the state of the `AgenticScope`. In particular, an agent is triggered by the presence of its own required inputs as state variables in the `AgenticScope`. Subsequently, a change in one or more of those variables, produced by the output of a different agent, can retrigger the invocation of that agent again. The process terminates either when the `AgenticScope` reaches a stable state and no agent can be invoked anymore, or when the predefined exit condition is satisfied, or when a maximum number of agent invocations has been reached. A `Planner` implementation that realizes this peer-to-peer agentic pattern could be written as it follows:

```java
public class P2PPlanner implements Planner {

    private final int maxAgentsInvocations;
    private final BiPredicate<AgenticScope, Integer> exitCondition;

    private int invocationCounter = 0;
    private Map<String, AgentActivator> agentActivators;

    public P2PPlanner(int maxAgentsInvocations, BiPredicate<AgenticScope, Integer> exitCondition) {
        this(null, maxAgentsInvocations, exitCondition);
    }

    @Override
    public void init(InitPlanningContext initPlanningContext) {
        this.agentActivators = initPlanningContext.subagents().stream().collect(toMap(AgentInstance::agentId, AgentActivator::new));
    }

    @Override
    public Action nextAction(PlanningContext planningContext) {
        if (terminated(planningContext.agenticScope())) {
            return done();
        }

        AgentActivator lastExecutedAgent = agentActivators.get(planningContext.previousAgentInvocation().agentId());
        lastExecutedAgent.finishExecution();
        agentActivators.values().forEach(a -> a.onStateChanged(lastExecutedAgent.agent.outputKey()));

        return nextCallAction(planningContext.agenticScope());
    }

    private Action nextCallAction(AgenticScope agenticScope) {
        AgentInstance[] agentsToCall = agentActivators.values().stream()
                .filter(agentActivator -> agentActivator.canActivate(agenticScope))
                .peek(AgentActivator::startExecution)
                .map(AgentActivator::agent)
                .toArray(AgentInstance[]::new);
        invocationCounter += agentsToCall.length;
        return call(agentsToCall);
    }

    private boolean terminated(AgenticScope agenticScope) {
        return invocationCounter > maxAgentsInvocations || exitCondition.test(agenticScope, invocationCounter);
    }
}
```

Here the `P2PPlanner` keeps track of the number of agent invocations performed so far, and uses an `AgentActivator` for each subagent to determine if it can be invoked based on the current state of the `AgenticScope`. The `nextAction` method checks if the exit condition has been met or if the maximum number of invocations has been reached, and if not, it identifies all agents that can be activated based on the current state, marks them as started, and returns an action to call them.

To give a practical example of how this works let's try to build a peer-to-peer agentic system that can perform a scientific research and formulate new hypothesis on a given topic, so that the API of this service could be something like:

```java
public interface ResearchAgent {

    @Agent("Conduct research on a given topic")
    String research(@V("topic") String topic);
}
```

To this purpose the following 5 agents can be defined:

```java
public interface LiteratureAgent {

    @SystemMessage("Search for scientific literature on the given topic and return a summary of the findings.")
    @UserMessage("""
            You are a scientific literature search agent.
            Your task is to find relevant scientific papers on the topic provided by the user and summarize them.
            Use the provided tool to search for scientific papers and return a summary of your findings.
            The topic is: {{topic}}
            """)
    @Agent("Search for scientific literature on a given topic")
    String searchLiterature(@V("topic") String topic);
}

public interface HypothesisAgent {

    @SystemMessage("Based on the research findings, formulate a clear and concise hypothesis related to the given topic.")
    @UserMessage("""
            You are a hypothesis formulation agent.
            Your task is to formulate a clear and concise hypothesis based on the research findings provided by the user.
            The topic is: {{topic}}
            The research findings are: {{researchFindings}}
            """)
    @Agent("Formulate hypothesis around a give topic based on research findings")
    String makeHypothesis(@V("topic") String topic, @V("researchFindings") String researchFindings);
}

public interface CriticAgent {

    @SystemMessage("Critically evaluate the given hypothesis related to the specified topic. Provide constructive feedback and suggest improvements if necessary.")
    @UserMessage("""
            You are a critical evaluation agent.
            Your task is to critically evaluate the hypothesis provided by the user in relation to the specified topic.
            Provide constructive feedback and suggest improvements if necessary.
            If you need to, you can also perform additional research to validate or confute the hypothesis using the provided tool.
            The topic is: {{topic}}
            The hypothesis is: {{hypothesis}}
            """)
    @Agent("Critically evaluate a hypothesis related to a given topic")
    String criticHypothesis(@V("topic") String topic, @V("hypothesis") String hypothesis);
}

public interface ValidationAgent {

    @SystemMessage("Validate the provided hypothesis on the given topic based on the critique provided.")
    @UserMessage("""
            You are a validation agent.
            Your task is to validate the hypothesis provided by the user in relation to the specified topic based on the critique provided.
            Validate the provided hypothesis, either confirming it or reformulating a different hypothesis based on the critique.
            The topic is: {{topic}}
            The hypothesis is: {{hypothesis}}
            The critique is: {{critique}}
            """)
    @Agent("Validate a hypothesis based on a given topic and critique")
    String validateHypothesis(@V("topic") String topic, @V("hypothesis") String hypothesis, @V("critique") String critique);
}

public interface ScorerAgent {

    @SystemMessage("Score the provided hypothesis on the given topic based on the critique provided.")
    @UserMessage("""
            You are a scoring agent.
            Your task is to score the hypothesis provided by the user in relation to the specified topic based on the critique provided.
            Score the provided hypothesis on a scale from 0.0 to 1.0, where 0.0 means the hypothesis is completely invalid and 1.0 means the hypothesis is fully valid.
            The topic is: {{topic}}
            The hypothesis is: {{hypothesis}}
            The critique is: {{critique}}
            """)
    @Agent("Score a hypothesis based on a given topic and critique")
    double scoreHypothesis(@V("topic") String topic, @V("hypothesis") String hypothesis, @V("critique") String critique);
}
```

These agents will be all provided with a tool capable of performing research on scientific literature, for instance downloading academic papers from arXiv, and then added to the P2P agentic system:

```java
ArxivCrawler arxivCrawler = new ArxivCrawler();

LiteratureAgent literatureAgent = AgenticServices.agentBuilder(LiteratureAgent.class)
        .chatModel(baseModel())
        .tools(arxivCrawler)
        .outputKey("researchFindings")
        .build();
HypothesisAgent hypothesisAgent = AgenticServices.agentBuilder(HypothesisAgent.class)
        .chatModel(baseModel())
        .tools(arxivCrawler)
        .outputKey("hypothesis")
        .build();
CriticAgent criticAgent = AgenticServices.agentBuilder(CriticAgent.class)
        .chatModel(baseModel())
        .tools(arxivCrawler)
        .outputKey("critique")
        .build();
ValidationAgent validationAgent = AgenticServices.agentBuilder(ValidationAgent.class)
        .chatModel(baseModel())
        .tools(arxivCrawler)
        .outputKey("hypothesis")
        .build();
ScorerAgent scorerAgent = AgenticServices.agentBuilder(ScorerAgent.class)
        .chatModel(baseModel())
        .tools(arxivCrawler)
        .outputKey("score")
        .build();

ResearchAgent researcher = AgenticServices.plannerBuilder(ResearchAgent.class)
        .subAgents(literatureAgent, hypothesisAgent, criticAgent, validationAgent, scorerAgent)
        .outputKey("hypothesis")
        .planner(() -> new P2PPlanner(10, agenticScope -> {
            if (!agenticScope.hasState("score")) {
                return false;
            }
            double score = agenticScope.readState("score", 0.0);
            System.out.println("Current hypothesis score: " + score);
            return score >= 0.85;
        }))
        .build();

String hypothesis = researcher.research("black holes");
```

With this configuration the `researcher` p2p coordinator is passed with the topic of the research. At this point the only agent that can be invoked is the `literatureAgent`, because it is the only one that has all its required inputs, in this case the `topic`, present in the `AgenticScope`. The invocation of this agent produces the `researchFindings` variable, which is added to the `AgenticScope` state, and this new variable triggers the invocation of the `HypothesisAgent`. Then this produces a `hypothesis` that in turn triggers the `criticAgent`. Finally, the `ValidationAgent` takes in input both the `hypothesis` and the `critique` and generates a new `hypothesis` that eventually retriggers the other agents again. In the meanwhile the `ScorerAgent` gives a `score` to the `hypothesis` and the process terminates when this `score` is greater than or equal to 0.85, or when a maximum of 10 agents invocations have been performed. The following image summarizes all the agents and variables involved in this execution.

For instance a typical run of this example could terminate because the `ScorerAgent` produced a score above the predetermined threshold

```
Current hypothesis score: 0.95
```

and the final output could be something like:

```
Based on the provided references, here are some key points about stochastic gravitational wave backgrounds (SGWBs) from primordial black holes (PBHs):

1. **Detection Rates and Sources:**
   - The detection rate of gravity waves emitted during parabolic encounters of stellar black holes in globular clusters was estimated by Kocsis et al. [85].
   - Gravitational wave bursts from PBH hyperbolic encounters were discussed by García-Bellido and Nesseris [93].

2. **Energy Emission:**
   - The energy spectrum of gravitational waves from hyperbolic encounters was studied by De Vittori, Jetzer, and Klein [88].
   - Gravitational wave energy emission and detection rates for PBH hyperbolic encounters were analyzed by García-Bellido and Nesseris [90].

3. **Template Banks:**
   - Template banks for gravitational waveforms from coalescing binary black holes (including non-spinning binaries) were developed by Ajith et al. [92].

4. **Constraints on PBHs:**
   - Constraints on primordial black holes were reviewed by Carr, Kohri, Sendouda, and Yokoyama [98].
   - Universal gravitational wave signatures of cosmological solitons were discussed by Lozanov, Sasaki, and Takhistov [100].

5. **Induced SGWBs:**
   - Doubly peaked induced stochastic gravitational wave backgrounds were tested for baryogenesis from primordial black holes by Bhaumik et al. [101].
   - Distinct signatures of spinning PBH domination and evaporation, including doubly peaked gravitational waves, dark relics, and CMB complementarity, were explored by Bhaumik et al. [101].

6. **Future Detectors:**
   - Future detectors like Taiji, LISA, DECIGO, Big Bang Observer, Cosmic Explorer, Einstein Telescope, and KAGRA are expected to contribute significantly to the detection of SGWBs from PBHs.

7. **Pulsar Timing Arrays:**
   - Pulsar timing arrays have been used to search for an isotropic stochastic gravitational wave background [73-75].

8. **Template Banks and Simulations:**
   - Template banks like those developed by Ajith et al. are crucial for matching observed signals with theoretical predictions.
```


glaforge:
Is it still going to be called "Peer to Peer"?
I thought at Devoxx Belgium we felt it deserved a better name?
(I guess we need to find it, though)

mariofusco:
> Is it still going to be called "Peer to Peer"? I thought at Devoxx Belgium we felt it deserved a better name? (I guess we need to find it, though)

To be honest in this pull request there is much more. I defined a generic interface, called `Planner` (for now, but even there I'd appreciate suggestions for better naming) allowing to implement any kind of agentic pattern and reimplemented all workflow patterns and the supervisor in term of it.

The P2P pattern, plus a simplified version of Embabel's GOAP approach, are now in the test folder only to demonstrate how you can create your own agentic pattern and use it in combination with the existing ones.

I'm still doing minor adjustments to this pull request and I'm hoping to find some time today to write a comprehensive documentation of this rework, but it would be great to have some early review on what I've done. Any feedback, comment or constructive critique is very welcome.

/cc @LizeRaes @glaforge @jdubois @agoncal

LizeRaes:
@mariofusco this looks fantastic! I really like how the `Planner` abstraction allows for full control over orchestration, plus it makes advanced logging and tracing straightforward to implement. I’ll try to build a couple of test planners next week to explore this in practice, but it’s looking great!
For naming, I like `Planner` as is. If not, `Orchestrator`?

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,280 @@ String result = (String) bankSupervisor.invoke(input);
...
+```java
+public interface Planner {
+
+    default void init(AgenticScope agenticScope, AgentInstance plannerAgent, List<AgentInstance> subagents) { }
+
+    default Action firstAction(AgenticScope agenticScope) {
+        return nextAction(agenticScope, null);
+    }
+
+    Action nextAction(AgenticScope agenticScope, AgentExecution lastAgentExecution);
```
glaforge: Shouldn't `lastAgentExecution` be called `previousAgentExecution`?

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,280 @@ String result = (String) bankSupervisor.invoke(input);
...
+}
+```
+
+This interface has three methods: `init`, `firstAction`, and `nextAction`. The `init` method is called once at the beginning of the execution, and can be used to initialize any state or data structures needed by the planner. The `firstAction` method is called to determine the first action to be taken by the agentic pattern, while the `nextAction` method is called after each agent execution to determine the next action to be taken based on the current state of the `AgenticScope` and the result of the last agent execution.
```
glaforge: Same comment on "last" vs "previous", because I think you mean the previous execution, the one right before the current one? Am I correct? (I'm a bit confused by the meaning of "last" otherwise)

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,280 @@ String result = (String) bankSupervisor.invoke(input);
...
+## Custom agentic patterns
+
+The agentic patterns discussed so far are provided out-of-the-box by the `langchain4j-agentic` module, but what if none of them fit the specific needs of your application? In this case it is possible to create your own custom pattern, that orchestrates the interactions among a set of subagents in a way that is tailored to your requirements.
+
+In more details an agentic pattern is simply the specification of an execution plan for the subagents that it coordinates. This plan can be defined by implementing the following `Planner` interface:
+
+```java
+public interface Planner {
+
+    default void init(AgenticScope agenticScope, AgentInstance plannerAgent, List<AgentInstance> subagents) { }
```
glaforge: If we have an `init()` method to set some state, should we have some kind of `finish()` or `end()` method to clean afterwards?

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,280 @@ String result = (String) bankSupervisor.invoke(input);
...
+    }
+
+    Action nextAction(AgenticScope agenticScope, AgentExecution lastAgentExecution);
+}
+```
+
+This interface has three methods: `init`, `firstAction`, and `nextAction`. The `init` method is called once at the beginning of the execution, and can be used to initialize any state or data structures needed by the planner. The `firstAction` method is called to determine the first action to be taken by the agentic pattern, while the `nextAction` method is called after each agent execution to determine the next action to be taken based on the current state of the `AgenticScope` and the result of the last agent execution.
+
+The `Action` class returned by the `firstAction` and `nextAction` methods represents the next step to be taken by the agentic pattern, and can be one either a list of one or more subagents to be called next, or a signal that the execution has been completed. If the action specifies only one subagent invocation, then it will be executed sequentially, and in the same thread that is executing the planner itself, while if they are more than one, they will be executed in parallel using the provided `Executor` or the langchain4j default one.
```
glaforge: Typo: "if they are more" --> "if there are more".
And I'd spell LangChain4j with uppercase for the L and C :-)

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,280 @@ String result = (String) bankSupervisor.invoke(input);
...
+
+These agents will be all provided with a tool capable of performing research on scientific literature, for instance downloading academic papers from arXiv, and then added to the P2P agentic system:
+
+```java
+ArxivCrawler arxivCrawler = new ArxivCrawler();
+
+LiteratureAgent literatureAgent = AgenticServices.agentBuilder(LiteratureAgent.class)
+        .chatModel(baseModel())
+        .tools(arxivCrawler)
+        .outputName("researchFindings")
```
glaforge: Hasn't it been renamed to `outputKey()`?

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,280 @@ String result = (String) bankSupervisor.invoke(input);
...
+        .build();
+HypothesisAgent hypothesisAgent = AgenticServices.agentBuilder(HypothesisAgent.class)
+        .chatModel(baseModel())
+        .tools(arxivCrawler)
```
glaforge, line 1505: The other agents also need to have access to the Arxiv crawler?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
```diff
@@ -0,0 +1,297 @@
...
+                case "evictAgenticScope" -> registry.evict(args[0]);
+                default ->
+                    throw new UnsupportedOperationException(
+                            "Unknown method on AgenticScopeAccess class : " + method.getName());
+            };
+        }
+
+        if (method.getDeclaringClass() == AgentSpecification.class) {
+            return switch (method.getName()) {
+                case "name" -> name;
```
glaforge, line 129: Maybe use some String constants or enum to avoid typos in strings, to make it easier for later refactorings should they happen?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
```diff
@@ -0,0 +1,184 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.util.stream.Collectors.toMap;
+
+public class SupervisorPlanner implements Planner, ChatMemoryAccessProvider {
+
+    private static final Logger LOG = LoggerFactory.getLogger(SupervisorPlanner.class);
+    public static final String SUPERVISOR_CONTEXT_KEY = "supervisorContext";
+    public static final String SUPERVISOR_CONTEXT_PREFIX = "Use the following supervisor context to better understand "
```
glaforge, line 31: Just a thought, but I'm often a bit worried of the situation where developers use / write agents in one language, but some of the key _infrastructure_ prompts are in English.
Maybe it's a bit of a theoretical problem, and LLMs are polyglot nowadays, but I'm always wondering if we shouldn't be extracting those constants somewhere to be able to override them / localize them...

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/ConditionalPlanner.java
```diff
@@ -0,0 +1,29 @@
...
+public record ConditionalPlanner(List<ConditionalAgent> conditionalAgents) implements Planner {
+
+    record ConditionalAgent(Predicate<AgenticScope> condition, List<AgentInstance> agentInstances) { }
+
+    @Override
+    public Action firstAction(AgenticScope agenticScope) {
+        List<AgentInstance> agentsToCall = conditionalAgents.stream()
+                .filter(conditionalAgent -> conditionalAgent.condition.test(agenticScope))
+                .flatMap(conditionalAgent -> conditionalAgent.agentInstances.stream())
+                .toList();
```
glaforge, line 21: Interesting! When thinking of conditional planning, I was thinking of getting just one action to be triggered, but indeed, that's a sub-list of agents! Neat & powerful!

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/p2p/researcher/ArxivCrawler.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+    }
+
+    private static byte[] downloadUrl(String pdfUrl) throws URISyntaxException, IOException {
+        URLConnection urlConn = new URI(pdfUrl).toURL().openConnection();
+        try (BufferedInputStream in = new BufferedInputStream(urlConn.getInputStream())) {
+            ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+            byte dataBuffer[] = new byte[4096];
+            int bytesRead;
+            while ((bytesRead = in.read(dataBuffer, 0, 4096)) != -1) {
```
glaforge: There's an `InputStream.readAllBytes()` method you could use here instead of winging your own.

glaforge (commented):
That's a big PR to review!

I really like this idea of a higher level abstraction that is able to represent all the agentic patterns. It's nice that you could find a simple enough interface to do that.

I left a few comments here and there. Nothing serious.
I noticed a few typos in the documentation, but stopped adding comments for the following ones I could spot. We can do a pass on the documentation later on.

I'm wondering if the `Planner` really needs a `firstAction()`, or if we could do with just the `nextAction()`. It's a little bit simpler, I guess, to have both though, implementation-wise.  But some preparation needed in implementations of `firstAction()` could be done in the `init()`. Or there could be a `boolean` flag as parameter of `nextAction()` that would indicate if it's the first action or not. But yeah, again, that complexifies the life of the implementor, maybe.

Very interesting use case with the goal oriented planner. It shows this higher level planner abstraction is pretty powerful and can be used to build advanced planners. 
In ADK, for instance (although I haven't played with that so far), you'd have to create custom coded agents, and put all the logic in there, but I find that less abstracted away. I feel this planner approach is a more expressive and readable one.
You could easily do a LangGraph approach as well there. Lots of possibilities!

There's one little thing that I've been wondering, which I find a bit brittle in the P2P planner. We are relying on strings for input/output. Users could easily make typos, and an agent wait for ever on a different string to be present in the agentic scope.
LangChain4j is pretty much type-safe everywhere, but here, we rely on string keys in the agentic scope.
I guess one could create some string constant fields, or some enum values, etc, to circumvent this risk, but I thought I should mention this here.

Really great job on this PR, I love this planner abstraction!

docs/docs/tutorials/agents.md
mariofusco: Good point, renamed.

docs/docs/tutorials/agents.md
mariofusco: Yes, correct. Fixed.

docs/docs/tutorials/agents.md
mariofusco: I don't know to be honest. The `init` method is needed mostly to initialize the planner with the set of subagents it manages. I've never had the need of a `finish` method so far, so for this situation I prefer to follow the 1st rule of any good API designer: "if in doubt leave it out".

docs/docs/tutorials/agents.md
mariofusco: Fixed, thanks.

docs/docs/tutorials/agents.md
mariofusco: Fixed.

docs/docs/tutorials/agents.md
mariofusco: I don't know to be honest :) In my example all agents have the possibility to make their own research on arxiv, not sure if it makes sense, but it's only an example, so I don't think this is very relevant?

docs/docs/tutorials/agents.md
glaforge: OK fair enough! 😄 
And nowadays with default methods, we could add it later, should it be needed.

docs/docs/tutorials/agents.md
glaforge: I was just curious, and didn't really expect the tool to be available to the others as well.
But nothing critical here, no worries, not relevant.

mariofusco:
> That's a big PR to review!
> 
> I really like this idea of a higher level abstraction that is able to represent all the agentic patterns. It's nice that you could find a simple enough interface to do that.

Thanks a lot for your insightful review.

> I'm wondering if the `Planner` really needs a `firstAction()`, or if we could do with just the `nextAction()`. It's a little bit simpler, I guess, to have both though, implementation-wise. But some preparation needed in implementations of `firstAction()` could be done in the `init()`. Or there could be a `boolean` flag as parameter of `nextAction()` that would indicate if it's the first action or not. But yeah, again, that complexifies the life of the implementor, maybe.

I found it useful and clearer to have a dedicated `firstAction()` method in many planner implementations. Anyway it has a `default` implementation that delegates to the `nextAction()` one, so if you want to avoid implementing it and do everything in the `nextAction()` you can already do so. In this case you also have a obvious way to know if this is the first action, because this is the only case when the `previousAction` is `null`.
 
> There's one little thing that I've been wondering, which I find a bit brittle in the P2P planner. We are relying on strings for input/output. Users could easily make typos, and an agent wait for ever on a different string to be present in the agentic scope. LangChain4j is pretty much type-safe everywhere, but here, we rely on string keys in the agentic scope. I guess one could create some string constant fields, or some enum values, etc, to circumvent this risk, but I thought I should mention this here.

This is a good point and I need to think a bit more on this. Anyway note that (at least for now) I'm using both the P2P planner and the GOAP one only as additional planner implementations examples and indeed I moved them in the test folder.
 
> Really great job on this PR, I love this planner abstraction!

Thanks again!

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
mariofusco: This is an interesting observation, even though maybe out of scope for this PR. I mean this PR didn't change anything in this regard. Also note that I have [a test trying to invoke the supervisor in Italian](https://github.com/langchain4j/langchain4j/blob/abcf5a7f8204374fa344996b062a87b6bbf04cd5/langchain4j-agentic/src/test/java/dev/langchain4j/agentic/SupervisorAgentIT.java#L319) and it seems to work pretty well. Of course it would be good to have more tests for this aspect.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/p2p/researcher/ArxivCrawler.java
mariofusco: I just learned it! Thanks a lot.

kdubois:
This is a really nice and powerful way of letting the user create their own patterns!  I like how you then showed 2 implementations of already known patterns to demonstrate its usefulness.  

A few thoughts:
1. I agree with @glaforge that using strings to take action from the AgenticScope feels a bit brittle and a bit "un-Java" like :)
2. Not relevant to the implementation, but I found the diagrams kind of hard to follow; it would be more clear if it was a diagram that shows the flow/sequence in which agents are called based on what variables it required and which ones are available. 
3. Are you planning on adding the GOAP and P2P implementations as new 'core' patterns or just leaving them as examples on how one could add new patterns using the generic agentic planner?

glaforge:
@kdubois I was also wondering about this: that would be nice to have those patterns as part of core, and not just as samples. But maybe @mariofusco wants to take time to refine them before integrating them? Or should there be dedicated modules for such implementations? What's the idea?

mariofusco:
>Are you planning on adding the GOAP and P2P implementations as new 'core' patterns or just leaving them as examples on how one could add new patterns using the generic agentic planner?

> @kdubois I was also wondering about this: that would be nice to have those patterns as part of core, and not just as samples. But maybe @mariofusco wants to take time to refine them before integrating them? Or should there be dedicated modules for such implementations? What's the idea?

Originally I implemented those patterns only to put at work this `Planner` idea and provide two practical examples. I believe that it wouldn't be a good idea to pollute the core agentic module with them, but now I'm thinking to create a separated optional module where I will collect these alternative `Planner` implementations and others that we will hopefully add in future. Or maybe it would be even better to have one more module for each of the different patterns, so users can selectively cherry-pick only the ones in which they are interested?

glaforge:
Maybe not a new module per planner, but one with all the planners could be nice.

mariofusco:
> Maybe not a new module per planner, but one with all the planners could be nice.

I moved both GOAP and P2P implementations into a new module called `langchain4j-agentic-patterns`.

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,479 @@ String result = (String) bankSupervisor.invoke(input);
...
+```java
+public interface Planner {
+
+    default void init(AgenticScope agenticScope, AgentInstance plannerAgent, List<AgentInstance> subagents) { }
+
+    default Action firstAction(AgenticScope agenticScope) {
+        return nextAction(agenticScope, null);
+    }
+
+    Action nextAction(AgenticScope agenticScope, AgentExecution previousAgentExecution);
```
dliubarskyi: Shouldn't it accept a `List<AgentExecution>` (instead of a single `AgentExecution`), to allow for the case when multiple agents were called in parallel by the previous `firstAction`/`nextAction` invocation?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgentExecution.java
```diff
@@ -0,0 +1,7 @@
+package dev.langchain4j.agentic.scope;
+
+import dev.langchain4j.agentic.internal.AgentSpecification;
+import java.util.Map;
+
+public record AgentExecution(AgentSpecification agentSpec, Object agent, Map<String, Object> inputs, Object output) {
```
dliubarskyi: There is an `AgentInvocation` already, and it has `Map<String, String> arguments`. `AgentExecution` contains `Map<String, Object> inputs`. Perhaps these maps should have consistent types and names? Perhaps even `AgentExecution` should contain `AgentInvocation` inside instead of a raw map?

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,479 @@ String result = (String) bankSupervisor.invoke(input);
...
+The agentic patterns discussed so far are provided out-of-the-box by the `langchain4j-agentic` module, but what if none of them fit the specific needs of your application? In this case it is possible to create your own custom pattern, that orchestrates the interactions among a set of subagents in a way that is tailored to your requirements.
+
+In more details an agentic pattern is simply the specification of an execution plan for the subagents that it coordinates. This plan can be defined by implementing the following `Planner` interface:
+
+```java
+public interface Planner {
+
+    default void init(AgenticScope agenticScope, AgentInstance plannerAgent, List<AgentInstance> subagents) { }
+
+    default Action firstAction(AgenticScope agenticScope) {
```
dliubarskyi: I am wondering if for future-proofing it makes sense to accept a "PlannerRequest" object that contains `AgenticScope` and whatever else will be required/added in the future. Otherwise we will have to do breaking changes or introduce overloaded versions of methods that will become confusing. Same concern for every method inputs and output.

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,479 @@ String result = (String) bankSupervisor.invoke(input);
...
+    }
+
+    Action nextAction(AgenticScope agenticScope, AgentExecution previousAgentExecution);
+}
+```
+
+This interface has three methods: `init`, `firstAction`, and `nextAction`. The `init` method is called once at the beginning of the execution, and can be used to initialize any state or data structures needed by the planner. The `firstAction` method is called to determine the first action to be taken by the agentic pattern, while the `nextAction` method is called after each agent execution to determine the next action to be taken based on the current state of the `AgenticScope` and the result of the previous agent execution.
+
+The `Action` class returned by the `firstAction` and `nextAction` methods represents the next step to be taken by the agentic pattern, and can be one either a list of one or more subagents to be called next, or a signal that the execution has been completed. If the action specifies only one subagent invocation, then it will be executed sequentially, and in the same thread that is executing the planner itself, while if there are more than one, they will be executed in parallel using the provided `Executor` or the LangChain4j default one.
```
dliubarskyi, line 1137: Perhaps it would be great to have an option to specify whether to run multiple agents sequentially or in parallel

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgentExecution.java
mariofusco: Good catch, I found that the new `AgentExecution` class was redundant and only kept the `AgentInvocation` one.

docs/docs/tutorials/agents.md
mariofusco: Agreed, I introduced the new `PlannerRequest` object. My only concern is that I'm not totally happy with this name, but I couldn't find a better one. Any other suggestion?

docs/docs/tutorials/agents.md
mariofusco: Actually, what happens at the moment is that this method is called every time that a subagent completed its execution, regardless if it has been called in parallel with other agents or not. So if 3 agents are called in parallel, this method is called 3 times. To be honest I don't know if this is correct behaviour, I'm afraid that this is very questionable. Do you think that it should be called only once with the list of the 3 invoked agents?

mariofusco: I realized that in case of parallel invocations I wasn't collecting all the generated actions but only using the last one. I fixed this, and now I believe that `nextAction` should be invoked at the end of agent invocation, regardless if parallel or not, as it is dong now. WDYT?

docs/docs/tutorials/agents.md
mariofusco: I think that if you want sequential execution you simply return the first agent to be invoked next and then wait to the `nextAction` to be called again to execute the subsequent one as you normally do. At the moment I don't see any valid reason to return a list of agent invocation in one shoot that you however you want to execute sequentially, and until I don't a use case for this I would prefer to leave it out.

docs/docs/tutorials/agents.md
dliubarskyi: I am not sure about the name either, maybe something like `NextActionRequest`? I would probably first settle on the shape of the API, then think about good names at the very end, right before merge

docs/docs/tutorials/agents.md
dliubarskyi: > now I believe that nextAction should be invoked at the end of agent invocation, regardless if parallel or not, as it is dong now. WDYT?

Do you mean this method should be called once for multiple agents? So if previous action requested execution of 2 agents in parallel, then this method should be called once both finished? Seems fair.

If multiple agents were requested in parallel and this method is called after each of them is done, then this means that the rest of the agentic system will be done in parallel without an option to "join".

langchain4j-agentic-patterns/src/test/resources/logback-test.xml
dliubarskyi, line 1: It should probably be in `src/test` instead of `src/main`?

docs/docs/tutorials/agents.md
mariofusco: > Do you mean this method should be called once for multiple agents? So if previous action requested execution of 2 agents in parallel, then this method should be called once both finished? Seems fair.
> 
> If multiple agents were requested in parallel and this method is called after each of them is done, then this means that the rest of the agentic system will be done in parallel without an option to "join".

Sorry I didn't explain myself very well. I mean that at the moment the `nextAction` method is called every time a single agent terminates its execution regardless if it is running in parallel with other agents or not. So yes this method is called after each of them is done and if you want to join their execution, the user can simply implement this in the method itself.

langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/p2p/P2PPlanner.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    @Override
+    public void init(AgenticScope agenticScope, AgentInstance plannerAgent, List<AgentInstance> subagents) {
+        this.agentActivators = subagents.stream().collect(toMap(AgentInstance::agentId, AgentActivator::new));
+    }
+
+    @Override
+    public Action firstAction(PlannerRequest plannerRequest) {
+        if (plannerRequest.agenticScope().hasState("p2pRequest")) {
```
dliubarskyi: nit: Perhaps `p2pRequest` should be a constant?

langchain4j-agentic-patterns/src/test/resources/logback-test.xml
mariofusco: You're right, moved.

langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/p2p/P2PPlanner.java
mariofusco: Agreed, done.

docs/docs/tutorials/agents.md
dliubarskyi: Thinking about it a bit more:

- call `nextAction` once for all agents:
  - pros: it is easy to "join" after all agents are done and decide what to do next. Probably the most common use case, but I might be wrong.
  - cons: it is not possible to trigger something in background as soon as specific agent is done (although can be modelled that this agent is a nested agentic system with 2 agents called in sequence, so it is not a blocker), so need to wait untill all agents are done, and only then decide what to do next.
- call `nextAction` once for every agent:
  - pros: more flexible, users could implement more complicated patterns
  - cons: more low-level; AFAIU there is currently no way for the user to say "do not do anything yet, wait for all agents to be done", user has to return some `Action`.

dliubarskyi: My bad, there is `Action.noOp()` which I missed

dliubarskyi: BTW, there seems to be no test verifying that `noOp()` works as expected, probably worth testing it

mariofusco: Correct, if an user wants to wait for all parallel agents to complete can return a `noOp` and wait for all them to join. I agree that this requires a bit more work on the user's side, but the main goal of this rework is to give them as much flexibiilty as possible.

dliubarskyi: Agreed, more flexibility is key

docs/docs/tutorials/agents.md
dliubarskyi: Makes sense, thanks for clarification

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,479 @@ String result = (String) bankSupervisor.invoke(input);
...
+    private List<AgentInstance> agents;
+
+    @Override
+    public void init(AgenticScope agenticScope, AgentInstance plannerAgent, List<AgentInstance> subagents) {
+        this.agents = subagents;
+    }
+
+    @Override
+    public Action firstAction(AgenticScope agenticScope) {
+        return call(agents);
```
dliubarskyi, line 1153: nit: Perhaps for more clarity when reading the code it would make sense to have separate `call(AgentInstance)` and `callParallel(List<AgentInstance>)` methods?

langchain4j-agentic-patterns/src/test/java/dev/langchain4j/agentic/patterns/goap/horoscope/GoapBasedHoroscopeWriter.java
```diff
@@ -0,0 +1,49 @@
...
+
+import dev.langchain4j.agentic.AgenticServices;
+import dev.langchain4j.agentic.UntypedAgent;
+import dev.langchain4j.agentic.patterns.goap.GoalOrientedPlanner;
+
+import java.util.Map;
+
+import static dev.langchain4j.agentic.patterns.Models.baseModel;
+
+public class GoapBasedHoroscopeWriter {
```
dliubarskyi, line 11: Something is fishy here, I have run it and it was running for a very long time, then failed with:

```
Exception in thread "main" dev.langchain4j.agentic.agent.AgentInvocationException: Failed to invoke agent method: public abstract java.lang.String dev.langchain4j.agentic.patterns.goap.horoscope.StoryFinder.findStory(dev.langchain4j.agentic.patterns.goap.horoscope.Person,java.lang.String)
	at dev.langchain4j.agentic.internal.AgentInvoker.internalInvoke(AgentInvoker.java:52)
	at dev.langchain4j.agentic.internal.AgentInvoker.invoke(AgentInvoker.java:38)
	at dev.langchain4j.agentic.internal.AgentExecutor.internalExecute(AgentExecutor.java:55)
	at dev.langchain4j.agentic.internal.AgentExecutor.execute(AgentExecutor.java:31)
	at dev.langchain4j.agentic.internal.AgentExecutor.execute(AgentExecutor.java:19)
	at dev.langchain4j.agentic.internal.PlannerBasedInvocationHandler$PlannerLoop.loop(PlannerBasedInvocationHandler.java:225)
	at dev.langchain4j.agentic.internal.PlannerBasedInvocationHandler.executeAgentMethod(PlannerBasedInvocationHandler.java:191)
	at dev.langchain4j.agentic.internal.PlannerBasedInvocationHandler.invoke(PlannerBasedInvocationHandler.java:170)
	at jdk.proxy2/jdk.proxy2.$Proxy18.invoke(Unknown Source)
	at dev.langchain4j.agentic.patterns.goap.horoscope.GoapBasedHoroscopeWriter.main(GoapBasedHoroscopeWriter.java:47)
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
	at dev.langchain4j.agentic.internal.AgentInvoker.internalInvoke(AgentInvoker.java:50)
	... 9 more
Caused by: java.lang.reflect.UndeclaredThrowableException
	at jdk.proxy2/jdk.proxy2.$Proxy17.findStory(Unknown Source)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	... 11 more
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:118)
	at dev.langchain4j.agentic.agent.AgentInvocationHandler.invoke(AgentInvocationHandler.java:111)
	... 13 more
Caused by: java.lang.RuntimeException: Something is wrong, exceeded 100 sequential tool executions
	at dev.langchain4j.internal.Exceptions.runtime(Exceptions.java:36)
	at dev.langchain4j.service.tool.ToolService.executeInferenceAndToolsLoop(ToolService.java:222)
	at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:392)
	at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:236)
	at jdk.proxy2/jdk.proxy2.$Proxy16.findStory(Unknown Source)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	... 15 more
```

mariofusco: This is very weird. I just run it again with both [Ollama (qwen2.5:7b)](https://gist.github.com/mariofusco/da2690ef9f7888f6b575be8bfcf214a8) and [OpenAI (gpt-4o-mini)](https://gist.github.com/mariofusco/9fc790f0722b6a775ca494f5fb9c07a7) and it worked for me.

Can you please provide the complete output of your execution?

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,479 @@ String result = (String) bankSupervisor.invoke(input);
...
+## Custom agentic patterns
+
+The agentic patterns discussed so far are provided out-of-the-box by the `langchain4j-agentic` module, but what if none of them fit the specific needs of your application? In this case it is possible to create your own custom pattern, that orchestrates the interactions among a set of subagents in a way that is tailored to your requirements.
+
+In more details an agentic pattern is simply the specification of an execution plan for the subagents that it coordinates. This plan can be defined by implementing the following `Planner` interface:
+
+```java
+public interface Planner {
+
+    default void init(AgenticScope agenticScope, AgentInstance plannerAgent, List<AgentInstance> subagents) { }
```
dliubarskyi: It seems that this method might also be extended to accept more data in the future, perhaps worth introducing a `InitPlannerRequest` (not sure what is the good name)?

docs/docs/tutorials/agents.md
mariofusco: Great. I also added a test demonstrating how to join parallel executions and the use of noOp action.

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,479 @@ String result = (String) bankSupervisor.invoke(input);
...
+
+As anticipated, here the goal coincides with the final output of the planner-based agentic pattern itself, while the path from the initial state to the goal is calculated using a `GoalOrientedSearchGraph`, that is built analyzing the input and output keys of all subagents. The sequence of agents to be invoked is then calculated as the shortest path on that graph from the current state to the desired goal.
+
+To give a practical example of how this works, let's try to build a goal-oriented agentic system that can extract the name and zodiac sign of a person from a prompt, generate the horoscope for that sign, look for a related story on the internet and finally create a nice writeup combining all this information. We can achieve this set of tasks by using the following 5 agents:
+
+```java
+public interface HoroscopeGenerator {
+    @SystemMessage(
+            """
+            You are an astrologist that generates horoscopes based on the user's name and zodiac sign.
```
dliubarskyi: nit: could be on a single line for a better readebility

docs/docs/tutorials/agents.md
```diff
@@ -1111,6 +1111,479 @@ String result = (String) bankSupervisor.invoke(input);
...
+            """)
+    @UserMessage("""
+            Generate the horoscope for {{person}} who is a {{sign}}.
```
dliubarskyi: nit: could be on a single line for a better readebility

langchain4j-agentic-patterns/src/test/java/dev/langchain4j/agentic/patterns/goap/horoscope/GoapBasedHoroscopeWriter.java
dliubarskyi: I cannot reproduce it any more, it works now :/

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/Planner.java
```diff
@@ -0,0 +1,37 @@
...
+import dev.langchain4j.agentic.scope.AgentInvocation;
+import dev.langchain4j.agentic.scope.AgenticScope;
+import java.util.List;
+
+public interface Planner {
+
+    default void init(AgenticScope agenticScope, AgentInstance plannerAgent, List<AgentInstance> subagents) { }
+
+    default Action firstAction(PlannerRequest plannerRequest) {
+        return nextAction(plannerRequest);
```
dliubarskyi: now that both `firstAction` and `nextAction` have the same inputs and outputs, `firstAction` might be redundant

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/Action.java
```diff
@@ -0,0 +1,82 @@
...
+
+        @Override
+        public Object result() {
+            return result;
+        }
+    }
+
+    class AgentCallAction implements Action {
+
+        static final Action NO_OP = new AgentCallAction();
```
dliubarskyi: Perhaps `NO_OP` deserves it's own class like `NoOpAction`?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/Planner.java
mariofusco: This is a discussion that I already did with @glaforge and in theory I would agree, but in practice I found quite handy to have the possibility to distinguish the 2 cases in a few situations, for instance:

- The `ParallelPlanner` can invoke all subagents in parallel in the `firstAction` and simply `done()` for all the next.
- The `ConditionalPlanner` has a very similar situation where I do the routing in the `firstAction` and then it's done.
- The `GoalOrientedPlanner` need to calculate the sequence of action to be invoked only the first time, and then it can behave like a sequence planner for all subsequent invocations.

In other cases the `firstAction` doesn't require a specific implementation and indeed I left it unimplemented using the default one that simply delegate to the `nextAction`. In essence the default implementation of the `firstAction` method gives you the freedom to choose if you need to perform some specific action only the first time or not and I'd prefer to leave it.

dliubarskyi: Makes sense, thanks!

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgentExecutionListener.java
```diff
@@ -0,0 +1,8 @@
+package dev.langchain4j.agentic.scope;
+
+public interface AgentExecutionListener {
```
dliubarskyi:
```suggestion
public interface AgentInvocationListener {
```

mariofusco: Renamed.

docs/docs/tutorials/agents.md
mariofusco: I rename `PlannerRequest` to `PlanningContext` and introduced a `InitPlanningContext` for this. If you're happy with this names I will update the docs accordingly.

dliubarskyi:
Yeah good names!

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/Action.java
mariofusco: Done.

docs/docs/tutorials/agents.md
mariofusco: If I do this I should introduce the possibility to call multiple agents both in parallel and sequentially and, for what discussed in a different thread, I don't want to do this until I don't have the evidence of the necessity of calling multiple agents sequentially in a single call. I'd prefer to leave it as it is.

docs/docs/tutorials/agents.md
mariofusco: Fixed.

docs/docs/tutorials/agents.md
mariofusco: Fixed.

docs/docs/tutorials/agents.md
dliubarskyi: > If I do this I should introduce the possibility to call multiple agents both in parallel and sequentially

Not necessarily, I am just suggesting renaming one of the `call` methods to `callParallel` (without any change to behaviour) to make it clearer

mariofusco: Note that the method takes a varargs and if it is passed with a single agent it calls it sequentially and in the same thread, so renaming it `callParallel` would be quite misleading.

mariofusco:
> Yeah good names!

Good, I also updated the documentation accordingly.

docs/docs/tutorials/agents.md
dliubarskyi: I meant something like this:

```java
    default Action call(AgentInstance agent) {
        return callParallel(agent);
    }

    default Action callParallel(AgentInstance... agents) {
        return new Action.AgentCallAction(agents);
    }

    default Action callParallel(List<AgentInstance> agents) {
        return callParallel(agents.toArray(new AgentInstance[0]));
    }
```

dliubarskyi: So when user types "call", it shows:
image

From the dropdown it is clear that user can either call single agent, or call multiple in parallel

dliubarskyi: But it is a minor thing, feel free to ignore

mariofusco: My problem is that both `callParallel(AgentInstance... agents)` and `callParallel(List<AgentInstance> agents)` can be called with a single agent and in that case it is not parallel at all and everything is synchronous in the same thread. I could change the signature of the first to enforce that it is called with at least 2 agents, something like `callParallel(AgentInstance firstAgent, AgentInstance secondAgent, AgentInstance... agents)` but there is nothing that I can do about the second that indeed can be used with one or more agent, as for instance the `ConditionalPlanner` already does.

dliubarskyi:
What about validating that there are at least 2 agents?

mariofusco:
> What about validating that there are at least 2 agents?

Look at the implementation of the `ConditionalPlanner` where multiple conditions can be active at the same time or a single condition can activate multiple agents
https://github.com/langchain4j/langchain4j/pull/3929/files#diff-a0ca03f7913581f22dff8a11292606cdb89abaec167d0d407baee259d7c18e21R22

I believe that it would be a bit awkward to invoke 2 different methods depending on the number of agents that are present in that list.

dliubarskyi (dismissed):
@mariofusco great job, thanks a lot!

The only thing left is to fix the license check:

> Error:  Failed to execute goal org.honton.chas:license-maven-plugin:0.0.6:compliance (default-cli) on project langchain4j-agentic: ch.qos.logback:logback-classic:jar:1.5.18 does not have acceptable license [{name:Eclipse Public License - v 1.0,url:[http://www.eclipse.org/legal/epl-v10.html},{name:GNU](http://www.eclipse.org/legal/epl-v10.html%7D,%7Bname:GNU) Lesser General Public License,url:[http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html}]](http://www.gnu.org/licenses/old-licenses/lgpl-2.1.html%7D]) -> [Help 1]

Probably worth adding "Eclipse Public License - v 1.0" variant to the `license-maven-plugin` in `langchain4j-parent/pom.xml`. Or maybe use another logging library with more permissive license

mariofusco:
> Probably worth adding "Eclipse Public License - v 1.0" variant to the `license-maven-plugin` in `langchain4j-parent/pom.xml`. Or maybe use another logging library with more permissive license

I realized that I don't need that dependency and I removed it.

jabrena:
Hi @mariofusco, 

Nice abstraction, with this approach how to solve a runtime issue if an agent doesn´t finish/reach the goal or reach a timeout? 
How to handle the "error handling" in this case?

Juan Antonio

PR #3936, opened 2025-10-22
Remove model from GeminiGenerateContentRequest

Bram--:
## Change
Removed the model parameter from the `GeminiGenerateContentRequest`.

The model is a path variable only, there's no need to set this in the request body - it appears it's not even documented any more: https://ai.google.dev/api/generate-content#method:-models.generatecontent

I've also changed `GeminiGenerateContentRequest` to be a record to prevent missing members in equals/hashCode/toString implementations

## General checklist

### DIFF
--- a/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/BaseGeminiChatModel.java
+++ b/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/BaseGeminiChatModel.java
@@ -107,7 +107,6 @@ class BaseGeminiChatModel {
         }
 
         return GeminiGenerateContentRequest.builder()
-                .model(chatRequest.modelName())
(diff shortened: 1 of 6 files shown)


dliubarskyi (approved):
@Bram-- thank you!

dliubarskyi:
I have just discovered that this change broke the `GoogleAiGeminiTokenCountEstimatorIT.should_estimate_token_count_for_tool_specs`. Apparently `GoogleAiGeminiTokenCountEstimatorIT` was not running because the pattern in `@EnabledIfEnvironmentVariable` was too strict and my API key did not match it.

The test fails with:
```
{
  "error": {
    "code": 400,
    "message": "* CountTokensRequest.generate_content_request.model: model is not specified\n",
    "status": "INVALID_ARGUMENT"
  }
}
```

I am going to return the model parameter back

PR #3937, opened 2025-10-23
Reinforce immutability (map)

jeanbisutti:
With the code of `main`, if the original map changes after the copy, then the copied map is also updated.
 
 This [added test](https://github.com/langchain4j/langchain4j/blob/3ca486098673a2b6ec30c5066946572348fe9416/langchain4j-core/src/test/java/dev/langchain4j/internal/UtilsTest.java#L333) fails with the code of `main`.
 
 As a consequence, some LangChain4j objects can change after the object instanciation. I have updated a `CustomMessage` test:  https://github.com/langchain4j/langchain4j/blob/3ca486098673a2b6ec30c5066946572348fe9416/langchain4j-core/src/test/java/dev/langchain4j/data/message/CustomMessageTest.java#L19 It fails with the code of `main`.

Please also [see](https://github.com/langchain4j/langchain4j/pull/3937#discussion_r2454553429).

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
@@ -444,8 +444,8 @@ public class Utils {
         if (map == null) {
             return Map.of();
         }
-
-        return unmodifiableMap(map);
+        // LinkedHashMap to preserve iteration order (important for deterministic toString)
+        return unmodifiableMap(new java.util.LinkedHashMap<>(map));
     }
 
     /**
--- a/langchain4j-core/src/test/java/dev/langchain4j/data/message/CustomMessageTest.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/data/message/CustomMessageTest.java
@@ -1,5 +1,7 @@
 package dev.langchain4j.data.message;
 
+import static org.assertj.core.api.Assertions.entry;
+
 import java.util.LinkedHashMap;
 import java.util.Map;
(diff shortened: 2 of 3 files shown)


langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
```diff
@@ -444,8 +444,8 @@ public static <K, V> Map<K, V> copy(Map<K, V> map) {
         if (map == null) {
             return Map.of();
         }
-
-        return unmodifiableMap(map);
+        // LinkedHashMap to preserve iteration order (important for deterministic toString)
```
jeanbisutti, line 447: A `CustomMessageTest` test has failed once with my computer and the code of `main` because of this:
```
[ERROR]   CustomMessageTest.methods:19
Expecting actual's toString() to return:
"CustomMessage { attributes = {content=The sky is blue., myAttribute=myValue} }"
but was:
"CustomMessage { attributes = {myAttribute=myValue, content=The sky is blue.} }"
```

dliubarskyi:
@jeanbisutti good catch, thank you!

dliubarskyi (commented):
BTW the same issue applies to other `copy*(*)` utility methods.

I am hesitant to merge this now, as these utils are used everywhere in the code, the risk is high.
I can also imagine there are users that misuse current implementation to dynamically change things in runtime, and I can't blame them, because there are no other ways to do that.
So for now I would probably keep it as it is and reconsider this change for the 2.0 release, WDYT?

PR #3940, opened 2025-10-23
Fix NPE with OpenAiChatRequestParameters

jeanbisutti:
This [added test](https://github.com/langchain4j/langchain4j/blob/6df084ddd571672a84707582d4a734d2bcfdf22f/langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiChatRequestParametersTest.java#L72) is failing with the code on `main`:

`java.lang.NullPointerException: Cannot invoke "dev.langchain4j.model.openai.OpenAiChatModelName.toString()" because "modelName" is null`

It seems that spotless was never applied on `OpenAiChatRequestParametersTest`. See [this second commit ](https://github.com/langchain4j/langchain4j/pull/3940/commits/d1bda13c5b5a7c3a8f030eff692335ae915f8493) to fix the spotless CI error.


sunyuhan1998 (approved):

dliubarskyi (approved):
@jeanbisutti thank you!

PR #3954, opened 2025-10-24
Fix langchain4j-test artifact versioning

### LINKED ISSUE #3953: [BUG] langchain4j-test artifact in maven central doesn't have correct version
The `langchain4j-test` artifact in Maven central doesn't have the correct versioning. I would expect to see versions similar to `langchain4j-core`, but instead I only see beta versions.

https://central.sonatype.com/artifact/dev.langchain4j/langchain4j-test

Image

edeandrea:
I think this is the correct fix for #3953

Fixes #3953

### DIFF
--- a/langchain4j-test/pom.xml
+++ b/langchain4j-test/pom.xml
@@ -10,6 +10,7 @@
     </parent>
 
     <artifactId>langchain4j-test</artifactId>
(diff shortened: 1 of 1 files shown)


dliubarskyi:
I guess this can be closed?

PR #3955, opened 2025-10-24
Flatten request response objects

Bram--:
## Change
Refactors Many of the google-ai-gemini value classes. 
Loads of classes were (some still are) using manual implementations of equals, hashCode and toString - this seemed to have happened when removing lombok? Furthermore, every interface and value class had its own file in the root package. This makes the package, IMHO, quite overwhelming and hard to navigate. I've folded a bunch of package-private classes and converted them into records. I've only really touched the `GenerateContentResponse` and `GenerateContentRequest` objects.

### Changes: 
 * Removed several unused members and properties which was causing unnecessary (de)serialization of data/classes. 
 * Change all classes/objects within `GenerateContentResponse` and `GenerateContentRequest` to java `record`s
 * Move all direct children in `GenerateContentResponse` from having their own files - i.e. files are removed in favor of nesting them within the `GenerateContentResponse`.
* Move all direct children in `GenerateContentRequest` from having their own files - i.e. files are removed in favor of nesting them  within the `GenerateContentRequest` - only if they are exclusively used in the Request.
* Moved `GeminiPart` to be inner class of `GeminiContent` - Part is exclusively used in `GeminiContent`.
* Moved `GeminiBlob`, `GeminiFunctionCall`, `GeminiFunctionResponse`, `GeminiFileData`, `GeminiExecutableCode`, `GeminiCodeExecutionResult` to be inner classes of `GeminiPart` - again exclusively used here.
* Removed a bunch of Builders that were not used or only had 1-2 members (resulting in quite a bit of boilerplate)
 

## General checklist

- [x] I have manually run all the unit and integration tests in the [core]


glaforge (approved):
Looks great!
Thanks for cleaning this!

mariofusco (approved):

PR #3960, opened 2025-10-27
Refactor ToolExecutionResult to support lazy evaluation of result tex…

### LINKED ISSUE #3869: [FEATURE] Build/Calculate ToolExecution.result() on demand (lazily)
When using IMMEDIATE tools, I found that I'm rarely interested in ToolExecution.result().  So, marshalling the value of a tool method response to JSON is unnecessary performance hit (albeit, small hit). 

ToolExecution.result() string should only be calculated on demand (lazily).

hemanthsavasere:
Refactor ToolExecutionResult to support lazy evaluation

## Issue
Closes #3869

## Changes

1. Refactor ToolExecutionResult to support lazy evaluation of result text and add corresponding unit tests.
2. Added integration tests for performance optimization

## General checklist

- [] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolExecutionResult.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolExecutionResult.java
@@ -1,7 +1,5 @@
 package dev.langchain4j.service.tool;
 
-import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
-
 import java.util.Objects;
 import java.util.function.Supplier;
 
@@ -20,7 +18,7 @@ public class ToolExecutionResult {
     public ToolExecutionResult(Builder builder) {
         this.isError = builder.isError;
         this.result = builder.result;
-        
+
         // If resultText is provided directly, use it; otherwise use the supplier
         if (builder.resultText != null) {
             this.resultText = builder.resultText;
@@ -85,11 +83,10 @@ public class ToolExecutionResult {
 
     @Override
     public String toString() {
-        return "ToolExecutionResult{" +
-                "isError=" + isError +
-                ", result=" + result +
-                ", resultText='" + resultText() + '\'' +
-                '}';
+        return "ToolExecutionResult{" + "isError="
+                + isError + ", result="
+                + result + ", resultText='"
+                + resultText() + '\'' + '}';
     }
 
     public static Builder builder() {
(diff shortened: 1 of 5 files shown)


hemanthsavasere:
@mariofusco Can you please review the pull request

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolExecutionResult.java
```diff
@@ -41,10 +51,18 @@ public Object result() {
     /**
      * Returns the tool execution result as text.
      * It is a {@link #result()} that is serialized into JSON string.
+     * The text is calculated lazily on first access and then cached.
      *
      * @see #result()
      */
     public String resultText() {
+        if (resultText == null && resultTextSupplier != null) {
+            synchronized (this) {
```
mariofusco: In general we want to avoid using `synchronized` blocks in LangChain4j (also because they don't play well with virtual threads). In this specific situation I believe that you could avoid using any lock simply wrapping the `resultText` in an `AtomicReference<String>`.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolExecutionResult.java
```diff
@@ -56,14 +57,14 @@ public Object result() {
...
-            }
+        String text = resultText.get();
+        if (text == null && resultTextSupplier != null) {
+            text = resultTextSupplier.get();
+            resultText.compareAndSet(null, text);
+            // Re-read to ensure we return the winning value in case of race
+            text = resultText.get();
         }
-        return resultText;
+        return text;
```
mariofusco: I believe that this whole method could be implemented with a one-liner as it follows (please double check):

```
  public String resultText() {
      return resultText.updateAndGet(text -> text != null ? text : resultTextSupplier.get());
  }
```

hemanthsavasere: Agreed, I have done the changes

hemanthsavasere:
@dliubarskyi Can you please review the PR. Thanks

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolExecutionResult.java
```diff
@@ -94,6 +114,21 @@ public Builder result(Object result) {
...
+            return this;
+        }
+
+        /**
+         * Sets a supplier for lazy calculation of the result text.
+         * The supplier will be called only when {@link ToolExecutionResult#resultText()} is first accessed.
+         *
+         * @param resultTextSupplier the supplier to calculate result text on demand
+         * @return this builder
+         * @since 1.6.0
```
dliubarskyi:
```suggestion
         * @since 1.9.0
```

dliubarskyi (approved):
@hemanthsavasere thank you!

PR #3962, opened 2025-10-27
fix(openai): 处理空响应列表时的完成原因解析问题- 在访问 choices 列表前增加空值检查- 避免当响应中 choices…

balajinima:
## Problem Description

In the `OpenAiChatModel.java` file, within the [doChat](langchain4j\langchain4j-core\src\main\java\dev\langchain4j\model\chat\ChatModel.java#L54-L56) method, there is a potential risk of a `NullPointerException`. The specific issue occurs on line 158:

```java
openAiResponse.choices()
```

This code directly calls the `openAiResponse.choices()` method without performing a null check on the return value of [choices()](langchain4j\langchain4j-open-ai\src\main\java\dev\langchain4j\model\openai\internal\chat\ChatCompletionResponse.java#L58-L60).
If the `choices()` method returns `null`, any subsequent call to `.get(0)` will throw a `NullPointerException`.

### Analysis

In the current implementation:

```java
.finishReason(finishReasonFrom(openAiResponse.choices().get(0).finishReason()))
```

If `openAiResponse.choices()` returns `null` or the [choices()](langchain4j\langchain4j-open-ai\src\main\java\dev\langchain4j\model\openai\internal\chat\ChatCompletionResponse.java#L58-L60) list is empty, calling `.get(0)` will cause a `NullPointerException`.

### Suggested Fix

A null and size check should be added before accessing the list elements to improve code robustness.
The following improvements are recommended:

1. Check whether the return value of [choices()](langchain4j\langchain4j-open-ai\src\main\java\dev\langchain4j\model\openai\internal\chat\ChatCompletionResponse.java#L58-L60) is `null`
2. Verify that the list is not empty
3. Safely access the first element only after validation

These checks can prevent the application from crashing when the API response is incomplete or malformed.


dliubarskyi (commented):
Thank you, but this fix does not make much sense, as we always expect choices to be present, otherwise it will fail in other critical place (when we create `AiMessage`).

PR #3964, opened 2025-10-27
feat: support scheduled workflow

vesense:
### Scheduled workflow

Automated periodic business is a common scenario in our daily lives, such as checking business reports every day, reading summary news every morning, and so on.
Langchain4j provides a scheduled workflow to support this feature. Here's an example, we have three sub agents:
 - `BusinessData` agent retrieves business data;
 - `DataAnalysis` agent is based on analyzing large models to obtain professional analysis reports;
 - `Sender` agent will send the analysis report to the relevant personnel.

```java
public class BusinessData {
    @Agent(value = "business data", outputName = "data")
    public String getData() {
        ......
    }
}

public interface DataAnalysis {
    @UserMessage(
            """
                    Summary the business data(it's a json-like string), and give a professional report.
                    The data is: '{{data}}'.
                    """)
    @Agent(value = "summary report", outputName = "content")
    String summary(@V("data") String data);
}

public class Sender {
    @Agent(value = "send content to peoples")
    public void send(@V("content") String content) {
        ......
    }
}
```

In a scheduled workflow, the scheduling period can be set using cron expression `cron("0 0 8 */1 * ?")`, and the maximum number of scheduled executions can be set using `maxIterations(5)`.

```java
        DataAnalysis dataAnalysis = spy(AgenticServices.agentBuilder(DataAnalysis.class)
                .chatModel(baseModel())
                .outputName("content")
                .build());

        UntypedAgent novelCreator = AgenticServices.scheduledBuilder()
                .cron("0 0 8 * * ?") 
                .maxIterations(30)
                .subAgents(new BusinessData(), dataAnalysis, new Sender())
                .build();

        novelCreator.invoke(new HashMap<>());
```

## Issue

Closes #

## Change

support scheduled workflow

## General checklist


vesense:
@mariofusco @dliubarskyi 
This PR has refactored based on the agentic planner(https://github.com/langchain4j/langchain4j/pull/3929).

mariofusco:
I'm sorry but I really don't see the point of this feature. First of all I don't understand in which situation you may want to use it and even if so couldn't the user simply wrap the agent invocation in a cron on his own and obtain the same result?

That said I find also wrong that you're passing the `AgentInvocationListener` in the `PlanningContext` and even worse that the `Planner` is triggering an agent execution, while this should be done exclusively by the execution layer.

Finally I'd prefer that any future `Planner` implementation would be added to the new `langchain4j-agentic-patterns` module leaving this core agentic module as minimal as possible.

PR #3975, opened 2025-10-30
refactor(AiServiceStreamingResponseHandler): Update tool invocation handling logic and optimize concurrent management

### LINKED ISSUE #3973: [BUG] Concurrent tool execution behaves sequentially in AiServiceStreamingResponseHandler
**Describe the bug**

First, I'd like to thank the community for supporting the `executeToolsConcurrently` functionality. However, during my actual testing, I've observed that the behavior does not meet my expectations. The issue appears to be in the following code segment of the `dev.langchain4j.service.AiServiceStreamingResponseHandler` class:

Specifically, in the for loop that executes `toolExecutionFuture.get()`, this causes sequential blocking execution rather than true concurrent execution.

**Log and Stack trace**

No logs or stack traces are needed for this issue, as the problem is evident from the code logic.

**To Reproduce**

1. Set up concurrent tool execution using the `executeToolsConcurrently` feature
2. Execute multiple tools that should run concurrently
3. Observe that the execution time is significantly longer than expected
4. Notice that the tools appear to be executing sequentially, even when running in different virtual threads

**Expected behavior**

I expect that when using the concurrent tool execution feature, all tools should execute truly concurrently rather than sequentially. The execution time should reflect the longest-running tool rather than the sum of all tool execution times.

- LangChain4j version: 1.8.0
- LLM(s) used: Qwen-max
- Java version: 21
- Spring Boot version (if applicable): 3.3.8

**Additional context**

The core issue is in the code block where `toolExecutionFuture.get()` is called sequentially in a loop, which blocks until each future completes before moving to the next one. This defeats the purpose of concurrent execution, even when the futures are submitted to an executor service for parallel processing.

The problem occurs specifically in this code pattern:
```java
for (Future<ToolRequestResult> toolExecutionFuture : toolExecutionFutures) {
    try {
        ToolRequestResult toolRequestResult = toolExecutionFuture.get(); // This blocks sequentially
        // ... rest of the processing
    } catch (ExecutionException e) {
        // ... exception handling
    } catch (InterruptedException e) {
        // ... exception handling
    }
}

### PR DESCRIPTION
**Pull request by Ivan97**
…andling logic and optimize concurrent management

## Issue

Closes #3973

## Change

This PR fixes the sequential execution issue in `AiServiceStreamingResponseHandler` where `executeToolsConcurrently` was not truly executing tools concurrently due to sequential `Future.get()` calls.

Key changes:
- Replaced sequential `Future.get()` processing with proper concurrent execution mechanism
- Updated tool invocation handling logic to wait for all futures to complete concurrently
- Improved asynchronous processing flow for better performance
- Enhanced code readability and maintainability

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -35,15 +35,10 @@ import dev.langchain4j.service.tool.ToolExecutionErrorHandler;
 import dev.langchain4j.service.tool.ToolExecutionResult;
 import dev.langchain4j.service.tool.ToolExecutor;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.Queue;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
-import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
@@ -88,8 +83,7 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
     private final List<String> responseBuffer = new ArrayList<>();
     private final boolean hasOutputGuardrails;
 
-    private record ToolRequestResult(ToolExecutionRequest request, ToolExecutionResult result) {
-    }
+    private record ToolRequestResult(ToolExecutionRequest request, ToolExecutionResult result) {}
 
     AiServiceStreamingResponseHandler(
             ChatExecutor chatExecutor,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by Ivan97** [comment]
@dliubarskyi hi，how's your vacation? please help me to review this PR. Thank you.

File: langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java [hunk 1]
```diff
@@ -184,18 +180,7 @@ public void onPartialThinking(PartialThinking partialThinking, PartialThinkingCo
...
-            var future = CompletableFuture.supplyAsync(
-                    () -> {
-                        ToolExecutionResult toolResult = execute(toolRequest);
-                        return new ToolRequestResult(toolRequest, toolResult);
-                    },
-                    toolExecutor);
-            toolExecutionFutures.add(future);
-        }
-    }
+    public void onCompleteToolCall(CompleteToolCall completeToolCall) {}
```
dliubarskyi, line 183: Why changing this? It is better to start executing a tool as soon as possible.

dliubarskyi (commented):
I am not sure I understand the problem and a solution. Also, there are no tests to verify that old or new implementation is correct or wrong

Ivan97:
> I am not sure I understand the problem and a solution. Also, there are no tests to verify that old or new implementation is correct or wrong

I will write an example project to illustrate whether the current code is executing concurrently correctly. Just a moment.

Ivan97:
> I am not sure I understand the problem and a solution. Also, there are no tests to verify that old or new implementation is correct or wrong

Ah, I think I made a mistake. I've figured out where I got confused. You're right, my friend, the functionality is working fine. I'll close this PR.

dliubarskyi:
@Ivan97 np, thanks!

PR #1351, opened 2024-06-23
1184 - (Very) Rough draft of SummarizedTokenWindowChatMemory

johnsosoka:
## Issue
https://github.com/langchain4j/langchain4j/issues/1184

Please Note -- This is just a start for the implementation, more discussion, tests & safety will need to be implemented. I just want to get the ball rolling on a conversation for the implementation.

## Change
This is the beginning of a SummarizedTokenWindowChatMemory...The MR is currently a _very preliminary_ draft.

- Summary is stored in SystemMessage
- Original SystemMessage content is persisted in the new SystemMessage. `SummarizedSystemprompt` structured prompt encapsulates original system message + conversation summary
- Evicted Messages are integrated into an existing summary
- Skips the summarization of tool messages (for now)
- Only supports the original system message, no replacing (for now)

## General checklist

- [ ] There are no breaking changes
- [ ] I have added unit and integration tests for my change
- [ ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new model integration

- [ ] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added my new module in the [BOM](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-bom/pom.xml)

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #4003, opened 2025-11-06
Summarizing chat memory

PaperFly-web:
## Change
Added `SummarizingChatMemory` class to support automatic summarization of chat history.

Features include:
- Automatic summarization when message count exceeds a configurable threshold
- Support for both static and dynamic `maxMessages` and `summarizeThreshold`
- Dynamic system prompt generation per memory ID for context-aware summaries
- Automatic cleanup of orphan `ToolExecutionResultMessage` when associated `AiMessage` is evicted
- Ensures `SystemMessage` is always preserved at the head of memory
- Works with any `ChatMemoryStore` for persistence
- Builder API with clear and flexible configuration
- Comprehensive unit tests covering edge cases and dynamic configuration

Example usage:

```java
ChatMemory chatMemory = SummarizingChatMemory.builder()
    .chatModel(chatModel)
    .maxMessages(4)
    .summarizeThreshold(2)
    .dynamicSystemPrompt(id -> "Dynamic prompt for memory " + id)
    .build();
```

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
+++ b/langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
@@ -170,7 +170,6 @@ public class SummarizingChatMemory implements ChatMemory {
                         You are a conversation summarization assistant. Please read the entire history of interactions between the user and the AI, and generate a concise summary.
     
                         Requirements for the summary:
-                        - Write in natural and fluent Chinese.
                         - Keep the length between 50 and 500 characters.
                         - Accurately capture the main topics, goals, and conclusions discussed.
                         - Exclude any irrelevant dialogue or system messages.
--- a/langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
+++ b/langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
@@ -7,6 +7,7 @@ import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.ChatMessage;
 import dev.langchain4j.data.message.SystemMessage;
 import dev.langchain4j.data.message.ToolExecutionResultMessage;
+import dev.langchain4j.data.message.UserMessage;
 import dev.langchain4j.memory.ChatMemory;
 import dev.langchain4j.model.chat.ChatModel;
 import dev.langchain4j.model.chat.response.ChatResponse;
@@ -20,43 +21,45 @@ import java.util.function.Function;
 
 /**
  * A chat memory implementation that automatically summarizes conversation history
- * when the number of messages exceeds a configurable threshold.
+ * when the number of messages exceeds a configurable limit.
  *
  * <p>This class behaves similarly to {@link MessageWindowChatMemory} but adds automatic
- * summarization capabilities using an underlying {@link ChatModel}.
+ * summarization capabilities through a customizable <b>summary generation function</b>.
  *
  * <p>Key features:
  * <ul>
- *     <li>Supports both static and dynamic configuration of <b>maxMessages</b>:
- *         the maximum number of messages retained before triggering evictions.</li>
- *     <li>Supports both static and dynamic configuration of <b>summarizeThreshold</b>:
- *         the number of messages to remove and summarize once the message limit is exceeded.
- *         Must be greater than 1.</li>
+ *     <li>Supports both static and dynamic configuration of <b>maxMessages</b> —
+ *         the maximum number of messages retained before triggering summarization.</li>
+ *     <li>Supports both static and dynamic configuration of <b>maxMessagesToSummarize</b> —
+ *         the number of messages to remove and summarize once the limit is exceeded.</li>
  *     <li>Supports dynamic generation of <b>system prompts</b> per memory ID,
- *         allowing context-specific instructions for summarization.</li>
+ *         allowing context-specific summarization instructions.</li>
  *     <li>Automatically evicts orphaned {@link ToolExecutionResultMessage}s
  *         when their parent {@link AiMessage} is evicted.</li>
- *     <li>Ensures that {@link SystemMessage}s are preserved at the head of the memory
- *         and never summarized or evicted.</li>
+ *     <li>Ensures that {@link SystemMessage}s are preserved at the beginning of the memory
+ *         and are never summarized or removed.</li>
  *     <li>Integrates with any {@link ChatMemoryStore} to persist memory state.</li>
+ *     <li>Allows injection of a custom summary generation function, or uses the
+ *         <b>default summary function</b> </li>
  * </ul>
  *
  * <p>Usage notes:
  * <ul>
- *     <li>Messages are only summarized when the total message count exceeds <b>maxMessages</b>.</li>
- *     <li>The number of messages removed during summarization is determined by
- *         <b>summarizeThreshold</b>, and the summary is inserted back into the memory.</li>
+ *     <li>Messages are summarized only when their count exceeds <b>maxMessages</b>.</li>
+ *     <li>The number of messages summarized each time is determined by
+ *         <b>maxMessagesToSummarize</b>, and the generated summary is added back into memory.</li>
+ *     <li>You can provide your own summary generation function or use the default one,
+ *         which only requires a {@link ChatModel} instance.</li>
  *     <li>Dynamic configuration functions allow real-time adjustment of
- *         maxMessages, summarizeThreshold, and system prompts for different memory IDs.</li>
+ *         maxMessages, maxMessagesToSummarize, and system prompts for different memory IDs.</li>
  * </ul>
  *
  * <p>Example usage:
  * <pre>{@code
  * SummarizingChatMemory memory = SummarizingChatMemory.builder()
- *     .chatModel(chatModel)
+ *     .defaultGenerateSummaryFunction(chatModel) // use default summarization with a ChatModel
  *     .maxMessages(10)
- *     .summarizeThreshold(3)
- *     .dynamicSystemPrompt(id -> "Summarize conversation for user: " + id)
+ *     .maxMessagesToSummarize(3)
  *     .build();
  * }</pre>
  *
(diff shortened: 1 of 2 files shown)


dliubarskyi:
cc @cescoffier you might be interested in this PR

langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
```diff
@@ -0,0 +1,340 @@
...
+        }
+
+        int insertIndex = (!messages.isEmpty() && messages.get(0) instanceof SystemMessage) ? 1 : 0;
+        messages.add(insertIndex, generateSummary(removedMessages));
+    }
+
+    /**
+     * Generate a concise AI summary for the removed messages.
+     */
+    private AiMessage generateSummary(List<ChatMessage> messages) {
```
dliubarskyi: Not sure that summary should reside in the `AiMessage`. Perhaps better to keep it in the `UserMessage`?

langchain4j/src/test/java/dev/langchain4j/memory/chat/SummarizingChatMemoryTest.java
```diff
@@ -0,0 +1,782 @@
...
+import dev.langchain4j.model.openai.OpenAiChatModel;
+import java.util.List;
+import java.util.stream.Stream;
+import org.assertj.core.api.WithAssertions;
+import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
+@EnabledIfEnvironmentVariable(named = "deepseek-key", matches = ".+")
```
dliubarskyi:
```suggestion
@EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = ".+")
```

langchain4j/src/test/java/dev/langchain4j/memory/chat/SummarizingChatMemoryTest.java
```diff
@@ -0,0 +1,782 @@
...
+                .build();
+
+        // Add the first three messages
+        chatMemory.add(userMessage("hello"));
+        chatMemory.add(userMessage("world"));
+        chatMemory.add(userMessage("I am lisi"));
+
+        // Verify that the first three messages are still retained
+        assertThat(chatMemory.messages())
+                .usingRecursiveFieldByFieldElementComparatorIgnoringFields("id", "timestamp", "createdAt")
```
dliubarskyi: why is this required?

langchain4j/src/test/java/dev/langchain4j/memory/chat/SummarizingChatMemoryTest.java
```diff
@@ -0,0 +1,782 @@
...
+                .containsExactly(userMessage("hello"), userMessage("world"), userMessage("I am lisi"));
+
+        // Add the fourth message, which may trigger summarization logic
```
dliubarskyi: why this comment says that it "may" trigger? This test should be fully deterministic

langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
```diff
@@ -0,0 +1,340 @@
...
+        SystemMessage systemPrompt = SystemMessage.from(systemPromptFunction.apply(id));
+        messages.add(0, systemPrompt);
```
dliubarskyi: this seems odd, why do you insert the summarization message in the memory?

dliubarskyi: my bad - it is not memory, but removed messages. The name of the variable is confusing. In any case it is not the best practice to modify collections passed into method as arguments

langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
```diff
@@ -0,0 +1,340 @@
...
+import dev.langchain4j.store.memory.chat.ChatMemoryStore;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * A chat memory implementation that automatically summarizes conversation history
+ * when the number of messages exceeds a configurable threshold.
```
dliubarskyi:
```suggestion
 * when the number of messages exceeds a configurable `maxMessages`.
```
Otherwise it can be confused with `summarizeThreshold`

dliubarskyi: But in general I think it would be more useful to have `maxTokens` setting instead of `maxMessages`, as it is much more precise and controllable

langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * <p>This class behaves similarly to {@link MessageWindowChatMemory} but adds automatic
+ * summarization capabilities using an underlying {@link ChatModel}.
+ *
+ * <p>Key features:
+ * <ul>
+ *     <li>Supports both static and dynamic configuration of <b>maxMessages</b>:
+ *         the maximum number of messages retained before triggering evictions.</li>
+ *     <li>Supports both static and dynamic configuration of <b>summarizeThreshold</b>:
```
dliubarskyi: I am not sure `summarizeThreshold` is the clear enough name, perhaps something like `maxMessagesToSummarize`?

langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
PaperFly-web: Yes, you're absolutely right — keeping the summary in the UserMessage makes more sense. I’ll update it to use UserMessage instead.

langchain4j/src/test/java/dev/langchain4j/memory/chat/SummarizingChatMemoryTest.java
PaperFly-web: OK, I will replace it.

langchain4j/src/test/java/dev/langchain4j/memory/chat/SummarizingChatMemoryTest.java
PaperFly-web: This is my mistake — I didn’t check this comment carefully. I’ll review it again to make sure it’s correct.This probably won’t be triggered now — it will only be triggered the next time something is added.

langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
PaperFly-web: Okay, create a new copy of the list to avoid modifying the passed parameters.

langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
PaperFly-web: Thank you for your suggestion. Indeed, "maxMessagesToSummarize" has a better semantic meaning, and I will replace it.

langchain4j/src/main/java/dev/langchain4j/memory/chat/SummarizingChatMemory.java
PaperFly-web: Indeed, maxTokens would be better. It can prevent the summary from exceeding the token limit of the large model due to an excessive number of messages. I will switch to using tokens to summarize the conversation.

PaperFly-web: However, if we switch to using maxTokens, it becomes difficult to control whether the generated summary exceeds maxTokens. Therefore, I think we can consider providing an interface function, allowing users to summarize according to their own strategies if they have any.

dliubarskyi:
BTW there was an old PR that tried to do a similar thing, might be worth checking it for ideas: https://github.com/langchain4j/langchain4j/pull/1351

PaperFly-web:
> BTW there was an old PR that tried to do a similar thing, might be worth checking it for ideas: #1351

Thank you for your reminder. I checked it out and feel that his version is a bit messy. He generated system messages through summarization. Logically speaking, system messages should not be modified casually; generally, it's better to generate user messages. However, I still appreciate your reminder. I will carefully look at his code again to gain some inspiration.

douglas-DS:
@PaperFly-web @dliubarskyi  I’m wondering whether using a _maxMessages_ threshold is the best approach here. Most AI frameworks I’ve seen apply summarization based on a token window, typically controlled by a _maxTokens_ setting. This ensures the memory is kept within the configured context size. Would it make more sense to follow that pattern here as well?

dliubarskyi:
@douglas-DS [agreed](https://github.com/langchain4j/langchain4j/pull/4003#discussion_r2519059833)

PaperFly-web:
> @PaperFly-web @dliubarskyi I’m wondering whether using a _maxMessages_ threshold is the best approach here. Most AI frameworks I’ve seen apply summarization based on a token window, typically controlled by a _maxTokens_ setting. This ensures the memory is kept within the configured context size. Would it make more sense to follow that pattern here as well?
@douglas-DS 
Sorry, I've been busy with work lately and forgot to reply. I've also tried using token calculation, but the final summary generated by the large language model cannot be controlled to a specific number of tokens. So, you might have defined summarizing once with more than 1000 tokens, but this summary might generate more than 1000 tokens of data. This would cause a conflict between the summary and the user-defined token threshold. Therefore, I don't have a better solution to this problem at the moment.

dliubarskyi:
Thanks a lot for your PR! Unfortunately, we have to close it. See more details here: https://github.com/langchain4j/langchain4j/issues/5443#issuecomment-4707891571

PR #4021, opened 2025-11-10
vertex-ai-gemini: Expose vertex apiEndpoint in builder

### LINKED ISSUE #4020: [FEATURE] Allow Vertex Gemini to set custom endpoint
I want to use a custom deployment of Gemini accessible via a custom API Endpoint provided by our internal department. The current builder only allows providing project and location while assuming a `-aiplatform.googleapis.com` endpoint.

The underlying VertexAI allows setting the apiEndpoint. If not given it will construct the apiEndpoint based on 
```
           this.apiEndpoint = (String)apiEndpoint.orElse(String.format("%s-aiplatform.googleapis.com", this.location));
```

I would like if the `VertexAiGeminiChatModel` (respectively `VertexAiGeminiStreamingChatModel`) on top of the `projectId` and `location` also would allow setting the `apiEndpoint`.

Partially, looked into how to override the used HttpClient but this felt weird, just because the underlying `VertexAi`-class already has a setter for the apiEndpoint

eiselems:
## Issue
Closes #4020 

## Change
I exposed the vertex ApiEndpoint in the builder for Vertex AI Gemini chat models, allowing users to specify a custom API endpoint.

This enhancement applies to both the synchronous and streaming chat model implementations

## General checklist

### DIFF
--- a/langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiChatModel.java
+++ b/langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiChatModel.java
@@ -24,6 +24,7 @@ import com.google.cloud.vertexai.api.Tool;
 import com.google.cloud.vertexai.api.ToolConfig;
 import com.google.cloud.vertexai.generativeai.GenerativeModel;
 import com.google.cloud.vertexai.generativeai.ResponseHandler;
+import com.google.common.annotations.VisibleForTesting;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.data.message.AiMessage;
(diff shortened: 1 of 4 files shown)


langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiChatModel.java
```diff
@@ -191,6 +192,10 @@ public VertexAiGeminiChatModel(VertexAiGeminiChatModelBuilder builder) {
             vertexAiBuilder.setCredentials(scopedCredentials);
         }
 
+        if (builder.apiEndpoint != null) {
```
eiselems, line 195: This is where I would like to get the most feedback on. What I don't like that much that this is an addition but still you would need to set the location and the project. But coming from other projects I feel this is often done this way.

If not, I can still add some conditional checks that ensure only that both location and project are set OR the endpoint.

Just let me know what is preferrable.

dliubarskyi: This implementation seems reasonable. We cannot always ensure that the user doesn’t shoot themselves in the leg. :)

dliubarskyi (approved):
@eiselems thank you!

eiselems:
@dliubarskyi thanks for the fast and kind feedback 👍

PR #4043, opened 2025-11-13
Propagate root exception when json parsing fails

### LINKED ISSUE #3972: [BUG] JsonParsingUtils: parsing-related exceptions are always ignored
**Describe the bug**

`JsonParsingUtils#extractAndParseJson` ignores exceptions related to JSON parsing. As a result, the `OutputParsingException` instance doesn't have a `cause` (`cause` is `null`).
Otherwise, it's impossible to understand what has happened without a debugger.

**Log and Stack trace**

```
dev.langchain4j.service.output.OutputParsingException: Failed to parse "{
"obfuscatedField": "obfuscatedValue"
}" (base64: "obfuscatedbase64") into
...ObfuscatedClass
at dev.langchain4j.service.output.ParsingUtils.outputParsingException(ParsingUtils.java:107)
at dev.langchain4j.service.output.ParsingUtils.outputParsingException(ParsingUtils.java:103)
at dev.langchain4j.service.output.PojoOutputParser.lambda$parse$0(PojoOutputParser.java:42)
at java.base/java.util.Optional.orElseThrow(Optional.java:403)
at dev.langchain4j.service.output.PojoOutputParser.parse(PojoOutputParser.java:42)
```

**To Reproduce**

Expect JSON that cannot be parsed to a record class.

**Expected behavior**

If an exception happens, there should be a clear understanding of what has happened.
I.e. `OutputParsingException` should have a `cause`.

LangChain4j version: 1.8.0
Java version: 25

mariofusco:
## Issue

Closes #3972

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
@@ -13,15 +13,17 @@ public class JsonParsingUtils {
 
     public record ParsedJson<T>(T value, String json) {}
 
-    public static <T> Optional<ParsedJson<T>> extractAndParseJson(String text, Class<T> type) {
+    public static <T> Either<Exception, ParsedJson<T>> extractAndParseJson(String text, Class<T> type) {
         return extractAndParseJson(text, s -> Json.fromJson(s, type));
     }
 
-    public static <T> Optional<ParsedJson<T>> extractAndParseJson(String text, ThrowingFunction<String, T> parser) {
+    public static <T> Either<Exception, ParsedJson<T>> extractAndParseJson(String text, ThrowingFunction<String, T> parser) {
+        Exception parsingException = null;
         try {
-            return Optional.of(new ParsedJson<>(parser.apply(text), text));
-        } catch (Exception ignored) {
-            // Ignore parsing errors and try to find a JSON block in the text
+            return Either.right(new ParsedJson<>(parser.apply(text), text));
+        } catch (Exception e) {
+            // Temporarily ignore parsing errors and try to find a JSON block in the text
+            parsingException = e;
         }
 
         int index = text.length();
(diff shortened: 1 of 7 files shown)


langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
```diff
@@ -13,33 +13,35 @@ public interface ThrowingFunction<T, R> {
 
     public record ParsedJson<T>(T value, String json) {}
 
-    public static <T> Optional<ParsedJson<T>> extractAndParseJson(String text, Class<T> type) {
+    public static <T> Either<Exception, ParsedJson<T>> extractAndParseJson(String text, Class<T> type) {
```
dliubarskyi, line 16: This looks quite strange. In the positive case, we usually return an object from the method. If exception happens, it is usually thrown. Is there a reason why exception should be returned (and not thrown) in this case?

mariofusco: Yeah, you're right my inner functional programmer made me doing this :) 

Using an `Either` to propagate errors without side effect is quite common in functional programming and sometimes also done in Java ( see for example https://medium.com/@samuelavike/either-monads-in-java-elegant-error-handling-for-the-modern-developer-423bbf7300e6 ), but I understand that this is not very idiomatic of the language.

Usually I don't do this but since it already returned a wrapping type (the `Optional`) it instinctively seemed natural to me to turn the `Optional` into an `Either`, also because I don't like a method that could fail in 2 different ways: either returning an `Optional` but also eventually throwing an Exception.

I fully understand if you don't like this programming style and I can revert what I did here. But if so I could also get rid of the `Optional` and directly returning the wrapped object, since now a failure in the parsing will be simply signaled by the thrown Exception. Let me know which alternative you prefer.

dliubarskyi: Oh, I just got flashbacks to 2015 when I learned Scala :D
Yeah, directly returning the wrapped object or throwing an exception seems like a idiomatic way to do it in Java, could you please do that instead?

mariofusco:
Replaced by https://github.com/langchain4j/langchain4j/pull/4046

PR #4049, opened 2025-11-14
Enable case-insensitive enum deserialization in structured output parsing

### LINKED ISSUE #4045: [FEATURE] Enable case-insensitive enum deserialization in structured output parsing

When using lower-level LLM models (e.g., smaller or less capable ones) with structured output parsing, JSON binding to Java POJOs fails if the model outputs enum values in mixed or inconsistent casing (e.g., "Active" instead of "ACTIVE"). This causes Jackson's ObjectMapper to throw a deserialization error due to case-sensitive enum matching, even though the semantic meaning is correct.

Enable case-insensitive enum deserialization by default in the ObjectMapper used for structured output parsing. This can be achieved by configuring:

`objectMapper.enable(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS);`

This setting is available in Jackson 2.12+ and allows enums to match regardless of case (e.g., "active", "Active", "ACTIVE" all map to MyEnum.ACTIVE).

- Manually normalizing LLM output (e.g., via string replacement or regex) before parsing — fragile and error-prone.
- Requiring users to annotate all enums with @JsonCreator and custom factory methods — verbose and not scalable across projects.
- Injecting a custom ObjectMapper

**Additional context**

This issue primarily affects smaller Ollama LLMs that lack strict schema adherence. Larger models (like GPT-4) typically respect exact casing in structured outputs. Enabling ACCEPT_CASE_INSENSITIVE_ENUMS is a low-overhead, backward-compatible fix that improves robustness without changing existing behavior for well-formed outputs.

mariofusco:
## Issue

Closes #4045

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
@@ -13,8 +13,10 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationContext;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@mariofusco thank you!

PR #3945, opened 2025-10-23
Gemini batch chat model delete and cancel

Bram--:
## Issue

Depends on #3942 
Progresses #3916

## Change
Adds list and delete endpoints for `batchGenerateContent` requests. The actual prod code is pretty straight forward and only needed a single record to be added to work `ListOperations` that wraps the `Operation` objects returned when creating or getting a Batch. The list operation support additional `filter` and `returnPartialSuccess` parameters which are not currently implemented. 

Basically all this PR does is add to methods to the BatchChatModel and pipe them through the GeminiService, which in turn does a simple HTTP GET or DELETE.

## General checklist

- [x] There are no breaking changes (API, behaviour)
- [x] I have added unit and/or integration tests for my change
- [x] The tests cover both positive and negative cases
- [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)

PR #3942, opened 2025-10-23
Gemini Batch API

Bram--:
## Issue
Progresses #3916

## Change
Added `GoogleAiGeminiBatchChatModel` and implemented 3 of the 6 APIs. 

### APIs:
 * `createBatchInline` - Creates a new batch job that returns a PENDING operation/result. This can later be retrieved with `retrievedBatchResults`. [Documentation](https://ai.google.dev/gemini-api/docs/batch-api#create-batch-job)
 * `retrieveBatchResults` - Retrieve a batch job for the given name. This can return a BatchError(Whoops, something went wrong), BatchIncomplete (Meaning it's still pending or running) or a BatchIncomplete which contains the parsed results.
 * `cancelBatchJob` - Returns nothing but will try to cancel a pending or running job, throws otherwise.

### TODO:
 * `deleteBatchJob` - Deletes a batch job no matter the state.
 * `listBatchJobs` - Retrieves all the batch jobs - may use filters and pagination to limit these.
 * `createBatchFromFile` - Will need to implement the Files API as the API takes a file reference.

This means batching works with this implementation but is missing some useful APIs. It also limits users to batch sizes no larger than 20MB as as the file API should be used for anything beyond that.

### Changes/Considerations: 
 * `GoogleAiGeminiBatchChatModel` is the new main API and re-uses the `BaseGeminiChatModel` marked as `@Experimental`
 * `Some helper methods moved from `GoogleAiGeminiChatModel` to `BaseGeminiChatModel`
 * All Request/Response records are grouped in `BatchRequestResponse` to restrict the number of files added to the root package.
  * Minor changes to `GeminiService` to allow different the use of `HttpMethod.GET` and an empty Http body (Both are needed for  `retrieveBatchResults`)
  * Added plenty of unit tests
  * Added minimal Integration tests; The problem is that the time it takes to create/process a batch is non-deterministic which means retrieving erroneous and successful batches is impossible   

## General checklist
- [x] There are no breaking changes (API, behaviour)
- [x] I have added unit and/or integration tests for my change
- [x] The tests cover both positive and negative cases
- [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

PR #4068, opened 2025-11-17
Implement all of the Gemini Files APIs

### LINKED ISSUE #3949: [FEATURE] Add support for Gemini Files API
To support files >=20MB it's required to use the Files API provided by Google. [Documentation](https://ai.google.dev/gemini-api/docs/files)

Have the files API available: upload, list and, delete.

**Additional context**
I've started looking into this as we're planning to use large batches. This is somewhat part of: #3916 and #3774

Bram--:
## Issue
Closes #3949

## Change
This implements the complete Gemini Files API as documented here: https://ai.google.dev/gemini-api/docs/files

***Additions***
 * `GeminiFiles.uploadFile()` - Initiates a resumable upload that gets the URL to upload the file to. After it's successful it uploads the file to the provided URL and returns a `GeminiFile`.
 * `GeminiFiles.getMetadata` - Retrieves meta data for the given file name e.g. `files/upload.txt`. This is the same metadata as returned from the uploadFile() call although the state might have changes.
 * `GeminiFiles.listFiles()` - List all uploaded files. Files are automatically deleted after 48 hours so this call will return all files that were uploaded in the last 48 hours
 * `GeminiFiles.deleteFile()` - Manually delete a file. This might be useful as there's a project limit of 20GB

***TODO***:
 * Add examples / Docs
 * Add File upload API for `GoogleAiGeminiBatchChatModel` (#3942 & #3945)

*HttpClient*: Unfortunately, I ended up using `java.net.http.HttpClient` directly. So no wrapped `HttpClient`. I tried using the client that is available but unfortunately the `body` is hard-coded to be always `String`. As the file upload requires the body to be binary/bytes I opted for using `java.net.http.HttpClient` directly. I had a look into add/changing the `dev.langchain4j.http.client.HttpClient` but ultimately this would a bunch of boilerplate to the response/request objects and having two body methods (e.g. `body(String ...)` and `bodyBytes(...)`) which is not great IMHO. Making the client, request and response objects generic touched over 250 files, so I stayed away from that one :sweat_smile: 

## General checklist


Bram--:
@glaforge Would you mind having a look? Thanks!

glaforge (dismissed):
LGTM!

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiFiles.java
```diff
@@ -0,0 +1,277 @@
...
+ * <p>
+ * The Gemini models support multimodal inputs including text, images, audio, videos, and documents.
+ * Use this API to upload media files when the total request size exceeds 20 MB.
+ *
+ * <p>
+ * Files are stored for 48 hours and can be referenced in content generation requests using their URI.
+ * The API supporjava.net.http.HttpClien
+ * During the retention period, you can retrieve file metadata but cannot download the files directly.
+ */
+final class GeminiFiles {
```
dliubarskyi: The class is package-private, shouldn't it be public? Or is it not intended to be used directly by users?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiFiles.java
```diff
@@ -0,0 +1,277 @@
...
+/**
+ * Service for uploading and managing media files with Google AI Gemini.
+ *
+ * <p>
+ * The Gemini models support multimodal inputs including text, images, audio, videos, and documents.
+ * Use this API to upload media files when the total request size exceeds 20 MB.
+ *
+ * <p>
+ * Files are stored for 48 hours and can be referenced in content generation requests using their URI.
+ * The API supporjava.net.http.HttpClien
```
dliubarskyi: this line is not complete

dliubarskyi (commented):
@Bram-- thank you! Out of curiosity, why changing HttpClient ended up touching 250 files?

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiFiles.java
Bram--: That's embarrassing. Absolutely this should be public. 

Double checked all visibility modifiers and sorted them.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiFiles.java
Bram--: Fixed

Bram--:
> @Bram-- thank you! Out of curiosity, why changing HttpClient ended up touching 250 files?

@dliubarskyi This was with changing HttpClient to be generic (e.g. HttpClient<BodyType>), this consequently requires changes to SuccessfulHttpResponse, HttpRequest and HttpClientBuilder. This actually results in 611 (including test files) and 374 (excluding tests) call sites that would need to be changed. On top of it breaking user impls.

I suppose another way to do it would be keeping those as is and deriving `GenericHttpClient<BodyType>` where existing classes simply implement the the new generic client/builder etc. for <String>.

I think the 250+ came from looking at HttpRequest which would've been the absolutely minimum I would have to change to make this work - as only the body needs to be set to bytes for the Gemini file upload.

dliubarskyi:
@Bram-- But why generics are needed? Can't we just change `String body` to `Object body` in `HttpRequest` and `SuccessfulHttpResponse` and allow builders to set other types instead of `String`?

Bram--:
> @Bram-- But why generics are needed? Can't we just change `String body` to `Object body` in `HttpRequest` and `SuccessfulHttpResponse` and allow builders to set other types instead of `String`?

@dliubarskyi  Perhaps I'm misunderstanding, but this would result in breaking API changes, right? It would also force call sites to do type checks before calling and casting it to a string. Or are you proposing to add different body methods for each type? 

I suppose using any serializable `Object` would work, although it seems to me you would still need to do type checking and mapping it to a `BodyPublisher` in the actual client impl:
```java
        if (request.body() != null) {
            bodyPublisher = BodyPublishers.ofString(request.body());
        } else {
            bodyPublisher = BodyPublishers.noBody();
        }
```



Looking at the Response, using an `Object`:
```java
public class SuccessfulHttpResponse {

    private final int statusCode;
    private final Map<String, List<String>> headers;
    private final Object body;

    // ...
 
    public Object body() {
        return body;
    }
   
    // ...
}
```

Would then require a (preferably checked) cast on each call site. 

Or are you saying you get different methods for different body types?
```java
public class SuccessfulHttpResponse {

    private final int statusCode;
    private final Map<String, List<String>> headers;
    private final Object body;

    // ...
 
    public String bodyAsString() {
        return (String) body;
    } 

    // etc..
   
    // ...
}
```

IMHO generics would be preferable to that as this would allow passing the object around with compile time type checking opposed to implicit type checking at the call site.

```java
public class SuccessfulHttpResponse<T> {

    private final int statusCode;
    private final Map<String, List<String>> headers;
    private final T body;

    // ...
 
    public T body() {
        return body;
    } 

    // ...
}
```

Bram--:
@dliubarskyi Would it be possible/realistic to have this, #4086 and an additional PR that adds methods to to the batch models that takes a File reference, in for 1.9 cut off? Or should I make a contingency plan? 😄 

Thanks again,

dliubarskyi:
@Bram-- I will try to review and merge all PRs today/tomorrow, the release will probably happen tomorrow afternoon

Bram--:
> @Bram-- I will try to review and merge all PRs today/tomorrow, the release will probably happen tomorrow afternoon

@dliubarskyi Don't mean to put any workload on you! Just trying to figure out whether we should have a temporary solution whilst these things make its way in. If cutoff is tomorrow ish we'll be happy to wait and see! 😄 

Please let me know if there is anything I can do to help out or make this easier on you.

Thanks,

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeminiFiles.java
```diff
@@ -0,0 +1,277 @@
...
+    private static final String LIST_FILES_PATH = "/v1beta/files";
+    private static final String UPLOAD_PATH = "/upload/v1beta/files";
+    private static final String API_KEY_HEADER_NAME = "x-goog-api-key";
+    private static final String UPLOAD_URL_HEADER = "x-goog-upload-url";
+
+    private final HttpClient httpClient;
+    private final String baseUrl;
+    private final String apiKey;
+
+    GeminiFiles(String apiKey, @Nullable HttpClient httpClient, @Nullable String baseUrl) {
```
dliubarskyi: ctor is supposed to be public? How users will create it? Ideally ctor should accept a builder, as in other models

Bram--: Added a Builder.

dliubarskyi:
Please check the failures, there are some issues with javadoc

Bram--:
> Please check the failures, there are some issues with javadoc

Added `@param` to `GeminiFiles.GeminiFile`

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GeminiFilesIT.java
```diff
@@ -0,0 +1,431 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class GeminiFilesIT {
```
dliubarskyi, line 18:
```suggestion
@EnabledIfEnvironmentVariable(named = "GOOGLE_AI_GEMINI_API_KEY", matches = ".+")
class GeminiFilesIT {
```

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GeminiFilesIT.java
```diff
@@ -0,0 +1,432 @@
...
+
+import dev.langchain4j.model.googleai.GeminiFiles.GeminiFile;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
```
dliubarskyi, line 15:
```suggestion
import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
import org.junit.jupiter.api.io.TempDir;
```

dliubarskyi (approved):
@Bram-- thank you!

PR #4070, opened 2025-11-17
Refactor value classes to (nested) records in google-ai-gemini

Bram--:
## Change
This further flattens the classes inside the `google-ai-gemini` module. I've started writing changes to implement the Embedding Batch API for Gemini (https://ai.google.dev/gemini-api/docs/batch-api#batch-embedding) but the preexisting batch embedding API makes it difficult to follow, and it would be unclear which one is a async batch and which is synchronous. 

These changes are similar to the ones in #3955 but are around the non chat model requests/responses. After these only (public) models are prefixed with `GoogleAi` everything else is prefixed Gemini. So this leaves the (currently) 4 model services:
 * `GoogleAiEmbeddingModel.java` - Synchronous batching
 * `GoogleAiGeminiChatModel.java` - Default chat model
 * `GoogleAiGeminiStreamingChatModel.java` - Streaming chat model
 * `GoogleAiGeminiTokenCountEstimator.java` - Token counter
 * `GoogleAiGeminiBatchChatModel` - Batching in #3942 
 

## General checklist


glaforge (approved):
LGTM

PR #4086, opened 2025-11-19
Gemini (async) batch embed api

Bram--:
## Issue
Progresses #3916 - This is one of the last PRs for the Gemini Batch API support. The batch API calls using the Files API is still left to do

## Change
This introduces `GoogleAiGeminiBatchEmbeddingModel` which is similar to `GoogleAiGeminiBatchChatModel` - as the batch API is actually the same for both. I opted to split the `GoogleAiGeminiBatchChatModel` into two and creating a generic `GeminiBatchProcessor` that does the actual calling of the batch APIs.  The latter deals with the logic of processing the requests and wrapping the responses in a sealed `BatchResponse`.  The mapping between LangChain4j request/response objects and the Gemini API request/response objects is done in each `GoogleAiGeminiBatch..Model` through a `RequestPreparer<REQUEST, API_REQUEST, API_RESPONSE, RESPONSE>`.

Sonar is giving me a hard time with the full word generics [REQUEST, API_REQUEST, API_RESPONSE, RESPONSE] but IMHO it beats: [R1, R2, R3, R4]. Let me know if you have any opinions on this. 

Other than processor this PR introduces the `GoogleAiGeminiBatchEmbeddingModel`which is exactly the same as the preexisting `GoogleAiGeminiBatchChatModel` but with different [REQUEST, API_REQUEST, API_RESPONSE, RESPONSE] generics.

Finally I choose to make `BatchRequestResponse` public as to expose the `BatchRequestResponse` through where this was previously in `GoogleAiGeminiBatchChatModel`. Meaning this is technically a breaking change.

## General checklist


langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiEmbeddingModelIT.java
```diff
@@ -46,8 +46,8 @@ void should_use_metadata() {
         GoogleAiEmbeddingModel embeddingModel = GoogleAiEmbeddingModel.builder()
                 .apiKey(GOOGLE_AI_GEMINI_API_KEY)
                 .modelName("embedding-001")
-                .logRequests(false) // embeddings are huge in logs
-                .logResponses(false)
+                .logRequests(true) // embeddings are huge in logs
```
dliubarskyi:
```suggestion
                .logRequests(false) // embeddings are huge in logs
```

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiEmbeddingModelIT.java
```diff
@@ -46,8 +46,8 @@ void should_use_metadata() {
...
+                .logResponses(true)
```
dliubarskyi:
```suggestion
                .logResponses(false)
```

dliubarskyi (dismissed):
@Bram-- thank you!

PR #4089, opened 2025-11-19
Receive and expose MCP tool metadata

### LINKED ISSUE #4076: [FEATURE] Support MCP tool annotations
The MCP protocol allows the server to expose additional hints about tools, see: https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations

It's not yet completely clear how langchain4j should consume and work with this metadata. Maybe we could make it part of the `ToolSpecification` class (in which case we will need to make it more generic to avoid tying that class to something MCP-specific) so that applications can access it programmatically and maybe build tool filters based on these flags.
We could maybe also automatically append this data to the tool descriptions that are passed to LLMs (but it's not quite clear how to do this automatically)...

jmartisk:
- Closes: https://github.com/langchain4j/langchain4j/issues/4076

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
@@ -2,8 +2,11 @@ package dev.langchain4j.agent.tool;
 
 import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
 
+import static dev.langchain4j.internal.Utils.copy;
 import static dev.langchain4j.internal.Utils.quoted;
 
 /**
@@ -16,6 +19,7 @@ public class ToolSpecification {
     private final String name;
     private final String description;
     private final JsonObjectSchema parameters;
+    private final Map<String, Object> metadata;
 
     /**
      * Creates a {@link ToolSpecification} from a {@link Builder}.
@@ -26,6 +30,7 @@ public class ToolSpecification {
         this.name = builder.name;
         this.description = builder.description;
         this.parameters = builder.parameters;
+        this.metadata = copy(builder.metadata);
     }
 
     /**
@@ -51,6 +56,13 @@ public class ToolSpecification {
      */
     public JsonObjectSchema parameters() {
         return parameters;
+    }
+
+    /**
+     * Returns the metadata relevant to the tool.
+     */
+    public Map<String, Object> metadata() {
+        return metadata;
     }
 
     @Override
@@ -108,11 +120,13 @@ public class ToolSpecification {
         private String name;
         private String description;
         private JsonObjectSchema parameters;
+        private Map<String, Object> metadata;
 
         /**
          * Creates a {@link Builder}.
          */
         private Builder() {
+            this.metadata = new HashMap<>();
         }
 
         /**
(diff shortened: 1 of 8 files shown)


langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
```diff
@@ -26,6 +29,7 @@ private ToolSpecification(Builder builder) {
         this.name = builder.name;
         this.description = builder.description;
         this.parameters = builder.parameters;
+        this.metadata = builder.metadata;
```
dliubarskyi:
```suggestion
        this.metadata = copy(builder.metadata);
```

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
```diff
@@ -148,6 +161,25 @@ public Builder parameters(JsonObjectSchema parameters) {
             return this;
         }
 
+        /**
+         * Sets the {@code metadata}.
+         *
+         * @param metadata
+         * @return {@code this}
+         */
+        public Builder setMetadata(Map<String, Object> metadata) {
```
dliubarskyi:
```suggestion
        public Builder metadata(Map<String, Object> metadata) {
```

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
```diff
@@ -108,11 +119,13 @@ public static final class Builder {
         private String name;
         private String description;
         private JsonObjectSchema parameters;
+        private Map<String, Object> metadata;
 
         /**
          * Creates a {@link Builder}.
          */
         private Builder() {
+            this.metadata = new HashMap<>();
```
dliubarskyi, line 129: nit: no need to create a new map, `copy()` will create an empty one if needed

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/ToolSpecificationHelperTest.java
```diff
@@ -448,4 +456,58 @@ void arrayWithoutSpecifiedType() throws JsonProcessingException {
...
+                }]
+                """;
+        ArrayNode json = OBJECT_MAPPER.readValue(text, ArrayNode.class);
+        Map<String, Object> metadata = ToolSpecificationHelper.toolSpecificationListFromMcpResponse(json).get(0).metadata();
+
+        ArrayNode expectedArray = OBJECT_MAPPER.createArrayNode().add(1).add(2).add(3);
+        assertThat(metadata.get("example.org/array")).isEqualTo(expectedArray);
+
+        TextNode expectedString = OBJECT_MAPPER.getNodeFactory().textNode("hello");
+        assertThat(metadata.get("example.org/string")).isEqualTo(expectedString);
```
dliubarskyi: can't we expose value as `String` instead of a `TextNode`? Jackson should be able to parse any JSON into a Map<String, Object> where value can be nested map recursively

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
jmartisk: We do need to create it because there's the `addMetadata` method in the builder that writes a single entry into it, so it needs to be instantiated in the builder already

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/ToolSpecificationHelperTest.java
jmartisk: Ok, if you prefer storing something that isn't Jackson-specific, which might be a good idea

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
dliubarskyi: `addMetadata` could create a map lazily if it is not there yet, but this is minor

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/ToolSpecificationHelperTest.java
dliubarskyi: Yes, it would be great to use the most generic types possible

dliubarskyi (approved):
@jmartisk thank you!

PR #4100, opened 2025-11-20
Expose agentic system as tree and check key types consistency

mariofusco:
## Issue

Closes #

## Change

The main goal of this change is giving the possibility to navigate the structure of an agentic system as a tree of agents. I believe that this could be a starting point to provide some minimal UI to visually represent the state of the system. To the same purpose it is also now possible to obtain from the `AgenticScope` a ordered list of all the invocations performed during an execution (before they were splitted by agent, but in this way you didn't have the temporal sequence of all invocations).

As a nice side effect of this change, I'm also now checking the consistency of the types of the keys used in the agentic system, something that the Quarkus extension already does at build time. 

## General checklist


dliubarskyi (approved):
@mariofusco good stuff, thank you! Only one Q: is tree generic enough to represent all possible agentic system topologies? What about P2P (isn't graph needed in this case)?

mariofusco:
> @mariofusco good stuff, thank you! Only one Q: is tree generic enough to represent all possible agentic system topologies? What about P2P (isn't graph needed in this case)?

Logically it is more a graph, but under a topological point of view it is a node of a tree where the P2P planner is the root and its subagents are the leaves (that could eventually be other subtrees). In this sense all planner based patterns must be trees in general, where the planner itself is the root and its subagents are the leaves, so this should cover all possible implementation including the ones made by users.

PR #4106, opened 2025-11-22
MongoDB: Update to the latest version and append framework name to MongoDB metadata 

### LINKED ISSUE #4105: [FEATURE] MongoDriver doesn't send a telemetry related to framework name
Spring boot [sends](https://github.com/spring-projects/spring-boot/blob/a25bcfc6107dc3e62e900018da7c0d2bfa521c72/module/spring-boot-mongodb/src/main/java/org/springframework/boot/mongodb/autoconfigure/MongoClientFactorySupport.java#L51) a telemetry metrics to Mongo Atlas. It would be great to have the same telemetry from langchain4j users who are using MongoDB embedding store 

Spring Boot has a dedicated bean that creates MongoClient with predefined metadata that includes `spring-boot` as a framework name. For langchain4j `MongoClient` is created manually by users and then passed over to `MongoDbEmbessingStore` , within the store we can call `appendMetadata` which is 5.6.0 specific API 

So far `appendMetadata` is the only MongoClient API to allows users to add metadata dynamically 

**Additional context**
The API was introduced by MongoDB starting from 5.6.0, this is the first attempt to use it in external library

strogiyotec:
## Issue
Closes [#4105](https://github.com/langchain4j/langchain4j/issues/4105)

## Change
Since 5.6.0 Mongo driver added support for dynamic client metadata 
For this change: 
1. Upgraded mongo-client version to 5.6.0
2. Added langchain4j to driver name to Client Metadata in MongoDbEmbeddingStore, spring boot already does it [here](https://github.com/spring-projects/spring-boot/blob/c687998f56ff462e1bfb00bd905218a96a4e0315/module/spring-boot-mongodb/src/main/java/org/springframework/boot/mongodb/autoconfigure/MongoClientFactorySupport.java#L62) , however for spring boot `MongoClient` created as a bean but `MongoDbEmbeddingStore` accepts `MongoClient` from the client , so the append happens within `MongoDbEmbeddingStore` constructor 
3. Added integration test to verify client metadata
4. The call to `appendMetadata` is dynamic so that clients with older driver versions can still use it 

## General checklist

- [ X ] There are no breaking changes (API, behaviour)
- [ X ] I have added unit and/or integration tests for my change
- [ X ] The tests cover both positive and negative cases
- [ X ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ X ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ X ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ X ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)


dliubarskyi:
> The call to appendMetadata is dynamic so that clients with older driver versions can still use it

Do you expect users to use an older driver version than the one that comes with LC4j out of the box? Any reason for that?

strogiyotec:
> > The call to appendMetadata is dynamic so that clients with older driver versions can still use it
> 
> Do you expect users to use an older driver version than the one that comes with LC4j out of the box? Any reason for that?

Hi @dliubarskyi , I have two reasons in mind 
## MongoDB official example for Langchain4j
In the [Get Started with Langchain4j integration](https://www.mongodb.com/docs/atlas/ai-integrations/langchain4j/) document, MongoDb provided an example that adds langchain4j dependency along with the client library
```
<dependency>
    <artifactId>langchain4j-mongodb-atlas</artifactId>
    <version>1.1.0</version>
</dependency>
<dependency>
    <artifactId>langchain4j-voyage-ai</artifactId>
    <version>1.1.0-beta7</version>
</dependency>
<dependency>
    <groupId>org.mongodb</groupId>
    <artifactId>mongodb-driver-sync</artifactId>
    <version>5.4.0</version> # this dependency is declared explicitly 
</dependency>
```
Once langchain4j merges this PR we will update the documentation to rely on client version that comes with langchain4j

## Old library usage
According to Mongo Atlas usage, some clients are still using an older versions of the driver, existing customers might be hesitant to switch to the new client version (one of the reasons might be unsupported java 8 version)


Let me know if it makes sense

dliubarskyi (approved):
@strogiyotec thank you!

PR #4122, opened 2025-11-24
Allow to optionally define agentic system variables in a strongly typed way

mariofusco:
## Issue

Closes #

## Change

Allow to optionally define agentic system variables in a strongly typed way binding each variable to a class implementing the new `AgentState` interface.

## General checklist


langchain4j-agentic/src/test/java/dev/langchain4j/agentic/TypedAgentsIT.java
```diff
@@ -0,0 +1,119 @@
...
+    public static class UserRequest implements AgentState<String> { }
+    public static class ExpertResponse implements AgentState<String> { }
+
+    public interface CategoryRouter {
+
+        @UserMessage("""
+            Analyze the following user request and categorize it as 'legal', 'medical' or 'technical'.
+            In case the request doesn't belong to any of those categories categorize it as 'unknown'.
+            Reply with only one of those words and nothing else.
+            The user request is: '{{UserRequest}}'.
```
mariofusco: This is the thing that probably I like the less of this pull request: the name of the template variable has to be the same of the class name defining the agent state. I didn't find a better solution, but I'm open to discuss this.

mariofusco:
@dliubars @geoand @cescoffier please give a look and let me know if you have any suggestion to improve. For now I flagged this as a draft. Once we will be all happy with the API, I will also add the documentation before merging. Any suggestion to improve or simplify is very welcome.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/K.java
```diff
@@ -0,0 +1,14 @@
...
+
+import static java.lang.annotation.ElementType.PARAMETER;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+@Target(PARAMETER)
+@Retention(RUNTIME)
+public @interface K {
```
dliubarskyi, line 42: Please add a javadoc explaining how to use it

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
```diff
@@ -98,6 +100,8 @@ public AgentBuilder(Class<T> agentServiceClass, Method agenticMethod) {
         }
         if (!isNullOrBlank(agent.outputKey())) {
             this.outputKey = agent.outputKey();
+        } else if (agent.typedOutputKey() != Agent.NoAgentState.class) {
```
dliubarskyi: `outputKey`  seems to still having a higher priority than `typedOutputKey`, is this expected?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
```diff
@@ -305,6 +309,10 @@ public AgentBuilder<T> outputKey(String outputKey) {
         return this;
     }
 
+    public AgentBuilder<T> outputKey(Class<? extends AgentState<?>> outputKey) {
```
dliubarskyi: What about renaming `AgentState` into `OutputKeyType` or something like this to make the name of the type and the name of methods/variables consistent

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
mariofusco: Ideally they should never been both used for the same agent. Should I throw a configuration exception if this happens?

dliubarskyi:
@mariofusco TBH I am not sure I understand the idea of this feature, could you please elaborate?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
mariofusco: This is for both input and output. Maybe `AgentKey`? I wanted a name as short as possible.

mariofusco:
> @mariofusco TBH I am not sure I understand the idea of this feature, could you please elaborate?

The idea is to possibly avoid relying on strings for variable names in an agentic system, because a typo can easily break the whole system. Also we wanted something more strongly typed so that for instance

`agenticScope.readState(RequestCategory.class);`

where

`public static class RequestCategory implements AgentState<Category> { }`

would return an object of type `Category`. 

This requirement has been also expressed here https://github.com/quarkiverse/quarkus-langchain4j/issues/1893

Initially I was hoping to use an `enum` for this purpose, but this cannot work both because the `enum` of variables in a specific system should be defined by the user and because I couldn't have the type safety. This is the best solution to fulfill those requirements, but of course I'm open for better suggestions.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
dliubarskyi: Perhaps `TypedKey`? "Agent Key" might mean something else?

dliubarskyi:
> Also we wanted something more strongly typed so that for instance
> 
> agenticScope.readState(RequestCategory.class);
> 
> where
> 
> public static class RequestCategory implements AgentState<Category> { }
> 
> would return an object of type Category.

Can't it be just `agenticScope.readState(Category.class)`? Is wrapping into `AgentState` necessary?

dliubarskyi:
> The idea is to possibly avoid relying on strings for variable names in an agentic system, because a typo can easily break the whole system.

Hmm, but we already validate that all variables are filled inside the `PromptTemplate`.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/K.java
mariofusco: Done.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
mariofusco: Done.

mariofusco:
> Can't it be just `agenticScope.readState(Category.class)`? Is wrapping into `AgentState` necessary?

No, for a twofold reason:

1. The class implementing `AgentState` identifies the key for the whole agentic system
2. `agenticScope.readState(String.class)` could be ambiguous, typically there are many keys with type `String` while having a class per state like in `agenticScope.readState(FirstName.class)` and `agenticScope.readState(LastName.class)` will clearly disambiguate.

mariofusco:
> Hmm, but we already validate that all variables are filled inside the `PromptTemplate`.

True but here is not only about prompt templating. It also involves being sure that the output of an agent is correctly used as the input for another one in the same agentic system.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
mariofusco: Renamed.

mariofusco:
@dliubarskyi I should have addressed all your remarks (thanks for review), please let me know if it makes sense now and if so I will also updated the docs accordingly.

mariofusco:
@dliubarskyi I also added the documentation, this should be ready to be fully reviewed now.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentUtil.java
```diff
@@ -34,8 +37,38 @@ public class AgentUtil {
...
+    private static final Map<Class<? extends TypedKey<?>>, TypedKey<?>> STATE_INSTANCES = new ConcurrentHashMap<>();
+
     private AgentUtil() {}
 
+    private static <T> TypedKey<T> stateInstance(Class<? extends TypedKey<? extends T>> key) {
+        return (TypedKey<T>) STATE_INSTANCES.computeIfAbsent(key, k -> {
+            try {
+                return key.getDeclaredConstructor().newInstance();
+            } catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException e) {
+                throw new RuntimeException(e);
```
dliubarskyi, line 53: nit: consider adding a clarifying error message describing the limitation (no-arg ctor, visibility, etc)

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
```diff
@@ -69,18 +72,18 @@ static AgentInvoker fromMethod(AgentSpecification spec, Method method) {
 
     static String parameterName(Parameter parameter) {
         return optionalParameterName(parameter)
-                .orElseThrow(() -> new IllegalArgumentException("Parameter name not specified and no @P or @V annotation present: " + parameter));
+                .orElseThrow(() -> new IllegalArgumentException("Parameter name not specified and no @V or @K annotation present: " + parameter));
     }
 
     static Optional<String> optionalParameterName(Parameter parameter) {
```
dliubarskyi, line 79: nit: can't `AgenticParameterNameResolver` be reused here?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgenticScope.java
```diff
@@ -27,6 +28,7 @@ public interface AgenticScope extends LangChain4jManaged {
     Object readState(String key);
 
     <T> T readState(String key, T defaultValue);
+    <T> T readState(Class<? extends TypedKey<T>> key);
```
dliubarskyi, line 33: should other methods be added as well? e.g., `hasState`, `writeState`?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgenticParameterNameResolver.java
```diff
@@ -0,0 +1,26 @@
+package dev.langchain4j.agentic.internal;
+
+import java.lang.reflect.Parameter;
+import dev.langchain4j.agentic.declarative.K;
+import dev.langchain4j.service.ParameterNameResolver;
+import dev.langchain4j.service.V;
+
+import static dev.langchain4j.agentic.internal.AgentUtil.stateName;
+
+public class AgenticParameterNameResolver implements ParameterNameResolver {
```
dliubarskyi, line 11:
```suggestion
public class DefaultParameterNameResolver implements AgenticParameterNameResolver {
```
We usually call the default implementation `Default*`

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/TypedAgentsIT.java
dliubarskyi: IDK, seems completely fine to me

dliubarskyi:
Perhaps I am missing something, but I don't understand how adding strong types helps a lot. If I understand correctly, valiudation is happening in runtime and not during compilation and/or creation of agentic system, so how strong typing is better than just strings?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentUtil.java
mariofusco: Done.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentInvoker.java
mariofusco: Done.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgenticParameterNameResolver.java
mariofusco: This is not the default implementation, is the one specific of the agentic module. The default one is in an inner class inside the `ParameterNameResolver` interface.

dliubarskyi: Oh sorry, my bad

mariofusco:
> Perhaps I am missing something, but I don't understand how adding strong types helps a lot. If I understand correctly, valiudation is happening in runtime and not during compilation and/or creation of agentic system, so how strong typing is better than just strings?

That's true, but only because I already added types consistency validation with this other change https://github.com/langchain4j/langchain4j/pull/4100

The advantages are the ones that I tried to clarify in documentation, basically preventing typos in input/output keys and avoiding casts when accessing a state of the `AgenticScope`. I understand that these advantages could seem limited (and I'm open to discuss if it is really useful or not), but being something totally optional I would leave to users the decision if they want to leverage it or not.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgenticScope.java
mariofusco: Good point, done.

dliubarskyi:
> agenticScope.readState(String.class) could be ambiguous, typically there are many keys with type String while having a class per state like in agenticScope.readState(FirstName.class) and agenticScope.readState(LastName.class) will clearly disambiguate.

This is true for `String` inputs/outputs and other common types, but agent outputs (and perhaps inputs) can be of a custom type (e.g., enum or some pojo), for example when expecting structured outputs from the LLM. Peraps we could allow users to use their custom types without the need to wrap them into `TypedKey`? This is a minor thing though.

dliubarskyi:
`Category category = agenticScope.readState(Category.class)` is also more readable/clear than
`Category category = agenticScope.readState(RequestCategory.class)`

dliubarskyi (approved):
@mariofusco thanks a lot!

PR #4127, opened 2025-11-25
Parse structured results in A2A client

### LINKED ISSUE #4107: [BUG] DefaultA2AClientBuilder doesn't handle non-String return types
### Description

`DefaultA2AClientBuilder` currently always returns a `String` from A2A responses, regardless of the actual return type declared in the agent service interface method. This causes `ClassCastException` at runtime when methods have non-String return types (e.g., `Integer`, `Boolean`, POJOs, `List<T>`, etc.).

### Current Behavior

When invoking an agent method with a non-String return type, the proxy always returns a `String`:

```java
// Agent interface
interface MyAgent {
    Integer calculateAge(String birthYear);
    List<String> extractKeywords(String text);
    Person parsePersonInfo(String data);
}

// Usage
MyAgent agent = A2AClient.builder("http://localhost:8080", MyAgent.class).build();
Integer age = agent.calculateAge("1990"); // ClassCastException: String cannot be cast to Integer
```

### Root Cause

In `DefaultA2AClientBuilder.invokeAgent()` method (line 195), the response is always returned as a `String`:

```java
private Object invokeAgent(Object[] args) throws A2AClientException {
    // ... code to send message and process response ...
    
    try {
        String responseText = messageResponse.get();
        LOG.debug("Response: " + responseText);
        return responseText;  // ⚠️ Always returns String!
    } catch (InterruptedException | ExecutionException e) {
        // ...
    }
}
```

The `invoke` method in the `InvocationHandler` doesn't pass the method's return type information to `invokeAgent`, so there's no way to perform proper type conversion.

### Expected Behavior

The builder should convert the A2A response to the appropriate return type, similar to how `DefaultAiServices` handles this:

1. Extract the method's return type
2. Parse the String response according to the return type:
   - Return `String` as-is for `String` return types
   - Parse primitives (`Integer`, `Boolean`, `Double`, etc.)
   - Deserialize JSON for POJOs and collections
   - Handle special types like `void`, `Result<T>`, etc.

mariofusco:
## Issue

Closes #4107

## General checklist

### DIFF
--- a/langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
+++ b/langchain4j-agentic-a2a/src/main/java/dev/langchain4j/agentic/a2a/DefaultA2AClientBuilder.java
@@ -8,6 +8,7 @@ import dev.langchain4j.agentic.agent.AgentResponse;
 import dev.langchain4j.agentic.internal.A2AClientBuilder;
 import dev.langchain4j.agentic.internal.AgentSpecification;
 import dev.langchain4j.agentic.planner.AgentInstance;
+import dev.langchain4j.service.output.ServiceOutputParser;
 import io.a2a.A2A;
 import io.a2a.client.Client;
 import io.a2a.client.ClientEvent;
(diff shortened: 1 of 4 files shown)


dliubarskyi (approved):
@mariofusco thank you!

PR #4138, opened 2025-11-26
Update chat message types count in documentation

ratomidev:
The documentation lists only four chat message types, but there are actually five: the missing one is CustomMessage, a newer type supported in some models like Ollama. This causes confusion or mismatch in the docs.


dliubarskyi (approved):
@ratomidev thanks!

PR #4149, opened 2025-11-28
Test for MCP registry server parsing

jmartisk:
This is a lower-level test that doesn't actually connect to any remote registry. I think it will be useful to have tests that verify JSON parsing from responses. I took a complex sample from https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md for the first test.


dliubarskyi (approved):
@jmartisk thank you!

PR #4153, opened 2025-11-28
Allow to configure a default value for a key in the agentic scope

### LINKED ISSUE #3537: [FEATURE] Allow to set the AiService and AiAgent user message programmatively
Currently, when creating either an AiService or an AiAgent, you can set the user message either via the `@UserMessage` annotation on the method or on one of the parameter.

This is limiting when you would like to build programmatively an AiService or an AiAgent with a dynamic user message.

Having a way inside both the AiService builder AiAgent builder to set a user message.
This could be done as today for the systemMessage provider of an AiService builder.

**Additional context**

mariofusco:
## Issue

Closes #3537

## Change

The intent of this pull request is allowing agents to define default values for one or more of its input argument. This can have a general usefulness, but in particular can be also used as a way to programmatically provide a user message as request here https://github.com/langchain4j/langchain4j/issues/3537

As demonstrated by the added tests now there are 3 different ways to programmatically provide a user message:

1. As an input to the whole agentic system
2. As a default value of a typed argument
3. As a key default value for the agent using the feature introduced by this pull request

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
@@ -35,15 +35,18 @@ import dev.langchain4j.service.tool.ToolExecutionErrorHandler;
 import dev.langchain4j.service.tool.ToolProvider;
 import java.lang.reflect.Method;
 import java.lang.reflect.Proxy;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.Executor;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
 public class AgentBuilder<T> {
     final Class<T> agentServiceClass;
+    final Method agenticMethod;
     final Class<?> agentReturnType;
-    final List<AgentArgument> arguments;
+    List<AgentArgument> arguments;
 
     String name;
     String agentId;
@@ -53,6 +56,8 @@ public class AgentBuilder<T> {
 
     Consumer<AgentRequest> beforeListener = request -> {};
     Consumer<AgentResponse> afterListener = response -> {};
+
+    private final Map<String, Object> defaultValues = new HashMap<>();
 
     private ChatModel model;
     private ChatMemory chatMemory;
@@ -81,8 +86,8 @@ public class AgentBuilder<T> {
 
     public AgentBuilder(Class<T> agentServiceClass, Method agenticMethod) {
         this.agentServiceClass = agentServiceClass;
+        this.agenticMethod = agenticMethod;
         this.agentReturnType = agenticMethod.getReturnType();
-        this.arguments = argumentsFromMethod(agenticMethod);
 
         Agent agent = agenticMethod.getAnnotation(Agent.class);
         if (agent == null) {
@@ -113,6 +118,8 @@ public class AgentBuilder<T> {
     }
 
     T build(DefaultAgenticScope agenticScope) {
+        this.arguments = argumentsFromMethod(agenticMethod, defaultValues);
+
         AiServiceContext context = AiServiceContext.create(agentServiceClass);
         AiServices<T> aiServices = AiServices.builder(context);
         if (model != null) {
@@ -360,6 +367,11 @@ public class AgentBuilder<T> {
 
     public AgentBuilder<T> afterAgentInvocation(Consumer<AgentResponse> afterListener) {
         this.afterListener = this.afterListener.andThen(afterListener);
+        return this;
+    }
+
+    public AgentBuilder<T> keyDefaultValue(String key, Object value) {
+        this.defaultValues.put(key, value);
         return this;
     }
 }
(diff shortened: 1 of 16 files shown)


langchain4j/src/main/java/dev/langchain4j/service/AiServiceContext.java
```diff
@@ -46,6 +57,8 @@ public class AiServiceContext {
 
     public BiFunction<ChatRequest, Object, ChatRequest> chatRequestTransformer = (req, memId) -> req;
 
+    private final Set<Method> validMethods = new HashSet<>();
```
dliubarskyi: Does it make sense to keep references to all methods that are valid? IIUC, if one of them is not valid, we fail anyway?
Perhaps we can just keep a `volatile boolean validated`?

langchain4j/src/main/java/dev/langchain4j/service/AiServiceContext.java
```diff
@@ -77,4 +90,100 @@ public GuardrailService guardrailService() {
         return this.guardrailService.updateAndGet(
                 service -> (service != null) ? service : guardrailServiceBuilder.build());
     }
+
+    void validate() {
```
dliubarskyi: I am not sure validation logic belongs to the AI Service context, perhaps we need a separate class for that?

langchain4j/src/main/java/dev/langchain4j/service/AiServiceContext.java
```diff
@@ -77,4 +90,100 @@ public GuardrailService guardrailService() {
...
+                            "In order to use @MemoryId, please configure the ChatMemoryProvider on the '%s'.",
+                            aiServiceClass.getName());
+                }
+            }
+        }
+    }
+
+    void validateParameters(Method method) {
+        if (!validMethods.add(method)) {
+            return;
```
dliubarskyi: Nice idea to cache the validation, thank you!

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -216,8 +124,7 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                             return handleChatMemoryAccess(method, args);
                         }
 
-                        // TODO do it once, when creating AI Service?
-                        validateParameters(context.aiServiceClass, method);
+                        context.validateParameters(method);
```
```suggestion
                        // TODO do it once, when creating AI Service?
                        context.validateParameters(method);
```
I would still keep the comment as it is better to validate it as soon as possible to fail fast if something is invalid, instead of waiting for the first time AI Service is used

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
```diff
@@ -362,4 +369,13 @@ public AgentBuilder<T> afterAgentInvocation(Consumer<AgentResponse> afterListene
         this.afterListener = this.afterListener.andThen(afterListener);
         return this;
     }
+
+    public AgentBuilder<T> keyDefaultValue(String key, Object value) {
```
dliubarskyi:
```suggestion
    public AgentBuilder<T> defaultKeyValue(String key, Object value) {
```
nit: reads more naturally

langchain4j/src/main/java/dev/langchain4j/service/AiServiceContext.java
dliubarskyi: My bad, please ignore

dliubarskyi (commented):
@mariofusco thank you!

langchain4j/src/main/java/dev/langchain4j/service/AiServiceContext.java
mariofusco: I wasn't sure either, I put it there because it was accessing some of the fields of the context, but yes, probably it would be a good idea to move this to a completely separated `ValidationUtil` class.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: I originally tried to perform the validation at service build time, but this made a few test to fail, and more important would have introduced a backward incompatibility that, has discussed, we want to avoid it for now. I will restore the todo comment, even though the problem is now mitigated by the cache that I introduced.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/AiServiceContext.java
mariofusco: Done.

dliubarskyi (approved):
@mariofusco thanks a lot!

PR #4156, opened 2025-11-30
Whoops; Make GoogleAiGeminiBatchChatModel.builder() public.

Bram--:
## Change
Fixes access to the builder, right now GoogleAiGeminiBatchChatModel cannot be instantiated as the builder is not public 😢 

## General checklist


dliubarskyi (approved):
@Bram-- thanks!

PR #4169, opened 2025-12-02
Support tools definition via Map in agents

### LINKED ISSUE #4167: [FEATURE] AgentBuilder: Support for Tool-Definition via Map<ToolSpecification, ToolExecutor>
I have build some tools dynamically with the help of ToolSpecifictions and their ToolExecutors. I would like to pass these tools into my agent.

Support of the AgentBuilder for the tools(...) with a Map<ToolSpecification, ToolExecutor> as parameter; like the AiService does. At the moment this is not possible, as the passed Map is interpreted as one Object, where the @Tool annotation is searched for, thus I have no tool at all.

Workaround: Using a static ToolProvider, which always returns the same tools, regardless of the input.

mariofusco:
## Issue

Closes #4167

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
@@ -7,7 +7,9 @@ import static dev.langchain4j.agentic.internal.AgentUtil.uniqueAgentName;
 import static dev.langchain4j.internal.Utils.isNullOrBlank;
 
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
+import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.agentic.Agent;
+import dev.langchain4j.agentic.declarative.K;
 import dev.langchain4j.agentic.declarative.TypedKey;
 import dev.langchain4j.agentic.internal.AgentSpecification;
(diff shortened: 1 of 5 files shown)


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
```diff
@@ -241,6 +248,11 @@ public AgentBuilder<T> tools(Object... objectsWithTools) {
         return this;
     }
 
+    public AgentBuilder<T> tools(Map<ToolSpecification, ToolExecutor> toolsMap) {
```
dliubarskyi, line 257: There is also another overload:

`AiServices<T> tools(Map<ToolSpecification, ToolExecutor> tools, Set<String> immediateReturnToolNames)`

Not sure if it makes sense to support it though

dliubarskyi (dismissed):
@mariofusco thank you!

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java
mariofusco: Let me check it before merging.

mariofusco: Ok, I added it, thanks.

PR #4181, opened 2025-12-05
Add observability and monitoring for agentic systems

### LINKED ISSUE #4098: [FEATURE] Add Agent Observability & Listener API (agent lifecycle + model/tool call events)
### Problem

LangChain4j exposes solid observability for AI Services and chat models (e.g., `AiServiceListener` events and `ChatModelListener` hooks for request/response/error), and a `ToolExecutedEvent` exists on the observability side.

However, there isn’t a first-class agent-level listener that lets us trace an agent run (start/complete/error, iterations) and correlate the **nested model calls** and **tool calls** that happen inside that run. This makes it hard to build production-grade tracing, metrics, and UI debugging for the agentic module.

### Request

Introduce a first-class Agent Observability / Listener API for the agentic module that covers the following areas:

**1. Agent lifecycle events**

The context should include an agentRunId, agentName, input(s), timestamps, and (if applicable) a parentRunId for nested agents.
```
onAgentStart(context)

onAgentIteration(context, iterationInfo) (optional but useful for loops)

onAgentComplete(context, result)

onAgentError(context, error)
```

**2. Model call signals inside agents**

Emit structured events (or forward/bridge existing ChatModelListener events) with correlation to the enclosing agentRunId. This allows us to tie LLM activity to a specific agent step.

```
onModelRequest(agentRunId, chatRequest, params)

onModelResponse(agentRunId, chatResponse, metadata)

onModelError(agentRunId, error)
```

Note: This should reuse/align with the existing `ChatModelListener` semantics (messages, params, token usage, finish reasons).

**3. Tool call signals inside agents**

Provide agent-correlated tool events (or bridge the existing `ToolExecutedEvent`) so we can tell when a tool is called during an agent step. Ensure these either wrap or cleanly correlate with `ToolExecutedEvent`.

```
onToolStart(agentRunId, toolName, args)

onToolComplete(agentRunId, toolName, output)

onToolError(agentRunId, toolName, error)
```

4. Trace building

Optionally expose a lightweight trace (e.g., `AgentRunTrace`) that aggregates the above for a run: lifecycle, per-iteration data, model calls, tool calls, timing, and errors. This is highly useful for tests and UI debugging.

### Why this matters

**Production Visibility**: Teams need to know exactly when the model is called and when a tool is called inside an agent, and connect those to the overall agent run for logging, metrics, and debugging.

**Consistency:** Aligns with existing observability patterns (AiService listeners, model listeners, tool events) while adding the missing agent layer and correlations.

**Integration:** Plays nicely with OpenTelemetry/GenAI semantic conventions and existing Java stacks (Micrometer/OTel, Quarkus).

**API sketch (illustrative)**

```
public interface AgentListener {
  void onAgentStart(AgentContext ctx);
  void onAgentIteration(AgentContext ctx, AgentIteration it);
  void onAgentComplete(AgentContext ctx, AgentResult result);
  void onAgentError(AgentContext ctx, Throwable error);

  // Inside-agent signals (correlated with ctx.runId)
  void onModelRequest(AgentContext ctx, ChatModelRequestContext req);
  void onModelResponse(AgentContext ctx, ChatModelResponseContext res);
  void onModelError(AgentContext ctx, ChatModelErrorContext err);

  void onToolStart(AgentContext ctx, String toolName, Object args);
  void onToolComplete(AgentContext ctx, String toolName, Object output);
  void onToolError(AgentContext ctx, String toolName, Throwable error);
}
```

Thanks for considering! Happy to help with a design or PR.

mariofusco:
## Issue

Closes #4098

## Change

This pull request provides a first implementation of the observability features requested here https://github.com/langchain4j/langchain4j/issues/4098

On top of it also introduces a way to monitor agents as discussed here https://github.com/langchain4j/langchain4j/discussions/4008

The following explanation, copied from the updated documentation, provides a recap of how this works.

### Observability

Tracking and logging the agents' invocations can be crucial for debugging and understanding the aggregate behavior of the whole agentic system in which those agents participate. For this reason, the `langchain4j-agentic` module allows to register an `AgenticListener` through the `listener` method of the agent builders, that is notified of all agents invocations and their results, and it is defined as follows:

```java
public interface AgenticListener {

    default void beforeAgentInvocation(AgentRequest agentRequest) { }
    default void afterAgentInvocation(AgentResponse agentResponse) { }
    default void onAgentInvocationError(AgentInvocationError agentInvocationError) { }

    default void onAgenticScopeCreated(AgenticScope agenticScope) { }
    default void onAgenticScopeDestroyed(AgenticScope agenticScope) { }
}
```

Note that all methods of this interface have a default empty implementation, so that it is possible to implement only the methods of interest. This will also allow to add new methods in future releases without breaking existing implementations.

For instance the following configuration of the `CreativeWriter` agent will log to the console when it is invoked and what is the story it generated.

```java
CreativeWriter creativeWriter = AgenticServices.agentBuilder(CreativeWriter.class)
        .chatModel(baseModel())
        .outputKey("story")
        .listener(new AgenticListener() {
            @Override
            public void beforeAgentInvocation(AgentRequest request) {
                System.out.println("Invoking CreativeWriter with topic: " + request.inputs().get("topic"));
            }
        
            @Override
            public void afterAgentInvocation(AgentResponse response) {
                System.out.println("CreativeWriter generated this story: " + response.output());
            }
        })
        .build();
```

These listener methods receive as argument respectively an `AgentRequest` and an `AgentResponse` that provide useful information about the agent invocation, like its name, the inputs it received and the output it produced, together with the instance of the `AgenticScope` used for that invocation. Note that these methods are invoked in the same thread used to also perform the agent invocation, so they are synchronous with it and should not perform long blocking operations.

Listeners are composable, meaning that you can register multiple listeners to the same agent, by invoking the `listener` method more than once, and they will be notified in the order they were registered. They are also hierarchical, meaning that they are inherited by subagents, so that if you register a listener to a top level agent, that listener will also be notified of the invocations of all its subagents at any level and composed with all listeners that these subagents could have registered on their own.

### Monitoring

Leveraging the observability features provided by the `AgenticListener` interface, the `langchain4j-agentic` module also provides a built-in implementation of this interface, named `AgenticMonitor`, having the goal of recording all agents invocations in an in-memory tree structure, allowing to inspect the sequence of invocations and their results during or after the execution of the agentic system. This monitor can be registered as a listener to the root agent of the agentic system using the `listener` method of the agent builder.

To provide a more comprehensive example, let's reconsider the loop workflow intended to generate and iteratively refine a story until it meets the required style quality, and register a few listeners on it, including an `AgenticMonitor`.

```java
AgenticMonitor monitor = new AgenticMonitor();

CreativeWriter creativeWriter = AgenticServices.agentBuilder(CreativeWriter.class)
        .listener(new AgenticListener() {
            @Override
            public void beforeAgentInvocation(AgentRequest request) {
                System.out.println("Invoking CreativeWriter with topic: " + request.inputs().get("topic"));
            }
        })
        .chatModel(baseModel())
        .outputKey("story")
        .build();

StyleEditor styleEditor = AgenticServices.agentBuilder(StyleEditor.class)
        .chatModel(baseModel())
        .outputKey("story")
        .build();

StyleScorer styleScorer = AgenticServices.agentBuilder(StyleScorer.class)
        .name("styleScorer")
        .chatModel(baseModel())
        .outputKey("score")
        .build();

UntypedAgent styleReviewLoop = AgenticServices.loopBuilder()
        .subAgents(styleScorer, styleEditor)
        .maxIterations(5)
        .exitCondition(agenticScope -> agenticScope.readState("score", 0.0) >= 0.8)
        .build();

UntypedAgent styledWriter = AgenticServices.sequenceBuilder()
        .subAgents(creativeWriter, styleReviewLoop)
        .listener(new AgenticListener() {
            @Override
            public void afterAgentInvocation(AgentResponse response) {
                if (response.agentName().equals("styleScorer")) {
                    System.out.println("Current score: " + response.output());
                }
            }
        })
        .listener(monitor)
        .outputKey("story")
        .build();
```

Here a first listener is registered directly on the `creativeWriter` agent, so that it logs the request topic for the story to be generated only when that agent is invoked. A second listener is registered on the top level `styledWriter` agent, so that it will be also invoked for all subagents in the hierarchy of that agent at any level. That is why the `afterAgentInvocation` method of that listener checks if the agent being invoked is the `styleScorer`, and only in that case it logs the current score assigned to the style of the generated story.

Finally, the `AgenticMonitor` instance is also registered, and automatically composed with the other 2 listeners, as a further listener to the `styledWriter` top level agent, so that it can track all agents invocations in the whole agentic system. 

When invoking the `styledWriter` agent as follows:

```java
Map<String, Object> input = Map.of(
        "topic", "dragons and wizards",
        "style", "comedy");
String story = styledWriter.invoke(input);
```

the `AgenticMonitor` records all agents invocations in a tree structure that also keeps track of the start time, finish time, duration, inputs and output of each agent call. At this point it is possible to retrieve the recorded executions from the monitor and for instance print it to the console for inspection.

```java
MonitoredExecution execution = monitor.successfulExecutions().get(0);
System.out.println(execution);
```

so it will reveal the nested sequence of agents invocations necessary to generate and refine the story, like it follows:

```
AgentCall{agent=Sequential, startTime=2025-12-04T17:23:45.684601233, finishTime=2025-12-04T17:25:31.310476077, duration=105625 ms, inputs={style=comedy, topic=dragons and wiz...}, output=In the shadowy ...}
|=> AgentCall{agent=generateStory, startTime=2025-12-04T17:23:45.687031946, finishTime=2025-12-04T17:23:53.216629832, duration=7529 ms, inputs={topic=dragons and wiz...}, output=In the shadowed...}
|=> AgentCall{agent=reviewLoop, startTime=2025-12-04T17:23:53.218004760, finishTime=2025-12-04T17:25:31.310442197, duration=98092 ms, inputs={score=0.85, topic=dragons and wiz..., style=comedy, story=In the shadowy ...}, output=null}
    |=> AgentCall{agent=scoreStyle, startTime=2025-12-04T17:23:53.218606335, finishTime=2025-12-04T17:23:58.900747685, duration=5682 ms, inputs={style=comedy, story=In the shadowed...}, output=0.25}
    |=> AgentCall{agent=editStory, startTime=2025-12-04T17:23:58.901041911, finishTime=2025-12-04T17:24:58.130857588, duration=59229 ms, inputs={style=comedy, story=In the shadowed...}, output=In the shadowy ...}
    |=> AgentCall{agent=scoreStyle, startTime=2025-12-04T17:24:58.130980855, finishTime=2025-12-04T17:25:31.310076714, duration=33179 ms, inputs={style=comedy, story=In the shadowy ...}, output=0.85}
```

## General checklist


docs/docs/tutorials/agents.md
```diff
@@ -539,18 +539,128 @@ UntypedAgent novelCreator = AgenticServices.sequenceBuilder()
...
+These listener methods receive as argument respectively an `AgentRequest` and an `AgentResponse` that provide useful information about the agent invocation, like its name, the inputs it received and the output it produced, together with the instance of the `AgenticScope` used for that invocation. Note that these methods are invoked in the same thread used to also perform the agent invocation, so they are synchronous with it and should not perform long blocking operations.
+
+Listeners are composable, meaning that you can register multiple listeners to the same agent, by invoking the `listener` method more than once, and they will be notified in the order they were registered. They are also hierarchical, meaning that they are inherited by subagents, so that if you register a listener to a top level agent, that listener will also be notified of the invocations of all its subagents at any level and composed with all listeners that these subagents could have registered on their own.
```
dliubarskyi: nit: sentence about hierarchical is pretty important, perhaps it deserves it's own paragraph or even section?

BTW, is it possible to disable this hierachical-ity? This might be useful for some use cases

docs/docs/tutorials/agents.md
```diff
@@ -539,18 +539,128 @@ UntypedAgent novelCreator = AgenticServices.sequenceBuilder()
...
+        .maxIterations(5)
+        .exitCondition(agenticScope -> agenticScope.readState("score", 0.0) >= 0.8)
+        .build();
+
+UntypedAgent styledWriter = AgenticServices.sequenceBuilder()
+        .subAgents(creativeWriter, styleReviewLoop)
+        .listener(new AgenticListener() {
+            @Override
+            public void afterAgentInvocation(AgentResponse response) {
+                if (response.agentName().equals("styleScorer")) {
```
dliubarskyi, line 631: Did you consider having separate listener methods for child agent invocations? From the API perspective, this will be more explicit and perhaps less confusing. But perhaps having a duplicate of each method is even more confusing

docs/docs/tutorials/agents.md
```diff
@@ -539,18 +539,128 @@ UntypedAgent novelCreator = AgenticServices.sequenceBuilder()
...
+
+```java
+MonitoredExecution execution = monitor.successfulExecutions().get(0);
+System.out.println(execution);
+```
+
+so it will reveal the nested sequence of agents invocations necessary to generate and refine the story, like it follows:
+
+```
+AgentCall{agent=Sequential, startTime=2025-12-04T17:23:45.684601233, finishTime=2025-12-04T17:25:31.310476077, duration=105625 ms, inputs={style=comedy, topic=dragons and wiz...}, output=In the shadowy ...}
```
dliubarskyi: nit: there is inconsistency in the terminology: "agent invocation" vs "agent call"

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/ListenerNotifierUtil.java
```diff
@@ -0,0 +1,39 @@
...
+import dev.langchain4j.agentic.planner.AgentInstance;
+import dev.langchain4j.agentic.scope.AgenticScope;
+import java.util.Map;
+
+public class ListenerNotifierUtil {
+    private ListenerNotifierUtil() { }
+
+    public static void beforeAgentInvocation(AgenticListener listener, AgenticScope agenticScope, AgentInstance agent, Map<String, Object> inputs) {
+        if (listener != null) {
+            listener.beforeAgentInvocation(new AgentRequest(agenticScope, agent, inputs));
```
dliubarskyi: Perhaps it is worth wrapping listener invocation into a `try-catch` block and not fail the whole agentic flow if some listener fails for some reason

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentListenerProvider.java
```diff
@@ -0,0 +1,8 @@
+package dev.langchain4j.agentic.observability;
+
+import dev.langchain4j.Internal;
+
+@Internal
+public interface AgentListenerProvider {
```
dliubarskyi, line 6:
```suggestion
public interface AgenticListenerProvider {
```

or perhaps `AgenticListener` should be renamed into `AgentListener` for consistency

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/ComposedAgentListener.java
```diff
@@ -0,0 +1,61 @@
...
+            } else {
+                this.listeners.add(listener);
+            }
+        }
+    }
+
+    @Override
+    public void beforeAgentInvocation(final AgentRequest agentRequest) {
+        for (AgenticListener listener : listeners) {
+            listener.beforeAgentInvocation(agentRequest);
```
dliubarskyi, line 48: Here as well: perhaps it is worth wrapping listener invocation into a `try-catch` block to not skip the rest of listeners in case of failure

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/AgentListenerSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+ * <p>
+ * Example:
+ * <pre>
+ * {@code
+ *    public interface CreativeWriter {
+ *         @Agent(description = "Generate a story based on the given topic", outputKey = "story")
+ *         String generateStory(@V("topic") String topic);
+ *
+ *         @AgentListenerSupplier
+ *         static AgentListener listener() {
```
dliubarskyi, line 22: `AgentListener` in the javadoc, but it is called `AgenticListener`now, although I would rename it into `AgentListener`

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/AgentListenerSupplier.java
```diff
@@ -0,0 +1,37 @@
...
+ *                 }
+ *             };
+ *         }
+ *     }
+ * }
+ * </pre>
+ */
+@Retention(RUNTIME)
+@Target({METHOD})
+public @interface AgentListenerSupplier {
```
dliubarskyi, line 36: name inconsistency: `AgentListener` vs `AgenticListener`

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
```diff
@@ -178,24 +169,40 @@ private Object executeAgentMethod(AgenticScopeRegistry registry, Method method,
...
+            return (Map<String, Object>) args[0];
+        }
+        Map<String, Object> namedArgs = new HashMap<>();
+        for (int i = 0; i < args.length; i++) {
+            namedArgs.put(ParameterNameResolver.name(method.getParameters()[i]), args[i]);
+        }
+        return namedArgs;
+    }
+
     private class PlannerLoop implements AgentInvocationListener {
```
dliubarskyi, line 208: I am a bit confused between existing `AgentInvocationListener` and new `AgenticListener`, do you plan to keep them both?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentInvocationError.java
```diff
@@ -0,0 +1,16 @@
+package dev.langchain4j.agentic.observability;
+
+import java.util.Map;
+import dev.langchain4j.agentic.planner.AgentInstance;
+import dev.langchain4j.agentic.scope.AgenticScope;
+
+public record AgentInvocationError(AgenticScope agenticScope, AgentInstance agent, Map<String, Object> inputs, Throwable error) {
```
dliubarskyi, line 7: Should it be called `AgentCallError` to be consistent with `AgentCall`? Or perhaps everything should be called "invocation" rather than "call"?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgenticListener.java
```diff
@@ -0,0 +1,16 @@
...
+ * Listener interface for monitoring agent invocations.
+ */
+public interface AgenticListener {
+
+    default void beforeAgentInvocation(AgentRequest agentRequest) { }
+    default void afterAgentInvocation(AgentResponse agentResponse) { }
+    default void onAgentInvocationError(AgentInvocationError agentInvocationError) { }
+
+    default void onAgenticScopeCreated(AgenticScope agenticScope) { }
+    default void onAgenticScopeDestroyed(AgenticScope agenticScope) { }
```
dliubarskyi: naming nit: should they be called `afterAgenticScopeCreated` and `beforeAgenticScopeDestroyed`? `onAgenticScopeDestroyed` accepts `AgenticScope`, so the name might be a bit confusing

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
```diff
@@ -0,0 +1,83 @@
...
+    public void onAgentInvocationError(AgentInvocationError agentInvocationError) {
+        Object memoryId = agentInvocationError.agenticScope().memoryId();
+        MonitoredExecution execution = ongoingExecutions.remove(memoryId);
+        if (execution != null) {
+            execution.onAgentInvocationError(agentInvocationError);
+            failedExecutions.computeIfAbsent(memoryId, k -> new ArrayList<>()).add(execution);
+        }
+    }
+
+    public Map<Object, MonitoredExecution> ongoingExecutions() {
```
dliubarskyi, line 53: There are currently 3 terms used around this feature: "invocation", "call" and "execution". Would it be possible to use only one?

dliubarskyi (commented):

docs/docs/tutorials/agents.md
mariofusco: I did consider this possibility, but at the moment I have 5 methods, I'm already planning to add more when I will have tools calling monitoring and I cannot foresee how many I will have to add in future. In my opinion, having a single listener for each of those methods will bloat the agents public API, and having a single listener interface with all `default` methods is a good compromise for conciseness and forward compatibility.

It would be quick and easy to add separate listener methods, taking a lambda and internally creating and registering and implementation of the `AgenticListener`. This is also why I already made them composable. To recap I will be always in time to add listener methods for specific events (not all of them) without breaking backward compatibility, but I would do this in future and only if anybody asks for them.

docs/docs/tutorials/agents.md
mariofusco: I have `AgentInvocation` somewhere else and I wanted to avoid the class names duplication, but I agree that this generates even more confusion. I will review the namings.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/ListenerNotifierUtil.java
mariofusco: Done.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentListenerProvider.java
mariofusco: Done.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/AgentListenerSupplier.java
mariofusco: I renamed it `AgentListener` everywhere.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/AgentListenerSupplier.java
mariofusco: I renamed it `AgentListener` everywhere.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgenticListener.java
mariofusco: Good point, renamed.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
mariofusco: To me an "execution" is the execution of a whole agentic system, while an "invocation" is an invocation of a single agent. I tried to clarify this in Javadocs. I removed the "call" that I only introduced to avoid a name clash with another `AgentInvocation` that anyway it is an internal class so this shouldn't be a problem. 

I cannot unify that other `AgentInvocation` with this one, because that one is part of the `AgenticScope` so it has to be serializable.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentInvocationError.java
mariofusco: I renamed `AgentCall` to `AgentInvocation`.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
mariofusco: That one is an internal interface only used by the Planner. I moved it into the `internal` package to make this clearer.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/ComposedAgentListener.java
mariofusco: Done.

docs/docs/tutorials/agents.md
mariofusco: I renamed `AgentCall` to `AgentInvocation`.

docs/docs/tutorials/agents.md
mariofusco: I clarified documentation, putting in evidence both composable and hierarchical features. I could implement the possibility to optionally disable listeners' hierachical-ity, but since this could be done in a backward compatible way, I'd prefer to do this in a separated pull request and only if any user ask for it.

mariofusco:
@dliubarskyi Thanks for review, I think I addressed all your remarks and this is ready to be reviewed again.

docs/docs/tutorials/agents.md
dliubarskyi: Sorry, I guess I was not clear. My concern is about the fact that the same listener is called for the agent on which listener is registered and for all of it's "children". For this reason you need to do `if(response.agentName().equals("xyz"))` which is not only bulky but also error-prone. From the API point of view, it is also not explicit that this method will be called not only for the agent that this listener is being registered on.

Random idea: perhaps adding a `boolean isChild` argument to each method would make it more explicit and simultaneously remove the need to write an `if` condition?

```java
public void afterAgentInvocation(AgentResponse response, boolean isChild) {
    if (!isChild) {
        ...
    }
}
```

mariofusco: > Random idea: perhaps adding a `boolean isChild` argument to each method would make it more explicit and simultaneously remove the need to write an `if` condition?

In the case of this example, like probably in the vast majority of situations, the presence of that `isChild` argument won't remove the need of also test the agent name because the listener is registered on the top level agent, but in reality it needs to act only for a specific subagent, so you will still have to check the method name to skip the execution for all other subagents. In reality maybe this specific test case is a bit unnatural because if you're in that situation you could better register the listener on the single leaf agent instead only on the top level one.

To give you the complete picture, I implemented this hierarchical composition of listeners because it was necessary for the monitor. Thinking twice, this could be a useful behavior in some other situation, but probably it shouldn't be the only possibility and neither the the default one. If you agree I will try to make this optional.

dliubarskyi: > this could be a useful behavior in some other situation

Totally agree

> neither the the default one

TBH I have no idea what the default should be


I was only concerned about having this hierarchical behaviour not visible/explicit in the API.
Another random idea, perhaps this would work?
```java
public void afterAgentInvocation(AgentResponse response, Class<?> agentType) {
    if (agentType == MyAgent.class) {
        ...
    }
}
```

mariofusco: > I was only concerned about having this hierarchical behaviour not visible/explicit in the API. 

I made the listeners inheritance optional and off by default, please give a second look.

> Another random idea, perhaps this would work? 
> ```java
> public void afterAgentInvocation(AgentResponse response, Class<?> agentType) {
>     if (agentType == MyAgent.class) {
>         ...
>     }
> }
> ```

The agent type is available through the `AgentInstance` interface and included in the `AgentRequest` and `AgentResponse`, so you can check it doing something like

```java
if (response.agent().type() == MyAgent.class)
``` 

but in this case it has the same limitation of checking the agent name.

mariofusco:
@dliubarskyi I made the listeners inheritance optional and off by default, please give a second look.

dliubarskyi:
@mariofusco looks good, thank you!

Please check this test failure:

```java
Error:  Errors: 
Error:    CustomPlannerIT.parallel_in_pairs_tests:88 » NullPointer Cannot read the array length because "args" is null
```

mariofusco:
> @mariofusco looks good, thank you!
> 
> Please check this test failure:
> 
> ```java
> Error:  Errors: 
> Error:    CustomPlannerIT.parallel_in_pairs_tests:88 » NullPointer Cannot read the array length because "args" is null
> ```

Fixed. Sorry and thanks for checking it.

docs/docs/tutorials/agents.md
```diff
@@ -539,18 +539,134 @@ UntypedAgent novelCreator = AgenticServices.sequenceBuilder()
...
+
+Leveraging the observability features provided by the `AgentListener` interface, the `langchain4j-agentic` module also provides a built-in implementation of this interface, configured to be inherited by all subagents, named `AgentMonitor`, having the goal of recording all agents invocations in an in-memory tree structure, allowing to inspect the sequence of invocations and their results during or after the execution of the agentic system. This monitor can be registered as a listener to the root agent of the agentic system using the `listener` method of the agent builder.
+
+To provide a more comprehensive example, let's reconsider the loop workflow intended to generate and iteratively refine a story until it meets the required style quality, and register a few listeners on it, including an `AgentMonitor`.
+
+```java
+AgentMonitor monitor = new AgentMonitor();
+
+CreativeWriter creativeWriter = AgenticServices.agentBuilder(CreativeWriter.class)
+        .listener(new AgenticListener() {
```
dliubarskyi:
```suggestion
        .listener(new AgentListener() {
```
Please make sure there are no old names left after all the renamings

mariofusco: Good catch, thanks. Fixed.

dliubarskyi (approved):
@mariofusco thank you, great job!

PR #4185, opened 2025-12-08
Introduce McpToolProviderRequest

jmartisk:
See https://github.com/quarkiverse/quarkus-langchain4j/issues/1998 for the explanation


jmartisk:
I gave up on this (see https://github.com/quarkiverse/quarkus-langchain4j/pull/2004#issuecomment-3625690622), maybe I'll re-think it in the future.

PR #4200, opened 2025-12-10
feat: including reasoning content

### LINKED ISSUE #4097: Gemini 3 Pro Tools Calling Fails in LangChain4j Due to Missing thoughtSignature Handling (Cannot Continue Multi-Turn Conversations)
When integrating Google Gemini 3 Pro (latest versions), any time the model triggers a tool call (function calling), the Gemini API now requires that the returned **thoughtSignature** must be passed back exactly as returned in the next request.

According to the official Google API documentation:

Gemini 3 Pro returns a mandatory thoughtSignature for any tool call.

The next request must include the same signature in the same content part position.

If the signature is missing or placed incorrectly, the API returns a 400 error immediately.

❗ Problem in LangChain4j

Currently, LangChain4j does not support this required thoughtSignature mechanism:

It does not extract thoughtSignature from the model response.

It does not store the signature (e.g., inside AiMessage.attributes).

It does not re-inject the signature into the next request’s content.parts.

As a result:

Any tool call causes the next Gemini request to fail.

Multi-turn conversations after a tool call become impossible.

Gemini 3 Pro's mandatory behavior becomes incompatible with LangChain4j.

This makes Tool Calling unusable with Gemini 3 Pro under LangChain4j.

✔️ Feature request

Please add support for Gemini’s mandatory thoughtSignature flow:

Extract signature from model responses

Persist it automatically between turns

Re-insert the signature into the next request in the correct content part position

Or provide hooks/callbacks allowing developers to manually manage signatures

This is currently a blocking issue that prevents real use of Gemini 3 Pro with tool calling in LangChain4j.

commandf1:
## Issue

Closes #4097

## General checklist


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/AssistantMessage.java
```diff
@@ -37,6 +40,7 @@ public final class AssistantMessage implements Message {
     @JsonProperty
     @Deprecated
     private final FunctionCall functionCall;
+    private final Map<String, Object> additionalProperties;
```
dliubarskyi, line 43: This class already has a `reasoningContent` field, can't it be used?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
```diff
@@ -400,6 +404,31 @@ public OpenAiChatModelBuilder returnThinking(Boolean returnThinking) {
...
+        /**
+         * Controls whether to include reasoning_content in assistant messages when sending requests to the API.
+         * This is required for some APIs (like DeepSeek) when using thinking mode with tool calls.
+         * <p>
+         * Disabled by default.
+         * <p>
+         * When enabled, the reasoning content from previous assistant messages (stored in {@link AiMessage#thinking()})
+         * will be included in the request when converting messages to API format.
+         */
+        public OpenAiChatModelBuilder includeReasoningContentInRequests(Boolean includeReasoningContentInRequests) {
```
dliubarskyi: Could you please use `boolean sendThinking` to be consistent with other implementations?

dliubarskyi (commented):
@commandf1 thank you!

It would be great if you could add/update tests in `OpenAiChatModelDeepSeekThinkingIT` and `OpenAiStreamingChatModelDeepSeekThinkingIT`

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/AssistantMessage.java
commandf1: it is used to add more custom additional properties to prevent similar situations from happening

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
coralundersea: ok

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/AssistantMessage.java
coralundersea: reasoningContent can only be serialized to reasoning_content therefore we cannot customize the field name

coralundersea:
the method name has been changed to sendThinking
tests have been added and passed
@dliubarskyi

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
```diff
@@ -400,6 +404,31 @@ public OpenAiChatModelBuilder returnThinking(Boolean returnThinking) {
...
+            return this;
+        }
+
+        /**
+         * Sets the field name to use for reasoning content in API requests.
+         * <p>
+         * Defaults to "reasoning_content".
+         * <p>
+         */
+        public OpenAiChatModelBuilder reasoningContentFieldName(String reasoningContentFieldName) {
```
dliubarskyi: What about having a `sendThinking(Boolean sendThinking, String fieldName)` method instead of `reasoningContentFieldName` ?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatRequestParameters.java
```diff
@@ -23,6 +23,8 @@ public class OpenAiChatRequestParameters extends DefaultChatRequestParameters {
     private final Map<String, String> metadata;
     private final String serviceTier;
     private final String reasoningEffort;
+    private final Boolean sendThinking;
+    private final String reasoningContentFieldName;
```
dliubarskyi: I doubt that we need to customize reasoning content field name per each request. Please remove this field from this class.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatRequestParameters.java
```diff
@@ -23,6 +23,8 @@ public class OpenAiChatRequestParameters extends DefaultChatRequestParameters {
     private final Map<String, String> metadata;
     private final String serviceTier;
     private final String reasoningEffort;
+    private final Boolean sendThinking;
```
dliubarskyi: I am also not sure this field should be here, I think it is enough to enable it on the model level

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/AssistantMessage.java
dliubarskyi: Why do we need to customize the field name?

coralundersea: for some other llm providers need user to provide reasoning content with different field name

coralundersea:
@dliubarskyi how about only use one method named sendThinkingWithFieldName(String fieldName)
sendThinking should be disabled while fieldName is null

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/AssistantMessage.java
dliubarskyi: @commandf1 could you please give an example?

dliubarskyi:
> how about only use one method named sendThinkingWithFieldName(String fieldName)

I guess there should be a way to enable sending thinking (`sendThinking(Boolean)`) without the need to specify the field name, there should be a convenient default ("reasoning_content").

coralundersea:
@dliubarskyi all issues have been solved

dliubarskyi (approved):
@commandf1 thank you!

PR #4201, opened 2025-12-10
Fix MCP client permanent disconnection after temporary network failure

dongfeng3692:
## The Problem
I encountered a critical availability issue: **If the MCP server becomes briefly unavailable (e.g., due to network glitches or a server restart), the client fails to reconnect even after the service recovers.**

## Description
I identified two critical stability issues in the MCP client implementation:

1.  **AutoHealth Task Termination:** In `triggerReconnection()`, if `initialize()` throws an exception, it propagates up to the `AutoHealth` scheduled task. This unhandled exception causes the scheduler to permanently stop the health check thread, preventing any future reconnection attempts.
2.  **Initialization Logic Flaw:** In the `execute()` method, when sending an `McpInitializeRequest` (usually during a retry), the code was waiting on `reinitializeInProgress.join()`. If the previous initialization attempt failed, this future holds an exception. Calling `join()` retrieves this **historical error**, causing the current (new) initialization request to fail immediately instead of proceeding to establish a new connection.

## Changes
*   **Safe Reconnection:** Added a `try-catch` block inside `triggerReconnection()`. Now, if initialization fails, it logs a warning and ensures the `AutoHealth` thread remains alive to trigger future retries.
*   **Bypass History Check for Init:** Modified `execute()` to skip the `initializeInProgress.join()` check when the message is an `McpInitializeRequest`. This allows the initialization packet to be sent regardless of the previous state, breaking the loop of fetching historical errors.

## General checklist
I have manually run all the unit and integration tests in the [core]

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/ConditionalPlanner.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/ConditionalPlanner.java
@@ -19,7 +19,7 @@ public record ConditionalPlanner(List<ConditionalAgent> conditionalAgents) imple
                 .filter(conditionalAgent -> conditionalAgent.condition.test(planningContext.agenticScope()))
                 .flatMap(conditionalAgent -> conditionalAgent.agentInstances.stream())
                 .toList();
-        return call(agentsToCall);
+        return agentsToCall.isEmpty() ? done() : call(agentsToCall);
     }
 
     @Override
--- a/langchain4j-agentic/src/test/java/dev/langchain4j/agentic/WorkflowAgentsIT.java
+++ b/langchain4j-agentic/src/test/java/dev/langchain4j/agentic/WorkflowAgentsIT.java
@@ -629,6 +629,41 @@ public class WorkflowAgentsIT {
     }
 
     @Test
+    void no_matching_condition_tests() {
+        CategoryRouter routerAgent = AgenticServices.agentBuilder(CategoryRouter.class)
+                .chatModel(baseModel())
+                .outputKey("category")
+                .build();
+
+        LegalExpert legalExpert = spy(AgenticServices.agentBuilder(LegalExpert.class)
+                .chatModel(baseModel())
+                .outputKey("response")
+                .build());
+        TechnicalExpert technicalExpert = spy(AgenticServices.agentBuilder(TechnicalExpert.class)
+                .chatModel(baseModel())
+                .outputKey("response")
+                .build());
+
+        UntypedAgent expertsAgent = AgenticServices.conditionalBuilder()
+                .subAgents(
+                        agenticScope ->
+                                agenticScope.readState("category", RequestCategory.UNKNOWN) == RequestCategory.LEGAL,
+                        legalExpert)
+                .subAgents(
+                        agenticScope -> agenticScope.readState("category", RequestCategory.UNKNOWN)
+                                == RequestCategory.TECHNICAL,
+                        technicalExpert)
+                .build();
+
+        var agentInstance = AgenticServices.sequenceBuilder(ExpertRouterAgentInstance.class)
+                .subAgents(routerAgent, expertsAgent)
+                .outputKey("response")
+                .build();
+
+        assertThat(agentInstance.ask("I broke my leg what should I do")).isNull();
+    }
+
+    @Test
     void agents_validation_tests() {
         CategoryRouter routerAgent = AgenticServices.agentBuilder(CategoryRouter.class)
                 .chatModel(baseModel())


dongfeng3692:
I'm sorry, I don't know how to write the code for this test case, but I verified this fix locally by manually stopping the MCP server, waiting for the client to log connection errors, and then restarting the server. 
If you have any suggestions, I'd be happy to add the test case code.

jmartisk:
Nice, thanks!
I think the problem only occurs if the auto-health-check is attempted while the server is down. That is why `McpReconnectStreamableHttpTransportIT` (which should test reconnecting after the connection breaks) did not catch this issue - the server is down for only 5 seconds and the auto-health-check interval is set to 30 seconds for that test.

So, could you try this with `McpReconnectStreamableHttpTransportIT` (you will need JBang to be installed) - I think that adding a `.autoHealthCheckInterval(Duration.ofSeconds(1))` call to the `DefaultMcpClient` builder in the `setup` method is enough to actually make the test reproduce the issue. Could you verify that after adding  `.autoHealthCheckInterval(Duration.ofSeconds(1))` the test starts failing when your fix is not in place, but it will pass when your patch is applied?

dongfeng3692:
Hello,
﻿
I've modified the test code to simulate an MCP server restart and conducted a comparative test. The updated code successfully passed the tests after the fix was applied.
﻿
Here are the test results:
﻿
*   **Before the fix:** 
fb902ce866b2fc35d46e8724a127522a

*   **After the fix:** 
6ecf2f5b65c7a4157f62616706e81af5

dongfeng3692:
The original test method had a minor issue. I've re-implemented a new process termination method to forcefully terminate JBang processes and their sub-processes after the test, addressing the problem of JBang processes not shutting down properly.

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpHeadersStreamableHttpTransportIT.java
```diff
@@ -25,9 +28,17 @@ static void setup() throws IOException, InterruptedException, TimeoutException {
                 .build();
         mcpClient = new DefaultMcpClient.Builder()
                 .transport(transport)
+                .autoHealthCheckInterval(Duration.ofSeconds(1))
                 .toolExecutionTimeout(Duration.ofSeconds(4))
                 .reconnectInterval(Duration.ofSeconds(1))
                 .build();
+        // waiting for mcpClient to be ready
+        Thread.sleep(5000);
```
jmartisk: I don't understand the point of adding these lines - this is a test for HTTP headers, it doesn't have anything to do with reconnection after failure.  Only `McpReconnectStreamableHttpTransportIT` requires killing the process in the middle of the test and that is already implemented in it.

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpHeadersStreamableHttpTransportIT.java
```diff
@@ -36,7 +47,16 @@ static void tearDown() throws Exception {
             mcpClient.close();
         }
         if (process != null) {
-            process.destroyForcibly();
+            destroyProcessTree(process);
```
jmartisk: I don't think this is necessary - could you elaborate on the issue where it didn't kill jbang properly? The server is just a single process, no need to kill the whole tree. Furthermore, if we needed to apply this change, we would need to do the same in all tests, not just this one specific test (but I don't think it's necessary, the test suite has been around for a long time and nobody complained about this issue so far).

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpHeadersStreamableHttpTransportIT.java
jmartisk: Let's stick to `McpReconnectStreamableHttpTransportIT` and make sure that one works please :) this one is for http headers, let's not mix two test cases together. So I would rather move the `.autoHealthCheckInterval(Duration.ofSeconds(1))` call to `McpReconnectStreamableHttpTransportIT` and keep this one unchanged.

dongfeng3692: My apologies, I didn't read your previous reply carefully enough and overlooked McpReconnectStreamableHttpTransportIT. I am now re-testing McpReconnectStreamableHttpTransportIT 🙏

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpHeadersStreamableHttpTransportIT.java
dongfeng3692: During my testing, I noticed that after the tests complete, it doesn't shut down completely, as shown in the image. I was wondering if you've encountered a similar situation?
image

jmartisk: No I'm not aware of that, the MCP server is a single process, so `destroyForcibly()` should reliably kill it.. Do you see hanging processes after the test finishes?

jmartisk: Maybe it was just because of the stuff that you added to the `McpHeadersStreamableHttpTransportIT` but if you remove it, you won't see the problem anymore?!

dongfeng3692: I've reverted to the main branch, which doesn't contain my code.
May I ask if you are using an Apple device? I am on Windows.

dongfeng3692: image
I've investigated this thoroughly, and this is the behavior I'm seeing on Windows: the parent process jbang.cmd starts the java.exe process. It seems that simply closing the cmd doesn't effectively terminate it.

jmartisk: Ah ok, good catch then! I didn't try on Windows, I use Linux and apparently the behavior is different.
In that case, could you please apply this change across all tests in the `dev.langchain4j.mcp.client.integration` package? Probably extract the method for destroying the process tree into a separate class that can be reused because otherwise it would be duplicated many times.

dongfeng3692: No problem, I'll update it in the next few days.

dongfeng3692: I've added a Windows destroyProcess method to McpServerHelper and replaced all process.destroy calls with this method.

jmartisk:
Ah dang, I tried cleaning up and force pushing, but I pushed a wrong branch (please don't use your `main` for sending PRs!!!!! :cry:  ), and the PR auto-closed itself. Sorry for the confusion. I'll re-create it

jmartisk:
Replaced by https://github.com/langchain4j/langchain4j/pull/4248

PR #4211, opened 2025-12-11
Anthropic: support latest tool features

dliubarskyi:
## Change
Added support for the latest Anthropic's tool features, especially the ones described here: https://www.anthropic.com/engineering/advanced-tool-use

## General checklist


docs/docs/integrations/language-models/anthropic.md
```diff
@@ -112,6 +114,264 @@ by setting `toolChoice(ToolChoice)` or `toolChoiceName(String)`.
...
+        return "12:34:56";
+    }
+}
+
+ChatModel chatModel = AnthropicChatModel.builder()
+        .apiKey(System.getenv("ANTHROPIC_API_KEY"))
+        .modelName(CLAUDE_SONNET_4_5_20250929)
+        .beta("advanced-tool-use-2025-11-20")
+        .serverTools(toolSearchTool)
+        .toolMetadataKeysToSend("deferLoading") // need to specify it explicitly
```
mariofusco: I don't know the spec so maybe this is correct, but the key used here "deferLoading" is different from the one declared at line 157 "defer_loading", is this right?

dliubarskyi: Good spot, thank you!

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
```diff
@@ -323,6 +340,37 @@ public static AnthropicTool toAnthropicTool(
...
 
+        if (!toolMetadataKeysToSend.isEmpty()) {
+            toolBuilder.customParameters(retainKeys(toolSpecification.metadata(), toolMetadataKeysToSend));
+        }
+
         return toolBuilder.build();
     }
+
+    private static Map<String, Object> retainKeys(Map<String, Object> map, Set<String> keys) {
+        Map<String, Object> result = new HashMap<>(map);
```
mariofusco: Is this correct? Shouldn't the `result` map start empty and then be populated only in the subsequent for loop?

mariofusco (dismissed):
Great job, please just check the last comment I left on that suspicious `retainKeys` methods and then this is ok for me to be merged.

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
dliubarskyi: Oh my, what a dumb mistake, thank you!

dliubarskyi:
@mariofusco thanks a lot for the review!

ejstuart:
@dliubarskyi 
Sorry for coming to this late (and after it's already been merged) but it doesn't look like we're mapping the tool execution result for these server tools into anything (I might be missing something). Looks like we are only mapping the text from the LLM after the tool execution happens.

Some of these [responses](https://platform.claude.com/docs/en/agents-and-tools/tool-use/code-execution-tool#bash-command-response) can contain important information for example if code-execution is used to create a file the file_id to access that file via the claude files api is returned in that e.g.

```json
  {
    "type": "bash_code_execution_tool_result",
    "tool_use_id": "...",
    "content": {
      "type": "bash_code_execution_result",
      "stdout": "...",
      "content": [{ "file_id": "file_xxx" }]  // <-- files created
    }
  }
```

I think would need to update `AnthropicContent.java` and then map these server tool contents to `ToolExecutionResult`

I'd interested in contributing so would be happy to do a follow up PR if you agree with what i've written above.

More context from claude docs: https://platform.claude.com/docs/en/build-with-claude/skills-guide#downloading-generated-files

Wal8800:
Thanks for the pull request.

How we configured the server side tool and the related attributes makes sense.

+1 to @ejstuart comment

dliubarskyi:
@ejstuart I was not sure if anyone is interested in server tool responses, so I did not include it in this PR. Feel free to contribute this feature though. I suggest to return server tool responses inside the `AiMessage.attributes()`. This behaviour should be explicitly enabled when creating `AnthropicChatModel`, otherwise we will pollute `ChatMemory` with unnecessary data.

dliubarskyi:
BTW, with https://github.com/langchain4j/langchain4j/pull/4225 you will be able to access raw HTTP response body and access all Anthropic-specific fields

mullerhai:
@ejstuart @dliubarskyi   Hi ，do  we support load skills  md files for langchain4j agent in new version ?

dliubarskyi:
@mullerhai no, you need to implement it yourself

PR #4212, opened 2025-12-11
Make Planners to declare their topology

mariofusco:
## Issue

Closes #

## Change
The intent of this pull request is to allow `Planner` implementations to declare their topology among the ones enumerated in `AgenticSystemTopology`. This should be informative for all users of that Planner, but also be used by some presentation UI (together with the list of subagents) to create a nicer and better understandable diagram of the whole agentic system.

This pull request also makes each agent aware of its parent in the hierarchy and uses the topology of the agentic system to generate stable agentIDs.

## General checklist


dliubarskyi (dismissed):
@mariofusco thank you!

Should we merge it? (it is in draft state)

mariofusco:
> Should we merge it? (it is in draft state)

Not yet, I flagged this as a draft because I wanted to check the comments in latest Gregory's email first.

mariofusco:
@dliubarskyi I also made each agent aware of its parent in the hierarchy and used the topology of the agentic system to generate stable agentIDs. This pull request should be ready to be reviewed now.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/InternalAgent.java
```diff
@@ -0,0 +1,10 @@
+package dev.langchain4j.agentic.internal;
+
+import dev.langchain4j.agentic.planner.AgentInstance;
+
+public interface InternalAgent extends AgentInstance {
+
+    void setParent(AgentInstance parent);
```
dliubarskyi, line 7: Maybe a stupid question, but shouldn't parent exist before children are created?

mariofusco: Unfortunately no. With the current API what you do is creating the single agents and then composing them as subagents of a parent planner-based orchestrator. Only at that point, when the existing agent is added as subagent to its parent, you know who is the parent, but at that point the child agent has been already created. Am I making sense?

dliubarskyi: Right, my bad

dliubarskyi (approved):
@mariofusco thank you, well done!

PR #4221, opened 2025-12-14
[POC] Support sealed polymorphic parsing and schemas

### LINKED ISSUE #2557: [FEATURE]  LangChain4j's AiService to Support Polymorphic JSON Deserialization

```java
}

    public String type() {
        return type;
    }
}

    public String type() {
        return type;
    }
}

    public String type() {
        return type;
    }
}

}

```

```java
interface Assistant {
}
```

```java
        switch (response) {
                // directly use textMessage
            }
            }
                // directly use cards
            }
        }

```

```java
    @Override

        return switch (type) {
        };
    }
}
```

```java
                .build();

```

raphidlee:
## Issue
Closes #2557

## Change
~~- Add `SealedOutputParser` for sealed types annotated with `@Polymorphic` (discriminator) and 
  `@PolymorphicValue` (permitted subtype mapping); handles duplicate/missing annotations, missing/
  unknown discriminator values, and provides discriminator-aware format instructions.~~
~~- Wire sealed support into `DefaultOutputParserFactory` and `JsonSchemas`: sealed return
  types (including fields and collections) now emit anyOf schemas reflecting the `@Polymorphic`/
  `@PolymorphicValue` metadata; collection parsers reuse item parsers/schemas so `List<sealed>` works
  end to end.~~
~~- Tests: `SealedOutputParserTest` for parsing/error/schema/format cases; `JsonSchemasTest` covers
  sealed, sealed collections, and sealed fields; `PolymorphicAiServiceTest` adds single/list parsing;
  `AiServicesIT` polymorphic case is stabilized with a stubbed chat model and prompt/response
  assertions.~~
- Add PolymorphicOutputParser to handle Jackson @JsonTypeInfo/@JsonSubTypes/@JsonTypeName
    discriminator-based parsing.
- Wire DefaultOutputParserFactory to return the polymorphic parser when applicable.
- Update PojoCollectionOutputParser to use the factory and reuse the item parser’s schema (enabling
  anyOf for polymorphic items).
- Extend schema coverage in JsonSchemasTest and add new PolymorphicOutputParserTest.
- Add PolymorphicAiServiceTest cases for polymorphic response parsing and prompt instructions.

## General checklist


raphidlee:
It looks like the CI is taking a very long time because of an integration test
related to Oracle embedding stores.

From the logs, the delay seems to come from starting an Oracle Testcontainers
image, and the failure happens during test initialization.
This does not seem related to the changes in this PR, since the modified code
is in a different area.

Other PRs appear to run the same CI much faster, so I am not sure why this test
is triggered or behaves differently in this case.

I wanted to leave this note to explain why the CI is slow here.
Please let me know if you prefer me to disable or guard this test for now.

langchain4j/src/main/java/dev/langchain4j/service/output/DefaultOutputParserFactory.java
```diff
@@ -57,6 +56,10 @@ public OutputParser<?> get(Class<?> rawClass, Class<?> typeArgumentClass) {
             return new EnumOutputParser<>(rawClass.asSubclass(Enum.class));
         }
 
+        if (rawClass.isSealed()) {
```
dliubarskyi: Is this 100% non-breaking change?

langchain4j/src/test/java/dev/langchain4j/service/PolymorphicAiServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    interface ChatbotService {
+        ChatbotResponse reply(String userMessage);
+    }
+
+    interface ChatbotListService {
+        List<ChatbotResponse> reply(String userMessage);
+    }
+
+    class StubJsonModel implements ChatModel {
```
dliubarskyi: nit: you can use `ChatModelMock` instead

langchain4j/src/test/java/dev/langchain4j/service/PolymorphicAiServiceTest.java
```diff
@@ -0,0 +1,136 @@
...
+            this.json = json;
+        }
+
+        @Override
+        public ChatResponse doChat(ChatRequest chatRequest) {
+            return ChatResponse.builder().aiMessage(aiMessage(json)).build();
+        }
+    }
+
+    class CapturingModel implements ChatModel {
```
dliubarskyi: nit: you can use `ChatModelMock` instead

langchain4j/src/main/java/dev/langchain4j/service/output/SealedOutputParser.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        Map<String, Class<? extends T>> mapping = new LinkedHashMap<>();
+        for (Class<?> permittedSubclass : permittedSubclasses) {
+            @SuppressWarnings("unchecked")
+            Class<? extends T> subType = (Class<? extends T>) permittedSubclass;
+            PolymorphicValue polymorphicValue = subType.getAnnotation(PolymorphicValue.class);
+            if (polymorphicValue == null) {
+                throw illegalConfiguration(
+                        "Permitted subclass %s of %s must be annotated with @PolymorphicValue",
```
dliubarskyi: Would it be possible to use the name of the class as a value by default and allow to avoid annotations completely? Since classes normally have clear names for humans, they should be clear for LLMs as well.

langchain4j-core/src/main/java/dev/langchain4j/json/Polymorphic.java
```diff
@@ -0,0 +1,17 @@
...
+
+import static java.lang.annotation.ElementType.TYPE;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(TYPE)
+public @interface Polymorphic {
```
dliubarskyi: Maybe a dumb queston, but can't we reuse Jackson's annotations instead of creating our own?

dliubarskyi:
> Other PRs appear to run the same CI much faster, so I am not sure why this test
> is triggered or behaves differently in this case.

Most probably because you did changes in the `langchain4j-core` module, so [GIB](https://github.com/gitflow-incremental-builder/gitflow-incremental-builder) decided to build all modules that depend on it (including Oracle).

I did some changes last week to the CI, Oracle ITs should be skipped now. Sorry for confusion

dliubarskyi (commented):
@raphidlee thank you, this PR looks very promising!

raphidlee:
> @raphidlee thank you, this PR looks very promising!


Thanks for reviewing! I’ll revise it based on your comments and submit a new PR shortly.

dliubarskyi:
@raphidlee great! No need to open a new PR though, you can do the changes in this one (to preserve the history, comments, etc).

langchain4j/src/main/java/dev/langchain4j/service/output/SealedOutputParser.java
raphidlee: By switching to @JsonSubTypes, we now default to using Class#getSimpleName() for the discriminator
  value, and we also allow specifying a custom discriminator.

langchain4j-core/src/main/java/dev/langchain4j/json/Polymorphic.java
raphidlee: Not a dumb question at all — yes, we switched to Jackson’s @JsonTypeInfo/@JsonSubTypes/@JsonTypeName
  so users don’t need custom annotations.
 There are quite a few annotation attributes to use(like JsonTypeInfo's use=JsonTypeInfo.Id.NAME), so if we go with this, we’ll need to document it.

raphidlee:
Update: replaced the custom Polymorphic annotations with Jackson’s @JsonTypeInfo/@JsonSubTypes/
  @JsonTypeName, consolidated polymorphic tests into PolymorphicAiServiceTest, and added coverage for
  default discriminator behavior (default property + simpleName when name is omitted).

langchain4j/src/main/java/dev/langchain4j/service/output/DefaultOutputParserFactory.java
raphidlee: It was not. 
Please check new version with JsonTypeInfo.

raphidlee:
I tried to keep the review on this branch as suggested earlier, but an accidental
merge from `main` caused the diff to become much larger than intended. 
It's My bad.

To keep the review manageable, I’ve opened a new PR from a clean branch with only
the relevant changes:
#4354

dliubarskyi:
Closing in favour of https://github.com/langchain4j/langchain4j/pull/4354

PR #4243, opened 2025-12-17
Deprecate the `langchain4j-github-models` module

jdubois:
I want to deprecate the `langchain4j-github-models` module (which I created), in favor of the `langchain4j-openai-official` module (which I also created).

Here is why:
- The `langchain4j-github-models` uses the Azure Inference API which isn't as well maintained as the OpenAi Java SDK, and is limited to Azure (the OpenAI Java SDK works everywhere, including on competing platforms, which is better for our end-users)
- If you are using Microsoft Foundry (the new name for Azure OpenAi) or OpenAi (which is operated by Microsoft), underneath it's the exact same thing as GitHub Models: we might as well have the same SDK to connect to all of those.
- I would also like to focus on less modules, this would allow better quality from my side.


dliubarskyi (approved):
@jdubois thank you!

PR #4249, opened 2025-12-18
Fix MCP client permanent disconnection after temporary network failure

dongfeng3692:
replacement for https://github.com/langchain4j/langchain4j/pull/4201 @jmartisk


jmartisk (approved):
thanks!

PR #4288, opened 2025-12-25
feat: PgVectorEmbeddingStore supports hybrid search

### LINKED ISSUE #1599: [FEATURE] PgVectorEmbeddingStore supports mixed search
PgVectorEmbeddingStore supports mixed search of semantics and keywords

YongGoose:
## Issue

Closes #1599 

## General checklist


langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStoreSearchModeIT.java
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    @Test
+    void fullTextOnly_withEmptyQuery_shouldReturnEmpty() {
+        TextSegment query = TextSegment.from("vector similarity search in databases");
+        Embedding queryEmbedding = embeddingModel.embed(query).content();
+
+        EmbeddingSearchRequest request = EmbeddingSearchRequest.builder()
+                .query("   ")
+                .queryEmbedding(queryEmbedding)
```
YongGoose: @dliubarskyi 

I have one point I'd like to discuss.
For `FullTextSearch`, I believe query embeddings are not necessary since it’s purely text-based search.
However, looking at the `EmbeddingSearchRequest`, it seems that `queryEmbedding` is required.
This means that even when performing full-text search, we have to provide a dummy query embedding, which I think is something that needs improvement.

I considered removing the null check in the `EmbeddingSearchRequest`, but I thought that wouldn’t be appropriate, since it would allow nulls to pass through in places where they are actually required.

If you have any ideas for a better approach, I’d love to hear them!

YongGoose (commented):
Also, I’d like to ask you to review the query.
Although I verified it through the test results, I think it would be good to double-check it.

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -75,23 +122,32 @@ public class PgVectorEmbeddingStore implements EmbeddingStore<TextSegment> {
...
-        MetadataStorageConfig config = getOrDefault(metadataStorageConfig, DefaultMetadataStorageConfig.defaultConfig());
+        MetadataStorageConfig config =
+                getOrDefault(metadataStorageConfig, DefaultMetadataStorageConfig.defaultConfig());
         this.metadataHandler = MetadataHandlerFactory.get(config);
         useIndex = getOrDefault(useIndex, false);
         createTable = getOrDefault(createTable, true);
         dropTableFirst = getOrDefault(dropTableFirst, false);
 
+        this.searchMode = getOrDefault(searchMode, SearchMode.EMBEDDING_ONLY);
+        this.textSearchConfig = validateTextSearchConfig(getOrDefault(textSearchConfig, DEFAULT_TEXT_SEARCH_CONFIG));
```
YongGoose: I added validation due to the potential risk of SQL injection.

Currently, I am validating the input by defining commonly used values as constants. 
Please let me know if there are any additional values that should be supported.

langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStoreSearchModeIT.java
dliubarskyi: @YongGoose I am not sure full-text search is fitting the `EmbeddingStore` API. By definition, `EmbeddingStore` stores embeddings, so in order to search, you need an embedding. I would focus only on hybrid search in this PR.

YongGoose: Agree.
Done in https://github.com/langchain4j/langchain4j/pull/4288/commits/56d24c3a183cc01e3e240b8e9a5f5363e045f89c !

Please take another look :)

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -50,7 +50,42 @@
  */
 // Needed for inherited bean injection validation
 public class PgVectorEmbeddingStore implements EmbeddingStore<TextSegment> {
+
+    /**
+     * Search modes for the embedding store.
+     */
+    public enum SearchMode {
+        EMBEDDING_ONLY,
```
```suggestion
        VECTOR,
```

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -50,7 +50,42 @@
...
+     */
+    private static final int DEFAULT_RRF_K = 60;
+
+    /**
+     * Whitelist of PostgreSQL text search configurations that are considered safe to use.
+     * <p>
+     * Restricting the text search configuration to this list helps prevent SQL injection
+     * attacks when the configuration name is incorporated into SQL statements.
+     */
+    private static final List<String> ALLOWED_TEXT_SEARCH_CONFIGS =
```
dliubarskyi: Do we really need to validate this on our side? What if one more language is added in the next version of pgvector? Users will not be able to use it before we update this list and release new version...

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -75,23 +121,32 @@ public class PgVectorEmbeddingStore implements EmbeddingStore<TextSegment> {
...
+            DataSource datasource,
+            String table,
+            Integer dimension,
+            Boolean useIndex,
+            Integer indexListSize,
+            Boolean createTable,
+            Boolean dropTableFirst,
+            MetadataStorageConfig metadataStorageConfig,
+            SearchMode searchMode,
+            String textSearchConfig) {
```
dliubarskyi: This is a breaking change. It is a good time to introduce a new ctor that will accept single argument (`PgVectorEmbeddingStoreBuilder`). and change `PgVectorEmbeddingStoreBuilder.build()` to use it.
All existing ctors should stay unchanged.

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
dliubarskyi: I am not sure we should validate this. This value is user-controlled (not end-user-controlled), so the risk of injection is quite low IIUC. But it will limit users...

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -125,28 +182,41 @@ protected PgVectorEmbeddingStore(
             Integer indexListSize,
             Boolean createTable,
             Boolean dropTableFirst,
-            MetadataStorageConfig metadataStorageConfig
-    ) {
-        this(createDataSource(host, port, user, password, database),
-                table, dimension, useIndex, indexListSize, createTable, dropTableFirst, metadataStorageConfig);
+            MetadataStorageConfig metadataStorageConfig,
+            SearchMode searchMode,
```
dliubarskyi: please revert changes to this ctor signature

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -125,28 +182,41 @@ protected PgVectorEmbeddingStore(
...
+                metadataStorageConfig,
+                searchMode,
+                textSearchConfig);
     }
 
     public PgVectorEmbeddingStore() {
         this.datasource = null;
         this.table = null;
         this.metadataHandler = null;
+        this.searchMode = null;
```
```suggestion
        this.searchMode = VECTOR;
```

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -312,11 +401,21 @@ public EmbeddingSearchResult<TextSegment> search(EmbeddingSearchRequest request)
             String whereClause = (filter == null) ? "" : metadataHandler.whereClause(filter);
             whereClause = (whereClause.isEmpty()) ? "" : "AND " + whereClause;
             String query = String.format(
-                    "SELECT (2 - (embedding <=> '%s')) / 2 AS score, embedding_id, embedding, text, %s FROM %s " +
-                            "WHERE round(cast(float8 (embedding <=> '%s') as numeric), 8) <= round(2 - 2 * %s, 8) %s " + "ORDER BY embedding <=> '%s' LIMIT %s;",
-                    referenceVector, join(",", metadataHandler.columnsNames()), table, referenceVector,
-                    minScore, whereClause, referenceVector, maxResults
-            );
+                    "SELECT (2 - (embedding <=> '%1$s')) / 2 AS score, " + "       embedding_id, "
```
dliubarskyi: could you please use a text block here instead of concatenating?

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -342,32 +441,137 @@ referenceVector, join(",", metadataHandler.columnsNames()), table, referenceVect
...
+                coalescedMetadataCols = ", "
+                        + metadataCols.stream()
+                                .map(col -> String.format("COALESCE(v.%1$s, k.%1$s) AS %1$s", col))
+                                .collect(java.util.stream.Collectors.joining(", "));
+            }
+
+            int rrfK = DEFAULT_RRF_K;
+
+            String sql = String.format(
+                    "WITH vector_search AS ( " + "  SELECT "
```
dliubarskyi: same here

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -342,32 +441,137 @@ referenceVector, join(",", metadataHandler.columnsNames()), table, referenceVect
         return new EmbeddingSearchResult<>(result);
     }
 
+    private EmbeddingSearchResult<TextSegment> hybridSearch(EmbeddingSearchRequest request) {
+        Embedding referenceEmbedding = request.queryEmbedding();
+        String keywordQuery = request.query();
+
+        if (isNullOrBlank(keywordQuery)) {
+            return new EmbeddingSearchResult<>(List.of());
```
dliubarskyi: Maybe we should throw an exception instead? If user wants to do a hybrid search, but query is missing, something is wrong on their side, so we need to fail

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -50,7 +50,42 @@
...
+     * RRF assigns each result a score based on its rank in each list using the formula:
+     * {@code score = 1 / (k + rank)}. Larger {@code k} values dampen the impact of rank
+     * differences so that lower-ranked but consistently relevant results can still contribute
+     * meaningfully to the final score.
+     * <p>
+     * The value {@code 60} is a commonly used heuristic in RRF that provides a good balance
+     * between emphasizing top-ranked results and allowing deeper results to influence the
+     * fused ranking, and has been chosen here as a sensible default.
+     */
+    private static final int DEFAULT_RRF_K = 60;
```
dliubarskyi, line 69: There is currently no way to adjust this value, perhaps we could add a config method to the builder?

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
dliubarskyi: Or maybe keep the validation, but make it possible for the user to override the list of allowed values. When exception is thrown, mention how to override the list in the error message. WDYT?

langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStoreSearchModeIT.java
```diff
@@ -0,0 +1,130 @@
...
+                .host(pgVector.getHost())
+                .port(pgVector.getFirstMappedPort())
+                .user("test")
+                .password("test")
+                .database("test")
+                .table(tableBase + "_emb")
+                .dimension(384)
+                .dropTableFirst(true)
+                .textSearchConfig("english")
+                .searchMode(EMBEDDING_ONLY)
```
dliubarskyi: `EMBEDDING_ONLY` is the default, and it is already well-tested. I would suggest extending this test class from `EmbeddingStoreWithFilteringIT` (similarly to `PgVectorEmbeddingStoreIT`) and test only `HYBRID`

langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStoreSearchModeIT.java
```diff
@@ -0,0 +1,130 @@
...
+import dev.langchain4j.store.embedding.EmbeddingStore;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@Testcontainers
+class PgVectorEmbeddingStoreSearchModeIT {
```
dliubarskyi:
```suggestion
class PgVectorEmbeddingStoreHybridIT {
```

dliubarskyi (commented):
@YongGoose thank you!

YongGoose:
> @YongGoose thank you!

You're welcome :)
Happy new year!

Also, I sent you a message on LinkedIn to help with the code review. 
I’d appreciate it if you could take a look when you have time.

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
YongGoose: After some consideration, I agreed with your point that the user restrictions outweigh the benefits of protecting against injection attacks. As a result, I will move forward with removing the validation.

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
YongGoose: Done in https://github.com/langchain4j/langchain4j/pull/4288/commits/b51b5cedb2495229e62deb9e32d5c091f97c6843 !

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
YongGoose: Agree

langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStoreSearchModeIT.java
YongGoose: @dliubarskyi 

I have one question.

When we extend EmbeddingStoreWithFilteringIT and run the tests using the HYBRID mode, the tests do execute in HYBRID mode, but the results differ from what the tests expect.

With the existing embedding tests, the result score is based on cosine similarity, so when the embeddings are similar, the score is close to 1.0.
However, in HYBRID mode, the scoring mechanism changes to a rank-based approach.

Because of this, it seems that we should no longer validate results using score-based assertions, but instead verify the ranking or order of the results. In that case, inheriting from EmbeddingStoreWithFilteringIT may no longer be appropriate, and it might be better to create a separate test class specifically for HYBRID mode.

One idea I have is to introduce a Hybrid-specific base test class, similar to EmbeddingStoreWithFilteringIT, and then have the other HYBRID test classes extend that base class.

However, this approach would require a significant number of changes to the existing tests. I wanted to ask what your thoughts are on this approach and whether you think this is the right direction.

WDYT?

dliubarskyi: @YongGoose I see. How many tests fail in `EmbeddingStoreWithFilteringIT` with hybrid mode? I am wondering if it would be easier to adjust `EmbeddingStoreWithFilteringIT` and allow using other assirtions in the subclasses. That would reduce the duplication and will keep all tests in sync.

YongGoose: > How many tests fail in EmbeddingStoreWithFilteringIT with hybrid mode?

Unfortunately every single tests....

dliubarskyi: Hmm, but why exactly?

YongGoose: The difference in calculation methods appears to be the most significant factor.
In the VECTOR approach, the closer the values are, the closer the resulting score is to 1.0.

In contrast, the HYBRID approach calculates the score as:
`Score = 1 / (k + rank_vector) + 1 / (k + rank_keyword)`.
As a result, even in the closest possible case, the score only reaches around 0.02.

For example:
`Score = (1 / (80 + 1)) + (1 / (80 + 1))`
`Score ≈ 0.012345 + 0.012345`
`Score ≈ 0.02469`

Because of this, all validations using `isCloseTo` end up failing.

image

dliubarskyi: I see, thanks for hte explanation!

> Score = 1 / (k + rank_vector) + 1 / (k + rank_keyword).
As a result, even in the closest possible case, the score only reaches around 0.02.


Where does this formula come from? I am not sure it is the right formula for our use case. For the closest possible case we should ideally return 1 instead of 0.02.


But even if we cannot normalize it and reach 1.0, we could extract calling `isCloseTo` in a template method and then override it in the hybrid subclass, WDYT? In this case all the tests will stay the same, only assertions will be different for hybrid.

YongGoose: > Where does this formula come from? I am not sure it is the right formula for our use case.

This part is extracted from the SQL statement used to calculate the score in the Hybrid approach.
If you look closely, you can see that the formula originates from this query.

If there is anything I’m misunderstanding, please let me know!

```sql
 SELECT * FROM (
   SELECT
     COALESCE(v.embedding_id, k.embedding_id) AS embedding_id,
     COALESCE(v.embedding, k.embedding) AS embedding,
     COALESCE(v.text, k.text) AS text
     %8$s,
     COALESCE(1.0 / (%9$d + v.rnk), 0.0) + COALESCE(1.0 / (%9$d + k.rnk), 0.0) AS score // ❗️formula comes from here!
   FROM vector_search v
   FULL OUTER JOIN keyword_search k ON v.embedding_id = k.embedding_id
 ) ranked
```

YongGoose: > For the closest possible case we should ideally return 1 instead of 0.02.

If we want to obtain values closer to 1, it seems we would need to adopt an approach other than the RRF method.

YongGoose: Done in https://github.com/langchain4j/langchain4j/pull/4288/commits/04c34629d729be322bf4f2aefedf07575396c007 !

PTAL :)

YongGoose:
@dliubarskyi 
Could you please take a look when you have time?

langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorHybridFilteringIT.java
```diff
@@ -0,0 +1,139 @@
...
+import dev.langchain4j.store.embedding.EmbeddingStore;
+import dev.langchain4j.store.embedding.EmbeddingStoreWithFilteringIT;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.containers.PostgreSQLContainer;
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+@Testcontainers
+class PgVectorHybridFilteringIT extends EmbeddingStoreWithFilteringIT {
```
dliubarskyi:
```suggestion
class PgVectorHybridSearchIT extends EmbeddingStoreWithFilteringIT {
```

langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorHybridFilteringIT.java
```diff
@@ -0,0 +1,139 @@
...
+                .password("test")
+                .database("test")
+                .table("test" + nextInt(1000, 2000))
+                .dimension(384)
+                .dropTableFirst(true)
+                .searchMode(HYBRID)
+                .rrfK(80)
+                .build();
+    }
+
```
dliubarskyi, line 44: This test class makes sure that existing `EmbeddingStore` features are working well, but there is no testing of the hybrid search itself, could you please add a few `@Test` methods here?

langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorHybridFilteringIT.java
```diff
@@ -0,0 +1,139 @@
...
+    @Override
+    protected boolean supportsContains() {
+        return true;
+    }
+
+    @Override
+    protected void assertScore(EmbeddingMatch<TextSegment> match, double expectedScore) {
+        // In Hybrid mode, the score is calculated using the Reciprocal Rank Fusion (RRF) algorithm.
```
dliubarskyi, line 62: Could you please document hybrid search feature under https://docs.langchain4j.dev/integrations/embedding-stores/pgvector?
Also, please document calculation of the score (you could copy this comment into the documentation as-is)

dliubarskyi (commented):
@YongGoose thank you, well done! Please address the remaining comments and we can merge!
FYI: the release is planned on 4 Feb after 13:00 CET, so if this PR will be ready, it will be merged and included in the release

langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorHybridFilteringIT.java
YongGoose: Done in https://github.com/langchain4j/langchain4j/pull/4288/commits/3d447e1315299709f10fd4f3533c2cb57fda0f18 !

docs/docs/integrations/embedding-stores/pgvector.md
```diff
@@ -43,7 +43,9 @@ RAG, and more.
...
+| `rrfK`                  | The constant `k` used in the RRF (Reciprocal Rank Fusion) algorithm: `Score = 1/(k + rank_vector) + 1/(k + rank_keyword)`. Lower values (20-40) emphasize top results more; higher values (80-100) create more balanced rankings. Only relevant when `searchMode` is set to `HYBRID`.                                                                                                                                                                          | `60`            | Optional. Only used in HYBRID search mode.                                                                                                                                                                                                                                        |
```
dliubarskyi, line 47: please add `textSearchConfig`

docs/docs/integrations/embedding-stores/pgvector.md
```diff
@@ -302,6 +304,107 @@ try {
...
+EmbeddingStore<TextSegment> embeddingStore = PgVectorEmbeddingStore.builder()
+        .host("localhost")
+        .port(5432)
+        .database("postgres")
+        .user("my_user")
+        .password("my_password")
+        .table("document_embeddings")
+        .dimension(embeddingModel.dimension())
+        .searchMode(SearchMode.HYBRID)  // Enable hybrid search (default: SearchMode.VECTOR)
+        .rrfK(60)                       // Optional: RRF algorithm parameter (default: 60)
```
dliubarskyi, line 333: please add `textSearchConfig` to the example

docs/docs/integrations/embedding-stores/pgvector.md
```diff
@@ -302,6 +304,107 @@ try {
...
+**Example Score Calculation** (with k=80):
+
+If a document ranks 1st in both vector and keyword search:
+```
+Score = 1/(80+1) + 1/(80+1)
+      = 1/81 + 1/81
+      = 0.0247
+```
+
+**Important**: RRF scores are rank-based and typically range from ~0.02 to ~1.0. This is different from cosine similarity scores (0.0 to 1.0) used in vector-only search.
```
dliubarskyi, line 385: Not sure about 0.02 to 1.0. I guess it should be 0.02...0?

docs/docs/integrations/embedding-stores/pgvector.md
```diff
@@ -302,6 +304,107 @@ try {
...
+
+#### Key Differences from Vector-Only Search
+
+| Aspect | Vector Search | Hybrid Search |
+|--------|--------------|---------------|
+| **Query Input** | Only `queryEmbedding` | Both `queryEmbedding` AND `query` text |
+| **Score Type** | Cosine similarity (0.0-1.0) | RRF rank-based score (~0.02-1.0) |
```
dliubarskyi, line 392: same here

dliubarskyi (approved):
@YongGoose thanks a lot!
Please address latest review comments in a separate PR. Thanks!

docs/docs/integrations/embedding-stores/pgvector.md
YongGoose: Oh... my mistake 😅

YongGoose:
> @YongGoose thanks a lot! Please address latest review comments in a separate PR. Thanks!

Please take a look :)
- https://github.com/langchain4j/langchain4j/pull/4539

PR #4289, opened 2025-12-25
feat: InMemoryEmbeddingStore add size() and isEmpty()

### LINKED ISSUE #4235: Improvements to Get number of embeddings in in-memory embedding store

We use in-memory embedding stores (`InMemoryEmbeddingStore`) and lazily populate them with data. There are a few shortcomings though, we should be rather simple to improve:

* Currently. it is not possible to inspect the in-memory store to see whether it already contains entries so we can decide whether to (re-)ingest data
* serializing and de-serializing is file-based and uses JSON strings. This does not scale for even medium number of embeddings as it keeps it all in memory

Being able to get the number of contained embeddings specifically for the in-memory store would solve that as a minimal solution (this does not have to be on the `EmbeddingStore` interface).
An even better solution would be to providing a new method to run the search using a lazy (and abortable) iteration (using an Iterator) which would allow us to determine whether some values (possibly matching some specific criteria using a search filter) already exist. This could be an extension to the `EmbeddingStore` interface with a default implementation using the existing `EmbeddingStore.search()´ method wrapped in an iterator as fallback.

To determine whether the in-memory store has values, one could use reflection to peek into the store's `entries` field, but that's against best practices.

**Additional context**
n/a

wangxuanni:
## Issue

Fixes #4235

## Change

Add `size()` and `isEmpty()` methods to `InMemoryEmbeddingStore` to allow inspecting whether the store already contains embeddings.

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
+++ b/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
@@ -1,5 +1,6 @@
 package dev.langchain4j.store.embedding.inmemory;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import dev.langchain4j.data.document.Metadata;
 import dev.langchain4j.data.embedding.Embedding;
 import dev.langchain4j.data.segment.TextSegment;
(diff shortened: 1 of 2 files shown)


YongGoose (approved):
It looks fine overall. However, it would be good to run spotless apply.

wangxuanni:
> It looks fine overall. However, it would be good to run spotless apply.

Thanks! I’ve run spotless:apply and pushed the formatting changes.

dliubarskyi (approved):
@wangxuanni thank you!

PR #4304, opened 2025-12-29
[feature] ChatMemory always keep system message on index 0

### LINKED ISSUE #1111: [FEATURE] keep SystemMessage first in ChatMemory

The behavior of `ChatMemory` implementations looks a bit weird in the position handling of SystemMessage.

- When `SystemMessage` is added first it is kept as first item
- When a `SystemMessage` message is added, either after `UserMessage` or after a first `SystemMessage`, it is placed at last position
  - during evictions, the `SystemMessage` position will go up in the Memory until it reaches the first place and remains there.

I'd like to standardize and simplify the position handling of the `SystemMessage` so that it is always kept as first message in the list.

I could reimplement (add my own Implementation) or proxify existing implementations: `MessageWindowChatMemory` and  `TokenWindowChatMemory`

**Additional context**

based on current state of `langchain4j:0.30.0`

May I propose something to solve this?

mr-cloud:
## Issue
Closes #1111 

## Change
allow ChatMemory to always keep system message on index 0, impl this feature for MessageWindowChatMemory/TokenWindowChatMemory.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
@@ -58,4 +58,11 @@ public interface ChatMemory {
      * Clears the chat memory.
      */
     void clear();
+
+    /**
+     * always keep system message on index 0
+     *
+     * @return true if system message should always be on index 0, false otherwise
+     */
+    boolean isSystemMessageFirst();
 }
--- a/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
@@ -58,11 +58,4 @@ public interface ChatMemory {
      * Clears the chat memory.
      */
     void clear();
-
-    /**
-     * always keep system message on index 0
-     *
-     * @return true if system message should always be on index 0, false otherwise
-     */
-    boolean isSystemMessageFirst();
 }
--- a/langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
+++ b/langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
@@ -27,6 +27,7 @@ import java.util.function.Function;
  * the most recent value returned by the provider.
  * <p>
  * Once added, a {@link SystemMessage} is always retained.
+ * If {@link MessageWindowChatMemory#systemMessageFirst} is true, always keep system message on index 0.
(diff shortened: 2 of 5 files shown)


langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -58,4 +58,11 @@ default void add(Iterable<ChatMessage> messages) {
      * Clears the chat memory.
      */
     void clear();
+
+    /**
+     * always keep system message on index 0
+     *
+     * @return true if system message should always be on index 0, false otherwise
+     */
+    boolean isSystemMessageFirst();
```
dliubarskyi: There are a couple of issues:
1. This is a breaking change. If you want to add this method, you need to provide meaningful default implementation, otherwise all custom implementations of this interface will break.
2. Do we really have to expose this knowledge? This seems more like an internal impleentation detail which should not be exposed

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
```diff
@@ -71,7 +73,11 @@ public void add(ChatMessage message) {
                 }
             }
         }
-        messages.add(message);
+        if (message instanceof SystemMessage && this.isSystemMessageFirst()) {
```
dliubarskyi:
```suggestion
        if (message instanceof SystemMessage && this.systemMessageFirst) {
```

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
```diff
@@ -162,6 +174,17 @@ public Builder chatMemoryStore(ChatMemoryStore store) {
             return this;
         }
 
+        /**
+         * Specifies whether the system message should be added to the beginning of the message list.
+         *
+         * @param systemMessageFirst
+         * @return
+         */
+        public Builder systemMessageFirst(boolean systemMessageFirst) {
```
dliubarskyi:
```suggestion
        public Builder alwaysKeepSystemMessageFirst(boolean systemMessageFirst) {
```
Please rename it everywhere

dliubarskyi (commented):
@mr-cloud thank you!

mr-cloud:
thankful review, i'll refine it.

mr-cloud:
i agree that we should not expose isSystemMessageFirst() at interface level cause it is just a flag and  custom impl. still needs to process with it in add() method. i update it as suggested , please review it again.

dliubarskyi (approved):
@mr-cloud thank you!

PR #4306, opened 2025-12-29
feat(openai): add support for logprobs and top_logprobs

### LINKED ISSUE #4298: [FEATURE] Support 'logprobs' in OpenAiChatModel for Confidence Scoring
Yes, currently `OpenAiChatModel` doesn't expose the `logprobs` parameter from the OpenAI API.

In the Python ecosystem (`ChatOpenAI`), using logprobs is standard practice for observability. Specifically for confidence scoring (checking if the model is hallucinating) or using the LLM as a classifier by checking token probability mass.

Right now, there's a feature gap here compared to the Python lib. If we want to build reliability checks or uncertainty quantification in Java, we're stuck having to use the raw client or custom wrappers.

I'd like to add support for `logprobs` in the `langchain4j-open-ai` module.

**1. Builder Config**
Update `OpenAiChatModel.builder()` to accept:
- `logprobs(Boolean logprobs)`
- `topLogprobs(Integer topLogprobs)`

**2. Response Mapping**
Since the `logprobs` data in the OpenAI response is nested in the `choices` object and can get pretty verbose, I think the best place for it is the `Response<AiMessage>` metadata.

**Implementation Plan:**
- Update the internal `OpenAiClient` DTOs to handle the new request params.
- Update `OpenAiMapper` to pull `content.logprobs` from the raw response.
- Map it into the standard `Response` metadata (maybe under a key like `openai_logprobs`) so we don't clutter the main `AiMessage` content.

- Using the raw OpenAI client, but that defeats the purpose of using LangChain4j's abstractions.
- Writing a custom wrapper around the ChatModel, but that duplicates a lot of the HTTP handling logic and is hard to maintain.

**Additional context**
I ran into this while building an internal eval framework for hallucination detection. Bridging this gap is pretty key for teams trying to build "Safe AI" workflows (like toxicity detection or uncertainty measurement) in Java without relying on sidecar Python services.

**I'm happy to implement this.**
I've already looked at `OpenAiChatModel` and `OpenAiMapper`. If the maintainers are okay with the metadata approach, I can raise a PR with the implementation and tests.

ss-bhatt:
## Issue

Closes #4298

### Summary
Implements `logprobs` and `top_logprobs` support for `OpenAiChatModel`. Users can now retrieve log probability information from OpenAI responses, which is useful for assessing model confidence and exploring alternative token choices.

### Implementation Details
Added two new parameters to `OpenAiChatModel`:
- `logprobs(Boolean)` - enables log probability output
- `topLogprobs(Integer)` - specifies number of alternative tokens to return (0-20)

**New DTOs:**
Created internal DTOs (`LogProbs`, `TokenLogProb`, `TopLogProb`) to strictly match OpenAI's API structure.

**Metadata Exposure:**
As suggested by @dliubarskyi, the logprobs data is exposed via `OpenAiChatResponseMetadata` rather than the generic metadata map, ensuring type safety.

### Example Usage
```java
    .messages(UserMessage.from("Hello"))
    .parameters(OpenAiChatRequestParameters.builder()
        .logprobs(true)
        .topLogprobs(2)
        .build())
    .build();

// Access strongly-typed metadata
OpenAiChatResponseMetadata metadata = (OpenAiChatResponseMetadata) response.metadata();
LogProbs logprobs = metadata.logprobs(); 
```

### Testing
- Added integration test should_return_logprobs in OpenAiChatModelIT.
- Verified that parameters are correctly sent to the API.
- Verified that the response structure (including nested top_logprobs) is correctly parsed into the new DTOs.
- Tested successfully against the live OpenAI API.

## General checklist


langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiChatModelIT.java
```diff
@@ -418,4 +418,48 @@ record WebSearchOptions(@JsonProperty("user_location") UserLocation userLocation
...
+        // then
+        assertThat(response.aiMessage().text()).isNotBlank();
+        assertThat(response.metadata()).isInstanceOf(OpenAiChatResponseMetadata.class);
+
+        OpenAiChatResponseMetadata metadata = (OpenAiChatResponseMetadata) response.metadata();
+        assertThat(metadata.logprobs()).isNotNull();
+        assertThat(metadata.logprobs().content()).isNotEmpty();
+
+        // Verify first token has logprob data
+        dev.langchain4j.model.openai.internal.chat.TokenLogProb firstToken =
```
dliubarskyi: I am not sure it is a good idea to expose `internal` classes to the user.

langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiChatModelIT.java
```diff
@@ -418,4 +418,48 @@ record WebSearchOptions(@JsonProperty("user_location") UserLocation userLocation
...
+        // when
+        ChatResponse response = model.chat(chatRequest);
+
+        // then
+        assertThat(response.aiMessage().text()).isNotBlank();
+        assertThat(response.metadata()).isInstanceOf(OpenAiChatResponseMetadata.class);
+
+        OpenAiChatResponseMetadata metadata = (OpenAiChatResponseMetadata) response.metadata();
+        assertThat(metadata.logprobs()).isNotNull();
+        assertThat(metadata.logprobs().content()).isNotEmpty();
```
dliubarskyi: would it be possible to avoid one nesting layer, so that `logprobs()` returns a list?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/TokenLogProb.java
```diff
@@ -0,0 +1,130 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import java.util.List;
+import java.util.Objects;
+
+@JsonDeserialize(builder = TokenLogProb.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public final class TokenLogProb {
```
dliubarskyi: It seems that `TokenLogProb` and `TopLogProb` are identical, except `TokenLogProb` can contain more nested `TopLogProb`. When exposing this to the user, what do you think about using a single type? So that `LogProb` has a `List<LogProb>`? After one layer of nesting, the list will be empty. WDYT?

dliubarskyi (commented):
@ss-bhatt thank you!

langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiChatModelIT.java
dliubarskyi: Let's create a `LogProb` class in the `dev.langchain4j.model.openai` package?

langchain4j-open-ai/src/test/java/dev/langchain4j/model/openai/OpenAiChatModelIT.java
dliubarskyi: `List<LogProb>` specifically

ss-bhatt:
Thank you for the review, @dliubarskyi .

Makes sense. I agree that exposing the internal hierarchy is a bit messy, so a single public LogProb class is definitely cleaner for the user. I'll refactor it to flatten the structure and merge TopLogProb / TokenLogProb into one. Will push the updates shortly!

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/TokenLogProb.java
ss-bhatt: Done. I've refactored this based on your feedback:

- Replaced TokenLogProb and `TopLogProb` with a single public, recursive `LogProb` class.
- Flattened the return type so `metadata.logprobs()` now returns `List<LogProb>` directly.
- Kept the Jackson mapping internal to keep the public API clean.

Pushed the fixes!

ss-bhatt:
Makes sense. I agree on your point about exposing internal classes. A single public LogProb class is definitely cleaner. I have flattened the structure and merged TopLogProb / TokenLogProb into one. Pushed the updates.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/LogProb.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Objects;
+
+/**
+ * Represents log probability information for a token.
+ * Contains the token itself, its log probability, and optionally the most likely alternative tokens.
+ */
+@JsonDeserialize(builder = LogProb.Builder.class)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
+public final class LogProb {
```
dliubarskyi, line 6: Can we please have a separate `LogProb` class that we expose to our users? The one that does not have Jackson annotations and deser logic

dliubarskyi: And this class should be moved to the internal, close to the `LogProbs` class

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatResponseMetadata.java
```diff
@@ -65,14 +68,19 @@ public List<ServerSentEvent> rawServerSentEvents() {
         return rawServerSentEvents;
     }
 
+    public List<LogProb> logprobs() {
```
dliubarskyi:
```suggestion
    public List<LogProb> logProbs() {
```

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatResponseMetadata.java
```diff
@@ -145,6 +162,11 @@ public Builder rawServerSentEvents(List<ServerSentEvent> rawServerSentEvents) {
             return this;
         }
 
+        public Builder logprobs(LogProbs logprobs) {
```
dliubarskyi:
```suggestion
        public Builder logProbs(LogProbs logprobs) {
```
Please rename it everywhere

dliubarskyi (commented):
@ss-bhatt thank you!

ss-bhatt:
Thanks for the review @dliubarskyi ! I'll address these comments this week to get it ready for 1.12.0

ss-bhatt:
@dliubarskyi : Addressed the Feb 3 feedback. Ready for another look when you get a chance.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatResponseMetadata.java
```diff
@@ -65,14 +69,36 @@ public List<ServerSentEvent> rawServerSentEvents() {
         return rawServerSentEvents;
     }
 
+    public List<LogProb> logProbs() {
+        if (logProbs == null) return null;
+        return logProbs.content().stream()
```
dliubarskyi: why does this conversion has to be done here? Can't it be done before creating `OpenAiChatResponseMetadata`?

ss-bhatt:
@dliubarskyi: Done. Moved it to OpenAiUtils.logProbsFrom(), same pattern as tokenUsageFrom.

dliubarskyi (approved):
@ss-bhatt thank you!

PR #4349, opened 2026-01-02
feat: add listeners for embedding model, embedding store and content retriever

### LINKED ISSUE #4273: [FEATURE] Observability support for EmbeddingModel via Listeners
Currently, the ChatModel in LangChain4j provides a robust observability mechanism through ChatModelListener, allowing developers to intercept requests, responses, and errors. However, this same level of observability is missing for EmbeddingModel. In enterprise applications where tracking token usage, latency, and input/output data for all AI operations is crucial for auditing and cost management, the lack of listeners in EmbeddingModel creates a significant gap.

I would like to have the implementation of EmbeddingModelListener (and associated context classes like EmbeddingModelRequestContext, EmbeddingModelResponseContext, and EmbeddingModelErrorContext) in langchain4j-core, following the same pattern as ChatModelListener.
Specifically:
1.Define the interfaces: Create EmbeddingModelListener with methods onRequest, onResponse, and onError.
2.Context Classes: Implement the necessary context objects to carry the request (e.g., textSegments), the response (e.g., embeddings and TokenUsage), and any error information.
3.Model Support: Update the base EmbeddingModel and its builders to accept a list of listeners and trigger them during the embedding process.
4.Provider Implementation: Update popular implementations (OpenAI, HuggingFace, etc.) to support these listeners.

One alternative is to manually wrap every EmbeddingModel call with logging or monitoring logic. However, this is error-prone, leads to code duplication, and doesn't integrate well with standard observability frameworks that could otherwise simply register a listener.

**Additional context**
The implementation should ideally mirror the ChatModelListener structure found in the dev.langchain4j.model.chat.listener package. This consistency will make it easier for users already familiar with LangChain4j observability to apply the same patterns to their embedding workflows.

### LINKED ISSUE #4274: [FEATURE] Observability support for EmbeddingStore and ContentRetriever via Listeners
Following the addition of listener for ChatModel, the VectorStore (represented by EmbeddingStore and ContentRetriever) remains a "black box" in terms of observability. In production environments, it is essential to monitor:
1.Latency: How long vector searches take.
2.Accuracy/Relevance: Capturing the scores of retrieved documents.
3.Payloads: Seeing the raw query vector and the metadata of the returned results.
4.Errors: Catching database connection issues or search failures within the store implementation.

I would like to see the implementation of EmbeddingStoreListener and ContentRetrieverListener (along with their respective RequestContext, ResponseContext, and ErrorContext classes) in langchain4j-core.
Specifically:
1.EmbeddingStoreListener: Should intercept calls to add, addAll, and search.
 ◦onRequest: Access to the Embedding and maxResults.
 ◦onResponse: Access to the list of EmbeddingMatch.
2.ContentRetrieverListener: Should intercept calls to retrieve(Query).
 ◦onRequest: Access to the Query object.
 ◦onResponse: Access to the list of Content (segments and metadata).
3.Model Support: Update builders for common stores (like MongoDbEmbeddingStore) and retrievers (like EmbeddingStoreContentRetriever) to accept and trigger these listeners.

The current alternative is to create custom wrappers for every EmbeddingStore implementation, which is redundant and difficult to maintain across different providers (MongoDB, Pinecone, Weaviate, etc.).

**Additional context**
This implementation should follow the pattern established by ChatModelListener, ensuring a consistent developer experience for observability across the entire LangChain4j ecosystem.

LucasGois1:
## Issue
Closes #4274
Closes #4273

---

## Change
This PR adds listener-based observability hooks for RAG components, similar to the existing `ChatModelListener` pattern, so that  `EmbeddingModel`, `EmbeddingStore` and `ContentRetriever` are no longer “black boxes”.

---

#### EmbeddingStore observability
- Added `EmbeddingStoreListener` API in `langchain4j-core`:
  - `EmbeddingStoreListener` with `onRequest/onResponse/onError`
  - `EmbeddingStoreRequestContext`, `EmbeddingStoreResponseContext`, `EmbeddingStoreErrorContext`
  - `EmbeddingStoreOperation` to identify the operation being observed (e.g., ADD, ADD_ALL, SEARCH, REMOVE, etc.)
- Added `EmbeddingStore#addListener(...)` and `EmbeddingStore#addListeners(...)` default methods returning an observing store wrapper.
- Implemented a generic wrapper `ObservingEmbeddingStore` that delegates to the real store and fires listeners around operations.

#### ContentRetriever observability
- Added `ContentRetrieverListener` API in `langchain4j-core`:
  - `ContentRetrieverListener` with `onRequest/onResponse/onError`
  - `ContentRetrieverRequestContext`, `ContentRetrieverResponseContext`, `ContentRetrieverErrorContext`
- Added `ContentRetriever#addListener(...)` and `ContentRetriever#addListeners(...)` default methods returning an observing retriever wrapper.
- Implemented `ObservingContentRetriever` to intercept `retrieve(Query)` and dispatch events.
- **Important:** `ContentRetriever#addListeners` uses `Collection<ContentRetrieverListener>` to avoid Java type-erasure signature clashes for implementations that also implement `EmbeddingStore` (e.g., `AzureAiSearchContentRetriever`).

#### EmbeddingModel observability
- Added `EmbeddingModelListener` API in `langchain4j-core`:
  - `EmbeddingModelListener` with `onRequest/onResponse/onError`
  - `EmbeddingModelRequestContext`, `EmbeddingModelResponseContext`, `EmbeddingModelErrorContext`
- Added `EmbeddingModel#addListener(...)` and `EmbeddingModel#addListeners(...)` default methods returning an observing model wrapper.
- Implemented `ObservingEmbeddingModel` which intercepts `embedAll(...)` (the core method all embeddings flow through) and fires listeners around the delegate call.
- **Note:** `EmbeddingModel#addListeners` uses `Iterable<EmbeddingModelListener>` to reduce the risk of future type-erasure clashes with other interfaces.

### How it works (wrapper/decorator approach)
Instead of modifying each individual model/store/retriever implementation (MongoDB, Pinecone, Azure AI Search, etc.), the change introduces **generic wrappers** that implement the same interfaces and delegate to the real implementation:
- `ObservingEmbeddingStore` wraps any `EmbeddingStore`
- `ObservingContentRetriever` wraps any `ContentRetriever`
- `ObservingEmbeddingModel` wraps any `EmbeddingModel`

Users enable observability by calling:
- `EmbeddingStore#addListener(s)` which returns an observing `EmbeddingStore`
- `ContentRetriever#addListener(s)` which returns an observing `ContentRetriever`
- `EmbeddingModel#addListener(s)` which returns an observing `EmbeddingModel`

This ensures:
- **No provider-specific duplication** (single implementation works across all integrations)
- **No breaking changes** for existing store/retriever implementations (no required code changes)
- **Consistent behavior** with `ChatModelListener`:
  - listeners are invoked in list order
  - listeners are invoked synchronously
  - listener exceptions are logged and ignored so they do not break user calls
  - `attributes` map allows passing data across request/response/error and across listeners

---

### Why wrapper/decorator was chosen
Unlike `ChatModel`, which has a central default method (`ChatModel#chat`) that can be instrumented once, `EmbeddingStore` and `ContentRetriever` are “thin” interfaces without a single shared orchestration method. The wrapper/decorator approach provides a non-invasive, reusable interception point that can be applied to any implementation without requiring per-provider wrappers or modifying every integration.

### Note on Azure AI Search compilation conflict
Some integrations (e.g., Azure AI Search retriever) implement/extend both `EmbeddingStore` and `ContentRetriever`. To avoid Java type erasure signature clashes, `ContentRetriever#addListeners` uses a `Collection<ContentRetrieverListener>` while `EmbeddingStore#addListeners` uses a `List<EmbeddingStoreListener>`. This keeps the API ergonomic (still accepts `List.of(...)`) and avoids conflicts for implementations that inherit both interfaces.

### Tests
Added unit tests in `langchain4j-core` mirroring `ChatModelListenerTest` coverage:
- listener invocation order
- listener exception handling (ignored/logged)
- attributes propagation across listeners
- `onError` invocation when delegate throws
- null/empty listener lists handling

---

## General checklist


dliubarskyi (commented):
@LucasGois1 there are too many formatting changes, could you please revert?

LucasGois1:
> @LucasGois1 there are too many formatting changes, could you please revert?

done

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/listener/EmbeddingModelErrorContext.java
```diff
@@ -0,0 +1,56 @@
...
+ * The embedding model error context.
+ * It contains the error, corresponding input, the {@link EmbeddingModel} and attributes.
+ * The attributes can be used to pass data between methods of an {@link EmbeddingModelListener}
+ * or between multiple {@link EmbeddingModelListener}s.
+ */
+public class EmbeddingModelErrorContext {
+
+    private final Throwable error;
+    private final List<TextSegment> textSegments;
+    private final EmbeddingModel embeddingModel;
```
dliubarskyi, line 24: Why is `EmbeddingModel` required?

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/listener/EmbeddingModelRequestContext.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * The embedding model request context.
+ * It contains the input, the {@link EmbeddingModel} and attributes.
+ * The attributes can be used to pass data between methods of an {@link EmbeddingModelListener}
+ * or between multiple {@link EmbeddingModelListener}s.
+ */
+public class EmbeddingModelRequestContext {
+
+    private final List<TextSegment> textSegments;
+    private final EmbeddingModel embeddingModel;
```
dliubarskyi, line 23: same here

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/listener/EmbeddingModelResponseContext.java
```diff
@@ -0,0 +1,58 @@
...
+ * The embedding model response context.
+ * It contains the {@link Response}, corresponding input, the {@link EmbeddingModel} and attributes.
+ * The attributes can be used to pass data between methods of an {@link EmbeddingModelListener}
+ * or between multiple {@link EmbeddingModelListener}s.
+ */
+public class EmbeddingModelResponseContext {
+
+    private final Response<List<Embedding>> response;
+    private final List<TextSegment> textSegments;
+    private final EmbeddingModel embeddingModel;
```
dliubarskyi, line 26: same here

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
```diff
@@ -3,17 +3,50 @@
...
  */
 public interface EmbeddingModel {
 
+    /**
+     * Wraps this {@link EmbeddingModel} with an observing model that dispatches events to the provided listener.
+     *
+     * @param listener The listener to add.
+     * @return An observing {@link EmbeddingModel} that will dispatch events to the provided listener.
+     */
+    default EmbeddingModel addListener(EmbeddingModelListener listener) {
```
dliubarskyi: please move new methods to the bottom of the interface

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/listener/EmbeddingModelErrorContext.java
LucasGois1: I think it’s useful when the same listener instance is attached to multiple models (to disambiguate which model fired the event), and it allows extracting model-specific info (e.g., modelName() / dimension() / custom wrapper type) without relying on external closures. But no problem about removing this if you want

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/ContentRetriever.java
```diff
@@ -26,6 +29,34 @@
  */
 public interface ContentRetriever {
 
+    /**
+     * Wraps this {@link ContentRetriever} with an observing retriever that dispatches events to the provided listener.
+     *
+     * @param listener The listener to add.
+     * @return An observing {@link ContentRetriever} that will dispatch events to the provided listener.
+     */
+    default ContentRetriever addListener(ContentRetrieverListener listener) {
```
dliubarskyi, line 51: same here

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStore.java
```diff
@@ -1,26 +1,57 @@
...
 
+    /**
+     * Wraps this {@link EmbeddingStore} with an observing store that dispatches events to the provided listener.
+     * <p>
+     * This is a non-breaking convenience method to add observability to any {@link EmbeddingStore} implementation.
+     *
+     * @param listener The listener to add.
+     * @return An observing {@link EmbeddingStore} that will dispatch events to the provided listener.
+     */
+    default EmbeddingStore<Embedded> addListener(EmbeddingStoreListener listener) {
```
dliubarskyi: same here

LucasGois1: done

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/ContentRetriever.java
LucasGois1: done

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
LucasGois1: done

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreRequestContext.java
```diff
@@ -0,0 +1,132 @@
...
+    private final List<Embedding> embeddings;
+
+    private final Embedded embedded;
+    private final List<Embedded> embeddedList;
+
+    private final EmbeddingSearchRequest searchRequest;
+    private final Filter filter;
+
+    public EmbeddingStoreRequestContext(
+            EmbeddingStoreOperation operation,
```
dliubarskyi: This ctor won't scale, pease introduce a `Builder` in this class and instead accept a single `Builder` in this ctor

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreOperation.java
```diff
@@ -0,0 +1,14 @@
+package dev.langchain4j.store.embedding.listener;
+
+/**
+ * Operation executed against an embedding store.
+ */
+public enum EmbeddingStoreOperation {
+    ADD,
+    ADD_ALL,
```
dliubarskyi: do we need to distinguish between ADD and ADD_ALL?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreErrorContext.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private final Throwable error;
+    private final EmbeddingStoreOperation operation;
+    private final EmbeddingStore<Embedded> embeddingStore;
+    private final Map<Object, Object> attributes;
+
+    private final EmbeddingStoreRequestContext<Embedded> requestContext;
+
+    public EmbeddingStoreErrorContext(
+            Throwable error,
```
dliubarskyi: same here

dliubarskyi: (builder)

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/listener/EmbeddingModelErrorContext.java
```diff
@@ -0,0 +1,56 @@
...
+    private final Map<Object, Object> attributes;
+
+    public EmbeddingModelErrorContext(
+            Throwable error, List<TextSegment> textSegments, EmbeddingModel embeddingModel, Map<Object, Object> attributes) {
```
dliubarskyi: same here, please introcude a `Builder` in all new `*Context` classes in this PR

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/listener/EmbeddingModelErrorContext.java
dliubarskyi: make sense, thanks!

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreOperation.java
dliubarskyi: pls ignore

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreRequestContext.java
```diff
@@ -0,0 +1,132 @@
...
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The embedding store request context.
+ * It contains operation details and attributes.
+ * The attributes can be used to pass data between methods of an {@link EmbeddingStoreListener}
+ * or between multiple {@link EmbeddingStoreListener}s.
+ */
+public class EmbeddingStoreRequestContext<Embedded> {
```
dliubarskyi: It is unfortunate that `Embedded` is still messing around...
Do you think it is useful? All of our `EmbeddingStore` implementations are EmbeddingStore<TextSegment>, so `Embedded` is always of type `TextSegment`

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreRequestContext.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private final EmbeddingStoreOperation operation;
+    private final EmbeddingStore<Embedded> embeddingStore;
+    private final Map<Object, Object> attributes;
+
+    private final String id;
+    private final List<String> ids;
```
dliubarskyi: It is not nice to have this single object representing hals a dozen of operations with most of the fileds `null`. Perhaps let's go with a subtype per operation?

dliubarskyi: but please don't make it sealed, we want to be able to add more operations in the future

dliubarskyi (commented):
@LucasGois1 thank you, well done! Very useful and long awaited feature!

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreRequestContext.java
LucasGois1: EmbeddingStore is intentionally generic. As I see, we already have generic implementations (e.g., InMemoryEmbeddingStore<Embedded>) and custom user stores may embed types other than TextSegment, (I don't know if exist another way to use embeddings, but once this was created to accept generic instead TextSegment as default type, users maybe created something different?)

I'm following the current design and if makes sense, we can discuss about this and do some kind of refactor to next major version

LucasGois1:
@dliubarskyi can you take a look again? let me know if my fixes are good for you

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreRequestContext.java
dliubarskyi: OK, makes sense, let's keep the existing appoach

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/ContentRetriever.java
```diff
@@ -35,4 +38,32 @@ public interface ContentRetriever {
...
+     * @return An observing {@link ContentRetriever} that will dispatch events to the provided listeners.
+     */
+    default ContentRetriever addListeners(Collection<ContentRetrieverListener> listeners) {
+        if (isNullOrEmpty(listeners)) {
+            return this;
+        }
+        if (this instanceof ObservingContentRetriever observingContentRetriever) {
+            return observingContentRetriever.withAdditionalListeners(listeners);
+        }
+        return new ObservingContentRetriever(this, List.copyOf(listeners));
```
dliubarskyi: nit: it would be safer to make a `Utils.copy(listeners)` inside the ctor

dliubarskyi: same for all other places with copying

dliubarskyi: (all `Observing*` ctors)

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/listener/ContentRetrieverErrorContext.java
```diff
@@ -0,0 +1,89 @@
...
+import dev.langchain4j.rag.query.Query;
+import java.util.Map;
+
+/**
+ * The content retriever error context.
+ * It contains the error, corresponding {@link Query}, {@link ContentRetriever} and attributes.
+ * The attributes can be used to pass data between methods of a {@link ContentRetrieverListener}
+ * or between multiple {@link ContentRetrieverListener}s.
+ */
+public class ContentRetrieverErrorContext {
```
dliubarskyi, line 19: Could you please annotate all new classes and public methods (in existing interfaces) with `@Experimental`?
It would also be great to add `@since 1.11.0` to the javadoc

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/ObservingEmbeddingStore.java
```diff
@@ -0,0 +1,233 @@
...
+
+@Internal
+final class ObservingEmbeddingStore<Embedded> implements EmbeddingStore<Embedded> {
+
+    private final EmbeddingStore<Embedded> delegate;
+    private final List<EmbeddingStoreListener> listeners;
+
+    ObservingEmbeddingStore(EmbeddingStore<Embedded> delegate, List<EmbeddingStoreListener> listeners) {
+        this.delegate = ensureNotNull(delegate, "delegate");
+        this.listeners = ensureNotNull(listeners, "listeners");
```
dliubarskyi:
```suggestion
        this.listeners = copy(listeners);
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/ObservingEmbeddingStore.java
```diff
@@ -0,0 +1,233 @@
...
+    @Override
+    public void remove(String id) {
+        Map<Object, Object> attributes = new ConcurrentHashMap<>();
+        EmbeddingStoreRequestContext.Remove<Embedded> requestContext =
+                new EmbeddingStoreRequestContext.Remove<>(this, attributes, id);
+        onRequest(requestContext, listeners);
+        try {
+            delegate.remove(id);
+            EmbeddingStoreResponseContext.Void<Embedded> responseContext =
+                    new EmbeddingStoreResponseContext.Void<>(requestContext, attributes);
```
dliubarskyi:
```suggestion
                    new EmbeddingStoreResponseContext.Remove<>(requestContext, attributes);
```
Can we keep request and response types symmetrical?

dliubarskyi: I would not base the name of the type on the return type, but rather on the operation that is being done

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/ObservingEmbeddingStore.java
```diff
@@ -0,0 +1,233 @@
...
+    public void removeAll(Collection<String> ids) {
+        Map<Object, Object> attributes = new ConcurrentHashMap<>();
+        List<String> idsAsList = (ids == null) ? null : new ArrayList<>(ids);
+        EmbeddingStoreRequestContext.RemoveAllIds<Embedded> requestContext =
+                new EmbeddingStoreRequestContext.RemoveAllIds<>(this, attributes, idsAsList);
+        onRequest(requestContext, listeners);
+        try {
+            delegate.removeAll(ids);
+            EmbeddingStoreResponseContext.Void<Embedded> responseContext =
+                    new EmbeddingStoreResponseContext.Void<>(requestContext, attributes);
```
dliubarskyi:
```suggestion
                    new EmbeddingStoreResponseContext.RemoveAllIds<>(requestContext, attributes);
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/ObservingEmbeddingStore.java
```diff
@@ -0,0 +1,233 @@
...
+    @Override
+    public void removeAll(Filter filter) {
+        Map<Object, Object> attributes = new ConcurrentHashMap<>();
+        EmbeddingStoreRequestContext.RemoveAllFilter<Embedded> requestContext =
+                new EmbeddingStoreRequestContext.RemoveAllFilter<>(this, attributes, filter);
+        onRequest(requestContext, listeners);
+        try {
+            delegate.removeAll(filter);
+            EmbeddingStoreResponseContext.Void<Embedded> responseContext =
+                    new EmbeddingStoreResponseContext.Void<>(requestContext, attributes);
```
dliubarskyi:
```suggestion
                    new EmbeddingStoreResponseContext.RemoveAllFilter<>(requestContext, attributes);
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/ObservingEmbeddingStore.java
```diff
@@ -0,0 +1,233 @@
...
+
+    @Override
+    public void removeAll() {
+        Map<Object, Object> attributes = new ConcurrentHashMap<>();
+        EmbeddingStoreRequestContext.RemoveAll<Embedded> requestContext =
+                new EmbeddingStoreRequestContext.RemoveAll<>(this, attributes);
+        onRequest(requestContext, listeners);
+        try {
+            delegate.removeAll();
+            EmbeddingStoreResponseContext.Void<Embedded> responseContext =
```
dliubarskyi:
```suggestion
            EmbeddingStoreResponseContext.RemoveAll<Embedded> responseContext =
```

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/ObservingEmbeddingModel.java
```diff
@@ -0,0 +1,78 @@
...
+        if (additionalListeners == null || additionalListeners.isEmpty()) {
+            return this;
+        }
+        List<EmbeddingModelListener> merged = new ArrayList<>(listeners);
+        merged.addAll(additionalListeners);
+        return new ObservingEmbeddingModel(delegate, merged);
+    }
+
+    @Override
+    public Response<List<Embedding>> embedAll(List<TextSegment> textSegments) {
```
dliubarskyi: It is probably not enough to implement only `embedAll` method, as there are others that can be overriden in the implementations. For example, `embed(String)` is overriden in `GoogleAiEmbeddingModel` and never calls `embedAll()`, so it will not be ovserved

dliubarskyi: Same for `embed(TextSegment)`

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreOperation.java
```diff
@@ -0,0 +1,14 @@
+package dev.langchain4j.store.embedding.listener;
+
+/**
+ * Operation executed against an embedding store.
+ */
+public enum EmbeddingStoreOperation {
+    ADD,
```
dliubarskyi: I am wondering if it would make sense to have only 3 operations: add, search and remove.
All the `add`/`addAll` methods do not differ much, they are basically the same operation, just slightly different methods for convenience. Not sure anyone would be interested to know which exactly method was called.

Same for remove. We could only have a single remove enum and request/response subtype which will contain both list of IDs and filter. Depending on the method called, one of them will be `null`. For `removeAll()` everything will be `null`. WDYT?

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreRequestContext.java
```diff
@@ -0,0 +1,239 @@
...
+         */
+        public EmbeddingSearchRequest searchRequest() {
+            return searchRequest;
+        }
+    }
+
+    /**
+     * The {@code remove(String)} request context.
+     */
+    public static final class Remove<Embedded> extends EmbeddingStoreRequestContext<Embedded> {
```
dliubarskyi, line 180:
```suggestion
    public static final class RemoveById<Embedded> extends EmbeddingStoreRequestContext<Embedded> {
```
This comment is valid only in case we decide to keep the type-per-method approach

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreRequestContext.java
```diff
@@ -0,0 +1,239 @@
...
+         */
+        public String id() {
+            return id;
+        }
+    }
+
+    /**
+     * The {@code removeAll(ids)} request context.
+     */
+    public static final class RemoveAllIds<Embedded> extends EmbeddingStoreRequestContext<Embedded> {
```
dliubarskyi, line 203:
```suggestion
    public static final class RemoveByIds<Embedded> extends EmbeddingStoreRequestContext<Embedded> {
```
This comment is valid only in case we decide to keep the type-per-method approach

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreRequestContext.java
```diff
@@ -0,0 +1,239 @@
...
+         */
+        public List<String> ids() {
+            return ids;
+        }
+    }
+
+    /**
+     * The {@code removeAll(Filter)} request context.
+     */
+    public static final class RemoveAllFilter<Embedded> extends EmbeddingStoreRequestContext<Embedded> {
```
dliubarskyi, line 226:
```suggestion
    public static final class RemoveByFilter<Embedded> extends EmbeddingStoreRequestContext<Embedded> {
```
This comment is valid only in case we decide to keep the type-per-method approach

dliubarskyi (commented):
@LucasGois1 thanks a lot!

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/ObservingEmbeddingModel.java
```diff
@@ -0,0 +1,78 @@
...
+import dev.langchain4j.model.embedding.listener.EmbeddingModelRequestContext;
+import dev.langchain4j.model.embedding.listener.EmbeddingModelResponseContext;
+import dev.langchain4j.model.output.Response;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@Internal
+final class ObservingEmbeddingModel implements EmbeddingModel {
```
dliubarskyi:
```suggestion
final class ListeningEmbeddingModel implements EmbeddingModel {
```
To be consistent with `EmbeddingModelListener`

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/EmbeddingModel.java
```diff
@@ -66,4 +69,34 @@ default int dimension() {
     default String modelName() {
         return "unknown";
     }
+
+    /**
+     * Wraps this {@link EmbeddingModel} with an observing model that dispatches events to the provided listener.
```
dliubarskyi:
```suggestion
     * Wraps this {@link EmbeddingModel} with a listening model that dispatches events to the provided listener.
```

Please rename all the "observ..." instances in this PR

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/ObservingContentRetriever.java
```diff
@@ -0,0 +1,73 @@
...
+import dev.langchain4j.rag.content.retriever.listener.ContentRetrieverResponseContext;
+import dev.langchain4j.rag.query.Query;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@Internal
+final class ObservingContentRetriever implements ContentRetriever {
```
dliubarskyi:
```suggestion
final class ListeningContentRetriever implements ContentRetriever {
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/ObservingEmbeddingStore.java
```diff
@@ -0,0 +1,233 @@
...
+import dev.langchain4j.store.embedding.listener.EmbeddingStoreRequestContext;
+import dev.langchain4j.store.embedding.listener.EmbeddingStoreResponseContext;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+@Internal
+final class ObservingEmbeddingStore<Embedded> implements EmbeddingStore<Embedded> {
```
dliubarskyi:
```suggestion
final class ListeningEmbeddingStore<Embedded> implements EmbeddingStore<Embedded> {
```

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreOperation.java
dliubarskyi: BTW, do we still need this enum, given that we already have different subtypes to distinguish operations?

dliubarskyi:
Sorry, comminted in the wrong branch :/

langchain4j-core/src/main/java/dev/langchain4j/store/embedding/listener/EmbeddingStoreOperation.java
LucasGois1: Let's start simple and remove the enum, maybe the different subtypes can be enough

dliubarskyi (approved):
@LucasGois1 great job, thanks a lot!

PR #4350, opened 2026-01-02
Support streaming agents

### LINKED ISSUE #3519: [FEATURE] agent supports streamChatModel

mariofusco:
## Issue

Closes #3519

## Change

Allows to use streaming chat models with agents. Check the section added in the agents documentation to find how this works.

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/StreamingResponse.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/StreamingResponse.java
@@ -6,22 +6,13 @@ import java.util.concurrent.CompletableFuture;
 
 public class StreamingResponse implements DelayedResponse<String> {
 
-    private final TokenStream tokenStream;
-
-    private final StringBuilder answerBuilder = new StringBuilder();
     private final CompletableFuture<ChatResponse> futureResponse = new CompletableFuture<>();
 
     public StreamingResponse(TokenStream tokenStream) {
-        this.tokenStream = tokenStream;
         tokenStream
-                .onPartialResponse(answerBuilder::append)
                 .onCompleteResponse(futureResponse::complete)
                 .onError(futureResponse::completeExceptionally)
                 .start();
-    }
-
-    public TokenStream tokenStream() {
-        return tokenStream;
     }
 
     @Override
@@ -31,8 +22,7 @@ public class StreamingResponse implements DelayedResponse<String> {
 
     @Override
     public String blockingGet() {
-        futureResponse.join();
-        return answerBuilder.toString();
+        return futureResponse.join().aiMessage().text();
     }
 
     @Override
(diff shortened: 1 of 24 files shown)


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/StreamingResponse.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Override
+    public boolean isDone() {
+        return futureResponse.isDone();
+    }
+
+    @Override
+    public String blockingGet() {
+        futureResponse.join();
```
dliubarskyi:
```suggestion
        return futureResponse.join().aiMessage().text();
```
In case LLM called one or more tools, it is possible that `answerBuilder` contains not only the final response, but everything that LLM said before calling tools. For example: "Sure, let me call the tool X. OK, I am going to call tool Y. [Final response of the final request is here]". I guess it is more safe to just return text from `futureResponse`.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/StreamingResponse.java
```diff
@@ -0,0 +1,42 @@
...
+
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.service.TokenStream;
+import java.util.concurrent.CompletableFuture;
+
+public class StreamingResponse implements DelayedResponse<String> {
+
+    private final TokenStream tokenStream;
+
+    private final StringBuilder answerBuilder = new StringBuilder();
```
dliubarskyi:
```suggestion
```
This is not needed any more

dliubarskyi (commented):

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/StreamingResponse.java
mariofusco: Fixed.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/StreamingResponse.java
mariofusco: Fixed.

mariofusco:
I removed some unnecessary code in `StreamingResponse`, it should be ok now.

dliubarskyi (approved):
@mariofusco thank you!

PR #4354, opened 2026-01-03
Support Polymorphic json

### LINKED ISSUE #2557: [FEATURE]  LangChain4j's AiService to Support Polymorphic JSON Deserialization

raphidlee:
## Issue
Closes #2557

## Change
- Add PolymorphicOutputParser to handle Jackson @JsonTypeInfo/@JsonSubTypes/@JsonTypeName
    discriminator-based parsing.
- Wire DefaultOutputParserFactory to return the polymorphic parser when applicable.
- Update PojoCollectionOutputParser to use the factory and reuse the item parser’s schema (enabling
  anyOf for polymorphic items).
- Extend schema coverage in JsonSchemasTest and add new PolymorphicOutputParserTest.
- Add PolymorphicAiServiceTest cases for polymorphic response parsing and prompt instructions.

## General checklist


raphidlee:
@dliubarskyi
Just a gentle ping in case this got buried.
I've marked the PR as ready for review.

Happy to make any changes if needed.

langchain4j/src/main/java/dev/langchain4j/service/output/DefaultOutputParserFactory.java
```diff
@@ -57,6 +57,10 @@ public OutputParser<?> get(Class<?> rawClass, Class<?> typeArgumentClass) {
             return new EnumOutputParser<>(rawClass.asSubclass(Enum.class));
         }
 
+        if (rawClass.isAnnotationPresent(JsonTypeInfo.class)) {
```
dliubarskyi: I am wondering if this is a breaking change or not.
Is it possible that there are users that have `JsonTypeInfo`-annotated return types and currently have one behaviour, and they will observe another behaviour after the update?

langchain4j/src/main/java/dev/langchain4j/service/output/PojoCollectionOutputParser.java
```diff
@@ -1,28 +1,30 @@
...
 
     private final Class<T> type;
-    private final PojoOutputParser<T> parser;
+    private final OutputParser<T> parser;
 
     PojoCollectionOutputParser(Class<T> type) {
         this.type = ensureNotNull(type, "type");
-        this.parser = new PojoOutputParser<>(type);
+        @SuppressWarnings("unchecked")
+        OutputParser<T> outputParser = (OutputParser<T>) new DefaultOutputParserFactory().get(type, null);
```
dliubarskyi: Not sure I like this circular dependency... `DefaultOutputParserFactory` knows about parsers and parsers know about factory...

langchain4j/src/main/java/dev/langchain4j/service/output/JsonSchemas.java
```diff
@@ -1,53 +1,40 @@
...
-    private static boolean isPojo(Type returnType) {
-
-        if (returnType == String.class
-                || returnType == AiMessage.class
-                || returnType == TokenStream.class
-                || returnType == Response.class) {
-            return false;
+        OutputParser<?> outputParser = outputParserFor(returnType);
+        if (outputParser instanceof PojoOutputParser || outputParser instanceof PolymorphicOutputParser) {
+            return outputParser.jsonSchema();
```
dliubarskyi: I am not sure we tested all the use cases in unit tests, and there is a risk of this being a breaking change. Would it be possible to keep this method without changes and introduce a new (internal, private) instead?

langchain4j/src/main/java/dev/langchain4j/service/output/PolymorphicOutputParser.java
```diff
@@ -0,0 +1,200 @@
...
+                .append("' set to one of [")
+                .append(join(", ", discriminatorValuesToTypes.keySet()))
+                .append("].");
+
+        discriminatorValuesToTypes.forEach((value, subType) -> instructions
+                .append("\n- When ")
+                .append(discriminator)
+                .append("=")
+                .append(value)
+                .append(", follow: ")
```
dliubarskyi:
```suggestion
                .append(", answer strictly in the following JSON format: ")
```

langchain4j/src/test/java/dev/langchain4j/service/output/PolymorphicOutputParserTest.java
```diff
@@ -0,0 +1,174 @@
...
+    void should_treat_non_string_discriminator_as_string_value() {
+        String json = """
+                { "type": 123, "url": "https://example.com" }
+                """;
+
+        assertThatThrownBy(() -> parser.parse(json))
+                .isInstanceOf(OutputParsingException.class)
+                .hasMessageContaining("Unknown discriminator value: 123");
+    }
+}
```
dliubarskyi, line 224: Please add an integration test to `OpenAiAiServiceWithJsonSchemaIT`, similar to `OpenAiChatModelWithJsonSchemaIT.should_generate_valid_json_with_anyof`.
Return type of AI Service should be polymorphyc

dliubarskyi (commented):
@raphidlee thank you!

langchain4j/src/main/java/dev/langchain4j/service/output/DefaultOutputParserFactory.java
raphidlee: That’s a fair concern. I originally used a custom @Polymorphic annotation to avoid this exact risk, but switched to @JsonTypeInfo to stay aligned with Jackson.

Currently, returns with @JsonTypeInfo don't parse correctly because the LLM isn't instructed to provide a discriminator. So this is more of a fix for an unsupported case than a breaking change.

If we want to be extra safe, we could require both @JsonTypeInfo and a marker like @Polymorphic to trigger this logic. This would make it strictly opt-in. What do you think?

langchain4j/src/main/java/dev/langchain4j/service/output/JsonSchemas.java
raphidlee: Sure.

langchain4j/src/main/java/dev/langchain4j/service/output/PojoCollectionOutputParser.java
raphidlee: Agreed. Let me take a look and update the PR.

langchain4j/src/main/java/dev/langchain4j/service/output/DefaultOutputParserFactory.java
raphidlee: While working on this, I thought about another option: we could mitigate the risk by checking JsonTypeInfo.Id.NONE, without requiring any additional annotations.
The idea is to only use PolymorphicOutputParser if @JsonTypeInfo is present and use != Id.NONE. If use == Id.NONE, it would fall back to the original PojoOutputParser behavior.
Do you think this would be a sufficient safeguard?

**JsonTypeInfo jsonTypeInfo = rawClass.getAnnotation(JsonTypeInfo.class);
if (jsonTypeInfo != null && jsonTypeInfo.use() != JsonTypeInfo.Id.NONE) {**

raphidlee:
@dliubarskyi  
I've pushed the changes to address the feedback.
- Circular Dependency: Refactored PojoCollectionOutputParser to accept OutputParser<T> directly, breaking the circular dependency with the factory.
- Backward Compatibility: Added a check for JsonTypeInfo.Id.NONE in DefaultOutputParserFactory. If use = Id.NONE, it falls back to PojoOutputParser, preserving existing behavior. I'd appreciate your thoughts on whether this check is sufficient to address the backward compatibility concern.
- JsonSchemas: Extracted schema generation logic to jsonSchemaFrom(OutputParser). This delegates schema generation to the OutputParser itself, ensuring consistency between PojoOutputParser and PolymorphicOutputParser. The logic for PojoOutputParser remains effectively the same, just moved to the parser class.
- Tests: Added integration tests for polymorphic responses and unit tests for the Id.NONE fallback.

Please let me know if you have any further suggestions or if the Id.NONE check is sufficient for backward compatibility.

dliubarskyi:
@raphidlee apparently something went wrong when merging, this PR has 733 files changed.
Could you please check?

raphidlee:
> @raphidlee apparently something went wrong when merging, this PR has 733 files changed.
> Could you please check?

Oh.. Sorry, probably out of sync. will fix it asap.

raphidlee:
@dliubarskyi 
Fixed. 
I realized this branch was originally created from feature/2557_polymorphic_json instead of main, which caused the large diff. The PR should now show only the intended changes.

Sorry for the long review thread. Please take another look when you have time.

mariofusco (changes requested):
I believe that what I've done in this [other pull request](https://github.com/langchain4j/langchain4j/pull/4781) is complementary to what has been done here. In particular what I've done there is used for serialization of tool parameters [here](https://github.com/langchain4j/langchain4j/blob/d8d17d7fee5c6708e02e7c01eb8700c4b43d1bb8/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java#L200).

Please consider merging that work in this pull request.

@dliubarskyi alternatively we could merge the 2 PRs independently. Please let me know if I should close the other one.

langchain4j/src/test/java/dev/langchain4j/service/output/PolymorphicOutputParserTest.java
dliubarskyi: These new tests are failing:

```
java.lang.IllegalArgumentException: For OpenAI, the root element of the JSON Schema must be either a JsonObjectSchema or a JsonRawSchema, but it was: class dev.langchain4j.model.chat.request.json.JsonAnyOfSchema

	at dev.langchain4j.model.openai.internal.OpenAiUtils.toOpenAiResponseFormat(OpenAiUtils.java:462)
	at dev.langchain4j.model.openai.internal.OpenAiUtils.toOpenAiChatRequest(OpenAiUtils.java:538)
	at dev.langchain4j.model.openai.OpenAiChatModel.doChat(OpenAiChatModel.java:149)
	at dev.langchain4j.guardrail.SynchronousChatExecutor.execute(SynchronousChatExecutor.java:32)
	at dev.langchain4j.guardrail.AbstractChatExecutor.executeInternal(AbstractChatExecutor.java:66)
	at dev.langchain4j.guardrail.AbstractChatExecutor.execute(AbstractChatExecutor.java:52)
	at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:306)
```

I guess to support polymorphic return type we will have to do one level nesting, similar to what we do with enums. So the root should be a `JsonObjectSchema` with a single property (e.g., `value`) of type `JsonAnyOfSchema`. WDYT?

raphidlee: Thanks for the suggestion! I'm already working on it — just a bit delayed due to my day job. Bear with me a little longer.

langchain4j/src/test/java/dev/langchain4j/service/output/JsonSchemasTest.java
```diff
@@ -40,10 +56,24 @@ void should_return_empty_for_not_pojos() {
         assertThat(jsonSchemaFrom(Response.class)).isEmpty();
         assertThat(jsonSchemaFrom(Integer.class)).isEmpty();
         assertThat(jsonSchemaFrom(LocalDate.class)).isEmpty();
-        assertThat(jsonSchemaFrom(new TypeReference<List<Pojo>>() {
-        }.getType())).isEmpty();
+        assertThat(jsonSchemaFrom(new TypeReference<List<Pojo>>() {}.getType())).isEmpty();
     }
 
+    @Test
+    void should_return_json_schema_for_polymorphic_sealed_types() {
```
dliubarskyi: name of the test includes "sealed", but it does not seem correct. BTW, could you please remond we why sealed types are not supported out of the box? In my understnading, if interface is sealed, we don't have to use `JsonTypeInfo` and other annotations, right? Jackson should be able to figure it out automatically, right?

raphidlee: The "sealed" in the test name seems to be a leftover from earlier iterations — fixed that. As for @JsonTypeInfo, we still need it at least for the discriminator, so it can't be dropped just yet.

raphidlee:
@dliubarskyi 
Pushed all the latest changes. 
Will check if there are any updates related to JsonAnyOfSchema from mariofusco's change request and merge them if needed.

langchain4j/src/test/java/dev/langchain4j/service/output/JsonSchemasTest.java
dliubarskyi: > As for @JsonTypeInfo, we still need it at least for the discriminator

Can't we automatically take a class name as a discriminator (if not specified otherwise)?

raphidlee: Good point — the existing code already defaults subtype names to simple class names, so only a small tweak should do it. Will work on it now.

raphidlee: Pushed, Now, we handle sealed interface without JsonTypeInfo.

dliubarskyi:
Hi @raphidlee, thanks a lot for putting time into this, the PR was a useful starting point for thinking through the feature. After going through it carefully I decided to take a different direction on the design, and since the changes I had in mind would mean rewriting most of the PR inside the review, I went ahead and opened https://github.com/langchain4j/langchain4j/pull/5060 as a fresh implementation.
Going to close this one in favor of that, hope you understand. And thanks again! 🙏

raphidlee:
> Hi @raphidlee, thanks a lot for putting time into this, the PR was a useful starting point for thinking through the feature. After going through it carefully I decided to take a different direction on the design, and since the changes I had in mind would mean rewriting most of the PR inside the review, I went ahead and opened #5060 as a fresh implementation. Going to close this one in favor of that, hope you understand. And thanks again! 🙏

Thanks a lot. Sorry I couldn't be of more help. I'll try to work on another (simpler) issue. 😁

PR #4368, opened 2026-01-05
Allow to listen for tool executions in both AiServices and agents

### LINKED ISSUE #4177: [FEATURE] beforeToolExecution() and onToolExecuted() with Agent
Many uses cases involving an Agent do not support (or are not designed for) streaming (see #3519). Furthermore, the AgenticServices.agentBuilder() currently lacks the beforeToolExecution() and onToolExecuted() methods. These methods are still quite useful even if an agent does not use streaming. I would need it for example, to notify the user which tools are executed.

Add the two methods to the builder interface

Currently this functionality is provided by the assistant.stream(messages) interface, which is quite unnatural for Agents where I do expect to call the method I defined in my agent interface. IMHO, moving/adding those methods on the builder level would make the pattern more generic and natural (even for AiServices)

mariofusco:
## Issue

Closes #4177

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
@@ -38,10 +38,11 @@ import dev.langchain4j.rag.RetrievalAugmentor;
 import dev.langchain4j.rag.content.Content;
 import dev.langchain4j.rag.content.retriever.ContentRetriever;
 import dev.langchain4j.rag.content.retriever.EmbeddingStoreContentRetriever;
+import dev.langchain4j.service.tool.BeforeToolExecution;
 import dev.langchain4j.service.tool.DefaultToolExecutor;
 import dev.langchain4j.service.tool.ToolArgumentsErrorHandler;
+import dev.langchain4j.service.tool.ToolExecution;
 import dev.langchain4j.service.tool.ToolExecutionErrorHandler;
-import dev.langchain4j.service.tool.ToolExecutionResult;
 import dev.langchain4j.service.tool.ToolExecutor;
 import dev.langchain4j.service.tool.ToolProvider;
 import dev.langchain4j.spi.services.AiServicesFactory;
@@ -53,7 +54,6 @@ import java.util.Set;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
-import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -473,11 +473,11 @@ public abstract class AiServices<T> {
     }
 
     /**
-     * Configures the maximum number of tool invocations that can be executed.
+     * Configures the maximum number of times LLM responds with tool call(s).
      * <p>
      * By default, this limit is set to 100.
      *
-     * @param maxSequentialToolsInvocations The maximum number of tool invocations that can be executed.
+     * @param maxSequentialToolsInvocations The maximum number of times LLM responds with tool call(s).
      * @return builder
      */
     public AiServices<T> maxSequentialToolsInvocations(int maxSequentialToolsInvocations) {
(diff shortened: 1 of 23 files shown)


langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -469,11 +472,46 @@ public AiServices<T> executeToolsConcurrently(Executor executor) {
         return this;
     }
 
+    /**
+     * Configures the maximum number of tool invocations that can be executed.
```
dliubarskyi: Unfortunately the name of the method is misleading, so this description is incorrect. It is not the number of tool calls, but the number of times LLM responds with tool call(s) (think LLM request/response cycles)

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -469,11 +472,46 @@ public AiServices<T> executeToolsConcurrently(Executor executor) {
...
 
+    /**
+     * Configures a callback to be invoked before each tool execution.
+     *
+     * @param beforeToolExecution A {@link Consumer} that accepts a {@link ToolExecutionRequest}
+     *                            representing the tool execution request about to be executed.
+     * @return builder
+     * @since 1.11.0
+     */
+    public AiServices<T> beforeToolExecution(Consumer<ToolExecutionRequest> beforeToolExecution) {
```
```suggestion
    public AiServices<T> beforeToolExecution(Consumer<BeforeToolExecution> beforeToolExecution) {
```
To be in sync with `TokenStream`. It will also be easier to add more fields later

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -469,11 +472,46 @@ public AiServices<T> executeToolsConcurrently(Executor executor) {
...
+    /**
+     * Configures a callback to be invoked after each tool execution.
+     *
+     * @param afterToolExecution A {@link BiConsumer} that accepts a {@link ToolExecutionRequest}
+     *                           representing the tool execution request that was executed,
+     *                           and a {@link ToolExecutionResult} representing the result of the tool execution.
+     * @return builder
+     * @since 1.11.0
+     */
+    public AiServices<T> afterToolExecution(BiConsumer<ToolExecutionRequest, ToolExecutionResult> afterToolExecution) {
```
```suggestion
    public AiServices<T> afterToolExecution(Consumer<ToolExecution> afterToolExecution) {
```
To be in sync with `TokenStream`. It will also be easier to add more fields later

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
```diff
@@ -150,10 +152,15 @@ void should_execute_a_tool_then_answer(ChatModel chatModel) {
...
         ChatModel spyChatModel = spy(chatModel);
 
+        List<String> toolCalls = new ArrayList<>();
+        Map<String, Object> toolResults = new HashMap<>();
+
         Assistant assistant = AiServices.builder(Assistant.class)
                 .chatModel(spyChatModel)
                 .chatMemory(chatMemory)
                 .tools(transactionService)
+                .beforeToolExecution(request -> toolCalls.add(request.name()))
```
dliubarskyi: Please add a similar in `StreamingAiServicesWithToolsIT`, it seems that for streaming case new callbacks set when building AI Service are not called, only those that are configured on `TokenStream` are called.

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
mariofusco: I reworded the Javadoc, please check it again.

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
mariofusco: Done.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
mariofusco: Done.

dliubarskyi (approved):
@mariofusco thank you!

PR #4337, opened 2026-01-01
[BUG] 4250 - set media resolution in gemini generate content request

partarstu:
## Issue
Closes #4250

## Change
Added media_resolution field support to the Gemini integration to control how the API processes media inputs (images, videos, PDFs).

### New Files
- GeminiMediaResolutionLevel - New enum representing media resolution levels:
  - MEDIA_RESOLUTION_UNSPECIFIED - Default setting
  - MEDIA_RESOLUTION_LOW - Faster processing, less detail
  - MEDIA_RESOLUTION_MEDIUM - Balanced
  - MEDIA_RESOLUTION_HIGH - More detail
  - MEDIA_RESOLUTION_ULTRA_HIGH - Highest detail (per-part only, for computer use)

### Modified Files
- GeminiGenerationConfig - Added mediaResolution field for global media resolution setting (all multimodal models)
- GeminiContent - Added mediaResolution field and GeminiMediaResolution wrapper record to GeminiPart for per-part resolution (Gemini 3 only)
- PartsAndContentsMapper - Updated fromContentToGPart() to map ImageContent.DetailLevel to GeminiMediaResolutionLevel:
  - LOW → MEDIA_RESOLUTION_LOW
  - HIGH → MEDIA_RESOLUTION_HIGH
  - AUTO → MEDIA_RESOLUTION_UNSPECIFIED

### Tests Added
- GeminiGenerationConfigTest - Added tests for all media resolution levels (HIGH, LOW, MEDIUM, ULTRA_HIGH, MEDIA_RESOLUTION_UNSPECIFIED)
- PartsAndContentsMapperTest - Added tests for DetailLevel to GeminiMediaResolutionLevel mapping

## General checklist
- [X] There are no breaking changes (API, behaviour)
- [X] I have added unit and/or integration tests for my change
- [X] The tests cover both positive and negative cases
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the core and main modules, and they are all green
- [ ] I have added/updated the documentation
- [ ] I have added an example in the examples repo (only for "big" features)
- [ ] I have added/updated Spring Boot starter(s) (if applicable)

## Checklist for adding new maven module
N/A - No new module added

## Checklist for adding new embedding store integration
N/A - Not an embedding store integration

## Checklist for changing existing embedding store integration
N/A - Not an embedding store change

PR #4389, opened 2026-01-08
[BUG] 4250 - integration tests fix

partarstu:
## Issue
Fixes IT failures caused by commit ac8ad4c - per-part media resolution is not supported by older Gemini models

## Change
This PR rolls back the per-part media resolution feature introduced in #4337 (commit ac8ad4c).

The issue was that the mediaResolution field was being included in API requests for all models, but this feature is only supported by Gemini 3.x models and gemini-2.5-computer-use models. Older models like gemini-2.0-flash-lite reject requests with this field, causing the error: "Per part media resolution is not supported for this Model."

### Changes made:
- Rolled back per-part media resolution feature
- Added/updated tests for model version detection and media resolution inclusion/exclusion behavior.

## General checklist


partarstu:
This fix is not quite clean, but creating a better fix would mean refactoring the ImageContent class thus impacting the whole project - there is currently no way to identify if the user has explicitly set the image detail level, or not (default ImageContent constructor sets the DetailLevel to LOW and DetailLevel enum has no "default" value).

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/PartsAndContentsMapper.java
```diff
@@ -47,24 +47,32 @@
...
 
 final class PartsAndContentsMapper {
+
     private PartsAndContentsMapper() {}
 
     static final String THINKING_SIGNATURE_KEY =
             "thinking_signature"; // do not change, will break backward compatibility!
     static final String GENERATED_IMAGES_KEY =
             "generated_images"; // key for storing generated images in AiMessage attributes
+    static final String GEMINI_MODEL_SUPPORTING_PER_PART_MEDIA_RESOLUTION_NAME_REGEX =
```
dliubarskyi: This is very unreliable... Not only this will need to be updated with every model release, but this might break custom model names (e.g., finetunes, etc.). Can we do without it? Let the user decide if this media resolution field should be included in the request. For all existing users of this module, nothing should change unless they explicitly configure something.

dliubarskyi (commented):
@partarstu thank you!

Please make sure your recent PRs related to this feature do not break/change anything for the existing users. If media resolution should be sent to the API, it should be explicitly enabled somehow, e.g., by some property in the builder.

partarstu:
@dliubarskyi , I've mentioned in my comment that it's not clean, but because of the ImageContent class limitations (setting by default "LOW" and no option for the default value) it is impossible to differentiate between the "explicitly enabled" and default "LOW" level and I haven't found a cleaner solution. 
I can see three solution options:
1. Change the constructor of ImageContent to have DetailLevel set to "AUTO" (identifying that this is a default setting and user hasn't set this level explicitly, i.e. no per-part media resolution will be set), but it might impact the behavior of all supported models.
2. Introduce a Gemini-specific image content class implementing "Content" interface. This class will have all 5 supported media resolutions. Then add the info to the docs that if the user wants Gemini-specific image content config, this class could be used instead of ImageContent.
3. Refuse the support of per-part media resolution for Gemini models. In this case only generation config will allow to set explicitly the media resolution for all parts.

Which one do you prefer (or hopefully you have a better one) ?

dliubarskyi:
> Refuse the support of per-part media resolution for Gemini models. In this case only generation config will allow to set explicitly the media resolution for all parts.

This seems like the most reasonable option right now. If per-part resolution will be required in the future, we can always add it later.

Thanks a lot!

partarstu:
@dliubarskyi , per-part changes reverted. One thing to keep in mind - MEDIA_RESOLUTION_ULTRA_HIGH is only supported by Gemini API per-part for now, i.e. it will not be usable in LangChain4j after these changes. This resolution is important for computer use (and similar) agents.

dliubarskyi:
@partarstu if it is important to set resolution per-part, perhaps let's add a new property that should be explicitly enabled to take `DetailLevel` into consideration per-part? So only if this property is enabled (which will ensure existing users are not affected), `DetailLevel` will be mapped and applied per-part. Would that work?

partarstu:
> @partarstu if it is important to set resolution per-part, perhaps let's add a new property that should be explicitly enabled to take `DetailLevel` into consideration per-part? So only if this property is enabled (which will ensure existing users are not affected), `DetailLevel` will be mapped and applied per-part. Would that work?

@dliubarskyi , do you mean extending the `GeminiGenerationConfig `with a new boolean e.g. `mediaResolutionPerPartEnabled `? Yes, it could be a solid solution. 

But I've made an error in my previous comment (already edited it) - I meant not high resolution, but `MEDIA_RESOLUTION_ULTRA_HIGH `. High and low resolutions can be set up per whole message, which already works and nothing needs to be changed in the scope of per-part. The issue with `MEDIA_RESOLUTION_ULTRA_HIGH `is that this one has currently no mapping from `DetailLevel`, so even if I implement additional flag for enabling the per-part media resolution, I still have no idea how to set `MEDIA_RESOLUTION_ULTRA_HIGH` based on the current `DetailLevel`. If you have any idea how to do it in a clean way, I would then implement that flag and this idea, otherwise I don't think it makes sense implementing per-part media resolution if `MEDIA_RESOLUTION_ULTRA_HIGH` will not be supported.

Sorry for the confusion.

dliubarskyi:
@partarstu I guess it is time to extend `DetailLevel` enum with `MEDIUM` and `ULTRA_HIGH` options, WDYT?

partarstu:
@dliubarskyi - it would be the cleanest way, but the impact scope is huge in this case (`DetailLevel `is used by all models). I could take a look into it as soon as I get some spare time. 

This scope also allows to take care of setting the default level to "`AUTO`" instead of "`LOW`" thus allowing to identify if the user actually set the level manually or not - are you OK with this change as well ? Or do you want to keep this part as it is ?

dliubarskyi:
> but the impact scope is huge in this case (DetailLevel is used by all models)

As far as I know, it is used only by OpenAI, Gemini and WatsonX, which we can easily adapt. It might be a breaking change if someone has a custom `ChatModel` implementation outside of this repo that uses switch blocks. But the probability of this is low, and even if it is breaking, it is very easily fixable.

> This scope also allows to take care of setting the default level to "AUTO" instead of "LOW" thus allowing to identify if the user actually set the level manually or not - are you OK with this change as well ? Or do you want to keep this part as it is ?

Changing the default would be a big (but silent) change in behaviour for users, so I would not do that in 1.X. Maybe we should consider this for 2.0 version. Could you please explain why is it required to know if user has set it or not?

partarstu:
> Changing the default would be a big (but silent) change in behavior for users, so I would not do that in 1.X. Maybe we should consider this for 2.0 version. Could you please explain why is it required to know if user has set it or not?

That's actually one of the topics we've discussed during per-part mapping of `DetailLevel `to the media resolution for Gemini. You wanted not to impact the users which haven't set the detail level explicitly, so that for them everything works as it used to. In order to know if the user set the `DetailLevel `manually or not we need a default value for this enum. Setting by default to `LOW` breaks the logic because it's not a default value, `AUTO` is. In the case of Gemini there was an additional problem because of per-part support only by Gemini 3 models, but even if this problem wasn't there, current default setting of `DetailLevel ` to `LOW ` doesn't allow to separate those users which do not explicitly set this value and users which do. 

Leaving it as is for now is also a viable option in my opinion. In this case I could implement only automatic mapping of `ULTRA_HIGH ` for Gemini per-part, leaving `LOW, MEDIUM and HIGH` not mapped until there's a final architectural decision about how to handle this on the per-part level.

partarstu:
@dliubarskyi , so what should I do in this case ?

dliubarskyi:
> You wanted not to impact the users which haven't set the detail level explicitly, so that for them everything works as it used to.

Yes, as well as those who have set it explicitly, nothing should change for them.

> In order to know if the user set the DetailLevel manually or not we need a default value for this enum.

Perhaps there is another (ugly) way to know this - introduce a flag `boolean detailLevelSet` or somehting like this

> Setting by default to LOW breaks the logic because it's not a default value, AUTO is

Yes, I guess it was a bad decision to set the default value to `LOW`. But it does not break anything. It just sets the detail level to low by default. If it is not enough for the user, they can set it higher.


> In the case of Gemini there was an additional problem because of per-part support only by Gemini 3 models, but even if this problem wasn't there, current default setting of DetailLevel  to LOW  doesn't allow to separate those users which do not explicitly set this value and users which do.

I am not sure I understand what is the problem here. The default value in Gemini API differs from the default value in LC4j API, but it still works. So if we were to introduce `DetailLevel.ULTRA_HIGH` and map it to Gemini's `ULTRA_HIGH`, it should work, no?

dliubarskyi:
> I am not sure I understand what is the problem here.

Perhaps a concrete example/scenario would clarify this

dliubarskyi (approved):
@partarstu thank you! I will merge this PR now to fix failing tests. Let's continue the discussion, and if we agree on per-part detail level approach, let's have a separate PR

partarstu:
> I am not sure I understand what is the problem here. The default value in Gemini API differs from the default value in LC4j API, but it still works. So if we were to introduce `DetailLevel.ULTRA_HIGH` and map it to Gemini's `ULTRA_HIGH`, it should work, no?

I simply meant that introducing per-part mapping without knowing if the user explicitly set the DetailLevel or if it's a default value (which is now impossible because of setting it by default to LOW, not AUTO), will not adhere to the principle "not to impact the users which haven't set the detail level explicitly". It means without any additional workaround like you've mentioned (introduce a flag boolean detailLevelSet or somehting like this) it can't be done. However, the approach you've suggested to introduce a new boolean like mediaResolutionPerPartEnabled inside GeminiGenerationConfig is a good solution to not impact existing usages of DetailLevel, no matter if the user set it explicitly or not.

partarstu:
> Let's continue the discussion, and if we agree on per-part detail level approach, let's have a separate PR

@dliubarskyi , created here: https://github.com/langchain4j/langchain4j/discussions/4432

dliubarskyi:
I guess I understand now. I wrongly assumed that `DetailLevel` mapping was already present in Gemini module, but it is not the case. So all existing users that send images, they all are processed in `MEDIA_RESOLUTION_UNSPECIFIED` on the Gemini side, because we do not specify any setting. Correct?

So if we were to introduce per-part `DetailLevel` mapping, it would break the behaviour for them because images would start processing in `MEDIA_RESOLUTION_LOW` instead of `MEDIA_RESOLUTION_UNSPECIFIED`.

Then let's go with `mediaResolutionPerPartEnabled` solution?

partarstu:
> So all existing users that send images, they all are processed in `MEDIA_RESOLUTION_UNSPECIFIED` on the Gemini side, because we do not specify any setting. Correct?

@dliubarskyi - yes, correct. And on the level of `GeminiGenerationConfig `the media resolution is currently not set by default, it's null - the user must set it explicitly.

> So if we were to introduce per-part `DetailLevel` mapping, it would break the behaviour for them because images would start processing in `MEDIA_RESOLUTION_LOW` instead of `MEDIA_RESOLUTION_UNSPECIFIED`.

Again correct. `MEDIA_RESOLUTION_UNSPECIFIED` has definitely other than `MEDIA_RESOLUTION_LOW `behavior. 


> Then let's go with mediaResolutionPerPartEnabled solution?

Yes - if you're ok. I could implement it and also extend DetailLevel, as suggested by you, to contain `MEDIUM `and `ULTRA_HIGH `levels.

PR #4392, opened 2026-01-08
Make (some) Anthropic classes testable, add tests for DefaultAnthropicClient and add docs

Bram--:
## Issue
Progresses #3916

## Change
This makes some of the classes testable and adds a unit test for DefaultAnthropicClient. There are no functional changes or changes to the public API. this PR is in preparation for the Batch impl. for Claude (https://platform.claude.com/docs/en/build-with-claude/batch-processing).

Finally, I addressed some minor issues in the client (such as using `.collecto(Collectors:toList)` instead of `.toList()`).

## General checklist
- [x] There are no breaking changes (API, behaviour)
- [x] I have added unit and/or integration tests for my change
- [x] The tests cover both positive and negative cases
- [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [x] I have manually run all the unit and integration tests in the [core]

PR #4394, opened 2026-01-08
Adds AnthropicBatchChatModel

Bram--:
## Issue
Progresses #3916
Depends on PR #4392 

## Summary

Added `AnthropicBatchChatModel` for asynchronous batch processing of chat requests using the Anthropic Message Batches API.

## APIs

| Method | Description |
|--------|-------------|
| `createBatchInline` | Creates a new batch job with the provided chat requests. Returns an `AnthropicBatchIncomplete` with the batch ID and initial "in_progress" status. [Documentation](https://platform.claude.com/docs/en/build-with-claude/batch-processing) |
| `retrieveBatchResults` | Retrieves the current status and results of a batch job. Returns `AnthropicBatchIncomplete` (still processing), `AnthropicBatchSuccess` (completed with results), or `AnthropicBatchError` (all requests failed). |
| `cancelBatchJob` | Cancels a batch job that is currently processing. The batch transitions to "canceling" and eventually "ended" with partial results. Delete is not supported |
| `listBatchJobs` | Lists all batch jobs in the workspace with optional pagination support (`pageSize`, `pageToken`). |

## Features

- **50% cost reduction** on all token usage compared to synchronous API
- Process up to **100,000 requests** or **256 MB** per batch
- Support for all Messages API features (vision, tools, system messages, thinking, etc.)
- Results available for **29 days** after batch creation
- Automatic custom ID generation (UUID) for request tracking

## Changes

- `AnthropicBatchChatModel` - New main API class marked as `@Experimental`
- `AbstractAnthropicChatModelBuilder<T>` - Extracted shared builder configuration from `AnthropicChatModel` and `AnthropicBatchChatModel` using generic `self()` pattern
- `AnthropicBatchName` - Value object for batch IDs with validation (must start with `msgbatch_`)
- `AnthropicBatchResponse<T>` - Sealed interface hierarchy for batch states: `AnthropicBatchIncomplete`, `AnthropicBatchSuccess`, `AnthropicBatchError`
- `AnthropicBatchIndividualResult<T>` - Individual result wrapper with status (succeeded, errored, canceled, expired)
- `AnthropicBatchList<T>` - Paginated list of batches
- Extended `AnthropicClient` and `DefaultAnthropicClient` with batch operations
- Added unit tests for `AnthropicBatchChatModel`
- Added integration tests for batch creation, retrieval, cancellation, and listing

## Limitations

- No streaming support for batch requests
- Batches may take up to 24 hours to process (most complete within 1 hour)
- Results may not be returned in the same order as requests
- Batch deletion is not supported by the Anthropic API

## Considerations

- The Anthropic API slightly differs from the Gemini API unfortunately (Making it harder to generalize). Where Anthropic's API has 2 different endpoints for polling and retrieving Gemini has this as a single endpoint and will always return the full batch on completion (opposed to an URL with the results). Anthropic also does not support deleting batches, but only cancelling them.
 - Although I've written Integration tests and wrote a manual test file (example output below) I have yet to test it with different requests e.g. from their documentation:
 ```
Any request that you can make to the Messages API can be included in a batch. This includes:

 * Vision
 * Tool use
 * System messages
 * Multi-turn conversations
 * Any beta features
```

## Example output:

```
Creating batch with 3 requests...
Batch created: msgbatch_017u3VBvjc6EjSfVi32NZNVJ
Initial status: in_progress
Processing count: 3

Polling for completion...
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0
Status: in_progress | Processing: 3 | Succeeded: 0 | Errored: 0 | Canceled: 0 | Expired: 0

Batch processing completed!
Batch succeeded!
Total: Succeeded=3, Errored=0, Canceled=0, Expired=0
Created at: 2026-01-08T22:27:12.021830+00:00
Ended at: 2026-01-08T22:29:34.910293+00:00

--- Results ---

Custom ID: e7ae50b9-a6f9-48f7-b555-3354135a5ece
Result type: succeeded
Response: Paris is the capital of France.
Token usage: AnthropicTokenUsage { inputTokenCount = 19, outputTokenCount = 10, totalTokenCount = 29, cacheCreationInputTokens = 0, cacheReadInputTokens = 0 }
Finish reason: STOP

Custom ID: a3688303-51bd-4cd1-8472-fec669c32174
Result type: succeeded
Response: Berlin is the capital of Germany.
Token usage: AnthropicTokenUsage { inputTokenCount = 19, outputTokenCount = 10, totalTokenCount = 29, cacheCreationInputTokens = 0, cacheReadInputTokens = 0 }
Finish reason: STOP

Custom ID: bed29ff9-94ec-49a9-836e-f1a77c94da39
Result type: succeeded
Response: Rome is the capital of Italy.
Token usage: AnthropicTokenUsage { inputTokenCount = 19, outputTokenCount = 10, totalTokenCount = 29, cacheCreationInputTokens = 0, cacheReadInputTokens = 0 }
Finish reason: STOP

--- Recent Batches ---
Batch: msgbatch_017u3VBvjc6EjSfVi32NZNVJ
Batch: msgbatch_01Y6QzJVEsVzgWbr4yRhuegS
Batch: msgbatch_01S2gSVUJiTEoXbSH67rk35b
Batch: msgbatch_011fXAzidUDRBU1azvqtyELq
Batch: msgbatch_01FKRdwdPt6uZgbHLGiaUD9x
```

## General checklist

### CODE REVIEW

File: langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/client/AnthropicClient.java [hunk 1]
```diff
@@ -42,6 +49,53 @@ public AnthropicModelsListResponse listModels() {
         throw new UnsupportedOperationException("Model listing is not supported by this client implementation");
     }
 
+    /**
+     * Creates a new Message Batch for asynchronous processing.
+     *
+     * @param request the batch creation request containing the list of message requests
+     * @return the batch response with ID and initial status
+     */
+    public abstract AnthropicBatchResponse createBatch(AnthropicCreateBatchRequest request);
```
dliubarskyi, line 58: Please make these methods `default` (you can just `throw new UnsupportedFeatureException("Not supported yet.");`), otherwise it will break existijng implementations

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicBatchResponse.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.List;
+import org.jspecify.annotations.NonNull;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Response from a batch operation, which can be in progress, completed successfully, or completed with errors.
+ *
+ * @param <T> the type of individual results
+ */
+public sealed interface AnthropicBatchResponse<T>
```
dliubarskyi, line 14: Are we sure Anthropic won't add one more type in the future? This will require us to add one more class and it will break code of existing users.

dliubarskyi: Why do we need 3 states anyway?

If I understand the docu [correctly](https://platform.claude.com/docs/en/build-with-claude/batch-processing), submitting a batch is always async? So in case of success, `AnthropicBatchIncomplete` is always returned from `createBatchInline`. If error happens, we can throw an exception, the same way we do with other APIs.

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicBatchChatModel.java
```diff
@@ -0,0 +1,405 @@
...
+
+    /**
+     * Creates a new Message Batch with the provided chat requests.
+     *
+     * <p>Each request is assigned a unique custom ID (UUID) for tracking results.</p>
+     *
+     * @param requests the list of chat requests to process
+     * @return the batch response containing the batch ID and initial status
+     */
+    public dev.langchain4j.model.anthropic.AnthropicBatchResponse<ChatResponse> createBatchInline(
```
dliubarskyi, line 193: What does "inline" means here? It seems that `submitBatch` is a more appropriate name, WDYT?

dliubarskyi (commented):
@Bram-- thank you! I did not finish the review yet, but I have left a few comments

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicBatchChatModel.java
```diff
@@ -0,0 +1,405 @@
...
+ *             System.out.println(result.result().aiMessage().text());
+ *         }
+ *     }
+ * }
+ * }</pre>
+ *
+ * @see <a href="https://docs.anthropic.com/en/docs/build-with-claude/batch-processing">Anthropic Batch Processing</a>
+ */
+@Experimental
+public class AnthropicBatchChatModel {
```
dliubarskyi, line 98: Given that we already have `GoogleAiGeminiBatchChatModel`, does it make sense to introduce a new generic `BatchChatModel` API in core?

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicBatchResponse.java
Bram--: You're right that Anthropic could theoretically add more states in the future. However, the `sealed interface` approach actually provides controlled extensibility, you can add new permitted subtypes in future versions while maintaining backward compatibility for users who handle the existing types via `instanceof` checks.

The three states serve distinct purposes:

1. **`AnthropicBatchIncomplete`** - Batch is still processing (polling should continue)
2. **`AnthropicBatchSuccess`** - Batch completed with results available (including mixed success/failure within the batch for the requests)
3. **`AnthropicBatchError`** - Batch-level failure (e.g., all requests failed)


`createBatchInline` is always asynchronous and will return `AnthropicBatchIncomplete` on success. For immediate failures during batch creation (e.g., invalid API key, malformed request, network errors), it should absolutely throw exceptions .

However, the three-state response model handles the asynchronous lifecycle after successful submission:

```java
// Batch creation failure → Exception (exceptional condition)
// Batch submitted successfully → AnthropicBatchIncomplete (normal flow)
// Batch completed with results → AnthropicBatchSuccess (normal flow)
// Batch ended with all failures → AnthropicBatchError (normal flow - not exceptional)
```


*Objects over exceptions*

To prevent 2 tiers of exception handling (batches and requests) both Batch and within a batch, **individual request failures are data, not exceptions**:

```java
// Individual results can be: succeeded, errored, canceled, or expired
record AnthropicBatchIndividualResult<T>(
    String customId,
    @Nullable T result,      // Present if succeeded
    @Nullable String error,  // Present if failed
    String resultType        // "succeeded", "errored", "canceled", "expired"
)
```

Mapping all requests to objects rather than throwing exceptions provides several benefits:

1. **Avoids two levels of exception handling** - Users don't need to handle both "batch failed" exceptions AND "request within batch failed" exceptions
2. **Enables complete result processing** - All results (success and failure) can be stored, logged, and processed uniformly
3. **Supports retry strategies** - Failed `customId`s can be collected and resubmitted in a new batch
4. **Maintains failure atomicity** - The batch operation itself succeeds; individual failures are reported as data

Throwing checked exceptions for each failed request would make the API unusable with streams and force awkward error handling:

```java
// Not great: Two levels of exceptions
try {
    var batch = model.retrieveBatchResults(name);
    for (var result : batch.results()) {
        try {
            process(result.getResultOrThrow()); // Throws on individual failure
        } catch (BatchRequestFailedException e) {
            // Handle individual failure
        }
    }
} catch (BatchFailedException e) {
    // Handle batch failure
}

// Better: Uniform data processing
var batch = model.retrieveBatchResults(name);
if (batch instanceof AnthropicBatchSuccess<ChatResponse> success) {
    for (var result : success.results()) {
        if (result.isSucceeded()) {
            process(result.result());
        } else {
            logFailure(result.customId(), result.error());
        }
    }
}
```

This approach treats the batch system as a pipeline where all outcomes flow through as processable objects - IMHO exactly what you'd want for robust batch processing at scale. WDYT?




*Google AI Gemini Batch API*

I've tried to keep this API as close to the existing `GoogleAiGeminiBatchChatModel` API:

| Gemini | Anthropic |
|--------|-----------|
| `BatchIncomplete` | `AnthropicBatchIncomplete` |
| `BatchSuccess` | `AnthropicBatchSuccess` |
| `BatchError` | `AnthropicBatchError` |
| `BatchName` | `AnthropicBatchName` |
| `createBatchInline()` | `createBatchInline()` |
| `retrieveBatchResults()` | `retrieveBatchResults()` |
| `cancelBatchJob()` | `cancelBatchJob()` |
| `listBatchJobs()` | `listBatchJobs()` |

This consistency will make it significantly easier to create a unified `BatchChatModel` API in `langchain4j-core`, as both implementations follow the same pattern(s).

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/AnthropicBatchChatModel.java
Bram--: See comment below.

subhashpolisetti:
Hi @Bram--, thanks for kicking off batch support here, both the #3916 issue and
the Gemini `Batch*` models. Your Gemini implementation was the pattern I followed
to add `MistralAiBatchChatModel` (#5750, merged) and `BedrockBatchChatModel`
(#5797), so batch has been landing across providers on the groundwork you laid.

Anthropic is the remaining gap. Since this PR opened, the generic `BatchChatModel`
interface landed in `langchain4j-core` and the Gemini and Mistral batch models
both implement it now. That means the provider-specific shape here (the
`AnthropicBatchResponse` sealed hierarchy and `createBatchInline` /
`retrieveBatchResults`) would be rewritten onto the core interface anyway, which
is the direction @dliubarskyi pointed to in the review above.

I opened #5875  implementing `AnthropicBatchChatModel` against that interface,
matching the Gemini and Mistral models (submit/retrieve/cancel/list, `custom_id`
re-ordering, per-request `BatchItemResult`s), with unit tests and a live
round-trip against the Batches API. It builds on your work here and closes out
the Anthropic side of #3916, so the batch set is complete.

@dliubarskyi, happy to make any changes you'd suggest in the review. Thank you.

Bram--:
Seems like this has been done/ Closing

PR #4401, opened 2026-01-09
Feat: Use streaming I/O for JSON serialization in InMemoryEmbeddingStore

### LINKED ISSUE #4393: Stream-based persistence for in-memory embedding store
(this is a follow-up of #4235)

We use in-memory embedding stores (`InMemoryEmbeddingStore`) and lazily populate them with data. Unfortunately. serializing and de-serializing is file-based and uses JSON strings. This does not scale for even medium number of embeddings as it keeps it all in memory

File-based persistence does not work for us as we have a stream-based abstraction layer which supports different backend implementations.

**Additional context**
n/a

wangxuanni:
## Issue

Fixes #4393

## Change

Use streaming (`InputStream` / `OutputStream`) for file-based JSON serialization in `InMemoryEmbeddingStore` to avoid loading the entire JSON document into memory.

The codec interface now provides default `InputStream` / `OutputStream` methods for backward compatibility.

Existing test `InMemoryEmbeddingStoreTest#should_serialize_to_and_deserialize_from_file` covers this change.

From my local testing, streaming starts to pay off around 10k embeddings (about ~67% lower peak memory with only a small time cost). Beyond that point, memory stays roughly flat thanks to fixed-size buffering, unlike serializeToJson() which keeps growing with the data.

## General checklist

- [ ] I have manually run all the unit and integration tests in the [[core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core)](https://github.com/langchain4j/langchain4j/tree/[main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j)/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green
  
- [ ] I have added/updated the [[documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [[examples repo](https://github.com/langchain4j/langchain4j-examples)](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [[Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
+++ b/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
@@ -23,7 +23,11 @@ import dev.langchain4j.store.embedding.EmbeddingSearchResult;
 import dev.langchain4j.store.embedding.EmbeddingStore;
 import dev.langchain4j.store.embedding.RelevanceScore;
 import dev.langchain4j.store.embedding.filter.Filter;
+import java.io.BufferedInputStream;
+import java.io.BufferedOutputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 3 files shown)


dliubarskyi (approved):
@wangxuanni thank you!

Issue #3306, opened 2025-07-05
[HELP]  Modify the access property of the class to public


twosoy:
I am using langchain4j to operate on a private large model, which is based on the OpenAI interface but has added some access header properties, and the request body has also been extended with properties. 

Currently, I want to implement extensions based on OpenAiChatModel. When calling the local large model, it is required to pass the appkey in the request, and the appkey expires every 15 minutes. Therefore, the customHeaders property needs to be dynamically modified, but this property is private and cannot be modified; And the defaultRequestParameters property is also private, making it impossible to add personalized properties. I am currently implementing it using a custom OpenAiChatModel, but the SyncRequestExecutor and RequestExecutor classes are private and cannot be used.
(1) May I ask if there are any other ways to define request headers and customize request parameters;
(2) Can we modify the relevant class access properties in the package dev.langchain4j.model.openai-INternal to public to support custom request headers and parameters.

Thank you, thank you. 
Looking forward to your guidance.

dliubarskyi:
Hi, as for dynamic API key: feel free to open a PR that introduces `custom headers(Producer<Map<String, String>> customHeadersProducer)` in `OpenAiChatModel` builder.
As for personalized properties, this should be possible once https://github.com/langchain4j/langchain4j/pull/2694 is merged and released.
Would that work for you?

answersky:
https://github.com/langchain4j/langchain4j/pull/2694 这个内容大概在什么时候能发布？我现在也遇到了通过langchain4j调用第三方模型需要增加额外的参数

PR #4403, opened 2026-01-10
Add Supplier support for customHeaders in OpenAI models

### LINKED ISSUE #3408: [FEATURE] Add Supplier methods to ChatModel for customHeaders
Currently setting customHeaders is only available during object instantiation (typically with the builder).  If an environment is leveraging authorization headers (like OAuth2), there is no means to update the authorization token (or any header) through the remaining life of the ChatModel object.  Typically the token will expire at some point and the LLM will no longer respond.

A proxy like LiteLLM requires such headers and may be fully enabled/supported with this additional feature (as it states it is OpenAI API compliant, which is already supported by LangChain4j).

It would be beneficial - and fully backwards compatible - if ChatModel implementations (possibly the interface itself) had a mechanism to automatically retrieve any custom headers from a Supplier<Map<String, String>> provided by the developer.  Prior to any http/s connection (GET/POST/ETC), the model should retrieve the fresh set of headers from the supplier and use those.  It is then up to the Supplier implementation to provide the correct and up-to-date headers each time.

This could probably be useful in other LangChain4j interfaces where "customHeaders" are allowed to be provided during object-build.

Maybe something like a **chatModel.setCustomHeaderSupplier()** method?  If no supplier is set, use headers provided by the current .customHeaders() method, per current implementation.

The only current workaround is to monitor the token expiration manually and re-build the chatmodel, which is virtually unusable.

ghtjr410:
## Issue
Closes #3408
Related to #3306 (dynamic customHeaders part)

## Change

Add `Supplier<Map<String, String>>` support for `customHeaders` in `langchain4j-open-ai` module.

### Problem
OAuth2 tokens expire, requiring ChatModel rebuild - impractical for production use.

### Solution
```java
// Before: fixed at build time
.customHeaders(Map.of("Authorization", "Bearer " + token))

// After: fresh on each request  
.customHeaders(() -> Map.of("Authorization", "Bearer " + tokenProvider.getToken()))
```

### Implementation
- Follow existing pattern from `StreamableHttpMcpTransport` in `langchain4j-mcp` module
- Add `customHeaders(Supplier<Map<String, String>>)` overload to all OpenAI model builders
- Use `getOrDefault` pattern for null safety (same as MCP Transport)
- Backward compatible: existing `customHeaders(Map)` still works (wrapped internally as `() -> map`)

### Modified files
- `OpenAiClient.java` - Add Supplier field and overloaded method in Builder
- `DefaultOpenAiClient.java` - Add `buildRequestHeaders()` helper, call Supplier per request
- 7 model classes: `OpenAiChatModel`, `OpenAiStreamingChatModel`, `OpenAiEmbeddingModel`, `OpenAiImageModel`, `OpenAiModerationModel`, `OpenAiLanguageModel`, `OpenAiStreamingLanguageModel`
- New test: `OpenAiCustomHeadersSupplierTest.java`

### Scope
This PR covers `langchain4j-open-ai` module only. If this approach is approved, I'm happy to apply the same pattern to other modules (e.g., `langchain4j-azure-open-ai`, `langchain4j-ollama`) in follow-up PRs for consistency.

## General checklist

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
@@ -38,6 +38,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
 import org.slf4j.Logger;
 
 /**
@@ -73,7 +74,7 @@ public class OpenAiChatModel implements ChatModel {
                 .logResponses(getOrDefault(builder.logResponses, false))
                 .logger(builder.logger)
                 .userAgent(DEFAULT_USER_AGENT)
-                .customHeaders(builder.customHeaders)
+                .customHeaders(builder.customHeadersSupplier)
                 .customQueryParams(builder.customQueryParams)
                 .build();
         this.maxRetries = getOrDefault(builder.maxRetries, 2);
@@ -146,7 +147,7 @@ public class OpenAiChatModel implements ChatModel {
         validate(parameters);
 
         ChatCompletionRequest openAiRequest = toOpenAiChatRequest(
-                        chatRequest, parameters, sendThinking, thinkingFieldName,strictTools, strictJsonSchema)
+                        chatRequest, parameters, sendThinking, thinkingFieldName, strictTools, strictJsonSchema)
                 .build();
 
         ParsedAndRawResponse<ChatCompletionResponse> parsedAndRawResponse = withRetryMappingExceptions(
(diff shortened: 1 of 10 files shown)


dliubarskyi (approved):
@ghtjr410 thank you!

ghtjr410:
@dliubarskyi Should I apply this pattern to other modules (Azure OpenAI, Ollama, etc.) for API consistency? Or better to wait until there's actual demand?

dliubarskyi:
@ghtjr410 feel free to do the same for other modules

ghtjr410:
> @ghtjr410 feel free to do the same for other modules

@dliubarskyi Thanks! I'd be happy to take this on — I'll work through the other modules one by one and open separate PRs.

ajweber:
Ah, I didn't notice the changes because I thought they should be in dev.langchain4j.http.client.HttpRequest

That would make them available to all modules, I believe.  Nevertheless, TY

PR #4405, opened 2026-01-10
MistralAi: return raw HTTP response and SSE events

### LINKED ISSUE #4222: [FEATURE] Access of response headers in ChatModelListener
We re using the the ChatModelLister to track token usage. This is fine, as the information is encapsuled into the parsed context (I guess for most models which send this information in the body)

But we also want to track the ratelimit. Some model provider like OpenAi, Anthropic, etc. will send these information as response header. Example: "x-ratelimit-limit-requests", "x-ratelimit-limit-tokens", "x-ratelimit-remaining-requests", etc.

There are two possible solutions for that:
- parsing the various rate-limit headers and unify them inside the ResponseContext or in the meta data
- giving access to the raw HTTP-Headers would allow a more flexible usage and a custom implementation. It would also enable other use cases too

Actually I would prefer the first solution, but I propose the second one here, as it is more flexible. Or maybe both ;)

The workaround solution is to completly override the HttpClient(Builder) and pass a modified proxy version of it, which has access to the data. The problem with this solution is that it requires two different concepts to observe/track data. Also it does not work for some providers, which are using an own Client-Stack (like OfficailOpenAi, AzureOpenAi, etc)

HostedDinner:
## Issue
Closes #4222

## Change
Similiar to #4225: Exposes raw HTTP Header via `MistralAiChatResponseMetadata` for MistralAiChatModel by passing through the data from HTTP client.

## General checklist

### DIFF
--- a/langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiChatModelIT.java
+++ b/langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiChatModelIT.java
@@ -11,6 +11,7 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import dev.langchain4j.data.message.AiMessage;
 import dev.langchain4j.data.message.UserMessage;
 import dev.langchain4j.model.chat.ChatModel;
+import dev.langchain4j.model.chat.request.ChatRequest;
 import dev.langchain4j.model.chat.request.ResponseFormat;
 import dev.langchain4j.model.chat.request.ResponseFormatType;
 import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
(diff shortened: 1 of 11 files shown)


langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiChatModelIT.java
```diff
@@ -160,4 +161,31 @@ void should_handle_timeout(int millis) {
         assertThatThrownBy(() -> model.chat("hi"))
                 .isExactlyInstanceOf(dev.langchain4j.exception.TimeoutException.class);
     }
+
+    @Test
+    void should_get_raw_response() {
```
dliubarskyi, line 166: Could you please add a similar test to the `MistralAiStreamingChatModelIT`?

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/common/MistralAiStreamingChatModelIT.java
```diff
@@ -51,9 +53,12 @@ protected String customModelName() {
...
     protected ChatRequestParameters createIntegrationSpecificParameters(int maxOutputTokens) {
-        return ChatRequestParameters.builder()
-                .maxOutputTokens(maxOutputTokens)
-                .build();
+        return ChatRequestParameters.builder().maxOutputTokens(maxOutputTokens).build();
+    }
+
+    @Override
+    protected Class<? extends ChatResponseMetadata> chatResponseMetadataType(StreamingChatModel model) {
+        return MistralAiChatResponseMetadata.class;
```
dliubarskyi, line 61: The same change is required in the MistralAiStreamingAiServiceIT

dliubarskyi (commented):
@HostedDinner thank you!

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/common/MistralAiStreamingChatModelIT.java
HostedDinner: Done. Got a little bit confused about the tests in commons vs. in the parent folder

langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiChatModelIT.java
HostedDinner: Done

dliubarskyi (approved):
@HostedDinner thank you!

dliubarskyi:
Please document this feature on https://docs.langchain4j.dev/integrations/language-models/mistral-ai, similarly to https://docs.langchain4j.dev/integrations/language-models/open-ai#accessing-raw-http-responses-and-server-sent-events-sse

PR #4414, opened 2026-01-12
#4413 - PGVector does not escape values in an IN (notIn) filter

### LINKED ISSUE #4413: [BUG] PGVector does not escape values in an IN (notIn) filter
**Describe the bug**

The method `dev.langchain4j.store.embedding.pgvector.PgVectorFilterMapper#formatValuesAsString` does not escape the values of an IN filter and the embedding store search method may fail with an error `java.lang.RuntimeException: org.postgresql.util.PSQLException: Unterminated string literal started at position xxx` if any of the IN values of the metadata filter contains an unescaped '.

This behaviour is unique for the IN and NOT IN filters as the rest of the filter do escape the value with the method `dev.langchain4j.store.embedding.pgvector.PgVectorFilterMapper#formatValue`.

Additionally, insertions do not require to escape the values as they are inserted within Prepared Statements.

**Log and Stack trace**

> java.lang.RuntimeException: org.postgresql.util.PSQLException: Unterminated string literal started at position 15,503 in SQL SELECT (2 - (embedding <=> ... ;. Expected  char
	at dev.langchain4j.store.embedding.pgvector.PgVectorEmbeddingStore.search(PgVectorEmbeddingStore.java:340)
	at dev.langchain4j.store.embedding.pgvector.PgVectorEmbeddingStoreIT.test_escape_in(PgVectorEmbeddingStoreIT.java:108)
Caused by: org.postgresql.util.PSQLException: Unterminated string literal started at position 15,503 in SQL SELECT (2 - (embedding <=> ... ;. Expected  char
	at org.postgresql.core.Parser.checkParsePosition(Parser.java:1435)
	at org.postgresql.core.Parser.parseSql(Parser.java:1342)

**To Reproduce**

```
    @Test
    void test_escape_in() {
        TextSegment[] segments = new TextSegment[] {
                TextSegment.from(
                        "toEscape",
                        Metadata.from(Map.of("text", "This must be escaped '"))),
                TextSegment.from(
                        "notEscape",
                        Metadata.from(Map.of("text", "This does not require to be escaped")))
        };
        List<Embedding> embeddings = new ArrayList<>(segments.length);
        for (TextSegment segment : segments) {
            Embedding embedding = embeddingModel().embed(segment.text()).content();
            embeddings.add(embedding);
        }

        List<String> ids = embeddingStore().addAll(embeddings, Arrays.asList(segments));
        awaitUntilAsserted(() -> assertThat(getAllEmbeddings()).hasSameSizeAs(segments));

        Filter filterEQ = metadataKey("text").isEqualTo("This must be escaped '");

        EmbeddingSearchRequest eqSearchRequest = EmbeddingSearchRequest.builder()
                .maxResults(1)
                .queryEmbedding(embeddings.get(0))
                .filter(filterEQ)
                .build();


        // EqualsTo works as expected
        EmbeddingSearchResult<TextSegment> searchResult = embeddingStore().search(eqSearchRequest);
        EmbeddingMatch<TextSegment> match = searchResult.matches().get(0);
        assertThat(match.score()).isCloseTo(1, withPercentage(1));
        assertThat(match.embeddingId()).isEqualTo(ids.get(0));


        // In filter does not escape values
        Filter filterIN = metadataKey("text").isIn("This must be escaped '");
        EmbeddingSearchRequest inSearchRequest = EmbeddingSearchRequest.builder()
                .maxResults(1)
                .queryEmbedding(embeddings.get(0))
                .filter(filterIN)
                .build();

        searchResult = embeddingStore().search(inSearchRequest);
        match = searchResult.matches().get(0);
        assertThat(match.score()).isCloseTo(1, withPercentage(1));
        assertThat(match.embeddingId()).isEqualTo(ids.get(0));
    }
```

**Expected behavior**

Both searches should work the same way. However, the second one (the `isIn `one) fails with an exception

- LangChain4j version: 1.11.0-beta19-SNAPSHOT
- Java version: Java 21

**Additional context**

odelpozo-denodo:
## Issue

Fixes #4413

## Change

Changed the method `dev.langchain4j.store.embedding.pgvector.PgVectorFilterMapper#formatValuesAsString`. Instead of formatting always the value as '<value>', now, for String values, we call to the method `formatValue` that is already used in the other filters that does not use collections.

It is important that for values that are not String, we return a text value ('<value>') instead of the `value.toString()` that returns the `formatValue` method because otherwise, IN filters with integer (for instance) will fail with a required CAST between text and integer. With the changes of this PR, I just keep the current behaviour of the filter.

## General checklist

- [X] The tests cover both positive and negative cases (existing tests already cover negative cases)

### DIFF
--- a/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorFilterMapper.java
+++ b/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorFilterMapper.java
@@ -144,7 +144,16 @@ abstract class PgVectorFilterMapper {
         }
     }
 
+    String formatCollectionValue(Object value) {
+        if (value instanceof String stringValue) {
+            final String escapedValue = stringValue.replace("'", "''");
+            return '\'' + escapedValue + '\'';
+        } else {
+            return '\'' + value.toString() + '\'';
+        }
+    }
+
     String formatValuesAsString(Collection<?> values) {
-        return "(" + values.stream().map(v -> format("'%s'", v)).collect(Collectors.joining(",")) + ")";
+        return "(" + values.stream().map(this::formatCollectionValue).collect(Collectors.joining(",")) + ")";
     }
 }
--- a/langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStoreIT.java
+++ b/langchain4j-pgvector/src/test/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStoreIT.java
@@ -92,5 +92,18 @@ class PgVectorEmbeddingStoreIT extends EmbeddingStoreWithFilteringIT {
         EmbeddingMatch<TextSegment> match = searchResult.matches().get(0);
         assertThat(match.score()).isCloseTo(1, withPercentage(1));
         assertThat(match.embeddingId()).isEqualTo(ids.get(0));
+
+        // In filter escapes values as well
+        Filter filterNotIN = metadataKey("text").isNotIn("This must be escaped '");
+        EmbeddingSearchRequest notInSearchRequest = EmbeddingSearchRequest.builder()
+                .maxResults(1)
+                .queryEmbedding(embeddings.get(0))
+                .filter(filterNotIN)
+                .build();
+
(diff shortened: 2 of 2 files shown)


dliubarskyi (approved):
@odelpozo-denodo thank you!

PR #3816, opened 2025-10-06
feat(open-ai): streaming model for Open AI responses API

maxlepikhin:
## Issue
#3252

## Change

This PR adds support for OpenAI's **Responses API** (`/v1/responses` endpoint) with two implementations:

### 1. Non-official implementation (`langchain4j-open-ai` module)
- **Class**: `OpenAiResponsesStreamingChatModel`
- Uses manual HTTP client with JSON serialization
- Provides full control over request/response handling
- Includes `OpenAiResponsesValidator` for parameter validation

### 2. Official implementation (`langchain4j-open-ai-official` module)
- **Class**: `OpenAiOfficialResponsesStreamingChatModel`
- Uses official OpenAI Java SDK v4.1.0
- Leverages SDK's built-in features and updates
- Includes `ResponseCancellationHandler` for stream cancellation

### Key Features
- **Streaming-only implementation**: Currently only streaming versions are implemented (non-streaming could be added in the future)
- **Tool execution**: Full support for function calling with proper tool execution flow
- **Cancellation**: Streaming responses can be cancelled via `StreamingHandle.cancel()`
- **Validation**: Both implementations validate parameters with fail-fast approach and clear error messages
- **Listener support**: Full integration with `ChatModelListener` for request/response monitoring
- **Metadata**: Complete response metadata including token usage, model name, finish reason

### Implementation Choices

- **Non-official**: For users who prefer manual control, minimal dependencies, or need custom HTTP client configuration
- **Official**: For users who want SDK-maintained compatibility and automatic updates

**Why `OpenAIClient` instead of `OpenAIClientAsync`?**
The official SDK's `OpenAIClient` provides a synchronous streaming API that returns a `Stream<ResponseEvent>`. This is actually more suitable for our use case than `OpenAIClientAsync` because:
1. We process events sequentially in order
2. We need to handle backpressure naturally via stream iteration
3. The async client would add unnecessary complexity for event processing
4. We already handle async behavior at the LangChain4j level via `CompletableFuture`

### Current Implementation Support
- ✅ Text content via `TextContent`
- ✅ Images via `ImageContent` (base64-encoded)
- ✅ Images via public URLs
- ❌ Audio content (not yet implemented)
- ❌ Video content (not yet implemented)
- ❌ Document/file attachments (not yet implemented)

### Files Added
**langchain4j-open-ai module (4 files):**
1. `OpenAiResponsesStreamingChatModel.java` - Main implementation
2. `OpenAiResponsesValidator.java` - Parameter validation
[...]

PR #4415, opened 2026-01-12
Migrate from Azure OpenAi to Microsoft Foundry

jdubois:
Microsoft Foundry is the new platform for deploying and operating AI models (not just OpenAi) on Microsoft Cloud.


dliubarskyi (commented):
@jdubois thanks! Many of these changes are breaking, are we ok with it?

jdubois:
@dliubarskyi it shouldn't break for users: the constant is used internally and I'm not sure anyone uses it directly, then that's why I kept the older constant with a "deprecated" notice.
The main issue for me would be the environment variable names, but those will need to be updated anyway, as Azure OpenAI doesn't exist anymore.
Do you see anything else, am I missing something?

dliubarskyi:
Changing the `ModelProvider` constant, environmant variables and method names (e.g., `azureDeploymentName` -> `microsoftFoundryDeploymentName`) are all breaking changes which will affect users.

> the constant is used internally

Not only internally, `ChatModelListener` exposes it in `ChatModelRequestContext`, so users with observability will be affected. Some users might record it, others might do something conditionally (e..g, `provider == AZURE_OPEN_AI`).

I would avoid changing deprecated Azure OpenAI module and do the changes only in the OpenAI official module

> but those will need to be updated anyway, as Azure OpenAI doesn't exist anymore

Can't we keep both old and new variable names? This change can be done in a backwards compaptible way

Same for method names, we could keep the existing ones, deprecate thhem and add new

dliubarskyi:
@jdubois FYI, the release is planned for tomorrow afternood. Would you like to include this PR in the release?

jdubois:
@dliubarskyi just when I was going to start working on it :-)

jdubois:
So indeed @dliubarskyi I missed many things!

Here are my new changes:

- We should now support Azure OpenAI and Microsoft Foundry in parallel: environment variables for both should be supported, methods for both are duplicated (but Azure OpenAI methods are marked deprecated).
- I rollbacked the changes for the Azure OpenAI module, so it only stays with Azure OpenAI
- I've implemented similar changes in the `OpenAiOfficialResponsesStreamingChatModel` class from @maxlepikhin coming from PR #3816 so we should be consistent

Be careful @dliubarskyi : I had some inconsistencies in the `pom.xml` files, some referencing `1.11.0-beta19-SNAPSHOT`and some referencing `1.11.0-SNAPSHOT` -> I'm not committing my changes on those, as I believe this is because you're preparing a release.

I've just run all the tests on Microsoft Foundry and they are green!

dliubarskyi:
@jdubois thank you!

> Be careful @dliubarskyi : I had some inconsistencies in the pom.xml files, some referencing 1.11.0-beta19-SNAPSHOTand some referencing 1.11.0-SNAPSHOT -> I'm not committing my changes on those, as I believe this is because you're preparing a release.

Not sure what exactly are you referring to, but for latest release we have stable modules on `1.10.0` version and beta modules on `1.10.0-beta18`, so seeing `1.11.0-beta19-SNAPSHOT` and `1.11.0-SNAPSHOT` together is OK. Not all modules are stable yet...

dliubarskyi (approved):
@jdubois thank you!

PR #4416, opened 2026-01-13
feat: add ChatMemory.set method

### LINKED ISSUE #4172: [FEATURE] ChatMemory design
The entire ChatMemory contract is quite a pain: 

There is no unique identifier for messages, other than comparing content/hashes and relative position to other messages to determine uniqueness.

The `DefaultCommittableChatMemory` commit() method repeatedly calls the "delegate's" add() method with no clear indicator of when the last message has been reached. Why not add a batch/commit method and allow all the messages to be passed at once?

This presents several problems. 

#1 you assemble the list of messages to return as the memory in messages() and you need to retain/track of what was returned in messages() for that request, somewhere. 

#2 As ChatMemory.add() is called you need to keep a separate buffer of everything accumulated from prior add() invocations and then re-pass this this entire buffer to the store iterate over all of them to and compare each both ordinally and with hashes against what you was previously known to have been returned from the last messages() invocation to determine what is new.

#3. for each "new" message you need to persist it right then and there or queue it off somewhere with good guarantees.... because you don't know if this is the last call to add() and there is no guaranteed that messages() will be called again. There is no indicator that all the messages are done being "added".

All of this just seems to make efficient handling/persistence and summarization tasks etc. more difficult. 

https://github.com/langchain4j/langchain4j/issues/2479

https://github.com/langchain4j/langchain4j/issues/1966#issuecomment-2428760190

https://github.com/langchain4j/langchain4j/issues/1980

https://github.com/langchain4j/langchain4j/issues/1184

https://github.com/langchain4j/langchain4j/pull/2593

chirino:
This allows implementations of ChatMemory to atomically update the stores using a single operation.

This should fix #4172

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
@@ -74,6 +74,8 @@ public interface ChatMemory {
      * This method will typically be used when chat memory needs to be re-written to implement things like
      * memory compaction.
      *
+     * NOTE: This method is never called automatically by LangChain4j.
+     *
      * @param messages The {@link ChatMessage}s to set. Must not be {@code null} or empty.
      * @since 1.11.0
      */
--- a/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
@@ -46,6 +46,27 @@ public interface ChatMemory {
     }
 
     /**
+     * Sets the messages of the chat memory
+     * @param messages The {@link ChatMessage}s to set
+     */
+    default void set(ChatMessage... messages) {
+        if ((messages != null) && (messages.length > 0)) {
+            set(Arrays.asList(messages));
+        }
+    }
+
+    /**
+     * Sets the messages of the chat memory
+     * @param messages The {@link ChatMessage}s to set
+     */
+    default void set(Iterable<ChatMessage> messages) {
+        clear();
+        if (messages != null) {
+            messages.forEach(this::add);
+        }
+    }
+
+    /**
      * Retrieves messages from the chat memory.
      * Depending on the implementation, it may not return all previously added messages,
      * but rather a subset, a summary, or a combination thereof.
--- a/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
@@ -2,9 +2,10 @@ package dev.langchain4j.memory;
 
 import dev.langchain4j.data.message.ChatMessage;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.List;
-import java.util.Objects;
+
+import static dev.langchain4j.internal.ValidationUtils.ensureNotEmpty;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 
 /**
  * Represents the memory (history) of a chat conversation.
@@ -53,14 +54,14 @@ public interface ChatMemory {
      * rather than appending to it.
      * <p>
      * The default implementation delegates to {@link #set(Iterable) set(Iterable&lt;ChatMessage&gt;)}.
-     *
+     * <p>
      * NOTE: This method is never called automatically by LangChain4j.
      *
      * @param messages The {@link ChatMessage}s to set. Must not be {@code null} or empty.
      * @since 1.11.0
      */
     default void set(ChatMessage... messages) {
-        Objects.requireNonNull(messages, "messages must not be null");
+        ensureNotEmpty(messages, "messages");
         set(Arrays.asList(messages));
     }
 
@@ -75,19 +76,15 @@ public interface ChatMemory {
      * <p>
      * This method will typically be used when chat memory needs to be re-written to implement things like
      * memory compaction.
-     *
+     * <p>
      * NOTE: This method is never called automatically by LangChain4j.
      *
      * @param messages The {@link ChatMessage}s to set. Must not be {@code null} or empty.
      * @since 1.11.0
      */
     default void set(Iterable<ChatMessage> messages) {
-        Objects.requireNonNull(messages, "messages must not be null");
-        if (messages instanceof Collection<?> collection) {
-            if (collection.isEmpty()) {
-                throw new IllegalArgumentException("messages must not be empty");
-            }
-        } else if (!messages.iterator().hasNext()) {
+        ensureNotNull(messages, "messages");
+        if (!messages.iterator().hasNext()) {
             throw new IllegalArgumentException("messages must not be empty");
         }
         clear();
--- a/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
@@ -2,7 +2,9 @@ package dev.langchain4j.memory;
 
 import dev.langchain4j.data.message.ChatMessage;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Represents the memory (history) of a chat conversation.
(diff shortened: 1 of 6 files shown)


chirino:
Should it be an error to try to set a null or empty list or should it just be treated a noop?

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -45,6 +45,27 @@ default void add(Iterable<ChatMessage> messages) {
         }
     }
 
+    /**
+     * Sets the messages of the chat memory
+     * @param messages The {@link ChatMessage}s to set
+     */
```
dliubarskyi, line 62:
```suggestion
     * @since 1.11.0
     */
```

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -45,6 +45,27 @@ default void add(Iterable<ChatMessage> messages) {
...
+    default void set(ChatMessage... messages) {
+        if ((messages != null) && (messages.length > 0)) {
+            set(Arrays.asList(messages));
+        }
+    }
+
+    /**
+     * Sets the messages of the chat memory
+     * @param messages The {@link ChatMessage}s to set
+     */
```
dliubarskyi, line 84:
```suggestion
     * @since 1.11.0
     */
```

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -45,6 +45,27 @@ default void add(Iterable<ChatMessage> messages) {
         }
     }
 
+    /**
+     * Sets the messages of the chat memory
```
dliubarskyi: Please add a short description how it is different from `add`

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -45,6 +45,27 @@ default void add(Iterable<ChatMessage> messages) {
...
+     * @param messages The {@link ChatMessage}s to set
+     */
+    default void set(ChatMessage... messages) {
+        if ((messages != null) && (messages.length > 0)) {
+            set(Arrays.asList(messages));
+        }
+    }
+
+    /**
+     * Sets the messages of the chat memory
```
dliubarskyi: same here

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
```diff
@@ -91,6 +93,25 @@ public void add(ChatMessage message) {
...
 
+    private void set(final List<ChatMessage> messages) {
+        Integer maxMessages = this.maxMessagesProvider.apply(this.id);
+        ensureGreaterThanZero(maxMessages, "maxMessages");
+        ensureCapacity(messages, maxMessages);
+        store.updateMessages(id, messages);
+    }
+
+    @Override
+    public void set(final ChatMessage... messages) {
```
dliubarskyi: Is this method override required?

langchain4j/src/main/java/dev/langchain4j/memory/chat/TokenWindowChatMemory.java
```diff
@@ -95,6 +97,25 @@ public void add(ChatMessage message) {
...
 
+    private void set(final List<ChatMessage> messages) {
+        Integer maxTokens = maxTokensProvider.apply(id);
+        ensureGreaterThanZero(maxTokens, "maxTokens");
+        ensureCapacity(messages, maxTokens, tokenCountEstimator);
+        store.updateMessages(id, messages);
+    }
+
+    @Override
+    public void set(final ChatMessage... messages) {
```
dliubarskyi: same here

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
```diff
@@ -91,6 +93,25 @@ public void add(ChatMessage message) {
...
+        set(Arrays.asList(messages));
+    }
+
+    @Override
+    public void set(final Iterable<ChatMessage> iter) {
+        List<ChatMessage> list = new ArrayList<>();
+        iter.forEach(list::add);
+        set(list);
```
dliubarskyi: slight optimization: if `iter` is already a `List`, just pass it instead of creating another list and iterating

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -45,6 +45,27 @@ default void add(Iterable<ChatMessage> messages) {
...
+     * @param messages The {@link ChatMessage}s to set
+     */
+    default void set(Iterable<ChatMessage> messages) {
+        clear();
+        if (messages != null) {
```
dliubarskyi: nit: you could call `add(Iterable<ChatMessage> messages)`, so that implementations that already optimize adding multiple messages at once will benefit from it

langchain4j/src/main/java/dev/langchain4j/memory/chat/TokenWindowChatMemory.java
```diff
@@ -95,6 +97,25 @@ public void add(ChatMessage message) {
...
+        set(Arrays.asList(messages));
+    }
+
+    @Override
+    public void set(final Iterable<ChatMessage> iter) {
+        List<ChatMessage> list = new ArrayList<>();
+        iter.forEach(list::add);
+        set(list);
```
dliubarskyi: same here

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
dliubarskyi: And also would be nice to clarify:
- why this method is requried
- who is using it (or not) - for example, LC4j does not call it
- who/why sould implement/override it

dliubarskyi (commented):
@chirino, thanks! Please add tests

dliubarskyi:
> Should it be an error to try to set a null or empty list or should it just be treated a noop?

I think it should throw an exception in both cases. If anyone wants to just clear the memory, they can call `clear()`. Doing `set(null)` or `set(List.of())` seems more like a bug

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
chirino: will do

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
chirino: Yes.. so that the store can be updated in an atomic operation.

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
chirino: good idea.

chirino (commented):
I think latest commit should address all the review concerns.

langchain4j/src/main/java/dev/langchain4j/memory/chat/MessageWindowChatMemory.java
dliubarskyi: If we remove this (redundant) implementation, the default `set(final ChatMessage...)` implementation from the `ChatMemory` will call the `set(final Iterable<ChatMessage> iter)` from this class and will do it atomically, no?

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -45,6 +47,49 @@ default void add(Iterable<ChatMessage> messages) {
...
+     * Unlike {@link #add(Iterable) add}, this method replaces the entire message history
+     * rather than appending to it.
+     * <p>
+     * Implementations should override this method to provide more efficient atomic operations if possible.
+     * The default implementation calls {@link #clear()} followed by {@link #add(Iterable) add(Iterable&lt;ChatMessage&gt;)}
+     * which is not atomic.
+     * <p>
+     * This method will typically be used when chat memory needs to be re-written to implement things like
+     * memory compaction.
+     *
```
dliubarskyi:
```suggestion
     *
     * NOTE: This method is never called automatically by LangChain4j.
     *
```

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -45,6 +47,51 @@ default void add(Iterable<ChatMessage> messages) {
         }
     }
 
+    /**
+     * Replaces all messages in the chat memory with the specified messages.
+     * Unlike {@link #add(ChatMessage...) add}, this method replaces the entire message history
+     * rather than appending to it.
+     * <p>
+     * The default implementation delegates to {@link #set(Iterable) set(Iterable&lt;ChatMessage&gt;)}.
+     *
```
dliubarskyi:
```suggestion
     *
     * NOTE: This method is never called automatically by LangChain4j.
     *
```

langchain4j-core/src/main/java/dev/langchain4j/memory/ChatMemory.java
```diff
@@ -45,6 +47,53 @@ default void add(Iterable<ChatMessage> messages) {
...
+     * memory compaction.
+     *
+     * NOTE: This method is never called automatically by LangChain4j.
+     *
+     * @param messages The {@link ChatMessage}s to set. Must not be {@code null} or empty.
+     * @since 1.11.0
+     */
+    default void set(Iterable<ChatMessage> messages) {
+        Objects.requireNonNull(messages, "messages must not be null");
+        if (messages instanceof Collection<?> collection) {
```
dliubarskyi: nit: can be simplified by only testing `!messages.iterator().hasNext()`

dliubarskyi (approved):
@chirino thank you!

PR #4436, opened 2026-01-16
Add support for multimodal agents

### LINKED ISSUE #4123: [FEATURE] Support ImageModel in Agent
My agentic workflow requires generating an image using an ImageModel. I could use a Non-AI agent to achieve this. However, we would gain additional benefits if the AiServices built with DefaultAiServices supported calling ImageModel (e.g., if an ImageModel was set in the AiServiceContext and the interface method returned an Image/Response<Image>), such as GuardrailService calling.

mariofusco:
## Issue

Closes #4123

## Change

The main goal of this pull request is allowing an `AiService`, and then even an agent, to also return an image, other than taking it as an input argument. To be honest, I have a few doubt both on the idea and the implementation, so I'm completely open to discuss this.

## General checklist


langchain4j-agentic/src/test/java/dev/langchain4j/agentic/MultimodalAgentsIT.java
```diff
@@ -0,0 +1,172 @@
...
+
+    public interface AnimalsIdentifier {
+
+        @SystemMessage("You are an expert in identifying animals from images.")
+        @UserMessage("""
+            Recognize the type of animals present in the given image.
+            Reply with the name of the animal only and nothing else, using the plural if they are more than one.
+            """)
+        @Agent("Recognize the type of animals present in an image.")
+        String identify(@UserMessage @V("animalsImage") ImageContent animalsImage);
```
dliubarskyi, line 40: Why is `@V("animalsImage")` required?

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/MultimodalAgentsIT.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public interface AnimalsCounter {
+
+        @SystemMessage("You are an expert in counting specific animals from images.")
+        @UserMessage("""
+            How many {{animalType}} are present in the given image?
+            """)
+        @Agent("Count the number of animals of a given type present in an image.")
+        int count(@UserMessage @V("animalsImage") ImageContent animalsImage, @V("animalType") String animalType);
```
dliubarskyi, line 50: same here and in all other places

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/MultimodalAgentsIT.java
```diff
@@ -0,0 +1,172 @@
...
+        @UserMessage("Generate a detailed description of a {{style}} scene containing {{sceneContent}}.")
+        @Agent
+        String describeScene(@V("sceneContent") String sceneContent, @V("style") String style);
+    }
+
+    public interface ImageGenerator {
+
+        @UserMessage("A high-resolution, studio-lit product photograph of {{requiredImage}}")
+        @Agent
+        ImageContent generateImageOf(@V("requiredImage") String requiredImage);
```
dliubarskyi, line 127: I would use a more generic `dev.langchain4j.data.image.Image` type (works both for image inputs and outputs) instead of message-specific `dev.langchain4j.data.message.ImageContent` (supposed to represent image input)

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeneratedImageHelper.java
```diff
@@ -1,50 +0,0 @@
...
-import dev.langchain4j.data.image.Image;
-import dev.langchain4j.data.message.AiMessage;
-import java.util.Collections;
-import java.util.List;
-
-/**
- * Helper class for accessing generated images from AiMessage attributes.
- * This provides a convenient way to extract images generated by Gemini's image generation models.
- */
-public class GeneratedImageHelper {
```
dliubarskyi, line 14: This is a breaking change, let's keep it an deprecate insead

langchain4j-core/src/main/java/dev/langchain4j/data/message/AiMessage.java
```diff
@@ -117,6 +121,10 @@ public String thinking() {
         return thinking;
     }
 
+    public List<Image> images() {
```
dliubarskyi, line 131: Worth adding a javadoc mentioning that this works only for Gemini?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -379,12 +382,44 @@ public Object invoke(Method method, Object[] args, InvocationContext invocationC
...
+                        return false;
+                    }
 
-                        return actualResponse;
+                    private static Object parseInternalData(ChatResponse response, Type returnType) {
+                        Class<?> rawReturnType = getRawClass(returnType);
+                        if (LangChain4jData.class.isAssignableFrom(rawReturnType)) {
+                            if (rawReturnType == ImageContent.class) {
+                                List<ImageContent> images = ImageContent.from(response.aiMessage());
+                                return images.isEmpty() ? null : images.get(0);
```
dliubarskyi, line 412: What if there are multiple images generated?
Or if no images were generated, but LLM responded with text?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -379,12 +382,44 @@ public Object invoke(Method method, Object[] args, InvocationContext invocationC
...
+                            Class<?> genericParam = resolveFirstGenericParameterClass(returnType);
+                            return genericParam != null && LangChain4jData.class.isAssignableFrom(genericParam);
+                        }
+                        return false;
+                    }
 
-                        return actualResponse;
+                    private static Object parseInternalData(ChatResponse response, Type returnType) {
+                        Class<?> rawReturnType = getRawClass(returnType);
+                        if (LangChain4jData.class.isAssignableFrom(rawReturnType)) {
```
dliubarskyi: Not sure a new `LangChain4jData` type is required, we can directly check for `Image` return type

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -379,12 +382,44 @@ public Object invoke(Method method, Object[] args, InvocationContext invocationC
...
+                            if (rawReturnType == ImageContent.class) {
+                                List<ImageContent> images = ImageContent.from(response.aiMessage());
+                                return images.isEmpty() ? null : images.get(0);
+                            }
+                        }
+                        if (Collection.class.isAssignableFrom(rawReturnType)) {
+                            Class<?> genericParam = resolveFirstGenericParameterClass(returnType);
+                            if (genericParam == ImageContent.class) {
+                                return ImageContent.from(response.aiMessage());
```
dliubarskyi, line 422: In this case there will be casting exception? User expects a collection, but get's a single element?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: If we'll have more supported output types in the future, we can always add a new marker interface later

dliubarskyi (commented):
@mariofusco thank you!

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: No, that from method returns a `List<ImageContent>`.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: Makes sense, I will remove the `LangChain4jData` for now.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeneratedImageHelper.java
mariofusco: Ok.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: Done.

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GeneratedImageHelper.java
mariofusco: Done.

mariofusco:
@dliubarskyi I modified the test to make it clear both why I'd prefer to keep the `ImageContent` as output type and why the `@V` annotation is necessary on the agent's argument: in this way the `ImageContent` generated by an agent can be directly used as an argument for another agent. Moreover, in order to make this works, the `ImageContent` is stored as a state of the `AgenticScope` and retrieved from there using the variable name, as it happens with any other variable used in the agentic system.

Please let me know if this makes sense or you have any better suggestions.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/MultimodalAgentsIT.java
mariofusco: See https://github.com/langchain4j/langchain4j/pull/4436#issuecomment-3761034885

langchain4j-core/src/main/java/dev/langchain4j/data/message/AiMessage.java
mariofusco: Done.

glaforge:
I haven't reviewed the code (yet) but as a comment on the idea and your doubts Mario.
I think it's great to be able to have AI services and agents support other modalities.
For a marketing agent, you can use an image generation model to create various image assets.
For some deep research agent, you can generate infographics, illustrations, diagrams.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/Models.java
```diff
@@ -110,13 +135,40 @@ public static ChatModel plannerModel() {
...
+    public static StreamingChatModel streamingBaseModel(MODEL_PROVIDER modelProvider) {
         return switch (modelProvider) {
             case OPENAI -> OPENAI_STREAMING_BASE_MODEL;
             case OLLAMA -> OLLAMA_STREAMING_BASE_MODEL;
             case GEMINI -> GEMINI_STREAMING_BASE_MODEL;
         };
     }
 
+    public static ChatModel visionModel() {
+        return visionModel(modelProvider);
```
glaforge, line 150: In this PR, we assume that image models are multimodal chat models, is that right?
There are image models, and there are chat models that support multiple modalities.
For example, the Imagen 4 model from Google is a pure image model, while Nano Banana is a multimodal model that accepts and returns text, images, etc.
Does this mean we can only support the multimodal chat models? And/or could we also support pure image models?

mariofusco: I believe we only support multimodal chat models for now (@dliubarskyi please confirm), but anyway this is not a issue related with this pull request or with the agentic module in general. This has to be eventually fix at the langchain4j core level, and then every model supported by langchain4j could also be used by agents without any change (or with minimal ones).

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/Agents.java
```diff
@@ -4,6 +4,7 @@
 import dev.langchain4j.agentic.planner.AgentInstance;
 import dev.langchain4j.agentic.scope.AgenticScopeAccess;
 import dev.langchain4j.agentic.scope.ResultWithAgenticScope;
+import dev.langchain4j.data.message.ImageContent;
```
dliubarskyi, line 7:
```suggestion
```

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/Models.java
dliubarskyi: I am not sure why `ImageModel` should be used by the AI Service, as all the AI Service features (e.g., memory, tools, etc) are `ChatModel`-centered

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/MultimodalAgentsIT.java
dliubarskyi: Can we please support both `Image` and `ImageContent` output types?
I understand that when one AI Service/agent output is used as input in other AI Service, it makes sense to use `ImageContent`, although it is still not very intuitive for me as `ImageContent` is supposed to be an input type, not output.
For use cases when a single AI Service generates images, I would rather have `Image` as return type.

langchain4j-core/src/main/java/dev/langchain4j/data/message/AiMessage.java
dliubarskyi: Also please include in the javadoc that images are only used as outputs, they are not sent to the LLM on the follow-up requests

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GeneratedImageHelperTest.java
```diff
@@ -1,5 +1,6 @@
 package dev.langchain4j.model.googleai;
```
dliubarskyi, line 1: please roll back changes to this test, aside from using `GENERATED_IMAGES_KEY`

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -214,13 +217,16 @@ public Object invoke(Method method, Object[] args, InvocationContext invocationC
...
                         // TODO should it be called when returnType==String?
                         boolean supportsJsonSchema = supportsJsonSchema();
-
                         Optional<JsonSchema> jsonSchema = Optional.empty();
-                        if (supportsJsonSchema && !streaming) {
-                            jsonSchema = serviceOutputParser.jsonSchema(returnType);
-                        }
-                        if ((!supportsJsonSchema || jsonSchema.isEmpty()) && !streaming) {
-                            userMessage = appendOutputFormatInstructions(returnType, userMessage);
+                        boolean isInternalData = internalData(returnType);
```
dliubarskyi: please rename from internal data to "image"-something

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: This is confusing... When calling `ImageContent.from()` I would expect to get a single `ImageContent`...
Could you please also move that method from `ImageContent` class to here?

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/Models.java
dliubarskyi: I mean if one wants to generate an image, one can simply use `ImageModel`, the API is easy to use, there is no need to wrap `ImageModel` into `AiService`, there is nothing to gain here

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GeneratedImageHelperTest.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
mariofusco: Done.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/MultimodalAgentsIT.java
mariofusco: Done.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentUtil.java
```diff
@@ -235,6 +238,12 @@ private static Object adaptValueToType(Object value, Class<?> type) {
                 default -> value;
             };
         }
+        if (value instanceof Image image && type == ImageContent.class) {
+            return ImageContent.from(image);
+        }
+        if (value instanceof ImageContent imageContent && type == ImageContent.class) {
```
dliubarskyi, line 244:
```suggestion
        if (value instanceof ImageContent imageContent && type == Image.class) {
```
Probably copy-paste?

dliubarskyi (approved):
@mariofusco thank you!

PR #4447, opened 2026-01-20
Update the MCP specification version to 2025-11-25

jmartisk:
We support all required parts of that spec so I don't see why we shouldn't advertise the newer version.


dliubarskyi (approved):
@jmartisk thanks!

PR #4458, opened 2026-01-22
Fix: Re-evaluate tool filter after tool execution

### LINKED ISSUE #4455: [BUG] Tool filter set in `McpToolProvider` is not evaluated between tool execution (AiServices). Tools are cached and not refreshed.
**Describe the bug**
Tool filter applied in `McpToolProvider` is not evaluated between tool execution. Any dynamic change to the filter based on previous tool call is not reflected in the subsequent response. ToolProvider.provideTools is called at the start of a request and is not recomputed after a tool call.

I am trying to implement a ToolSearch tool which supports progressive disclosure of MCP tools. similar to ([Anthropic Tool Search](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool)). Initially it hides/filters all other tools. The LLM, based on the prompt ,searches for / selects tools that are relevant by invoking this search tool. ToolSearch tool enables/selects relevant tools and changes the filter accordingly. If the filter is reapplied here, next request to LLM (containing the tool result)  should include the newly enabled tools. 

Note: Spring AI provides such a tool out of the box.[Link]( https://spring.io/blog/2025/12/11/spring-ai-tool-search-tools-tzolov). My tool follows exactly the same flow/sequence described in the link. The main difference is, since we don't have a concept of Advisor, the ToolSearch tool's filter is manually applied in the toolProvider.

```java
            McpToolProvider toolProvider = McpToolProvider.builder()
                    .mcpClients(clients)
                    .filter(searchTool.asFilter())
                    .build();
```

**Log and Stack trace**

**To Reproduce**

**Expected behavior**

`mcpToolsFilter` should be reapplied after tool execution, so that any dynamic changes to the tool list (based on filter) is reflected in subsequent responses. 

- LangChain4j version: 1.10.0
- LLM(s) used: Gemini Flash 2.5
- Java version: 25

**Additional context**

As of now, the ToolServiceContext caches the list of tools after applying the filter  (from ToolProvider). This same toolServiceContext is later used to execute tool calls and prepare the response. So if a tool call dynamically changes the filter, it is ignored. These changes are not propagated in the response and hence the LLM does not see the newly enabled tools.

Note that, if we start a fresh request then the tool changes (due to filter) made in previous call are now correctly applied. The problem pertains to a sequence which involves  tool calls, in particular if the tool executed has impact on the filter.

Class DefaultAiServices
```java

  ToolServiceContext toolServiceContext =
          context.toolService.createContext(invocationContext, userMessage);

 // make llm call and get response



  ToolServiceResult toolServiceResult = context.toolService.executeInferenceAndToolsLoop(
          context,
          memoryId,
          chatResponse,
          parameters,
          messages,
          chatMemory,
          invocationContext,
          toolServiceContext,
          isReturnTypeResult);

```

jiaming2li:
## Problem
Tool filters are only evaluated once at request start. Dynamic filter changes during tool execution are ignored, preventing progressive tool disclosure patterns.
Fixes #4455

## Solution
Re-call `createContext()` after each tool execution to refresh the tool list with updated filters.

## Use Case
Enables progressive tool disclosure similar to Spring AI ToolSearch:
1. Initially hide tools
2. LLM searches for needed tools
3. Filter updates to enable relevant tools  
4. LLM sees newly enabled tools
5. LLM uses the tools

## Impact
- No breaking changes
- Minimal performance overhead (~2.8ms per tool execution)
- Backward compatible

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
@@ -347,6 +347,17 @@ public class ToolService {
                 messages = chatMemory.messages();
             }
 
+            UserMessage lastUserMessage = extractLastUserMessage(messages);
+            if (lastUserMessage != null) {
+                toolServiceContext = createContext(invocationContext, lastUserMessage);
+
+                parameters = parameters.overrideWith(
+                        ChatRequestParameters.builder()
+                                .toolSpecifications(toolServiceContext.toolSpecifications())
+                                .build()
+                );
+            }
+
             ChatRequest chatRequest = context.chatRequestTransformer.apply(
                     ChatRequest.builder()
                             .messages(messages)
@@ -368,6 +379,16 @@ public class ToolService {
                 .aggregateTokenUsage(aggregateTokenUsage)
                 .build();
     }
+
+    private UserMessage extractLastUserMessage(List<ChatMessage> messages) {
+        for (int i = messages.size() - 1; i >= 0; i--) {
+            if (messages.get(i) instanceof UserMessage) {
+                return (UserMessage) messages.get(i);
+            }
+        }
+        return null;
+    }
+    //debug ends
 
     private void fireToolExecutedEvent(
             InvocationContext invocationContext,
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
@@ -351,11 +351,9 @@ public class ToolService {
             if (lastUserMessage != null) {
                 toolServiceContext = createContext(invocationContext, lastUserMessage);
 
-                parameters = parameters.overrideWith(
-                        ChatRequestParameters.builder()
-                                .toolSpecifications(toolServiceContext.toolSpecifications())
-                                .build()
-                );
+                parameters = parameters.overrideWith(ChatRequestParameters.builder()
+                        .toolSpecifications(toolServiceContext.toolSpecifications())
+                        .build());
             }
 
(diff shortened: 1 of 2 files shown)


langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -369,6 +378,16 @@ public ToolServiceResult executeInferenceAndToolsLoop(
                 .build();
     }
 
+    private UserMessage extractLastUserMessage(List<ChatMessage> messages) {
+        for (int i = messages.size() - 1; i >= 0; i--) {
```
dliubarskyi: Please move it to UserMessage, and implement similarly to SystemMessage.findLast

dliubarskyi:
> progressive tool disclosure

How do you implement it currently?

dliubarskyi:
It is a breaking change as it changes the behaviour of AI Service from the user point of view.

BTW, tool search tool is going to be supported natively soon, I am working on it. Feel free to review https://github.com/langchain4j/langchain4j/compare/main...tool_search_tool

jiaming2li:
> > progressive tool disclosure
> 
> How do you implement it currently?

I moved extractLastUserMessage logic to UserMessage.findLast(), if there's anything else I need to address, please let me know!

dliubarskyi:
@jiaming2li 

> How do you implement it currently?

I mean how do you (plan to) implement progressive tool disclosure?

If [tool search tool](https://github.com/langchain4j/langchain4j/compare/main...tool_search_tool) will be supported natively, I guess this PR is not required, right?

I am hesitant to change this logic, as it might break existing user applications

jiaming2li:
> tool search tool is going to be supported natively soon

Yes, that is true. I'll close this pr.

PR #4462, opened 2026-01-22
Allows to define untyped agents dynamically without the need of an interface

### LINKED ISSUE #4459: [question] Why is the agent definition designed in the form of "@Agent + Interface"?
Why is the agent definition designed in the form of "@Agent + Interface"?
Is it difficult to create agents dynamically in this way?
For example, the metadata information related to my agent is stored in the database, and then the agent is dynamically created based on the metadata information in the database. What solutions are there for this?
**Describe the bug**

**Log and Stack trace**

**To Reproduce**

**Expected behavior**

- LangChain4j version:
- LLM(s) used: 
- Java version: 

**Additional context**

mariofusco:
## Issue

Closes #4459
Closes #4385

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
@@ -239,6 +239,21 @@ public abstract class AiServices<T> {
     }
 
     /**
+     * Configures the system message to be used each time an AI service is invoked.
+     * It can be either a complete system message or a system message template containing unresolved template
+     * variables (e.g. "{{name}}"), which will be resolved using the values of method parameters annotated with @{@link V}.
+     * <br>
+     * When both {@code @SystemMessage} and the system message provider are configured,
+     * {@code @SystemMessage} takes precedence.
+     *
+     * @param systemMessage The system message to be used.
+     * @return builder
+     */
+    public AiServices<T> systemMessage(String systemMessage) {
+        return systemMessageProvider(ignore -> systemMessage);
+    }
+
+    /**
      * Configures the system message provider, which provides a system message to be used each time an AI service is invoked.
      * <br>
      * When both {@code @SystemMessage} and the system message provider are configured,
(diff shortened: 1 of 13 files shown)


langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -238,6 +238,21 @@ public AiServices<T> streamingChatModel(StreamingChatModel streamingChatModel) {
         return this;
     }
 
+    /**
+     * Configures the system message to be used each time an AI service is invoked.
+     * It can be either a complete system message or a system message template containing unresolved template
+     * variables (e.g. "{{name}}"), which will be resolved using the values of method parameters annotated with @{@link V}.
+     * <br>
+     * When both {@code @SystemMessage} and the system message provider are configured,
+     * {@code @SystemMessage} takes precedence.
```
dliubarskyi, line 247: Hmm, shouldn't it be vice versa? System message in annotation is usually static, but the one configured via this method can be more dynamic, so it feel like it should override the static one. WDYT?

dliubarskyi: Same for user message methods below

dliubarskyi: Could you please also add AI Service tests for new methods?

dliubarskyi (commented):
@mariofusco thank you!

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
mariofusco: This behaves exactly in the same way of the `systemMessageProvider` that I didn't change. I agree that probably it would make more sense to give precedence to the dynamic one, but if so I think I should change also there thus causing a minor backward incompatibility. Please let me know. Same thing also for the `userMessageProvider` and `userMessage` methods of course.

I will also test this for the AI Service, but please let's find an agreement on the desired behaviour first

dliubarskyi: I see... I guess it is better to have an unintuitive precedence than to break things for existing users, so let's keep the existing behaviour. I will create an issue for 2.0 to consider changing the precedence though.

mariofusco: Ok, I will simply add a few more tests then.

mariofusco:
@dliubarskyi I added a couple of tests for the AI Service, please give it a second look.

dliubarskyi (approved):
@mariofusco thanks!

PR #4467, opened 2026-01-23
Injecting context into header providers & support for MCP listeners

### LINKED ISSUE #4241: [FEATURE]: dynamically inject custom HTTP headers on MCP Client tool invocation requests
Hi,
I have a need to dynamically inject one or more HTTP headers on MCP client tool invocation requests (metadata/listings would be good as well).

Whenever a tool is invoked a hook would be nice, that has context of the request (such as the memoryId etc) so that custom headers could be injected prior to the tool being called.

Currently solutions for this are client filters etc which apply to all requests and don't have contextual info to be able to do this easily.

jmartisk:
This introduces a concept of `McpCallContext` that contains information about the context around an invocation to an MCP server. This context is injected into the newly added MCP header providers.

Building on top of that, I added support for MCP listeners that can be invoked before and after calls to MCP servers. They also inject a `McpCallContext`.

Further building on top of that, I've been working on Micrometer metrics support for MCP clients, this will live in the Quarkus extension and will be based on top of MCP listeners. I am drafting that, I'd like to get feedback on this part first before continuing the metrics thing because it builds on top of it.

Closes #4241

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -15,6 +15,7 @@ import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
 import dev.langchain4j.exception.ToolArgumentsException;
 import dev.langchain4j.exception.ToolExecutionException;
+import dev.langchain4j.invocation.InvocationContext;
 import dev.langchain4j.mcp.client.logging.DefaultMcpLogMessageHandler;
 import dev.langchain4j.mcp.client.logging.McpLogMessageHandler;
 import dev.langchain4j.mcp.client.transport.McpOperationHandler;
@@ -194,6 +195,11 @@ public class DefaultMcpClient implements McpClient {
 
     @Override
     public List<ToolSpecification> listTools() {
+        return listTools(null);
+    }
+
+    @Override
+    public List<ToolSpecification> listTools(InvocationContext invocationContext) {
         assertNotClosed();
         if (isToolListRefreshNeeded()) {
             CompletableFuture<Void> updateInProgress = this.toolListUpdateInProgress.get();
@@ -206,7 +212,7 @@ public class DefaultMcpClient implements McpClient {
                 CompletableFuture<Void> update = new CompletableFuture<>();
                 this.toolListUpdateInProgress.set(update);
                 try {
-                    obtainToolList();
+                    obtainToolList(invocationContext);
                 } finally {
                     update.complete(null);
                     toolListOutOfDate.set(false);
@@ -234,6 +240,11 @@ public class DefaultMcpClient implements McpClient {
 
     @Override
     public ToolExecutionResult executeTool(ToolExecutionRequest executionRequest) {
+        return executeTool(executionRequest, null);
+    }
+
+    @Override
+    public ToolExecutionResult executeTool(ToolExecutionRequest executionRequest, InvocationContext invocationContext) {
         assertNotClosed();
         ObjectNode arguments = null;
         try {
@@ -251,7 +262,8 @@ public class DefaultMcpClient implements McpClient {
         CompletableFuture<JsonNode> resultFuture = null;
         JsonNode result = null;
         try {
-            resultFuture = transport.executeOperationWithResponse(operation);
+
+            resultFuture = transport.executeOperationWithResponse(new McpCallContext(invocationContext, operation));
             result = resultFuture.get(timeoutMillis, TimeUnit.MILLISECONDS);
         } catch (TimeoutException timeout) {
             transport.executeOperationWithoutResponse(new McpCancellationNotification(operationId, "Timeout"));
@@ -269,15 +281,25 @@ public class DefaultMcpClient implements McpClient {
 
     @Override
     public List<McpResource> listResources() {
+        return listResources(null);
+    }
+
+    @Override
+    public List<McpResource> listResources(InvocationContext invocationContext) {
         assertNotClosed();
         if (resourceRefs.get() == null) {
-            obtainResourceList();
+            obtainResourceList(invocationContext);
         }
         return resourceRefs.get();
     }
 
     @Override
     public McpReadResourceResult readResource(String uri) {
+        return readResource(uri, null);
+    }
+
+    @Override
+    public McpReadResourceResult readResource(String uri, InvocationContext invocationContext) {
         assertNotClosed();
         final long operationId = idGenerator.getAndIncrement();
         McpReadResourceRequest operation = new McpReadResourceRequest(operationId, uri);
@@ -285,7 +307,7 @@ public class DefaultMcpClient implements McpClient {
         JsonNode result = null;
         CompletableFuture<JsonNode> resultFuture = null;
         try {
-            resultFuture = transport.executeOperationWithResponse(operation);
+            resultFuture = transport.executeOperationWithResponse(new McpCallContext(invocationContext, operation));
             result = resultFuture.get(timeoutMillis, TimeUnit.MILLISECONDS);
             return ResourcesHelper.parseResourceContents(result);
         } catch (ExecutionException | InterruptedException | TimeoutException e) {
@@ -347,16 +369,21 @@ public class DefaultMcpClient implements McpClient {
 
     @Override
     public List<McpResourceTemplate> listResourceTemplates() {
+        return listResourceTemplates(null);
+    }
+
+    @Override
+    public List<McpResourceTemplate> listResourceTemplates(InvocationContext invocationContext) {
         assertNotClosed();
         if (resourceTemplateRefs.get() == null) {
-            obtainResourceTemplateList();
+            obtainResourceTemplateList(invocationContext);
         }
         return resourceTemplateRefs.get();
     }
 
-    private synchronized void obtainToolList() {
+    private synchronized void obtainToolList(InvocationContext invocationContext) {
         McpListToolsRequest operation = new McpListToolsRequest(idGenerator.getAndIncrement());
-        CompletableFuture<JsonNode> resultFuture = transport.executeOperationWithResponse(operation);
+        CompletableFuture<JsonNode> resultFuture = transport.executeOperationWithResponse(new McpCallContext(invocationContext, operation));
         JsonNode result = null;
         try {
             result = resultFuture.get();
@@ -371,7 +398,7 @@ public class DefaultMcpClient implements McpClient {
         toolListRefs.set(toolList);
     }
 
-    private synchronized void obtainResourceList() {
+    private synchronized void obtainResourceList(InvocationContext invocationContext) {
         if (resourceRefs.get() != null) {
             return;
         }
@@ -380,7 +407,7 @@ public class DefaultMcpClient implements McpClient {
         JsonNode result = null;
         CompletableFuture<JsonNode> resultFuture = null;
         try {
-            resultFuture = transport.executeOperationWithResponse(operation);
+            resultFuture = transport.executeOperationWithResponse(new McpCallContext(invocationContext, operation));
             result = resultFuture.get(timeoutMillis, TimeUnit.MILLISECONDS);
             resourceRefs.set(ResourcesHelper.parseResourceRefs(result));
         } catch (ExecutionException | InterruptedException | TimeoutException e) {
@@ -390,7 +417,7 @@ public class DefaultMcpClient implements McpClient {
         }
     }
 
-    private synchronized void obtainResourceTemplateList() {
+    private synchronized void obtainResourceTemplateList(InvocationContext invocationContext) {
         if (resourceTemplateRefs.get() != null) {
             return;
         }
@@ -399,7 +426,7 @@ public class DefaultMcpClient implements McpClient {
         JsonNode result = null;
         CompletableFuture<JsonNode> resultFuture = null;
         try {
-            resultFuture = transport.executeOperationWithResponse(operation);
+            resultFuture = transport.executeOperationWithResponse(new McpCallContext(invocationContext, operation));
             result = resultFuture.get(timeoutMillis, TimeUnit.MILLISECONDS);
             resourceTemplateRefs.set(ResourcesHelper.parseResourceTemplateRefs(result));
         } catch (ExecutionException | InterruptedException | TimeoutException e) {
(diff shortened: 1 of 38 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpCallContext.java
```diff
@@ -0,0 +1,13 @@
...
+import dev.langchain4j.mcp.protocol.McpClientMessage;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Context information for any invocation made towards an MCP server.
+ *
+ * It contains the AI service invocation context when this is during
+ * an AI service invocation (in other cases, the invocation context is null).
+ */
+public record McpCallContext(@Nullable InvocationContext invocationContext, McpClientMessage message) {}
```
dliubarskyi, line 13:
```suggestion
public record McpRequestContext(@Nullable InvocationContext invocationContext, McpClientMessage message) {}
```
Not sure about this option either, but it seems that there should also be an `McpResponseContext`, in this case "request"/"response" would be clear

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
```diff
@@ -22,28 +23,55 @@ public interface McpClient extends AutoCloseable {
      */
     List<ToolSpecification> listTools();
 
+    /**
+     * Obtains a list of tools from the MCP server.
+     */
+    List<ToolSpecification> listTools(InvocationContext invocationContext);
```
dliubarskyi, line 29: consider making new methods `default` and delegating to existing methods to not break things

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpHeadersSupplier.java
```diff
@@ -0,0 +1,10 @@
+package dev.langchain4j.mcp.client;
+
+import java.util.Map;
+import java.util.function.Function;
+
+/**
+ * A functional interface that supplies HTTP headers for MCP client requests
+ * based on the given {@link McpCallContext}.
+ */
+public interface McpHeadersSupplier extends Function<McpCallContext, Map<String, String>> {}
```
dliubarskyi, line 11:
```suggestion
@FunctionalInterface
public interface McpHeadersSupplier extends Function<McpCallContext, Map<String, String>> {}
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
```diff
@@ -0,0 +1,31 @@
+package dev.langchain4j.mcp.client;
+
+import dev.langchain4j.service.tool.ToolExecutionResult;
+
+public interface McpListener {
+
+    default void beforeToolCall(McpCallContext context) {}
+    ;
```
dliubarskyi: please remove all ";" in this class

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
```diff
@@ -0,0 +1,31 @@
+package dev.langchain4j.mcp.client;
+
+import dev.langchain4j.service.tool.ToolExecutionResult;
+
+public interface McpListener {
```
dliubarskyi:
```suggestion
public interface McpClientListener {
```
to be more specific

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
```diff
@@ -0,0 +1,31 @@
...
+
+    default void beforeToolCall(McpCallContext context) {}
```
dliubarskyi:
```suggestion
    default void beforeExecuteTool(McpCallContext context) {}
```
Worth keeping same name as is used in `McpClient`?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
```diff
@@ -0,0 +1,31 @@
...
+    ;
+
+    default void afterToolCall(McpCallContext context, ToolExecutionResult result) {}
```
dliubarskyi:
```suggestion
    default void afterToolCall(McpResponseContext context, ToolExecutionResult result) {}
```
Since you expose low-level `McpClientMessage` before request, would it make sense to expose some kind of low-level MCP "response message" inside the `McpResponseContext` as well?

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpHeadersTestBase.java
```diff
@@ -1,39 +1,156 @@
...
+                .aiMessage(AiMessage.from(ToolExecutionRequest.builder()
+                        .name("echoHeader")
+                        .arguments("{\"headerName\": \"X-Test-Header\"}")
+                        .build()))
+                .build();
+
+        // when no tools are requested, return the tool requesting response
+        //        Mockito.when(mockChatModel.chat(Mockito.argThat((ArgumentMatcher<ChatRequest>) chatRequest ->
+        // chatRequest != null && chatRequest.toolSpecifications().isEmpty()))).thenReturn(toolRequestingResponse);
+        Mockito.when(mockChatModel.chat(Mockito.any(ChatRequest.class))).thenAnswer((i) -> {
```
dliubarskyi: nit: you might want to use `ChatModelMock.thatAlwaysResponds(AiMessage...)` instead

dliubarskyi (commented):
@jmartisk thank you!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpCallContext.java
jmartisk: Ive' been thinking about this too, but it's not really a "request", or is it? It also applies to notifications (like responding to pings...) and a "request" sounds too much HTTP-bound, whereas this can be a simple message sent over STDIO.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
jmartisk: This pattern is generally applied to methods that take fewer parameters than the existing one (so the default implementation can call the existing method by replacing some parameters with null), but this is the opposite case, so it feels unnatural to me. The default implementation would have to take an `InvocationContext` and call the existing one by DROPPING a parameter, so while that wouldn't technically break the API, it would lead to unexpected behavior where it would ignore a value that you pass to it.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
jmartisk: For tools, resources and prompts, we have the specific classes `ToolExecutionResult`, `McpReadResourceResult`, `McpGetPromptResult`. I would say it makes sense to deserialize to that, rather than give the raw JSON? We don't have any such class for a response message that covers all kinds of messages, so we would otherwise have to give the raw `JsonNode`

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
jmartisk: Oops, good catch

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpHeadersTestBase.java
jmartisk: I think would need `ChatModelMock.thatAlwaysResponds(Function<InvocationOnMock, AiMessage>)` to be able to do that, it does not simply return a fixed message. If you want, I could potentially add that...

jmartisk:
I've applied some of your suggestions, some are still open

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpCallContext.java
dliubarskyi: In a client-server architecture I guess request/response are the usual terms, no?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
dliubarskyi: But why dropping is bad in this case? If an implementor/users of a custom `McpClient` interface are not interested in invocation context, so be it, but the existing functionality will still work for them after your change.

But I guess there are not many custom `McpClient` impls (if any), so it is not critical for sure

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
dliubarskyi: I would probably go with the lowest-level possible, as we might not have all the response exposed in the `ToolExecutionResult`/etc. If we provide a raw response, users will be able to get anything that interests them.

`JsonNode` is Jackson-specific, perhaps we can convert JsonNode into a recursive `Map<String, Object>`?

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpHeadersTestBase.java
dliubarskyi: It's up to you

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
jmartisk: I  mean that if an author of a `McpClient` implementation 'forgets' to override this method (because nothing will require him to) then users of that implementation may wrongly assume that the `InvocationContext` is important, but it will actually be ignored. And yeah, I don't expect any custom `McpClient` implementations to exist anyway.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpCallContext.java
jmartisk: The MCP specification refers to these as  "messages", and these messages have three types: request, notification, response. So in the client context, this message could be a request or a notification. That's why I would prefer to refrain from calling it "request"... `McpMessageContext` could potentially work, but it does not convey the fact that it's something being transferred to the server... ugh :(

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
jmartisk: In that case we should make public all the helper methods for converting between `Map` and `JsonNode` (and vice versa) and between `JsonNode` and `McpResource`, `McpReadResourceResult`, `McpResourceTemplate`, `McpPrompt`, `McpGetPromptResult` because users will want to access the higher-level representations too.. Ok?

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpHeadersTestBase.java
jmartisk: Ok I did that

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClientListener.java
```diff
@@ -0,0 +1,24 @@
...
+
+import dev.langchain4j.service.tool.ToolExecutionResult;
+
+public interface McpClientListener {
+
+    default void beforeExecuteTool(McpCallContext context) {}
+
+    default void afterExecuteTool(McpCallContext context, ToolExecutionResult result) {}
+
+    default void onToolCallError(McpCallContext context, Throwable error) {}
```
dliubarskyi:
```suggestion
    default void onExecuteToolError(McpCallContext context, Throwable error) {}
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
dliubarskyi: @jmartisk I guess we can add this in the future when/if it will be required

jmartisk: I've now changed it to `Map<String, Object>` without making the utilities public...

dliubarskyi: Sorry, I guess I expressed myself unclearly, I meant that **_in addition_** to the parsed `ToolExecutionResult` (or any other response/result high-level type) we could include the raw response

jmartisk: Ah ok, I didn't understand you properly. But yeah, that makes sense, let me do that.

jmartisk: Done now. I've also squashed to just two commits.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -250,45 +263,90 @@ public ToolExecutionResult executeTool(ToolExecutionRequest executionRequest) {
...
-            return ResourcesHelper.parseResourceContents(result);
+            McpReadResourceResult resourceResult = ResourcesHelper.parseResourceContents(result);
+            if (listener != null) {
+                listener.afterResourceGet(
+                        context, resourceResult, (Map<String, Object>) ToolExecutionHelper.toObject(result));
+            }
+            return resourceResult;
         } catch (ExecutionException | InterruptedException | TimeoutException e) {
+            if (listener != null) {
+                listener.onResourceGetError(context, e);
```
dliubarskyi, line 348: Listener is called for `InterruptedException`, but when tools are executed, `InterruptedException` are not propagated into the listener. Let's make it consistent?

jmartisk: Good catch. I believe the listener should NOT be called anymore after an `InterruptedException`, so I've changed that.

dliubarskyi (approved):
@jmartisk thank you!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpListener.java
dliubarskyi: Thanks!

PR #4480, opened 2026-01-27
Missing interrupt status handling in RetryUtils (issue #4479)

### LINKED ISSUE #4479: [BUG] Missing interrupt status handling in RetryUtils
**Describe the bug**
RetryUtils fails to propagate the InterruptedException, causing the loop to continue retrying.

**Log and Stack trace**
N/A

**To Reproduce**
```
        ExecutorService executor = Executors.newSingleThreadExecutor();
        Future<?> future = executor.submit(() -> {
            model.chat("how are you"); // any model that uses RetryUtils
        });

        try {
            future.get(2, TimeUnit.SECONDS); // time out exceeds
        } catch (TimeoutException e) {
            future.cancel(true); // try to interrupt thread
        } catch (ExecutionException | InterruptedException e) {
        } finally {
            executor.shutdownNow();
        }
```

**Expected behavior**
I expect the thread to be interrupted and no new retries to be made after calling future.cancel(true).

**Expected fix:**
adding this inside while loop
```
  if (Thread.currentThread().isInterrupted()) {
            throw new RuntimeException("Execution cancelled by user");
        }
```

- LangChain4j version: Latest main branch
- LLM(s) used: N/A
- Java version:  21

**Additional context**
You can see now spring Retryable  handles such behaviour, it stops retries and throws BackOffInterruptedException

isergeymd:
## Issue

Closes #4479

## Change
I've added check if thread isInterrupted and rethrowing exception

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
@@ -164,12 +164,12 @@ public final class RetryUtils {
          *
          * @param retry The retry number.
          */
-        @JacocoIgnoreCoverageGenerated
         public void sleep(int retry) {
             try {
                 Thread.sleep(jitterDelayMillis(retry));
-            } catch (InterruptedException ignored) {
-                // pass
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("Interrupted during sleep", e);
             }
         }
 
@@ -205,6 +205,10 @@ public final class RetryUtils {
                 } catch (NonRetriableException e) {
                     throw e;
                 } catch (Exception e) {
+                    if (Thread.currentThread().isInterrupted()) {
+                        throw new RuntimeException("Interrupted during action execution", e);
+                    }
+
                     if (retry >= maxRetries) {
                         throw e instanceof RuntimeException re ? re : new LangChain4jException(e);
                     }
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
@@ -169,7 +169,7 @@ public final class RetryUtils {
                 Thread.sleep(jitterDelayMillis(retry));
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
-                throw new RuntimeException("Interrupted during sleep", e);
+                throw new RuntimeException("Interrupted while sleep", e);
             }
         }
 
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
@@ -169,7 +169,7 @@ public final class RetryUtils {
                 Thread.sleep(jitterDelayMillis(retry));
             } catch (InterruptedException e) {
                 Thread.currentThread().interrupt();
-                throw new RuntimeException("Interrupted while sleep", e);
+                throw new RuntimeException("Interrupted while retrying", e);
             }
         }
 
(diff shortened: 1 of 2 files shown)


langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
```diff
@@ -164,12 +164,12 @@ public int jitterDelayMillis(int retry) {
...
          */
-        @JacocoIgnoreCoverageGenerated
         public void sleep(int retry) {
             try {
                 Thread.sleep(jitterDelayMillis(retry));
-            } catch (InterruptedException ignored) {
-                // pass
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("Interrupted during sleep", e);
```
dliubarskyi:
```suggestion
                throw new RuntimeException("Interrupted while retrying", e);
```

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
```diff
@@ -205,6 +205,10 @@ public <T> T withRetry(Callable<T> action, int maxRetries) {
                 } catch (NonRetriableException e) {
                     throw e;
                 } catch (Exception e) {
+                    if (Thread.currentThread().isInterrupted()) {
+                        throw new RuntimeException("Interrupted during action execution", e);
```
dliubarskyi, line 209: Please explain why is this change needed?
It is also not testep, please add a test if required

isergeymd: In this line, we break from the while loop to stop retrying (the main point of this PR). We handle the Exception here and check .isInterrupted() because it is the only way to do it; for example, JdkHttpClient wraps InterruptedException in a RuntimeException

should_not_retry_after_interruption - here is the test

dliubarskyi: If you remove this part and run `should_not_retry_after_interruption`, it will pass. So either there is something wrong with the implementation, or with test, please check.

I guess we can indeed remove this part and rely only on the check inside the `sleep` method. If `sleep` method throws an exception, you break out of `while` loop.

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
```diff
@@ -164,12 +164,12 @@ public int jitterDelayMillis(int retry) {
...
          */
-        @JacocoIgnoreCoverageGenerated
         public void sleep(int retry) {
             try {
                 Thread.sleep(jitterDelayMillis(retry));
-            } catch (InterruptedException ignored) {
-                // pass
+            } catch (InterruptedException e) {
+                Thread.currentThread().interrupt();
+                throw new RuntimeException("Interrupted while sleep", e);
```
dliubarskyi:
```suggestion
                throw new RuntimeException("Interrupted while retrying", e);
```
I am not sure "sleep" is the most important part here

langchain4j-core/src/main/java/dev/langchain4j/internal/RetryUtils.java
isergeymd: I think it needs to be tested differently... Theoretically, it is possible to interrupt the thread in both cases: while sleeping and during action.call() (for example, in a JdkHttpClient sync call).

isergeymd: i've added send test to cover both cases.

Here why we should not remove this part with .isInterrupted()
1) Prevents misleading "retriable error" logs
2) Checking the interrupt flag is a lightweight memory check. It is significantly more efficient than relying on Thread.sleep() to trigger a costly native exception

dliubarskyi (approved):
@isergeymd thank you!

PR #4486, opened 2026-01-27
MCP: Replace raw Thread() with ExecutorService in StdioMcpTransport

### LINKED ISSUE #4485: MCP: Replace raw Thread() with ExecutorService in StdioMcpTransport
## Description

The `StdioMcpTransport` class currently uses raw `new Thread()` calls to start background threads for handling I/O operations. This approach has several drawbacks:

1. **No thread management**: Raw threads are not tracked or managed, making it difficult to control their lifecycle
2. **Resource leaks**: Threads are not properly shut down when the transport is closed
3. **No virtual thread support**: Misses the opportunity to leverage virtual threads on Java 21+ for better scalability

## Current Code (before fix)

```java
// FIXME: where should we obtain the thread?
new Thread(jsonRpcIoHandler).start();
stderrHandler = new ProcessStderrHandler(process);
new Thread(stderrHandler).start();
```

## Proposed Solution

Replace raw `new Thread()` calls with `ExecutorService` using the existing `VirtualThreadUtils` utility class:

- Use `VirtualThreadUtils.createVirtualThreadExecutor()` to automatically leverage virtual threads on Java 21+
- Fall back to a fixed thread pool with daemon threads on earlier Java versions
- Add proper `ExecutorService.shutdown()` in the `close()` method with graceful termination
- Consistent with the pattern already used in `DefaultMcpClient`

## Benefits

- **Better thread management**: ExecutorService provides proper lifecycle control
- **Virtual thread support**: Automatic use of virtual threads on Java 21+ for improved scalability
- **Graceful shutdown**: Proper termination handling prevents thread leaks
- **Consistency**: Follows the same pattern used elsewhere in the codebase

cc @dliubarskyi

farzad-sedaghatbin:
## Issue
Closes #4485

## Change
Replace raw `new Thread()` calls with `ExecutorService` in `StdioMcpTransport` for better thread management and virtual thread support.

**Changes made:**
- Replace raw `new Thread()` calls with `ExecutorService` for proper thread lifecycle management
- Use `VirtualThreadUtils.createVirtualThreadExecutor()` to leverage virtual threads on Java 21+
- Fall back to fixed thread pool with daemon threads on earlier Java versions
- Add proper `ExecutorService.shutdown()` in `close()` method with graceful termination (5 second timeout)
- Remove the FIXME comment about thread management

This follows the same pattern used in `DefaultMcpClient` for consistency.

## General checklist

cc @dliubarskyi


jmartisk:
I guess the most proper solution would be to introduce an SPI for frameworks like Quarkus to be able to provide their own supplier for the executor. Quarkus will probably want to use its regular worker thread pool; WDYT, @geoand?

geoand:
Yup, that's the way to go

farzad-sedaghatbin:
Thanks for the great feedback @jmartisk and @geoand! 

I've implemented the SPI approach as suggested. The changes now include:

**New SPI Interface:**
- `McpExecutorServiceSupplier` - allows frameworks to provide custom ExecutorService implementations
- Discovered via `ServiceLoader.load(McpExecutorServiceSupplier.class)`
- Includes documentation with Quarkus example

**Updated StdioMcpTransport:**
- Uses ServiceLoader to discover custom implementations first
- Falls back to VirtualThreadUtils (virtual threads on Java 21+ or daemon thread pool) if no SPI found  
- Added debug logging to show which executor is being used

This allows Quarkus to register its managed executor by implementing the SPI interface and adding it to `META-INF/services/dev.langchain4j.mcp.client.transport.McpExecutorServiceSupplier`.

The approach maintains backward compatibility while providing the extensibility frameworks need. Let me know if you'd like any adjustments to the interface or implementation!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdio/StdioMcpTransport.java
```diff
@@ -116,9 +123,42 @@ public void close() throws IOException {
...
+            executorService.shutdown();
+            try {
+                if (!executorService.awaitTermination(5, TimeUnit.SECONDS)) {
+                    executorService.shutdownNow();
+                }
+            } catch (InterruptedException e) {
+                executorService.shutdownNow();
+                Thread.currentThread().interrupt();
+            }
+        }
```
geoand: This should not be done as it's not the library's responsibility to close a executor obtained from some other piece of the code

jmartisk: Yeah if the executor is coming from Quarkus, we definitely must not shut it down. This should only happen when the executor was created by langchain4j.

farzad-sedaghatbin: Excellent catch @geoand and @jmartisk! You're absolutely right - we shouldn't shut down ExecutorServices that we didn't create.

I've fixed this by:

1. **Tracking ownership**: Added a `shouldShutdownExecutorService` flag to track whether we created the executor or got it via SPI
2. **SPI executors**: When obtained via SPI (e.g., from Quarkus), we set `shouldShutdown = false` since the framework manages it
3. **Our executors**: When we create the default executor (virtual threads or fixed pool), we set `shouldShutdown = true`
4. **Safe cleanup**: The `close()` method now only calls `shutdown()` if `shouldShutdownExecutorService` is true

This ensures framework-provided executors remain under their management while still properly cleaning up resources we create ourselves.

Thanks for the thorough review!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdio/StdioMcpTransport.java
```diff
@@ -116,9 +126,56 @@ public void close() throws IOException {
...
         process.destroy();
     }
 
+    private static ExecutorServiceInfo createExecutorService() {
+        // First, try to load a custom ExecutorService via SPI
+        // This allows frameworks like Quarkus to provide their own managed executor
+        ServiceLoader<McpExecutorServiceSupplier> loader = ServiceLoader.load(McpExecutorServiceSupplier.class);
```
dliubarskyi: I am not sure about using an SPI for getting the executor, in all other placess we accept an optional `Executor` instance when building objects, for example in `ToolSerivce`:
```java
public void executeToolsConcurrently(Executor executor) {
        this.executor = getOrDefault(executor, ToolService::defaultExecutor);
    }
```

If none is provided explicitly, we fallback to `DefaultExecutorProvider.getDefaultExecutorService()`.

What do you think about this approach?

farzad-sedaghatbin: You're absolutely right @dliubarskyi! The SPI approach was inconsistent with the existing langchain4j patterns. Thank you for pointing this out.

I've updated the implementation to follow the same builder pattern used throughout the codebase:

**New approach (consistent with ToolService pattern):**
```java
// In Builder class
public Builder executorService(ExecutorService executorService) {
    this.executorService = executorService;
    return this;
}

// In constructor  
this.executorService = getOrDefault(builder.executorService, DefaultExecutorProvider::getDefaultExecutorService);
this.shouldShutdownExecutorService = builder.executorService == null;

jmartisk: Hmm that's a good point, I don't know why I didn't realize it before :)

jmartisk (approved):

PR #4487, opened 2026-01-27
feat(anthropic): add support for caching UserMessage

### LINKED ISSUE #1591: [FEATURE] Anthropic: prompt caching
Implement [prompt caching](https://www.anthropic.com/news/prompt-caching) in Anthropic integration

hetvi3012:
## Issue

Closes #1591 

## Change

This PR adds support for prompt caching in UserMessage for the Anthropic integration.

Previously, prompt caching was only available for SystemMessage and ToolSpecification. This change allows users to mark a UserMessage for caching by adding a specific attribute.

Key changes:

Updated AnthropicMapper to check for the cache_control attribute in UserMessage.

If the attribute is set to "ephemeral", the cache_control parameter is injected into the Anthropic API request (applied to the last content block of the message, as per Anthropic's requirements).

Added unit tests in AnthropicMapperTest to verify that the attribute is correctly mapped to the Anthropic request object.

Usage:
UserMessage userMessage = UserMessage.from("Hello cached world");
userMessage.attributes().put("cache_control", "ephemeral");

## General checklist


Bram-- (changes requested):
Thanks for your PR.

Can you please revert unrelated changes? It's difficult to see what the actual change is, furthermore a lot of comments have been removed or added and shouldn't be. For example:

`// do not change, will break backward compatibility!` <- Don't remove

Same on line L91: `/ ignore, it is handled in the "toAnthropicSystemPrompt" method`

Changes:

```diff
-                } else if (message instanceof AiMessage aiMessage) {
+                } else if (message instanceof AiMessage) {
+                    AiMessage aiMessage = (AiMessage) message;
```
The previous version was actually a lot better.

There's a lot more changes in the file that relate to adding `CACHE_CONTROL` which seems it should just be (ignoring tests) an few line change.

hetvi3012:
> Thanks for your PR.
> 
> Can you please revert unrelated changes? It's difficult to see what the actual change is, furthermore a lot of comments have been removed or added and shouldn't be. For example:
> 
> `// do not change, will break backward compatibility!` <- Don't remove
> 
> Same on line L91: `/ ignore, it is handled in the "toAnthropicSystemPrompt" method`
> 
> Changes:
> 
> ```diff
> -                } else if (message instanceof AiMessage aiMessage) {
> +                } else if (message instanceof AiMessage) {
> +                    AiMessage aiMessage = (AiMessage) message;
> ```
> 
> The previous version was actually a lot better.
> 
> There's a lot more changes in the file that relate to adding `CACHE_CONTROL` which seems it should just be (ignoring tests) an few line change.

Done! I've reverted the unrelated changes (pattern matching, text blocks, and comments) in both the Mapper and the Test file. The PR now only contains the logic for UserMessage caching and the new test case.

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
```diff
@@ -119,30 +120,52 @@ private static AnthropicToolResultContent toAnthropicToolResultContent(ToolExecu
...
+        List<AnthropicMessageContent> anthropicContents = new ArrayList<>();
+
+        // 2. Iterate with an index so we can identify the LAST item
+        for (int i = 0; i < contents.size(); i++) {
+            dev.langchain4j.data.message.Content content = contents.get(i);
+
+            boolean isLastItem = (i == contents.size() - 1);
+            boolean applyCache = shouldCache && isLastItem;
+
+            // 3. Convert content (Using the modern syntax the maintainer wants)
```
Bram--: Please remove :)

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
```diff
@@ -119,30 +120,52 @@ private static AnthropicToolResultContent toAnthropicToolResultContent(ToolExecu
...
+                && "ephemeral".equals(message.attributes().get(CACHE_CONTROL));
+
+        List<dev.langchain4j.data.message.Content> contents = message.contents();
+        List<AnthropicMessageContent> anthropicContents = new ArrayList<>();
+
+        // 2. Iterate with an index so we can identify the LAST item
+        for (int i = 0; i < contents.size(); i++) {
+            dev.langchain4j.data.message.Content content = contents.get(i);
+
+            boolean isLastItem = (i == contents.size() - 1);
```
Bram--, line 134: Here is where I would expect an implementation comment; Why is this important? Why ONLY the last item?

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicMapperTest.java
```diff
@@ -409,6 +410,38 @@ void should_not_extract_server_tool_results_when_disabled() {
         assertThat(aiMessage.attributes()).doesNotContainKey(SERVER_TOOL_RESULTS_KEY);
     }
 
+    @Test
+    void should_map_user_message_with_cache_control_metadata() {
```
Bram--, line 417: If the it being the last item is important, can you please add a test for this to ensure that is respected?

langchain4j-anthropic/src/main/java/dev/langchain4j/model/anthropic/internal/mapper/AnthropicMapper.java
```diff
@@ -119,30 +120,52 @@ private static AnthropicToolResultContent toAnthropicToolResultContent(ToolExecu
...
-                            return AnthropicPdfContent.fromUrl(pdfFile.url().toString());
-                        }
-                        return AnthropicPdfContent.fromBase64(
-                                pdfFile.mimeType(), ensureNotBlank(pdfFile.base64Data(), "base64Data"));
-                    } else {
-                        throw illegalArgument("Unknown content type: " + content);
-                    }
-                })
-                .collect(toList());
+        // 1. Check if the user asked for caching
```
Bram--: Nitpick; but personally I wouldn't leave an comment here, comments should only, generally, only describe very complex bits of code OR the why.

langchain4j-anthropic/src/test/java/dev/langchain4j/model/anthropic/AnthropicMapperTest.java
```diff
@@ -409,6 +410,38 @@ void should_not_extract_server_tool_results_when_disabled() {
...
+        // When
+        List<AnthropicMessage> anthropicMessages = toAnthropicMessages(singletonList(userMessage));
+
+        // Then
+        assertThat(anthropicMessages).hasSize(1);
+        AnthropicMessage message = anthropicMessages.get(0);
+
+        assertThat(message.content).hasSize(1);
+
+        // 1. Get the content as the generic interface
```
Bram--: Same here as above about comments.

dliubarskyi:
There are some conflicts now, could you please check?

Bram-- (changes requested):
Thanks! 😄

hetvi3012:
> There are some conflicts now, could you please check?

I have resolved the merge conflicts with main.

To align with the recent refactoring in AnthropicMapper, I updated my changes to use IntStream instead of a loop, which keeps the style consistent while preserving the logic (applying cache_control only to the last item).

I also added the implementation comment explaining the prefix-based caching strategy and included the requested test case to verify it.

hetvi3012:
> Thanks! 😄

I've resolved the conflicts and refactored the mapper to use the new Stream style while preserving the caching logic. I also added the implementation comment and the requested test case.

Bram-- (approved):
Looks great, thanks!

hetvi3012:
Hi @dliubarskyi , just a quick update: I’ve resolved the conflicts and addressed Bram's feedback (added the streaming logic and the 'last-item' test case).

dliubarskyi:
@hetvi3012 thank you! Please document this feature under https://docs.langchain4j.dev/integrations/language-models/anthropic#caching in a separate PR

dliubarskyi:
@Bram-- thank you for reviewing!

dliubarskyi:
@hetvi3012 could you please remind why we went for this option (setting attribute on the `UserMessage`) instead of having a configuration parameter (e.g., `cacheUserMessages`) that would work similarly to `cacheTools` and `cacheSystemMessages`?

marcobalcon:
Hi,

I'm using Langchain with Anthropic in Spring Boot (langchain4j-anthropic-spring-boot-starter ).

I wanted to test the cache. The guide says this:

`
UserMessage userMessage = UserMessage.from("Hello cached world");
userMessage.attributes().put("cache_control", "ephemeral");
`

but since I have something like that 

`
@UserMessage("""
            User message: {{text}}           
            """)
    Result<AuthenticationResult> authenticate(
            @MemoryId String memoryId,
            @V("text") String text,
            );
`

I don't think I can follow this instruction.

After an interaction on Discord with the team it seems that an idea could be to introduce a cacheLastUserMessages property on the AnthropicChatModel's builder that would set the cache_control attribute automatically.

@hetvi3012 What do you think? Thanks!

hetvi3012:
> @hetvi3012 could you please remind why we went for this option (setting attribute on the `UserMessage`) instead of having a configuration parameter (e.g., `cacheUserMessages`) that would work similarly to `cacheTools` and `cacheSystemMessages`?
Hi Dmytro,

The main reasoning behind using attributes on the UserMessage rather than a global cacheUserMessages configuration was to provide granular control over cache breakpoints.

Anthropic allows a maximum of 4 cache breakpoints per request. If we implemented a global toggle like cacheUserMessages, we would risk:

Inefficiency: Caching every user message in a long conversation would quickly exhaust the 4-slot limit or lead to unnecessary overhead/costs if the user only wanted to cache a specific "system-heavy" block of data within a message.

Precision: By using attributes, the developer can explicitly mark the exact point where the prefix should be cached (e.g., after a large document or a set of instructions), which is often the most cost-effective way to use Anthropic's caching.

hetvi3012:
> Hi,
> 
> I'm using Langchain with Anthropic in Spring Boot (langchain4j-anthropic-spring-boot-starter ).
> 
> I wanted to test the cache. The guide says this:
> 
> `UserMessage userMessage = UserMessage.from("Hello cached world"); userMessage.attributes().put("cache_control", "ephemeral");`
> 
> but since I have something like that
> 
> `@UserMessage(""" User message: {{text}} """) Result<AuthenticationResult> authenticate( @MemoryId String memoryId, @V("text") String text, );`
> 
> I don't think I can follow this instruction.
> 
> After an interaction on Discord with the team it seems that an idea could be to introduce a cacheLastUserMessages property on the AnthropicChatModel's builder that would set the cache_control attribute automatically.
> 
> @hetvi3012 What do you think? Thanks!

Hi Marco,

You are absolutely right. With the current implementation, there is a gap when using AI Services (declarative interfaces) because the UserMessage object is abstracted away, leaving you with no way to manually inject the cache_control attribute.

The current workaround would be to drop down to the low-level ChatLanguageModel API to manually construct the UserMessage, but I agree that's not ideal for a Spring Boot setup where you want to keep the clean interface-based approach.

I think your suggestion of adding a property like cacheLastUserMessages to the AnthropicChatModel builder is a great solution. It would allow the model to automatically handle the attribute injection behind the scenes.

One thing to keep in mind is Anthropic's limit of 4 cache breakpoints. A builder-level property that simply flags the last message (or the system message + last message) would cover the majority of use cases without hitting those limits.

I'm definitely in favor of this—it makes the caching feature much more "plug-and-play" for the Spring Boot ecosystem!

PR #4497, opened 2026-01-29
Allow to create LangChain4j contents from a path + agentic test for pdf ingestion

### LINKED ISSUE #3832: [FEATURE] Add support for non-text user messages in Agentic API
I need to pass PDF document for analysis (for data extraction of PFDs with complex layout where simple text extraction loses a good deal of structure and precision) along with instructions (system message). I am using xAI Grok via OpenAI langchain4j interface
I tried to use PdfFile or PdfFileContent as parameter in my agentic interface but they were treated as text and simply embedded as their toString() represenattion in TextContent user message
After everything failed I also tried to create a Tool that would return PdfFileContent but result was the same

I would like Agentic API to fully support non text user messages including file upload and base64 encoded delivery and also support Tools returning non-text (file upload/base64) responses 

I had to stick with PDF text extraction and loss of precision

**Additional context**
OpenAI supports several file processing techniques - two step - file upload/reference and base64 encoded content:
https://platform.openai.com/docs/guides/pdf-files?api-mode=responses
Ideally support for both as base 
for example 
```
"input": [
            {
                "role": "user",
                "content": [
                    {
                        "type": "input_file",
                        "filename": "draconomicon.pdf",
                        "file_data": "...base64 encoded PDF bytes here..."
                    },
                    {
                        "type": "input_text",
                        "text": "What is the first dragon in the book?"
                    }
                ]
            }
        ]
```

mariofusco:
## Issue

Closes #3832

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/internal/ContentUtil.java
```diff
@@ -0,0 +1,20 @@
+package dev.langchain4j.data.message;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Base64;
+
+public class ContentUtil {
```
dliubarskyi, line 8: Please move it to `dev.langchain4j.internal` package

dliubarskyi (commented):
@mariofusco thanks!

langchain4j-core/src/main/java/dev/langchain4j/internal/ContentUtil.java
mariofusco: Done.

dliubarskyi (approved):
@mariofusco thank you!

PR #4500, opened 2026-01-29
Exceptions during AiServiceEvents are eaten

### LINKED ISSUE #4499: [BUG] Exceptions during AiServiceEvents are eaten
In `DefaultAiServiceListenerRegistrar`, when a listener is fired, if that listener errors with an exception, the exception is logged but not re-thrown. This is a bug.

Anything downstream should be able to catch any exceptions thrown by the event listener.

**NOTE:** The fix will be backwards-compatible from a compilation standpoint, but will introduce a difference in functionality.

edeandrea:
Add support for exception handling configuration in `AiServiceListenerRegistrar`, enabling optional propagation of errors during event firing

Fixes #4499

## Issue

Closes #

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/observability/api/AiServiceListenerRegistrar.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/observability/api/AiServiceListenerRegistrar.java
@@ -1,10 +1,10 @@
 package dev.langchain4j.observability.api;
 
+import java.util.Arrays;
+import java.util.ServiceLoader;
 import dev.langchain4j.observability.api.event.AiServiceEvent;
 import dev.langchain4j.observability.api.listener.AiServiceListener;
 import dev.langchain4j.spi.observability.AiServiceListenerRegistrarFactory;
-import java.util.Arrays;
-import java.util.ServiceLoader;
 
 /**
  * A registrar for registering {@link AiServiceListener}s.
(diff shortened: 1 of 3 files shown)


edeandrea:
There's no breaking behavioral change here like I thought there might be. Whether exceptions are thrown or eaten can be configured on a per-registrar basis, which ends up as a per ai-service basis.

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
```diff
@@ -24,6 +25,9 @@ public class DefaultAiServiceListenerRegistrar implements AiServiceListenerRegis
     private final Map<Class<? extends AiServiceEvent>, EventListeners<? extends AiServiceEvent>> listeners =
             new ConcurrentHashMap<>();
 
+    // Defaults to false to preserve backwards compatibility
+    private final AtomicBoolean shouldThrowExceptionOnEventError = new AtomicBoolean(false);
```
mariofusco, line 29: I'm not understanding why you're using an `AtomicBoolean` here. Can you please explain?

langchain4j-core/src/main/java/dev/langchain4j/observability/api/AiServiceListenerRegistrar.java
```diff
@@ -83,17 +83,49 @@ default void unregister(Iterable<? extends AiServiceListener<?>> listeners) {
...
+     * Configures whether exceptions should be thrown when an error occurs during event processing.
+     * If set to {@code true}, any error that occurs while processing an event will result in an exception
+     * propagating to the caller. If set to {@code false}, errors will be silently handled or logged
+     * without interrupting the normal execution flow.
+     *
+     * @param shouldThrowExceptionOnEventError Indicates whether to throw exceptions on event errors.
+     *                                         If {@code true}, exceptions will be thrown; otherwise, errors
+     *                                         will be handled silently.
+     */
+    void shouldThrowExceptionOnEventError(boolean shouldThrowExceptionOnEventError);
```
dliubarskyi, line 96: Please mention default value in the javadoc

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
edeandrea: Because multiple threads could possible be registering listeners and/or changing this value.

langchain4j-core/src/main/java/dev/langchain4j/observability/api/AiServiceListenerRegistrar.java
edeandrea: Done

edeandrea:
I'm not sure why the spotless check is failing. I ran spotless on the `langchain4j-core` module (the only one I made changes in) locally and it doesn't find anything that needs to be changed
.

mariofusco (approved):

PR #4507, opened 2026-01-30
Mcp listeners error handling

jmartisk:
More proper error handling for MCP listeners

- Call `onPromptGetError` when prompt retrieval fails
- Call `onResourceGetError` when resource retrieval fails
- Call `afterExecuteTool` instead of `onExecuteToolError` when a tool fails with an application-level error
- Rename `McpClientListenerTest` to `McpClientListenerIT` for consistency
- Javadoc for `McpClientListener`


dliubarskyi (approved):
@jmartisk thank you!

PR #4508, opened 2026-01-30
Tools: Support Optional parameter as alternative to @P(required)

### LINKED ISSUE #4279: [FEATURE] Use of `Optional` instead (or as alternative) of `@P(required)`

Currently if a tool argument is optional you use the `@P` annotation `@P(value = "Unit of temperature", required = false) String unit`. Maybe another option could be `@P("...") Optional<String> unit`

Using the `@P(required)` annotation field

linnaung:
## Issue

Closes #4279 

## Change
Implement Optional parameter as alternative to `@P(required)`

## General checklist


dliubarskyi:
@linnaung thank you, but it seems that `DefaultToolExecutor` should be adjusted to handle `Optional` as well. We need to make sure `Optional.empty()` is created when LLM does not provide the argument, and that `Optional.of(...`) is created when it does

linnaung:
> ...it seems that `DefaultToolExecutor` should be adjusted to handle `Optional` as well. We need to make sure `Optional.empty()` is created when LLM does not provide the argument, and that `Optional.of(...`) is created when it does

thanks, re-iterated.

dliubarskyi (approved):
@linnaung thank you!

PR #4512, opened 2026-01-31
feat: add support for Google Search grounded image generation in Gemi…

### LINKED ISSUE #4511: [FEATURE] Support Google Search grounding when using Gemini 3 Pro Image preview (aka Nano Banana Pro)
Gemini 3 Pro Image preview (_"Nano Banana Pro"_) supports grounding image generation in Google Search results. So for example, you can say, generate an illustration of the weather in a given city, and the model will issue a Google Search to find the current weather in that city, and will generate an image accordingly.

glaforge:
Implements Google Search grounding for image generation, when using Gemini 3 Pro Preview (aka _"Nano Banana Pro"_)

Fixes #4511


glaforge:
cc @Bram-- and @dliubarskyi

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiImageModel.java
Bram--: Nit; Might as well extract this to a private method that takes content. That way the building is only done in a single place.

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiImageModelIT.java
```diff
@@ -73,6 +74,33 @@ void should_edit_image() throws IOException {
         saveImage(editedResponse.content(), "should_edit_image_result");
     }
 
+    @Test
+    void should_ground_image_in_search() throws IOException {
```
Bram--, line 80: This test doesn't really validate the Google search is done, instead it just sets the boolean on the builder. IMHO https://github.com/glaforge/langchain4j/blob/336f9c4aae9efa3430bac4d78b9fc2863a59b532/langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiImageModelTest.java#L307 would be a better place to add a test for this. (I would also cost a lot less to run in both time and money).

Bram-- (changes requested):
Nice! 👍

langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiGeminiImageModel.java
glaforge: Method extracted and reused.
Good catch, thank you.

langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiImageModelIT.java
glaforge: I've added the assertion in `GoogleAiGeminiImageModelTest`.
Should I remove the test in the `IT` test?

glaforge: Now that the grounding metadata is properly passed around, I've expanded the assertions to check the groundedness of the image generation.

Bram-- (approved):
Looks great now, thanks!

PR #4523, opened 2026-02-02
Some more MCP client fixes

jmartisk:
- Don't shut down the executor from DefaultExecutorProvider: The problem is that the `DefaultExecutorProvider` returns a single shared instance so if we close it, it breaks any other stuff in the app that might want to use it
- In the WebSocket MCP transport, we erroneously treat some operations as "wait for response"  when they don't expect any response. This causes a resource leak, and it potentially breaks client-initiated operations that happen to have the same ID as something that was a server-initiated operation but a response to it was registered here.


dliubarskyi (approved):
@jmartisk thanks a lot!

PR #4522, opened 2026-02-02
Revert "MCP: Replace raw Thread() with ExecutorService in StdioMcpTransport"

dliubarskyi:
Reverts langchain4j/langchain4j#4486

PR #4526, opened 2026-02-02
Fix/mcp executor proper lifecycle

### LINKED ISSUE #4485: MCP: Replace raw Thread() with ExecutorService in StdioMcpTransport

farzad-sedaghatbin:
## Issue
Closes #4485

## Change
Fix ExecutorService lifecycle management in `StdioMcpTransport` to resolve integration test failures that caused revert in PR #4522.

### Problem
The FIXME comment left after revert #4522 used a blanket `shouldShutdownExecutorService = false` approach, which addressed CI failures but wasn't the proper solution. The root cause was multiple transport instances sharing the same `DefaultExecutorProvider.getDefaultExecutorService()` singleton, and the first instance shutting it down caused subsequent instances to fail with `RejectedExecutionException`.

### Solution
Replace the blanket approach with smart logic:
```java
// Only shutdown executor if user explicitly provided one, never shutdown the shared default executor
this.shouldShutdownExecutorService = builder.executorService != null && 
        builder.executorService != DefaultExecutorProvider.getDefaultExecutorService();

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdio/StdioMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/stdio/StdioMcpTransport.java
@@ -45,9 +45,9 @@ public class StdioMcpTransport implements McpTransport {
         this.logger = builder.logger;
         this.executorService =
                 getOrDefault(builder.executorService, DefaultExecutorProvider::getDefaultExecutorService);
-        // FIXME: are there actually any cases where we should shut down the executor service?
-        // the DefaultExecutorProvider always returns a single shared instance, so we can't shut it down
-        this.shouldShutdownExecutorService = false;
+        // Only shutdown executor if user explicitly provided one, never shutdown the shared default executor
+        this.shouldShutdownExecutorService = builder.executorService != null
+                && builder.executorService != DefaultExecutorProvider.getDefaultExecutorService();
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


farzad-sedaghatbin:
@dliubarskyi
@jmartisk

jmartisk:
I would say the only case where the executor should be shut down is if the `StdioMcpTransport` created it for itself. Which is currently never the case. If the user created their own executor, they should be responsible for shutting it down too. 

I would probably leave it as it is now.

dliubarskyi:
@farzad-sedaghatbin I will then close this PR, thanks!

PR #4529, opened 2026-02-02
Unify Batch API Response Model

Bram--:
## Issue
Progresses #3916

### Summary

This PR refactors the batch processing response model to use a single unified `BatchResponse<T>` class instead of the previous sealed interface hierarchy (`BatchSuccess`, `BatchIncomplete`, `BatchError`). This simplifies the API and provides a more consistent experience across different batch model implementations.  Will update docs and examples in separate PRs.

### Breaking Changes

**Batch Models affected:**
- `BatchChatModel` (moved to `dev.langchain4j.model.chat`)
- `BatchEmbeddingModel` (moved to `dev.langchain4j.model.embedding`)
- `BatchImageModel` (moved to `dev.langchain4j.model.image`)

**Google AI Gemini (@Experimental) implementations:**
The existing Gemini batch implementations are updated to reflect these new interfaces and the unified `BatchResponse` model.

#### Migration Guide

**Before:**
The API used a sealed interface hierarchy (`BatchSuccess`, `BatchIncomplete`, etc.) and a shared `retrieveBatchResults` method.

```java
BatchResponse<ChatResponse> response = batchModel.retrieveBatchResults(batchName);

if (response instanceof BatchSuccess<ChatResponse> success) {
    List<ChatResponse> results = success.responses();
} else if (response instanceof BatchIncomplete) { ... }
```

**After:**
The API uses a unified `BatchResponse<T>` with helper methods and specialized `submit`/`retrieve` methods. Batch identifiers are now wrapped in `BatchId`.

```java
  // Using the new BatchRequest wrapper and submit method
  String id = batchModel.submit(new BatchRequest<>(chatRequests)).batchId();

  // Polling with the new retrieve method
  BatchResponse<ChatResponse> response = batchModel.retrieve(id);

  if (response.state() == BatchState.SUCCEEDED) {
      List<ChatResponse> results = response.responses(); // Results list
  } else if (!response.state().isTerminal()) {
      // Still running
  } else if (response.state() == BatchState.FAILED) {
      List<BatchError> errors = response.errors(); // Detailed error info
  }
```

### Changes

- **Decoupled Interfaces**: Created `BatchChatModel`, `BatchEmbeddingModel`, and `BatchImageModel` in their respective packages.
- **Unified `BatchResponse<T>`**: Replaced the sealed hierarchy with a single class.
- **New `BatchRequest<T>`**: Introduced a wrapper for requests to allow for future job-level parameters (e.g., Gemini's display name) without breaking method signatures.
- **New Value Objects**:
    - `BatchId`: Type-safe identifier for batch jobs.
    - `BatchPage<T>`: Standardized record for paginated results in `list()` operations.
    - `BatchError`: Detailed error tracking including codes and provider-specific metadata.
- **`BatchState` Enum**: Explicit lifecycle tracking (`PENDING`, `RUNNING`, `SUCCEEDED`, `FAILED`, `CANCELLED`, `EXPIRED`).

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchJobState.java
```diff
@@ -0,0 +1,14 @@
+package dev.langchain4j.model.batch;
+
+/**
+ * Represents the possible states of a batch job.
+ */
+public enum BatchJobState {
+    BATCH_STATE_PENDING,
```
dliubarskyi:
```suggestion
    PENDING,
```
Do we need to repeat `BATCH_STATE_` every time?

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+    void cancelBatchJob(BatchName name);
+
+    /**
+     * Lists batch jobs with optional pagination.
+     *
+     * @param pageSize  the maximum number of batch jobs to return; if null, uses server default
+     * @param pageToken token for retrieving a specific page; if null, returns the first page
+     * @return a {@link BatchList} containing batch responses and pagination information
+     */
+    BatchList<RES> listBatchJobs(@Nullable Integer pageSize, @Nullable String pageToken);
```
dliubarskyi:
```suggestion
    BatchList<RES> listJobs(@Nullable Integer pageSize, @Nullable String pageToken);
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Cancels a batch operation that is currently pending or running.
+     *
+     * <p>Cancellation may not be immediate; the batch may transition through intermediate
+     * states before fully cancelling.</p>
+     *
+     * @param name the batch identifier to cancel
+     */
+    void cancelBatchJob(BatchName name);
```
dliubarskyi:
```suggestion
    void cancelJob(BatchName name);
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+     * Creates a batch of requests and submits them for asynchronous processing.
+     *
+     * <p>The returned {@link BatchResponse} typically represents an incomplete state
+     * (e.g., pending or running). Use {@link #retrieveBatchResults(BatchName)} to poll
+     * for completion and obtain results.</p>
+     *
+     * @param requests the list of requests to process in the batch
+     * @return a {@link BatchResponse} representing the initial state of the batch operation
+     */
+    BatchResponse<RES> createBatch(List<REQ> requests);
```
dliubarskyi:
```suggestion
    BatchResponse<RES> submit(List<REQ> requests);
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: sorted

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: Done

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchJobState.java
Bram--: I personally like having the prefix in enums as it allows for clearer static imports e.g. using `BATCH_STATE_PENDING` instead of `PENDING` or `BatchJobState.PENDING`. But you're probably right, the batch prefix is pretty redundant everywhere - removed from interface and enum.

Bram--:
@dliubarskyi Friendly ping.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchChatModel.java
```diff
@@ -0,0 +1,11 @@
+package dev.langchain4j.model.batch;
```
dliubarskyi: I am not entirely sure, but I guess that for the end user, the importance of model characteristics is in this order:
1. Model type (chat, embedding, image, etc.)
2. API type (single, batch)

I would therefore probably avoid grouping them all under the `model.batch` package and instead move them to the corresponding packages (e.g., `model.chat.batch`). WDYT?

Bram--: That makes sense. Keeping the generic interface in `model.batch`?

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+     * Creates a batch of requests and submits them for asynchronous processing.
+     *
+     * <p>The returned {@link BatchResponse} typically represents an incomplete state
+     * (e.g., pending or running). Use {@link #retrieveResults(BatchName)} to poll
+     * for completion and obtain results.</p>
+     *
+     * @param requests the list of requests to process in the batch
+     * @return a {@link BatchResponse} representing the initial state of the batch operation
+     */
+    BatchResponse<RES> submit(List<REQ> requests);
```
dliubarskyi: I am wondering if it makes sense to future-proof and introduce a mirroring `BatchRequest` type that would allow to specify not only requests, but other potential parameters applicable to all requests but that are not necessarily belong to request. For example, job ID/name. Maybe I am owerthinking it

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Retrieves the current state and results of a batch operation.
+     *
+     * <p>The responses indicates whether the batch is still processing, completed successfully,
```
dliubarskyi:
```suggestion
     * <p>The responses indicate whether the batch is still processing, completed successfully,
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchEmbeddingModel.java
```diff
@@ -0,0 +1,11 @@
+package dev.langchain4j.model.batch;
+
+import dev.langchain4j.Experimental;
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.data.segment.TextSegment;
+
+/**
+ * A batch model for processing multiple embedding requests asynchronously.
```
dliubarskyi: Perhaps the javadoc needs more clarification, as regular `EmbeddingModel` is also processing multiple embeddings at once

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchImageModel.java
```diff
@@ -0,0 +1,12 @@
...
+import dev.langchain4j.Experimental;
+import dev.langchain4j.data.image.Image;
+import dev.langchain4j.model.output.Response;
+import org.jspecify.annotations.NonNull;
+
+/**
+ * A batch model for processing multiple image generation requests asynchronously.
+ */
+@Experimental
+public interface BatchImageModel extends BatchModel<String, Response<@NonNull Image>> {}
```
dliubarskyi:
```suggestion
public interface BatchImageModel extends BatchModel<String, Response<Image>> {}
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchName.java
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Represents a unique identifier for a batch operation.
+ *
+ * <p>A batch name is typically assigned by the API when a batch is created and is used
+ * to track, retrieve, cancel, or delete the batch operation. The format of the value
+ * depends on the provider (e.g., {@code "batches/abc123"} for Gemini, {@code "msgbatch_xyz"} for Anthropic).</p>
+ *
+ * @param value the unique identifier string for the batch
+ */
+public record BatchName(String value) {}
```
dliubarskyi:
```suggestion
public record BatchId(String value) {}
```
This seems to be a more appropriate name for the class given the javadoc

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchList.java
```diff
@@ -0,0 +1,23 @@
...
+ * @param <T> the type of the responses payload in each batch (e.g., {@code List<ChatResponse>})
+ * @param batches the list of batch responses for the current page
+ * @param nextPageToken token to pass to {@link BatchModel#listJobs(Integer, String)} to
+ *                      retrieve the next page; may be {@code null} if no more pages exist
+ *
+ * @see BatchModel#listJobs(Integer, String)
+ * @see BatchResponse
+ */
+@Experimental
+public record BatchList<T>(List<BatchResponse<T>> batches, @Nullable String nextPageToken) {}
```
dliubarskyi:
```suggestion
public record BatchPage<T>(List<BatchResponse<T>> batches, @Nullable String nextPageToken) {}
```
Since this is about pagination, it seem that this is a more appropriate name, WDYT?

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * @param <REQ> the type of input request (e.g., ChatRequest, TextSegment, ImageGenerationRequest)
+ * @param <RES> the type of individual responses (e.g., ChatResponse, Embedding, Response&lt;Image&gt;)
+ *
+ * @see BatchResponse
+ * @see BatchName
+ * @see BatchList
+ */
+@Experimental
+public interface BatchModel<REQ, RES> {
```
dliubarskyi: I have a bit of doubts about squishing all model types (e.g., chat, embedding, etc) into the same API shape.
Are we sure they do (and will) behave similar enough to make them have the same API? I am talking not only about Gemini, but other model providers as well. Also thinking about the future.

I doubt many users will write model-type-agnostic code to work with batch models in the same manner (this is where this interface will help). But I am afraid of the situation when in the future batch processing of chat, embedding and image requests might diverge for whatever reason and we will be forced to maintain a zoo of models that behave differently under the same API

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
```diff
@@ -0,0 +1,50 @@
...
+     * Returns {@code true} if the batch completed successfully.
+     */
+    public boolean isSuccess() {
+        return state == SUCCEEDED;
+    }
+
+    /**
+     * Returns {@code true} if the batch failed.
+     */
+    public boolean isError() {
```
dliubarskyi:
```suggestion
    public boolean isFailed() {
```
To use consistent terminology

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
```diff
@@ -0,0 +1,50 @@
...
+     * Returns {@code true} if the batch is still processing (not in a terminal state).
+     */
+    public boolean isIncomplete() {
+        return !TERMINAL_BATCH_STATES.contains(state);
+    }
+
+    /**
+     * Returns {@code true} if the batch completed successfully.
+     */
+    public boolean isSuccess() {
```
dliubarskyi:
```suggestion
    public boolean isSucceded() {
```
To use consistent terminology

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
```diff
@@ -0,0 +1,50 @@
...
+        BatchName batchName,
+        BatchJobState state,
+        List<T> responses,
+        @Nullable List<ExtractedBatchResults.Status> errors) {
+    private static final List<BatchJobState> TERMINAL_BATCH_STATES = List.of(EXPIRED, FAILED, SUCCEEDED);
+
+    /**
+     * Returns {@code true} if the batch is still processing (not in a terminal state).
+     */
+    public boolean isIncomplete() {
```
dliubarskyi:
```suggestion
    public boolean isInProgress() {
```
WDYT?

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Cancels a batch operation that is currently pending or running.
+     *
+     * <p>Cancellation may not be immediate; the batch may transition through intermediate
+     * states before fully cancelling.</p>
+     *
+     * @param name the batch identifier to cancel
+     */
+    void cancelJob(BatchName name);
```
dliubarskyi:
```suggestion
    void cancel(BatchName name);
```
I am not sure we should mix "batch" and "job" terminology. I would stick to "batch" everywhere

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Lists batch jobs with optional pagination.
+     *
+     * @param pageSize  the maximum number of batch jobs to return; if null, uses server default
+     * @param pageToken token for retrieving a specific page; if null, returns the first page
+     * @return a {@link BatchList} containing batch responses and pagination information
+     */
+    BatchList<RES> listJobs(@Nullable Integer pageSize, @Nullable String pageToken);
```
dliubarskyi:
```suggestion
    BatchList<RES> list(@Nullable Integer pageSize, @Nullable String pageToken);
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
```diff
@@ -0,0 +1,64 @@
...
+    /**
+     * Retrieves the current state and results of a batch operation.
+     *
+     * <p>The responses indicates whether the batch is still processing, completed successfully,
+     * or failed. Clients should poll this method at intervals until the batch completes.</p>
+     *
+     * @param name the batch identifier obtained from {@link #submit(List)}
+     * @return a {@link BatchResponse} representing the current state of the batch operation
+     */
+    BatchResponse<RES> retrieveResults(BatchName name);
```
dliubarskyi:
```suggestion
    BatchResponse<RES> get(BatchName name);
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
dliubarskyi: Not sure whether we want to have minimalistic `submit/get/cancel/list` names or duplicate Batch everywhere like `submitBatch/getBatch/cancelBatch/listBatches`, WDYT?

dliubarskyi (commented):
@Bram-- thank you!

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchChatModel.java
dliubarskyi: @Bram-- I guess yes

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: All APIs I've looked at - Mistral, OpenAI, Gemini and Claude - only use a list of request as it's request obj. With the exception of Gemini which allows setting a display name. It would complicate things a bit, but I think it might be good actually.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchEmbeddingModel.java
Bram--: The embedAll method is not asynchronous tho, which is the main difference. I can add some more information in each BatchModel interface individually.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: The providers where you have a choice (Gemini, OpenAI, Bedrock and Mistral) basically let's you upload a set of messages/request objects and you then specify the endpoint it should run against. This means that all APIs are the same but only differ on request object input. 
I can sever the connection to a common BatchModel however, and just have the same for now but not enforced by it's shared interface

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchChatModel.java
Bram--: sorted.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: changed, see comment below.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchEmbeddingModel.java
Bram--: Changed the Java docs.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchName.java
Bram--: changed.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchList.java
Bram--: Changed.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: Removed.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
Bram--: now `hasFailed`

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
Bram--: now `hasSucceeded`

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
Bram--: done

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: removed job

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: made all methods on the batch models a lot shorter, see comment below.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchModel.java
Bram--: changed to `retrieve`

Bram--:
Hi @dliubarskyi ,  thanks for the review.

I've gone ahead and refactored the batch implementation based on our discussion. I think you're right about the future-proofing, it's better to decouple these now than to try and untangle a generic "zoo" later.

Here is the updated direction I've taken:

### 1. Specialized & Decoupled Interfaces
I’ve removed the shared generic `BatchModel` interface, as discussed. While they look similar now, this allows them to evolve independently. I've also moved them to their respective packages to prioritize the model type over the API type:

*   `BatchChatModel` → `dev.langchain4j.model.chat`
*   `BatchEmbeddingModel` → `dev.langchain4j.model.embedding`
*   `BatchImageModel` → `dev.langchain4j.model.image`

### 2. Standardized API Shape
Each of these interfaces now follows this consistent method signature:
*   `submit(BatchRequest<T> requests)`: Submits the batch.
*   `retrieve(BatchId name)`: Polls for status/results.
*   `cancel(BatchId name)`: Stops a running job.
*   `list(Integer pageSize, String pageToken)`: For paginated job history.

### 3. Future-proofed `BatchRequest`
I took your advice on `BatchRequest`. Instead of just passing a `List`, I've introduced it as a wrapper class. This allows us to easily add job-level metadata (like that Gemini "display name" or a custom job ID) in the future without breaking the API.

### 4. New Core components
I've introduced several new classes in `dev.langchain4j.model.batch` to handle the shared logic:
*   **`BatchId`**: Stronger typing for those provider-specific strings.
*   **`BatchResponse<T>`**: Wraps the results and state. I added some helpers like `isInProgress()` and `hasSucceeded()` to make the user-side polling logic much cleaner.
*   **`BatchPage<T>`**: Standardized pagination record.
*   **`BatchError`**: For detailed failure info.
*   **`BatchState`**: Enum for the job lifecycle (`PENDING`, `RUNNING`, `SUCCEEDED`, `FAILED`, etc.).

Let me know if this looks more like what you had in mind!

docs/docs/integrations/embedding-models/google-ai-gemini.md
```diff
@@ -236,17 +236,17 @@ BatchResponse<Embedding> initialResponse = batchModel.createBatchInline(
 );
 
 // Extract the batch name for polling
-BatchName batchName = switch (initialResponse) {
-    case BatchIncomplete incomplete -> incomplete.batchName();
-    case BatchSuccess success -> success.batchName();
+BatchName batchId = switch (initialResponse) {
```
dliubarskyi:
```suggestion
BatchId batchId = switch (initialResponse) {
```

docs/docs/integrations/embedding-models/google-ai-gemini.md
```diff
@@ -208,7 +208,7 @@ BatchResponse<Embedding> response = batchModel.createBatchInline("My Batch", nul
 switch (response) {
     case BatchIncomplete incomplete -> {
         System.out.println("Batch is " + incomplete.state());
-        System.out.println("Batch name: " + incomplete.batchName().value());
+        System.out.println("Batch name: " + incomplete.batchId().value());
```
dliubarskyi:
```suggestion
        System.out.println("Batch ID: " + incomplete.batchId().value());
```

docs/docs/integrations/embedding-models/google-ai-gemini.md
```diff
@@ -263,10 +263,10 @@ if (result instanceof BatchSuccess success) {
 **Cancel a batch job:**
 
 ```java
-BatchName batchName = // ... obtained from createBatchInline or createBatchFromFile
+BatchName batchId = // ... obtained from createBatchInline or createBatchFromFile
```
dliubarskyi:
```suggestion
BatchId batchId = // ... obtained from createBatchInline or createBatchFromFile
```

dliubarskyi: Please rename it everywhere, there are some `BatchName` instances left

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchError.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.Map;
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Represents an error status for a failed request within a batch operation.
+ *
+ * @param code    the error code (e.g., HTTP status code or provider-specific error code)
+ * @param message a human-readable description of the error
+ * @param details optional additional details about the error, typically containing
+ *                provider-specific diagnostic information.qsw l;eswwwwwwwwwwwwwwwwwwwwwwww
```
dliubarskyi:
```suggestion
 *                provider-specific diagnostic information.
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchError.java
```diff
@@ -0,0 +1,15 @@
...
+ */
+public record BatchError(int code, String message, @Nullable List<Map<String, Object>> details) {}
```
dliubarskyi: Please don't use records for public APIs, they are strict, adding a new component changes constructor signature, this breaks binary and source compatibility, etc. Let's use regular classes

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchId.java
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Represents a unique identifier for a batch operation.
+ *
+ * <p>A batch ID is typically assigned by the API when a batch is created and is used
+ * to track, retrieve, cancel, or delete the batch operation. The format of the value
+ * depends on the provider (e.g., {@code "batches/abc123"} for Gemini, {@code "msgbatch_xyz"} for Anthropic).</p>
+ *
+ * @param value the unique identifier string for the batch
+ */
+public record BatchId(String value) {}
```
dliubarskyi: I am not sure this value class is very useful

langchain4j-core/src/main/java/dev/langchain4j/model/chat/BatchChatModel.java
```diff
@@ -0,0 +1,62 @@
...
+    /**
+     * Retrieves the current state and results of a chat batch operation.
+     *
+     * <p>The response indicates whether the batch is still processing, completed successfully,
+     * or failed. Clients should poll this method at intervals until the batch completes.</p>
+     *
+     * @param name the batch identifier obtained from {@link #submit(BatchRequest)}
+     * @return a {@link BatchResponse} representing the current state of the chat batch operation
+     */
+    BatchResponse<ChatResponse> retrieve(BatchId name);
```
dliubarskyi:
```suggestion
    BatchResponse<ChatResponse> retrieve(BatchId batchId);
```
Please update it everywhere

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchRequest.java
```diff
@@ -0,0 +1,46 @@
...
+public class BatchRequest<T> {
+    private final List<T> requests;
+
+    /**
+     * Creates a new BatchRequest.
+     *
+     * @param requests The list of requests. Must not be null.
+     */
+    public BatchRequest(List<T> requests) {
+        this.requests = Objects.requireNonNull(requests);
```
dliubarskyi:
```suggestion
        this.requests = ensureNotEmpty(requests, "requests");
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
```diff
@@ -0,0 +1,112 @@
...
+ * Represents the responses of a batch operation.
+ *
+ * <p>A batch response contains the batch identifier, current state, and optionally the results
+ * when the batch has completed successfully.</p>
+ *
+ * @param <T> the type of the responses payload (e.g., {@code ChatResponse}, {@code Embedding})
+ */
+@Experimental
+public class BatchResponse<T> {
+    private static final List<BatchState> TERMINAL_BATCH_STATES = List.of(EXPIRED, FAILED, SUCCEEDED);
```
dliubarskyi: nit: this seems more like a property of batch state: `BatchState.isTerminal()`

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private final BatchId batchId;
+    private final BatchState state;
+    private final List<T> responses;
+    private final List<BatchError> errors;
+
+    public BatchResponse(
+            BatchId batchId, BatchState state, @Nullable List<T> responses, @Nullable List<BatchError> errors) {
+        this.batchId = Objects.requireNonNull(batchId, "batchId cannot be null");
```
dliubarskyi:
```suggestion
        this.batchId = ensureNotBlank(batchId, "batchId");
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
```diff
@@ -0,0 +1,112 @@
...
+        this.state = Objects.requireNonNull(state, "state cannot be null");
```
dliubarskyi:
```suggestion
        this.state = ensureNotNull(state, "state");
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
```diff
@@ -0,0 +1,112 @@
...
+        this.responses = responses;
```
dliubarskyi:
```suggestion
        this.responses = copy(responses);
```

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
```diff
@@ -0,0 +1,112 @@
...
+        this.errors = errors;
```
dliubarskyi:
```suggestion
        this.errors = copy(errors);
```

langchain4j-core/src/main/java/dev/langchain4j/model/embedding/BatchEmbeddingModel.java
```diff
@@ -0,0 +1,62 @@
...
+    void cancel(BatchId name);
+
+    /**
+     * Lists embedding batch jobs with optional pagination.
+     *
+     * @param pageSize  the maximum number of batch jobs to return; if null, uses server default
+     * @param pageToken token for retrieving a specific page; if null, returns the first page
+     * @return a {@link BatchPage} containing embedding batch responses and pagination information
+     */
+    BatchPage<Embedding> list(@Nullable Integer pageSize, @Nullable String pageToken);
```
dliubarskyi:
```suggestion
    BatchPage<Embedding> list(@Nullable BatchPagination pagination);
```
WDYT about wrapping `pageSize` and `pageToken` inside the `BatchPagination` type? This will allow to extend it and add more parameters in the future if needed (e.g., filters, etc)

dliubarskyi: (same for all other batch model types)

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchId.java
dliubarskyi: I would remove it TBH

dliubarskyi (commented):
@Bram-- thank you!

dliubarskyi:
Sorry for the delay with the review... It is already pretty solid and very close to being merged!

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
Bram--: Why make a defensive copy here? As this is a batch response it's potentially very large (Claude allows requests upto 100k messages / 256 MB).

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchError.java
Bram--: :joy_cat:

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchId.java
Bram--: As this is an unique identifier returned from the API I think it has value wrapping it. It self documents the API as the BatchId is returned from the API it's clear only the ID can be used, there's no way to pass in a wrong parameter - especially if/when additional String params are added. Finally having a record wrap a class is virtually free.

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
Bram--: same as above

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchResponse.java
dliubarskyi: Makes sense, did not think about it

langchain4j-core/src/main/java/dev/langchain4j/model/batch/BatchId.java
dliubarskyi: > It self documents the API

The name of the parameter is already self-documenting, no?

> there's no way to pass in a wrong parameter

I doubt this is something that happens reasonably often to justify a dedicated type. If it were used in many places - ok, but it is used only to get the job result and cancel the job...

> Finally having a record wrap a class is virtually free.

It adds one more class to the application, I would not call it virtually free


In most cases I imagine users will be submitting a batch, getting the ID, persisting it somewhere (e.g., DB) and shutting down. Then running a periodic job to check the state. In this scenario it will require converting `String` to `BatchId` back and forth...

dliubarskyi:
Hi @Bram--, do you plan to finalize this PR? There are a few minor changes left and we could merge it

Bram--:
> Hi @Bram--, do you plan to finalize this PR? There are a few minor changes left and we could merge it

Yes, been extraordinarily busy at work, will try to get around to finish this the coming week. Any chance on having a look at #4593 ?

dliubarskyi:
@Bram-- completely understandable, thanks in advance!

Bram--:
Sorted, I ran spotless on core as well and it changed somethings around

Bram--:
Sorted. I ran spotless on both modules and regrettably it did change some unchanged files.

Bram--:
@dliubarskyi Please take another look. Thanks

Bram--:
@dliubarskyi Please advise on the errors. Seems to be a plugin complaining about adding new public APIs?

dliubarskyi:
@Bram-- yes, now we automatically check for breaking API changes. If there is a good reason for a breaking change, you can add an exclusion in revapi config

Bram--:
@dliubarskyi Ready for review

dliubarskyi (approved):
@Bram-- thanks a lot, great job!

PR #4489, opened 2026-01-28
Add Azure OpenAI Responses API models and reasoning summaries

aavetis:
## Issue
Closes #3637  
Related: #3252, #3816, #3817

## Change
- Added Azure OpenAI Responses API chat + streaming models to surface reasoning summaries when the service emits reasoning output items.
- Added graceful handling for missing summaries (thinking stays null).
- Added a debug log when reasoning tokens exist but summary text is absent (only if `logRequestsAndResponses=true`).
- Added unit + IT coverage and updated Azure OpenAI docs with Responses API usage and summary caveat.
- Examples PR: https://github.com/langchain4j/langchain4j-examples/pull/185

## General checklist
- [X] There are no breaking changes (API, behaviour)
- [X] I have added unit and/or integration tests for my change
- [X] The tests cover both positive and negative cases
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [X] I have manually run all the unit and integration tests in the core and main modules, and they are all green
- [X] I have added/updated the documentation
- [X] I have added an example in the examples repo (only for "big" features)
- [ ] I have added/updated Spring Boot starter(s) (if applicable) 

## Checklist for adding new maven module
- [ ] I have added my new module in the root pom.xml and langchain4j-bom/pom.xml — N/A (no new module)

## Checklist for adding new embedding store integration
- [ ] I have added a {NameOfIntegration}EmbeddingStoreIT ... — N/A (no embedding store changes)
- [ ] I have added a {NameOfIntegration}EmbeddingStoreRemovalIT ... — N/A (no embedding store changes)

## Checklist for changing existing embedding store integration
- [ ] I have manually verified that the {NameOfIntegration}EmbeddingStore works correctly ... — N/A (no embedding store changes)

Issue #3637, opened 2025-09-03
[BUG] Azure module does not support returning think (reasoning content)


Lzhaohui0808:
**Describe the bug**

When using the langchain4j-azure-open-ai module, the Azure API responses that contain a think / reasoning field are currently ignored.

In contrast, the langchain4j-open-ai module already supports parsing and exposing these fields.

This inconsistency makes it impossible to access reasoning content when using Azure OpenAI Service.

**Log and Stack trace**

<img width="1793" height="511" alt="Image" src="https://github.com/user-attachments/assets/6897746f-daad-416d-9962-016cdd720ffc" />

ChatResponse { aiMessage = AiMessage { text = "你好呀！👋 很高兴见到你～有什么我可以帮你的吗？无论是学习、工作上的问题，还是想聊聊有趣的话题，我都在这里哦！😊", thinking = null, toolExecutionRequests = [], attributes = {} }, metadata = ChatResponseMetadata{id='663438a8b8e042368ad3d749ce0bd062', modelName='deepseek-r1', tokenUsage=TokenUsage { inputTokenCount = 7, outputTokenCount = 226, totalTokenCount = 233 }, finishReason=STOP} }

**To Reproduce**

```java
void testAzureAI() {
      ChatRequest chatRequest = ChatRequest.builder()
              .messages(UserMessage.from("hello！"))
              .build();

      ChatModel model = AzureOpenAiChatModel.builder()
              .endpoint(AZURE_ENDPOINT)
              .apiKey(AZURE_API_KEY)
              .deploymentName(AZURE_DEPLOYMENT_NAME)
              .temperature(0.0)
              .logRequestsAndResponses(true)
              .build();

      ChatResponse chatResponse = model.chat(chatRequest);
      log("text", chatResponse);
}
```

**Expected behavior**
- langchain4j-azure-open-ai should parse the think / reasoning field from Azure API responses.
- The model response should expose this content in the same way as the OpenAI module.

**Please complete the following information:**
- LangChain4j version: 1.4.0 & 1.4.0-beta10
- LLM(s) used: 
- Java version:  17 
- Spring Boot version (if applicable):  3.5.5

**Additional context**

dliubarskyi:
cc @jdubois

jdubois:
Thanks @Lzhaohui0808 ! Is it working with the OpenAI Official module -> https://docs.langchain4j.dev/integrations/language-models/open-ai-official ?
[...]

Issue #3252, opened 2025-06-23
[FEATURE] Support for the new OpenAI Responses API


SuperPat45:
**Is your feature request related to a problem? Please describe.**
OpenAI introduces a new [Responses API](https://platform.openai.com/docs/api-reference/responses) combining the simplicity of the Chat Completions API with the tool use capabilities of the deprecated Assistants API.

**Describe the solution you'd like**
Support for this API and its tools (web search, code interpreter, mcp...) in langchain4j would be great.

**Describe alternatives you've considered**

**Additional context**

stantonk:
plus one!

dliubarskyi:
Anyone wants to create a proof of concept? It should be an implementation of an existing `ChatModel` API.

Since `ChatModel` API is stateless, we should use [OpenAI's manual state management](https://platform.openai.com/docs/guides/conversation-state?api-mode=responses#manually-manage-conversation-state). Make sure to explicitly set the [`store`](https://platform.openai.com/docs/api-reference/responses/create#responses-create-store) parameter to `false`.

wesleysalimansdvb:
plus one!

maxlepikhin:
Draft PR is above, I won't be able to continue on it soon.

ejstuart:
I also have a draft PR, took me a bit longer than I would've liked 😓 (sorry about that)
Happy to have others make improvements and changes to it, going to try to keep updating it but will depend on my schedule

maxlepikhin:
@ejstuart sounds good, I forgot to include a version I wrote with langchain4j httpclient too (attached here) in case somebody finds it useful.

```java
import static dev.langchain4j.internal.JsonSchemaElementUtils.toMap;

import com.fasterxml.jackson.databind.ObjectMapper;
import dev.langchain4j.agent.tool.ToolExecutionRequest;
import dev.langchain4j.data.message.AiMessage;
import dev.langchain4j.data.message.ChatMessage;
import dev.langchain4j.data.message.SystemMessage;
import dev.langchain4j.data.message.TextContent;
import dev.langchain4j.data.message.UserMessage;
import dev.langchain4j.http.client.HttpClient;
import dev.langchain4j.http.client.HttpClientBuilder;
import dev.langchain4j.http.client.HttpMethod;
import dev.langchain4j.http.client.HttpRequest;
import dev.langchain4j.http.client.sse.DefaultServerSentEventParser;
import dev.langchain4j.http.client.sse.ServerSentEvent;
[...]

PR #4542, opened 2026-02-04
feat(open-ai-official): add reasoning summaries to Responses streaming

aavetis:
## Issue
Related to #3252, #3637, and the closed PR #4489 (this PR follows the official OpenAI SDK path instead of Azure OpenAI SDK).

## Change
- Add reasoning summary support to the OpenAI Official Responses streaming model.
- Surface summary content in `AiMessage.thinking`.
- Add integration tests for reasoning summaries (positive and negative cases, gated by `OPENAI_API_KEY`).
- Document reasoning summaries in the OpenAI Official SDK docs.

## General checklist
- [X] I have manually run all the unit and integration tests in the core and main modules, and they are all green
- [X] I have added/updated the documentation

### DIFF
--- a/langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
+++ b/langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
@@ -1,6 +1,7 @@
 package dev.langchain4j.model.openaiofficial;
 
 import static dev.langchain4j.internal.InternalStreamingChatResponseHandlerUtils.onPartialResponse;
+import static dev.langchain4j.internal.InternalStreamingChatResponseHandlerUtils.onPartialThinking;
 import static dev.langchain4j.internal.InternalStreamingChatResponseHandlerUtils.withLoggingExceptions;
 import static dev.langchain4j.internal.JsonSchemaElementUtils.toMap;
 import static dev.langchain4j.internal.Utils.copy;
@@ -114,8 +115,8 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
     private final String safetyIdentifier;
     private final String promptCacheKey;
     private final String promptCacheRetention;
-    private final String reasoningEffort;
-    private final String reasoningSummary;
+    private final ReasoningEffort reasoningEffort;
+    private final Reasoning.Summary reasoningSummary;
     private final String textVerbosity;
     private final Boolean streamIncludeObfuscation;
     private final Boolean store;
@@ -250,14 +251,13 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
                 paramsBuilder.putAdditionalBodyProperty(
                         PROMPT_CACHE_RETENTION_FIELD, JsonValue.from(promptCacheRetention));
             }
-            if ((reasoningEffort != null && !reasoningEffort.isEmpty())
-                    || (reasoningSummary != null && !reasoningSummary.isEmpty())) {
+            if (reasoningEffort != null || reasoningSummary != null) {
                 Reasoning.Builder reasoningBuilder = Reasoning.builder();
-                if (reasoningEffort != null && !reasoningEffort.isEmpty()) {
-                    reasoningBuilder.effort(ReasoningEffort.of(reasoningEffort));
+                if (reasoningEffort != null) {
+                    reasoningBuilder.effort(reasoningEffort);
                 }
-                if (reasoningSummary != null && !reasoningSummary.isEmpty()) {
-                    reasoningBuilder.summary(Reasoning.Summary.of(reasoningSummary));
+                if (reasoningSummary != null) {
+                    reasoningBuilder.summary(reasoningSummary);
                 }
                 paramsBuilder.reasoning(reasoningBuilder.build());
             }
@@ -531,8 +531,8 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
         private String safetyIdentifier;
         private String promptCacheKey;
         private String promptCacheRetention;
-        private String reasoningEffort;
-        private String reasoningSummary;
+        private ReasoningEffort reasoningEffort;
+        private Reasoning.Summary reasoningSummary;
         private String textVerbosity;
         private Boolean streamIncludeObfuscation;
         private Boolean store;
@@ -628,6 +628,11 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
             return this;
         }
 
+        public Builder modelName(ChatModel modelName) {
+            this.modelName = modelName.toString();
+            return this;
+        }
+
         public Builder temperature(Double temperature) {
             this.temperature = temperature;
             return this;
@@ -698,6 +703,12 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
         }
 
         public Builder reasoningEffort(String reasoningEffort) {
+            this.reasoningEffort =
+                    reasoningEffort == null || reasoningEffort.isEmpty() ? null : ReasoningEffort.of(reasoningEffort);
+            return this;
+        }
+
+        public Builder reasoningEffort(ReasoningEffort reasoningEffort) {
             this.reasoningEffort = reasoningEffort;
             return this;
         }
@@ -706,6 +717,13 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
          * Sets the reasoning summary mode. Accepts values like "auto", "concise", or "detailed".
          */
         public Builder reasoningSummary(String reasoningSummary) {
+            this.reasoningSummary = reasoningSummary == null || reasoningSummary.isEmpty()
+                    ? null
+                    : Reasoning.Summary.of(reasoningSummary);
+            return this;
+        }
+
+        public Builder reasoningSummary(Reasoning.Summary reasoningSummary) {
             this.reasoningSummary = reasoningSummary;
             return this;
         }
@@ -766,6 +784,7 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
         private final Map<String, Integer> toolCallIndices = new HashMap<>();
         private final List<ToolExecutionRequest> completedToolCalls = new ArrayList<>();
         private final StringBuilder textBuilder = new StringBuilder();
+        private final StringBuilder reasoningSummaryBuilder = new StringBuilder();
         private OpenAiOfficialTokenUsage tokenUsage;
         private String responseId;
         private String finishReason;
@@ -788,9 +807,9 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
             }
 
             try {
-                // TODO: process reasoning events, specifically ResponseReasoningSummaryTextDoneEvent and
-                //   ResponseReasoningTextDoneEvent. It makes little sense to map them to onPartialThinking,
-                //   need reasoning callbacks.
+                // Note: reasoning summary stream events (reasoning_summary_*) are mapped to onPartialThinking
+                // because LangChain4j currently exposes a single "thinking" streaming callback.
+                // TODO: consider mapping other reasoning-related stream events once we have richer callbacks.
                 if (event.isCreated()) {
                     handleCreated(event.asCreated());
                 } else if (event.isInProgress()) {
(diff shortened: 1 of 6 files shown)


langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -114,6 +115,7 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
     private final String promptCacheKey;
     private final String promptCacheRetention;
     private final String reasoningEffort;
```
jdubois: Can we uses com.openai.models.ReasoningEffort instead of a String? This is what we do in OpenAiOfficialImageModel for the size and quality.

jdubois: Same as above, can we use a com.openai.models.Reasoning.Summary instead of a String?

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -983,6 +1004,50 @@ private void extractTokenUsageAndComplete(com.openai.models.responses.Response r
...
+
+        private static String extractReasoningSummary(com.openai.models.responses.Response response) {
+            if (response == null) {
+                return null;
+            }
+            List<ResponseOutputItem> outputItems;
+            try {
+                outputItems = response.output();
+            } catch (RuntimeException ex) {
+                return null;
```
jdubois, line 1071: I don't like to ignore exceptions, can we log it at least?

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -983,6 +1004,50 @@ private void extractTokenUsageAndComplete(com.openai.models.responses.Response r
...
+                        if (text == null || text.isBlank()) {
+                            continue;
+                        }
+                        if (!builder.isEmpty()) {
+                            builder.append("\n");
+                        }
+                        builder.append(text);
+                    }
+                } catch (RuntimeException ignored) {
+                    // Summary is optional and may be missing or malformed in some responses.
```
jdubois: Same as above, can we at least log it instead of ignoring the error totally?

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -983,6 +1004,50 @@ private void extractTokenUsageAndComplete(com.openai.models.responses.Response r
...
+            List<ResponseOutputItem> outputItems;
+            try {
+                outputItems = response.output();
+            } catch (RuntimeException ex) {
+                return null;
+            }
+            if (outputItems == null || outputItems.isEmpty()) {
+                return null;
+            }
+            StringBuilder builder = new StringBuilder();
```
jdubois: "builder" is too generic here, can you rename it to something like `responseStringBuilder`? Also with modern Java, you don't gain from using a StringBuilder, so maybe just a String would be easier.

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesStreamingChatModelIT.java
```diff
@@ -367,6 +367,53 @@ void should_support_reasoning_effort() {
         assertThat(handler.get().aiMessage().text()).contains("Paris");
     }
 
+    @Test
+    void should_support_reasoning_summary() {
+
+        // given
+        var client = OpenAIOkHttpClient.builder()
```
jdubois: Can you use the same system as in the other tests? Like in `OpenAiOfficialChatModelIT`? This will ensure it works the same everywhere, including on Microsoft Foundry

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesStreamingChatModelIT.java
```diff
@@ -367,6 +367,53 @@ void should_support_reasoning_effort() {
...
+                .apiKey(System.getenv("OPENAI_API_KEY"))
```
jdubois: Also this is copied everywhere in the test, can you have a look at how I do it in `InternalOpenAiOfficialTestHelper`? This will be more consistent, and remove this duplication

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesStreamingChatModelIT.java
```diff
@@ -443,6 +490,5 @@ protected boolean supportsPartialToolStreaming(dev.langchain4j.model.chat.Stream
 
     @Override
     @Disabled("Can't do it reliably")
-    protected void should_execute_multiple_tools_in_parallel_then_answer(StreamingChatModel model) {
-    }
+    protected void should_execute_multiple_tools_in_parallel_then_answer(StreamingChatModel model) {}
```
jdubois: I've run it a few times without any issue, is this when you run it alone, or with other tests?

jdubois (changes requested):
Most of my comments are to have something which is more coherent with the other tests. Also, it should work automatically on Microsoft Foundry if you follow what I did - in the end, OpenAI also works on the same platform, so it's just a matter of having the right URL/API key.
I did ran all the tests of Microsoft Foundry, and I only had 3 errors:
- `should_respect_modelName_in_default_model_parameters` which fails with a 404: I always struggle with that one on Microsoft Foundry, so I'm not sure about this one.
- `should_accept_multiple_images_as_public_URLs` and `should_accept_single_image_as_public_URL` but I'm pretty sure it's an issue on my end, we don't allow the public URLs anymore (not the first time I see this, so I'm not worried)

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
aavetis: `reasoningEffort` is now stored as `com.openai.models.ReasoningEffort`. I kept a small convenience overload `reasoningEffort(String)` similar to `OpenAiOfficialImageModel`’s typed builder pattern.

`reasoningSummary` is now stored as `com.openai.models.Reasoning.Summary`. Kept `reasoningSummary(String)` as a convenience wrapper for consistency with other typed builders.

Happy to remove the wrapper/overload if we want only the typed setter.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
aavetis: We now log at debug if reading `response.output()` fails during reasoning summary extraction.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
aavetis: We now log at debug the per-output-item extraction failures instead of ignoring them completely. Summary remains optional, so we still degrade gracefully to null.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
aavetis: Replaced the generic builder with a clearer approach to improve readability and avoid manual `StringBuilder` concatenation.

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesStreamingChatModelIT.java
aavetis: Refactored the Responses-streaming ITs to follow the same helper-driven setup pattern as the other OpenAI official ITs. I also validated locally on Microsoft Foundry. Using the endpoint and appending /openai/v1 worked for the Responses API (e.g. https://<resource>.services.ai.azure.com/openai/v1).

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesStreamingChatModelIT.java
aavetis: Moved repeated client/model creation into `InternalOpenAiOfficialTestHelper` and updated the tests to reuse it.

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesStreamingChatModelIT.java
aavetis: The test assumes parallel tool calls come back in a deterministic order. Instead of relying on the test’s fixed ordering (or keeping this scenario disabled), I updated `OpenAiOfficialResponsesStreamingChatModelIT.should_execute_multiple_tools_in_parallel_then_answer(...)` to keep the scenario covered but make assertions order-insensitive. We assert both `getWeather` and `getTime` tool calls are present (by name + args) and then verify the follow-up answer after providing both tool results, without relying on call order/index. This should remove intermittent failures when running the full suite.

aavetis:
Thanks for the comments. Happy to adjust as needed

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
tkumar1918: > "builder" is too generic here, can you rename it to something like `responseStringBuilder`? Also with modern Java, you don't gain from using a StringBuilder, so maybe just a String would be easier.

Since this code aggregates multiple streamed parts inside nested loops, using StringBuilder avoids creating intermediate String instances on each append. Modern Java optimizes single-expression concatenation, but not repeated += in loops. So StringBuilder is still appropriate here.

mikehearn:
I tried this out and it worked, although I found I preferred a slightly different design in which `onPartialThinking` is invoked in `handleReasoningSummaryTextDone`. In practice, OpenAI models emit quite short thinking strings so streaming them is of no great value, and it wasn't obvious to me how to know when the thinking tokens are completed otherwise. By moving where the handler is invoked you get the full string which is normally (on 5.3-codex) a short piece of Markdown, and so there's no need to accumulate tokens yourself.

mikehearn:
Actually I've now seen the model produce longer thinking strings. But with the current API there's no way to know when the model has finished emitting reasoning commentary, because there's no `onThinking` callback. So ideally we'd have both but if not then it might also be OK to just pass whole strings through to onPartialThinking. The current design is hard to understand though because when do you stop digesting the whole thought - the model is allowed to emit several different thoughts before it starts sending the final answer.

aavetis:
i'm happy to go whichever directions the team wants. given it's close to the milestone end, need some guidance here

jdubois:
@aavetis I will have time to work on this at the end of next week, and on the week after: I just realized by looking at your profile that you work at Microsoft ! Can you ping me on Teams? I couldn't find your real name, but on my side I should be easy to find... We can discuss this directly.

dliubarskyi:
@aavetis thanks a lot! Last days I've been working on improving OpenAI Responses API support in both unofficial and official modules, so there are many conflicts, sorry about that. Do you mind if I'll close this PR in favour of https://github.com/langchain4j/langchain4j/pull/4954? In addition to reasoning summaries, it also adds support for encrypted reasoning and sync model

dliubarskyi:
Closing in favour of https://github.com/langchain4j/langchain4j/pull/4954

@aavetis thank you!

PR #4544, opened 2026-02-05
Bugfix/toolid repeat:  add docs

### LINKED ISSUE #4528: [BUG] Tool Call ID incorrectly accumulated/duplicated with DeepSeek/Qwen in Streaming mode
### Summary
When using OpenAiStreamingChatModel (via OpenAI-compatible API) with DeepSeek-V3 or Qwen-2.5 models, the id of the ToolExecutionRequest is incorrectly accumulated/appended multiple times during the streaming process.

This results in a malformed ID (e.g., ididid...) in the final AiMessage. When the client attempts to send the tool execution result back to the LLM, it causes a tool_call_id mismatch error because the LLM expects the original short ID, but LangChain4j sends the duplicated long ID (or the internal checking logic fails).

**Note**: This issue does not occur with Doubao (ByteDance) models, suggesting it stems from how LangChain4j handles delta chunks from DeepSeek/Qwen, which seem to send the id field redundantly in every chunk.

**Environment**
LangChain4j version: 1.10.0 
Module: qwen-plus, deepseek-reasoner
Model Provider: DeepSeek / Qwen 
Java Version: 17

**Reproduction Steps**
Configure OpenAiStreamingChatModel pointing to a DeepSeek or Qwen endpoint.
Define many Tool (e.g.,tool1   tool2   tools).
Send a prompt that triggers these tools in streaming mode.
Observe the generated AiMessage.
Observed Behavior (The Bug)
In the debug logs, the ToolExecutionRequest inside the AiMessage has an ID that is a repetition of the original UUID.

**Example Log (Thinking/Debug):**

`{
    "thinking": "好的，我需要。。。。",
    "toolExecutionRequests": [
        {
            "id": "44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd44d5b59c6d5b496daf79a5fa923512bd",
            "name": "getToolDescriptions",
            "arguments": "{\"toolKeys\": \"travel.indicator.aced,travel.indicator.comment,report.out.component,target.out.component\"}"
        }
    ],
    "attributes": {},
    "type": "AI"
}`

`{
    "id": "44d5b59c6d5b496daf79a5fa923512bd",
    "toolName": "getToolDescriptions",
    "text": "some params",
    "type": "TOOL_EXECUTION_RESULT"
}`

chq562656437:
Problem Description

Fix an issue where tool call IDs are duplicated when using OpenAI-compatible APIs (such as DeepSeek and Qwen) with streaming responses.

Some OpenAI-compatible APIs behave differently from the official OpenAI API when streaming tool calls:

OpenAI standard behavior:
The tool call ID is sent incrementally across chunks
(e.g. chunk1 = "abc", chunk2 = "abc" → final ID = "abcabc")

Some compatible APIs behavior:
The full tool call ID is repeated in every chunk
(e.g. chunk1 = "abc", chunk2 = "abc" → expected ID = "abc", not "abcabc")

This causes tool call IDs to be incorrectly accumulated and duplicated when using APIs such as DeepSeek and Qwen.

Solution

Introduce a new configuration option accumulateToolCallId in OpenAiStreamingChatModel to control how tool call IDs are handled during streaming:

Default value: true
Preserves compatibility with the official OpenAI streaming behavior

Set to false:
Suitable for APIs like DeepSeek and Qwen, where each chunk sends the full tool call ID
In this mode, the ID from each chunk replaces the previous value instead of being accumulated

Code Example
StreamingChatModel model = OpenAiStreamingChatModel.builder()
        .baseUrl("https://api.deepseek.com/v1")
        .apiKey(apiKey)
        .modelName("deepseek-chat")
        .accumulateToolCallId(false) // Disable ID accumulation
        .build();

Changes

Added the accumulateToolCallId configuration field to OpenAiStreamingChatModel

Implemented corresponding handling logic in OpenAiStreamingResponseBuilder

Updated official documentation to explain the purpose and usage scenarios of this option

Related Issue

Fixes #4528

Documentation

Documentation has been updated:
docs/docs/integrations/language-models/openai-compatible.md

### DIFF
--- a/docs/docs/integrations/language-models/openai-compatible.md
+++ b/docs/docs/integrations/language-models/openai-compatible.md
@@ -22,6 +22,29 @@ Many services and tools expose OpenAI-compatible APIs. The general approach to u
             .logResponses(true)
             .build();
     ```
+
+### Configuration for Specific OpenAI-Compatible APIs
+
+Some OpenAI-compatible APIs may have different behaviors in streaming responses, particularly for tool calling. LangChain4j provides configuration options to handle these differences:
+
+#### `accumulateToolCallId` (for `OpenAiStreamingChatModel`)
+
+Controls how tool call IDs are handled in streaming responses. Default is `true`.
+
+- **Enabled (`true`)**: Tool call IDs are accumulated across streaming chunks (standard OpenAI behavior)
+    - Example: Chunk 1 sends "abc", Chunk 2 sends "def" → Final ID: "abcdef"
+- **Disabled (`false`)**: Each chunk's tool call ID replaces the previous one
+    - Example: Chunk 1 sends "abc", Chunk 2 sends "abc" → Final ID: "abc"
+    - Use this for APIs like DeepSeek or Qwen that send the complete tool call ID in every chunk
+
+```java
+StreamingChatModel model = OpenAiStreamingChatModel.builder()
+        .baseUrl("https://api.deepseek.com/v1") // or other provider
+        .apiKey("YOUR_API_KEY")
+        .modelName("deepseek-chat")
+        .accumulateToolCallId(false) // Set to false for DeepSeek, Qwen, etc.
+        .build();
+```
 Below we provide specific examples for popular OpenAI-compatible APIs, including Groq, Docker Model Runner, GPT4All, Ollama, and LM Studio.
 
 ### Contents:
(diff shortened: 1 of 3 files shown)


dliubarskyi (approved):
@chq562656437 thank you!

PR #4541, opened 2026-02-04
Bump com.azure:azure-sdk-bom from 1.3.3 to 1.3.4

jdubois:
This dependency update is also updating several transitive dependencies, and fixing a some security issues.

PR #4549, opened 2026-02-06
Fix the Azurite version check

jdubois:
This fixes the test issue as in the comment at https://github.com/langchain4j/langchain4j/pull/4541#issuecomment-3858625024 as #4541 is already merged


dliubarskyi (approved):
@jdubois thanks a lot!

PR #4550, opened 2026-02-06
Make MCP stdio transport's operations fail immediately when the subprocess exits

jmartisk:
See https://github.com/quarkiverse/quarkus-langchain4j/issues/2111


jmartisk:
@dliubarskyi given that this was requested by the QE team, it would be nice to have it fixed in the Quarkus 3.33 LTS stream. Quarkus 3.33 will (most likely) be on LangChain4j 1.11, so we could use a backport to a 1.11 branch and a 1.11.1 release at some point. No big rush right now, but let's keep this in mind for the future.

dliubarskyi (approved):
@jmartisk thank you!

PR #4555, opened 2026-02-08
Gemini grounding metadata not properly propagated

### LINKED ISSUE #4553: [BUG] [Gemini] Google Maps grounding not passed through metadata
**Describe the bug**

When using Google Maps grounding, the grounding metadata is not necessarily passed through and populated in the response metdata.

**To Reproduce**

Enable `allowGoogleMaps(true)` on the Gemini model, and notice that maps grounding metadata is not present in the response.

**Expected behavior**

Maps Grounding Metadata should be present.

**Additional context**

Grounding metadata can appear at the top level of the Gemini response, as well as individually at the level of a _candidate_ response. So the idea is to use the top-level metadata if present. If it's absent, check if it is available in the first candidate resposne (there's always only one candidate response)

### LINKED ISSUE #4554: [BUG] [Gemini] URL Context grounding metadata not passed through correctly
When using Google URL Context grounding, the grounding metadata is not necessarily passed through and populated in the response metdata.

**To Reproduce:**

Enable `allowUrlContext(true)` on the Gemini model, and notice that URL context grounding metadata is not present in the response.

**Expected behavior:**

URL Context Grounding Metadata should be present.

**Additional context:**

Grounding metadata can appear at the top level of the Gemini response, as well as individually at the level of a candidate response. So the idea is to use the top-level metadata if present. If it's absent, check if it is available in the first candidate response (there's always only one candidate response)

glaforge:
The Gemini URL Context and Google Maps tools were not propagating their metadata properly.

Fixes #4554
Fixes #4553


glaforge:
For your info @mohankumar27 @Bram-- and @dliubarskyi

dliubarskyi (approved):
@glaforge thanks a lot!

PR #4556, opened 2026-02-08
Feature/observability Metrics with Micrometer

### LINKED ISSUE #2329: Implement observability with Micrometer

Sus4nne:
## Issue
Closes https://github.com/langchain4j/langchain4j/issues/2329 partly
It adds metrics for the chat model interaction only.

## Change
This adds a specialised ChatModelListener which utilises a Micrometer Meterregistry to expose metrics about the ChatModelResponse/Request. It exposes metrics in line with the OpenTelemetry Semantic Conventions for Generative AI.

## General checklist

- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable) Will do later


langchain4j-micrometer-metrics/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-parent</artifactId>
+        <version>1.11.0-beta19-SNAPSHOT</version>
```
jeanbisutti: The codebase is on `1.12.0-beta20-SNAPSHOT`.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListener.java
```diff
@@ -0,0 +1,86 @@
...
+        recordTokenUsageMetrics(responseContext);
+    }
+
+    @Override
+    public void onError(final ChatModelErrorContext errorContext) {
+        // ChatModelErrorContext does not contain the ChatModelResponseContext, therefor token usage is unavailable
+    }
+
+    private String getProviderName(ChatModelResponseContext responseContext) {
+        return responseContext.modelProvider().toString().toLowerCase();
```
jeanbisutti: `responseContext.modelProvider()` can be null (see https://github.com/langchain4j/langchain4j/blob/7f6d54a879697c337fd0b04181888ef26310bef6/langchain4j-core/src/main/java/dev/langchain4j/model/chat/listener/ChatModelResponseContext.java#L30). So this code line may raise a `NullPointerException`.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListener.java
```diff
@@ -0,0 +1,86 @@
...
+                                OTelGenAiTokenType tokenType,
+                                int tokenCount) {
+
+        Counter.builder(OTelGenAiMetricName.TOKEN_USAGE.value())
+                .tag(OTelGenAiAttributes.OPERATION_NAME.value(), OTelGenAiOperationName.CHAT.value())
+                .tag(OTelGenAiAttributes.PROVIDER_NAME.value(), getProviderName(responseContext))
+                .tag(OTelGenAiAttributes.REQUEST_MODEL.value(),
+                     responseContext.chatRequest().parameters().modelName())
+                .tag(OTelGenAiAttributes.RESPONSE_MODEL.value(),
+                     responseContext.chatResponse().metadata().modelName())
```
jeanbisutti: `chatRequest().parameters().modelName()` can be null. What would happen in this case?

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListener.java
```diff
@@ -0,0 +1,86 @@
...
+
+    private void addTokenMetric(ChatModelResponseContext responseContext,
+                                OTelGenAiTokenType tokenType,
+                                int tokenCount) {
+
+        Counter.builder(OTelGenAiMetricName.TOKEN_USAGE.value())
+                .tag(OTelGenAiAttributes.OPERATION_NAME.value(), OTelGenAiOperationName.CHAT.value())
+                .tag(OTelGenAiAttributes.PROVIDER_NAME.value(), getProviderName(responseContext))
+                .tag(OTelGenAiAttributes.REQUEST_MODEL.value(),
+                     responseContext.chatRequest().parameters().modelName())
```
jeanbisutti: `responseContext.chatRequest().parameters().modelName()` can be null. What would happen in this case?

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
```diff
@@ -0,0 +1,107 @@
...
+import dev.langchain4j.micrometer.metrics.conventions.OTelGenAiTokenType;
+import dev.langchain4j.model.azure.AzureOpenAiChatModel;
+import dev.langchain4j.model.chat.request.ChatRequest;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.simple.SimpleMeterRegistry;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class ChatModelMicrometerMetricsListenerIT {
```
jeanbisutti: It does not seem an integration test. Perhaps the class name should not end with `IT`.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
```diff
@@ -0,0 +1,86 @@
...
+                responseContext.chatResponse().tokenUsage().outputTokenCount());
+    }
+
+    private void addTokenMetric(ChatModelResponseContext responseContext,
+                                OTelGenAiTokenType tokenType,
+                                int tokenCount) {
+
+        Counter.builder(OTelGenAiMetricName.TOKEN_USAGE.value())
+                .tag(OTelGenAiAttributes.OPERATION_NAME.value(), OTelGenAiOperationName.CHAT.value())
+                .tag(OTelGenAiAttributes.PROVIDER_NAME.value(), getProviderName(responseContext))
```
jeanbisutti, line 83: The provider name values do not respect the OpenTelemetry semantic conventions.

See https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/

Image

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
```diff
@@ -0,0 +1,86 @@
...
+                responseContext,
+                OTelGenAiTokenType.OUTPUT,
+                responseContext.chatResponse().tokenUsage().outputTokenCount());
+    }
+
+    private void addTokenMetric(ChatModelResponseContext responseContext,
+                                OTelGenAiTokenType tokenType,
+                                int tokenCount) {
+
+        Counter.builder(OTelGenAiMetricName.TOKEN_USAGE.value())
```
jeanbisutti: The OpenTelemetry Semantic conventions for generative AI metrics specifies an histogram, not a counter. A Micrometer `DistributionSummary` may be better.

docs/docs/tutorials/observability.md
```diff
@@ -271,149 +271,64 @@ The `attributes` map allows passing information between the `onRequest`, `onResp
...
-import dev.langchain4j.model.embedding.listener.EmbeddingModelErrorContext;
+The naming of the metrics follows the [OpenTelemetry Semantic Conventions for Generative AI Metrics](https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/). (v1.39.0)
 
-public class MyEmbeddingModelListener implements EmbeddingModelListener {
-
-    @Override
-    public void onRequest(EmbeddingModelRequestContext requestContext) {
-        requestContext.attributes().put("startNanos", System.nanoTime());
-    }
+> **⚠️ Experimental**: This module is marked as `@Experimental` and may have breaking changes in future versions.
```
jeanbisutti, line 422: I would not mark the module as experimental.

From what I know, only the HTTP and  database OpenTelemetry semantic conventions are stable.

I would just inform users to be aware of this point if they create dashboards.

docs/docs/tutorials/observability.md
```diff
@@ -271,149 +271,64 @@ The `attributes` map allows passing information between the `onRequest`, `onResp
...
+| `gen_ai.request.model`  | The model name from the request | `gpt-4`, `gpt-35-turbo` |
+| `gen_ai.response.model` | The model name from the response | `gpt-4-0613` |
+| `gen_ai.token.type`     | The type of token counted | `input`, `output` |
 
-```java
-import dev.langchain4j.store.embedding.listener.EmbeddingStoreListener;
-import dev.langchain4j.store.embedding.listener.EmbeddingStoreRequestContext;
-import dev.langchain4j.store.embedding.listener.EmbeddingStoreResponseContext;
-import dev.langchain4j.store.embedding.listener.EmbeddingStoreErrorContext;
+#### Viewing the Metrics
```
jeanbisutti: I would not mention things related to the Spring actuator in the `langchain4j-micrometer-metrics `documentation. This module is not related to Spring  and could be used in applications not based on Spring.

docs/docs/tutorials/observability.md
```diff
@@ -271,149 +271,64 @@ The `attributes` map allows passing information between the `onRequest`, `onResp
...
 
-```java
-EmbeddingModel observedModel = embeddingModel.addListener(new MyEmbeddingModelListener());
+The following metrics are currently collected:
 
-observedModel.embed("hello");
-```
+| Metric Name | Type | Description |
+|-------------|------|-------------|
+| `gen_ai.client.token.usage` | Counter | The number of tokens used by the model for input or output |
```
jeanbisutti: An important point about the `gen_ai.client.token.usage` metric is that, with the current implementation, it only takes into account the token usage from the calls of chat models. The documentation should mention it.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListener.java
```diff
@@ -0,0 +1,86 @@
...
+ * A {@link ChatModelListener} that uses a Micrometer {@link MeterRegistry} to collect metrics
+ * about chat model interactions following OpenTelemetry Semantic Conventions for Generative AI.
+ * <p>
+ * This listener records token usage metrics (input and output tokens) when a chat model response is received.
+ * <p>
+ * Note: The {@link ChatModelMicrometerMetricsListener}
+ * must be instantiated separately (e.g., via Spring Boot auto-configuration or manual instantiation).
+ */
+@Experimental
+public class ChatModelMicrometerMetricsListener implements ChatModelListener {
```
dliubarskyi:
```suggestion
public class MicrometerMetricsChatModelListener implements ChatModelListener {
```

langchain4j-micrometer-metrics/pom.xml
dliubarskyi:
```suggestion
        <version>1.12.0-beta20-SNAPSHOT</version>
```

langchain4j-micrometer-metrics/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+        <relativePath>../langchain4j-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-micrometer-metrics</artifactId>
+    <version>1.11.0-SNAPSHOT</version>
```
dliubarskyi:
```suggestion
```

docs/docs/tutorials/observability.md
dliubarskyi: I think it refers not to the OTel part, but to the `ChatModelListener` implementation itself, which is still experimental. We usually keep modules as experimental for some time before marking them stable, so I would keep it as it is

docs/docs/tutorials/observability.md
dliubarskyi: Agreed, it would make sense to move Spring-specific parts here: https://docs.langchain4j.dev/tutorials/spring-boot-integration

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
```diff
@@ -0,0 +1,107 @@
...
+                        .count())
+                .isGreaterThan(1.0);
+        assertThat(meterRegistry
+                        .get(OTelGenAiMetricName.TOKEN_USAGE.value())
+                        .tag(OTelGenAiAttributes.TOKEN_TYPE.value(), OTelGenAiTokenType.OUTPUT.value())
+                        .counter()
+                        .count())
+                .isGreaterThan(1.0);
+
+        //        assertThat(meterRegistry.find("gen_ai.client.operation.duration").timer()).isNotNull();
```
dliubarskyi: please remove dead code

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
dliubarskyi: It uses `AzureOpenAiChatModel`, so it is IT

dliubarskyi:
```suggestion
@EnabledIfEnvironmentVariable(named = "AZURE_OPENAI_KEY", matches = ".+")
class ChatModelMicrometerMetricsListenerIT {
```

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
```diff
@@ -0,0 +1,107 @@
...
+                .build();
+
+        ChatRequest chatRequest =
+                ChatRequest.builder().messages(UserMessage.from("Hi")).build();
+
+        try {
+            ChatResponse response = chatModel.chat(chatRequest);
+            assertThat(response.metadata()).isNotNull();
+        } catch (Exception e) {
+            System.out.println(e.getMessage());
```
dliubarskyi: Let's not catch exceptions, but let it fail instead

langchain4j-bom/pom.xml
```diff
@@ -583,6 +577,13 @@
                 <version>${langchain4j.beta.version}</version>
             </dependency>
 
+            <!-- observability -->
+            <dependency>
+                <groupId>dev.langchain4j</groupId>
+                <artifactId>langchain4j-micrometer-metrics</artifactId>
+                <version>${project.version}</version>
```
```suggestion
                <version>${langchain4j.beta.version}</version>
```

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
dliubarskyi:
```suggestion
class MicrometerMetricsChatModelListenerIT {
```

langchain4j-micrometer-metrics/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
+            <version>${project.version}</version>
```
dliubarskyi:
```suggestion
            <version>1.12.0-SNAPSHOT</version>
```

langchain4j-micrometer-metrics/pom.xml
```diff
@@ -0,0 +1,60 @@
...
+        <!-- For testing purposes-->
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-observation-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-azure-open-ai</artifactId>
+            <version>${project.version}</version>
```
dliubarskyi:
```suggestion
            <version>1.12.0-SNAPSHOT</version>
```

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
dliubarskyi: for the `wrongDeploymentName` case, let's better move this try-catch there:

```java
@Test
    void should_contain_metrics_with_error_on_request() {
        try {
            doChatRequest("wrongDeploymentName");
        } catch (ModelNotFoundException expected) {
            // ignored
        }
```

Ideally it should not ignore all exceptions for all cases

brunobat:
This is pretty much the same PR as in here: https://github.com/langchain4j/langchain4j/pull/2384
Copying the comment I left there:

@Sus4nne, We should move on with this.
I've done an in-depth review. The PR as is can work but I need 2 changes:

The name of the project should be langchain4j-micrometer-observation
You should have a SpringBoot IT test to make sure this will work with the framework.
Is this OK?

dliubarskyi:
@brunobat 

> This is pretty much the same PR as in here: https://github.com/langchain4j/langchain4j/pull/2384

> The name of the project should be langchain4j-micrometer-observation


AFAIU this PR does not use Observation API

brunobat:
> @brunobat
> 
> > This is pretty much the same PR as in here: #2384
> 
> > The name of the project should be langchain4j-micrometer-observation
> 
> AFAIU this PR does not use Observation API

Right, but it seems to create the exact same Metrics. 
I'd say the name should be ok then.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
brunobat: I agree. If we use the OTel name, we should be consistent with what they define. See: https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclienttokenusage

docs/docs/tutorials/observability.md
Sus4nne: Yes, that is why I marked it experimental.

docs/docs/tutorials/observability.md
Sus4nne: I think for a chat model listener, it should be clear from the name this only applies to the interaction with chat models. But I added it in the description of the metric to make this clear for the user.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListener.java
Sus4nne: Yes, I agree to change it that way. Done.

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
Sus4nne: yes. Done

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
Sus4nne: yes, I am sorry to have overlooked this

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
Sus4nne: Thank you

langchain4j-micrometer-metrics/src/test/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListenerIT.java
Sus4nne: Ah, I see what you mean. Done

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListener.java
Sus4nne: Added a null check. If the provider is not set in the responseContext, there are two choices: not record the metrics because the provider name is required, or add the metrics and add a default provider name "unknown" The Otel semcon does not define a fallback value. But a custom value is allowed. The spec states: "If one of them applies, then the respective value MUST be used; otherwise, a custom value MAY be used."
Given that gen_ai.provider.name is meant to be a discriminator for telemetry format, recording a metric with "unknown" could still be useful for debugging.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
Sus4nne: You are right. I added an enum to map the ModelProvider enum to a otel semcon compliant name, using the well-known values or a fitting custom value.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListener.java
Sus4nne: You are right. That would break it. If this value would come back as null, than we use "unknown". The same pattern as for provider name.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/ChatModelMicrometerMetricsListener.java
Sus4nne: Yes. Changed it the same as above.

jonatan-ivanov:
I would recommend using the Observation API instead of only the Metrics API.
But if you know for sure that your users will not want any other output other than metrics (I don't think this is the case), I would remove observation-test too (there is a test module for metrics).

docs/docs/tutorials/observability.md
Sus4nne: Done

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
Sus4nne: I did not see that, or understood the difference. Thank you, I learned something and it solved the simmering itch that the counter only shows the total amount of tokens used. Nice!

Sus4nne:
After all the comments on my former PR, it was decided in consultation with @jeanbisutti @SandraAhlgrimm @jdubois and @dliubarskyi to only use the Metrics API instead of the Observation API of Micrometer. We want Micrometer metrics, this PR brings this to users of not only Spring (Boot). That is why I dropped the Observation part and created a fresh PR.

@jonatan-ivanov You are right, the test dependency for observation should be deleted. I did that now. 
@brunobat I added a Spring boot IT.

Thank you all for your reviews again! I really appreciate it.

jonatan-ivanov:
There might be a misunderstanding here. The Observation API has nothing to do with Spring (Boot). Wether you instrument components using the Metrics API and/or the Observation API has not much to do with what frameworks you use or don't use. Both of the APIs work with the big frameworks or plain JVM apps without using any frameworks. If this was the reason not using the Observation API, I would recommend reconsidering that decision.

On the other hand, a big advantage of using the Observation API is that you can get not only metrics but other signals out of it as well (tracing/logging/etc). Also, using the Observation API does not mean you cannot use the Metrics API for certain things (e.g.: Gauges).

brunobat (approved):

brunobat:
We should move on and have some MM instrumentation. 
I do think we should have a `langchain4j-micrometer-observation` lib. If @Sus4nne doens't mind I can experiment with that and propose something later this week.
Will also be interesting to illustrate how the implementation differences between the 2 APIs.
How does this sound?

SandraAhlgrimm:
@Sus4nne Please drop Spring Boot from this module. Keep only Micrometer + LangChain4j deps, pin Micrometer via BOM (1.16.2), and move the Boot wiring test to the langchain4j-spring repo.
Current setup passes tests without Boot: ./mvnw -pl langchain4j-micrometer-metrics,langchain4j-core,langchain4j-azure-open-ai -am test -DskipITs.

brunobat:
> @Sus4nne Please drop Spring Boot from this module. Keep only Micrometer + LangChain4j deps, pin Micrometer via BOM (1.16.2), and move the Boot wiring test to the langchain4j-spring repo. Current setup passes tests without Boot: ./mvnw -pl langchain4j-micrometer-metrics,langchain4j-core,langchain4j-azure-open-ai -am test -DskipITs.

@dliubarskyi, @SandraAhlgrimm  Is there a way to have IT tests with SpringBoot?... We should be able to have a way to demonstrate that the Spring integration is working.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
```diff
@@ -0,0 +1,103 @@
...
+                .tag(OTelGenAiAttributes.PROVIDER_NAME.value(), getProviderName(responseContext))
+                .tag(OTelGenAiAttributes.REQUEST_MODEL.value(), getRequestModelName(responseContext))
+                .tag(OTelGenAiAttributes.RESPONSE_MODEL.value(), getResponseModelName(responseContext))
+                .tag(OTelGenAiAttributes.TOKEN_TYPE.value(), tokenType.value())
+                .description(String.format("Measures %s tokens used", tokenType.value()))
+                .publishPercentileHistogram()
+                .minimumExpectedValue(1.0)
+                .maximumExpectedValue(67108864.0)
+                .serviceLevelObjectives(
+                        1, 4, 16, 64, 256, 1024, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864)
```
jonatan-ivanov: This is usually not a good idea (min and max could be ok but `publishPercentileHistogram` and `serviceLevelObjectives` are usually not).

Users should decide on their own to enable/disable histograms, also when they do, Micrometer has a default bucketing strategy (since bucketing is rather hard).
Instrumentation should not decide about buckets, it should just record the data, users can configure on a per-Meter level if they want histogram and if so do they want to override the default buckets.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
```diff
@@ -0,0 +1,103 @@
...
+                responseContext,
+                OTelGenAiTokenType.OUTPUT,
+                responseContext.chatResponse().tokenUsage().outputTokenCount());
+    }
+
+    private void addTokenMetric(
+            ChatModelResponseContext responseContext, OTelGenAiTokenType tokenType, int tokenCount) {
+
+        DistributionSummary.builder(OTelGenAiMetricName.TOKEN_USAGE.value())
+                .baseUnit("{token}")
```
jonatan-ivanov: I would just do
```java
.baseUnit("tokens")         
```
Because of two reasons: `{token}` is not a unit and certain metrics backends encode the base unit in the meter names in which weird characters can cause trouble.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
```diff
@@ -0,0 +1,103 @@
...
+                .tag(OTelGenAiAttributes.OPERATION_NAME.value(), OTelGenAiOperationName.CHAT.value())
+                .tag(OTelGenAiAttributes.PROVIDER_NAME.value(), getProviderName(responseContext))
+                .tag(OTelGenAiAttributes.REQUEST_MODEL.value(), getRequestModelName(responseContext))
+                .tag(OTelGenAiAttributes.RESPONSE_MODEL.value(), getResponseModelName(responseContext))
+                .tag(OTelGenAiAttributes.TOKEN_TYPE.value(), tokenType.value())
+                .description(String.format("Measures %s tokens used", tokenType.value()))
```
jonatan-ivanov: I would just go with
```
Measured token usage
```
or something similar, the point here is not to encode anything dynamic into the description since that can cause problems in metrics backends (e.g.: Prometheus) where multiple time series share the same description.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
```diff
@@ -0,0 +1,103 @@
...
+        addTokenMetric(
+                responseContext,
+                OTelGenAiTokenType.OUTPUT,
+                responseContext.chatResponse().tokenUsage().outputTokenCount());
+    }
+
+    private void addTokenMetric(
+            ChatModelResponseContext responseContext, OTelGenAiTokenType tokenType, int tokenCount) {
+
+        DistributionSummary.builder(OTelGenAiMetricName.TOKEN_USAGE.value())
```
jonatan-ivanov, line 80: Fyi, Micrometer has a Meter Convention concept (see [docs](https://docs.micrometer.io/micrometer/reference/concepts/meter-convention.html)) that helps in a few things here:
1. Defining all the constant keys and values
2. Users can override the convention with their own
3. When OTel SemConv will break, a new Meter Convention that follows the new OTel SemConv can be added without breaking everyone

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
Sus4nne: This is what the Otel Semcon says: "This metric SHOULD be specified with [ExplicitBucketBoundaries](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/metrics/api.md#instrument-advisory-parameters) of [1, 4, 16, 64, 256, 1024, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216, 67108864]." https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-metrics/#metric-gen_aiclienttokenusage
That is why I did this.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
Sus4nne: The Otel semcon states Unit (UCUM) = {token}.
I took that literal. I changed it. Works probably better for aggregators

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
Sus4nne: Okay, thank you. Changed it

Sus4nne:
> We should move on and have some MM instrumentation. I do think we should have a `langchain4j-micrometer-observation` lib. If @Sus4nne doens't mind I can experiment with that and propose something later this week. Will also be interesting to illustrate how the implementation differences between the 2 APIs. How does this sound?

Well, I would be interested in that. To my understanding I had a working implementation with the Observation API. So I would like to know what you would do different, @brunobat

Sus4nne:
> @Sus4nne Please drop Spring Boot from this module. Keep only Micrometer + LangChain4j deps, pin Micrometer via BOM (1.16.2), and move the Boot wiring test to the langchain4j-spring repo. Current setup passes tests without Boot: ./mvnw -pl langchain4j-micrometer-metrics,langchain4j-core,langchain4j-azure-open-ai -am test -DskipITs.

I deleted the Springboot IT and removed all the spring boot related dependencies from the pom.

I changed the things @jonatan-ivanov pointed out. Thanks!

The Meter Convention would be a good iteration. But for now, shall we merge this?

I found the langchain4j-spring repo. The IT I created could be moved there. That is another repo, so can't be done in this PR.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
jonatan-ivanov: Micrometer supports 20+ formats/protocols (and way more backends through those protocols). OTLP is one of them. I get that the OTel Semconv says what the buckets should be (not must be) but I assume not everyone who is using the Micrometer module of Langchain4J uses OTLP. On the other hand, 100% of these users use Micrometer, so I think I would still go with the defaults and let the users choose if they want to (you can create predefined buckets that they can set on their own or a config flag, etc.)

Another interesting factor worth to be considered: this part of the OTel SemConv is unstable. When it will change and Langchain4J changes this part, that will be a breaking change for the users so it's better to use the stable default than an unstable custom value I think.

I guess an alternative solution can be using the Meter Convention concept and if the convention is set to OTel explicitly (or a config flag is set on the convention), the OTel buckets can be defined like this. This can avoid breaking the users and they can migrate to one convention to another in their own pace.

SandraAhlgrimm:
lgtm @dliubarskyi, your final review would be highly appreciated

dliubarskyi:
> Is there a way to have IT tests with SpringBoot?... We should be able to have a way to demonstrate that the Spring integration is working.

> I found the langchain4j-spring repo. The IT I created could be moved there. That is another repo, so can't be done in this PR.

Yes, it would be great to have an IT in the `langchain4j-spring` repo, reel free to open a PR there

docs/docs/tutorials/observability.md
```diff
@@ -410,10 +411,40 @@ observedRetriever.retrieve(Query.from("my query"));
...
+| `gen_ai.provider.name`  | The AI provider name | `openai`, `azure.ai.inference`, `anthropic` |
+| `gen_ai.request.model`  | The model name from the request | `gpt-4`, `gpt-35-turbo`                     |
+| `gen_ai.response.model` | The model name from the response | `gpt-4-0613`                                |
+| `gen_ai.token.type`     | The type of token counted | `input`, `output`                           |
+
 ## Observability in Spring Boot Application
 
 See more details [here](/tutorials/spring-boot-integration#observability).
 
+See more details on how to implement the ChatModelListener with Micrometer Metrics: [here](/tutorials/spring-boot-integration#Micrometer Metrics).
```
dliubarskyi:
```suggestion
See more details on how to implement the ChatModelListener with Micrometer Metrics: [here](/tutorials/spring-boot-integration#micrometer-metrics). 
```

docs/docs/tutorials/observability.md
```diff
@@ -410,10 +411,40 @@ observedRetriever.retrieve(Query.from("my query"));
   listener, as well as between multiple listeners.
 
 
+## Observability Metrics with Micrometer
+
+The `langchain4j-micrometer-metrics` module provides a Micrometer-based metrics implementation for the `langchain4j` library. Currently, it provides metrics for chat model interactions using a `ChatModelListener` implementation that collects metrics via Micrometer's `MeterRegistry`.
```
dliubarskyi:
```suggestion
The `langchain4j-micrometer-metrics` module provides a Micrometer-based metrics implementation for the LangChain4j library. Currently, it provides metrics for `ChatModel` and `StreamingChatModel` interactions using a `ChatModelListener` implementation that collects metrics via Micrometer's `MeterRegistry`.
```

dliubarskyi:
@Sus4nne thansk a lot, looks good!
I did a few minor refactorings, I hope you don't mind.

There are just 2 open discussions left:
- https://github.com/langchain4j/langchain4j/pull/4556#discussion_r2818823909
- https://github.com/langchain4j/langchain4j/pull/4556#discussion_r2818852533

While the second (convention) is probably something that can be added iteratively in the next RPs (as it requires more work), should we close this first one (remoiving `minimumExpectedValue`, `maximumExpectedValue` and `serviceLevelObjectives` settings) in this PR?

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
Sus4nne: Thank you for pointing this out. I understand it now. I removed these settings and went for default.

Sus4nne:
> @Sus4nne thansk a lot, looks good! I did a few minor refactorings, I hope you don't mind.
> 
> There are just 2 open discussions left:
> 
>     * [Feature/observability Metrics with Micrometer #4556 (comment)](https://github.com/langchain4j/langchain4j/pull/4556#discussion_r2818823909)
> 
>     * [Feature/observability Metrics with Micrometer #4556 (comment)](https://github.com/langchain4j/langchain4j/pull/4556#discussion_r2818852533)
> 
> 
> While the second (convention) is probably something that can be added iteratively in the next RPs (as it requires more work), should we close this first one (remoiving `minimumExpectedValue`, `maximumExpectedValue` and `serviceLevelObjectives` settings) in this PR?

@dliubarskyi Yes, I did what Jonatan suggested and removed the settings. I think the convention is indeed something to add iteratively. I'll close that one.

langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
Sus4nne: Yes, I guess that would be a good thing to add iteratively.

dliubarskyi (approved):
@Sus4nne well done, thanks a lot!

PR #4557, opened 2026-02-08
Refactor OpenAI Responses API integrations (follow up on PR 3816)

maxlepikhin:
Follow-up on #3816.

This PR tightens OpenAI Responses support in both official and unofficial integrations:

- keep `strict` disabled by default and align Responses-mode JSON-schema handling with actual supported behavior
- add explicit `previousResponseId` request parameters instead of shared mutable model state
- restore unofficial client logging and remove extra local `maxToolCalls` truncation
- improve streaming behavior and test coverage for reasoning/tool-call/error/completion paths
- simplify Response-related AI-service test overrides via shared hooks
- add license metadata required for `internal/langchain4j-internal-test-retry`
- fix brittle Vertex AI Anthropic integration-test initialization by moving the quota-gating condition out of a class with provider-dependent static initialization

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
```diff
@@ -223,6 +223,18 @@ public static Map<String, Object> toMap(JsonSchemaElement jsonSchemaElement, boo
         return toMap(jsonSchemaElement, strict, true);
     }
 
+    /**
+     * OpenAI Responses API validates schemas using the Structured Outputs subset:
+     * all object properties must be required and {@code additionalProperties} must be {@code false}.
+     * This helper builds a strict schema (to produce nullable types for optional fields),
+     * keeps existing {@code additionalProperties} values if present, and only adds the field when it is missing.
+     */
+    public static Map<String, Object> toMapForOpenAiResponses(JsonSchemaElement jsonSchemaElement, boolean strict) {
```
dliubarskyi: `strict` is unused

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -328,6 +351,70 @@ public ModelProvider provider() {
...
+            public void onCompleteToolCall(CompleteToolCall completeToolCall) {
+                handler.onCompleteToolCall(completeToolCall);
+            }
+
+            @Override
+            public void onCompleteResponse(ChatResponse completeResponse) {
+                if (completeResponse != null && completeResponse.metadata() != null) {
+                    String responseId = completeResponse.metadata().id();
+                    if (responseId != null && !responseId.isBlank()) {
+                        lastResponseId.set(responseId);
```
dliubarskyi: This makes it impossible to use this model concurrently. What about introducing `OpenAiOfficialResponsesChatRequestParameters` (name might be shorter lol) with `previousResponseId` field? This will allow users to supply the ID explicitly when calling the model

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -218,8 +233,14 @@ public void doChat(ChatRequest chatRequest, StreamingChatResponseHandler handler
             if (parallelToolCalls != null) {
                 paramsBuilder.parallelToolCalls(parallelToolCalls);
             }
-            if (previousResponseId != null) {
-                paramsBuilder.previousResponseId(previousResponseId);
+            String effectivePreviousResponseId = previousResponseId;
+            if (effectivePreviousResponseId == null
+                    && Boolean.TRUE.equals(store)
+                    && containsToolExecutionResult(chatRequest)) {
```
dliubarskyi: why is it conditional on `containsToolExecutionResult` ?

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
```diff
@@ -223,6 +223,18 @@ public static Map<String, Object> toMap(JsonSchemaElement jsonSchemaElement, boo
...
+        Map<String, Object> schemaMap = toMap(jsonSchemaElement, true);
+        ensureAdditionalPropertiesFalse(schemaMap);
```
dliubarskyi: maybe I am missing something, but `toMap(jsonSchemaElement, true)` already adds `additionalProperties=false`

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesAiServicesWithJsonSchemaIT.java
```diff
@@ -0,0 +1,237 @@
...
+    private static final ChatModel STRICT_MODEL = new StreamingChatModelAdapter(
+            OpenAiOfficialResponsesStreamingChatModel.builder()
+                    .apiKey(System.getenv("OPENAI_API_KEY"))
+                    .modelName(InternalOpenAiOfficialTestHelper.CHAT_MODEL_NAME.toString())
+                    .temperature(0.0)
+                    .strict(true)
+                    .build(),
+            true);
+
+    // Responses API validates schemas as strict, even when strict=false.
```
dliubarskyi: Does it mean that `strict` flag has no effect?

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesAiServicesWithJsonSchemaIT.java
```diff
@@ -0,0 +1,237 @@
...
+
+    @Override
+    protected boolean isStrictJsonSchemaEnabled(ChatModel model) {
+        return model instanceof StreamingChatModelAdapter adapter && adapter.strictJsonSchemaEnabled;
+    }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    @Override
+    protected void should_extract_pojo_with_local_date_time_fields(ChatModel model) {
```
dliubarskyi: Please explain why this test needs to be overriden

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesAiServiceWithToolsIT.java
```diff
@@ -0,0 +1,190 @@
...
+
+    @Override
+    protected boolean supportsRecursion() {
+        return true;
+    }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    @Override
+    protected void should_execute_normal_tool_in_parallel_with_primitive_parameters(ChatModel chatModel) {
```
dliubarskyi: Please explain why this tests needs to be overriden

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesAiServiceWithToolsIT.java
```diff
@@ -0,0 +1,190 @@
...
+        parallelToolsExecution(chatModel, ReturnBehavior.TO_LLM);
+    }
+
+    @ParameterizedTest
+    @MethodSource("models")
+    @Override
+    protected void should_execute_immediate_tool_in_parallel_with_primitive_parameters(ChatModel chatModel) {
```
dliubarskyi: Please explain why this tests needs to be overriden

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiResponsesClient.java
```diff
@@ -144,12 +156,7 @@ class OpenAiResponsesClient {
     OpenAiResponsesClient(Builder builder) {
         HttpClientBuilder httpClientBuilder =
                 getOrDefault(builder.httpClientBuilder, HttpClientBuilderLoader::loadHttpClientBuilder);
-        HttpClient httpClient = httpClientBuilder.build();
```
dliubarskyi, line 147: Why removing logging functionality?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiResponsesClient.java
```diff
@@ -758,6 +793,12 @@ private void completeToolCall(String itemId, ToolExecutionRequest.Builder builde
             if (builder == null || completedToolCallItemIds.contains(itemId)) {
                 return;
             }
+            if (maxToolCalls != null && completedToolCalls.size() >= maxToolCalls) {
```
dliubarskyi: Could you please explain this logic? I thought that we should just pass through `maxToolCalls` parameter into OpenAI API?

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiResponsesClient.java
```diff
@@ -173,37 +183,31 @@ void streamingChat(ChatRequest chatRequest, OpenAiResponsesConfig config, Stream
 
     private Map<String, Object> buildRequestPayload(ChatRequest chatRequest, OpenAiResponsesConfig config) {
         ChatRequestParameters parameters = chatRequest.parameters();
+        if (parameters == null) {
```
dliubarskyi: `parameters` is always non-null

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiResponsesClient.java
```diff
@@ -494,20 +507,6 @@ Builder organizationId(String organizationId) {
             return this;
         }
 
-        Builder logRequests(Boolean logRequests) {
```
dliubarskyi, line 509: please return back the logging feature

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -863,14 +954,52 @@ private void handleOutputItemAdded(ResponseOutputItemAddedEvent event) {
...
+            String currentArgs = builder.build().arguments();
+            builder.arguments(currentArgs + delta);
+
+            PartialToolCall partialToolCall = PartialToolCall.builder()
+                    .index(index)
+                    .id(builder.build().id())
+                    .name(builder.build().name())
+                    .partialArguments(delta)
+                    .build();
+            onPartialToolCall(handler, partialToolCall, streamingHandle);
```
dliubarskyi, line 973: I guess `supportsPartialToolStreaming` methods should be removed in all Response-related ITs (also for unofficial OpenAI)

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
maxlepikhin: Removed the unused `strict` parameter from `toMapForOpenAiResponses(...)`. The helper now has a single Responses-mode behavior and the callers were updated accordingly.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
maxlepikhin: Implemented this. I introduced `OpenAiOfficialResponsesChatRequestParameters` and `OpenAiResponsesChatRequestParameters` with `previousResponseId`, removed the mutable shared response-id state, and added coverage for builder defaults and per-request overrides.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
maxlepikhin: Removed this condition. `previousResponseId` is now supplied explicitly via request parameters instead of being inferred from tool execution results or message content.

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
maxlepikhin: You were right that `toMap(..., true)` already sets `additionalProperties=false` for typed object schemas. I kept the Responses helper focused on typed schema elements and aligned the validation/messages so we no longer claim raw schemas are accepted in this mode.

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesAiServicesWithJsonSchemaIT.java
maxlepikhin: For Responses-mode JSON response format, `strict=false` does not change the emitted response-format schema. I adjusted the tests so they no longer imply otherwise, and raw schemas are rejected explicitly for this mode.

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesAiServicesWithJsonSchemaIT.java
maxlepikhin: I reduced this override by moving the special-case expectation into a narrow shared hook in the abstract JSON-schema IT, so the subclass no longer needs the previous ad hoc override.

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesAiServiceWithToolsIT.java
maxlepikhin: I reduced this override by moving the nondeterministic parallel-tool assertion into a shared hook in `AbstractAiServiceWithToolsIT`, so the subclass only declares the capability difference.

langchain4j-open-ai-official/src/test/java/dev/langchain4j/model/openaiofficial/openai/OpenAiOfficialResponsesAiServiceWithToolsIT.java
maxlepikhin: Same here: the subclass override was reduced to a capability hook in the shared tools IT instead of carrying the full test override locally.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiResponsesClient.java
maxlepikhin: Logging is restored in the unofficial Responses client. The request/response logging hooks are back in place.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiResponsesClient.java
maxlepikhin: Agreed. I removed the extra local truncation and now pass `maxToolCalls` through to the OpenAI API instead of applying an additional client-side cap.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiResponsesClient.java
maxlepikhin: Cleaned up. `parameters` is treated as non-null, while still preserving fallback to builder/config defaults for lower-level code paths that bypass interface-level parameter merging.

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiResponsesClient.java
maxlepikhin: Done. The logging feature is back in the unofficial Responses client.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
maxlepikhin: Done. The `supportsPartialToolStreaming` overrides were removed from the Responses-related ITs, and the shared streaming base assertions were adjusted to allow partial tool-call callbacks.

maxlepikhin:
@codex review

maxlepikhin:
@dliubarskyi not sure what you'd like to do next here.

dliubarskyi (approved):
@maxlepikhin thank you!

PR #4570, opened 2026-02-10
Tool Search

dliubarskyi:
## Change
Implements Tool Search Tool feature initially introduced by [Anthropic](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool).

### Tool Search

When working with a large number of tools,
sending all tools on every request can significantly increase token usage and reduce model performance.
To address this, LangChain4j provides a tool search mechanism
that allows tools to be discovered dynamically by the LLM itself,
instead of being exposed upfront.

The core idea is simple:
- Initially, the LLM is exposed to one or more special tool-search tools
- The LLM can call these tools to search for relevant tools
- Once relevant tools are found, they are included in subsequent requests to the LLM

This enables scalable, token-efficient, and model-driven tool discovery.

#### How Tool Search Works

A tool search flow typically looks like this:
1. Initial request:
   - The LLM sees only tool-search tools (not the full tool set)
2. Tool search
   - The LLM calls a tool-search tool, describing what kind of tool it needs
   - The tool-search strategy matches the request against available tools
4. Tool exposure
   - Matching tools are added to the next request to the LLM
5. Tool execution
   - The LLM can now call the found tools normally

Previously found tools are accumulated across multiple tool-search calls.
Each time the LLM invokes the tool-search tool,
the newly matched tools are added to the existing set of tools visible to the LLM (they are merged, not replaced).
This means the list of tools visible to the LLM can grow over time.
Found tools remain visible to the LLM until their corresponding `ToolExecutionResultMessage`
is evicted from the `ChatMemory`, and at least until the end of the AI Service invocation.

If `ChatMemory` is not configured, the found tools remain visible to the LLM
only until the end of the AI service invocation.

#### ToolSearchStrategy

Tool search is implemented via the `ToolSearchStrategy` interface:

```java
@Experimental
public interface ToolSearchStrategy {

    List<ToolSpecification> getToolSearchTools(InvocationContext invocationContext);

    ToolSearchResult search(ToolSearchRequest toolSearchRequest);
}
```

A `ToolSearchStrategy` is responsible for:
- Exposing tool-search tools to the LLM
- Executing tool search requests generated by the LLM
- Returning matching tool names, which will then be resolved and exposed

LangChain4j currently provides 2 out-of-the-box implementations:
- `SimpleToolSearchStrategy` – keyword-based matching
- `VectorToolSearchStrategy` – semantic search using embeddings

See Javadoc of these classes for more details.

You can also implement custom strategies.

#### Configuring Tool Search in AI Services

Tool search is configured at the AI Service level:

```java
    .chatModel(chatModel)
    .chatMemory(chatMemory)
    .tools(tools) // tool search works for static tools
    .toolProvider(mcpToolProvider) // tool search works for tools provided dynamically (e.g., MCP)
    .toolSearchStrategy(new SimpleToolSearchStrategy())
    .build();
```

Once configured:
- The LLM no longer sees all tools upfront
- Tool discovery becomes an explicit, model-driven step
- Token usage is reduced, especially with large tool sets

#### When to Use Tool Search

Tool search is especially useful when:
- You have many tools (dozens or hundreds)
- Tools are domain-specific or rarely used
- Tool availability depends on context, user, or permissions
- You want the LLM to reason about which tools it needs, instead of guessing from a long list

If you have only a small number of tools, or all tools are always relevant,
using regular approach may be simpler.

#### Always Visible Tools

When tool search is enabled, tools are normally hidden from the LLM until they are discovered via a tool-search call.
However, in some cases you may want certain tools to always be visible to the LLM.

Typical use cases:
- Core tools that should always be accessible
- Frequently used tools where search overhead is unnecessary
- Utility tools

LangChain4j supports this via the `ALWAYS_VISIBLE` tool search behavior.

##### How It Works

When a tool is marked as `ALWAYS_VISIBLE`:
- It is exposed to the LLM in the very first request
- It does not require discovery via tool search
- It remains visible throughout the AI Service invocation
- It is not included in searchable tool candidates

All other tools continue to follow the normal tool-search flow.

##### Using `@Tool` Annotation

You can mark a tool as always visible via the @Tool annotation:
```java
@Tool(searchBehavior = ALWAYS_VISIBLE)
String getWeather(String city) {
    return weatherService.getWeather(city);
}
```

##### Using `McpToolProvider`

When using MCP tools (via `McpToolProvider`), always-visible tools can be configured via `alwaysVisibleToolNames`:

```java
McpToolProvider.builder()
    .mcpClients(mcpClient)
    .alwaysVisibleToolNames("getWeather")
    .build();
```

##### Using `ToolSpecification`

If you configure tools programmatically, you can mark them as always visible using `metadata`:
```java
ToolSpecification toolSpecification = ToolSpecification.builder()
    .name("getWeather")
        .addStringProperty("city")
        .required("city")
        .build())
    .metadata(Map.of(ToolSpecification.METADATA_SEARCH_BEHAVIOR, SearchBehavior.ALWAYS_VISIBLE))
    .build();
```

#### Notes and Limitations

NOTE: Tool search relies on the LLM’s ability to understand when and how to search for tools.
The effectiveness of this feature depends heavily on the chosen model.

NOTE: Tool search is currently marked as experimental and may evolve in future releases.

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/data/message/AiMessage.java
```diff
@@ -202,6 +202,17 @@ public String toString() {
                 + attributes + " }";
     }
 
+    /**
+     * @since 1.12.0
+     */
+    public Builder toBuilder() {
```
mariofusco, line 208: We can of course keep this without problems, but it doesn't seem to be used anywhere?

langchain4j/src/main/java/dev/langchain4j/service/tool/search/ToolSearchStrategy.java
```diff
@@ -0,0 +1,47 @@
...
+@Experimental
+public interface ToolSearchStrategy {
+
+    /**
+     * Returns one or more tool specifications that are exposed to the LLM and are used to perform tool searches.
+     *
+     * @param invocationContext the AI Service invocation context
+     * @return list containing one or more tool search tools
+     */
+    List<ToolSpecification> getToolSearchTools(InvocationContext invocationContext);
```
mariofusco, line 38: Nitpick: maybe remove the `get` from the method name to be more uniform with the rest of our codebase?

langchain4j-core/src/main/java/dev/langchain4j/data/message/AiMessage.java
dliubarskyi: It was used by the previous version of the code, but not any more. I can remove it, but I thought it would make sense to keep it to be symmetrical with other message types

langchain4j/src/main/java/dev/langchain4j/service/tool/search/ToolSearchStrategy.java
dliubarskyi: I was wondering about it as well, initially it was called `toolSearchTools`, but I thought that adding `get` might make it clearer. Not sure about it TBH, but I can definitely revert if you think it will be better.

BTW, user-implemented interfaces do use `get` prefixes, for example `ChatMemoryStore` has `getMessages`

mariofusco: I'm not at all strongly opinionated on this, just trying to follow langchain4j conventions, and for sure you know better than me.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -219,12 +225,40 @@ public ToolExecutionErrorHandler executionErrorHandler() {
...
+     * @since 1.12.0
+     */
+    public void toolSearchStrategy(ToolSearchStrategy toolSearchStrategy) {
+        this.toolSearchService = new ToolSearchService(toolSearchStrategy);
+    }
+
+    public ToolServiceContext createContext(InvocationContext invocationContext,
+                                            UserMessage userMessage,
+                                            ChatMemory chatMemory) {
+        ToolServiceContext context = createContext(invocationContext, userMessage);
```
mariofusco: Nitpick: it is a bit weird that the method that you want to use instead of the deprecated one then calls the same deprecated one internally. I guess that at some point the deprecated one will actually stay there and be made `private` at some point? Maybe you may want to simply write this in a comment to avoid to forget this?

dliubarskyi: Yes, the plan is to make it private at some point. Good point regarding the comment, let me add it

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -254,7 +288,8 @@ public ToolServiceContext createContext(InvocationContext invocationContext, Use
             }
         }
         return ToolServiceContext.builder()
-                .toolSpecifications(toolSpecifications)
+                .effectiveTools(toolSpecifications)
+                .availableTools(toolSpecifications)
```
mariofusco, line 287: Is this correct? Do you want the `effectiveTools` and `availableTools` to be set with the same list? I'm seeing this pattern repeated even somewhere else, so maybe you could only set only one of the two, the `effectiveTools` I guess, and in the `availableTools` also return the content of `effectiveTools` when null.

mariofusco (dismissed):
Great job!

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
dliubarskyi: > Is this correct? Do you want the effectiveTools and availableTools to be set with the same list?

Yes, in case when tool search tool is not configured, effective tools and available tools are the same

> so maybe you could only set only one of the two, the effectiveTools I guess, and in the availableTools also return the content of effectiveTools when null.

Not sure I understand, could you please elaborate? How will it improve?

mariofusco: I mean that if effective and available tools are intended to be the same in most cases, maybe this could be internally reflected by the `ToolServiceContext` class, so you could set only one of the two and the other one will return the same result if left null. This will avoid having to set the same `toolSpecifications` twice when creating the `ToolServiceContext`. Anyway this is a very minor thing, feel free to ignore it.

docs/docs/tutorials/tools.md
```diff
@@ -817,6 +817,106 @@ ToolProvider toolProvider = (toolProviderRequest) -> {
...
+   - The LLM calls a tool-search tool, describing what kind of tool it needs
+3. Tool discovery
+   - The tool-search strategy matches the request against available tools
+4. Tool exposure
+   - Matching tools are added to the next request to the LLM
+5. Tool execution
+   - The LLM can now call the discovered tools normally
+
+Previously discovered tools are accumulated and remain available until the corresponding
+`ToolExecutionResultMessage` is evicted from the `ChatMemory`.
```
jmartisk: Maybe explain here that if the model executes the tool again, then tools from all executions will be merged together (at least I hope that is how it works)

jmartisk:
I haven't checked how cleanly this can be integrated in Quarkus, but hopefully it will be ok. I can try and draft it later. Nice work!

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
dliubarskyi: > so you could set only one of the two and the other one will return the same result if left null

This feels like making `ToolServiceContext` (DTO) aware of tool search tool logic, which I would like to keep isolated in the TST classes

docs/docs/tutorials/tools.md
dliubarskyi: Yes, it is exactly how it works, and this is what I meant by "accumulated", but I will clarify it more

dliubarskyi:
FYI: I have just pushed https://github.com/langchain4j/langchain4j/pull/4570/commits/e9a69c53591c808a65b3e722522ada6f250472f4, it changes which tools are provided inside the `ToolSearchRequest`. Previously, all available tools were provided, but now only "searchable" are provided, so that those tools that were found previously (and are already visible to the LLM) do not participate in search

dliubarskyi:
FYI: I have just pushed https://github.com/langchain4j/langchain4j/pull/4570/changes/49b15cac221caeef6c201f320e7c07884f42f901, it adds "always visible tools" feature that makes it possible to make some tools always visible to the LLM when tool search is configured.

geoand:
Just saw this, very cool!

PR #4576, opened 2026-02-11
feat: Make customizable the creation of the vector extension for PGVe…

### LINKED ISSUE #4575: [FEATURE] Make configurable the create vector extension in PGVector

The PGVector embedding store always executes `CREATE EXTENSION IF NOT EXISTS vector` when retrieving a JDBC connection.

This could be an issue when the configured credentials are not a superuser user (and the vector extension is already loaded).

I'd like to have a flag to be able to disable that CREATE EXTENSION statement when retrieving a JDBC connection.

There is no alternative. Without this feature, we need to use a superuser in the PGVector integration.

**Additional context**

I will send a PR with an implementation for this

odelpozo-denodo:
## Issue

Closes #4575

## Change

Added flag skipCreateVectorExtension to the PgVectorEmbeddingStore. Its default value does not change the current behaviour,
When set to true, the CREATE EXTENSION statement is not executed when a new connection is created.

## General checklist

- [X] I have manually verified that the `PgVectorEmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

### DIFF
--- a/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
+++ b/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
@@ -775,19 +775,7 @@ public class PgVectorEmbeddingStore implements EmbeddingStore<TextSegment> {
         }
 
         public PgVectorEmbeddingStore build() {
-            return new PgVectorEmbeddingStore(
-                    this.datasource,
-                    this.table,
-                    this.dimension,
-                    this.useIndex,
-                    this.indexListSize,
-                    this.createTable,
-                    this.dropTableFirst,
-                    this.skipCreateVectorExtension,
-                    this.metadataStorageConfig,
-                    this.searchMode,
-                    this.textSearchConfig,
-                    this.rrfK);
+            return new PgVectorEmbeddingStore(this);
         }
 
         public String toString() {
--- a/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
+++ b/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
@@ -673,7 +673,7 @@ public class PgVectorEmbeddingStore implements EmbeddingStore<TextSegment> {
         // Find a way to do the following code in connection initialization.
         // Here we assume the datasource could handle a connection pool
         // and we should add the vector type on each connection
-        if (skipCreateVectorExtension) {
+        if (!skipCreateVectorExtension) {
             try (Statement statement = connection.createStatement()) {
                 statement.executeUpdate("CREATE EXTENSION IF NOT EXISTS vector");
             }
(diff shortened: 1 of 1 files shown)


langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -120,6 +127,7 @@ protected PgVectorEmbeddingStore(
             Integer indexListSize,
             Boolean createTable,
             Boolean dropTableFirst,
+            Boolean skipCreateVectorExtension,
```
dliubarskyi: This (and other ctor changes below) are breaking change, please avoid.
For existing ctors that do not accept this field, let's set it to the default value (`false`).
This new field should be supported only for the ctor that accepts `Builder`

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -657,8 +673,10 @@ protected Connection getConnection() throws SQLException {
         // Find a way to do the following code in connection initialization.
         // Here we assume the datasource could handle a connection pool
         // and we should add the vector type on each connection
-        try (Statement statement = connection.createStatement()) {
-            statement.executeUpdate("CREATE EXTENSION IF NOT EXISTS vector");
+        if (skipCreateVectorExtension) {
```
```suggestion
        if (!skipCreateVectorExtension) {
```
Shouldn't it be like this?

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
odelpozo-denodo: > This new field should be supported only for the ctor that accepts Builder

There's a problem with that. The ctor receiving the builder calls to `this(...)` and that may execute the `initTable` method that requires a JDBC Connection. 

At that point, I cannot initialize the skipCreateVectorExtension flag with its builder value unless I include it in the ctor signature.

My only alternative is overloading the current ctor to add a new one receiving the new flag. Do you agree @dliubarskyi ?

dliubarskyi: > The ctor receiving the builder calls to this(...)

Let's inverse that and instead of calling this(...) let's initialize all fields directly in `PgVectorEmbeddingStore(PgVectorEmbeddingStoreBuilder)` ctor.

All other ctors can instead create a builder and pass it to the `PgVectorEmbeddingStore(PgVectorEmbeddingStoreBuilder)` ctor.

Would that wotk?

odelpozo-denodo: I had to create a new ctor with DatasourceBuilder because otherwise, with only the PgVectorEmbeddingStoreBuilder ctor we would only create embedding stores with PGSimpleDataSource.

dliubarskyi (approved):
@odelpozo-denodo thank you!

PR #4583, opened 2026-02-14
Fix NPE when toolCall.index() is null in OpenAI streaming (#4573)

### LINKED ISSUE #4573: [BUG] NullPointerException in OpenAiStreamingResponseBuilder with Gemini Streaming Tool Calls
## Description

When using Gemini's OpenAI-compatible API (`https://generativelanguage.googleapis.com/v1beta/openai/`) with `StreamingChatModel` and tools enabled, a `NullPointerException` is thrown during streaming response processing.

The root cause is that Gemini's streaming tool call responses return `null` for `toolCall.index()`, which is then used as a key in `ConcurrentHashMap.computeIfAbsent()`. Since `ConcurrentHashMap` does not allow null keys, this causes the exception.

---

## Environment

- **langchain4j version**: `1.9.1`
- **langchain4j-open-ai version**: `1.9.1`
- **langchain4j-open-ai-spring-boot-starter version**: `1.9.1-beta17`
- **Model Provider**: Google Gemini (via OpenAI-compatible endpoint)
- **Model**: `gemini-3-flash-preview` (also affects `gemini-2.0-flash-exp`)
- **Java Version**: 21
- **Spring Boot Version**: 3.5.10

---

## Steps to Reproduce

1. Configure `StreamingChatModel` to use Gemini's OpenAI-compatible endpoint:

```java
langchain4j:
  open-ai:
    streaming-chat-model:
      base-url: https://generativelanguage.googleapis.com/v1beta/openai/
      api-key: <your-gemini-api-key>
      model-name: gemini-3-flash-preview
```

2. Create an `AiService` with streaming and tools:

```java
AiServices.builder(MyService.class)
    .streamingChatModel(streamingChatModel)
    .tools(myCustomTool)
    .build();
```

3. Call the AI service with a prompt that triggers tool execution

4. Observe the `NullPointerException`

---

## Stack Trace

```
java.lang.NullPointerException
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1693)
	at dev.langchain4j.model.openai.OpenAiStreamingResponseBuilder.append(OpenAiStreamingResponseBuilder.java:160)
	at dev.langchain4j.model.openai.OpenAiStreamingResponseBuilder.append(OpenAiStreamingResponseBuilder.java:85)
	at dev.langchain4j.model.openai.OpenAiStreamingChatModel.lambda$doChat$0(OpenAiStreamingChatModel.java:150)
	at dev.langchain4j.model.openai.internal.StreamingRequestExecutor$2.onEvent(StreamingRequestExecutor.java:140)
	...
```

---

## Root Cause Analysis

In `OpenAiStreamingResponseBuilder.java` (line ~160):

```java
if (delta.toolCalls() != null && !delta.toolCalls().isEmpty()) {
    ToolCall toolCall = delta.toolCalls().get(0);

    // ❌ toolCall.index() returns null for Gemini's streaming responses
    ToolExecutionRequestBuilder builder = this.indexToToolExecutionRequestBuilder.computeIfAbsent(
            toolCall.index(), idx -> new ToolExecutionRequestBuilder());
    // ...
}
```

**Expected**: `toolCall.index()` should be a non-null `Integer`  
**Actual**: Gemini returns `null`, causing `ConcurrentHashMap.computeIfAbsent()` to throw `NullPointerException`

---

## Expected Behavior

The library should handle `null` index values gracefully, either by:
1. Using a default index (e.g., `0`) when `toolCall.index()` is null
2. Throwing a more informative exception with guidance
3. Documenting that Gemini's streaming tool calls are not fully compatible

---

## Actual Behavior

Application crashes with `NullPointerException`, preventing any tool execution via Gemini streaming API.

---

---

## Related Issues

Similar null-handling issues have been reported for Gemini integration:
- #2893 (NullPointerException with Gemini content filtering)
- #827 (NullPointerException when tool has no parameters)

---

**Would appreciate any guidance on the best approach to fix this compatibility issue!**

mohankumar27:
Gemini's OpenAI-compatible streaming API returns null for toolCall.index(), causing NullPointerException in ConcurrentHashMap.computeIfAbsent() and auto-unboxing. Default null index to 0, matching ToolCallBuilder's default.

## Issue
Closes #4573

## Change
When using Gemini's OpenAI-compatible API (`https://generativelanguage.googleapis.com/v1beta/openai/`) with `StreamingChatModel` and tools enabled, `toolCall.index()` returns `null` during streaming. This causes `NullPointerException` in two places:

1. **OpenAiStreamingResponseBuilder.java** — `ConcurrentHashMap.computeIfAbsent()` does not allow null keys
2. **OpenAiStreamingChatModel.java** — Auto-unboxing `null` `Integer` to `int` throws NPE

The fix defaults `toolCall.index()` to `0` when null, which is semantically correct since a null index indicates no parallel tool calls (single tool call at a time). This is consistent with the `ToolCallBuilder` no-arg constructor default.

## General checklist

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
@@ -213,7 +213,18 @@ public class OpenAiStreamingChatModel implements StreamingChatModel {
         if (toolCalls != null) {
             for (ToolCall toolCall : toolCalls) {
 
-                int index = toolCall.index() != null ? toolCall.index() : 0;
+                int index;
+                if (toolCall.index() != null) {
+                    index = toolCall.index();
+                } else {
+                    index = toolCallBuilder.index();
+                    // When index is null and a different tool call id appears, increment the index
+                    if (toolCall.id() != null
+                            && toolCallBuilder.id() != null
+                            && !toolCallBuilder.id().equals(toolCall.id())) {
+                        index = toolCallBuilder.index() + 1;
+                    }
+                }
                 if (toolCallBuilder.index() != index) {
                     onCompleteToolCall(handler, toolCallBuilder.buildAndReset());
                     toolCallBuilder.updateIndex(index);
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
@@ -145,13 +145,14 @@ public class OpenAiStreamingChatModel implements StreamingChatModel {
 
         ChatCompletionRequest openAiRequest =
                 toOpenAiChatRequest(
-                        chatRequest, parameters, sendThinking, thinkingFieldName, strictTools, strictJsonSchema)
+                                chatRequest, parameters, sendThinking, thinkingFieldName, strictTools, strictJsonSchema)
                         .stream(true)
                         .streamOptions(
                                 StreamOptions.builder().includeUsage(true).build())
                         .build();
 
-        OpenAiStreamingResponseBuilder openAiResponseBuilder = new OpenAiStreamingResponseBuilder(returnThinking, accumulateToolCallId);
+        OpenAiStreamingResponseBuilder openAiResponseBuilder =
+                new OpenAiStreamingResponseBuilder(returnThinking, accumulateToolCallId);
         ToolCallBuilder toolCallBuilder = new ToolCallBuilder();
 
         client.chatCompletion(openAiRequest)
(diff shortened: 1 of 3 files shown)


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
```diff
@@ -212,7 +213,7 @@ private void handle(
         if (toolCalls != null) {
             for (ToolCall toolCall : toolCalls) {
 
-                int index = toolCall.index();
+                int index = toolCall.index() != null ? toolCall.index() : 0;
```
dliubarskyi: there might be multiple tool calls, so fallback index should increase

dliubarskyi (commented):
@mohankumar27 thank you!

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingChatModel.java
mohankumar27: @dliubarskyi Thanks for the review! — addressed in commit edda6ff.

Both `OpenAiStreamingResponseBuilder` and `OpenAiStreamingChatModel`
 now detect new tool calls via differing toolCall.id() and auto-increment the fallback index, so multiple tool calls with null indices won't collide. Added a test covering the multiple-null-index scenario as well.

dliubarskyi (approved):
@mohankumar27 thank you!

PR #4584, opened 2026-02-14
fix(mcp): allow setting HTTP/1.1 in StreamableHttpMcpTransport

### LINKED ISSUE #4582: [BUG] StreamableHttpMcpTransport sends h2c Upgrade and hangs against Express MCP servers part of mcp reference samples (fix: force HTTP/1.1)
**Describe the bug**

I’m using langchain4j-mcp 1.11.0-beta19 with StreamableHTTPServerTransport against the MCP SDK StreamableHTTPServerTransport (Node/Express). Requests hang (timeout) and no response is received while testing with server from https://github.com/modelcontextprotocol/ext-apps/tree/main/examples/basic-server-react  (reference sdk samples).

Wire logs show the Java HttpClient sends HTTP/1.1 with h2c upgrade:

  Connection: Upgrade, HTTP2-Settings
  **Upgrade: h2c**
  HTTP2-Settings: ...

Express does not handle the h2c upgrade and stalls without responding. Curl with proper Accept works immediately.

Fix: force HTTP/1.1 on the HttpClient builder:

### StreamableHTTPServerTransport

HttpClient.newBuilder()
  .connectTimeout(timeout)
  .version(HttpClient.Version.HTTP_1_1)

After patching this locally, requests succeed.

**Log and Stack trace**

**To Reproduce**

 Try connecting langchain4j mcp to https://github.com/modelcontextprotocol/ext-apps/tree/main/examples/basic-server-react.

- Start ext-apps example server  on http://localhost:3001/mcp
- Use StreamableHttpMcpTransport to initialize
- Observe timeout unless HTTP/1.1 is forced

**Expected behavior**

Should work with other servers.

- LangChain4j version: 1.11.0
- LLM(s) used: 
- Java version: 25

**Additional context**

We can argue that problem is on  the server side. But since these are reference sdk samples we are likely to see more such scenarios.

weiguangli-io:
## Summary
Fixes #4582 by forcing the Streamable HTTP MCP transport to prefer HTTP/1.1.

## Changes
- Set `HttpClient.Builder.version(HttpClient.Version.HTTP_1_1)` in `StreamableHttpMcpTransport`.
- Added `shouldForceHttp11ForStreamableTransport` in `StreamableHttpMcpTransportTest`.

## Why
Some Express-based MCP servers do not handle h2c upgrade negotiation correctly. Forcing HTTP/1.1 avoids upgrade hangs and keeps Streamable HTTP transport interoperable with MCP reference servers.

## Validation
- `JAVA_HOME=$(brew --prefix openjdk@21)/libexec/openjdk.jdk/Contents/Home ./mvnw -pl langchain4j-mcp -Dtest=StreamableHttpMcpTransportTest test -q`

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
@@ -56,7 +56,9 @@ public class StreamableHttpMcpTransport implements McpTransport {
         Duration timeout = getOrDefault(builder.timeout, Duration.ofSeconds(60));
         customHeadersSupplier = getOrDefault(builder.customHeadersSupplier, (i) -> Map.of());
         sslContext = builder.sslContext;
-        HttpClient.Builder clientBuilder = HttpClient.newBuilder().connectTimeout(timeout);
+        HttpClient.Builder clientBuilder = HttpClient.newBuilder()
+                .connectTimeout(timeout)
+                .version(HttpClient.Version.HTTP_1_1);
         if (builder.executor != null) {
             clientBuilder.executor(builder.executor);
         }
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
@@ -45,6 +45,7 @@ public class StreamableHttpMcpTransport implements McpTransport {
     private volatile McpOperationHandler operationHandler;
     private final HttpClient httpClient;
     private final SSLContext sslContext;
+    private final HttpClient.Version httpVersion;
     private McpInitializeRequest initializeRequest;
     private final AtomicReference<String> mcpSessionId = new AtomicReference<>();
 
@@ -56,8 +57,9 @@ public class StreamableHttpMcpTransport implements McpTransport {
         Duration timeout = getOrDefault(builder.timeout, Duration.ofSeconds(60));
         customHeadersSupplier = getOrDefault(builder.customHeadersSupplier, (i) -> Map.of());
         sslContext = builder.sslContext;
+        httpVersion = getOrDefault(builder.httpVersion, HttpClient.Version.HTTP_1_1);
         HttpClient.Builder clientBuilder =
-                HttpClient.newBuilder().connectTimeout(timeout).version(HttpClient.Version.HTTP_1_1);
+                HttpClient.newBuilder().connectTimeout(timeout).version(httpVersion);
         if (builder.executor != null) {
             clientBuilder.executor(builder.executor);
         }
@@ -254,6 +256,7 @@ public class StreamableHttpMcpTransport implements McpTransport {
         private boolean logResponses = false;
         private Logger logger;
         private SSLContext sslContext;
+        private HttpClient.Version httpVersion;
 
         /**
          * The URL of the MCP server.
@@ -341,6 +344,15 @@ public class StreamableHttpMcpTransport implements McpTransport {
          */
         public StreamableHttpMcpTransport.Builder sslContext(SSLContext sslContext) {
             this.sslContext = sslContext;
+            return this;
+        }
+
+        /**
+         * Sets the HTTP protocol version used by the transport.
+         * Defaults to {@link HttpClient.Version#HTTP_1_1}.
+         */
+        public StreamableHttpMcpTransport.Builder httpVersion(HttpClient.Version httpVersion) {
+            this.httpVersion = httpVersion;
             return this;
         }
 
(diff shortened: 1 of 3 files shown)


jmartisk:
Will this not break compatibility with servers running HTTP/2 only? I think this should be a configurable thing (choosing the protocol version), no?

weiguangli-io:
Good point. I updated this PR to make the protocol version configurable instead of hard-locking behavior.\n\nWhat changed:\n- Added `httpVersion(HttpClient.Version)` to `StreamableHttpMcpTransport.Builder`.\n- Kept default as `HTTP_1_1` (to preserve the original fix), but callers can now opt into `HTTP_2` for servers that require it.\n- Added a regression test (`shouldAllowOverridingHttpVersion`) to verify override behavior.\n\nLocal verification:\n- `./mvnw -pl langchain4j-mcp -am -Dtest=StreamableHttpMcpTransportTest -Dsurefire.failIfNoSpecifiedTests=false test` (3/3 passing).

jmartisk:
Yeah, sorry, one more thing. We try to avoid having the public API depend on things from a specific HTTP client library, so please avoid the builder accepting a `java.net.http.HttpClient.Version` parameter, instead I think we should create our own enum for HTTP version and have its values mapped to values of `java.net.http.HttpClient.Version`. 
@dliubarskyi , what do you think, I guess it could live in the [`dev.langchain4j.http.client`](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-http-client/src/main/java/dev/langchain4j/http/client) package?!

weiguangli-io:
Thanks for the clear guidance, this makes sense.

I updated the PR to remove `java.net.http.HttpClient.Version` from the public builder API.

What changed:
- Added `dev.langchain4j.http.client.HttpVersion` (framework-level enum).
- Changed `StreamableHttpMcpTransport.Builder#httpVersion(...)` to accept `HttpVersion`.
- Kept behavior the same:
  - default is still HTTP/1.1
  - explicit override still supports HTTP/2
- Internally, the transport now maps `HttpVersion` to JDK `HttpClient.Version`.
- Updated tests accordingly (`StreamableHttpMcpTransportTest`).

Local verification:
- `./mvnw -pl langchain4j-mcp,langchain4j-http-client -am -Dtest=StreamableHttpMcpTransportTest -Dsurefire.failIfNoSpecifiedTests=false test`
- Result: `Tests run: 3, Failures: 0, Errors: 0, Skipped: 0`.

If you prefer a different enum naming/location, I can adjust quickly.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
```diff
@@ -56,7 +58,9 @@ public StreamableHttpMcpTransport(StreamableHttpMcpTransport.Builder builder) {
         Duration timeout = getOrDefault(builder.timeout, Duration.ofSeconds(60));
         customHeadersSupplier = getOrDefault(builder.customHeadersSupplier, (i) -> Map.of());
         sslContext = builder.sslContext;
-        HttpClient.Builder clientBuilder = HttpClient.newBuilder().connectTimeout(timeout);
+        httpVersion = getOrDefault(builder.httpVersion, HttpVersion.HTTP_1_1);
```
dliubarskyi: shouldn't the default still be 2.0?

dliubarskyi:
> I guess it could live in the [dev.langchain4j.http.client](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-http-client/src/main/java/dev/langchain4j/http/client) package

I am not sure it should live there if it is not used by the `dev.langchain4j.http.client.HttpClient` API.
As of current implementation, it is used only in the MCP module, so I would keep it there.

But I doubt a new enum is even needed. One could just add a `setHttpVersion1_1()` method to the `StreamableHttpMcpTransport.Builder`

jmartisk:
Ok, that makes sense, let's make it a `setHttpVersion1_1()` method in the builder (the default will be 2.0)

weiguangli-io:
Implemented this per your suggestion.

Updated behavior:
- Default protocol is now HTTP/2.
- Builder now exposes a dedicated `setHttpVersion1_1()` opt-in.
- Removed the temporary `HttpVersion` enum/API entirely.

Code changes:
- `StreamableHttpMcpTransport` now defaults to `HttpClient.Version.HTTP_2` unless `setHttpVersion1_1()` is called.
- Test updates:
  - `shouldUseHttp2ByDefault`
  - `shouldForceHttp11ForStreamableTransport`

Local verification:
- `JAVA_HOME=/usr/local/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home ./mvnw -pl langchain4j-mcp,langchain4j-http-client -am -Dtest=StreamableHttpMcpTransportTest -Dsurefire.failIfNoSpecifiedTests=false test`
- Result: `Tests run: 3, Failures: 0, Errors: 0, Skipped: 0`.

weiguangli-io:
Follow-up: I pushed a small formatting-only commit to fix the `spotless` failure from this change.

- Commit: `e73bbb9ad`
- Scope: line wrapping only in `StreamableHttpMcpTransport` (no logic changes)

Re-verified locally after formatting:
- `./mvnw -pl langchain4j-mcp,langchain4j-http-client -am -Dtest=StreamableHttpMcpTransportTest -Dsurefire.failIfNoSpecifiedTests=false test`
- Result: `Tests run: 3, Failures: 0, Errors: 0, Skipped: 0`.

jmartisk (approved):

jmartisk:
Thanks!

PR #4593, opened 2026-02-17
Add OnnxImageEmbeddingModel to embed images in-process with ONNX

### LINKED ISSUE #4019: [FEATURE] Support image embedding for OnnxEmbeddingModel

I use the _OnnxEmbeddingModel_ to embed an image as a Base64 encoded String. The model I've used is the multi modal sentence-transformers/clip-ViT-B-32-multilingual-v1. Then, given some search terms which I also embed with the same model, I always get nearly the same scores for all images and all search terms 0.85 - 0 0.87. 

```Java
OnnxEmbeddingModel embeddingModel = new OnnxEmbeddingModel(pathToModel, pathToTokenizer, poolingMode);
byte[] data = URI.create(url).toURL().openStream().readAllBytes();
String base64Data = Base64.getEncoder().encodeToString(data);
Response<Embedding> response = embeddingModel.embed(base64Data);
Embedding embedding = response.content();
```
I guess that the encoding / format of the image I've provided is not as expected. 

OnnxEmbeddingModel should offer a method to embed image data. The logic might depend on the particular model being used.

Instead of Base64 encoding I tried to embed the raw byte string as well.
**Additional context**

It is a multi modal vector search scenario. The search terms are texts and the data to be searched are images, e.g. if I search for the term "dog" I should get images of dogs.

Bram--:
## Issue
  Closes #4019

  ## Change
  Adds an experimental `OnnxImageEmbeddingModel`: an `EmbeddingModel` that embeds **images** inside your Java process, by running the vision half of a model such as CLIP or ViT through [ONNX
  runtime](https://onnxruntime.ai/).

  Because it is an ordinary `EmbeddingModel`, embedded images go into an `EmbeddingStore` and are searched exactly like text, and get listeners, metadata and the rest of the `EmbeddingModel` contract for free.

  ```java
  try (OnnxImageEmbeddingModel model = OnnxImageEmbeddingModel.builder()
          .pathToModel("/path/to/vision_model.onnx")
          .preprocessorConfig(ImagePreprocessorConfig.CLIP)
          .build()) {

      EmbeddingResponse response = model.embed(EmbeddingRequest.builder()
              .input(ImageContent.from(Paths.get("/path/to/cat.png").toUri()))
              .build());

      Embedding embedding = response.embeddings().get(0);
  }
  ```

  ### Details

  - **Images only.** `supportedContentTypes()` returns `{IMAGE}`, so core rejects text with an `UnsupportedFeatureException` instead of silently producing a meaningless vector. This follows the existing
  multimodal models (Cohere, Voyage, Jina, Bedrock Titan).
  - **Batching.** Each `EmbeddingInput` of the request yields one `Embedding`, in order. An input holding more than one image is rejected, since a vision encoder produces one vector per image.
  - **`dimension()`** is overridden — the inherited implementation embeds the text `"test"`, which an image-only model cannot do. It is derived once from the model and cached.
  - **Input tensor name** is read from the model rather than assumed to be `pixel_values`, with an explicit error for models that expect more inputs (e.g. a full CLIP model instead of its exported vision tower).
  - **Preprocessing** (resize, center crop, normalize) is configured through `ImagePreprocessorConfig`, which mirrors HuggingFace's `preprocessor_config.json` field for field. `ImagePreprocessorConfig.CLIP` and
  `.DEFAULT` cover the common cases. Getting these values wrong does not fail, it silently produces meaningless embeddings, so they are documented prominently.
  - The model holds a native ONNX session and is `AutoCloseable`.

  ### Searching images with a text query

  Embedding images is only half of multimodal search: the text query has to land in the same vector space. Two small changes to `OnnxBertBiEncoder` make CLIP's text model loadable through the existing
  `OnnxEmbeddingModel`, which previously failed outright:

  - `attention_mask` is now sent only when the model declares it, exactly as `token_type_ids` already was. CLIP's text model expects only `input_ids`, and previously failed with `OrtException: Unexpected number
  of inputs, expected [1,1) found 2`.
  - Models that pool internally and output one vector per text (shape `[batch, dimension]`) are now accepted alongside token-level output (`[batch, sequenceLength, dimension]`), which previously caused a
  `ClassCastException`.

  Both changes are strictly additive: they only affect models that do not fit the BERT shape, and every such model previously threw. With them, the two halves of CLIP work together:

  ```java
  Embedding image = imageModel.embed(EmbeddingRequest.builder()
          .input(ImageContent.from(catPhoto))
          .build())
          .embeddings().get(0);

  Embedding query = textModel.embed("a photo of a cat").content();

  double similarity = CosineSimilarity.between(image, query);
  ```

  ### Testing

  - `OnnxImageEmbeddingModelIT` — 17 tests against the real quantized CLIP ViT-B/32 vision model: URL and base64 input, non-square images, batching and batch ordering, dimension, determinism, cosine similarity
  ranking, pooling modes, and the negative cases (text rejected, multiple images per input rejected, unreachable URL, missing and null model path).
  - `OnnxClipCrossModalIT` — loads both halves of CLIP and verifies that a photo of a cat is closer to `"a photo of a cat"` (0.263) than to `"a photo of a red sports car"` (0.213), and that both towers report
  512 dimensions.
  - `ImagePreprocessorTest` — 76 tests asserting pixel-level parity with the interpolation and cropping conventions of PIL/HuggingFace.
  - The CLIP model used by the integration tests is downloaded by `download-maven-plugin` with a checksum, caching and retries, honouring `embeddingsSkipDownload`, like the models of the other
  `langchain4j-embeddings-*` modules.

  Verified green: 85 unit tests and 28 integration tests in `langchain4j-embeddings`; 34 integration tests across five sibling model modules (`all-minilm-l6-v2` and `-q`, `bge-small-en-v15`,
  `bge-small-zh-v15-q`, `e5-small-v2-q`) covering the `OnnxBertBiEncoder` change against real published models; 1247 tests in `langchain4j-core`; 1327 + 588 in `langchain4j`. `revapi` and `spotless` pass.

  ## General checklist
  - [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and


dliubarskyi:
@Bram-- its seems that this new functionality should live in a separate module, WDYT? IIUC, apart from using ONNX, there is not much else that it shares with the existing `langchain4j-embeddings` module

Bram--:
> @Bram-- its seems that this new functionality should live in a separate module, WDYT? IIUC, apart from using ONNX, there is not much else that it shares with the existing `langchain4j-embeddings` module

What are you thinking, having a separate module for image embeddings? To me it made sense to have them live next to the other ONNX embedders but happy to split it off into a separate module for a more granular module. If so, what would the module logically encapsulate?

Bram--:
> > @Bram-- its seems that this new functionality should live in a separate module, WDYT? IIUC, apart from using ONNX, there is not much else that it shares with the existing `langchain4j-embeddings` module
> 
> What are you thinking, having a separate module for image embeddings? To me it made sense to have them live next to the other ONNX embedders but happy to split it off into a separate module for a more granular module. If so, what would the module logically encapsulate?

@dliubarskyi Friendly ping.

dliubarskyi:
@Bram-- Sorry for the long silence here, and thanks for your patience.

  First, to finally answer my own question about a separate module: **let's keep it where you put it.** Once this is an `EmbeddingModel` it belongs right next to `OnnxEmbeddingModel`, sharing `PoolingMode`,
  `VectorUtils` and the ONNX session handling, and it adds no new dependency to the module. A separate module would have needed those same shared pieces anyway, so it would have bought us a folder and a new
  artifact rather than real isolation. Your instinct was right and my question was the wrong one — sorry for holding the PR on it for so long.

  I've merged current `main` and pushed a reworking of the implementation to your branch. Could you take a look and tell me if you're happy with it? The main points:

  **It's now an `EmbeddingModel`.** This is the substantive change. `EmbeddingModel` became multimodal in 1.18.0 (`EmbeddingRequest` carrying `Content` parts, plus `supportedContentTypes()`), which landed after
  you opened this. So no new interface is needed — `OnnxImageEncoder` became `OnnxImageEmbeddingModel` with `supportedContentTypes() = {IMAGE}` and `doEmbed(EmbeddingRequest)`, following `CohereEmbeddingModel` /
  `VoyageAiEmbeddingModel` / `JinaEmbeddingModel`. That matters beyond consistency: as a standalone class returning a bare `Embedding` it plugged into nothing, and now images can go straight into an
  `EmbeddingStore` and through the RAG pipeline.

  Consequences: a builder instead of public constructors, `dimension()` overridden (the inherited one embeds `"test"`), and the raw `embed(int[] pixels, width, height)` overload dropped from the public API — the
  preprocessor still has it internally, and the tests build real `ImageContent` instead.

  **Cross-modal search works now.** While reviewing I checked whether the feature actually closes #4019 end to end, and it didn't: you can embed images, but the text query has to land in the same vector space,
  and `OnnxEmbeddingModel` could not load CLIP's text model at all. It always sent `attention_mask` (CLIP's text model takes only `input_ids`) and hard-cast the output to a token-level tensor (CLIP outputs one
  already-pooled vector). Two small changes to `OnnxBertBiEncoder` fix both, mirroring the `token_type_ids` handling that was already there. There's a new `OnnxClipCrossModalIT` proving a cat photo is closer to
  "a photo of a cat" than to "a photo of a red sports car". Not your bug at all, but without it the feature was only half usable.

  **Smaller things:** `ImagePreprocessorConfig.CLIP` and `.DEFAULT` presets so nobody has to hunt down those normalisation constants; the config validates channel counts and copies its arrays; the input tensor
  name is read from the model instead of assuming `pixel_values`, with a clear error if someone passes a full CLIP model; `normalize` shared via `VectorUtils` rather than duplicated; the test model is now
  fetched by `download-maven-plugin` with a checksum and caching like the other embedding modules, instead of over a raw `HttpURLConnection` on every CI run; and I reverted the `final` + private constructor on
  `VectorUtils`, since that was binary-incompatible and `revapi` rejected it.

  **Your preprocessor tests are untouched** — all 76 of them still pass exactly as written. They are the reason I was comfortable making these changes at all; the pixel-level parity work was genuinely valuable
  and I'd have had no way to refactor safely without it. I did shrink `cat.png` from 3.2 MB to 268 KB (still RGBA with real transparency, so the RGB-conversion tests still mean something) as that would have
  lived in the repo's history forever.

  Two questions before I merge:

  1. Are you OK with all of the above, especially dropping the raw-pixel overload from the public API?
  2. Any objection to the name `ImagePreprocessorConfig`? It's public API named after `ImagePreprocessor`, which is internal and invisible to users. It does mirror HuggingFace's `preprocessor_config.json`, which
  is an argument for keeping it. Now is the cheap moment to rename it if you'd prefer something like `ImagePreprocessingConfig`.

  If you're happy, the last piece is an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) — image-to-image search plus a text query against an image store would show it off
  well. Would you like to add it, or shall I?

Bram--:
> @Bram-- Sorry for the long silence here, and thanks for your patience.
> 
> First, to finally answer my own question about a separate module: **let's keep it where you put it.** Once this is an `EmbeddingModel` it belongs right next to `OnnxEmbeddingModel`, sharing `PoolingMode`, `VectorUtils` and the ONNX session handling, and it adds no new dependency to the module. A separate module would have needed those same shared pieces anyway, so it would have bought us a folder and a new artifact rather than real isolation. Your instinct was right and my question was the wrong one — sorry for holding the PR on it for so long.
> 
> I've merged current `main` and pushed a reworking of the implementation to your branch. Could you take a look and tell me if you're happy with it? The main points:
> 
> **It's now an `EmbeddingModel`.** This is the substantive change. `EmbeddingModel` became multimodal in 1.18.0 (`EmbeddingRequest` carrying `Content` parts, plus `supportedContentTypes()`), which landed after you opened this. So no new interface is needed — `OnnxImageEncoder` became `OnnxImageEmbeddingModel` with `supportedContentTypes() = {IMAGE}` and `doEmbed(EmbeddingRequest)`, following `CohereEmbeddingModel` / `VoyageAiEmbeddingModel` / `JinaEmbeddingModel`. That matters beyond consistency: as a standalone class returning a bare `Embedding` it plugged into nothing, and now images can go straight into an `EmbeddingStore` and through the RAG pipeline.
> 
> Consequences: a builder instead of public constructors, `dimension()` overridden (the inherited one embeds `"test"`), and the raw `embed(int[] pixels, width, height)` overload dropped from the public API — the preprocessor still has it internally, and the tests build real `ImageContent` instead.
> 
> **Cross-modal search works now.** While reviewing I checked whether the feature actually closes #4019 end to end, and it didn't: you can embed images, but the text query has to land in the same vector space, and `OnnxEmbeddingModel` could not load CLIP's text model at all. It always sent `attention_mask` (CLIP's text model takes only `input_ids`) and hard-cast the output to a token-level tensor (CLIP outputs one already-pooled vector). Two small changes to `OnnxBertBiEncoder` fix both, mirroring the `token_type_ids` handling that was already there. There's a new `OnnxClipCrossModalIT` proving a cat photo is closer to "a photo of a cat" than to "a photo of a red sports car". Not your bug at all, but without it the feature was only half usable.
> 
> **Smaller things:** `ImagePreprocessorConfig.CLIP` and `.DEFAULT` presets so nobody has to hunt down those normalisation constants; the config validates channel counts and copies its arrays; the input tensor name is read from the model instead of assuming `pixel_values`, with a clear error if someone passes a full CLIP model; `normalize` shared via `VectorUtils` rather than duplicated; the test model is now fetched by `download-maven-plugin` with a checksum and caching like the other embedding modules, instead of over a raw `HttpURLConnection` on every CI run; and I reverted the `final` + private constructor on `VectorUtils`, since that was binary-incompatible and `revapi` rejected it.
> 
> **Your preprocessor tests are untouched** — all 76 of them still pass exactly as written. They are the reason I was comfortable making these changes at all; the pixel-level parity work was genuinely valuable and I'd have had no way to refactor safely without it. I did shrink `cat.png` from 3.2 MB to 268 KB (still RGBA with real transparency, so the RGB-conversion tests still mean something) as that would have lived in the repo's history forever.
> 
> Two questions before I merge:
> 
> 1. Are you OK with all of the above, especially dropping the raw-pixel overload from the public API?
> 2. Any objection to the name `ImagePreprocessorConfig`? It's public API named after `ImagePreprocessor`, which is internal and invisible to users. It does mirror HuggingFace's `preprocessor_config.json`, which
>    is an argument for keeping it. Now is the cheap moment to rename it if you'd prefer something like `ImagePreprocessingConfig`.
> 
> If you're happy, the last piece is an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) — image-to-image search plus a text query against an image store would show it off well. Would you like to add it, or shall I?

Thanks! That's great 😄  I'm stll out for a couple of days, will be around to answer and look into this PR on Wednesday.

Bram--:
Guess it'll have to be on Friday 😄

Bram-- (commented):
LGTM! example added in https://github.com/langchain4j/langchain4j-examples/pull/205

PR #4600, opened 2026-02-19
Generate a HTML report of an agentic system topology and execution

mariofusco:
## Issue

Closes #

## General checklist


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/HtmlReportGenerator.java
```diff
@@ -0,0 +1,969 @@
...
+import java.time.Duration;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+public record HtmlReportGenerator(String name, AgentMonitor monitor, AgentInstance rootAgent) {
```
dliubarskyi, line 23: I assume this class was vibe coded? :)
I have nothing against it in this particular case, but perhaps it is worth extracting it in a separate module? Or at least mark it as `@Experimental`? Or maybe it is time to introduce `@VibeCoded` annotation? :D

mariofusco: It is :) I would never been able to write it.

That said I'm not sure I understand your remarks. Creating a separated module only for this (and only because it is vibe coded) seems overkilling and I don't see why we should. Also flagging it as `@Experimental` seems inappropriate: so far the whole agentic module is considered experimental and I don't think this should be more experimental than the rest only because it has been generated.

I'm in favor to clarify which parts of our code is generated (either with an annotation, or simply writing it in a comment) even because I expect that we will have more vibe coded stuff in the near future.

dliubarskyi: Makes sense, let's comment that this was vibe-coded and is not expected to be maintainable manually by a human without LLM's help

mariofusco: I added this in a comment to that class.

dliubarskyi (approved):
@mariofusco thank you!

PR #4616, opened 2026-02-23
Initial implementation for parallel multi instance

### LINKED ISSUE #4615: [FEATURE] Multi  Instance parallel agent (batch)

When trying the declarative approach to describe agents, for example using the [@ParallelAgent](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/ParallelAgent.java) annotation, there are scenarios where instead of coordinating a list of SubAgents, what we want is the same agent operating on items from a list in parallel. 

The @ParallelMultiInstanceAgent proposed in the PR (https://github.com/langchain4j/langchain4j/pull/4616) demonstrates how we can declaratively provide a solution for this scenario. 

The main difference between @ParallelAgent and @ParallelMultiInstanceAgent is that it takes an inputKey that must contain a collection of items to iterate and a single SubAgent (the agent that will be instantiated for each element in the inputKey). 

This allows us to reuse an agent definition while we execute the same agent with different data in parallel. The PR reuses all the mechanisms defined in the ParallelAgent to provide a frictionless user experience. 

I am curious if there is another mechanism that provides a similar functionality that I've missed. 

**Additional context**

I am not attached to the name @ParallelMutilInstanceAgent, so I will appreciate suggestions.

salaboy:
## Issue

Closes #4615

## Change

Provided an implementation for a declarative Multi Instance Parallel agent. 

## General checklist


mariofusco (changes requested):
I fully understand the feature request and it is indeed a brilliant idea. However I'm quite skeptical of the implementation mostly because it doesn't play coherently with other key features of langchain4j. The main problem is that you're invoking the same agent concurrently with different arguments. This could work for most simplistic use cases, but for instance what happens if you have chat memory? 

Also I don't like the alternative agent invoker implementation that reads and writes the items of the list in different keys of the `AgenticScope`. I see why you're doing this but also it seems a bit of an hack. 

Finally, but this is probably minor and easily fixable, it should be possible to pass multiple lists as different arguments, and maybe do something (raising a proper exception?) when the lists are not of the same size. But thinking twice one may want to pass the item of a List as an argument and another whole list as a second argument, how would you disambiguate?

I'm sorry, but at the moment I don't have any more detailed suggestions. I'd need to think and spend a bit more time on this. Probably the only reasonable solution would be to only allow to use this kind of parallel agent without any chat memory and throw an exception if anybody tries to add it. Please check my remarks and see if you can address them. Also please add a few test for the programmatic API.

salaboy:
@mariofusco thanks for the review, I will address those points you mentioned and add the tests. I do understand the concern about the chat model, I thought about it too, but I do need to keep thinking about that.

mariofusco:
> @mariofusco thanks for the review, I will address those points you mentioned and add the tests. I do understand the concern about the chat model, I thought about it too, but I do need to keep thinking about that.

I spent a bit more time reviewing what you did in more details, and have a few hints about the remarks I made yesterday:

- About the `MultiInstanceAgentInvoker` I still believe that it is a bit of an hack. The cleaner alternative would probably be creating a child `AgenticScope` that wraps the original one with a local map that overrides the parent's state and then falls through to the parent scope for everything else. However this solution will be much more invasive, since it will also require to modify the `PlannerLoop` to pass around the right child scope and I would prefer to avoid this.
- Regarding the list(s?) of values to be iterated, let's simplify this: we only allow one and only one list or array to be iterated. In case there is only one list or array among the arguments of the agent we can automatically infer it. If there isn't any we raise an error. In case there is more than one, the one to be iterated has to be explicitly declared with a `itemsProvider` attribute on the `@ParallelMultiInstanceAgent` annotation and if it is missing we raise an error. This is probably similar to what you already meant to do with that `inputKey` but please let's avoid that name because it has different meaning in different parts of the agentic framework. If so probably just a renaming of the attribute, plus some check for the edge conditions that I mentioned will be enough.
- Regarding the annotation also remember that we have both a declarative and a programmatic API. At the moment it seems that you also implemented the second, but I don't see any test for it? If so please add them.
- In the `WorkflowAgentsBuilder` interface please provide a default implementation simply throwing `UnsupportedOperationException` otherwise we will break backward compatibility with other external implementors of that SPI.
- It is evident that we cannot support a meaningful `ChatMemory` for this kind of agent. Please raise an exception if anybody tries to set one.
- Don't forget documentation, add a quick explanation of how this works in the `agents.md` file.

@salaboy Please let me know if this makes sense or if you have any question. As I said I believe that this feature would be quite useful, so if possible I would like to have this with the next release. Please let me know if you have time to spend on this, otherwise I could try to take over.

salaboy:
@mariofusco yes this all makes sense , I will try to make progress today and ping you in the chat if I have concrete questions. I would love to have this asap too, so I will try to prioritize it on my side.

salaboy:
@mariofusco I am trying to work through your list

> - [x] About the MultiInstanceAgentInvoker I still believe that it is a bit of an hack. The cleaner alternative would probably be creating a child AgenticScope that wraps the original one with a local map that overrides the parent's state and then falls through to the parent scope for everything else. However this solution will be much more invasive, since it will also require to modify the PlannerLoop to pass around the right child scope and I would prefer to avoid this. 

For this one, it looks like action is needed as modifying the PlannelLoop will make things extremely complicated for an initial iteration. 

> - [x] Regarding the list(s?) of values to be iterated, let's simplify this: we only allow one and only one list or array to be iterated. In case there is only one list or array among the arguments of the agent we can automatically infer it. If there isn't any we raise an error. In case there is more than one, the one to be iterated has to be explicitly declared with a itemsProvider attribute on the @ParallelMultiInstanceAgent annotation and if it is missing we raise an error. This is probably similar to what you already meant to do with that inputKey but please let's avoid that name because it has different meaning in different parts of the agentic framework. If so probably just a renaming of the attribute, plus some check for the edge conditions that I mentioned will be enough. 

I've renamed `inputKey` to `itemsProvider`, I didn't want the agent to infer from the scope which property to use, as that would limit the use cases. While this is a declarative agent using the annotation I think it is fair to assume that developers will be explicit about which collection needs to be used. The check to validate that itemsProvider is a collection happens here: https://github.com/langchain4j/langchain4j/pull/4616/changes#diff-3cb4ec42f242c6b88328ea22859aac8a7cf010d0606b7f26ffa5e73c438ff4dcR39 

> - [x] Regarding the annotation also remember that we have both a declarative and a programmatic API. At the moment it seems that you also implemented the second, but I don't see any test for it? If so please add them.

Tests for declarative approach here: https://github.com/langchain4j/langchain4j/pull/4616/changes#diff-21545a8852a08f5307de4954463fa0ae8deb1de2b4bf9a28ae0398960c38cd11R916 and test for programatic approach here: https://github.com/langchain4j/langchain4j/pull/4616/changes#diff-c9bf8b215cb35eda3e835bdd831ee2e62e4fbd8aaf5617b1e8d8462d238e92e2R1246 

> - [ ] In the WorkflowAgentsBuilder interface please provide a default implementation simply throwing UnsupportedOperationException otherwise we will break backward compatibility with other external implementors of that SPI.

For this one, I understand all the words but looking at the code I couldn't figure out how to provide a default implementation as `WorkflowAgentsBuilderImpl` is an enum. @mariofusco can you guide me a bit here? 

> - [ ] It is evident that we cannot support a meaningful ChatMemory for this kind of agent. Please raise an exception if anybody tries to set one. 

I found this one to be tricky, I tried this approach: https://github.com/langchain4j/langchain4j/pull/4616/changes#diff-3cb4ec42f242c6b88328ea22859aac8a7cf010d0606b7f26ffa5e73c438ff4dcR77 but I feel that this is not the way to do it, @mariofusco can you please validate this? Test here: https://github.com/langchain4j/langchain4j/pull/4616/changes#diff-21545a8852a08f5307de4954463fa0ae8deb1de2b4bf9a28ae0398960c38cd11R954 **

> - [x] agents.md explanation 

Added

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/ParallelMultiInstancePlanner.java
```diff
@@ -0,0 +1,80 @@
...
+        return done();
+    }
+
+    @Override
+    public AgenticSystemTopology topology() {
+        return AgenticSystemTopology.PARALLEL;
+    }
+
+    @Override
+    public ChatMemoryAccess chatMemoryAccess(AgenticScope agenticScope) {
```
mariofusco: This doesn't make sense a planner doesn't have any `ChatMemory`, it is not an agent on its own, but only a coordinator of agents. We should check that the subagent of this planner doesn't have a memory instead. Please remove this.

salaboy: done, I've removed this

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/WorkflowAgentsIT.java
```diff
@@ -1196,6 +1220,64 @@ static ChatModel chatModel() {
...
+    @Test
+    void parallel_multi_instance_agent_tests() {
+        PersonAstrologyAgent personAstrologyAgent = AgenticServices.agentBuilder(PersonAstrologyAgent.class)
+                .chatModel(baseModel())
+                .outputKey("horoscope")
+                .build();
+
+        BatchHoroscopeAgent agent = AgenticServices.parallelMultiInstanceBuilder(BatchHoroscopeAgent.class)
+                .subAgents(personAstrologyAgent)
+                .itemsProvider("persons")
```
mariofusco, line 1254: In this case, with only an argument of type List, I would expect that the `ParallelMultiInstance` pattern (I'm not totally happy with the name by the way, can we maybe find a more revealing and concise one?) could figure out on its own the `itemsProvider` argument. This should be strictly necessary only in case of ambiguities, in essence when there is more than one List argument.

However, if you prefer, also feel free to leave this improvement out of this pull request. As I said we could improve this and other minor aspects with a follow up PR.

salaboy: @mariofusco yeah.. I will leave it out for the following PR I think I really like the idea to be explicit about this, but I might be missing a lot of history and context on this project to be able to judge. If you do inference to try to find a list, there will be a lot of cases where that will fail for the user. Or pick the wrong one.

mariofusco: Perfect. The idea is that you won't be allowed to "pick the wrong one". If there is any ambiguity and you didn't specify the `itemsProvider` it will raise an error. The "meaningful default" will only be applied in straightforward cases like this.

docs/docs/tutorials/agents.md
```diff
@@ -363,6 +363,60 @@ List<EveningPlan> plans = eveningPlannerAgent.plan("romantic");
...
+        .itemsProvider("persons")
+        .executor(Executors.newFixedThreadPool(3))
+        .build();
+
+List<Person> persons = List.of(
+        new Person("Mario", "aries"),
+        new Person("Luigi", "pisces"),
+        new Person("Peach", "leo"));
+
+ResultWithAgenticScope<Object> result = agent.generateHoroscopes(persons);
```
mariofusco, line 408: Please change this according to the new implementation returning a list. For clarity also explicitly add the `BatchHoroscopeAgent` interface to this example, so users can see how it is defined.

salaboy: done

mariofusco (approved):
@salaboy Great job thanks a lot!

There are a few outstanding minor tasks that I'm listing here to keep track of them:

- Evaluate renaming the `ParallelMultiInstanceAgent` with something more revealing and concise
- Check if the generation of the output can be better implemented through an automatically registered output function instead of aggregating the list of results as a final action in the `nextAction()` method
- Check how it works as `UntypedAgent` and add a test case for this
- Throw an error if a subagent of the `ParallelMultiInstanceAgent` tries to use a `ChatMemory` and eventually clarify better why this is necessary in documentation
- Throw an error if the typed `ParallelMultiInstanceAgent` doesn't return a list or an array in it signature
- Allow to omit the `itemsProvider` attribute where the list argument to be iterated is obvious and there are no ambiguities

I will address all of them in a subsequent pull request.

PR #4624, opened 2026-02-25
Subsidiary SSE channel for Streamable HTTP

jmartisk:
Adds support for the subsidiary SSE channel as desribed in https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#listening-for-messages-from-the-server

`quarkus-mcp-server` calls it "subsidiary", the spec doesn't have any name for it, so I went with subsidiary too.

I went with it being disabled by default because I don't think it is something that will be used very often, and it's somewhat experimental, so better stay on the safe side.

FYI @mkouba


dliubarskyi (approved):
@jmartisk thank you!

PR #4622, opened 2026-02-24
Add HibernateEmbeddingStore integration

beikov:
## Issue

None yet

## Change

Add a new `EmbeddingStore` implementation based on Hibernate ORM and its hibernate-vector module.

* Could make existing SQL specifc Langchain4j `EmbeddingStore` integrations obsolete
* Makes more SQL vector stores available due to Hibernate ORM supporting them
* Could make existing NoSQL Langchain4j `EmbeddingStore` integrations obsolete when we get further with our Hibernate NoSQL efforts
* Allows more vector distance operations and vector types thanks to Hibernate ORM supporting them
* Allows more convenient data model customizations by defining fields on entity classes
* Allows evolving existing entity classes into a source for `EmbeddingStore` API
* Another benefit is that when users use entities, they can then also use HQL to write custom and more complex vector/hybrid-search queries if needed

Thanks to the abstraction, this integration would free Langchaing4j developers from maintainance of all those separate `EmbeddingStore` implementations and focus on just a single integration, with our help.

## General checklist

- [x] There are no breaking changes (API, behaviour)
- [x] I have added unit and/or integration tests for my change
- [x] The tests cover both positive and negative cases
- [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [x] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs) see https://github.com/langchain4j/langchain4j-examples/pull/187
- [x] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new maven module

- [x] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

## Checklist for adding new embedding store integration

- [x] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [x] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

PR #4631, opened 2026-02-26
Make sure the HibernateEmbeddingStore also works with Hibernate ORM 7.1

beikov:
This is a quick follow-up of https://github.com/langchain4j/langchain4j/pull/4622 as I noticed that the integration didn't work with Hibernate ORM 7.1, which is part of the previous Quarkus LTS.


dliubarskyi (approved):
@beikov thank you!

PR #4646, opened 2026-03-02
Agent Skills support

### LINKED ISSUE #4396: [FEATURE] do we support load skills md files for langchain4j agent in new version
hi，
Claude Skills seem to be all the rage lately and have gained recognition among some users. However, I somewhat worry that they are just a gimmick. Personally, I feel that they are merely a new paradigm and workflow for prompt engineering, and there will surely be a variety of new paradigms emerging in the future. That said, do we have any plans to load Skills Markdown files into large language models? 
https://www.anthropic.com/engineering/advanced-tool-use
https://docs.langchain.com/oss/python/langchain/multi-agent/skills

dliubarskyi:
## Issue
Closes https://github.com/langchain4j/langchain4j/issues/4396

## Change

Introduces support for [Agent Skills](https://agentskills.io/home)

# Skills

:::note
The Skills API is experimental. APIs and behavior may still change in future releases.
:::

Skills is a mechanism for equipping an LLM with reusable, self-contained behavioral instructions.
A skill bundles a name, a short description, and a body of instructions (its _content_),
together with optional resources (e.g., references, assets, templates, etc.).
The LLM loads a skill on demand, keeping the initial context small and only pulling in
the detailed instructions when they are actually needed.

:::note
Skills are designed according to the [Agent Skills specification](https://agentskills.io).
:::

### From the File System

Typically, each skill lives in its own directory containing a `SKILL.md` file.
The file must start with a YAML front matter block that declares the skill's `name` and `description`.
Everything below the front matter becomes the skill's content — the instructions given to the LLM
when it activates the skill.

```
skills/
├── docx/
│   ├── SKILL.md
│   └── references/
│       └── tracked-changes.md   ← loaded as a resource
└── data-analysis/
    └── SKILL.md
```

Example `SKILL.md`:

```markdown
---
name: docx
description: Edit and review Word documents using tracked changes
---

When the user asks you to edit a Word document:

1. Always use tracked changes so edits can be reviewed.
   ...
```

Any file in the skill directory (other than `SKILL.md` itself and files under a `scripts/`
subdirectory) is automatically loaded as a `SkillResource` that the LLM can read on demand.

Use `FileSystemSkillLoader` from the `langchain4j-skills` module to load skills from the file system:

```xml
<dependency>
    <artifactId>langchain4j-skills</artifactId>
    <version>1.12.0-beta20</version>
</dependency>
```

```java
// Load all skills found in immediate subdirectories:
List<FileSystemSkill> skills = FileSystemSkillLoader.loadSkills(Path.of("skills/"));

// Or load a single skill by its directory:
FileSystemSkill skill = FileSystemSkillLoader.loadSkill(Path.of("skills/docx"));
```

### Programmatically

Skills do not have to be file-system based.
You can create them from any source — a database, a remote API, generated at runtime — using the builder API:

```java
Skill skill = Skill.builder()
        .name("incident-response")
        .description("Step-by-step runbook for diagnosing and resolving production incidents")
        .content("""
                When a production alert fires:
                1. Call `fetchRecentLogs(serviceName)` to retrieve the last 5 minutes of logs.
                2. Call `checkServiceHealth(serviceName)` to get current health metrics.
                3. Based on the findings, call `createIncidentTicket(summary, severity)`.
                4. If severity is CRITICAL, also call `pageOnCall(incidentId)`.
                """)
        .build();
```

You can also attach resources programmatically:

```java
SkillResource reference = SkillResource.builder()
        .relativePath("references/tone-guide.md")
        .content("Use warm, concise language. Avoid jargon.")
        .build();

Skill skill = Skill.builder()
        .name("customer-support")
        .description("Handles customer support inquiries")
        .content("Follow the tone guide in references/tone-guide.md ...")
        .resources(List.of(reference))
        .build();
```

## Modes

Skills can be integrated with an AI Service in two distinct modes, depending on how much
control and trust you need.

### Tool Mode (Recommended)

**Class:** `Skills` (from the `langchain4j-skills` module)

This corresponds to the **Tool-based agents** integration approach described in the
[Agent Skills specification](https://agentskills.io/integrate-skills).

In this mode, the LLM activates a skill to receive step-by-step instructions, then carries
them out by calling the [tools](/tutorials/tools) you have explicitly registered.
**The LLM has no access to the file system at inference time** — all skill content and
resources are loaded into memory upfront (e.g. via `FileSystemSkillLoader`), and the `activate_skill`
and `read_skill_resource` tools returns that preloaded content rather than reading from disk.
Because only your pre-defined tools can be invoked, **there is no risk of arbitrary code execution**.

#### Registered Tools

| Tool                  | When registered                                                                               |
|-----------------------|-----------------------------------------------------------------------------------------------|
| `activate_skill`      | Always. The LLM calls this to load a skill's full instructions into the context.              |
| `read_skill_resource` | When at least one skill has resources. The LLM calls this to read individual reference files. |

#### How It Works

1. The system message lists the available skills (names and descriptions) so the LLM can choose.
2. The user asks a question that requires a specific skill.
3. The LLM calls `activate_skill("my-skill")` to receive its instructions.
4. The LLM follows those instructions to complete the task, optionally reading resource files along the way.

#### Example Skill

Skills describe the _policy_ — the exact order of calls, required arguments, error-handling steps,
and worked examples — while the actual execution stays in type-safe, tested Java code:

```markdown
---
name: process-order
description: Processes a customer order end-to-end
---

To process an order:

1. Call `validateOrder(orderId)` to check the order is valid.
2. Call `reserveInventory(orderId)` to reserve the required stock.
3. Only if reservation succeeds, call `chargePayment(orderId)`.
4. Finally, call `sendConfirmationEmail(orderId)`.

If any step fails, call `rollbackOrder(orderId)` before reporting the error.
```

#### Wiring It Up

Pass the `ToolProvider` from `Skills` to your AI Service builder alongside your regular tools.
Use `formatAvailableSkills()` to inject the skill catalogue into the system message so
the LLM knows which skills it can activate:

```java
Skills skills = Skills.from(FileSystemSkillLoader.loadSkills(Path.of("skills/")));

MyAiService service = AiServices.builder(MyAiService.class)
        .chatModel(chatModel)
        .tools(new OrderTools()) // your tools
        .toolProvider(skills.toolProvider()) // or .toolProviders(mcpToolProvider, skills.toolProvider())
        .systemMessage("You have access to the following skills:\n" + skills.formatAvailableSkills()
                + "\nWhen the user's request relates to one of these skills, activate it first using the `activate_skill` tool before proceeding.")
        .build();
```

`formatAvailableSkills()` returns an XML-formatted block listing each skill's name and description:

```xml

<available_skills>
    <skill>
        <name>process-order</name>
        <description>Processes a customer order end-to-end</description>
    </skill>
    <skill>
        <name>data-analysis</name>
        <description>Analyse tabular data and produce charts</description>
    </skill>
</available_skills>
```

#### Customisation

The name, description, and parameter metadata of each tool can be overridden through the
corresponding config class on the builder:

```java
Skills skills = Skills.builder()
        .skills(mySkills)
        .activateSkillToolConfig(ActivateSkillToolConfig.builder()
                .name(...)                    // tool name (default: "activate_skill")
                .description(...)             // tool description
                .parameterName(...)           // parameter name (default: "skill_name")
                .parameterDescription(...)    // parameter description
                .throwToolArgumentsExceptions(...) // throw ToolArgumentsException instead of ToolExecutionException (default: false)
                .build())
        .readResourceToolConfig(ReadResourceToolConfig.builder()
                .name(...)                              // tool name (default: "read_skill_resource")
                .description(...)                       // tool description
                .skillNameParameterName(...)             // skill_name parameter name (default: "skill_name")
                .skillNameParameterDescription(...)      // skill_name parameter description
                .relativePathParameterName(...)          // relative_path parameter name (default: "relative_path")
                .relativePathParameterDescription(...)   // static description (takes precedence over provider)
                .relativePathParameterDescriptionProvider(...) // dynamic description based on available resources
                .throwToolArgumentsExceptions(...)       // throw ToolArgumentsException instead of ToolExecutionException (default: false)
                .build())
        .build();
```

### Shell Mode (Experimental)

**Class:** `ShellSkills` (from the `langchain4j-experimental-skills-shell` module)

This corresponds to the **Filesystem-based agents** integration approach described in the
[Agent Skills specification](https://agentskills.io/integrate-skills).

:::warning
**Shell execution is inherently unsafe.**
Commands run directly in the host process environment **without any sandboxing, containerization,
or privilege restriction**. A misbehaving or prompt-injected LLM can execute arbitrary commands
on the machine running your application.
Only use this in controlled environments where you fully trust the input and accept
the associated risks.
:::

In this mode, the LLM is given a single `run_shell_command` tool and reads skill instructions
directly from the file system using shell commands. There is no `activate_skill` or
`read_skill_resource` tool — the LLM navigates skill files like a human developer would.

#### Registered Tools

| Tool                | When registered                                                                                   |
|---------------------|---------------------------------------------------------------------------------------------------|
| `run_shell_command` | Always. The LLM runs shell commands to read `SKILL.md` files, resource files and execute scripts. |

#### How It Works

1. The system message lists available skills with their absolute filesystem paths.
2. The user asks a question that requires a specific skill.
3. The LLM runs `cat /path/to/skills/docx/SKILL.md` to read the instructions.
4. The LLM follows those instructions by running further shell commands.

#### Dependency

Shell execution lives in a separate experimental artifact — add it to your build:

```xml

<dependency>
    <artifactId>langchain4j-experimental-skills-shell</artifactId>
    <version>1.12.0-beta20</version>
</dependency>
```

#### Wiring It Up

All skills must be filesystem-based (loaded via `FileSystemSkillLoader`).
Use `ShellSkills` instead of `Skills`:

```java
ShellSkills skills = ShellSkills.from(FileSystemSkillLoader.loadSkills(Path.of("skills/")));

MyAiService service = AiServices.builder(MyAiService.class)
        .chatModel(chatModel)
        .toolProvider(skills.toolProvider()) // or .toolProviders(mcpToolProvider, skills.toolProvider())
        .systemMessage("You have access to the following skills:\n" + skills.formatAvailableSkills()
                + "\nWhen the user's request relates to one of these skills, read its SKILL.md before proceeding.")
        .build();
```

`formatAvailableSkills()` includes a `<location>` field so the LLM knows
exactly where to find each `SKILL.md`:

```xml

<available_skills>
    <skill>
        <name>docx</name>
        <description>Edit and review Word documents using tracked changes</description>
        <location>/path/to/skills/docx/SKILL.md</location>
    </skill>
    <skill>
        <name>data-analysis</name>
        <description>Analyse tabular data and produce charts</description>
        <location>/path/to/skills/data-analysis/SKILL.md</location>
    </skill>
</available_skills>
```

#### When to Use Shell Mode

This mode is best suited for **experimentation and prototyping**, or when you want to use
third-party skills published by the community (e.g. from the
[agentskills.io](https://agentskills.io) ecosystem) without first porting them to Java.
It lets you wire up a working workflow quickly, then migrate individual actions
to tools as the solution matures.

#### Customisation

Use `RunShellCommandToolConfig` to tune the working directory, output limits,
and parameter names:

```java
ShellSkills skills = ShellSkills.builder()
        .skills(mySkills)
        .runShellCommandToolConfig(RunShellCommandToolConfig.builder()
                .name(...)                              // tool name (default: "run_shell_command")
                .description(...)                       // tool description (default: includes OS name)
                .commandParameterName(...)              // command parameter name (default: "command")
                .commandParameterDescription(...)       // command parameter description
                .timeoutSecondsParameterName(...)       // timeout parameter name (default: "timeout_seconds")
                .timeoutSecondsParameterDescription(...) // timeout parameter description
                .workingDirectory(...)                  // working directory for commands (default: JVM's user.dir)
                .maxStdOutChars(...)                    // max stdout chars in result (default: 10_000)
                .maxStdErrChars(...)                    // max stderr chars in result (default: 10_000)
                .executorService(...)                   // ExecutorService for reading stdout/stderr streams
                .throwToolArgumentsExceptions(...)      // throw ToolArgumentsException instead of ToolExecutionException (default: false)
                .build())
        .build();
```

## General checklist

- [X] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`


dliubarskyi:
cc @stephanj @mullerhai @glaforge @smileyboy2019

langchain4j-skills/src/main/java/dev/langchain4j/skills/AbstractSkill.java
```diff
@@ -0,0 +1,85 @@
...
+    private final String description;
+    private final String content;
+    private final List<SkillResource> resources;
+
+    protected AbstractSkill(BaseBuilder<?> builder) {
+        this.name = ensureNotBlank(builder.name, "name");
+        this.description = ensureNotBlank(builder.description, "description");
+        this.content = ensureNotBlank(builder.content, "content");
+        this.resources = copy(builder.resources);
+        validateUniquePaths(this.resources);
```
mariofusco, line 27: If you want to have unique paths, why not enforcing this at API level having a `Set` as argument of the `resources` method instead of a generic `Collection`?

langchain4j-skills/src/main/java/dev/langchain4j/skills/Skills.java
```diff
@@ -0,0 +1,292 @@
...
+         * returned to the LLM. For these tools, returning the error message
+         * to the LLM is usually the desired behavior.
+         * <p>
+         * If this flag is set to {@code true}, {@link ToolArgumentsException}
+         * will be thrown instead.
+         *
+         * @param throwToolArgumentsExceptions whether to throw {@link ToolArgumentsException}
+         * @return this builder
+         */
+        public Builder throwToolArgumentsExceptions(Boolean throwToolArgumentsExceptions) {
```
mariofusco: This seems an internal implementation detail that shouldn't be exposed to end users? Or in other words as an user of this API it is not clear to me why should I care about the kind of Exception thrown when something goes wrong, can you please clarify?

langchain4j-skills/src/main/java/dev/langchain4j/skills/AbstractSkill.java
dliubarskyi: Even with `Set`, there can be 2 resources with same paths but different contents

langchain4j-skills/src/main/java/dev/langchain4j/skills/SkillResource.java
```diff
@@ -0,0 +1,23 @@
+package dev.langchain4j.skills;
+
+/**
+ * An additional resource associated with a {@link Skill}, such as a reference file or asset,
+ * that the LLM can read by calling the {@code read_skill_resource} tool.
+ */
+public interface SkillResource {
```
mariofusco, line 10: I'm not entirely sure if this interface is necessary, at the moment I cannot imagine another implementation  alternative to the default one.

langchain4j-skills/src/main/java/dev/langchain4j/skills/Skills.java
dliubarskyi: We explicitly differentiate between tool **_argument_** errors and tool **_execution_** errors, as users migh want to treat them differently, see more details here: https://docs.langchain4j.dev/tutorials/tools#error-handling

The `throwToolArgumentsExceptions` method is meant for users that explicitly difene error handlers to be able to control which exception type is thrown (which controls which error handler is eventually called).

The problem is that currently, the default behavior on tool argument error is to fail the whole AI Service invocation, which does not make sense because LLMs are pretty good at fixing any errors with arguments. But this was the default for a long time before it was spotted, so we cannot  change it now as it would be a breaking change for some users. This change is planned for 2.0 release.

langchain4j-skills/src/main/java/dev/langchain4j/skills/SkillResource.java
dliubarskyi: Perhaps a resource whose content is not eagerly loaded into memory, but lazily loaded only when LLM calls the tool?

But I agree, interface might not be necessary

mariofusco: Good point, leave it. I just wanted to simplify the code where possible, but I agree it makes no harm.

langchain4j-skills/src/main/java/dev/langchain4j/skills/Skills.java
mariofusco: Ok, got it. Yes, we will probably need to review this for the 2.0.

mullerhai:
Hi，when will  [langchain4j-skills]   java jar  release ?  @mariofusco  @dliubarskyi  ,langchain4j really need implement skills

dliubarskyi:
@mullerhai did you have a chance to review this PR? The release is planned for this week, most probably thursday

mullerhai:
> @mullerhai did you have a chance to review this PR? The release is planned for this week, most probably thursday

I’m very willing to review the PR for this skill. Do you need to grant me the relevant review permissions?

langchain4j-skills/src/test/java/dev/langchain4j/skills/SkillsIT.java
```diff
@@ -0,0 +1,258 @@
...
+                .build();
+
+        // when
+        assistant.chat("Use 'process' tool for Klaus Heisler");
+
+        // then
+        verify(spyTools).generateId("Heisler", "Klaus");
+        verify(spyTools).process("Klaus", 177, "Heisler");
+        verify(spyTools).reset();
+        verifyNoMoreInteractions(spyTools);
```
mariofusco, line 105: Very nice example 👍

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -170,6 +170,13 @@ public Object invoke(Method method, Object[] args, InvocationContext invocationC
                                 : null;
 
                         Optional<SystemMessage> systemMessage = prepareSystemMessage(memoryId, method, args);
+                        if (context.systemMessageTransformer != null) {
+                            String transformedSystemMessage = context.systemMessageTransformer.apply(
+                                    systemMessage.map(SystemMessage::text).orElse(null), invocationContext);
+                            systemMessage = transformedSystemMessage != null
+                                    ? Optional.of(SystemMessage.from(transformedSystemMessage))
+                                    : Optional.empty();
```
mariofusco, line 178: Nitpick: not sure if this would be more readable (I think so), but if I understand the logic this could be replaced with something like:

```
systemMessage = systemMessage
    .map(SystemMessage::text)
    .map(text -> context.systemMessageTransformer.apply(text, invocationContext))
    .map(SystemMessage::from);
```

mariofusco (dismissed):
Amazing work!

Once this will be merged and before dropping the new release, please give me a bit of time to also add the new `systemMessageTransformer` and `toolProviders` to agents.

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: It is definitely more readable (I also don't like current impl), but in this case `systemMessageTransformer` will be called only if some system message was already present. But it would be nice to call it in any case

mariofusco: Ok, I see what you mean.

dliubarskyi:
@mullerhai you don't need any permissions to do the review, you can just look at the code and add your comments

langchain4j-skills/src/main/java/dev/langchain4j/skills/Skills.java
```diff
@@ -0,0 +1,304 @@
...
+ *         .chatModel(chatModel)
+ *         .systemMessage("You have access to the following skills: " + skills.formatNamesAndDescriptions() + "\nWhen the user's request relates to one of these skills, activate it first using the 'activate_skill' tool before proceeding.")
+ *         // or, if you already have a system message configured:
+ *         .systemMessageTransformer(systemMessage -> systemMessage + "\n\nYou have access to the following skills: " + skills.formatNamesAndDescriptions() + "\nWhen the user's request relates to one of these skills, activate it first using the 'activate_skill' tool before proceeding.")
+ *         .toolProvider(skills.toolProvider()) // or .toolProviders(mcpToolProvider, skills.toolProvider())
+ *         .build();
+ * }</pre>
+ */
+@Experimental
+public class Skills {
```
vesense, line 53: `toString` method can be added to facilitate viewing registered skills during debugging

experimental/langchain4j-experimental-skills-shell/src/main/java/dev/langchain4j/skills/shell/ShellCommandRunner.java
```diff
@@ -0,0 +1,184 @@
...
+        List<String> shellCommand = isWindows()
+                ? List.of("cmd", "/c", command)
+                : List.of("sh", "-c", command);
+
+        ProcessBuilder pb = new ProcessBuilder(shellCommand);
+        if (workingDirectory != null) {
+            pb.directory(workingDirectory.toFile());
+        }
+
+        Process process = pb.start();
```
vesense, line 84: If subsequent code throws an exception, the process may not be cleaned up correctly. It is recommended to use try-with-resources.

experimental/langchain4j-experimental-skills-shell/src/main/java/dev/langchain4j/skills/shell/ShellCommandRunner.java
```diff
@@ -0,0 +1,184 @@
...
+        timeoutSeconds = getOrDefault(timeoutSeconds, DEFAULT_TIMEOUT_SECONDS);
+        if (timeoutSeconds > DEFAULT_MAX_TIMEOUT_SECONDS) {
+            timeoutSeconds = DEFAULT_MAX_TIMEOUT_SECONDS;
+        }
+
+        boolean finished = process.waitFor(timeoutSeconds, TimeUnit.SECONDS);
+
+        if (!finished) {
+            timedOut.set(true);
+            process.descendants().forEach(ProcessHandle::destroyForcibly);
```
vesense: If `descendants() `throws an exception, it may result in `destroyForcibly()` not being called, and the main process not being destroyed.
It is recommended to use more robust cleanup logic:
```
  try {
      process.descendants().forEach(ProcessHandle::destroyForcibly);
  } catch (Exception e) {
      // log warning
  }
  process.destroyForcibly();
```

vesense:
@dliubarskyi Thanks for your great work, I left some comments about resource leaks and process cleanup.

experimental/langchain4j-experimental-skills-shell/src/main/java/dev/langchain4j/skills/shell/ShellCommandRunner.java
dliubarskyi: good point, added try-catch

experimental/langchain4j-experimental-skills-shell/src/main/java/dev/langchain4j/skills/shell/ShellCommandRunner.java
dliubarskyi: good point, added

dliubarskyi:
@mariofusco 

> Once this will be merged and before dropping the new release, please give me a bit of time to also add the new systemMessageTransformer and toolProviders to agents.

Just in case, this PR is merged, and the release is scheduled for tomorrow, but I can push it a bit if needed

gAmUssA:
@dliubarskyi, thanks for merging this. Curious, when will 1.12.0-beta20 be available for testing?

dliubarskyi:
The release is planned for tomorrow

smileyboy2019:
have a question. Is the content of a skill loaded into memory after it is loaded? If there are many skills, how can we search for them, such as searching for relevant skills in a vector library or database for quick matching?

DTim887:
I can't wait to use it.

sunishkumar:
There are examples of using Skills with AiService. Can it be used with Agents like UntypedAgent?

PR #4647, opened 2026-03-03
Allows to mix programmatic and declarative API defining agentic patterns

mariofusco:
## Issue

Closes #

## Change

So far only AI-based agents could be defined mixing declarative and programmatic API. This improvement extends this possibility to all agentic pattern.

## General checklist


dliubarskyi (approved):
@mariofusco thank you!

Just double-checking: there are no documentation updates, is it right?

mariofusco:
> Just double-checking: there are no documentation updates, is it right?

No, docs only generally says that it is possible to mix declarative and programmatic API, but so far this was actually possible only for final agents, not for agentic pattern. I still have to better clarify the docs for the whole declarative API, I will do this with a different pull request.

PR #4650, opened 2026-03-03
adding multimodality support for untyped agents

### LINKED ISSUE #4649: [FEATURE] Add Multimodality support for Untyped Agents
Currently, the entry point for an Untyped Agent is defined as a method with the following signature:

`Object invoke(@V("input") Map<String, Object> input);`

where elements of `input `map are treated as prompt variables.

There is no way to pass in `Content` objects.

Enhance `InvocationHandler` in `DefaultAiServices` by checking the types of the input map values in case of a signle argument of `Map` type.
Include each value which is an instance of `Content` or it subclasses into the resulting `UserMessage` as an separate `Content` piece.

1) Changing the signature of `UntypedAgent.invoke` - would be a breaking change;
2) Adding a new method with extended signature to `UntypedAgent`- is not in line with the constraint limiting the number of `@Agent` annotated methods in a class to 1;

**Additional context**
The feature is greatly missed in large-scale enterprise-level document-processing pipelines.
Due to the high cost of a code change iteration, untyped agents are beneficial thanks to the ability to keep their configuration (prompts, keys, etc) in a persistent storage and change frequently without having to rebuild an application.

vg2902:
## Issue

Closes #4649

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -211,7 +211,8 @@ class DefaultAiServices<T> extends AiServices<T> {
                         UserMessage userMessage = invokeInputGuardrails(
                                 context.guardrailService(), method, userMessageForAugmentation, commonGuardrailParam);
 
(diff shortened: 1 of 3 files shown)


mariofusco:
Replaced by https://github.com/langchain4j/langchain4j/pull/4656

PR #4662, opened 2026-03-05
adding memoryId support for untyped agents

### LINKED ISSUE #4661: [FEATURE] Add MemoryId support for Untyped Agents
Currently, the entry point for an Untyped Agent is defined as a method with the following signature:

`Object invoke(@V("input") Map<String, Object> input);`

There is no way to annotate a value in this map as `MemoryId`.

Enhance InvocationHandler in DefaultAiServices by checking if the input map has @**MemoryId** key, and use the corresponsing value as MemoryId.

Changing the signature of `UntypedAgent.invoke` - would be a breaking change;
Adding a new method with extended signature to `UntypedAgent` - is not in line with the constraint limiting the number of @**Agent** annotated methods in a class to 1;

**Additional context**
@**MemoryId** constant already exists and is used internally in other palces.

vg2902:
## Issue

Closes #4661

## General checklist


mariofusco (changes requested):
I consider the `UntypedAgent` a sort of expedient that should be used only in rare circumstances and more important I don't want to leak too much of its implementation needs into langchain4j core.

Please consider introducing an alternative `UntypedAgentWithMemory` interface instead.

mariofusco:
Please do a complete rework of this introducing an `UntypedAgentWithMemory` interface as suggested. Let me know if you have any question.

PR #4670, opened 2026-03-06
Add stream-based persistence API for InMemoryEmbeddingStore

### LINKED ISSUE #4466: Stream-based persistence for in-memory embedding store (follow-up)
@wangxuanni @dliubarskyi  this is a follow-up from #4393  which was already implemented very quickly and closed before I saw the implementation.

I do not see the issue #4393 as solved, as the only thing the implementation does is using streams internally, but not exposing that. I would have re-opened that issue, but I have no permission to do so.

Quick repetition of the original description:

---

We use in-memory embedding stores (`InMemoryEmbeddingStore`) and lazily populate them with data. Unfortunately. serializing and de-serializing is file-based and uses JSON strings. This does not scale for even medium number of embeddings as it keeps it all in memory

File-based persistence does not work for us as we have a stream-based abstraction layer which supports different backend implementations.

**Additional context**
n/a

---

My intent with raising the ticket was to
* be able to write to and read from a stream without any files involved (i.e. having the methods `serializeToStream(OutputStream out)` and `fromStream(InputStream in)` methods)
* avoid having the JSON structure materialized as string with even double the memory as this is again being copied using `String.getBytes()` and `Files.readAllBytes(filePath)`

To solve the latter, I suggest to rather extend `InMemoryEmbeddingStoreJsonCodec`:

```
@Internal
public interface InMemoryEmbeddingStoreJsonCodec {

    // existing method
    InMemoryEmbeddingStore<TextSegment> fromJson(String json);
    // existing method
    String toJson(InMemoryEmbeddingStore<?> store);

    // new stream-based method
    InMemoryEmbeddingStore<TextSegment> fromJson(InputStream jsonStream);
    // new stream-based method
    void toJson(InMemoryEmbeddingStore<?> store, OutputStream jsonStream);
}
```

This would also require a change to the underlying Jackson codec class, but that should be simply as well.

With the current approach, everything needs to be loaded into memory and handled as String which is not very efficient, doubles the memory cost and is also slower.

DragonFSKY:
## Issue

Closes #4466

## Change

Add two public methods to `InMemoryEmbeddingStore` for stream-based persistence, allowing users to serialize/deserialize without file involvement:

- `serializeToStream(OutputStream)` — delegates to the codec's stream-based serialization
- `fromStream(InputStream)` — static factory method, delegates to the codec's stream-based deserialization

Both methods follow the same pattern as the existing `serializeToFile`/`fromFile` API. The caller is responsible for closing the provided stream.

Additionally, disabled Jackson's `AUTO_CLOSE_TARGET` and `AUTO_CLOSE_SOURCE` on the internal `ObjectMapper` to prevent it from closing caller-provided streams, which would violate the "caller is responsible for closing" contract documented in the Javadoc.

### Files changed

- `InMemoryEmbeddingStore.java` — 2 new public methods + updated class-level Javadoc
- `JacksonInMemoryEmbeddingStoreJsonCodec.java` — disabled auto-close features
- `InMemoryEmbeddingStoreTest.java` — 3 new tests (round-trip, stream-not-closed, IO error wrapping)

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
+++ b/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
@@ -47,9 +47,11 @@ import java.util.concurrent.CopyOnWriteArrayList;
  * <p>
  * Uses a brute force approach by iterating over all embeddings to find the best matches.
  * <p>
- * This store can be persisted using the {@link #serializeToJson()} and {@link #serializeToFile(Path)} methods.
+ * This store can be persisted using the {@link #serializeToJson()}, {@link #serializeToFile(Path)},
+ * and {@link #serializeToStream(OutputStream)} methods.
  * <p>
- * It can also be recreated from JSON or a file using the {@link #fromJson(String)} and {@link #fromFile(Path)} methods.
+ * It can also be recreated from JSON, a file, or an {@link InputStream} using the {@link #fromJson(String)},
+ * {@link #fromFile(Path)}, and {@link #fromStream(InputStream)} methods.
  *
  * @param <Embedded> The class of the object that has been embedded.
  *                   Typically, it is {@link dev.langchain4j.data.segment.TextSegment}.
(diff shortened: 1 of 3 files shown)


dliubarskyi:
@jetztgradnet please review

PR #4671, opened 2026-03-06
Add JSON serialization support for ToolSpecification

### LINKED ISSUE #4071: [FEATURE] Canonical way to serialize/deserialize ToolSpecification as JSON
We’re building a application where tools are dynamic and need to be stored externally (e.g., in a database) and attached to AI Services at runtime. We generate tool specs using `ToolSpecifications.toolSpecificationsFrom(...)`, which works well. However, we need a canonical way to serialize these `ToolSpecification` instances to JSON and later deserialize them back into usable objects.

The main challenge is that there’s no public API to convert a JSON schema document (with `"type"`, `"properties"`, `"required"`, etc.) back into a `ToolSpecification`. This makes it difficult to persist tool definitions in a stable format and rehydrate them later.

We’d like a public, supported way to:

*   Serialize a `ToolSpecification` (including its `JsonObjectSchema` and nested `JsonSchemaElement`s) into a canonical JSON Schema format.
*   Deserialize that JSON back into a `ToolSpecification` instance.

Ideally, this would support standard JSON Schema conventions (e.g., `"type": "string"`, `"type": "object"`, etc.) so the format is portable and versionable.

*   Writing our own custom mappers to convert between `ToolSpecification` and a canonical JSON Schema format.
*   Post-processing the JSON to inject `"type"` fields based on the schema element class.
*   Persisting the spec as-is and reconstructing it manually using the builder APIs.

These approaches work but are fragile and require maintaining custom logic that could drift from LangChain4j internals.

**Additional context**  
Is there a supported approach today o serialize ToolSpecification to JSON and deserialize it back without relying on MCP or internal APIs?

Thanks!

DragonFSKY:
## Issue
Closes #4071

## Change
Implements **Option 2** as discussed with @dliubarskyi — adds `toJson()` / `fromJson(String)` convenience methods on `ToolSpecification`, delegating to internal utility classes.

### New files
- **`JsonSchemaElementJsonUtils`** — bidirectional `JsonSchemaElement` ↔ `Map<String, Object>` conversion
  - Supports all 11 `JsonSchemaElement` subtypes
  - Falls back to `JsonRawSchema` for unsupported JSON Schema constructs (unknown `type`, missing `type`, non-`#/$defs/` `$ref`)
  - Explicit input validation with `IllegalArgumentException` for malformed structures
  - Independent of `ToolSpecification` — reusable for future `JsonSchema` serialization
- **`ToolSpecificationJsonUtils`** — `ToolSpecification` ↔ JSON string conversion using `JsonSchemaElementJsonUtils` for parameters and `Json` SPI for the string boundary

### Modified files
- **`ToolSpecification`** — added `toJson()` instance method and `fromJson(String)` static method

### Tests
- **`JsonSchemaElementJsonUtilsTest`** (27 tests) — round-trip for all subtypes, fallback-to-raw scenarios, input validation for all error paths
- **`ToolSpecificationJsonTest`** (6 tests) — round-trip with various field combinations, rejection of invalid parameters/metadata

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementJsonUtils.java
```diff
@@ -0,0 +1,301 @@
...
+ * Converts between {@link JsonSchemaElement} and JSON Schema {@link Map} representation.
+ *
+ * @see JsonSchemaElementUtils
+ */
+@Internal
+public class JsonSchemaElementJsonUtils {
+
+    private JsonSchemaElementJsonUtils() {}
+
+    public static Map<String, Object> toMap(JsonSchemaElement element) {
```
dliubarskyi, line 61: Did you check existing JsonSchemaElementUtils.toMap? Can't it be re-used here?

dliubarskyi (commented):
@DragonFSKY thank you!

DragonFSKY:
Thanks for the review!

I looked into reusing `JsonSchemaElementUtils.toMap()`, but it's lossy even in non-strict mode — `additionalProperties` is dropped entirely, and null `items` is replaced with `{}`. These propagate recursively through nested schemas, so round-trip fidelity breaks.

Based on your feedback, the latest push includes a few cleanups:
- `typedSchema()` helper to deduplicate the `type`+`description` boilerplate
- Representability checks in `fromMap()` so extra JSON Schema keywords (`format`, `pattern`, …) fall back to `JsonRawSchema` instead of being silently dropped
- Type-safe `optionalString()` replacing raw `(String)` casts

Would appreciate another look when you get a chance.

dliubarskyi (approved):
@DragonFSKY thank you!

PR #4679, opened 2026-03-07
fix: materialize content arguments before input guardrails

### LINKED ISSUE #4678: [BUG] AiServices input guardrails run before content arguments are materialized into the final UserMessage
**Describe the bug**

`InputGuardrail` is currently executed against an intermediate `UserMessage`, not the final `UserMessage` that is eventually sent to the model.

In `AiServices`, the current order is effectively:
1. build the initial user message
2. run RAG augmentation (if configured)
3. execute input guardrails
4. only then merge method-level `Content` / `List<Content>` / multimodal arguments into the `UserMessage`

This means an input guardrail can observe a single-text message even though the actual request sent to the model is multimodal.

This is a framework-level bug, not just a feature request from a downstream module: the natural expectation for `InputGuardrail` is that it validates the final materialized user input.

**To Reproduce**

A minimal reproduction looks like this:

```java
interface VisionAssistant {

    @UserMessage("Describe this image")
    @InputGuardrails(RecordingGuardrail.class)
    String describe(@UserMessage ImageContent image);
}

public static class RecordingGuardrail implements InputGuardrail {
    @Override
    public InputGuardrailResult validate(InputGuardrailRequest request) {
        UserMessage userMessage = request.userMessage();
        // Today this can observe a single-text message here,
        // even though the final request sent to the model is text + image.
        return success();
    }
}
```

The same issue can happen with `Content`, `List<Content>`, or map-based content arguments.

**Expected behavior**

`InputGuardrail` should receive the final materialized `UserMessage` after method content arguments have been merged, while still preserving the current invariant that RAG augmentation happens before guardrail execution.

The desired order is:
1. build the initial user message
2. run RAG augmentation (if configured)
3. merge method-level `Content` arguments into the `UserMessage`
4. execute input guardrails against that final message

**Why this matters**

This causes incorrect behavior for any input guardrail that depends on the actual message shape, for example:
- guardrails that distinguish single-text vs multimodal input
- guardrails that inspect `contents()` directly
- guardrails that rewrite text only when the request is plain text
- observability/debugging of guardrail input state

A downstream example is a conservative prompt-rewriting guardrail that intentionally skips non-single-text messages. With the current ordering, it can incorrectly rewrite a request that later becomes multimodal.

- LangChain4j version: `main` / `1.13.0-SNAPSHOT`
- Java version: 17+

**Additional context**

This looks related in spirit to #3991, but this report is specifically about AI Services method content arguments being merged after input guardrails run.

I am preparing a PR with a regression test and a minimal ordering fix.

qicesun:
## Summary
- fix `AiServices` input guardrail ordering for method-level `Content` / `ImageContent` arguments
- keep RAG augmentation before input guardrails
- add regression coverage for multimodal and augmented user messages seen by input guardrails

Closes #4678.

## Problem
`InputGuardrail` currently runs against an intermediate `UserMessage`.

In `DefaultAiServices`, method-level content arguments are merged into the `UserMessage` only **after** input guardrails run. As a result, an input guardrail can observe a single-text message even though the actual request sent to the model is multimodal.

That affects any guardrail that depends on the real message shape, for example:
- distinguishing plain-text vs multimodal input
- inspecting `contents()` directly
- rewriting text only for single-text requests

## Change
This PR moves `addContentsToUserMessage(...)` to happen after RAG augmentation but before input guardrail execution.

The resulting order is:
1. build the initial user message
2. run RAG augmentation, if configured
3. merge method-level content arguments into the user message
4. execute input guardrails
5. append output format instructions / continue normal request flow

This keeps the existing RAG invariant while ensuring input guardrails see the materialized user input instead of a partially built message.

## Tests
Added regression coverage in `AiServiceGuardrailTests` for:
- input guardrails receiving a materialized multimodal user message
- input guardrails seeing the augmented message after RAG and before the chat request is sent
- plain-text input guardrail rewrite behavior remaining intact

Also re-ran related service-layer tests:
- `AiServiceGuardrailTests`
- `AiServicesUserMessageConfigTest`
- `AiServicesWithCustomContentInjectorTest`
- `AiServicesRagChatMemoryBehaviorTest`
- `AiServicesObservabilityTests`
- `spotless:check`

## Notes
A downstream community feature exposed this bug, but the fix here is framework-level and not specific to that feature.

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -214,10 +214,12 @@ class DefaultAiServices<T> extends AiServices<T> {
                                 .variables(variables)
                                 .build();
 
-                        UserMessage userMessage = invokeInputGuardrails(
-                                context.guardrailService(), method, userMessageForAugmentation, commonGuardrailParam);
+                        UserMessage userMessage = addContentsToUserMessage(method, args, userMessageForAugmentation);
+                        userMessage = invokeInputGuardrails(
+                                context.guardrailService(), method, userMessage, commonGuardrailParam);
 
-                        Type returnType = context.returnType != null ? context.returnType : method.getGenericReturnType();
+                        Type returnType =
+                                context.returnType != null ? context.returnType : method.getGenericReturnType();
                         boolean streaming = returnType == TokenStream.class || canAdaptTokenStreamTo(returnType);
 
                         // TODO should it be called when returnType==String?
@@ -231,8 +233,6 @@ class DefaultAiServices<T> extends AiServices<T> {
                         if ((!supportsJsonSchema || jsonSchema.isEmpty()) && !streaming && !returnsImage) {
                             userMessage = appendOutputFormatInstructions(returnType, userMessage);
                         }
-
-                        userMessage = addContentsToUserMessage(method, args, userMessage);
 
                         List<ChatMessage> messages = new ArrayList<>();
                         if (context.hasChatMemory()) {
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@qicesun thank you!

PR #4684, opened 2026-03-09
Fix ModerationModel Javadoc

jeanbisutti:
The return Javadoc tag for ModerationModel#modelName() incorrectly states the method returns null when the model name is not available, but the default implementation actually returns "unknown".


jeanbisutti:
@dliubarskyi  Could you please have a look?

dliubarskyi (approved):
@jeanbisutti thank you!

PR #4697, opened 2026-03-10
Add OpenAI Text-to-Speech (TTS) support 

anunnakian:
## Summary

  Adds **Text-to-Speech (TTS)** support to LangChain4j: a new provider-agnostic `TextToSpeechModel` abstraction in `langchain4j-core`, an OpenAI implementation backed by the [OpenAI Speech 
  API](https://platform.openai.com/docs/api-reference/audio/createSpeech), and the HTTP-client plumbing needed to return binary (audio) response bodies intact.

  ## Changes

  ### `langchain4j-core` — new abstraction (`dev.langchain4j.model.audio`, `@Experimental`)
  - **`TextToSpeechModel`** — interface mirroring the existing `AudioTranscriptionModel`:
    - `synthesize(String text)` — convenience method using the model's default voice
    - `synthesize(TextToSpeechRequest request)`
    - `provider()` defaulting to `OTHER`
  - **`TextToSpeechRequest`** — carries `text` (validated non-blank) and optional `voice`
  - **`TextToSpeechResponse`** — wraps the core `Audio` type; `from(Audio)` factory

  ### `langchain4j-open-ai` — OpenAI implementation
  - **`OpenAiTextToSpeechModel`** — configurable `modelName`, `voice` (default `alloy`), `timeout`, `maxRetries` (default `2`), request/response logging, custom `baseUrl`, and `httpClientBuilder`. Enforces
  OpenAI's 4096-character input limit. `provider()` → `OPEN_AI`.
  - **`OpenAiTextToSpeechModelName`** — `tts-1`, `tts-1-hd`, `gpt-4o-mini-tts`, `gpt-4o-mini-tts-2025-12-15`
  - **`OpenAiTextToSpeechRequest` / `OpenAiTextToSpeechResponse`** — internal DTOs (`internal.audio.texttospeech`); the response is built from the raw HTTP body (binary audio, not JSON)
  - **`OpenAiTextToSpeechModelBuilderFactory`** — SPI factory for builder customization
  - **`OpenAiClient.textToSpeech(...)`** — `POST /audio/speech`, mapping the raw bytes + `Content-Type` into the response

  ### HTTP client layer (shared) — binary response support
  - **`SuccessfulHttpResponse`** — body is now stored as `byte[]`. `body()` is preserved and returns a decoded `String` (using the charset from the `Content-Type` header, UTF-8 fallback); adds `bodyBytes()` for
  raw binary access and `contentType()`. The `body(String)` builder is kept and a `body(byte[])` overload added — **additive, no public method removed or changed**.
  - **`JdkHttpClient`** — switched to `BodyHandlers.ofByteArray()`; **`ApacheHttpClient` / `OkHttpClient`** read the body as raw bytes, so binary payloads (audio) are no longer corrupted by text decoding.
  - **`HttpResponseLogger`** — no longer decodes/dumps binary bodies as text; logs `[binary body, N bytes, content-type: ...]` instead.

  ### Tests
  - Core: `TextToSpeechModelTest`, `TextToSpeechRequestTest`
  - HTTP: `SuccessfulHttpResponseTest` (charset decoding/edge cases), `HttpClientIT` binary-response test asserting a valid MP3 is returned across all client implementations
  - OpenAI: `OpenAiTextToSpeechModelTest` (input-length and required-field validation), `OpenAiTextToSpeechModelIT` (parameterized over every supported model)

### Docs & API compatibility
  - Added a **"Creating `OpenAiTextToSpeechModel`"** section to the OpenAI docs page
  - `revapi.json` updated to acknowledge the core `TextToSpeechRequest`/`TextToSpeechResponse` types intentionally exposed by `OpenAiTextToSpeechModel`

  ## Design notes
  - Named `TextToSpeechModel` / `synthesize(...)` for discoverability (TTS is the universally recognized term) and to read as a clear inverse of the existing `AudioTranscriptionModel` / `transcribe(...)`.
  - The shared HTTP-client change is **behavior-preserving**: for every client (JDK / Apache / OkHttp), responses without an explicit charset were already decoded as UTF-8, and explicit charsets are still
  honored — verified empirically against `httpclient5 5.6.1` / `httpcore5 5.4`. The `String → byte[]` move is additive at the public API level.

  ## General checklist
  - [x] There are no breaking changes (API, behaviour) — the shared `SuccessfulHttpResponse` change is additive and verified behavior-preserving; please review the HTTP-client layer regardless, as it affects all
  providers
  - [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and


http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
```diff
@@ -48,10 +48,10 @@ public SuccessfulHttpResponse execute(HttpRequest request) throws HttpException
         try {
             java.net.http.HttpRequest jdkRequest = toJdkRequest(request);
 
-            java.net.http.HttpResponse<String> jdkResponse = delegate.send(jdkRequest, BodyHandlers.ofString());
```
anunnakian, line 51: @dliubarskyi do you have better idea to avoid converting http response to string, and get directly a binary response ? I need to handle OpenAi TTS response

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
```diff
@@ -27,7 +29,14 @@ public Map<String, List<String>> headers() {
...
     public String body() {
-        return body;
+        if (cachedStringBody == null) {
+            cachedStringBody = new String(body);
+        }
+        return cachedStringBody;
+    }
+
+    public byte[] bodyBytes() {
+        return Arrays.copyOf(body, body.length);
```
anunnakian: @dliubarskyi I need your feedback about that changes as well, I'm not really satisfied by using a cached string and handle the body as a byte array. any thoughts ?

http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
```diff
@@ -48,17 +48,17 @@ public SuccessfulHttpResponse execute(HttpRequest request) throws HttpException
...
+            java.net.http.HttpResponse<byte[]> jdkResponse = delegate.send(jdkRequest, BodyHandlers.ofByteArray());
 
             if (!isSuccessful(jdkResponse)) {
-                throw new HttpException(jdkResponse.statusCode(), jdkResponse.body());
+                throw new HttpException(jdkResponse.statusCode(), new String(jdkResponse.body()));
```
dliubarskyi: Shouldn't we specify encoding explicitly?

http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
dliubarskyi: Not at the moment, I think this is a reasonable solution. Or do you have doubts?

langchain4j-http-client/src/main/java/dev/langchain4j/http/client/SuccessfulHttpResponse.java
dliubarskyi: Could you please explain why do you want to cache the string? Usually `body()` method is called only once per response, so in most cases there is no benefit of caching

anunnakian: You're right, I don't see any benefit

http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
anunnakian: No doubts, let's keep it like this

http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
anunnakian: Yes, my bad

anunnakian:
@dliubarskyi I handled all your comments, but I've noticed that CI "Java CI / integration_test (21) (pull_request)" run forever, is there any problem ?

dliubarskyi (approved):
@anunnakian thank you!

PR #4699, opened 2026-03-11
Azure OpenAI: add maxCompletionTokens to allow setting the token limit for GPT 5 models

ChrisZachariadis:
## Issue

When trying to specify the maximum tokens in the ChatRequest parameters, e.g:
```
										 .messages(systemPrompt, userPrompt)
										 .maxOutputTokens(4096)
										 .build();
```
Then there's the following exception when using gpt 5 models:
```
dev.langchain4j.exception.InvalidRequestException: Status code 400, "{
  "error": {
    "message": "Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.",
    "param": "max_tokens",
    "code": "unsupported_parameter"
  }
}"
```
## Change

Properly passing maxCompletionTokens to ChatRequest and AzureOpenAiChatModel

## General checklist


langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/DefaultChatRequestParameters.java
```diff
@@ -36,6 +37,7 @@ protected DefaultChatRequestParameters(Builder<?> builder) {
         this.frequencyPenalty = builder.frequencyPenalty;
         this.presencePenalty = builder.presencePenalty;
         this.maxOutputTokens = builder.maxOutputTokens;
+        this.maxCompletionTokens = builder.maxCompletionTokens;
```
dliubarskyi, line 40: `maxCompletionTokens` is an OpenAI-specific parameter, it should not be present in a generic `ChatRequestParameters`.

You can create an `AzureOpenAiChatRequestParameters`, similar to `OpenAiChatRequestParameters`, and have maxCompletionTokens there

langchain4j-core/src/main/java/dev/langchain4j/model/catalog/ModelDescription.java
```diff
@@ -28,6 +28,7 @@ public class ModelDescription {
     private final ModelType type;
     private final Integer maxInputTokens;
     private final Integer maxOutputTokens;
+    private final Integer maxCompletionTokens;
```
dliubarskyi, line 31: Are you sure about this change?

dliubarskyi (commented):
@ChrisZachariadis thanks, please check my comments. This PR also lacks tests and documentation

PR #4381, opened 2026-01-07
Avoid overwriting a structured state with an unstructured argument generated from supervisor's LLM response

mariofusco:
## Issue

Closes #4375

## Change

See https://github.com/langchain4j/langchain4j/issues/4375#issuecomment-3718417570

## General checklist

- [X] There are no breaking changes (API, behaviour)
- [X] I have added unit and/or integration tests for my change
- [X] The tests cover both positive and negative cases
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new maven module

- [ ] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #4712, opened 2026-03-14
Fix reversed isAssignableFrom in writeArgumentToScope type guard

### LINKED ISSUE #4686: [BUG] writeArgumentToScope() type guard blocks repeated @Agent invocations when adaptValueToType() converts String to Map
## Describe the bug

## Root cause

## Suggested fix

## Workaround

## Version

## Related

DragonFSKY:
## Summary

Fixes #4686.

The type guard in `SupervisorPlanner.writeArgumentToScope()` (line 146) used `value.getClass().isAssignableFrom(argType)` which checks "is argType a subclass of value?" — the wrong direction. This caused repeated invocations of the same agent to silently reuse stale arguments from the first call.

- Flip to `argType.isAssignableFrom(value.getClass())` to correctly check "is value compatible with argType?"
- Add `value != null` guard to prevent NPE when planner produces null arguments
- The #4381 POJO protection semantics are fully preserved

## Test plan

- [x] 11 unit tests covering all branches of `writeArgumentToScope` via reflection on real `SupervisorPlanner` + `DefaultAgenticScope`:
  - 4 compatible overwrite (Map, List, Object, String)
  - 2 POJO protection regression (#4381)
  - 1 incompatible existing value
  - 1 null value blocking
  - 1 first write (no existing state)
  - 1 undeclared key
  - 1 Number subtype compatibility

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
@@ -143,7 +143,8 @@ public class SupervisorPlanner implements Planner, ChatMemoryAccessProvider {
             if (argType != null) {
                 Object existingValue = agenticScope.readState(key);
                 // avoid overwriting a structured state with an unstructured argument generated from supervisor's LLM response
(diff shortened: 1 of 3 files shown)


DragonFSKY:
Closing for now — should wait for confirmation on the approach before submitting a PR. See discussion in #4686.

PR #4713, opened 2026-03-14
Fix reversed isAssignableFrom in writeArgumentToScope type guard

### LINKED ISSUE #4686: [BUG] writeArgumentToScope() type guard blocks repeated @Agent invocations when adaptValueToType() converts String to Map

DragonFSKY:
## Issue
Closes #4686

## Change
The `writeArgumentToScope()` type guard in `SupervisorPlanner` had two bugs:

1. **Reversed `isAssignableFrom` direction** — `value.getClass().isAssignableFrom(argType)` checks "is argType a subclass of value's class?" instead of the intended "is value compatible with argType?" This caused repeated invocations of the same agent to be blocked when the argument type is a parent class (e.g., `Map`, `List`, `Number`), because a `LinkedHashMap` value would fail the reversed check against `Map` argType.

2. **Missing null guard** — `value.getClass()` throws `NullPointerException` when `value` is null. Additionally, `existingValue` can be null when a `DelayedResponse` from an async agent resolves to null.

The fix:
- Flips `isAssignableFrom` to the correct direction: `argType.isAssignableFrom(value.getClass())`
- Adds null guard for `value` (null cannot overwrite typed state)
- Adds null guard for `existingValue` (allow overwrite when async result resolved to null)

The original protection from #4381 is preserved: unstructured LLM output (Map/String) still cannot overwrite structured POJO state.

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
@@ -143,7 +143,8 @@ public class SupervisorPlanner implements Planner, ChatMemoryAccessProvider {
             if (argType != null) {
                 Object existingValue = agenticScope.readState(key);
                 // avoid overwriting a structured state with an unstructured argument generated from supervisor's LLM response
-                return !argType.isAssignableFrom(existingValue.getClass()) || value.getClass().isAssignableFrom(argType);
+                return !argType.isAssignableFrom(existingValue.getClass())
+                        || (value != null && argType.isAssignableFrom(value.getClass()));
             }
         }
         return true;
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
@@ -155,6 +155,9 @@ public class SupervisorPlanner implements Planner, ChatMemoryAccessProvider {
                     .orElse(null);
             if (argType != null) {
                 Object existingValue = agenticScope.readState(key);
+                if (existingValue == null) {
+                    return true;
+                }
                 // avoid overwriting a structured state with an unstructured argument generated from supervisor's LLM
                 // response
                 return !argType.isAssignableFrom(existingValue.getClass())
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
@@ -1,9 +1,7 @@
 package dev.langchain4j.agentic.supervisor;
 
-import java.util.List;
-import java.util.Map;
-import java.util.function.Function;
-import java.util.stream.Collectors;
+import static java.util.stream.Collectors.toMap;
+
 import dev.langchain4j.agentic.internal.Context;
 import dev.langchain4j.agentic.planner.Action;
 import dev.langchain4j.agentic.planner.AgentArgument;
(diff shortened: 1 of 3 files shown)


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
```diff
@@ -139,11 +150,18 @@ private boolean writeArgumentToScope(AgenticScope agenticScope, AgentInstance ag
...
                     .filter(arg -> arg.name().equals(key))
-                    .findFirst().map(AgentArgument::rawType).orElse(null);
+                    .findFirst()
+                    .map(AgentArgument::rawType)
+                    .orElse(null);
             if (argType != null) {
                 Object existingValue = agenticScope.readState(key);
-                // avoid overwriting a structured state with an unstructured argument generated from supervisor's LLM response
-                return !argType.isAssignableFrom(existingValue.getClass()) || value.getClass().isAssignableFrom(argType);
+                if (existingValue == null) {
```
mariofusco, line 158: I don't see how `existingValue` could be null provided that `agenticScope.hasState(key)` is `true` and then we already checked that there is a state for that key.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/WriteArgumentToScopeTest.java
```diff
@@ -0,0 +1,211 @@
...
+/**
+ * Tests the type guard logic in {@code SupervisorPlanner.writeArgumentToScope()}.
+ *
+ * <p>Regression tests for <a href="https://github.com/langchain4j/langchain4j/issues/4686">#4686</a>:
+ * the guard blocked repeated invocations of the same agent because
+ * {@code isAssignableFrom} was called in the wrong direction.
+ *
+ * @see <a href="https://github.com/langchain4j/langchain4j/pull/4381">#4381</a> (introduced the guard)
+ */
+class WriteArgumentToScopeTest {
```
mariofusco, line 31: I don't see the point in all these tests calling an internal method of the `SupervisorPlanner`. I would prefer an integration test showing the bug using the supervisor agent instead, but I'm not able to create one at the moment.

mariofusco:
Replaced by https://github.com/langchain4j/langchain4j/pull/4722

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
DragonFSKY: Hi @mariofusco, thanks for the review.

While reading through the related code, I noticed that `hasState()` and `readState()` behave differently when the value is a `DelayedResponse` from an async agent:

```java
// hasState() does not resolve DelayedResponse
public boolean hasState(String key) {
    Object value = state.get(key);          // AsyncResponse object, non-null
    if (value == null) { return false; }
    return value instanceof String s ? !s.isBlank() : true;
}

// readState() resolves it — blockingGet() can return null
private Object readStateBlocking(String key, Object state) {
    if (state instanceof DelayedResponse asyncResponse) {
        state = asyncResponse.blockingGet();   // could be null
        writeState(key, state);                 // null removes the key
    }
    return state;
}
```

`hasState()` sees the placeholder and returns true, but `readState()` resolves it to null. Without this guard, `existingValue.getClass()` would NPE.

Covered by test: `should_allow_overwrite_when_existing_value_resolves_to_null` in `WriteArgumentToScopeTest`.

That said, my understanding of this area may be incomplete — happy to hear if you think it should be handled differently.

mariofusco: The supervisor agent should never use async calls. It requires all former agents to have completed their work before planning for the next call. In a former release this check was implemented and when the supervisor bumped in an async agent it emitted a warning saying that this is not allowed and calling the agent synchronously. I fear that this mechanism went lost with another refactor (very likely when I introduced the `Planner` interface). I will double check and if so I will reintroduce it.

PR #4716, opened 2026-03-15
Preserve augmented UserMessage in tool loop when storeRetrievedContentInChatMemory is false

### LINKED ISSUE #4714: How to provide context in the tool call without storing it in the memory
We have a RAG system, where we don't want to store the retrieved context in the memory, but this means that the context is also not provided after tool calling. In the example,

```Java
public class ToolCalling {

    static class Calculator {
        @Tool("Calculates the sum of two numbers")
        int add(int a, int b) {
            System.out.println("Called add with a=" + a + ", b=" + b);
            return a + b;
        }
    }

    interface Assistant {
    }

        ChatModel model = OpenAiChatModel.builder()
                .baseUrl("...")
                .listeners(new ChatModelListener() {
                               @Override
                               public void onRequest(ChatModelRequestContext requestContext) {
                                   requestContext.chatRequest().messages().forEach(System.out::println);
                               }
                           })
                .build();

                .chatModel(model)
                .tools(new Calculator())
                .storeRetrievedContentInChatMemory(true)
                .chatMemory(MessageWindowChatMemory.withMaxMessages(10))
                .contentRetriever(query -> List.of(Content.from("Say Hallo World!")))
                .build();

        String answer = assistant.chat("Call the tool to add the numbers 2 and 6");
        System.out.println("Answer:" + answer);

    }

}
```

the output for `.storeRetrievedContentInChatMemory(true)` is "Hallo World! The sum of 2 and 6 is 8.". For `.storeRetrievedContentInChatMemory(false)` it is "The sum of 2 and 6 is 8.". Here, the message shows that after tool calling, the context is not added 

```
UserMessage { name = null, contents = [TextContent { text = "Call the tool to add the numbers 2 and 6

Answer using the following information:
Say Hallo World!" }], attributes = {} }

Called add with a=2, b=6

UserMessage { name = null, contents = [TextContent { text = "Call the tool to add the numbers 2 and 6" }], attributes = {} }

AiMessage { text = null, thinking = null, toolExecutionRequests = [ToolExecutionRequest { id = "CWTlmzKu4mfAxEZ0N8DNCkOudoJLPxIx", name = "add", arguments = "{"arg0":2,"arg1":6}" }], attributes = {} }

ToolExecutionResultMessage{id='CWTlmzKu4mfAxEZ0N8DNCkOudoJLPxIx', toolName='add', text='8', isError=false, attributes={}}
```

Is there a clean/good way to add the context after tool calling, without the need to store it in the memory?

Thanks for your help!

DragonFSKY:
## Issue

Closes #4714

## Change

When `storeRetrievedContentInChatMemory` is set to `false`, the original (non-augmented) `UserMessage` is stored in `ChatMemory`. During tool execution loops, reloading messages via `chatMemory.messages()` loses the RAG-augmented content, causing the LLM to lose context on follow-up model calls.

This PR captures the augmented `UserMessage` before memory storage and replays it after each `chatMemory.messages()` reload in both sync (`ToolService`) and streaming (`AiServiceStreamingResponseHandler`) paths. When `storeRetrievedContentInChatMemory` is `true` (default) or no chat memory / tools are configured, the behavior is unchanged (no-op).

All modified classes are `@Internal` — no public API changes.

## General checklist

- [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (if applicable)


dliubarskyi (commented):
@DragonFSKY thank you!

I am not sure explicitly passing `UserMessage userMessageForToolReplay` through all the API layers is right. What about putting it inside the `InvocationContext`? We could add setters for original and augmented `UserMessage`s and expose both from the `InvocationContext`. WDYT?

dliubarskyi:
I am working on it, going to push the change soon

dliubarskyi (approved):
@DragonFSKY thank you!

PR #4722, opened 2026-03-16
Fix arguments types comparison in supervisor's writeArgumentToScope method

### LINKED ISSUE #4686: [BUG] writeArgumentToScope() type guard blocks repeated @Agent invocations when adaptValueToType() converts String to Map

mariofusco:
## Issue

Closes #4686

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
@@ -143,7 +143,7 @@ public class SupervisorPlanner implements Planner, ChatMemoryAccessProvider {
             if (argType != null) {
                 Object existingValue = agenticScope.readState(key);
                 // avoid overwriting a structured state with an unstructured argument generated from supervisor's LLM response
-                return !argType.isAssignableFrom(existingValue.getClass()) || value.getClass().isAssignableFrom(argType);
+                return !argType.isAssignableFrom(existingValue.getClass()) || argType.isAssignableFrom(value.getClass());
             }
         }
         return true;


dliubarskyi (approved):
As far as I remember, this is not the first time we face this kind of issue, perhaps let's extract this check into a utility method with a better name?

DragonFSKY:
Following up on the comment in #4686 about needing an integration test for this — I wrote one that covers the fixed scenario through the full supervisor pipeline.

A mock `ChatModel` makes the planner invoke the same agent twice with the same key but different values, and asserts the agent receives the updated data on the second call.

- Fails before this fix (`expected: 2, actual: 1` — second write silently blocked)
- Passes after this fix

```java
public class WriteArgumentToScopeIT {

    public static class DataProcessor {
        private Map<String, Object> lastReceived;

        @Agent(description = "Processes data and writes it to scope", outputKey = "data")
        public Map<String, Object> process(@V("data") Map<String, Object> data) {
            this.lastReceived = data;
            return data;
        }

        public Map<String, Object> getLastReceived() {
            return lastReceived;
        }
    }

    static class RepeatedInvocationPlannerModel implements ChatModel {
        private final AtomicInteger callCount = new AtomicInteger(0);

        @Override
        public ChatResponse chat(ChatRequest request) {
            String json =
                    switch (callCount.incrementAndGet()) {
                        case 1 -> """
                                { "agentName": "process", "arguments": { "data": {"id": 1} } }
                                """;
                        case 2 -> """
                                { "agentName": "process", "arguments": { "data": {"id": 2} } }
                                """;
                        default -> """
                                { "agentName": "done", "arguments": {} }
                                """;
                    };
            return ChatResponse.builder().aiMessage(AiMessage.from(json)).build();
        }
    }

    @Test
    void should_allow_repeated_invocation_of_same_agent_with_compatible_arguments() {
        DataProcessor dataProcessor = new DataProcessor();

        SupervisorAgent supervisor = AgenticServices.supervisorBuilder()
                .chatModel(new RepeatedInvocationPlannerModel())
                .responseStrategy(SupervisorResponseStrategy.LAST)
                .subAgents(dataProcessor)
                .build();

        ResultWithAgenticScope<String> result = supervisor.invokeWithAgenticScope("test");

        assertThat(dataProcessor.getLastReceived()).containsEntry("id", 2);
    }
}
```

I can open a follow-up PR to add this if it's worth having.

PR #4729, opened 2026-03-17
Make ElasticsearchConfiguration extensible for custom implementations

### LINKED ISSUE #4725: [Fix] Make ElasticsearchConfiguration extensible for custom implementations (documentation vs implementation mismatch)
I'm trying to create a custom Elasticsearch configuration for my specific use case, following the official documentation at https://docs.langchain4j.dev/integrations/embedding-stores/elasticsearch#creating-custom-configurations. The documentation states that I can "create your own Elasticsearch configuration by implementing the ElasticsearchConfiguration interface" and provides an example with `implements ElasticsearchConfiguration`.

However, when I look at the actual code implementation in `ElasticsearchConfiguration.java`, I found that it's an abstract class with package-private abstract methods, not an interface:

```java
public abstract class ElasticsearchConfiguration {
    abstract SearchResponse<Document> vectorSearch(...);
    abstract SearchResponse<Document> fullTextSearch(...);
    abstract SearchResponse<Document> hybridSearch(...);
}
```

This creates two major problems:
1. **Cannot implement as an interface**: Since it's an abstract class, I cannot implement it as shown in the documentation.
2. **Cannot extend from another package**: The abstract methods are package-private, so even if I try to extend this class from another package, I cannot override these methods.

This inconsistency between documentation and implementation makes it impossible to create custom configurations as described.

Additionally, the `ElasticsearchConfigurationHybrid` class (which extends `ElasticsearchConfiguration`) has a private constructor, further complicating any attempt to extend and customize the behavior:

```java
private ElasticsearchConfigurationHybrid(Integer numCandidates, boolean includeVectorResponse) {
    this.numCandidates = numCandidates;
    this.includeVectorResponse = includeVectorResponse;
}
```

I propose one of the following solutions:

1. **Convert `ElasticsearchConfiguration` to an interface** (preferred): 
   - Change the abstract class to an interface, matching the documentation
   - This would allow users to implement custom configurations as shown in the docs
   - Make the constants (`VECTOR_FIELD`, `TEXT_FIELD`) available in a separate utility class or as static interface fields

2. **Make the abstract methods protected**:
   - If keeping it as an abstract class, change the visibility of the abstract methods from package-private to `protected`
   - This would at least allow extension from other packages

3. **Make `ElasticsearchConfigurationHybrid` extensible**:
   - Change its constructor from `private` to `protected`
   - This would allow users to extend the hybrid configuration and customize its behavior

1. **Using composition instead of inheritance**: I attempted to create a wrapper class that delegates to the original configuration, but since the methods in `ElasticsearchEmbeddingStore` directly check for specific configuration types (`instanceof ElasticsearchConfigurationFullText`, `instanceof ElasticsearchConfigurationHybrid`), composition doesn't work as the wrapper instances won't be recognized as the expected types.

2. **Copying the entire class**: I considered copying the entire `ElasticsearchConfigurationHybrid` class into my own package, but this would lead to code duplication and maintenance issues, and I would lose compatibility with future updates.

3. **Using reflection**: I explored using reflection to access and modify the private constructor and methods, but this is fragile, error-prone, and goes against good software design practices.

**Additional context**

This issue affects anyone trying to customize Elasticsearch search behavior in LangChain4j, such as:
- Customizing the hybrid search algorithm
- Adding custom query boosting logic
- Implementing specialized filtering mechanisms
- Integrating with specific Elasticsearch features or versions

The documentation explicitly promises this capability, making it a documentation vs. implementation inconsistency that should be resolved to maintain developer trust and framework usability.

Here's the link to the relevant documentation: https://docs.langchain4j.dev/integrations/embedding-stores/elasticsearch#creating-custom-configurations
And the code reference: https://github.com/langchain4j/langchain4j/blob/main/langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchConfiguration.java

dadoonet:
## Issue

Closes #4725.

## Change

This PR switches `ElasticsearchConfiguration` from an abstract class to an interface with default methods throwing an `UnsupportedOperationException`.

This allows creating a self-made implementation. See the `ElasticsearchConfigurationCustomIT` test and its `ElasticsearchConfigurationCustom` class.

```java
public class ElasticsearchConfigurationCustom implements ElasticsearchConfiguration {
    @Override
    public SearchResponse<Document> vectorSearch(
            ElasticsearchClient client, String indexName, EmbeddingSearchRequest embeddingSearchRequest)
            throws ElasticsearchException {
        // Implement here
    }
}
```

## General checklist

### DIFF
--- a/docs/docs/integrations/embedding-stores/elasticsearch.md
+++ b/docs/docs/integrations/embedding-stores/elasticsearch.md
@@ -263,6 +263,12 @@ public class MyElasticsearchConfiguration implements ElasticsearchConfiguration 
 }
 ```
 
+Please note that you can implement only the methods relevant to your use case:
+
+* `vectorSearch` for vector similarity search (used by both `ElasticsearchEmbeddingStore` and `ElasticsearchContentRetriever`).
+* `fullTextSearch` for full text search (used by `ElasticsearchContentRetriever` only).
+* `hybridSearch` for hybrid search (used by `ElasticsearchContentRetriever` only).
+
 ## Examples
 
 - [ElasticsearchEmbeddingStoreExample](https://github.com/langchain4j/langchain4j-examples/blob/main/elasticsearch-example/src/main/java/ElasticsearchEmbeddingStoreExample.java)
--- a/langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/AbstractElasticsearchEmbeddingStore.java
+++ b/langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/AbstractElasticsearchEmbeddingStore.java
@@ -145,7 +145,7 @@ public abstract class AbstractElasticsearchEmbeddingStore implements EmbeddingSt
             return new EmbeddingSearchResult<>(results);
         } catch (ElasticsearchException e) {
             if (e.getLocalizedMessage().contains("Unknown key for a VALUE_BOOLEAN in [exclude_vectors]")
-                    && this.configuration.includeVectorResponse) {
+                    && this.configuration.isIncludeVectorResponse()) {
                 log.warn(
                         "Property [includeVectorResponse] is not needed for elasticsearch server versions previous to 9.2, remove it to fix the exception.");
             }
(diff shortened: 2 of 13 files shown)


dliubarskyi (approved):
@dadoonet thank you!

PR #4732, opened 2026-03-18
Skills: support skill-scoped tools

dliubarskyi:
#### Skill-Scoped Tools

You can attach tools directly to a skill. These tools are **only exposed to the LLM
after the skill has been activated** via the `activate_skill` tool.
This keeps the LLM's tool list small and focused, and ensures skill-specific tools only
appear when they are relevant.

##### Using `@Tool`-Annotated Methods

The simplest way to attach tools is to pass objects with `@Tool`-annotated methods:

```java
class OrderTools {

    @Tool("Validates a customer order by ID")
    String validateOrder(String orderId) {
        // validation logic
        return "valid";
    }

    @Tool("Charges payment for a customer order")
    String chargePayment(String orderId) {
        // payment logic
        return "charged";
    }
}

Skill skill = Skill.builder()
        .name("process-order")
        .description("Processes a customer order end-to-end")
        .content("""
                To process an order:
                1. Call `validateOrder(orderId)` to check the order is valid.
                2. Call `chargePayment(orderId)`.
                """)
        .tools(new OrderTools())
        .build();
```

Tools can also be attached to an already-built skill using `toBuilder()` — for example,
to add tools to a skill loaded from the file system:

```java
FileSystemSkill skill = FileSystemSkillLoader.loadSkill(Path.of("skills/process-order"));

Skill skillWithTools = skill.toBuilder()
        .tools(new OrderTools())
        .build();
```

##### Using Tool Providers

You can also attach `ToolProvider`s to a skill — for example, to expose tools from an
MCP server only after the skill is activated:

```java
ToolProvider mcpToolProvider = McpToolProvider.builder()
        .mcpClients(mcpClient)
        .toolFilter((tool, mcpClient) -> tool.name().startsWith("inventory_"))
        .build();

Skill skill = Skill.builder()
        .name("inventory-management")
        .description("Manages warehouse inventory")
        .content("""
                Use inventory tools to check stock levels and update quantities.
                """)
        .toolProviders(mcpToolProvider)
        .build();
```

##### Using a `Map<ToolSpecification, ToolExecutor>`

For full control over tool specifications and execution logic, you can pass a map directly:

```java
ToolSpecification validateOrder = ToolSpecification.builder()
        .name("validateOrder")
        .description("Validates a customer order by ID")
        .addParameter("orderId", JsonSchemaProperty.STRING, JsonSchemaProperty.description("The order ID"))
        .build();

ToolExecutor validateOrderExecutor = (request, memoryId) -> {
    String orderId = parseOrderId(request.arguments());
    return validate(orderId);
};

Skill skill = Skill.builder()
        .name("process-order")
        .description("Processes a customer order end-to-end")
        .content("""
                To process an order:
                1. Call `validateOrder(orderId)` to check the order is valid.
                """)
        .tools(Map.of(validateOrder, validateOrderExecutor))
        .build();
```

All three approaches can be combined — `@Tool` methods, `ToolProvider`s, and `Map` entries
are merged into a single set of skill-scoped tools:

```java
Skill skill = Skill.builder()
        .name("process-order")
        .description("Processes a customer order end-to-end")
        .content("...")
        .tools(new OrderTools())
        .tools(Map.of(validateOrder, validateOrderExecutor))
        .toolProviders(mcpToolProvider)
        .build();
```

##### Wiring It Up

```java
Skills skills = Skills.from(skill);

MyAiService service = AiServices.builder(MyAiService.class)
        .chatModel(chatModel)
        .chatMemory(MessageWindowChatMemory.withMaxMessages(100))
        .toolProvider(skills.toolProvider())
        .systemMessage("You have access to the following skills:\n" + skills.formatAvailableSkills()
                + "\nWhen the user's request relates to one of these skills, activate it first.")
        .build();
```

##### How Skill-Scoped Tools Work

1. Before skill activation, the LLM only sees the `activate_skill` (and `read_skill_resource`) tools.
   Skill-scoped tools are not included in the tool list.
2. When the LLM calls `activate_skill("process-order")`, the activation is recorded in the `ToolExecutionResultMessage`.
3. Before the next LLM call (within the same AI Service invocation), the AI Service re-evaluates dynamic tool providers
   against the current messages. The skill-scoped tools (e.g. `validateOrder`) become
   visible and the LLM can call them immediately, in the same AI Service invocation.
   The skill-scoped tools stay visible to the LLM in the next AI Service invocations, they become invisible only when
   the skill is deactivated.

##### Using Skills with Tool Search

Skills work alongside [Tool Search](/tutorials/tools#tool-search). When both are configured,
they operate independently:

- **Skill-scoped tools are never searchable.** They don't appear in the searchable tool pool
  and cannot be found via `tool_search_tool`. They only become visible after the LLM activates
  the corresponding skill.
- **Regular tools remain searchable.** Tools registered via `.tools(...)` on the AI Service
  (not on a skill) continue to be searchable, regardless of whether any skill is activated.
- **`activate_skill` is always visible.** It is marked as `ALWAYS_VISIBLE`, so the LLM can
  always call it even when Tool Search is enabled.

```java
Skills skills = Skills.from(mySkills);

MyAiService service = AiServices.builder(MyAiService.class)
        .chatModel(chatModel)
        .chatMemory(MessageWindowChatMemory.withMaxMessages(100))
        .tools(new MySearchableTools()) // these are searchable
        .toolProvider(skills.toolProvider()) // skill-scoped tools are NOT searchable
        .toolSearchStrategy(new SimpleToolSearchStrategy())
        .systemMessage("You have access to the following skills:\n" + skills.formatAvailableSkills()
                + "\nWhen the user's request relates to one of these skills, activate it first.")
        .build();
```

## General checklist


langchain4j/src/main/java/dev/langchain4j/service/tool/search/ToolSearchService.java
```diff
@@ -42,11 +43,11 @@ public ToolSearchService(ToolSearchStrategy toolSearchStrategy) {
     }
 
     public ToolServiceContext adjust(ToolServiceContext toolServiceContext,
-                                     ChatMemory chatMemory,
+                                     List<ChatMessage> messages,
```
dliubarskyi, line 46: This is one of the breaking changes, but since it is not used directly by Quarkus, and is located in `@Internal` class and was introduced in the last release, I think it is ok to do it in-place

langchain4j/src/main/java/dev/langchain4j/service/tool/search/ToolSearchService.java
```diff
@@ -108,26 +109,25 @@ private Map<String, ToolExecutor> createExecutors(List<ToolSpecification> toolSe
         return executors;
     }
 
-    public static ChatRequestParameters addFoundTools(ChatRequestParameters parameters,
-                                                      Collection<ToolExecutionResult> toolResults,
-                                                      List<ToolSpecification> availableTools) {
+    public static ToolServiceContext addFoundTools(ToolServiceContext toolServiceContext,
```
dliubarskyi, line 112: same here

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -246,62 +258,69 @@ public void toolSearchStrategy(ToolSearchStrategy toolSearchStrategy) {
...
+        return context;
     }
 
-    /**
-     * @deprecated use {@link #createContext(InvocationContext, UserMessage, ChatMemory)} instead
-     */
-    @Deprecated(since = "1.12.0")
-    public ToolServiceContext createContext(InvocationContext invocationContext, UserMessage userMessage) {
-        // TODO make private
+    private ToolServiceContext createContextFromStaticToolsAndProviders(InvocationContext invocationContext,
```
dliubarskyi, line 270: same here

langchain4j-skills/src/main/java/dev/langchain4j/skills/Skill.java
```diff
@@ -33,9 +34,24 @@ public interface Skill {
...
+        return List.of();
+    }
+
+    /**
+     * Returns the optional list of tool providers associated with this skill.
+     * These tool providers supply tools that will be exposed to the LLM when this skill is activated.
+     *
+     * @return the list of tool providers, empty by default
+     */
+    default List<ToolProvider> toolProviders() {
```
jmartisk, line 52: I would personally just add new abstract methods instead of providing default implementations as I wouldn't expect anybody to be broken by this yet. Or is there a real reason for defaulting to empty lists apart from just being able to maintain backward compatibility?

dliubarskyi: The real reason is that both resources and tools are optional to the skill, so enforcing user to provide an implementation in any case seems wrong

jmartisk: That's a matter of taste I guess, I would have preferred to require implementors to provide an implementation :) but ok

jmartisk:
I'm thinking how this will be handled by Quarkus because right now it just detects skill files automatically and generates a `ToolProvider` for them, but if the app uses this declarative approach, then how should we detect which tools are scoped to the skill? Probably we could add a config property like

```
quarkus.langchain4j.skills.scoped-tools.SKILL_NAME=fully.qualified.name.of.tool.class,another.class.with.tools
```

and we would attach these tool classes to the skill identified by name... @mariofusco @geoand WDYT?

geoand:
I can't say I understand the problem. Can you show a hypothetical example?

jmartisk:
Let's say there is a skill somewhere in the filesystem, say, `skills/poem-writing/SKILL.md` and the `application.properties` contains `quarkus.langchain4j.skills.directories=skills` to locate that skill.
That generates a `ToolProvider` that provides access to the skill. But how should it work if the skill should contain some skill-scoped tools? 
Imagine there is a class `org.acme.PoemWritingTools` that contains some methods with `@Tool`.
My idea is that "attaching" that class to the skill would be done by `quarkus.langchain4j.skills.scoped-tools.poem-writing=org.acme.PoemWritingTools`. And only when the `poem-writing` skill is activated, then will the tools from `org.acme.PoemWritingTools` appear in the context. Any better idea how to do the mapping?

geoand:
Does the mapping have to be from scoped tool to class? Can't it be the other way around?

jmartisk:
It could, but it would be much more complicated, we would need a new annotation, like `@ScopedToSkill("poem-writing")` that one would apply on the tools and we would scan that at build time (otherwise all the configuration is runtime, which I assume is desirable in this case because different deployments may contain different sets of skills)

geoand:
> which I assume is desirable in this case because different deployments may contain different sets of skills

If this is really the case, then what you propose makes sense to me

langchain4j-core/src/main/java/dev/langchain4j/data/message/UserMessage.java
```diff
@@ -383,4 +386,11 @@ public static UserMessage userMessage(List<Content> contents) {
     public static UserMessage userMessage(String name, List<Content> contents) {
         return from(name, contents);
     }
+
+    public static Optional<UserMessage> findLast(Collection<ChatMessage> messages) {
+        return messages.stream()
+                .filter(message -> message instanceof UserMessage)
+                .map(message -> (UserMessage) message)
+                .reduce((first, second) -> second);
+    }
```
mariofusco, line 395: Nit: using method references can make this method a bit more readable, something like

```
public static Optional<UserMessage> lastUserMessage(Collection<ChatMessage> messages) {
    return messages.stream()
            .filter(UserMessage.class::isInstance)
            .map(UserMessage.class::cast)
            .reduce((first, second) -> second);
}
```

langchain4j/src/main/java/dev/langchain4j/service/tool/AiServiceTool.java
```diff
@@ -0,0 +1,72 @@
...
+ * <ul>
+ *     <li>{@link ToolSpecification} — what is sent to the LLM</li>
+ *     <li>{@link ToolExecutor} — what is called when the LLM invokes the tool</li>
+ *     <li>Metadata that controls how the AI Service handles this tool (e.g., immediate return)</li>
+ * </ul>
+ *
+ * @since 1.13.0
+ */
+@Internal
+public class AiServiceTool {
```
mariofusco, line 19: Could this be a `record`?

mariofusco:
> Let's say there is a skill somewhere in the filesystem, say, `skills/poem-writing/SKILL.md` and the `application.properties` contains `quarkus.langchain4j.skills.directories=skills` to locate that skill. That generates a `ToolProvider` that provides access to the skill. But how should it work if the skill should contain some skill-scoped tools? Imagine there is a class `org.acme.PoemWritingTools` that contains some methods with `@Tool`. My idea is that "attaching" that class to the skill would be done by `quarkus.langchain4j.skills.scoped-tools.poem-writing=org.acme.PoemWritingTools`. And only when the `poem-writing` skill is activated, then will the tools from `org.acme.PoemWritingTools` appear in the context. Any better idea how to do the mapping?

I'm hoping that my proposal could solve or mitigate the problem, but at the moment I find it odd, also from an user perspective, that to add a skill to an `AiService`, what you actually do is adding its `ToolProvider`, something like

```
Skills skills = Skills.from(FileSystemSkillLoader.loadSkills(skillsDir));
AiServices.builder(MyAiService.class)
    .toolProvider(skills.toolProvider())
    .build()
```

I understand that @dliubarskyi did this to avoid explicitly adding one more concept and method into the `AiService` itself (or maybe there are also other reasons that I'm missing at the moment?), nevertheless I'd find more natural to do something like this instead:

```
Skills skills = Skills.from(FileSystemSkillLoader.loadSkills(skillsDir));
AiServices.builder(MyAiService.class)
    .skills(skills)
    .build()
```

I believe that this alternative approach would be beneficial for both the Quarkus extension, allowing to have more control and perform Quarkus specific task when a `Skill` is added (instead of a generic `ToolProvider`) and for LangChain4j itself, making it more future proof if eventually more features will be added to the concept of a skill. What if the capabilities of a skill will be extended and providing the `AiService` with a `ToolProvider` will be no longer enough to cover them?

jmartisk:
The AI service would still call `skills.getToolProvider()` internally, no? Maybe I'm missing something but to me it sounds like you would just pass the tool provider at a different level than regular tool providers. Or do you want to get rid of the tool provider completely? (That would make skills harder to use with the low-level ChatModel API, I don't know if that's a big deal)

langchain4j-core/src/main/java/dev/langchain4j/data/message/UserMessage.java
dliubarskyi: Nice, thanks!

langchain4j/src/main/java/dev/langchain4j/service/tool/AiServiceTool.java
dliubarskyi: This class will most probably become user-facing in order to replace `Map<ToolSpecification, ToolExecutor>` ugliness and specifying `ReturnBehavior` in a separate set. And it will most probably be extended with other fields in the future. Adding new fields to a record is a breaking change, so I would prefer keeping it a normal class

dliubarskyi:
> I understand that @dliubarskyi did this to avoid explicitly adding one more concept and method into the AiService itself (or maybe there are also other reasons that I'm missing at the moment?)

Exactly, I would love to avoid adding one more concept to the core module, especially one which is not yet proven to be a "core" AI concept. Skills are all the hype now, but we are not sure if it will stick. Keeping it in a separate module makes sure we do not complicate already complicated AI Services logic and allows us to move with different speed there. It is very new, so breaking changes are easier when you keep it separately. Once it matures (perhaps for 2.0), I can imagine moving it directly to the core module and configuring skills directly on AI Service.

> What if the capabilities of a skill will be extended and providing the AiService with a ToolProvider will be no longer enough to cover them?

This is a valid concern, but I would rather not address it until it actually becomes an issue (it may never materialize).

mariofusco (dismissed):

PR #4695, opened 2026-03-10
Refactor moderation models

jeanbisutti:
Refactor `OpenAiModerationModel` and `MistralAiModerationModel `to make them more readable and facilitate the addition of new features.

PR #4733, opened 2026-03-18
Fix NPE in MistralAiModerationModel when categories is null

jeanbisutti:
The NPE was possible before #4695.


anandfresh (approved):

dliubarskyi (approved):
@jeanbisutti thanks!

PR #4738, opened 2026-03-18
add support of halfvec in pgvector

### LINKED ISSUE #4559: [BUG] [FEATURE] Add halfvec and sparsevec support to PGVectorEmbeddingStore
At the moment the PGVectorEmbeddingStore throws a `class org.postgresql.util.PGobject cannot be cast to class com.pgvector.PGvector` error when trying to retrieve from a table that has `halfvec` or `sparsevec` as the type of the embedding column.

This needs at least changing the call to `PGvector.addVectorType` to a call to PGvector.registerTypes in the `getConnection`method of `PGVectorEmbeddingStore`, in addition the search methods use the `PGvector` type explicitly, which would also need some changes.

I am also willing to do a PR for this if needed.

xxueewa:
## Issue

Fixes #4559 

add support to different vector types

## Change

1. add vector type parameter in the builder
2. update init table, create index, search functions accordingly 
3. add integration tests on cases about embedding search, hybrid search, dimension limits

## General checklist

- [] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j


langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -313,22 +338,40 @@ public static PgVectorEmbeddingStoreBuilder builder() {
...
+     * @param vectorType     The field type of embeddings
      */
     protected void initTable(
-            Boolean dropTableFirst, Boolean createTable, Boolean useIndex, Integer dimension, Integer indexListSize) {
+            Boolean dropTableFirst,
+            Boolean createTable,
+            Boolean useIndex,
+            Integer dimension,
+            Integer indexListSize,
+            VectorType vectorType) {
```
wilbrt: You don't actually need to change the signature, as you can just use the private vectorType inside the try statement like so `VectorType vectortype = getOrDefault(this.vectorType, VectorType.VECTOR);`.

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
```diff
@@ -313,22 +338,40 @@ public static PgVectorEmbeddingStoreBuilder builder() {
...
                 statement.executeUpdate(String.format("DROP TABLE IF EXISTS %s", table));
             }
             if (createTable) {
-                query = String.format(
-                        "CREATE TABLE IF NOT EXISTS %s (embedding_id UUID PRIMARY KEY, "
-                                + "embedding vector(%s), text TEXT NULL, %s )",
-                        table,
-                        ensureGreaterThanZero(dimension, "dimension"),
-                        metadataHandler.columnDefinitionsString());
+                switch (vectorType) {
```
wilbrt: I would also preferably have the `embedding halfvec(%s)` as `embedding %s(%s)` and add the name of the `vectortype` in lowercase as a parameter. That way we could a) eliminate code duplication b) ensure that the SQL-strings for different vectortypes don't drift by accident.

This goes for all the three switch statements in this method.

wilbrt (commented):
Couple of comments. Mainly I'd prefer keeping the `initTable`-method signature as it was. 

One thing to note is that you are not changing the upsert at all, so it now depends on an implicit cast by Postgres. Which for `halfvec`  might be acceptable, though a bit reckless, because this will not work for `sparsevec` so maybe at least add comment somewhere so that when someone tries to add `sparsevec` and it blows up, they know where to look.

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
xxueewa: the reason why I added this type selection is the "CREATE TABLE" query need to specifically define the embedding type, and halfvec is different from vector, see the below dimension ranges:

     * Vector Type (supported types include: vector, halfvec, sparsevec)
     * vector - up to 2,000 dimensions
     * halfvec - up to 4,000 dimensions
     * sparsevec - up to 1,000 non-zero elements
     
  Therefore, maybe it is better to provide the choice, and user could select based on their embedding method

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
xxueewa: sure, updated

langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorEmbeddingStore.java
xxueewa: Ok, got what you are saying, updated as suggested

wilbrt (approved):
LGTM 👍

PR #4755, opened 2026-03-20
Re-query ToolProviders within inference loop to support dynamic tool lists

davydotcom:
## Summary

`ToolService.executeInferenceAndToolsLoop()` currently freezes the tool list at the start of the loop (via `createContext()`) and never re-queries `ToolProvider`s, even when the underlying tool inventory changes mid-conversation. This PR adds a refresh step after each batch of tool executions so that dynamically added or removed tools are picked up before the next LLM call.

## Problem

MCP servers that implement a **dynamic category-loader pattern** expose a small set of meta-tools initially (e.g. `use_instances_tools`, `use_clouds_tools`) and lazily load the actual tools for a category when the corresponding meta-tool is called via `tools/call`. The server signals that its tool list has changed by sending a `notifications/tools/list_changed` SSE event, and `DefaultMcpClient` correctly sets its `toolListOutOfDate` flag in response.

However, `ToolService` builds a `ToolServiceContext` once before the loop starts and wraps its maps in `Collections.unmodifiableMap/List`. The loop never re-calls `provideTools()`, so:

1. The LLM calls a category-loader tool (e.g. `use_instances_tools`)
2. `McpToolExecutor` sends `tools/call` to the MCP server
3. The server returns tool definitions as text content and sends `notifications/tools/list_changed`
4. `DefaultMcpClient` sets `toolListOutOfDate = true`
5. **But the loop never re-fetches tools** — the `ToolServiceContext` is stale
6. The LLM sees the tool definitions as text and treats them as the "answer" instead of being able to call them

This makes it impossible to use MCP servers that dynamically register tools during the conversation, a pattern that keeps LLM context small by only loading relevant tool categories on demand.

## Solution

After tool executions complete and before the next LLM request is built, `executeInferenceAndToolsLoop()` now calls a new private method `refreshToolServiceContext()` which:

1. Extracts the last `UserMessage` from the messages list (needed for `ToolProviderRequest`)
2. Re-calls `createContext()` which triggers `ToolProvider.provideTools()` on all configured providers
3. Compares the result to the current context via `equals()` — if unchanged, returns the original instance (identity preserved)
4. If changed, the caller updates both `toolServiceContext` and `parameters` (via `overrideWith()`) so the next `ChatRequest` includes the new tool specifications

### Performance characteristics

- **No-op when nothing changed:** `DefaultMcpClient.listTools()` checks `isToolListRefreshNeeded()` internally — if no `tools/list_changed` notification was received, it returns the cached tool list without an RPC call. The `ToolServiceContext.equals()` check then detects no change and the original context is kept.
- **No API changes:** This is purely an internal change to the loop logic. No method signatures, constructors, or public APIs were modified.
- **Reuses existing `createContext()` logic:** No duplication of the tool provider querying/merging code.

## Testing

This was tested against a production MCP server (Morpheus) that implements the category-loader pattern with ~50 categories and ~300+ tools across ~20 tool providers. Before this change, the LLM would receive tool definitions as text and stop. After this change, dynamically loaded tools are correctly available for the LLM to call in subsequent loop iterations.

A unit/integration test that exercises this path would need a `ToolProvider` that returns different tools on successive `provideTools()` calls. Happy to add one if desired.

## Related

- `DefaultMcpClient` already handles `notifications/tools/list_changed` correctly (sets `toolListOutOfDate` flag)
- `McpToolProvider.provideTools()` delegates to `DefaultMcpClient.listTools()` which respects the flag
- `ToolServiceContext` has proper `equals()` implementation for change detection

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
@@ -392,6 +392,22 @@ public class ToolService {
                         .build();
             }
 
+            // Re-query ToolProviders after tool executions to pick up dynamically
+            // added tools (e.g. MCP servers that update their tool list via
+            // notifications/tools/list_changed after a category-loader tool is called).
+            if (!toolProviders.isEmpty()) {
+                ToolServiceContext refreshed = refreshToolServiceContext(
+                        toolServiceContext, invocationContext, messages, chatMemory);
+                if (refreshed != toolServiceContext) {
+                    toolServiceContext = refreshed;
+                    // Update parameters so the next LLM request includes the refreshed tool specs
+                    parameters = parameters.overrideWith(
+                            ChatRequestParameters.builder()
+                                    .toolSpecifications(toolServiceContext.effectiveTools())
+                                    .build());
+                }
+            }
+
             if (chatMemory != null) {
                 messages = chatMemory.messages();
             }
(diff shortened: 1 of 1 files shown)


dliubarskyi (commented):
@davydotcom thank you for your PR!

Please check https://github.com/langchain4j/langchain4j/pull/4732, it introduced a concept of dynamic tool providers (see `ToolProvider.isDynamic()`), perhaps this will solve the problem. Refreshing all tool providers as you do in this PR is a breaking behavior change, it might have a negative side effect.

dliubarskyi:
@davydotcom I will close this PR, please see my comment above. If you have any comments/objections, please either reopen this PR or ping me by @dliubarskyi

PR #4538, opened 2026-02-04
feat: Add token-based triggering mode to SummarizingChatMemory

Ksushik:
## Summary

This PR enhances `SummarizingChatMemory` with an alternative **token-based triggering mode**, complementing the existing message-count mode from PR #4003. Users can now choose to trigger summarization based on token count rather than message count.

## Motivation

While message-count mode is simple and predictable, token-based mode offers more precise control over memory consumption, which is important when working with LLMs that have token limits. This enhancement addresses use cases similar to LangChain Python's `ConversationSummaryBufferMemory`.

Related issues: #1184

### New Builder API for Token-Based Mode

```java
// Token-based mode (NEW)
SummarizingChatMemory memory = SummarizingChatMemory.builder()
    .maxTokens(4000, tokenCountEstimator)
    .maxTokensToSummarize(2000)  // optional
    .defaultGenerateSummaryFunction(chatModel)
    .build();

// Dynamic token configuration
SummarizingChatMemory memory = SummarizingChatMemory.builder()
    .dynamicMaxTokens(id -> userService.getTokenLimit(id))
    .tokenCountEstimator(tokenCountEstimator)
    .defaultGenerateSummaryFunction(chatModel)
    .build();
```

### New Builder Methods

| Method | Description |
|--------|-------------|
| `maxTokens(Integer)` | Static token limit |
| `maxTokens(Integer, TokenCountEstimator)` | Convenience method with estimator |
| `dynamicMaxTokens(Function<Object, Integer>)` | Dynamic token limit per memory ID |
| `tokenCountEstimator(TokenCountEstimator)` | Required for token-based mode |
| `maxTokensToSummarize(Integer)` | Tokens worth of messages to summarize |
| `dynamicMaxTokensToSummarize(Function<Object, Integer>)` | Dynamic version |

### Mode Mutual Exclusivity

The two modes are **mutually exclusive** - users must choose one:
- **Message-count mode**: triggers when `messages.size() > maxMessages`
- **Token-based mode**: triggers when `estimatedTokens > maxTokens`

Build-time validation ensures clear error messages if misconfigured.

### Implementation Details

- Added `TriggerMode` enum (`MESSAGE_COUNT`, `TOKEN_BASED`)
- Added `checkAndSummarizeMessagesByTokens()` method parallel to existing count-based logic
- Token estimation uses `TokenCountEstimator.estimateTokenCountInMessages()`
- Orphaned `ToolExecutionResultMessage`s are properly handled in both modes
[...]

PR #4756, opened 2026-03-20
Introduce a CompactingChatMemory

mariofusco:
## Change

The idea of this experiment is providing an alternative `ChatMemory` implementation that uses a configurable `ChatModel` (ideally much smaller and possibly running locally) that at each iteration summarizes all the old `ChatMessage`into a single `UserMessage`. The eventual `SystemMessage` is preserved and skipped by the summarization. 

The compaction process is triggered, on an independent thread, by the addition of a `UserMessage` to the `ChatMemory`. When this thread terminates its work, it atomically replaces the old chat memory with the compacted one. In this way the compaction will be ready for the next iteration and also the sequence of kinds of messages in the memory is always consistent and won't break the conversation.

I understand that this approach may seem naive and I guess that something similar has been already attempted in the past and then discarded. However having a mechanism that allows to reduce the number of tokens consumed during a conversation could be an important differentiator and this solution could be quite effective despite its simplicity. In fact I also added an integration test trying to compare it with a plain `MessageWindowChatMemory` obtaining the following result.

```
=== Comparison ===
MessageWindowChatMemory:
  Final messages in memory: 12
  Total tokens across all turns: 30876
CompactingChatMemory:
  Final messages in memory: 2
  Total tokens across all turns: 19171
Token savings: 37.9% fewer input tokens with CompactingChatMemory
==================
```

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/memory/chat/CompactingChatMemory.java
+++ b/langchain4j/src/main/java/dev/langchain4j/memory/chat/CompactingChatMemory.java
@@ -1,0 +1,206 @@
+package dev.langchain4j.memory.chat;
+
+import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.data.message.SystemMessage;
+import dev.langchain4j.data.message.UserMessage;
+import dev.langchain4j.internal.DefaultExecutorProvider;
+import dev.langchain4j.memory.ChatMemory;
+import dev.langchain4j.model.chat.ChatModel;
+import dev.langchain4j.service.memory.ChatMemoryService;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+/**
+ * A {@link ChatMemory} implementation that automatically compacts messages by summarizing them
+ * using a {@link ChatModel}. Every time a {@link UserMessage} is added, a background compaction
+ * task is triggered. The compaction uses the provided {@link ChatModel} to produce a summarized
+ * {@link UserMessage} that replaces all current messages.
+ * <p>
+ * During compaction, {@link #messages()} returns the current (pre-compaction) state of messages.
+ * Once compaction completes, the memory is atomically replaced with a single summarized
+ * {@link UserMessage}.
+ * <p>
+ * A {@link SystemMessage}, if present, is preserved across compactions and is not included
+ * in the summarization. It is always kept as the first message.
+ * <p>
+ * The compaction runs on a separate thread, taken from either a user-provided {@link ExecutorService}
+ * or the default one from {@link DefaultExecutorProvider}.
+ */
+public class CompactingChatMemory implements ChatMemory {
+
+    private static final String COMPACTION_PROMPT =
+            """
+            Summarize all former chat messages into a single concise message that preserves
+            all important information, context, decisions, and any pending questions or tasks.
+            The summary should be written from the user's perspective, as if the user is
+            recapping the conversation so far. Do not include any preamble, just provide the summary.
+            """;
+
+    private final Object id;
+    private final ChatModel chatModel;
+    private final ExecutorService executorService;
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+    private final List<ChatMessage> messages = new ArrayList<>();
+
+    private CompactingChatMemory(Builder builder) {
+        this.id = ensureNotNull(builder.id, "id");
+        this.chatModel = ensureNotNull(builder.chatModel, "chatModel");
+        this.executorService = getOrDefault(
+                builder.executorService,
+                DefaultExecutorProvider.getDefaultExecutorService());
+    }
+
+    @Override
+    public Object id() {
+        return id;
+    }
+
+    @Override
+    public void add(ChatMessage message) {
+        lock.writeLock().lock();
+        try {
+            if (message instanceof SystemMessage) {
+                // Replace existing system message if present
+                messages.removeIf(SystemMessage.class::isInstance);
+                messages.add(0, message);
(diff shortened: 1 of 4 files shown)


dliubarskyi:
@mariofusco you are right, there were attempts earlier (https://github.com/langchain4j/langchain4j/pull/4003 and https://github.com/langchain4j/langchain4j/pull/4538, and probably more)

langchain4j/src/main/java/dev/langchain4j/memory/chat/CompactingChatMemory.java
```diff
@@ -0,0 +1,206 @@
...
+ * <p>
+ * A {@link SystemMessage}, if present, is preserved across compactions and is not included
+ * in the summarization. It is always kept as the first message.
+ * <p>
+ * The compaction runs on a separate thread, taken from either a user-provided {@link ExecutorService}
+ * or the default one from {@link DefaultExecutorProvider}.
+ */
+public class CompactingChatMemory implements ChatMemory {
+
+    private static final String COMPACTION_PROMPT =
```
dliubarskyi, line 56: please make it configurable

langchain4j/src/main/java/dev/langchain4j/memory/chat/CompactingChatMemory.java
```diff
@@ -0,0 +1,206 @@
...
+            all important information, context, decisions, and any pending questions or tasks.
+            The summary should be written from the user's perspective, as if the user is
+            recapping the conversation so far. Do not include any preamble, just provide the summary.
+            """;
+
+    private final Object id;
+    private final ChatModel chatModel;
+    private final ExecutorService executorService;
+    private final ReentrantReadWriteLock lock = new ReentrantReadWriteLock();
+    private final List<ChatMessage> messages = new ArrayList<>();
```
dliubarskyi: Please consider using `ChatMemoryStore` so that it can be persisted

dliubarskyi (commented):
@mariofusco I am not sure compacting on every turn is a right approach, seems pretty wasteful, also considering that with prompt caching old messages in the memory are almost for free (as long as you don't change/compact them).

Did you test in it combination with tool calling? Multiple tool calls in a row, etc

mariofusco:
> @mariofusco you are right, there were attempts earlier (https://github.com/langchain4j/langchain4j/pull/4003 and https://github.com/langchain4j/langchain4j/pull/4538, and probably more.

I will give a look and check if there is any good idea there that I can reuse. Out of curiosity is there any general reason that prevented those solutions to be merged (and that could affect also mine)? Asking mostly because I would like to figure out if you think that it worth to invest a bit more time on this or you believe that this is a totally useless effort.

> @mariofusco I am not sure compacting on every turn is a right approach, seems pretty wasteful, also considering that with prompt caching old messages in the memory are almost for free (as long as you don't change/compact them).

I will make all those options configurable. I wanted to demonstrate the basic idea here, but in general I will make totally pluggable both the condition triggering the compaction and the strategy used to perform it.

> Did you test in it combination with tool calling? Multiple tool calls in a row, etc

Good point, I will add more integration tests with use cases involving tool calls.

dliubarskyi:
> Out of curiosity is there any general reason that prevented those solutions to be merged (and that could affect also mine)? Asking mostly because I would like to figure out if you think that it worth to invest a bit more time on this or you believe that this is a totally useless effort.

Well, this feature is hard to implement right, I am sure it requies a lot of testing and evaluation across various real-world use cases to make sure it makes sense and actually improves things. IIUC, memory service will overtake this feature, this is why I did not prioritize reviewing those PRs

glaforge:
A few thoughts:
- it might be worth having a look at what other frameworks are doing in this space, as we're not the first one to investigate this
- I would love to have various knobs to tweak the configuration, for example a setting to say I want to keep the last n messages intact, but summarizing the rest
- Another setting to find the threshold between what to compact and what to keep could also be based out of our token estimators, so that we can say, keep n tokens intact, and the rest can be summarized
- Dmytro mentioned it already, but we might want to have tweaks to say if we keep tool calls as is or not, and/or how to summarize those tool calls
- in ADK, for example, they have a system of sliding window of events (more or less equivalent to our turns), and they can configure a compaction interval, and overlap for the windows, and define a custom summarization class (but at least making the prompt customizable would be I guess equivalent)

langchain4j/src/main/java/dev/langchain4j/memory/chat/CompactingChatMemory.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/memory/chat/CompactingChatMemory.java
mariofusco: Done.

mariofusco:
@dliubarskyi @glaforge I think I've implemented all your suggestions, please give it a second look. In the `chat-memory.md` there's a new section documenting how this works. 

My only doubt is about the handling of tool messages. Now it is possible to preserve `AiMessage`s containing
`ToolExecutionRequest`s and their corresponding `ToolExecutionResultMessage`s as they are, but this adds a bit more complexity and I'm not sure of if and how this could be really useful.

xstefank (commented):
@mariofusco I only read the docs because it doesn't make sense for me to read all the code :)

CompactingChatMemory will also do the eviction right when the context is full?

If you are already plugging in a model to do the compating, why not also do something like selection? What I'm thinking something like RAG but for memory. Select only a few messages from the chat history that are relevant to the current user message.

mariofusco:
> If you are already plugging in a model to do the compating, why not also do something like selection? What I'm thinking something like RAG but for memory. Select only a few messages from the chat history that are relevant to the current user message.

I wasn't thinking to do this, but it could be an interesting idea. My only concern is that, as far as I know, normally RAG is not supposed to work on live content but to index offline in the vector store the documents on which performing the retrieval. Of course, technically what you're suggesting is doable, my doubts are about the time and resources consumed for the indexing of the chat history, but since it is using another LLM for the compaction now, I guess it cannot be much worse :)

@dliubarskyi @glaforge Any feedback on this idea?

PR #4764, opened 2026-03-23
fix: preserve Content subtypes in AI Service template variables (#4352)

### LINKED ISSUE #4352: [BUG] AI Service: wrong treatment of multimodal content
**Describe the bug**
```java
    interface Assistant {
        TokenStream chat(AudioContent audio);
    }
```

In this case `AudioContent` is not treated as a content, it is converted into `String` and sent to the model as `TextContent`.

**Log and Stack trace**
```
- method: POST
- url: https://generativelanguage.googleapis.com/v1beta/models/gemini-2.5-flash-lite:streamGenerateContent?alt=sse
- headers: [x-goog-api-key: AIzaS...r0], [User-Agent: LangChain4j], [Content-Type: application/json]
- body: {
  "contents" : [ {
    "parts" : [ {
      "text" : "AudioContent { audio = Audio { url = null, base64Data = \"AAAAGGZ0eXBtcDQyAAAAAG1wNDFpc29tAAAAKHV1aWRcpwj7Mo5CBahhZQ7KCpWWAAAADDEwLjAuMjYy....
```

**Expected behavior**

If argument of AI Service method is a subtype of `dev.langchain4j.data.message.Content`, it should not be converted to `String`/`TextContent`, but sent to the LLM as-is.

- LangChain4j version: 1.10.0
- LLM(s) used: Google AI Gemini
- Java version: 21

qlsl1198:
## Description

Fixes #4352

### Problem

When passing multimodal data (Content objects like `AudioContent`, `ImageContent`) to AI Service through template variables, the system incorrectly converted them to `String` via `toString()` and sent them to the LLM as `TextContent`. 

Example from the issue:
```java
interface Assistant {
    TokenStream chat(AudioContent audio);
}
```
or with template:
```java
@UserMessage("Transcribe this: {{audio}}")
TokenStream chat(@V("audio") AudioContent audio);
```
Resulted in `"text": "AudioContent { audio = Audio { ... } }"` being sent instead of the actual audio data.

### Solution

1. **`prepareUserMessage`**: When template variables contain `Content` (or `List<Content>`), use new `applyTemplateWithContentSupport()` to build `UserMessage` with mixed contents (text + Content objects) instead of `PromptTemplate.apply()` which converts everything to string.

2. **`addContentsToUserMessage`**: When no Content is found in parameters but `userMessage` already has the correct contents (from the template-with-Content path), return early before `prependTextContentsToUserMessage` which was stripping non-TextContent.

### Changes

- `DefaultAiServices.java`:
  - Add `hasContentInVariables()` to detect Content in template variables
  - Add `applyTemplateWithContentSupport()` to parse template and build `List<Content>` preserving Content objects
  - Reorder `addContentsToUserMessage` to return `userMessage` when it already has correct contents

- `AiServicesUserMessageConfigTest.java`:
  - Add `chat24`, `chat24_simple`, `chat25` - regression tests for template + Content variable scenarios

### Testing

- All existing tests pass
- New tests cover: template with only Content (`{{audio}}`), template with text + Content (`Transcribe: {{audio}}`), and multiple Content variables (`{{image}}` and `{{audio}}`)


langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -571,13 +572,74 @@ private static UserMessage prepareUserMessage(
...
+        }
+        for (Object value : variables.values()) {
+            if (value instanceof Content || isListOfContents(value)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private static List<Content> applyTemplateWithContentSupport(String template, Map<String, Object> variables) {
```
dliubarskyi: It does not seem like a right place for this, `PromptTemplate` and `Prompt` should be responsible for it

dliubarskyi (commented):
@qlsl1198 thank you, this is a nice feature, but requires a redesign

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
dliubarskyi: `Prompt` should probably be changed to hold a list of `Content` instead of a single `String`. Then `toUserMessage()` could create a `UserMessage` with that list of contents

qlsl1198:
@dliubarskyi Following up on your review (“wrong layer” + “Prompt should carry `Content` list / `toUserMessage` uses it”), I reworked the branch accordingly.

**Design**
- Template multimodal handling lives in `PromptTemplate` / `DefaultPromptTemplateFactory` (`renderContents`), not in `DefaultAiServices`.
- `Prompt` stores `List<Content>`; `PromptTemplate.apply` produces that shape when placeholders bind to modalities.

**Housekeeping**
- Synced with upstream `main` and ran Spotless (`chore: apply Spotless after merging upstream main`) so CI formatting stays green.

**Tests**
- Existing AI Service regressions kept; extra `PromptTemplateTest` coverage for `Content`, `List<Content>`, and adjacent scalar merging.

Happy to adjust naming/SPI surface if you prefer not extending `Template` with `renderContents`.

PR #4780, opened 2026-03-25
MCP _meta fields

jmartisk:
Allows passing `_meta` fields for MCP client invocations. See the `docs/docs/tutorials/mcp.md` file for explanation.

The first commit slightly reorganizes the internal protocol messages to more closely mirror the MCP schema (https://github.com/modelcontextprotocol/modelcontextprotocol/blob/2025-06-18/schema/2025-06-18/schema.json). It doesn't introduce any breaking change. The gist is the introduction of `McpClientRequest`,`McpClientResponse`,`McpClientNotification` in the middle of the hierarchy to provide the proper  distinction between these types of messages. `McpClientRequest` and ,`McpClientNotification` now contain a common `params` field instead of each implementation providing its own field - instead, where appropriate, they have their own subclass of `McpClientParams` to hold their specific params, so it's more typesafe and the meta fields can be applied to each request type in a unified way. I've also added javadoc comments to make it clear which schema type each class corresponds to.

This is a prerequisite for proper implementation of distributed tracing of MCP client (https://github.com/quarkiverse/quarkus-langchain4j/issues/2221)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,6 @@ langchain4j-core/target_test-classes/
 
 ## AI Coding Assistants ###
 CLAUDE.md
+
+### Quarkus stuff ###
+.quarkus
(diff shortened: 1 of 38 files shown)


dliubarskyi (approved):
@jmartisk nice refactoring, thank you!

PR #4781, opened 2026-03-25
Add support for @JsonTypeInfo annotation

### LINKED ISSUE #4766: [FEATURE] Support polymorphism in typed output
I have a response to an agent call that I need to coerce into a published schema. 
The agent is doing intent mapping - NLP to formalize 1..n instructions to be provided to a downstream workflow process.

The base class is abstract - there's a property that defines the concrete class.
There are two real issues here.

1. The first is that the serialization to instruct the LLM for the shape of the response can't handle this. 
I'd suggest some form of "one of" and serializing all the classes, but that's not in scope for this request.

2. The second - the focus of this - is that even if I coerce the response to be a correct json payload, the deserializer only recognizes the base class - so most of my content is lost.  

Either the **JsonTypeInfo** annotations to be honored or a (less desirable) equivalent langchain4j annotation that has the same effect, so concrete classes with an abstract base can be produced without manual parsing (and additional prompting).
 
I do it manually with prompting, force the response to a string representation of the json and deserialize after the response is returned.

**Additional context**

The schema maps to something like this:
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.EXISTING_PROPERTY, property = "scope", visible = true)
@JsonSubTypes({ @JsonSubTypes.Type(value = RefineRequestItem.class, name = "request"), 
@JsonSubTypes.Type(value = RefineProductItem.class, name = "product"),
@JsonSubTypes.Type(value = RefineProposalItem.class, name = "proposal") })

The three sub types are all fairly simple classes. 

I'm surprised this hasn't come up before. If it did my jira-fu is clearly lacking and I apologize for the duplication.

mariofusco:
## Issue

Closes #4766

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
@@ -5,6 +5,8 @@ import static java.lang.reflect.Modifier.isStatic;
 import static java.util.Arrays.stream;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import dev.langchain4j.Internal;
 import dev.langchain4j.model.chat.request.json.JsonAnyOfSchema;
 import dev.langchain4j.model.chat.request.json.JsonArraySchema;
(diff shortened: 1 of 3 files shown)


dliubarskyi:
@mariofusco please check https://github.com/langchain4j/langchain4j/pull/4354, we were moving ahead with it (although slowly) for some time

mariofusco:
> @mariofusco please check #4354, we were moving ahead with it (although slowly) for some time

Ah sorry, I totally missed it. Ok, let me know if I can help with something.

Should I close this pull request?

dliubarskyi:
@mariofusco it would be great if you could review https://github.com/langchain4j/langchain4j/pull/4354 and see if it solves https://github.com/langchain4j/langchain4j/issues/4766 completely. If not, then let's see how to combine that PR and yours

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
```diff
@@ -85,7 +87,32 @@ public static JsonSchemaElement jsonSchemaElementFrom(
...
+        }
+
+        List<JsonSchemaElement> subSchemas = new ArrayList<>();
+        for (JsonSubTypes.Type subType : subTypes.value()) {
+            JsonSchemaElement subSchema = jsonObjectOrReferenceSchemaFrom(
+                    subType.value(), null, areSubFieldsRequiredByDefault, visited, false);
+            subSchemas.add(subSchema);
+        }
+
+        return Optional.of(JsonAnyOfSchema.builder()
```
dliubarskyi, line 112: @mariofusco did you test it against a real LLM? I am almost sure this will not work as, IFAIK, no LLM supports "any of" as a root of the schema. Same issue as in https://github.com/langchain4j/langchain4j/pull/4354#discussion_r3051510950

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
```diff
@@ -117,7 +144,7 @@ public static JsonSchemaElement jsonObjectOrReferenceSchemaFrom(
 
         Map<String, JsonSchemaElement> properties = new LinkedHashMap<>();
         List<String> required = new ArrayList<>();
-        for (Field field : type.getDeclaredFields()) {
+        for (Field field : getAllFields(type)) {
```
dliubarskyi, line 147: This seems to be a breaking change, any idea how critical is it?

dliubarskyi (commented):
@mariofusco thank you!

Please add integration tests for all expected use cases, I am not sure exising unit tests cover everything.
Please add a bit of documentaiton.

TBH I am still not sure if this PR is complimentary of the same as https://github.com/langchain4j/langchain4j/pull/4354

AFAIU, the scope of this PR is to support AI Service / Agent to return polymorphic type as a result, right? Or tools are in scope as well?

langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
raphidlee: Do we support inherited fields? 
It looks like getAllFields() is meant to collect fields from parent classes as well — if so, shouldn't this be treated as a separate issue from the polymorphic one? 
What do you both think?

dliubarskyi: We do not support inherited fields at the moment. Not sure it is good or bad (probably bad) but changing this is a breaking change, so perhaps we need to hide it behind a feature flag or something. Or wait till 2.0

PR #4783, opened 2026-03-25
Fix: Serialize Gemini tools as a JSON array

### LINKED ISSUE #4773: [BUG] `langchain4j-google-ai-gemini` serializes non-canonical Gemini `tools` shape (`object` instead of `array`)
## Describe the bug

When using `GoogleAiGeminiChatModel` with tool calling enabled, the generated request payload serializes the top-level `tools` field as a single JSON object:

```json
"tools": {
  "functionDeclarations": [
    {
      "name": "square",
      "description": "Returns the square of a given number.",
      "parameters": {
        "type": "OBJECT",
        "properties": {
          "number": {
            "type": "NUMBER"
          }
        },
        "required": ["number"]
      }
    }
  ]
}
```

However, Gemini's documented `generateContent` request schema defines `tools` as an array/list of `Tool` objects, and the official function-calling examples also use the array form:

```json
"tools": [
  {
    "functionDeclarations": [
      {
        "name": "square",
        "description": "Returns the square of a given number.",
        "parameters": {
          "type": "OBJECT",
          "properties": {
            "number": {
              "type": "NUMBER"
            }
          },
          "required": ["number"]
        }
      }
    ]
  }
]
```

Important nuance: when posting directly to Gemini Developer API (`v1beta/models/...:generateContent`), the backend currently appears to accept **both** forms for this case. So this is not necessarily a hard backend rejection bug in Gemini itself.

The issue here is that LangChain4j is emitting a **non-canonical wire shape** that does not match Gemini's documented contract. That can cause problems with strict validators, proxies, schema-checking middleware, or future server-side validation changes.

## Why this matters

Even if Gemini currently tolerates a singleton object in practice, serializing the documented array form is safer because:

- it matches the public API contract
- it is compatible with stricter intermediaries / schema validators
- it better supports future extension to multiple top-level Gemini tools
- it avoids ambiguity when debugging raw request payloads

### Minimal Java example

```java
import dev.langchain4j.model.chat.request.ChatRequest;
import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
import dev.langchain4j.model.chat.response.ChatResponse;
import dev.langchain4j.model.googleai.GoogleAiGeminiChatModel;

public class GeminiToolsShapeRepro {

        ToolSpecification squareTool = ToolSpecification.builder()
                .name("square")
                .description("Returns the square of a given number.")
                        .addNumberProperty("number", "The number to square")
                        .required("number")
                        .build())
                .build();

                .logRequests(true)
                .logResponses(true)
                .build();

                .messages(UserMessage.from("What is the square of 12? Use the square tool."))
                .toolSpecifications(squareTool)
                .build();

        System.out.println(response);
    }
}
```

### Observed request payload excerpt

```json
{
  "contents": [
    {
      "parts": [
        {
          "text": "What is the square of 12? Use the square tool."
        }
      ],
      "role": "user"
    }
  ],
  "tools": {
    "functionDeclarations": [
      {
        "name": "square",
        "description": "Returns the square of a given number.",
        "parameters": {
          "type": "OBJECT",
          "properties": {
            "number": {
              "type": "NUMBER",
              "description": "The number to square"
            }
          },
          "required": [
            "number"
          ]
        }
      }
    ]
  },
  "toolConfig": {
    "functionCallingConfig": {
      "mode": "AUTO"
    }
  }
}
```

## Expected behavior

LangChain4j should serialize the top-level `tools` field as an array, even if there is only one `GeminiTool` wrapper:

```json
{
  "contents": [
    {
      "parts": [
        {
          "text": "What is the square of 12? Use the square tool."
        }
      ],
      "role": "user"
    }
  ],
  "tools": [
    {
      "functionDeclarations": [
        {
          "name": "square",
          "description": "Returns the square of a given number.",
          "parameters": {
            "type": "OBJECT",
            "properties": {
              "number": {
                "type": "NUMBER",
                "description": "The number to square"
              }
            },
            "required": [
              "number"
            ]
          }
        }
      ]
    }
  ],
  "toolConfig": {
    "functionCallingConfig": {
      "mode": "AUTO"
    }
  }
}
```

## Additional observations

I also verified the behavior with direct manual POSTs to Gemini Developer API:

- `tools` as **object** works in my direct test
- `tools` as **array** also works in my direct test

So the concern is not "Gemini always rejects object form".
The concern is that LangChain4j serializes a non-canonical request shape that differs from the documented schema and may break in stricter environments.

## Suspected cause

From the Gemini integration source, it looks like the internal request model may represent `tools` as a single `GeminiTool` object instead of `List<GeminiTool>`, or the serializer/mapping layer is writing the single tool wrapper directly instead of wrapping it in a one-element list.

## Suggested fix

Serialize the top-level `tools` field as a JSON array in Gemini requests, even when only one top-level tool wrapper exists.

For example, if LangChain4j currently builds a single internal `GeminiTool`, it could still serialize as:

```json
"tools": [ singleTool ]
```

That preserves the current mapping logic while aligning the wire format with Gemini's documented contract.

## Environment

- LangChain4j version: `1.12.2`
- Module: `langchain4j-google-ai-gemini`
- Model: `gemini-2.5-flash`
- Java version: `<fill in>`
- Kotlin version (if applicable): `<fill in>`
- Spring Boot version (if applicable): `<fill in>`

## Attachments

I can attach:

- a minimal Kotlin notebook repro that inspects the generated request shape via reflection
- a minimal end-to-end tool-calling notebook using `GoogleAiGeminiChatModel`
- full request/response logs if helpful

VasilijeJukic01:
## Issue
Fixes #4773

## Change
This PR fixes an issue where the `tools` field in a Gemini `generateContent` request was being serialized as a single JSON object, rather than a JSON array as documented in the official Gemini schema.

While the backend currently tolerates this, it is non-canonical and could cause issues with strict validators or future API changes.

The fix involves:
*   Changing the `tools` field in `GeminiGenerateContentRequest` to be a `List<GeminiTool>`.
*   Updating the internal `FunctionMapper` to wrap the created `GeminiTool` in a `List`.
*   Adjusting all internal call sites and unit tests to accommodate the new list structure.

## General checklist
- [X] I have manually run all the unit tests in the module I have changed (`langchain4j-google-ai-gemini`), and they are all green

### DIFF
--- a/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/BaseGeminiChatModel.java
+++ b/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/BaseGeminiChatModel.java
@@ -4,7 +4,7 @@ import static dev.langchain4j.internal.Utils.copy;
 import static dev.langchain4j.internal.Utils.copyIfNotNull;
 import static dev.langchain4j.internal.Utils.getOrDefault;
 import static dev.langchain4j.model.googleai.FinishReasonMapper.fromGFinishReasonToFinishReason;
-import static dev.langchain4j.model.googleai.FunctionMapper.fromToolSepcsToGTool;
+import static dev.langchain4j.model.googleai.FunctionMapper.fromToolSpecsToGTools;
 import static dev.langchain4j.model.googleai.PartsAndContentsMapper.fromGPartsToAiMessage;
 import static dev.langchain4j.model.googleai.PartsAndContentsMapper.fromMessageToGContent;
 import static dev.langchain4j.model.googleai.SchemaMapper.fromJsonSchemaToGSchema;
@@ -156,7 +156,7 @@ class BaseGeminiChatModel {
                         .mediaResolution(this.mediaResolution)
                         .build())
                 .safetySettings(this.safetySettings)
-                .tools(fromToolSepcsToGTool(
+                .tools(fromToolSpecsToGTools(
                         chatRequest.toolSpecifications(),
                         this.allowCodeExecution,
                         this.allowGoogleSearch,
--- a/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/FunctionMapper.java
+++ b/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/FunctionMapper.java
@@ -3,6 +3,7 @@ package dev.langchain4j.model.googleai;
 import static dev.langchain4j.internal.Utils.isNullOrEmpty;
 import static dev.langchain4j.model.googleai.Json.toJsonWithoutIndent;
 import static dev.langchain4j.model.googleai.SchemaMapper.fromJsonSchemaToGSchema;
+import static java.util.Collections.singletonList;
 
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolSpecification;
@@ -16,7 +17,7 @@ import java.util.List;
 import java.util.Objects;
 
 class FunctionMapper {
-    static GeminiTool fromToolSepcsToGTool(
+    static List<GeminiTool> fromToolSpecsToGTools(
             List<ToolSpecification> specifications,
             boolean allowCodeExecution,
             boolean allowGoogleSearch,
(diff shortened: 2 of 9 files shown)


glaforge:
cc @Bram--

shyam63 (approved):
Thank you for addressing the issue so quickly.

Issue #1542, opened 2024-07-29
[BUG] Text with huge number of empty lines takes enormous amount of time to split


kamenf:
**Describe the bug**

When ingesting a text file containing a lot of empty lines (let say >100000) splitting takes hour or more. The same text cleared of empty lines takes just several seconds.

**To Reproduce**

The sample code:

```
ingestor = EmbeddingStoreIngestor.builder()
	.documentSplitter(DocumentSplitters.recursive(200, 0, new HuggingFaceTokenizer(tokenizerPath)))
	.embeddingModel(embeddingModel)
	.embeddingStore(embeddingStore)
	.build();

ingestor.ingest(document);
```
Example text (important is the number of empty lines) that takes hour or more to split:

```
Some text
in multiple lines

Another text
in multiple lines

More text
in multiple lines

... fill > 100000 such lines....
```

The same example text with cleared empty lines  takes just few seconds to split:

```
Some text
in multiple lines
Another text
in multiple lines
More text
in multiple lines
... fill > 100000 such lines....
```

**Expected behavior**

Splitting of the text with a lot of empty lines to take comparable time to the same text with no empty lines.

**Please complete the following information:**
- LangChain4j version: latest
- LLM(s) used: https://huggingface.co/Xenova/paraphrase-multilingual-MiniLM-L12-v2
- Java version:  22

**Additional context**

PR #4804, opened 2026-03-28
perf: replace String += with StringBuilder in SegmentBuilder

### LINKED ISSUE #4803: [BUG] SegmentBuilder O(n²) string concatenation causes catastrophic slowdown on large documents
**Describe the bug**

`SegmentBuilder` (used by `HierarchicalDocumentSplitter`) stores its accumulated text in a plain `String` field and uses `+=` concatenation in its `append()` method:

```java
segment += joinSeparator;
segment += text;
```

Each `+=` allocates a new String and copies the entire accumulated content, making repeated appends O(n²) in total. For documents with many small segments (e.g., thousands of empty lines), this causes catastrophic slowdown.

**Log and Stack trace**

No exception — the operation completes, but takes orders of magnitude longer than expected. Splitting a document with ~50,000 empty lines takes minutes instead of milliseconds.

**To Reproduce**

```java
SegmentBuilder builder = new SegmentBuilder(Integer.MAX_VALUE, String::length, "\n");
for (int i = 0; i < 50_000; i++) {
    builder.append("x");
}
// With String +=: takes minutes
// With StringBuilder: takes milliseconds
```

**Expected behavior**

`SegmentBuilder` should use `StringBuilder` internally so that `append()` and `prepend()` are amortized O(1) per call, making the total cost O(n).

- LangChain4j version: 1.13.0-beta23-SNAPSHOT (main branch)
- LLM(s) used: N/A (document splitting infrastructure)
- Java version: 17+

**Additional context**

The `prepend()` method has the same issue (`segment = text + joinSeparator + segment`). The fix replaces the `String segment` field with `StringBuilder` and updates all five methods that reference it: `append()`, `prepend()`, `isNotEmpty()`, `toString()`, and `reset()`.

farzad-sedaghatbin:
## Issue
Closes #4803

Related: likely root cause of #1542

## Change

Replaced `String` field + `+=` concatenation in `SegmentBuilder` with `StringBuilder`, eliminating O(n²) performance on repeated `append()` calls.

### What was wrong

`SegmentBuilder` stored accumulated text as a `String` and used `+=` in `append()`:
```java
segment += joinSeparator;
segment += text;
```
Each `+=` copies the entire accumulated string, making n appends cost O(n²) total. For documents with many small segments (thousands of empty lines), splitting took minutes instead of milliseconds.

### What changed

- `String segment` → `StringBuilder segment`
- `append()`: uses `segment.append()` instead of `+=`
- `prepend()`: uses `segment.insert(0, ...)` instead of string concatenation
- `isNotEmpty()`: checks `segment.length() > 0` instead of `!segment.isEmpty()`
- `toString()`: calls `segment.toString().trim()`
- `reset()`: calls `segment.setLength(0)` instead of `segment = ""`

### Files changed
- `SegmentBuilder.java` — replaced String with StringBuilder in all 5 methods
- `SegmentBuilderTest.java` — added `should_handle_many_appends_efficiently` regression test (50,000 appends, completes in milliseconds)

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/data/document/splitter/SegmentBuilder.java
+++ b/langchain4j/src/main/java/dev/langchain4j/data/document/splitter/SegmentBuilder.java
@@ -1,11 +1,10 @@
 package dev.langchain4j.data.document.splitter;
 
-import dev.langchain4j.Internal;
-
-import java.util.function.Function;
-
 import static dev.langchain4j.internal.ValidationUtils.ensureGreaterThanZero;
 import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+import dev.langchain4j.Internal;
+import java.util.function.Function;
 
 /**
  * Segment builder utility class for HierarchicalDocumentSplitter.
@@ -17,7 +16,7 @@ class SegmentBuilder {
     private final Function<String, Integer> sizeFunction;
     private final String joinSeparator;
     private final int joinSeparatorSize;
-    private String segment = "";
+    private StringBuilder segment = new StringBuilder();
     private int segmentSize = 0;
 
     /**
@@ -88,10 +87,10 @@ class SegmentBuilder {
      */
     public void append(String text) {
         if (isNotEmpty()) {
-            segment += joinSeparator;
+            segment.append(joinSeparator);
         }
-        segment += text;
-        segmentSize = sizeOf(segment);
+        segment.append(text);
+        segmentSize = sizeOf(segment.toString());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


langchain4j/src/main/java/dev/langchain4j/data/document/splitter/SegmentBuilder.java
```diff
@@ -17,7 +16,7 @@ class SegmentBuilder {
     private final Function<String, Integer> sizeFunction;
     private final String joinSeparator;
     private final int joinSeparatorSize;
-    private String segment = "";
+    private StringBuilder segment = new StringBuilder();
```
mariofusco, line 19: This could be `final`.

mariofusco (approved):

PR #4807, opened 2026-03-28
fix: correct logical operator in OpenAiOfficialImageModel empty-check

### LINKED ISSUE #4805: [BUG] OpenAiOfficialImageModel.generate() crashes with NoSuchElementException when data is absent
**Describe the bug**

In `OpenAiOfficialImageModel.generate(String)`, the empty-data guard at line 72 uses `&&` instead of `||`:

```java
if (response.data().isEmpty() && response.data().get().isEmpty()) {
```

`response.data()` returns `Optional<List<Image>>`. When the Optional is empty:
- `isEmpty()` returns `true`
- `&&` still evaluates the right side
- `.get()` throws `NoSuchElementException`

The guard was intended to catch both "no data field" and "empty data list", but the `&&` operator defeats it — it crashes on the very condition it was supposed to protect against.

**Log and Stack trace**

```
java.util.NoSuchElementException: No value present
    at java.base/java.util.Optional.get(Optional.java:143)
    at dev.langchain4j.model.openaiofficial.OpenAiOfficialImageModel.generate(OpenAiOfficialImageModel.java:72)
```

**To Reproduce**

Call `generate(prompt)` when the OpenAI API returns a response with no `data` field (i.e., `response.data()` is `Optional.empty()`).

**Expected behavior**

The method should throw `IllegalArgumentException("Image generation failed: no image returned")` when data is absent.

The fix is a one-character change: `&&` → `||`.

- LangChain4j version: 1.13.0-beta23-SNAPSHOT (main branch)
- LLM(s) used: OpenAI (official SDK) image models (dall-e-3, etc.)
- Java version: 17+

**Additional context**

The `generate(String, int)` method at line 85 correctly uses `if (response.data().isEmpty())` — only the single-image `generate(String)` overload has this bug.

farzad-sedaghatbin:
## Issue
Closes #4805

## Change
Single-character fix: changed `&&` to `||` in the empty-data guard on line 72 of `OpenAiOfficialImageModel.java`.

The original condition:
```java
if (response.data().isEmpty() && response.data().get().isEmpty()) {
```
uses `&&`, which means when `data()` returns an empty `Optional`, the left side is `true` but execution continues to the right side, calling `.get()` on the empty `Optional` — throwing `NoSuchElementException`.

The fix:
```java
if (response.data().isEmpty() || response.data().get().isEmpty()) {
```
uses `||` so the guard short-circuits correctly: if data is absent, the exception is thrown without attempting `.get()`.

**Note:** No unit tests are added because the module's response types (`ImagesResponse`, `Image`) are final Kotlin data classes that cannot be easily instantiated or mocked without adding new test dependencies. The fix is a single logical operator change with clear correctness from code inspection.

## General checklist

### DIFF
--- a/langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialImageModel.java
+++ b/langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialImageModel.java
@@ -69,7 +69,7 @@ public class OpenAiOfficialImageModel implements ImageModel {
 
         ImagesResponse response = client.images().generate(imageGenerateParams, requestOptions());
 
-        if (response.data().isEmpty() && response.data().get().isEmpty()) {
+        if (response.data().isEmpty() || response.data().get().isEmpty()) {
             throw new IllegalArgumentException("Image generation failed: no image returned");
         }
         return Response.from(fromOpenAiImage(response.data().get().get(0)));


dliubarskyi (approved):
@farzad-sedaghatbin thank you!

dliubarskyi:
cc @jdubois

PR #4813, opened 2026-03-28
fix: add null/empty choices guard in OpenAiChatModel and aiMessageFrom

### LINKED ISSUE #4810: [BUG] OpenAiChatModel and OpenAiUtils.aiMessageFrom() crash when choices is null or empty
## Describe the bug

`OpenAiChatModel.doChat()` accesses `openAiResponse.choices().get(0)` at line 162 without checking if `choices()` is null or empty. Additionally, the shared utility `OpenAiUtils.aiMessageFrom()` does the same at line 346.

When the OpenAI API returns a response with no choices (e.g. content filtering, malformed response, partial failure), this causes:
- `NullPointerException` if `choices()` is null
- `IndexOutOfBoundsException` if `choices()` is an empty list

The streaming counterpart `OpenAiStreamingChatModel` already has this guard (line 188-189):
```java
List<ChatCompletionChoice> choices = partialResponse.choices();
if (isNullOrEmpty(choices)) { return; }
```

## Log and Stack trace

```
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
    at java.base/java.util.ImmutableCollections$ListN.get(ImmutableCollections.java:XXX)
    at dev.langchain4j.model.openai.OpenAiChatModel.doChat(OpenAiChatModel.java:162)
```

## How To Reproduce

Send a request via `OpenAiChatModel` that results in the API returning an empty choices list.

## Expected behavior

The model should throw a clear `IllegalArgumentException` with a descriptive message instead of a cryptic `IndexOutOfBoundsException` or `NullPointerException`.

## Environment
- Langchain4j version: 1.0.0-beta1+
- Java version: 17+
- Spring Boot version: N/A

farzad-sedaghatbin:
## Issue
Closes #4810

## Change
Added null/empty guard for `choices` in two places:

1. **`OpenAiChatModel.doChat()`** — added `isNullOrEmpty(openAiResponse.choices())` check before accessing `.get(0)`, throwing `IllegalArgumentException` with a descriptive message.

2. **`OpenAiUtils.aiMessageFrom()`** — added the same guard in this shared utility method which also accessed `choices().get(0)` without protection.

Both are consistent with the streaming counterpart `OpenAiStreamingChatModel` which already has `isNullOrEmpty(choices)` guard at line 188-189.

## General checklist


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java
```diff
@@ -343,6 +341,9 @@ public static AiMessage aiMessageFrom(ChatCompletionResponse response) {
     }
 
     public static AiMessage aiMessageFrom(ChatCompletionResponse response, boolean returnThinking) {
+        if (isNullOrEmpty(response.choices())) {
+            throw illegalArgument("Chat completion failed: no choices returned");
```
dliubarskyi: I am not sure this is a right solution for "e.g. content filtering, malformed response, partial failure"

For content filtering we have a `ContentFilteredException`, for other cases we need to check which exactly exception to throw.

farzad-sedaghatbin: Thanks for the review feedback! @dliubarskyi  I investigated the actual runtime scenarios where `choices` would be null/empty in a 200 OK response:

1. **Content filtering at prompt level** → Already caught as HTTP 400 → `InvalidRequestException` (before choices are ever accessed)
2. **Content filtering at output level** → A choice IS returned with `finish_reason: "content_filter"` — choices is NOT empty
3. **Refusal (structured outputs)** → A choice IS returned with `refusal` field → already throws `ContentFilteredException`
4. **Malformed/unexpected response** → The only case where choices is actually empty in a 200, and there's no distinguishing info in the parsed response

So the guard is really only for case #4 — an unexpected server response where we get 200 OK but no choices and no indication why. The current `ChatCompletionResponse` model doesn't carry `prompt_filter_results` or any other field that would let us distinguish causes (they're dropped by `@JsonIgnoreProperties(ignoreUnknown = true)`).

What exception type would you prefer for this scenario? Options I see:

- `LangChain4jException` — generic base, non-retriable
- `InternalServerException` — "server did something wrong" (though technically not an HTTP error)
- Something else?

Or would you prefer a different approach entirely (e.g., removing the guard and letting the raw `IndexOutOfBoundsException` propagate)?

dliubarskyi: Thank you for the analysis, let's go with `InternalServerException` then

farzad-sedaghatbin:
fixed

dliubarskyi (approved):
@farzad-sedaghatbin thank you!

PR #4816, opened 2026-03-29
Reduce token usage when tool calling results are sent to the LLM (disable INDENT_OUTPUT in JacksonJsonCodec)

### LINKED ISSUE #4009: [FEATURE] Remove json indent chars from the sending tool calling result

When I am checking the sending data of my tool calling result, I notice that the serialized json string contains lots of indent chars, which waste tokens.

It would be good if these indent chars could be removed.

Currently I have to use Java reflections to disable the indent chars manually:
```java
    private static void configLangchain4jJson() {
        try {
            /**
             * Langchain4j's Json codec use pretty print by default, which waste tokens
             */

            /**
             * Trigger init first
             */
            Object obj = Json.toJson("");

            Field codecField = Json.class.getDeclaredField("CODEC");
            codecField.setAccessible(true);

            Class<?> JacksonJsonCodec = Class.forName("dev.langchain4j.internal.JacksonJsonCodec", false,
                    Json.class.getClassLoader());
            Field objectMapperField = JacksonJsonCodec.getDeclaredField("objectMapper");
            objectMapperField.setAccessible(true);
            ObjectMapper objectMapper = (ObjectMapper) objectMapperField.get(codecField.get(null));
            objectMapper.configure(SerializationFeature.INDENT_OUTPUT, false);
        } catch (Throwable e) {
            ExceptionHandler.process(e);
        }
    }
```

**Additional context**

DragonFSKY:
## Summary
Closes #4009.

- Remove `.enable(INDENT_OUTPUT)` from `JacksonJsonCodec.createObjectMapper()` so that `Json.toJson()` produces compact JSON
- This reduces token usage when tool calling results are sent to the LLM via `DefaultToolExecutor.toText()` → `Json.toJson()`
- Update existing test assertion from `isEqualToIgnoringWhitespace` to exact `isEqualTo` with compact JSON string

Only the core `JacksonJsonCodec` is changed. Provider modules (OpenAI, Anthropic, Gemini, etc.) have their own `ObjectMapper` instances and are not affected.

`Json` is `@Internal`, so no public API contract is broken.

## Test plan
- [x] `JsonTest.conversionToJsonAndFromJsonWorks` passes with compact output assertion
- [ ] Verify no other tests rely on pretty-printed `Json.toJson()` output format

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JacksonJsonCodec.java
@@ -2,7 +2,6 @@ package dev.langchain4j.internal;
 
 import static com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility.ANY;
 import static com.fasterxml.jackson.annotation.PropertyAccessor.FIELD;
-import static com.fasterxml.jackson.databind.SerializationFeature.INDENT_OUTPUT;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_DATE_TIME;
 import static java.time.format.DateTimeFormatter.ISO_LOCAL_TIME;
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@DragonFSKY thank you!

PR #4818, opened 2026-03-29
fix(anthropic): avoid serializing custom_parameters field

### LINKED ISSUE #4817: Anthropic: customParameters cannot safely send output_config.effort
**Describe the bug**
When using the Anthropic integration, official Anthropic request fields such as `output_config.effort` cannot be sent safely via `customParameters(...)`.

`AnthropicCreateMessageRequest` already has a first-class `outputConfig` field internally, but the public `AnthropicChatModelBuilder` / `AnthropicStreamingChatModelBuilder` do not expose a setter for it. The obvious workaround is to pass:

```java
.customParameters(Map.of("output_config", Map.of("effort", "low")))
```

However, live Anthropic requests fail with:

`custom_parameters: Extra inputs are not permitted`

This makes it impossible to use Anthropic `output_config.effort` through the public LangChain4j Anthropic API.

**Log and Stack trace**
Anthropic API response:

```json
{
  "type": "error",
  "error": {
    "message": "custom_parameters: Extra inputs are not permitted"
  }
}
```

I do not have a meaningful Java stack trace beyond the provider exception because the failure is returned by the Anthropic API itself.

**To Reproduce**
```java
import dev.langchain4j.model.chat.StreamingChatModel;
import dev.langchain4j.model.chat.TestStreamingChatResponseHandler;
import dev.langchain4j.model.anthropic.AnthropicStreamingChatModel;

import java.util.Map;

public class Repro {
        StreamingChatModel model = AnthropicStreamingChatModel.builder()
                .apiKey(System.getenv("ANTHROPIC_API_KEY"))
                .modelName("claude-sonnet-4-6") // also repros on claude-opus-4-5-20251101
                .maxTokens(64)
                .customParameters(Map.of(
                        "output_config", Map.of("effort", "low")
                ))
                .build();

        TestStreamingChatResponseHandler handler = new TestStreamingChatResponseHandler();
        model.chat("Reply with exactly OK.", handler);
        handler.get();
    }
}
```

This sends a request that Anthropic rejects with:

`custom_parameters: Extra inputs are not permitted`

Relevant code locations in current `main`:
- `AnthropicCreateMessageRequest` already has `outputConfig`
- `AnthropicCreateMessageRequest` also has `customParameters` + `@JsonAnyGetter`
- public Anthropic builders expose `customParameters(...)` but not `outputConfig(...)`

**Expected behavior**
One of these should work:

1. `customParameters(...)` should safely flatten additional top-level request fields like `output_config` without emitting an invalid top-level `custom_parameters` field.
2. The Anthropic public builders should expose a first-class `outputConfig(...)` API so official request fields such as `output_config.effort` can be sent without using `customParameters(...)`.

- LangChain4j version: 1.10.0 in our application; also inspected current `main` (`1.13.0-beta23-SNAPSHOT`) and the public Anthropic builder still appears not to expose `outputConfig(...)`
- LLM(s) used: Anthropic Claude Sonnet 4.6, Claude Opus 4.5
- Java version: 17

**Additional context**
This matters because Anthropic now supports official `output_config.effort` on some models, but the current LangChain4j Anthropic public API does not provide a reliable way to send it.

If maintainers prefer, the API ergonomics part (`outputConfig(...)` / `effort(...)` on builders) could be tracked separately as a feature request. The concrete bug here is that the documented/public escape hatch, `customParameters(...)`, does not work correctly for this valid Anthropic request field.

maxlepikhin:
## Issue
Closes #4817

## Change
Fix Anthropic request serialization so `customParameters(...)` is only flattened into top-level request fields and is not also serialized as a top-level `custom_parameters` field.

This fixes cases like:

```java
.customParameters(Map.of("output_config", Map.of("effort", "low")))
```

which should produce:

```json
{
  "output_config": {
    "effort": "low"
  }
}
```

but previously could also emit an invalid top-level `custom_parameters` field.

Changes in this PR:
- add `@JsonIgnore` to `customParameters` in `AnthropicCreateMessageRequest` so Jackson does not serialize it as a normal property
- keep `@JsonAnyGetter` behavior so custom parameters are still flattened into the request body
- add a unit regression test in `DefaultAnthropicClientTest` that verifies:
  - `output_config` is present in the serialized request body
  - `custom_parameters` is not present
- add a live Anthropic regression test in `AnthropicStreamingChatModelIT` that sends `output_config.effort=low` through `customParameters(...)` to a model that supports it and verifies the request succeeds

Manual verification:
- before the fix, Anthropic rejected live requests with:
  - `custom_parameters: Extra inputs are not permitted`
- after the fix, the logged request body contains:
  - `"output_config" : { "effort" : "low" }`
- and Anthropic responds with `200`

## General checklist


maxlepikhin:
Jackson 2.15.3 caused the issue closing.

dliubarskyi (approved):
@maxlepikhin thank you!

Issue #3806, opened 2025-10-02
[FEATURE] Add ability to recover chat memory if tool call was interrupted


karolzdebel:
**Is your feature request related to a problem? Please describe.**
If the application using langchain4j restarts while a tool call is in progress, and a new LLM request is sent from the user for that chat afterward, an error will be returned by the LLM provider since every tool call's response is expected in the chat history. 

**Describe the solution you'd like**
I wrote something like the following for our app in Kotlin and it feels like something that should belong to the framework. The client could set this recovery strategy optionally. 

```
/**
     * Removes interrupted tool calls from the conversation history.
     *
     * Tool calls become interrupted when the app restarts in the middle of tool execution and a user sends a
     * message afterward. Since there's no tool response present in the conversation and there's a new user message
     * the LLM provider responds with a InvalidRequestException. By removing the tool call we can retail most of the
     * conversation so that it doesn't become corrupted.
     *
     * This accounts for parallel tool calls where a single tool execution request attempts to trigger multiple
     * tool calls.
     */
    private fun removeInterruptedToolCalls(messages: List<ChatMessage>): List<ChatMessage> {
        if (messages.isEmpty()) return messages

        val lastMessage = messages.last()
        // Tool calls are only corrupted if a user message comes before their response and that can
        // only happen at the end of the conversation since we make this correction on every update request.
        if (lastMessage !is UserMessage) return messages

        var i = messages.size - 2
        val toolResponses = mutableSetOf<String>()

        while (i >= 0) {
            val message = messages[i]

            when {
                message is ToolExecutionResultMessage -> {
                    toolResponses.add(message.id())
                    i--
                }

                message is AiMessage && message.hasToolExecutionRequests() -> {
                    val toolCallIds = message.toolExecutionRequests().map { it.id() }.toSet()
                    val allResponsesPresent = toolCallIds.all { it in toolResponses }

                    if (!allResponsesPresent) {
                        logger.warn(
                            "Removing corrupted AI message with tool call IDs: ${toolCallIds - toolResponses}. "
                        )
                        return messages.subList(0, i) + listOf(lastMessage)
                    }
                    return messages
                }

                else -> {
                    return messages
                }
            }
        }

        return messages
    }
```

**Describe alternatives you've considered**
n/a

arunsubbaiah:
@dliubarskyi, if no one has started looking into this issue, I would love to take this one. 

Upon initial look, I think we can handle tool call interruption two ways -
[...]

Issue #1966, opened 2024-10-21
[BUG] MessageWindowChatMemory in bad state on concurrent calls with tools


DDani:
**Describe the bug**
On concurrent calls to an **AIService** with **MessageWindowChatMemory** and tools, the messages are not stored in proper order resulting in an invalid status on subsequent calls to OpenAI API. This can happen, for example when exposing having a chat interacting with an AI service, and the user sends multiple messages to an endpoint processed by an AIService.

**Log and Stack trace**

dev.ai4j.openai4j.OpenAiHttpException: {
  "error": {
    "message": "An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: call_voF7DhPAcx07vcVl2h5IX9Sc",
    "type": "invalid_request_error",
    "param": "messages.[3].role",
    "code": null
  }
}
	at dev.ai4j.openai4j.Utils.toException(Utils.java:12) ~[openai4j-0.22.0.jar:na]
	at dev.ai4j.openai4j.StreamingRequestExecutor$2.onFailure(StreamingRequestExecutor.java:218) ~[openai4j-0.22.0.jar:na]
	at okhttp3.internal.sse.RealEventSource.processResponse(RealEventSource.kt:52) ~[okhttp-sse-4.9.3.jar:na]
	at okhttp3.internal.sse.RealEventSource.onResponse(RealEventSource.kt:46) ~[okhttp-sse-4.9.3.jar:na]
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:519) ~[okhttp-4.11.0.jar:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[na:na]
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[na:na]
	at java.base/java.lang.Thread.run(Thread.java:840) ~[na:na]

**To Reproduce**
I have no code (as a test), but I can try to build an example to reproduce it. However, it should be easily reproducible by creating a Test with some threads calling to an AIService for a given user, with memory per user, and tools associated with it.

**Expected behavior**
ChatMessage should be appended in order or sorted at given point to avoid exceptions.

**Please complete the following information:**
- LangChain4j version: 0.35.0
- LLM(s) used: OpenAI
- Java version:  
- Spring Boot version (if applicable): 

[...]

PR #4820, opened 2026-03-30
fix(memory): repair unanswered/orphaned tool calls in chat memory

### LINKED ISSUE #3133: [BUG] Bulk Tool Execution can lead to corrupt memory
**Describe the bug**
We had a scenario where we were not limiting our sequential tool execution and we were able to instruct the assistant to execute a tool a large amount of times for the same request. Since our persisted memory provider has a limit of 35 messages, and since the memory strategy is set external to our implementation in langchain4j, we ended up with a scenario where a lot of `TOOL_EXECUTION_RESULT` messages were added to memory, removing prior context of what led to their request.

All subsequent messages led to an error:

```
dev.langchain4j.exception.InvalidRequestException: {
  \"error\": {
    \"message\": \"Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.\",
    \"type\": \"invalid_request_error\",
    \"param\": \"messages.[0].role\",
    \"code\": null
  }
}
  at dev.langchain4j.internal.ExceptionMapper$DefaultExceptionMapper.mapHttpStatusCode(ExceptionMapper.java:71)
  at dev.langchain4j.internal.ExceptionMapper$DefaultExceptionMapper.mapException(ExceptionMapper.java:44)
  at dev.langchain4j.internal.ExceptionMapper.withExceptionMapper(ExceptionMapper.java:31)
  at dev.langchain4j.internal.RetryUtils.lambda$withRetryMappingExceptions$2(RetryUtils.java:324)
  at dev.langchain4j.internal.RetryUtils$RetryPolicy.withRetry(RetryUtils.java:211)
  at dev.langchain4j.internal.RetryUtils.withRetry(RetryUtils.java:264)
  at dev.langchain4j.internal.RetryUtils.withRetryMappingExceptions(RetryUtils.java:324)
  at dev.langchain4j.internal.RetryUtils.withRetryMappingExceptions(RetryUtils.java:308)
  at dev.langchain4j.model.openai.OpenAiChatModel.doChat(OpenAiChatModel.java:142)
  at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:224)
  at jdk.proxy2/jdk.proxy2.$Proxy7.chat(Unknown Source)
...
Caused by: dev.langchain4j.exception.HttpException: {
  \"error\": {
    \"message\": \"Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.\",
    \"type\": \"invalid_request_error\",
    \"param\": \"messages.[0].role\",
    \"code\": null
  }
}
  at dev.langchain4j.http.client.jdk.JdkHttpClient.execute(JdkHttpClient.java:53)
  at dev.langchain4j.http.client.log.LoggingHttpClient.execute(LoggingHttpClient.java:39)
  at dev.langchain4j.model.openai.internal.SyncRequestExecutor.execute(SyncRequestExecutor.java:20)
  at dev.langchain4j.model.openai.internal.RequestExecutor.execute(RequestExecutor.java:39)
  at dev.langchain4j.model.openai.OpenAiChatModel.lambda$doChat$0(OpenAiChatModel.java:143)
  at dev.langchain4j.internal.ExceptionMapper.withExceptionMapper(ExceptionMapper.java:29)
  ... 19 common frames omitted
```

**Log and Stack trace**
See above

**To Reproduce**
1. Create a tool
2. Limit memory provider storage to a certain max (i.e. 10)
3. Interact with assistant and instruct them to call the tool more than 10 times
4. Error will throw as soon as the memory now overwrites the instruction that led to it

**Expected behavior**
While we are now limiting the tool execution, the biggest concern is that we were put in a place where our memory was now "corrupt". Any new messaging would throw an error for the current state of the history now was invalid.

Once I manually updated the history to remove the tool execution results, I could then continue interacting with the assistant.

I'm interested in suggestions on how to resolve this - one aspect could be to be a bit smarter on the history pruning. For example, always keep the last user message/ai message prior to tool execution chain, and then start collapsing the tool results since that message if you grow beyond the max size of history for the memory. Or even just throw an error, but never leave it in a state where it can't be continued.

- LangChain4j version: 1.0.1
- LLM(s) used: gpt-40
- Java version: 17.0.5

**Additional context**

abhyudayareddy:
## Issue
Closes #3133

Related to #3806 and #1966: this PR closes the gap those two describe (an `AiMessage` with unanswered tool calls left behind by a restart or a race between concurrent requests) as far as chat memory can, but the underlying causes are outside this PR's scope. See "Change" below for what is and isn't covered.

## Change

`MessageWindowChatMemory` and `TokenWindowChatMemory` can end up holding a corrupt history: a `ToolExecutionResultMessage` whose parent `AiMessage` is gone (e.g. a persistence layer that truncated the head of the history, the #3133 shape), or an `AiMessage` whose tool calls were never answered (e.g. an application restart mid tool execution, or two concurrent requests racing on the same memory id). Most LLM providers reject a request built from either shape.

Adds `ToolAwareMessageSanitizer`, a package-private helper both memory classes call every time they load messages from the `ChatMemoryStore`. It repairs both directions in one positional scan: a `ToolExecutionResultMessage` only answers the tool calls made by the `AiMessage` that most recently opened a window (the nearest preceding `AiMessage`, provided no other message type intervenes); a result outside any such window is dropped as orphaned regardless of its id, and an `AiMessage` whose calls are still unanswered once its window closes has those calls stripped (its text is preserved) or is dropped entirely if nothing is left. A null id on either side can't be correlated with its counterpart, so a null-id call is never treated as unanswered, and a null-id result is kept while inside a window (it might answer one of that window's calls) but dropped like any other orphan outside one.

An `AiMessage` with unanswered calls in tail position is always left untouched, since it's indistinguishable from one still awaiting its result. This means a request interrupted mid tool call still fails on the very next request (the dangling `AiMessage` is still last when that request is sent); only the request after that self-heals, once the failed request's own message has closed the window. Fixing that would mean reordering `DefaultAiServices` to add the new user message before reading `messages()` instead of after, which changes what every AI Service call sends and deserves review on its own, so it's deliberately left out of this PR. `docs/docs/tutorials/chat-memory.md` and the Javadoc on both memory classes now say this plainly.

The repaired, shorter list is written back to the store on the next `add()`, so the dropped messages are actually deleted from the store, not just hidden from one read.

## General checklist


abhyudayareddy:
👋 Hi @langchain4j/langchain4j-core-maintainers, @dliubarskyi — could you take a look at this PR when you get a chance? It adds `ToolAwareMessageSanitizer` to self-heal corrupt `ChatMemoryStore` state caused by bulk tool execution eviction (fixes #3133). All 7 CI checks are passing. Thanks!

abhyudayareddy:
Hi @lilian-benoit @deepu105 @geoand and the @langchain4j/langchain4j-core-maintainers team,

Following up on this PR — all 7 CI checks are passing. Would appreciate a first review when you get a chance.

Happy to address any feedback or make adjustments to fit the project's direction. Thanks!

abhyudayareddy:
Hi @lilian-benoit @deepu105 @geoand @langchain4j/langchain4j-core-maintainers — following up again on this PR (all CI checks passing).

This fixes the corrupt `ChatMemoryStore` state caused by bulk tool execution eviction (#3133), which produces hard-to-diagnose `InvalidRequestException: messages with role 'tool' must be a response to a preceding message with 'tool_calls'` errors in production. The fix is a self-healing two-pass O(n) scan in `MessageWindowChatMemory.messages()` — zero allocation on the happy path, idempotent, and logs each orphan removed.

Would appreciate a first review when you get a chance. Thanks!

abhyudayareddy:
cc @lilian-benoit @deepu105 @geoand — checking back in on this one. Just re-verified it's still cleanly mergeable against current main (no conflicts) and issue #3133 is still open upstream, so it's still relevant. No rush at all, just didn't want it to get lost.

abhyudayareddy:
@lilian-benoit @deepu105 @geoand — friendly ping on this one. Still cleanly mergeable against `main`, all 7 CI checks green, and issue #3133 is still open upstream so it remains relevant. Would appreciate a first review when you have a chance.

geoand:
Thanks for this, but it's hard to tell the impact without a test

abhyudayareddy:
@geoand — fair point, and digging in you were right to question it: `ToolAwareMessageSanitizer` was a standalone class that **nothing called**. `MessageWindowChatMemory.messages()` never invoked it, so the self-healing behavior in its javadoc never actually ran, and there was no test to catch that.

Pushed `3030fbbc9`:
- Wired it into `MessageWindowChatMemory.messages()` — it now sanitizes the raw list loaded from the `ChatMemoryStore` before `ensureCapacity` runs, so a history corrupted by an earlier session/library version self-heals instead of throwing on every subsequent call.
- Added `ToolAwareMessageSanitizerTest` — direct unit coverage of `sanitize()` (no-op fast path, single/multiple orphan removal, null id, idempotency).
- Added two cases to `MessageWindowChatMemoryTest` that seed an `InMemoryChatMemoryStore` directly (bypassing `add()`) to reproduce the exact corrupt-history scenario from #3133 end-to-end.

`mvn test -Dtest=ToolAwareMessageSanitizerTest,MessageWindowChatMemoryTest` → 25 run, 0 failures. Thanks for catching this — should be much easier to evaluate the impact now.

abhyudayareddy:
(Small follow-up to my last comment — the PR had also drifted far behind \`main\` since this branch was created back in April, which was showing as a merge conflict. Rebased onto current \`main\`, which had meanwhile refactored \`MessageWindowChatMemory\` to add \`addAsync\`/\`setAsync\`/\`messagesAsync\` behind a shared \`windowed(stored)\` helper — re-applied the sanitizer call there instead, so it now covers the async path too, not just \`messages()\`. Final commit: \`cae174108\`.)

\`mvn test -Dtest=ToolAwareMessageSanitizerTest,MessageWindowChatMemoryTest\` → 28 run, 0 failures. \`mvn spotless:check\` clean on all 4 touched files.

abhyudayareddy:
CI on the last push caught two real regressions — glad it did. Pushed \`25e1519f4\`:

\`sanitize()\` treated a \`ToolExecutionResultMessage\` with a null \`id()\` as automatically orphaned. Several fixtures in this codebase (and some providers) never assign tool-call ids, so it was dropping legitimately-paired result messages:

- \`AiServicesRagChatMemoryBehaviorTest.should_keep_original_message_in_memory_after_tool_loop_when_storage_is_disabled\` — expected 4 messages, sanitizer had silently dropped the (null-id) tool result down to 3
- \`HallucinatedToolNameStrategyTest.should_retry_on_hallucinated_tool_execution_when_streaming\` — same root cause, threw off a downstream message-index assertion

Fix: a null id can't be reliably matched back to its AiMessage in either direction, so it's now never treated as orphaned — the corruption pattern in #3133 always involves real provider-issued ids, so this doesn't weaken the fix for the actual bug being addressed. Also ran \`spotless:apply\` for the formatting failure.

Verified: the two regressed tests + my own 28 (39 total) → 0 failures. Full \`mvn test -pl langchain4j\` (2311 tests, unrelated to this change) → 0 failures. \`spotless:check\` clean.

geoand (commented):
Thanks for the update.

@dliubarskyi needs to vet this to ensure it makes sense

dliubarskyi:
Thanks for the work here, and thanks @geoand for pulling me in.

The instinct is right — self-healing memory is something we do want, it is exactly what @clun asked for on #3133. But after digging in I think the scope is off: this repairs one third of the problem, and not the third that hurts most. I would rather we land a single repair that covers the whole thing. Details below.

## 1. The corruption repaired here is not the one #3133 describes

The description says the sliding window "may evict the parent `AiMessage` that contains the originating `ToolExecutionRequest`(s) before all of its result messages have been added". That was true once, but `ensureCapacity` has cascaded the eviction of the following orphan `ToolExecutionResultMessage`(s) since #666, which predates #3133 by a long way.

I ran the scenario from #3133 against current `main` with `maxMessages=10`:

- 20 sequential `AiMessage(tool_call)` + `ToolExecutionResultMessage` pairs → history ends as `[AI, TOOL_EXECUTION_RESULT] x 5`, no orphan at the head
- 3 rounds of one `AiMessage` with 8 parallel tool calls + 8 results → `[AI, TOOL_EXECUTION_RESULT x 8]`, well formed

So `MessageWindowChatMemory` does not produce this state by itself any more. Re-reading the report, the truncation happened in the reporter's own persistence layer ("our persisted memory provider has a limit of 35 messages"), below `ChatMemory`. Defending against a store that hands us a head-truncated list is a legitimate thing to do, but it is a different claim than the one in the PR description, and `Closes #3133` overstates what this delivers.

## 2. There are three shapes of this bug, and this PR handles the rarest

| Issue | Shape | Trigger | Covered here? |
|---|---|---|---|
| #1966 (`bug`, P2) | `AiMessage` with `tool_calls` and no results | concurrent requests on one memory id | no |
| #3133 (`bug`) | orphan `ToolExecutionResultMessage`(s) at the head | store-side truncation | yes |
| #3806 (`enhancement`) | `AiMessage` with `tool_calls` and no results | app restarts mid tool execution | no |

#1966 and #3806 are the two that LangChain4j can reach on its own, and both leave an `AiMessage` whose `toolExecutionRequests()` are unanswered — the case this PR deliberately skips. The partially answered case has the same problem: `AiMessage(id1, id2, id3)` with only `id2` surviving is left untouched and still rejected by the provider.

Worth reading #3806: the reporter wrote the mirror image of your sanitizer for their own app and said it "feels like something that should belong to the framework". Between the two of you the complete utility already exists, just in two halves.

## 3. What I would like to see instead

Roughly the same amount of code, but closing the whole hole:

1. Repair **both** directions — drop orphan `ToolExecutionResultMessage`(s), and drop (or complete) an `AiMessage` whose tool calls are unanswered, including the partially answered case.
2. Match positionally, not by set membership. Today a result that appears *before* its `AiMessage` counts as covered and is kept, which the provider still rejects — and arbitrary corrupt state is precisely where ordering cannot be assumed.
3. Apply it to `TokenWindowChatMemory` as well. It has the same eviction logic and the same exposure; a shared private helper is enough.
4. No new public type. `ToolAwareMessageSanitizer` is only used by one class, and post-1.0 a public `static List<ChatMessage> sanitize(List)` is a permanent commitment — made harder by the fact that it returns the caller's own list sometimes and a copy other times, so callers cannot know whether the result is safe to mutate. Package-private, or `@Internal`.
5. Update `docs/docs/tutorials/chat-memory.md` next to the existing paragraph on orphan eviction. `messages()` now returns fewer messages than the store holds, for everybody, and that belongs in the docs.
6. Reopen or re-target the issue references so #1966 and #3806 do not stay silently unaddressed.

On configurability: #3806 asks for this to be optional. I would start always-on and non-configurable — a silently repaired history beats a permanently dead conversation — and add a strategy hook later only if somebody actually needs "fail loudly instead".

## 4. Smaller things on the current code

- The null-id handling is asymmetric. A result with a null `id()` is never dropped, but a `ToolExecutionRequest` with a null `id()` contributes nothing to `coveredToolCallIds`. A history where request ids are missing but result ids are present therefore loses every tool result, silently. Whatever the rule ends up being, it needs to be stated for both sides.
- "Zero allocation on the happy path" is not accurate — pass 1 allocates and fills a `HashSet` on every call, and `messages()` runs on every `add()`. Minor in itself, but the claim is in the javadoc, the description and several comments.
- The `WARN` is four lines, carries a `[langchain4j]` prefix that appears nowhere else in the codebase, and embeds an issue URL. One short line, please.
- Class javadoc: `<h2>Problem (Issue #3133)</h2>` and the implementation narrative read like a changelog entry. Javadoc should explain the behaviour to someone meeting this class for the first time.
- Please drop the step-by-step comments in the production code (`// Pass 1 - O(n)`, `// skip this message`, and similar) — the code already says that.
- The PR body does not use our template, so the "There are no breaking changes (API, behaviour)" checkbox never got considered. This change *is* a behaviour change, so it deserves a conscious decision there.

Happy to review a follow-up along these lines.

abhyudayareddy:
Thanks for the detailed review, @dliubarskyi — and for pulling this in, @geoand. This was the right call; I've reworked the PR along the lines you described rather than patching the original scope. Summary of what changed, point by point:

### 1. `Closes #3133` overstated the fix

Agreed, and corrected. The original description conflated two different things. This PR (and its follow-up commits) no longer close #3133 on their own; the PR references #3133, #1966 and #3806 together, since the repair now covers all three shapes. **A maintainer will need to actually re-target/re-link the issues** (add #1966 and #3806 references, adjust the close-on-merge state) — I don't have permissions to edit those issues myself.

### 2 & 3. One repair, both directions, positional matching

`ToolAwareMessageSanitizer.sanitize()` is now a single forward pass (not two independent set-membership passes). For each `AiMessage` with tool calls, it looks only at the run of `ToolExecutionResultMessage`s immediately following it (its "window") and matches results to calls by id within that window — a result anywhere else (before any `AiMessage`, or after the window has closed) is orphaned and dropped, even if some other `AiMessage` earlier in the list happens to declare the same id. This directly implements your point 2.

For the other direction (point 1/2's "rarest case" framing) — an `AiMessage` whose calls are un- or partially answered — my repair strategy is:
- Keep only the tool calls that have a matching, correctly-positioned result; drop the rest from `toolExecutionRequests()` (rebuilt via `AiMessage.toBuilder()`, so text/thinking/attributes survive untouched).
- If *none* of the calls end up answered and the message has no text, drop the `AiMessage` entirely (an empty assistant turn isn't valid history either).
- If none are answered but it has text, keep the message with an empty tool-call list.

**The one judgment call I want to flag explicitly**: an `AiMessage` with unanswered (or partially answered) tool calls is *only* repaired when a subsequent message proves its window has actually closed (i.e. something else was appended after it without completing it). If that `AiMessage` is the last message in the list, it's left completely untouched, no matter how many calls are unanswered. I added this after the first version of the rework broke the existing "tool call just added, result not added yet" test — that in-flight state (including partially-completed parallel tool calls) is indistinguishable at the memory layer from genuine corruption unless something else has since happened. I think this is the only safe rule given `ChatMemory` has no visibility into whether a tool call is still executing, but it does mean #3806-style corruption only self-heals on the *next* turn after the stall, not the instant it happens — flagging this in case that's not the tradeoff you had in mind.

### 3. `TokenWindowChatMemory`

Wired into its `windowed()` the same way as `MessageWindowChatMemory` — both call the same package-private `ToolAwareMessageSanitizer.sanitize()`, no duplicated logic.

### 4. No new public type

`ToolAwareMessageSanitizer` is now package-private (no `@Internal`, since the class itself needed no public surface at all — its only caller is the two `ChatMemory` implementations, both in the same package, and the test class lives there too). Its `sanitize()` method is package-private as well.

### Null-id symmetry

Stated explicitly in the class javadoc now: a `null` id on either side can't be reliably correlated, so it never participates in matching at all. A null-id `ToolExecutionRequest` is treated as always-satisfied (kept untouched, never triggers repair, never blocks its sibling calls from being repaired). A null-id `ToolExecutionResultMessage` is never dropped and never allowed to "answer" anything. This is symmetric with the previous behavior's null-id-result rule, extended to the request side.

### Smaller items

- Log line is now one line, no `[langchain4j]` prefix (confirmed via grep — that prefix appears nowhere else in the codebase), no embedded issue URL.
- Class javadoc rewritten to describe the behavior to a newcomer (no `<h2>Problem (Issue #3133)</h2>`, no changelog narrative).
- Removed the step-by-step inline comments (`// Pass 1 - O(n)`, `// skip this message`, etc.).
- Dropped the "zero allocation" claim entirely — it was never accurate (a `HashSet` is built per `AiMessage` with tool calls) and I didn't want to replace it with a different unverified claim, so allocation behavior just isn't mentioned in the docs anymore.
- Configurability: left this out entirely, per your "always-on, no strategy hook" guidance. No new builder options.
- Updated `docs/docs/tutorials/chat-memory.md` next to the existing orphan-eviction paragraph to describe the self-healing behavior and note that `messages()`/`messagesAsync()` may now return fewer messages than the store holds, for both `MessageWindowChatMemory` and `TokenWindowChatMemory`.
- PR description: will update to stop using `Closes #3133` and go through the "breaking changes" checkbox honestly — this is a behavior change (repaired history can differ from stored history in more cases than before), so I'm marking it as such rather than checking "no breaking changes" by default.

### Verification

- New/updated unit tests in `ToolAwareMessageSanitizerTest` cover: fully valid history (untouched, same instance), orphan result at the head, orphan result before any `AiMessage`, orphan result mid-window, multiple orphans, an `AiMessage` whose window is closed by an intervening message with all calls unanswered (dropped, no text) and with text (stripped to text-only), partial answering (unanswered calls stripped, answered calls + their results kept), a result that arrives after its `AiMessage`'s window already closed (dropped, doesn't retroactively answer it), null-id request not blocking sanitization of sibling calls (both when the sibling is answered and when it's stripped), null-id result never dropped (alone and alongside other repairs), idempotency (both for orphan-dropping and call-stripping), the original #3133 shape (orphans at the head of a truncated list), and — the case I most want a second pair of eyes on — an `AiMessage` with fully or partially unanswered tool calls as the *last* message in the list, asserting it is left completely untouched.
- `MessageWindowChatMemoryTest` and `TokenWindowChatMemoryTest` each got new integration-style tests seeding `InMemoryChatMemoryStore` directly: orphan-result self-heal (existing, kept), a partially-answered `AiMessage` with a message after it proving closure (repaired), an `AiMessage` with no answered calls and no text (dropped), and the pending-at-tail case (left alone).
- `./mvnw -pl langchain4j test -Dtest=ToolAwareMessageSanitizerTest,MessageWindowChatMemoryTest,TokenWindowChatMemoryTest`: all green, no failures.
- `./mvnw -pl langchain4j test` (full module suite, run twice after the rework): all green, no failures in anything I touched or elsewhere in the module.
- `spotless:check`: the module's ratchet-based check couldn't run directly in my sandboxed git worktree (JGit failed to resolve the linked-worktree `.git`, an environment limitation, not a code issue — confirmed by reproducing the identical failure against an unmodified checkout and then getting a clean run once the same files were checked from a plain, non-worktree clone). Re-ran spotless against a plain clone scoped to the files I changed: found and fixed one real line-wrapping violation in `ToolAwareMessageSanitizerTest.java`; the run is otherwise clean for every file this PR touches. (It also flagged one pre-existing violation in `SingleSlotChatMemoryStore.java`, a file I did not touch — left as-is.)

Let me know if the tail-position judgment call above doesn't match what you had in mind — happy to adjust before this goes further.

dliubarskyi:
Thanks for the rework, this is much closer to what I had in mind. Both directions, positional matching, `TokenWindowChatMemory`, package-private: all good, and the module is green on my side. A few things are still open.

## 1. An interrupted tool call still fails the first request after a restart

The tail rule itself is right: a trailing `AiMessage` with unanswered calls cannot be told apart from a tool call that is still running. But the consequence is worse than "self-heals on the next turn".

`DefaultAiServices` reads `chatMemory.messages()` *before* it adds the new user message (`assembleMessages` and `assembleMessagesAsync`). At that moment the dangling `AiMessage` is still the last message, so it is left untouched, the new user message is appended after it, and the provider rejects the request. Only the *following* request heals, because by then the stored user message has closed the window.

Repro: seed the store with `[UserMessage, AiMessage(tool call)]` and use a `ChatModel` that rejects an `AiMessage` whose tool calls are not followed by results, as OpenAI does:

```java
store.updateMessages("default", new ArrayList<>(List.of(UserMessage.from("weather?"), AiMessage.from(toolCall))));

        .chatModel(modelThatRejectsDanglingToolCalls)
        .chatMemory(MessageWindowChatMemory.builder().maxMessages(20).chatMemoryStore(store).build())
        .build();

assistant.chat("hi"); // fails
assistant.chat("hi"); // works, the dangling AiMessage is now gone from the store
```

The memory-level tests cannot show this, because they never go through AI Services.

Fixing it means reordering AI Services: add the user message first, then read `messages()` and swap in the augmented user message with `UserMessage.replaceLast(...)`, as the tool loop already does. That changes the first request of every AI Service call (it would send up to `maxMessages` messages instead of `maxMessages + 1`, and plain chat would start relying on the store returning what was just written), so I would rather not fold it into this PR. Please:
- keep this PR memory-only
- say it plainly in the docs: after an interrupted tool call, the next request still fails, and the conversation recovers after that
- do not close #3806 with this PR

A separate PR for the AI Services part is welcome, it deserves its own review.

## 2. Null ids opt out of positional matching

With positional matching, position alone proves that a result is orphaned, whatever its id. Today a null-id `ToolExecutionResultMessage` that is not inside a tool-call window (for example at the head of a truncated history, which is exactly the #3133 shape) is kept, and the provider still rejects it. So providers that can return tool calls without ids (Gemini and Ollama pass through whatever id they receive, including none) never get this repair.

Please drop a null-id result that is not inside a window, the same as a result with an id. Inside a window, keeping it is fine. I tried this: `AiServicesRagChatMemoryBehaviorTest` and `HallucinatedToolNameStrategyTest`, the tests that led to the null-id exemption, still pass, and only the two tests that assert the current rule need to change.

## 3. Docs and Javadoc

- "drops (or completes) any `AiMessage`": nothing is ever completed.
- "`messages()` may return fewer messages than the store physically holds" undersells it. The next `add()` writes the repaired list back, so those messages are deleted from the store. Users should know that.
- The caveat from point 1.
- The class Javadoc of `MessageWindowChatMemory` and `TokenWindowChatMemory` links `ToolAwareMessageSanitizer`, which users cannot see now that it is package-private. Please describe the behaviour there without naming the internal class.

## 4. Tests

- `should_be_idempotent_when_stripping_partially_answered_calls` never strips anything: the `AiMessage` and its result are the last messages, so the list is returned untouched. It needs another message after the result.
- The comment in `MessageWindowChatMemoryTest.should_self_heal_orphaned_ToolExecutionResultMessage_left_by_a_prior_corrupt_state` still says the orphan comes from "bulk tool execution outrunning ensureCapacity's forward-eviction cascade (issue #3133)", which point 1 of my previous review showed is not the case.
- Nothing asserts on the store after `add()`. That write-back is the part that deletes data, so please assert on `store.getMessages(...)` in at least one `MessageWindowChatMemory` and one `TokenWindowChatMemory` test.

## 5. PR title and description

Neither has been updated yet. The title is cut off with `…`. The body still says `Fixes #3133` / `Closes #3133`, "two-pass" and "zero allocation", and does not use the [PR template](https://github.com/langchain4j/langchain4j/blob/main/.github/pull_request_template.md), so the breaking-changes checkbox is still unanswered.

You don't need a maintainer for the issue links: they come from the keywords in the PR body, which you can edit yourself.
- `Closes #3133`
- #3806: related, not closed (see point 1)
- #1966: related, not closed. The race itself is still there (interleaved requests can still lose tool results, and the model may call the tools again). This PR only makes sure the conversation does not stay broken afterwards.

Happy to take another look once these are in.

abhyudayareddy:
Thanks for the detailed re-review, @dliubarskyi. Pushed `7ea938431` addressing 1-4; also updated the PR title and description (point 5) to match the current implementation.

1. Left `DefaultAiServices` out of this PR as you asked. `MessageWindowChatMemory`/`TokenWindowChatMemory` Javadoc and `chat-memory.md` now say plainly that a request interrupted mid tool call still fails the request right after it, and only the one after that self-heals.

2. `ToolAwareMessageSanitizer` now drops a null-id `ToolExecutionResultMessage` outside any window, same as one with an id. Inside a window it's still kept. `AiServicesRagChatMemoryBehaviorTest` and `HallucinatedToolNameStrategyTest` are unaffected; the two tests that asserted the old behavior are rewritten to assert the corrected one, plus a new test for the still-correct in-window case.

3. Reworded the class Javadoc on both memory classes to describe the repair in prose without linking the package-private sanitizer, and fixed the "drops (or completes)"/"may return fewer messages" doc language to say plainly that the next `add()` permanently deletes the repaired-away messages from the store.

4. `should_be_idempotent_when_stripping_partially_answered_calls` now closes the window with a trailing message so the first pass actually strips something. The stale ensureCapacity/#3133 framing in `MessageWindowChatMemoryTest`'s self-heal comment is corrected to describe store-side truncation instead. Both self-heal tests (`MessageWindowChatMemoryTest`, `TokenWindowChatMemoryTest`) now assert on `store.getMessages(...)` after `add()` to prove the write-back actually deletes the dropped messages.

5. Title and body rewritten to match the current bidirectional/positional design; issue links corrected to `Closes #3133` only, with #3806 and #1966 as related, not closed.

Full `langchain4j` module suite: 1742 run, 0 failures, 0 errors.

PR #4800, opened 2026-03-28
fix: MCP DefaultMcpClient now paginates tools/list via nextCursor

farzad-sedaghatbin:
## Issue
Closes #4798

## Change

`DefaultMcpClient.obtainToolList()` previously made a single `tools/list` request and returned, silently dropping any tools beyond the first page when the MCP server uses pagination.

This PR fixes two problems in `obtainToolList()`:

1. **Missing pagination loop**: The method now follows `nextCursor` — after each response it checks `result.nextCursor`, and if present, issues another `tools/list` request with `cursor` set to that value. Tools from all pages are accumulated into a single list. The loop stops when `nextCursor` is absent/null/blank.

2. **Missing timeout on `Future.get()`**: The method previously called `resultFuture.get()` with no timeout, unlike every other method in the class. It now uses `toolExecutionTimeout` consistently.

The protocol models already fully support pagination:
- `McpListToolsRequest.setCursor(String)`
- `McpListToolsResult.Result.getNextCursor()`

### Files changed
- `DefaultMcpClient.java` — replaced single-request logic with do/while pagination loop + bounded timeout
- `DefaultMcpClientTest.java` — added `should_paginate_tool_list_when_nextCursor_is_present` and `should_return_all_tools_when_no_pagination` tests; refactored helper to support `nextCursor` in mock responses

## General checklist
- [X] There are no breaking changes (API, behaviour)
- [X] I have added unit and/or integration tests for my change
- [X] The tests cover both positive and negative cases
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [X] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green
- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

PR #4822, opened 2026-03-30
MCP client pagination

jmartisk:


jmartisk:
@farzad-sedaghatbin thanks for https://github.com/langchain4j/langchain4j/pull/4800, I would preferably take this one as it also covers resources and prompts (not just tools) and includes a full integration test with quarkus-mcp-server, if you don't mind. However, we should wait for @dliubarskyi to review/approve https://github.com/langchain4j/langchain4j/pull/4780 first because there will be conflicts

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpPaginationStdioTransportIT.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @AfterAll
+    static void teardown() throws Exception {
+        if (mcpClient != null) {
+            mcpClient.close();
+        }
+    }
+
+    @Test
+    void paginatedListingOfToolsResourcesAndPrompts() {
```
dliubarskyi, line 55: Is there any way to verify that pagination actually works? If `quarkus.mcp.server.tools.page-size` (and other pagination properties) is ignored internally for any reason (e.g., bug in the future), this test will stay green, right?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -573,20 +598,40 @@ private synchronized void obtainPromptList() {
...
-        CompletableFuture<JsonNode> resultFuture = null;
-        try {
-            resultFuture = transport.executeOperationWithResponse(operation);
-            result = resultFuture.get(timeoutMillis, TimeUnit.MILLISECONDS);
-            promptRefs.set(PromptsHelper.parsePromptRefs(result));
-        } catch (ExecutionException | InterruptedException | TimeoutException e) {
-            throw new RuntimeException(e);
-        } finally {
-            pendingOperations.remove(operation.getId());
+        List<McpPrompt> allPrompts = new ArrayList<>();
```
dliubarskyi: nit: the same pattern repeats over and over again in this class, perhaps there is a way to extract common logic?

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpPaginationStdioTransportIT.java
jmartisk: Good observation... I've now added `DefaultMcpClientTest#should_paginate_tool_list_using_cursor` that validates this on a lower level

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: I've now extracted a `fetchPaginatedList` method for this, and added a `String cursor` parameter to the request classes constructors to make them easier to construct

dliubarskyi: nice, thanks!

dliubarskyi (approved):
@jmartisk thank you!

PR #4827, opened 2026-03-30
Make the execution state of an agentic system persistable and recoverable

mariofusco:
## Issue

Closes #

## General checklist

### DIFF
--- a/docs/docs/tutorials/agents.md
+++ b/docs/docs/tutorials/agents.md
@@ -1104,128 +1104,6 @@ String response = expertChatbot.ask("I broke my leg what should I do");
 
 The `routerAgent` doesn't need to programmatically specify the output key, since it is already defined in its interface through the `typedOutputKey` attribute of the `@Agent` annotation, while the 3 expert agents still need to specify it programmatically, since their interfaces don't define it, so as usual it is possible to use either one of the 2 approaches. Also, it worth to note that, when reading the values from the `AgenticScope`, like in the conditional workflow definition, there is no need to perform any type check or cast, since the typed keys already provide the necessary type information.
 
-## Memory and context engineering
-
-All agents discussed so far are stateless, meaning that they do not maintain any context or memory of previous interactions. However, like for any other AI service, it is possible to provide agents with a `ChatMemory`, allowing them to maintain context across multiple invocations. 
-
-To provide the former `MedicalExpert` with a memory, it is sufficient to add a field annotated with `@MemoryId` to its signature.
-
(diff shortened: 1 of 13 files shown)


docs/docs/tutorials/agents.md
```diff
@@ -2098,6 +1976,216 @@ waiting for the user to provide the answer, which will be then used to invoke th
...
+        .chatModel(BASE_MODEL)
+        .chatMemoryProvider(memoryId -> MessageWindowChatMemory.withMaxMessages(10))
+        .summarizedContext("medical", "technical")
+        .outputKey("response")
+        .build();
+```
+
+By doing so it internally uses the `ContextSummarizer` agent discussed before, executing it with the same chat model of the agent where it has been defined. It is also possible to add to this method a varargs of the names of the agents whose context should be summarized, so that the summarization is done only for those agents, and not for all the ones used in the agentic system.
+
+### AgenticScope registry and persistence
```
mariofusco, line 2083: @dliubarskyi Everything implemented in this pull requested is explained in this new section of the agents documentation. The apparently new stuff that you see above is only the already existing section that I had to move down here, because the new section mention things like the HumanInTheLoop and the Planner interface that are mentioned only later in this document.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/Planner.java
```diff
@@ -1,11 +1,37 @@
...
+    }
+
+    /**
+     * Restores the planner's execution state from a previously saved map.
+     * Called by the execution loop before {@link #firstAction(PlanningContext)} when recovering
+     * from a persisted scope.
+     *
+     * @param state the previously saved execution state
+     */
+    default void restoreExecutionState(Map<String, Object> state) { }
```
dliubarskyi, line 33: Perhaps it is worth throwing an "unimplemented" exception in the default implementation to avoid confusion?

dliubarskyi (approved):
@mariofusco thank you!

anandfresh (approved):

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/planner/Planner.java
mariofusco: No, I don't think so, because. like I also wrote in the documentation, there are a few `Planner`, like the parallel and the conditional ones, that are stateless, so the empty implementation makes sense for them.

PR #4828, opened 2026-03-30
Support reasoning as well as reasoning_content fields in the Open AI …

### LINKED ISSUE #4796: [FEATURE] Support reasoning as well as reasoning_content fields in the Open AI chat model

Recently VLLM has deprecated the ‘reasoning_content' field used to return thinking in the completions API.  It has been deprecated in favour of a field named ‘reasoning’.

Currently the Open AI chat model only looks for the ‘reasoning_content’ field so therefore does not receive any content and thus the onPartialThinking method never gets called.

This could be addressed if **dev.langchain4j.model.openai.internal.chat.Delta** could check both fields so as to support legacy and future versions of VLLM.

marcvernet:
…chat model

## Issue

Closes #4796

## Change

Modified the class that deserialises the Delta section of the OpenAIChat response to include a ‘reasoning’ field alongside the ‘reasoning_content’ field and return whichever is populated in the chat response.

## General checklist


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/Delta.java
```diff
@@ -26,6 +27,8 @@ public final class Delta {
     @JsonProperty
     private final String reasoningContent;
     @JsonProperty
+    private final String reasoning;
```
dliubarskyi, line 32: What about using @JsonAlias("reasoning") instead?

dliubarskyi (commented):
@marcvernet please add tests (you can use `MockHttpClient`) and update `AssistantMessage` class, so that sync model (`OpenAiChatModel`) works as well

langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/chat/Delta.java
marcvernet: Thats a good call and for this use case with vllm @JsonAlias("reasoning_content", "reasoning") would work. Previously VLLM was returning both fields populated in their responses. However, I'm not sure what would happen if another inference engine returned both but 'reasoning' was a zero length string.

libinye:
Hello, when will this feature be released？

squallliu:
A simpler method
```
@JsonAlias("reasoning")
public Builder reasoningContent(String reasoningContent) {
  this.reasoningContent = reasoningContent;
  return this;
}
```

ZASCHB:
Are there plans to accept this PR? When will it be available, and for which version?

PR #4841, opened 2026-03-31
feat(core): allow overriding tool parameter name with @P

KurobaKaitou:
## Issue

Closes #

## Change

By default, the method registration will automatically obtain the default parameter names of the method through reflection, such as `arg0`, `arg1`, `arg2`, etc., depending on the number of parameters. However, this will result in the key values of the parameters obtained through `ToolExecutionRequest` also being `arg0`, `arg1`, `arg2`, etc. 。 So I added the "name" attribute in @P to solve this problem, which will not cause any changes to the logic that originally did not set this attribute.

## General checklist

### DIFF
--- a/docs/docs/tutorials/tools.md
+++ b/docs/docs/tutorials/tools.md
@@ -13,7 +13,7 @@ All LLMs supporting tools can be found [here](/integrations/language-models) (se
 :::note
 Not all LLMs support tools equally well.
 The ability to understand, select, and correctly use tools depends heavily on the specific model and its capabilities.
-Some models may not support tools at all, while others might require careful prompt engineering 
+Some models may not support tools at all, while others might require careful prompt engineering
 or additional system instructions.
 :::
 
@@ -35,6 +35,7 @@ it should first call one of the provided math tools.
 Let's see how this works in practice (with and without tools):
 
 An example of a message exchange without tools:
+
 ```
 Request:
 - messages:
@@ -45,9 +46,11 @@ Response:
 - AiMessage:
     - text: The square root of 475695037565 is approximately 689710.
 ```
+
 Close, but not correct.
 
 An example of a message exchange with the following tools:
+
 ```java
 @Tool("Sums 2 given numbers")
 double sum(double a, double b) {
@@ -105,6 +108,7 @@ then summarize it and send the summary via email using the `sendEmail` tool.
 :::note
 To increase the chances of the LLM calling the right tool with the right arguments,
 we should provide a clear and unambiguous:
+
 - name of the tool
 - description of what the tool does and when it should be used
 - description of every tool parameter
@@ -119,7 +123,8 @@ Some models can even call multiple tools at once, for example,
 
 :::note
 Please note that not all models support tools.
-To see which models support tools, refer to the "Tools" column on [this](https://docs.langchain4j.dev/integrations/language-models/) page.
+To see which models support tools, refer to the "Tools" column
+on [this](https://docs.langchain4j.dev/integrations/language-models/) page.
 :::
 
 :::note
@@ -129,6 +134,7 @@ Please note that tools/function calling is not the same as [JSON mode](/tutorial
 # 2 levels of abstraction
 
 LangChain4j provides two levels of abstraction for using tools:
+
 - Low-level, using the `ChatModel` and `ToolSpecification` APIs
 - High-level, using [AI Services](/tutorials/ai-services) and `@Tool`-annotated Java methods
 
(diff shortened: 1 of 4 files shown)


docs/docs/tutorials/tools.md
```diff
@@ -306,26 +321,32 @@ and the return value of the method (if any) will be sent back to the LLM.
 You can find implementation details in `DefaultToolExecutor`.
 
 A few tool examples:
+
 ```java
+
 @Tool("Searches Google for relevant URLs, given the query")
-public List<String> searchGoogle(@P("search query") String query) {
+public List<String> searchGoogle(@P(value = "search query", name = "query") String query) {
```
dliubarskyi: It would be great to also have a `description` property on `@P` which is an alias to `value`, WDYT?

dliubarskyi: (out of scope of this PR ofcourse)

docs/docs/tutorials/tools.md
```diff
@@ -335,12 +356,35 @@ Methods annotated with `@Tool` can accept any number of parameters of various ty
 
 Methods without parameters are supported as well.
 
+#### Parameter Name
+
+By default, if the `name` attribute of `@P` is not specified, the method registration will automatically obtain the
+default parameter names of the method through reflection, such as `arg0`, `arg1`, `arg2`, etc., depending on the number
+of parameters. However, this will cause the key values of the parameters obtained through `ToolExecutionRequest` to also
```
dliubarskyi: I would change "However, this will cause the key values of the parameters obtained through `ToolExecutionRequest` to also
be `arg0`, `arg1`, `arg2`, etc. 。" to "This is what LLM will see", or something like that, because it might not be clear what is the exact problem.

Also it is worth mentioning that `argX` is happening when javac `-parameters` is not enabled. So when using LC4j with Quarkus/Spring/etc, this is not a problem, as `-parameters` option is enabled by default by those frameworks, so the actual names of method parameters are preserved and one does not need to specify the `name` in `@P`.

docs/docs/tutorials/tools.md
```diff
@@ -1058,7 +1160,11 @@ In this case by default LangChain4j will throw an exception reporting the proble
...
 with a strategy to be used in this situation.
 
-This strategy is an implementation of a `Function<ToolExecutionRequest, ToolExecutionResultMessage>` defining which `ToolExecutionResultMessage` should be produced as the result for a `ToolExecutionRequest` containing the request to invoke a tool that is not available. For instance, it could be possible to configure the AI service with a strategy that returns to the LLM a response that hopefully will push it to retry a different tool invocation, knowing that the formerly required tool doesn't exist, as in the following example:
+This strategy is an implementation of a `Function<ToolExecutionRequest, ToolExecutionResultMessage>` defining which
+`ToolExecutionResultMessage` should be produced as the result for a `ToolExecutionRequest` containing the request to
+invoke a tool that is not available. For instance, it could be possible to configure the AI service with a strategy that
+returns to the LLM a response that hopefully will push it to retry a different tool invocation, knowing that the
+formerly required tool doesn't exist, as in the following example:
```
dliubarskyi: This class has a lot of unrelated formatting changes, could you pelase revert that? It is hard to review

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
```diff
@@ -149,14 +146,20 @@ private static JsonObjectSchema parametersFrom(Parameter[] parameters) {
...
             boolean isRequired = !isOptional
-                    && Optional.ofNullable(parameter.getAnnotation(P.class))
+                    && Optional.ofNullable(pAnnotation)
                             .map(P::required)
                             .orElse(true);
 
-            properties.put(parameter.getName(), jsonSchemaElementFrom(parameter, visited));
+            String paramName = Optional.ofNullable(pAnnotation)
+                    .map(P::name)
+                    .filter(s -> !s.isEmpty())
```
dliubarskyi:
```suggestion
                    .filter(name -> isNotNullOrBlank(name))
```

dliubarskyi (commented):
@KurobaKaitou thank you!

Please add test(s) to `AiServicesWithToolsIT` that explicitly specifies the name of the tool parameter and then verifies that:
- `ToolSpecification` provided to the LLM has that specific name
- LLM can call the `@Tool`-annotated method properly
- parameter has an expected value

KurobaKaitou:
Hi @dliubarskyi , I’ve made the requested changes based on your feedback. Could you please review it again when you have time?

dliubarskyi:
@KurobaKaitou thank you! Since I was not able to push in your branch, I had to open a new PR: https://github.com/langchain4j/langchain4j/pull/4846

I fixed the implementation, as it was not working properly, polished javadoc and documentaiton. Hope you don't mind. I will close this PR in favour of https://github.com/langchain4j/langchain4j/pull/4846

KurobaKaitou:
@dliubarskyi Thanks for fixing and improving it!

PR #4842, opened 2026-04-01
MCP resource subscriptions

### LINKED ISSUE #4821: [FEATURE] MCP client: Resource subscriptions
Allow subscribing to MCP resource changes as described in https://modelcontextprotocol.io/specification/2025-11-25/server/resources#subscriptions

Perhaps `McpClient` could add a method like: `ResourceSubscription subscribeToResource(String resourceUri, Consumer<ResourceUpdatedEvent> callback)`. The `ResourceSubscription` object would allow to unsubscribe.

jmartisk:
- Closes: https://github.com/langchain4j/langchain4j/issues/4821 

Implements this part of the spec: https://modelcontextprotocol.io/specification/2025-11-25/server/resources#subscriptions


lordofthejars:
Great thanks. I am writing the chapter of my new book about MCP, so this is perfect, thanks.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClient.java
```diff
@@ -94,4 +94,16 @@ public interface McpClient extends AutoCloseable {
      * After calling this method, the client also sends a `notifications/roots/list_changed` message to the server.
      */
     void setRoots(List<McpRoot> roots);
+
+    /**
+     * Subscribes to updates for the resource with the specified URI.
+     * When the resource changes, the server will send a {@code notifications/resources/updated} notification.
+     * The client will invoke the {@code onResourceUpdated} callback (if configured) with the URI of the updated resource.
+     */
+    void subscribeToResource(String uri);
```
dliubarskyi: nit: please move it closer to other the resource-related methods

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpResourceSubscriptionStdioTransportIT.java
```diff
@@ -0,0 +1,101 @@
...
+                .isEqualTo("initial");
+
+        // update the resource on the server
+        mcpClient.executeTool(ToolExecutionRequest.builder()
+                .name("updateStatus")
+                .arguments("{\"newValue\": \"updated\"}")
+                .build());
+
+        // wait a bit for the notification to arrive
+        Thread.sleep(1000);
```
dliubarskyi: nit: could use awaitility instead:

```java
Awaitility.await()
                .atMost(Duration.ofSeconds(5))
                .pollDelay(Duration.ofSeconds(0))
                .pollInterval(Duration.ofMillis(300))
                .untilAsserted(() -> assertThat(updatedResourceUris).contains("file:///status"));
```

langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/integration/McpResourceSubscriptionStdioTransportIT.java
```diff
@@ -0,0 +1,101 @@
...
+
+        // clear the list and update again
+        updatedResourceUris.clear();
+        mcpClient.executeTool(ToolExecutionRequest.builder()
+                .name("updateStatus")
+                .arguments("{\"newValue\": \"updated-again\"}")
+                .build());
+
+        // wait and verify no notification was received after unsubscribing
+        Thread.sleep(1000);
```
dliubarskyi: same here

docs/docs/tutorials/mcp.md
```diff
@@ -371,6 +371,32 @@ and `get_resource` tools should suffice under most circumstances to explain to a
...
+To subscribe to updates for a specific resource, use `client.subscribeToResource(uri)`.
+When the server updates the resource, it sends a `notifications/resources/updated` notification.
+To handle these notifications, register a callback via the `onResourceUpdated` builder method:
+
+```java
+McpClient mcpClient = DefaultMcpClient.builder()
+    .transport(transport)
+    .onResourceUpdated(uri -> {
+        // re-read the updated resource
+        McpReadResourceResult result = mcpClient.readResource(uri);
```
dliubarskyi: Perhaps `mcpClient` should be provided as a second parameter of the callback instead of referencing a not-yet-initialized object?

docs/docs/tutorials/mcp.md
```diff
@@ -371,6 +371,32 @@ and `get_resource` tools should suffice under most circumstances to explain to a
...
+        // process the updated contents...
+    })
+    .build();
+
+// subscribe to a resource
+mcpClient.subscribeToResource("file:///status");
```
dliubarskyi, line 394: The current design has a split responsibility problem — you configure what happens on update in one place (builder), but which resources to watch in another place (subscribe call). This creates a few issues:
It's easy to misuse. Someone sets `onResourceUpdated` and assumes they're done. Or they call `subscribeToResource` but forget the callback. Both cases silently do nothing useful.
The callback is global but subscriptions are per-URI. This mismatch feels inconsistent — you'd expect either both to be global or both to be per-resource.

What about:
```java
mcpClient.subscribeToResource("file:///status", resource -> {
    // process the updated contents...
});
```

dliubarskyi (commented):
@jmartisk thank you!

docs/docs/tutorials/mcp.md
jmartisk: I thought about this  at first but this has different issues, like 
- having to manage references to each callback separately and cleaning up 
  - risk of memory leak, keeping a reference to a callback that corresponds to a subscription that is no longer valid due to a server issue or  because the resource was removed on the server and the spec doesn't allow notifying the client about canceling a subscription.
- I don't think it's super useful to have separate callbacks per resource because the notification is simply an URI, nothing else
- There could be a LOT of resources (especially if they are added/removed dynamically) in which case you would have to keep references to thousands of callbacks and it would be inefficient (memory-wise and also cpu-wise to look up which callback to execute)
- In Quarkus (https://github.com/quarkiverse/quarkus-langchain4j/issues/2290), we have a consensus to go with a CDI approach where all notifications will trigger a single common CDI event type, and that wouldn't work well if each resource had a separate callback

I did add a `warn` log if a subscription is requested but no callback was provided.
Another option would be to provide a default callback that just puts something in the log.

docs/docs/tutorials/mcp.md
jmartisk: Good point... fixed

docs/docs/tutorials/mcp.md
dliubarskyi: @jmartisk issues 1-3 relate to managing many callbacks, but with the approach above you don't **_have_** to have separate callbacks, it can still be the same singleton callback that can be provided in each `subscribeToResource` call. But if needed, one can provide a separate callback for each resource. This is more flexible than a single callback that manages all resources.

Issue 4: not sure I understand, this API will be hidded from the user in Quarkus LC4j, right? So nothing stops you from having a singleton callback that will be attached to every resource

jmartisk: > @jmartisk issues 1-3 relate to managing many callbacks, but with the approach above you don't **_have_** to have separate callbacks, it can still be the same singleton callback that can be provided in each `subscribeToResource` call. But if needed, one can provide a separate callback for each resource. This is more flexible than a single callback that manages all resources.

We would have to make it clear what we recommend having a single callback for performance reasons, but then I don't understand what's the point in allowing more and making it more complicated :)

> 
> Issue 4: not sure I understand, this API will be hidded from the user in Quarkus LC4j, right? So nothing stops you from having a singleton callback that will be attached to every resource

The problem is that it is Quarkus who will provide the callback, not the application. But the application will be the one calling the `subscribeToResource` method, so it's not clear what argument it should pass in that case if the callback is already managed by Quarkus.

dliubarskyi: > I don't understand what's the point in allowing more and making it more complicated :)

It seems more complicated in it's current implementation, when one needs to do 2 actions instead of 1 to subscribe to updates. Also it is spread: callback is registered when building the client, subscription is added on the client afterwards.

> The problem is that it is Quarkus who will provide the callback, not the application. But the application will be the one calling the subscribeToResource method, so it's not clear what argument it should pass in that case if the callback is already managed by Quarkus.

Not sure I understand. So user does not care what to do when resource is updated, but it cares about which resources to monitor? Do you plan to expose naked client to the user? Won't it be configured declaratively in app properties or something?

jmartisk: The plan is that when Quarkus detects a compatible CDI event listener in the app (`@Observes McpResourceUpdatedEvent evt`), it will automatically register a callback (for all managed MCP clients) that will transform update notifications into this `McpResourceUpdatedEvent` and fire the event. So the application doesn't need to care about the underlying callback (this one will be managed by Quarkus), the application will only need to provide the body of the method that handles the CDI event. This feels like the standard CDI-style approach. 
So the application will be responsible for calling `subscribeToResource` for the resources that it wants, but it shouldn't have to deal with the callbacks directly, only provide the CDI event listener that will be common for all resources. If we allow the app to pass a specific callback with a subscription, it's not clear what should happen, should it override the callback that transforms the notification into a CDI event? That would be weird..

dliubarskyi: Thanks for the explanation, it is clear how users will subscribe to events, but it is still not clear how they will configure which resources are they interested in. Do you expect them to obtain an `McpClient` instance and call `subscribeToResource` manually? Or will this be configured declaratively in app properties and Quarkus LC4j will call the method automatically?

jmartisk: I assume that the application will probably first need to obtain a list of existing resources (unless the list is very static and known in advance) and then decide which resources to subscribe to, so this should probably be left to the application to call `subscribeToResource` itself. If people request the option to do it declaratively, we can always do that, but I don't think it would be super useful.

dliubarskyi: OK, I see

lordofthejars: @jmartisk keep in mind that sometimes the resources are "dynamic" for example when using templates where one part of the URI is a paramter.

jmartisk: @lordofthejars  yes, but per my understanding these (resource templates) can't be subscribed to (even though the spec isn't explicit about it) because that subscription would have to already contain specific parameter values and it's not clear how the server should detect when the resource template, given specific parameters, would suddenly render differently, especially if there is some non-determinism in play when rendering the contents.

PR #4848, opened 2026-04-02
Added entry for the ArcadeDB embedding store integration in the comparison table in the documentation

odysseaspenta:
## Change
Added an entry into the embedding store comparison table  in the documentation for the ArcadeDB integration.


dliubarskyi (approved):
@odysseaspenta thank you!

PR #4858, opened 2026-04-03
Fix ClassPathSourceTests URL path handling

### LINKED ISSUE #4857: [BUG] ClassPathSourceTests.findFile() uses File.separatorChar for URL paths and should be re-enabled
**Describe the bug**

`ClassPathSourceTests.findFile()` is currently disabled in `b92a6c1b3` (`disabled failing tests`).

The disabled test still derives the expected filename from `classPathSource.url().getFile()` using `File.separatorChar`, but `URL#getFile()` returns a URL path, and URL paths always use `/` separators regardless of operating system.

This makes the test platform-dependent and inconsistent with the current `ClassPathSource` implementation, which already extracts the filename using `/`.

**Log and Stack trace**

I do not currently have the original Windows stack trace from the failing run.

However, the issue can be demonstrated directly from the current test logic:

```java
var urlString = classPathSource.url().getFile();
var filename = urlString.substring(urlString.lastIndexOf(File.separatorChar) + 1);
```

On Windows:

- `File.separatorChar == '\\'`
- `urlString` is still a URL path and uses `/`

So for a value like:

`/C:/path/to/classPathSourceTests/file1.txt`

`urlString.lastIndexOf('\\')` returns `-1`, and the computed filename becomes the whole string instead of `file1.txt`.

That also makes the expected metadata incorrect.

**To Reproduce**

Current logic in `ClassPathSourceTests.findFile()`:

```java
var classPathSource = ClassPathSource.from(classPathResource);
var urlString = classPathSource.url().getFile();
var filename = urlString.substring(urlString.lastIndexOf(File.separatorChar) + 1);
var expectedMetaData = new Metadata()
        .put(Document.URL, urlString)
        .put(Document.FILE_NAME, filename);
```

This should be evaluated using URL path semantics, not OS filesystem separator semantics.

Relevant file:

- `langchain4j/src/test/java/dev/langchain4j/data/document/source/ClassPathSourceTests.java`

**Expected behavior**

`findFile()` should derive the expected filename and metadata using `/` when parsing `url().getFile()`, and the test should be re-enabled.

- LangChain4j version: main (`363ef83f7`, checked locally on April 3, 2026)
- LLM(s) used: N/A
- Java version: 17

**Additional context**

`ClassPathSource` itself was already updated in `aa4ab8249` (`Fix classpath document loader on windows (#2448)`) to extract the filename from URL paths using `/`.

So the current mismatch is:

- implementation uses URL path semantics
- `findFile()` still uses `File.separatorChar`

I also verified locally on macOS that removing `@Disabled` and replacing `File.separatorChar` with `/` in `findFile()` makes `ClassPathSourceTests` pass.

Kimgyuilli:
## Issue
Closes #4857

## Change

Re-enabled `ClassPathSourceTests.findFile()` and updated the test to use URL path semantics when deriving the expected filename.

The test previously used `File.separatorChar` against `URL#getFile()`, which is incorrect because URL paths always use `/`. The expected metadata now reuses the derived `filename` value, matching the current `ClassPathSource` implementation.

I ran `./mvnw -pl langchain4j -Dtest=ClassPathSourceTests test` to verify the restored test passes.

## General checklist

### DIFF
--- a/langchain4j/src/test/java/dev/langchain4j/data/document/source/ClassPathSourceTests.java
+++ b/langchain4j/src/test/java/dev/langchain4j/data/document/source/ClassPathSourceTests.java
@@ -5,10 +5,8 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import dev.langchain4j.data.document.Document;
 import dev.langchain4j.data.document.Metadata;
-import java.io.File;
 import java.io.IOException;
 
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.CsvSource;
@@ -23,7 +21,6 @@ class ClassPathSourceTests {
                         "'resourceShouldntExist/because/it/just/shouldnt.txt' was not found as a classpath resource");
     }
 
-    @Disabled("TODO fix")
     @ParameterizedTest
     @ValueSource(
             strings = {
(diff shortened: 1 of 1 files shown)


dliubarskyi (approved):
@Kimgyuilli thank you!

Issue #2903, opened 2025-04-22
[FEATURE] MCP client and Function Calling: tool call lifecycle controls: duplicate prevention, configurable limits, and pre-execution hooks


krdk01:
**Is your feature request related to a problem? Please explain.**  
First of all: This problem may seem to be caused by LLM, but I think it would be good to have extra control on the client side.

I'm implementing a MCP client in `McpToolsExampleOverHttp.java` and a Spring Boot MCP server with tools defined in `TestToolsService.java`.

<details>
  <summary>McpToolsExampleOverHttp.java</summary>
package dev.langchain4j.example.mcp;

import dev.langchain4j.mcp.McpToolProvider;
import dev.langchain4j.mcp.client.DefaultMcpClient;
import dev.langchain4j.mcp.client.McpClient;
import dev.langchain4j.mcp.client.transport.McpTransport;
import dev.langchain4j.mcp.client.transport.http.HttpMcpTransport;
import dev.langchain4j.model.chat.ChatLanguageModel;
import dev.langchain4j.model.chat.request.ResponseFormat;
import dev.langchain4j.model.ollama.OllamaChatModel;
import dev.langchain4j.service.AiServices;
import dev.langchain4j.service.tool.ToolProvider;

import java.time.Duration;
import java.util.List;

public class McpToolsExampleOverHttp {
    
    public static void main(String[] args) throws Exception {

        ChatLanguageModel model = OllamaChatModel.builder()
                .baseUrl("https://7f16-34-170-163-147.ngrok-free.app/")
                .modelName("qwen2.5-coder:14b")
                .logRequests(true)
                .logResponses(true)
                .build();

        McpTransport transport = new HttpMcpTransport.Builder()
                .sseUrl("http://localhost:8080/sse")
                .timeout(Duration.ofMinutes(1))
                .logRequests(true)
                .logResponses(true)
                .build();

        McpClient mcpClient = new DefaultMcpClient.Builder()
                .transport(transport)
                .build();

        ToolProvider toolProvider = McpToolProvider.builder()
                .mcpClients(List.of(mcpClient))
                .build();

        Bot assistant = AiServices.builder(Bot.class)
                .chatLanguageModel(model)
                .toolProvider(toolProvider)
                .build();

        String chat = assistant.chat("Get the information of the number 1 star war character");
        /*String chat = assistant.chat("sum of 8+8 Use the provided tool to answer and always assume that the tool is correct. " +
[...]

PR #4859, opened 2026-04-03
feat: add per-tool execution limits to agent tool loop

saibadey:
## Summary

- Adds configurable per-tool invocation limits with two-level enforcement: proactive tool removal before LLM calls and reactive call rejection for parallel over-budget calls
- Three configurable behaviors: CONTINUE (default, rejected call receives a limit-reached message while other tools remain available), ERROR (throws exception), END (halts loop, makes final no-tools LLM call)
- Fully opt-in and backward compatible. No behavior change unless per-tool limits are explicitly configured
- Works with both streaming and non-streaming chat models

See design discussion in #2903.

## Test plan

- [x] 24 tests (17 non-streaming + 7 streaming) covering: per-tool limits, global defaults, per-tool overrides, all three behaviors (CONTINUE/ERROR/END), parallel call rejection, zero limit, all-tools-exhausted, backward compatibility, observability, tool removal verification, negative limit validation, mixed behaviors per tool, sibling tool skipping on END/ERROR
- [x] Full `./mvnw test -pl langchain4j` passes
- [x] `./mvnw spotless:check` passes
- [x] No new dependencies added

Addresses #2903


langchain4j/src/main/java/dev/langchain4j/service/tool/ToolServiceContext.java
```diff
@@ -118,13 +134,12 @@ public int hashCode() {
...
     @Override
     public String toString() {
-        return "ToolServiceContext{" +
-                "effectiveTools=" + effectiveTools +
-                ", availableTools=" + availableTools +
-                ", toolExecutors=" + toolExecutors +
-                ", immediateReturnTools=" + immediateReturnTools +
-                ", dynamicToolProviders=" + dynamicToolProviders +
-                '}';
+        return "ToolServiceContext{" + "effectiveTools="
```
saibadey, line 164: Changes like this are due to spotless. Spotless is configured with <ratchetFrom>origin/main</ratchetFrom> in langchain4j-parent/pom.xml, which reformats all lines in files that differ from origin/main. Since I modified these files for the feature, spotless picked up and reformatted pre-existing code that was untouched by previous contributors (their spotless runs only scoped to their own diff from main at the time). Reverting these changes manually to provide alignment between arg and value will simply delay when these changes will occur (next contribution to touch these files).

saibadey:
Responding to the two comments on the original commit:

On `ToolLimitExceededBehavior.java` (tool-search interaction): Exhausted tools are passed along as a parameter at each point the tool set gets refreshed per-iteration. Covers both the searchable pool and the tools surfaced back by the search strategy. `ToolSearchServiceExhaustionTest` has corresponding tests.

On `AiServices.java` (API shape): Adopted the suggested pattern and cleaned up the corresponding changes that are no longer required. Added default limit and default behavior methods on the builder for setting a single cap across all tools.

dliubarskyi:
Hi @saibadey, thank you for taking the time to put this PR together, I can see real effort went into it.

After reviewing, I don't think this is something we should merge. The added complexity is significant, and I'm not convinced the use case it addresses is common enough to justify carrying that complexity in the core library long-term. Maintenance cost is a real concern for us, and every new feature has a compounding effect on the codebase.

I'm going to close this PR for now. 
Thanks again for contributing, and please don't let this discourage you from opening future PRs.

PR #4867, opened 2026-04-07
Fix copy-paste errors in Guardrail Javadoc

iaJingda:
Fixed 6 incorrect Javadoc comments in guardrail classes caused by copy-paste:

- InputGuardrail#validate(UserMessage): @param userMessage was described as "the response from the LLM" instead of "the user message to be sent to the LLM"
- InputGuardrail#fatal(String, Throwable): described as "Produces a non-fatal failure" instead of "Produces a fatal failure"
- OutputGuardrail#successWith(String, Object) and successWith(AiMessage, Object): described as "Produces a non-fatal failure" instead of producing a successful result
- OutputGuardrailResult#successWith(String, Object) and successWith(AiMessage, Object): same issue


dliubarskyi (approved):
@iaJingda thank you!

PR #4868, opened 2026-04-07
Fix race condition in planner loop

### LINKED ISSUE #4862: [BUG] Planner loop is not thread safe.
I observed strange unpredictable behavior when running P2P setup with multiple, complex , parallel agents. I checked the code and I've found that the `nextAction` field in  PlannerBasedInvocationHandler.PlannerLoop can be modified by other threads when loosing its previous state.

**Describe the bug**
The `nextAction` assignment is not atomic, what can lead to the race condition:
```java
 this.nextAction = composeActions(this.nextAction, planner.nextAction(new PlanningContext(agenticScope, agentInvocation)));
```

The same applies to the main loop:
```
while (nextAction == null || !nextAction.isDone()) {
                if (nextAction == null) {
                    Thread.yield();
                    continue;
                }
               List<AgentExecutor> agents = ((Action.AgentCallAction) nextAction).agentsToCall();
                nextAction = null;
               ...
}
```

When changes are not synchronized, they may not be visible to other threads at all.

**Expected behavior**
No race condition

mariofusco:
<!--
Thank you so much for your contribution!

Please fill in all the sections below.
Please open the PR as a draft initially. Once it is reviewed and approved, we will ask you to add documentation and examples.
Please note that PRs with breaking changes or without tests will be rejected.

Please note that PRs will be reviewed based on the priority of the issues they address.
We ask for your patience. We are doing our best to review your PR as quickly as possible.
Please refrain from pinging and asking when it will be reviewed. Thank you for understanding!
-->

## Issue
<!-- Please specify the ID of the issue this PR is addressing. For example: "Closes #1234" or "Fixes #1234" -->
Closes #4862

## Change

As suggested in the bug report there are two distinct bugs, a more generic one causing a race condition when multiple agents are executed in parallel, and another one, still relatively generic, but specifically triggered by the P2P architecture.                                                                                                                                                
                                                                                                                                                                                                                                                                                        
In particular the first bug is a lost-update race on `nextAction`, where multiple threads calling `onSubagentInvoked` concurrently performed a non-atomic read-modify-write on the nextAction field, causing follow-up actions to be silently lost. The fix is simply making that operation atomic, by wrapping `onSubagentInvoked` body with a `ReentrantLock` (chosen over synchronized to avoid virtual thread pinning). This fix also simplifies the implementation of new Planners, making redundant the user-side lock from `CustomPlannerIT.ParallelInPairsPlanner`, that has been simplified accodingly.                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                          
The second bug is in the fact that `composeActions` drops `done()` in favor of empty call(). When two parallel agents finish and one returns done() while the other returns an empty `call()` (no new agents to schedule), `composeActions` always preferred the non-done action, even when it carried no work. This caused the planner loop to spin forever and it has been fixed by adding an `isEmptyCall()` check so empty `AgentCallActions` are treated as transparent, allowing `done()` to propagate.                                                                                                                                                                    

## General checklist
<!-- Please double-check the following points and mark them like this: [X] -->
<!-- Before adding documentation and example(s) (below), please wait until the PR is reviewed and approved. -->


<!-- Please double-check the following points and mark them like this: [X] -->


<!-- Please double-check the following points and mark them like this: [X] -->

<!-- Please double-check the following points and mark them like this: [X] -->

### DIFF
--- a/langchain4j-agentic-patterns/src/test/java/dev/langchain4j/agentic/patterns/p2p/P2PPlannerComposeActionsIT.java
+++ b/langchain4j-agentic-patterns/src/test/java/dev/langchain4j/agentic/patterns/p2p/P2PPlannerComposeActionsIT.java
@@ -1,0 +1,109 @@
+package dev.langchain4j.agentic.patterns.p2p;
+
+import dev.langchain4j.agentic.Agent;
+import dev.langchain4j.agentic.AgenticServices;
+import dev.langchain4j.agentic.UntypedAgent;
+import dev.langchain4j.agentic.declarative.K;
+import dev.langchain4j.agentic.declarative.TypedKey;
+import dev.langchain4j.agentic.scope.ResultWithAgenticScope;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+public class P2PPlannerComposeActionsIT {
+
+    // ── Typed keys ──────────────────────────────────────────────────────
+
+    public static class Key1 implements TypedKey<String> {
+        @Override public String name() { return "key1"; }
+    }
+
+    public static class Result2 implements TypedKey<String> {
+        @Override public String name() { return "result2"; }
+    }
+
+    public static class Result3 implements TypedKey<String> {
+        @Override public String name() { return "result3"; }
+    }
+
+    // ── Agent wrapper interfaces ────────────────────────────────────────
+    // The @K parameter on Consumer agents tells P2PPlanner to wait until
+    // key1 is present in scope before activating them.
+
+    public interface ProducerAgent {
+        @Agent
+        void produce();
+    }
+
+    public interface Consumer2Agent {
+        @Agent
+        void consume(@K(Key1.class) String key1);
+    }
+
+    public interface Consumer3Agent {
+        @Agent
+        void consume(@K(Key1.class) String key1);
+    }
+
+    @Test
+    void p2p_should_not_freeze_when_parallel_agents_complete() throws Exception {
+        Object agent1 = AgenticServices.conditionalBuilder(ProducerAgent.class)
+                .subAgents(
+                        "Run when key1 absent",
+                        scope -> !scope.hasState("key1"),
+                        AgenticServices.agentAction(scope -> scope.writeState("key1", "hello")))
+                .name("agent1")
+                .build();
+
+        Object agent2 = AgenticServices.conditionalBuilder(Consumer2Agent.class)
+                .subAgents(
+                        "Run when result2 absent",
+                        scope -> !scope.hasState("result2"),
+                        AgenticServices.agentAction(scope -> {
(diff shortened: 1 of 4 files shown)


mariofusco:
@mateusz-dykszak please check if you agree with my analysis and this fix also works for you. I checked it against your reproducer and it is fine for me.

dliubarskyi (dismissed):
@mariofusco thank you!

mateusz-dykszak:
@mariofusco Are you sure you don't need the lock in the loop method? I wasn't sure if firstAction and getting result need locking, so I added them just to be sure, but here I can imagine such a scenario:
```
thread1 - onSubagentInvoked:  nextAction=action1
main - loop: List<AgentExecutor> agents = ((Action.AgentCallAction) nextAction).agentsToCall();
thread2 - onSubagentInvoked:  nextAction=action2
main - loop: nextAction = null;
```
We are not even sure if the main thread in the loop method will see the changes of the `nextAction` if reads are not in the "locked" code.

mateusz-dykszak:
Making the `nextAction` field volatile is not enough, these operations in the loop method:
```
main - loop: List<AgentExecutor> agents = ((Action.AgentCallAction) nextAction).agentsToCall();
main - loop: nextAction = null;
```
are not atomic, so something may happen between them

mariofusco:
> @mariofusco Are you sure you don't need the lock in the loop method? I wasn't sure if firstAction and getting result need locking, so I added them just to be sure, but here I can imagine such a scenario:
> 
> ```
> thread1 - onSubagentInvoked:  nextAction=action1
> main - loop: List<AgentExecutor> agents = ((Action.AgentCallAction) nextAction).agentsToCall();
> thread2 - onSubagentInvoked:  nextAction=action2
> main - loop: nextAction = null;
> ```

This race cannot happen with the current code structure. The key is that `onSubagentInvoked` is always called inside `execute()`, before it returns. Looking at `AgentExecutor.internalExecute()`, the callback fires synchronously before the method returns. This means that `parallelExecution` blocks on `future.get()` for every future. All `onSubagentInvoked` calls complete before get() returns. By the time the main thread proceeds to line 370 (agents = ...) and line 371 (nextAction = null), there are no concurrent writers, they all finished.

> We are not even sure if the main thread in the loop method will see the changes of the `nextAction` if reads are not in the "locked" code.

To play safe I declared `volatile` that `nextAction` variable. For the rest this should also be safe because `CompletableFuture.get() `establishes a happens-before relationship: all actions performed by the completing thread (including the write to `nextAction` under the lock) happen-before the `get()` call returns on the main thread. Since we call get() on every future before the main thread reads `nextAction`, all writes from all background threads are guaranteed to be visible.

mariofusco:
> Making the `nextAction` field volatile is not enough, these operations in the loop method:
> 
> ```
> main - loop: List<AgentExecutor> agents = ((Action.AgentCallAction) nextAction).agentsToCall();
> main - loop: nextAction = null;
> ```
> 
> are not atomic, so something may happen between them

Please see the explanation that I just posted in my other comment.

mateusz-dykszak:
OK, I get it now. I agree, it should be fine with the `parallelExecution` waiting for all sub-actions to finish.

mariofusco:
> OK, I get it now. I agree, it should be fine with the `parallelExecution` waiting for all sub-actions to finish.

Great, thanks a lot for your review.

@dliubarskyi Give one last look and please feel free to merge this if this is ok for you.

dliubarskyi (approved):

PR #4879, opened 2026-04-09
docs: add ChatRequestOptions usage to observability tutorial

DragonFSKY:
Adds a short ChatRequestOptions example to the Chat Model Observability tutorial.

This documents how to pass per-invocation listener metadata via listenerAttributes, and clarifies that these options stay within the LangChain4j invocation chain and are not sent to the LLM provider.


dliubarskyi (approved):
@DragonFSKY thank you!

PR #4885, opened 2026-04-09
Rename editStory method to styleStory

manojsuthar123:
## Issue

Closes #

## General checklist

### DIFF
--- a/docs/docs/tutorials/agents.md
+++ b/docs/docs/tutorials/agents.md
@@ -104,7 +104,7 @@ public interface StyleEditor {
         The story is "{{story}}".
         """)
     @Agent("Edits a story to better fit a given style")
-    String editStory(@V("story") String story, @V("style") String style);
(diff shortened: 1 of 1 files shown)


mariofusco:
Can you please explain why you're changing that method name?

manojsuthar123:
@mariofusco 

The current logging behavior is causing some confusion during debugging and log analysis. Specifically, the same agent name (editStory) is being logged for both the AudienceEditor and StyleEditor.

As a result, it appears in the logs as though the editStory agent is being invoked multiple times, even though these are distinct components. This overlap in naming makes it difficult to accurately trace execution flow and differentiate between the two editors.

This change aims to clarify the logging by ensuring that each editor is represented with a distinct and descriptive agent name, improving traceability and reducing ambiguity in the logs.

mariofusco:
Sorry but I find this change not very relevant and anyway the method name is simply the default agent name, but you can programmatically or declaratively use the agent name that you want.

PR #4902, opened 2026-04-13
fix: make DateOutputParser thread-safe by using DateTimeFormatter

### LINKED ISSUE #4901: [BUG] DateOutputParser uses shared SimpleDateFormat which is not thread-safe
**Describe the bug**
`DateOutputParser` uses a `static final SimpleDateFormat` instance that is shared across all threads:

https://github.com/langchain4j/langchain4j/blob/main/langchain4j/src/main/java/dev/langchain4j/service/output/DateOutputParser.java#L13

`SimpleDateFormat` is [explicitly documented as not thread-safe](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/text/SimpleDateFormat.html) in the JDK. Since the parser is also registered as a singleton in `DefaultOutputParserFactory`, every AI Service invocation returning `java.util.Date` shares the same instance.

Concurrent calls to `parse()` can result in:
- Incorrect date results (silent corruption — hardest to detect)
- `NumberFormatException`
- `ArrayIndexOutOfBoundsException`

Notably, the sibling `LocalDateOutputParser` already avoids this by using the thread-safe `DateTimeFormatter`:

https://github.com/langchain4j/langchain4j/blob/main/langchain4j/src/main/java/dev/langchain4j/service/output/LocalDateOutputParser.java#L13-L15

So this appears to be an oversight rather than a deliberate design choice.

**Log and Stack trace**
Since this is a race condition, exact exceptions vary per run. Typical outputs from the reproducer below include:

java.lang.NumberFormatException: multiple points
java.lang.ArrayIndexOutOfBoundsException: Index 6 out of bounds for length 6
java.lang.NumberFormatException: For input string: ""

(Silent data corruption — returning incorrect Date values — is also possible and even more concerning since no exception is thrown.)

**To Reproduce**
Minimal reproducer demonstrating the race condition:

```java
@Test
void demonstrate_race_condition() throws InterruptedException {
    DateOutputParser parser = new DateOutputParser();
    int threads = 20;
    int iterations = 1000;

    CountDownLatch latch = new CountDownLatch(threads);
    AtomicInteger failures = new AtomicInteger();

    try (ExecutorService executor = Executors.newFixedThreadPool(threads)) {
        for (int i = 0; i < threads; i++) {
            executor.submit(() -> {
                try {
                    for (int j = 0; j < iterations; j++) {
                        parser.parse("2024-01-15");
                    }
                } catch (Exception e) {
                    failures.incrementAndGet();
                } finally {
                    latch.countDown();
                }
            });
        }
        latch.await();
    }
    assertThat(failures.get()).isZero(); // fails on current main
}
```

Real-world trigger: any AI Service method returning `Date` called concurrently from a web backend (Spring Boot, Quarkus, or virtual threads).

**Expected behavior**
`parse()` should be safe to call concurrently, producing correct `Date` values for every call. The same pattern (`yyyy-MM-dd`) and default-timezone semantics should be preserved.

The sibling `LocalDateOutputParser` demonstrates the canonical approach using `DateTimeFormatter`.

- LangChain4j version: main (1.14.0-SNAPSHOT), also affects all prior releases
- LLM(s) used: N/A (bug is in date parsing layer, LLM-independent)
- Java version: 17+
- Spring Boot version: N/A (affects any concurrent usage)

iaJingda:
Fixes #4901.

## Summary

`DateOutputParser` used a `static final SimpleDateFormat` instance shared
across all threads (and shared across the whole application via
`DefaultOutputParserFactory`). `SimpleDateFormat` is
[documented as not thread-safe](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/text/SimpleDateFormat.html),
so concurrent AI Service calls returning `Date` can produce silently
corrupt results or throw `NumberFormatException` /
`ArrayIndexOutOfBoundsException`.

This PR replaces `SimpleDateFormat` with the thread-safe
`DateTimeFormatter.ISO_LOCAL_DATE`, aligning the implementation with
the sibling `LocalDateOutputParser`.

## Changes

- `DateOutputParser`: use `DateTimeFormatter` + `LocalDate.parse(...)` +
  `Date.from(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant())`
- `OutputParserTest`: add `date_parser_is_thread_safe` — a concurrent
  regression test (16 threads × 500 iterations). Fails on the current
  `main` (`NumberFormatException: multiple points`) and passes with
  this fix.

## Behavior preservation

- Same format pattern (`yyyy-MM-dd`) and `formatInstructions()` output
- Same default-timezone semantics (`ZoneId.systemDefault()` matches what
  `SimpleDateFormat` did without explicit TZ configuration)
- Same `RuntimeException` surface and exact message for malformed input —
  the `indexOf("-")` validation is intentionally kept so the pre-existing
  `date()` test assertion
  (`.withMessage("Invalid date format: 01-12-2020")`) still holds

## Open questions raised in the issue

I left three questions in #4901 and chose the most conservative defaults
here; happy to adjust based on maintainer preference:

1. `DateTimeFormatter` vs `ThreadLocal<SimpleDateFormat>` → chose
   `DateTimeFormatter` to match `LocalDateOutputParser`
2. `ZoneId.systemDefault()` vs `ZoneOffset.UTC` → chose
   `systemDefault()` for behavior preservation
3. Remove the `indexOf("-")` pre-check → kept it to preserve the
   existing error-message assertion

Marking as **draft** until the direction is confirmed.

## Test plan

- [x] Existing `OutputParserTest` tests pass (9/9)
- [x] New concurrent test fails on current `main` and passes with this fix
- [x] `./mvnw -pl langchain4j spotless:check` passes

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/output/DateOutputParser.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/output/DateOutputParser.java
@@ -15,8 +15,15 @@ class DateOutputParser implements OutputParser<Date> {
 
     @Override
     public Date parse(String string) {
-        LocalDate localDate = LocalDate.parse(string.trim(), FORMATTER);
-        return Date.from(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant());
+        if (string == null) {
+            throw new OutputParsingException("Cannot parse null into java.util.Date", null);
+        }
+        try {
+            LocalDate localDate = LocalDate.parse(string.trim(), FORMATTER);
+            return Date.from(localDate.atStartOfDay(ZoneId.systemDefault()).toInstant());
+        } catch (Exception e) {
+            throw new OutputParsingException("Cannot parse '%s' into java.util.Date".formatted(string), e);
+        }
     }
 
     @Override
--- a/langchain4j/src/main/java/dev/langchain4j/service/output/DateOutputParser.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/output/DateOutputParser.java
@@ -1,29 +1,33 @@
 package dev.langchain4j.service.output;
 
 import dev.langchain4j.Internal;
-
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
+import java.time.LocalDate;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
 import java.util.Date;
 
 @Internal
 class DateOutputParser implements OutputParser<Date> {
 
     private static final String DATE_PATTERN = "yyyy-MM-dd";
-    private static final SimpleDateFormat SIMPLE_DATE_FORMAT = new SimpleDateFormat(DATE_PATTERN);
+    // DateTimeFormatter is immutable and thread-safe, unlike SimpleDateFormat.
+    private static final DateTimeFormatter FORMATTER = DateTimeFormatter.ISO_LOCAL_DATE;
 
     @Override
     public Date parse(String string) {
         string = string.trim();
 
-        // SimpleDateFormat silently accepts dd-MM-yyyy; but parses it strangely.
+        // Guard against inputs like "dd-MM-yyyy" so the error message stays consistent
+        // with previous behavior (and independent of formatter-specific exceptions).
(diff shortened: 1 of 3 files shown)


dliubarskyi (approved):
@iaJingda thank you!

PR #4905, opened 2026-04-13
Allow multiple McpClientListeners

### LINKED ISSUE #4904: [FEATURE] Support multiple McpClientListeners
This is an oversight on my part.. With tracing and metrics moving into `McpClientListener`s, it's clear that one MCP client will have to support multiple listeners so we don't have to hack around it by creating listener aggregators...

jmartisk:
- Fixes: https://github.com/langchain4j/langchain4j/issues/4904

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -58,6 +58,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import org.slf4j.Logger;
@@ -110,7 +111,7 @@ public class DefaultMcpClient implements McpClient {
     private final Boolean cacheToolList;
     private final Boolean cacheResourceList;
     private final Boolean cachePromptList;
-    private final McpClientListener listener;
+    private final List<McpClientListener> listeners;
     private final McpMetaSupplier metaSupplier;
 
     public DefaultMcpClient(Builder builder) {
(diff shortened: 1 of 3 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -285,22 +285,22 @@ public ToolExecutionResult executeTool(ToolExecutionRequest executionRequest, In
         JsonNode result = null;
         McpCallContext context = new McpCallContext(invocationContext, operation);
         try {
-            if (listener != null) {
+            for (McpClientListener listener : listeners) {
                 listener.beforeExecuteTool(context);
```
dliubarskyi: It would probably make sense to wrap listener invocation into a try/catch block to not block the main flow and other listeners when single listener fails. We use this pattern in all other places.

dliubarskyi (commented):
@jmartisk thank you!

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
jmartisk: Is there a policy what types of exceptions should fail the whole operation and what types should just be logged? Or should we just log every exception and go on?

jmartisk: At minimum, an `InterruptedException` should be propagated further and not ignored.

jmartisk: But we don't declare that it can be thrown from the listeners... Let's just log everything for now I guess, we can come up with a unified policy later?!

dliubarskyi: @jmartisk there is no explicit policy right now, but if you look at most listeners, they are called in the try/catch blocks, so that if user code thrown an exception, it does not interfere with the main flow and other listeners in the chain. I would just log every exception and continue

dliubarskyi (approved):
@jmartisk thanks!

PR #4908, opened 2026-04-13
Add automated API backward compatibility checks with Revapi

### LINKED ISSUE #3320: Automatic check to ensure backward compatibility of public API
Using the [Revapi Maven Plugin](https://revapi.org/revapi-maven-plugin/) implement an automatic check to make sure that a pull request is not breaking the backward compatibility of public API.

### LINKED ISSUE #3804: Automatically check for breaking API changes
Revapi or similar

dliubarskyi:
## Issue
Closes https://github.com/langchain4j/langchain4j/issues/3804
Closes https://github.com/langchain4j/langchain4j/issues/3320

## Change
- Integrate https://revapi.org/ (`revapi-maven-plugin`) into the build to automatically detect breaking API changes
- Revapi runs during the `verify` phase on every build, comparing the current code against the latest released version on Maven Central (`RELEASE`)
- Suppress `java.method.varargOverloadsOnlyDifferInVarargParameter` globally — this is a design warning about varargs overloads, not an actual breaking change
- Add per-module `revapi.json` justifications for pre-existing `java.class.nonPublicPartOfAPI` warnings in `langchain4j-ollama`, `langchain4j-bedrock`, `langchain4j-onnx-scoring`, and
  `langchain4j-google-ai-gemini` — these are package-private classes exposed through protected members on package-private base classes, so they cannot actually leak to external
  users
- Do not check depndencies - we do not expose them explicitly and are not responsible for any changes there
- skip `integration-tests` modules - we do not release them

### How it works
- Revapi downloads the latest released JAR from Maven Central and compares it against the locally built JAR at the bytecode level
- Any removal, signature change, or visibility reduction of public API elements will fail the build
- Adding new public API is always fine
- To justify an intentional breaking change, add a `revapi.json` in the module root — justifications auto-expire on the next release since the baseline moves forward

## General checklist

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -51,6 +51,9 @@ jobs:
           distribution: 'temurin'
           cache: 'maven'
 
+      - name: Print Maven version
+        run: mvn --version
+
       - name: Compile and unit test with JDK ${{ matrix.java_version }}
         run: |
           mvn -B -U -DembeddingsSkipCache -T8C test javadoc:aggregate
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -39,5 +39,6 @@
         <maven.javadoc.skip>true</maven.javadoc.skip>
         <maven.source.skip>true</maven.source.skip>
         <enforcer.skipRules>requireUpperBoundDeps</enforcer.skipRules>
+        <revapi.skip>true</revapi.skip>
     </properties>
 </project>
--- a/langchain4j-bedrock/revapi.json
+++ b/langchain4j-bedrock/revapi.json
@@ -1,0 +1,15 @@
+[
+  {
+    "extension": "revapi.differences",
+    "configuration": {
+      "ignore": true,
+      "differences": [
+        {
+          "code": "java.class.nonPublicPartOfAPI",
+          "old": "class dev.langchain4j.model.bedrock.BedrockTitanEmbeddingResponse",
+          "justification": "AbstractBedrockEmbeddingModel is package-private, so the generic type parameter cannot leak to external subclasses"
+        }
+      ]
+    }
+  }
+]
--- a/langchain4j-google-ai-gemini/revapi.json
+++ b/langchain4j-google-ai-gemini/revapi.json
@@ -1,0 +1,45 @@
+[
+  {
+    "extension": "revapi.differences",
+    "configuration": {
+      "ignore": true,
+      "differences": [
+        {
+          "code": "java.class.nonPublicPartOfAPI",
+          "old": "class dev.langchain4j.model.googleai.BatchRequestResponse.Operation.Status",
+          "justification": "Exposed only through package-private base classes"
+        },
+        {
+          "code": "java.class.nonPublicPartOfAPI",
+          "old": "class dev.langchain4j.model.googleai.GeminiGenerateContentRequest",
+          "justification": "Exposed only through package-private base classes"
+        },
+        {
+          "code": "java.class.nonPublicPartOfAPI",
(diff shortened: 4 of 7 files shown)


mariofusco (changes requested):
I'm trying to build this pull request locally to check how it works, but when I try to do so I get the following error

```
[ERROR] Failed to execute goal org.revapi:revapi-maven-plugin:0.15.1:check (default) on project langchain4j-test: The following API problems caused the build to fail:
[ERROR] java.class.nonPublicPartOfAPI: class org.assertj.core.api.recursive.comparison.ComparatorForPatterns: Class 'org.assertj.core.api.recursive.comparison.ComparatorForPatterns' is indirectly included in the API (by the means of method return type for example) but the class is not accessible (neither public nor protected). [org.assertj:assertj-core:jar:4.0.0-M1]
```

Can you reproduce it? I didn't investigate further yet, but if you want I could check what's going wrong and take over on this task (sorry, I know it's long overdue).

dliubarskyi:
@mariofusco that is strange, I don't have any issues with `langchain4j-test`:

```
[INFO] --- revapi:0.15.1:check (default) @ langchain4j-test ---
Downloading from central-portal-snapshots: https://central.sonatype.com/repository/maven-snapshots/dev/langchain4j/langchain4j-test/maven-metadata.xml
Downloading from central: https://repo.maven.apache.org/maven2/dev/langchain4j/langchain4j-test/maven-metadata.xml
Downloaded from central: https://repo.maven.apache.org/maven2/dev/langchain4j/langchain4j-test/maven-metadata.xml (1.3 kB at 5.0 kB/s)
Downloaded from central-portal-snapshots: https://central.sonatype.com/repository/maven-snapshots/dev/langchain4j/langchain4j-test/maven-metadata.xml (535 B at 502 B/s)
[INFO] Comparing [dev.langchain4j:langchain4j-test:jar:1.13.0-beta23] against [dev.langchain4j:langchain4j-test:jar:1.14.0-beta24-SNAPSHOT] (including their transitive dependencies).
[INFO] API checks completed without failures.
```

How do you run the build? Can you provide a specific command?

mariofusco:
This seems to be a false negative indeed, I'm still investigating.

I ran the build with

`mvn clean install -U -Dspotless.check.skip=true -Dmaven.javadoc.skip=true -DskipTests -DskipITs`

I'm on Java 25 and maven 3.9.14

mariofusco:
I'm pasting below Claude's explanation of the problem. The analysis seems correct, but still it is very odd that revapi works this way:

```
The key word in the error is "indirectly". ComparatorForPatterns isn't used directly in your code. It's exposed transitively — some public class in langchain4j-test extends or returns an AssertJ type, which in turn references     
  ComparatorForPatterns through its own type hierarchy.

That's the culprit. Here's the transitive exposure chain:                                                                                                                                                                                          
                                                                                                                                                                                                                                                     
  1. GuardrailAssertions extends org.assertj.core.api.Assertions                                                                                                                                                                                     
  2. Assertions has public/protected methods that return types involving RecursiveComparisonAssert                                                                                                                                                   
  3. RecursiveComparisonAssert internally references RecursiveComparisonConfiguration                                                                                                                                                                
  4. RecursiveComparisonConfiguration uses ComparatorForPatterns (a package-private class)                                                                                                                                                           
                                                                                                                                                                                                                                                     
  RevAPI walks the full type hierarchy of your public API. Since GuardrailAssertions extends Assertions, all inherited public/protected methods from AssertJ become part of your module's API surface. When RevAPI follows those method signatures,  
  it reaches ComparatorForPatterns — a non-public class that consumers can't access.
```

mariofusco (approved):

PR #4915, opened 2026-04-14
Add MCP tool result extractor

### LINKED ISSUE #4794: [FEATURE] MCP: handle a successful result with text fallback differently to the default mechanism

I'm implementing an mcp based protocol that hooks in multiple parties. So I can't change it. 
I need to handle a successful result with text fallback differently to the default mechanism. 
I can't wire in a different ToolExecutionHelper to do this. 

Currently
ToolExecutionHelper.extractSuccessfulResult() does:
contentStream.map(content -> content.get("text").asText()).collect(Collectors.joining("\n"));

The protocol requires iterating content[] one item at a time and returning the first item that JSON-parses to a non-array object. I'd like the ability to do that in the default client. 

I'll write my own ToolExecutionHelper if necessary, but I need to be able to get that into the client. Shoving it into the builder as an override and making it non-static doesn't seem that hard.

Override the client, write and instantiate my own ToolExecutionHelper and get it in that way. It's not difficult but I'm concerned about breaking changes in subsequent versions.

**Additional context**
If you need context on the protocol its here: https://docs.adcontextprotocol.org/docs/building/implementation/mcp-response-extraction

see "text fallback" in the extraction algorithm.

leochame:
## Issue
Closes #4794

## Change
This PR adds a configurable `McpToolResultExtractor` SPI to `DefaultMcpClient.Builder` for MCP tool responses backed by `CallToolResult.result.content[]`.

The main goal is to let users override the default `content[]` parsing behavior for servers that return custom payloads there, while keeping the existing `structuredContent` handling unchanged.

Changes included:
- added a new `McpToolResultExtractor` extension point
- added `DefaultMcpToolResultExtractor` to preserve the current default behavior
- wired the extractor into `DefaultMcpClient` through `DefaultMcpClient.Builder#toolResultExtractor(...)`
- updated tool result extraction so application-level errors still surface consistently when a custom extractor is used
- clarified in Javadoc that this customization only applies to ordinary `content[]` tool results and does **not** apply to `structuredContent`, which is handled separately
- added unit tests for:
  - builder wiring and custom extractor usage
  - timeout fallback path
  - listener / application-level error path
  - default extractor behavior
  - `structuredContent` precedence and parsing
- added a dedicated MCP integration test fixture and stdio integration test covering custom extractor behavior for:
  - JSON returned as text content
  - image content
  - mixed multimodal content
  - `structuredContent` bypassing the custom extractor

## General checklist


jmartisk:
Thanks!
I think this is great! Could you please squash into a single commit?

jmartisk (approved):

jmartisk:
Actually I can squash it here, so no problem. Thanks!

PR #4916, opened 2026-04-14
Allows to configure subagents also as a Collection

mariofusco:
## Issue

Closes #

## Change

Small, non-breaking, improvement for the agentic API, also allowing to configure subagents as a Collection.

## General checklist

### DIFF
--- a/langchain4j-agentic/revapi.json
+++ b/langchain4j-agentic/revapi.json
@@ -1,0 +1,50 @@
+[
+  {
+    "extension": "revapi.differences",
+    "configuration": {
+      "ignore": true,
+      "differences": [
+        {
+          "ignore": true,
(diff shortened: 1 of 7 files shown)


mariofusco:
@dliubarskyi I find the `revapi` plugin very useful, but I believe we should really find a way to exclude internal classes from its checks, also because otherwise the `revapi.json` files will become quickly extremely verbose (see the one in this pull request for instance) and it will be hard to find useful information among all the unnecessary noise caused by the internal changes. If it is ok for you I could take care of this.

dliubarskyi:
@mariofusco sure, feel free to override the default revapi config for agentic modules. I totally understand that since these modules change fast now, revapi might make more troubles than help

dliubarskyi (approved):
@mariofusco thank you!

PR #4920, opened 2026-04-14
feat(bedrock): add CacheTTL support for prompt caching

dahaejung-29cm:
## Summary

Add configurable cache TTL support for Bedrock prompt caching. Currently, all cache points use the default 5-minute TTL because `CachePointBlock` is created without a TTL parameter. The AWS Bedrock SDK (`bedrockruntime` 2.41+) supports `CacheTTL` with `VALUE_5_M` (5 minutes) and `VALUE_1_H` (1 hour), but there is no way to configure it through LangChain4j.

## Problem

The `CACHE_POINT_BLOCK` in `AbstractBedrockChatModel` is hardcoded as a static final field without TTL:

```java
private static final SystemContentBlock CACHE_POINT_BLOCK = SystemContentBlock.builder()
        .cachePoint(CachePointBlock.builder()
                .type(CachePointType.DEFAULT)
                // no TTL set → defaults to 5 minutes
                .build())
        .build();
```

Users who need 1-hour TTL (e.g., for APIs called less frequently than every 5 minutes) must bypass LangChain4j and call the Bedrock SDK directly, losing access to LangChain4j's listener-based metrics, message conversion, and other features.

## Solution

- **`BedrockChatRequestParameters`** — Added `cacheTtl` field and `promptCaching(placement, ttl)` overload:
  ```java
  BedrockChatRequestParameters.builder()
      .promptCaching(BedrockCachePointPlacement.AFTER_SYSTEM, CacheTTL.VALUE_1_H)
      .build()
  ```

- **`AbstractBedrockChatModel`** — Replaced static `CACHE_POINT_BLOCK` with dynamic `buildCachePoint(cacheTtl)` method that creates `CachePointBlock` with the specified TTL. Applied to all 4 cache point creation sites (system messages, user messages, tool configurations).

- **`BedrockChatModel` + `BedrockStreamingChatModel`** — Pass `cacheTtl` through to all `extract*` methods.

## Backward Compatibility

**Fully backward-compatible.** When `cacheTtl` is `null` (default), the behavior is identical to the existing implementation:
- The existing `promptCaching(placement)` method continues to work unchanged
- `buildCachePoint(null)` returns a `CachePointBlock` without TTL (same as before)
- No changes required in existing user code

## Usage

```java
// Existing code — unchanged, still works (5-minute default)
BedrockChatRequestParameters.builder()
    .promptCaching(BedrockCachePointPlacement.AFTER_SYSTEM)
    .build()

// New — 1-hour TTL
BedrockChatRequestParameters.builder()
    .promptCaching(BedrockCachePointPlacement.AFTER_SYSTEM, CacheTTL.VALUE_5_M)
    .build()
```

## Changed Files

| File | Change |
|------|--------|
| `BedrockChatRequestParameters.java` | Added `cacheTtl` field + `promptCaching(placement, ttl)` overload |
| `AbstractBedrockChatModel.java` | Replaced static cache point with dynamic `buildCachePoint(cacheTtl)` |
| `BedrockChatModel.java` | Pass `cacheTtl` to extract methods |
| `BedrockStreamingChatModel.java` | Pass `cacheTtl` to extract methods |

## References

- [AWS Bedrock Prompt Caching docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html)
- AWS SDK `CacheTTL` enum: `VALUE_5_M`, `VALUE_1_H`
- 1-hour TTL supported models: Claude Sonnet 4.5, Claude Opus 4.5, Claude Haiku 4.5


langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockCacheTtl.java
```diff
@@ -0,0 +1,39 @@
+package dev.langchain4j.model.bedrock;
+
+import software.amazon.awssdk.services.bedrockruntime.model.CacheTTL;
+
+/**
+ * Cache TTL options for Bedrock prompt caching.
+ *
+ * @see <a href="https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html">AWS Bedrock Prompt Caching</a>
+ */
+public enum BedrockCacheTtl {
```
dliubarskyi: Why do we need this enum? Can't we use `CacheTTL` directly?

dahaejung-29cm: Makes sense, agreed. With only two values and the module already exposing SDK types like `CachePointBlock`, the wrapper wasn't really pulling its weight. Going through `CacheTTL` directly also means new TTL values get picked up automatically when the SDK adds them. Thanks for handling it!

dahaejung-29cm:
@dliubarskyi Just flagging that the build is now failing on Revapi (`java.class.externalClassExposedInAPI`) because `CacheTTL` from the AWS SDK is exposed in the public signature of `BedrockChatRequestParameters`. From what I can tell, the options would be either adding an ignore entry in `langchain4j-bedrock/revapi.json` if the SDK exposure is intentional, or restoring the wrapper. Thought it'd be good to share.

dliubarskyi:
@dahaejung-29cm thanks for the heads up! I have fixed the revapi config

dliubarskyi:
@dahaejung-29cm thank you for your PR!

PR #4923, opened 2026-04-14
issue #4893 - Introduce executionContext to AgenticScope

### LINKED ISSUE #4893: Add execution context support to AgenticScope for async planner implementations
# Add Execution Context Support to AgenticScope

## Problem Statement

When implementing custom `Planner` implementations that use **asynchronous workflow orchestration**, there's currently no way to pass execution-time state (like planner references or other non-serializable context) from the planner to workflow execution branches.

This forces implementers to use workarounds like global singletons for state management, which are error-prone and not idiomatic.

## Use Case: Quarkus Flow Integration

We're integrating LangChain4j's Agentic Workflow API with [Quarkus Flow](https://github.com/quarkiverse/quarkus-flow), a workflow orchestration engine based on CNCF Serverless Workflow spec.

### Architecture

```
┌─────────────────────────────────────────────────────────────────┐
│ LangChain4j Agentic System                                      │
│                                                                  │
│  ┌──────────────┐                                               │
│  │ FlowPlanner  │  implements Planner                           │
│  │ (our impl)   │                                               │
│  └──────┬───────┘                                               │
│         │                                                        │
│         │ Builds workflow definition                            │
│         │ from agents                                           │
│         ▼                                                        │
│  ┌──────────────────────────┐                                   │
│  │ Workflow Definition      │                                   │
│  │ (Serverless Workflow)    │                                   │
│  └──────────────────────────┘                                   │
│         │                                                        │
│         │ Executes async in ForkJoinPool                        │
│         ▼                                                        │
│  ┌──────────────────────────────────────────────┐               │
│  │ Parallel Fork (workflow branches)            │               │
│  │                                               │               │
│  │  Branch A       Branch B       Branch C      │               │
│  │  executes       executes       executes      │               │
│  │  agent 1        agent 2        agent 3       │               │
│  │     │              │              │           │               │
│  │     └──────────────┴──────────────┘           │               │
│  │                    │                          │               │
│  │         Need to call back to planner         │               │
│  │         to enqueue agent execution!          │               │
│  └──────────────────────────────────────────────┘               │
│                      │                                           │
│                      │ ❌ But how to get                         │
│                      │    planner reference?                     │
│                      ▼                                           │
│  Each branch only has:                                          │
│  - String instanceId                                            │
│  - AgenticScope scope                                           │
│                                                                  │
└─────────────────────────────────────────────────────────────────┘
```

### The Problem in Detail

**In `FlowPlanner.firstAction()`**, we build and start a workflow:

```java
@Override
public Action firstAction(PlanningContext planningContext) {
    WorkflowDefinition definition = buildWorkflow(planningContext.subagents());
    WorkflowInstance instance = definition.instance(planningContext.agenticScope());
    
    // Workflow executes asynchronously in ForkJoinPool
    instance.start();
    
    return internalNextAction();
}
```

**The workflow definition** (built from parallel agents) looks like:

```java
tasks.fork("parallel", fork -> {
    for (AgentInstance agent : agents) {
        fork.branches(withInstanceId(branchName, 
            (String instanceId, DefaultAgenticScope scope) -> {
                // ❌ PROBLEM: We're in a workflow branch (different thread)
                // We need to call: planner.executeAgent(agent)
                // But we only have: instanceId and scope
                // How do we get the planner reference?
                
                CompletableFuture<Void> future = ???.executeAgent(agent);
                future.join();
                return null;
            },
            DefaultAgenticScope.class));
    }
});
```

### Current Workaround: Global Singleton (Not Ideal)

We currently use a **JVM-wide singleton** to look up planners:

```java
// FlowPlannerSessions.java - A global singleton
public class FlowPlannerSessions {
    private static final FlowPlannerSessions INSTANCE = new FlowPlannerSessions();
    private final ConcurrentHashMap<String, FlowPlanner> sessions = new ConcurrentHashMap<>();
    
    public static FlowPlannerSessions getInstance() {
        return INSTANCE;
    }
    
    public void open(String instanceId, FlowPlanner planner) {
        sessions.put(instanceId, planner);
    }
    
    public FlowPlanner get(String instanceId) {
        return sessions.get(instanceId);
    }
}
```

**In workflow branches**:
```java
fork.branches(withInstanceId(branchName, 
    (String instanceId, DefaultAgenticScope scope) -> {
        // Look up planner from global singleton
        FlowPlanner planner = FlowPlannerSessions.getInstance().get(instanceId);
        CompletableFuture<Void> future = planner.executeAgent(agent);
        future.join();
        return null;
    }));
```

### Why This Is Problematic

1. **Global state**: Singleton shared across entire JVM
2. **Testing complexity**: Parallel tests need isolation (`@Execution(ExecutionMode.SAME_THREAD)`)
3. **Memory leaks**: Must manually clean up sessions in `finally` blocks
4. **Not idiomatic**: Doesn't follow LangChain4j's design patterns
5. **Thread-safety concerns**: Race conditions in session cleanup
6. **No framework support**: Can't leverage LangChain4j's lifecycle management

## Proposed Solution

Add **execution context support** to `AgenticScope` following the existing pattern for cached agent instances:

```java
// In DefaultAgenticScope.java
public class DefaultAgenticScope implements AgenticScope {
    // Existing fields...
    private final transient Map<String, Object> agents = new ConcurrentHashMap<>();
    
    // NEW: Execution context (similar to agents map)
    private final transient Map<Class<?>, Object> executionContexts = new ConcurrentHashMap<>();
    
    // NEW: Public API
    public <T> void setExecutionContext(Class<T> type, T context) {
        executionContexts.put(type, context);
    }
    
    public <T> T getExecutionContext(Class<T> type) {
        return (T) executionContexts.get(type);
    }
}
```

**Expose in interface**:
```java
// In AgenticScope.java
public interface AgenticScope extends LangChain4jManaged {
    // Existing methods...
    
    <T> void setExecutionContext(Class<T> type, T context);
    <T> T getExecutionContext(Class<T> type);
}
```

### Why This Design?

1. **Follows existing pattern**: Mirrors the `private transient Map<String, Object> agents` pattern already in `DefaultAgenticScope`
2. **`transient` keyword**: Won't be serialized (same as `agents` map)
3. **Type-safe**: Keyed by `Class<T>` for compile-time safety
4. **Thread-safe**: `ConcurrentHashMap` (same as agents)
5. **Scoped correctly**: Lives with the `AgenticScope` instance (one per workflow)
6. **Generic**: Any planner implementation can use it for their needs

## Usage Example

With this API, our code becomes:

**In FlowPlanner**:
```java
@Override
public Action firstAction(PlanningContext planningContext) {
    // Store planner reference in execution context
    planningContext.agenticScope().setExecutionContext(FlowPlanner.class, this);
    
    WorkflowDefinition definition = buildWorkflow(planningContext.subagents());
    WorkflowInstance instance = definition.instance(planningContext.agenticScope());
    instance.start();
    
    return internalNextAction();
}
```

**In workflow branches**:
```java
fork.branches(withInstanceId(branchName, 
    (String instanceId, DefaultAgenticScope scope) -> {
        // Get planner from execution context (no singleton!)
        FlowPlanner planner = scope.getExecutionContext(FlowPlanner.class);
        CompletableFuture<Void> future = planner.executeAgent(agent);
        future.join();
        return null;
    }));
```

**No more singleton!** ✨

### For Quarkus Flow
- ✅ Eliminate global singleton
- ✅ Proper scoping (one planner per scope)
- ✅ Thread-safe by design
- ✅ Easier testing (no shared state)
- ✅ Cleaner lifecycle management

### For LangChain4j Ecosystem
- ✅ Enables async/workflow-based planner implementations
- ✅ Provides standard way to pass execution context
- ✅ Doesn't break serialization (transient)
- ✅ Follows existing patterns in codebase
- ✅ Generic solution for future use cases

### For Other Implementers
Any planner that needs to:
- Execute workflows asynchronously
- Maintain execution-time state
- Coordinate across threads
- Integrate with external orchestration systems

Can use this instead of rolling their own singleton/global state management.

## Alternative Considered

**Why not use `scope.writeState()`?**

The state map is for **data** (serializable agent interaction state), not execution context:
- Gets serialized
- Part of the conversation/interaction model
- Intended for agent coordination data

Execution context is:
- Non-serializable (planner instances, thread pools, etc.)
- Implementation-specific
- Not part of the agent interaction model
- Should NOT be persisted

The `transient` keyword is critical here.

## Implementation Impact

**Minimal changes required**:
1. Add two methods to `AgenticScope` interface
2. Add `executionContexts` field to `DefaultAgenticScope` (same pattern as `agents`)
3. Implement getter/setter methods

**Backward compatible**:
- No existing code affected
- Pure addition (no modifications to existing behavior)
- Default implementation can return null if not set

## References

- Quarkus Flow: https://github.com/quarkiverse/quarkus-flow
- Our current workaround: [FlowPlannerSessions.java](https://github.com/quarkiverse/quarkus-flow/blob/main/langchain4j/runtime/src/main/java/io/quarkiverse/flow/langchain4j/workflow/FlowPlannerSessions.java)
- Our planner implementation: [FlowPlanner.java](https://github.com/quarkiverse/quarkus-flow/blob/main/langchain4j/runtime/src/main/java/io/quarkiverse/flow/langchain4j/workflow/FlowPlanner.java)

---

Would love to contribute the implementation if this proposal is accepted! 🙏

ricardozanini:
## Issue

Closes #4893

## Change

- Added to `AgenticScope` interface the ability to set an execution context to facilitate `Planner` API implementations.

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgenticScope.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgenticScope.java
@@ -175,4 +175,29 @@ public interface AgenticScope extends LangChain4jManaged {
     default Set<String> pendingResponseIds() {
         return Set.of();
     }
+
+    /**
+     * Stores non-serializable execution context in this scope, keyed by type.
+     * <p>
+     * This allows custom {@link dev.langchain4j.agentic.planner.Planner} implementations to store
+     * runtime objects that are needed during execution but should not be persisted.
+     * The execution context is stored in a {@code transient} map and will not be serialized.
+     *
+     * @param type    the class type to use as the key for this context (must not be {@code null})
+     * @param context the execution context instance to store (must not be {@code null})
+     * @param <T>     the type of the execution context
+     * @throws IllegalArgumentException if {@code context} is {@code null}
+     */
+    <T> void setExecutionContext(Class<T> type, T context);
+
+    /**
+     * Retrieves non-serializable execution context from this scope by type.
+     * <p>
+     * Returns execution context previously stored via {@link #setExecutionContext(Class, Object)}.
+     *
+     * @param type the class type used as the key for the execution context
+     * @param <T>  the type of the execution context
+     * @return the execution context instance previously stored for this type, or {@code null} if none exists
+     */
+    <T> T getExecutionContext(Class<T> type);
 }
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgenticScope.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgenticScope.java
@@ -177,27 +177,69 @@ public interface AgenticScope extends LangChain4jManaged {
     }
 
     /**
-     * Stores non-serializable execution context in this scope, keyed by type.
+     * Stores non-serializable execution context in this scope, keyed by name.
      * <p>
      * This allows custom {@link dev.langchain4j.agentic.planner.Planner} implementations to store
      * runtime objects that are needed during execution but should not be persisted.
      * The execution context is stored in a {@code transient} map and will not be serialized.
+     * <p>
+     * This is distinct from {@link #writeState(String, Object)}, which is for serializable
+     * agent interaction data that forms part of the conversation state.
      *
-     * @param type    the class type to use as the key for this context (must not be {@code null})
+     * @param key     the key to use for this context (must not be {@code null})
      * @param context the execution context instance to store (must not be {@code null})
-     * @param <T>     the type of the execution context
+     * @throws IllegalArgumentException if {@code key} or {@code context} is {@code null}
+     */
+    void writeExecutionContext(String key, Object context);
+
+    /**
+     * Stores non-serializable execution context in this scope, using the class name as the key.
+     * <p>
+     * This is a convenience method that delegates to {@link #writeExecutionContext(String, Object)}
+     * using the fully qualified class name as the key.
+     *
+     * @param type    the class type to use as the key for this context
+     * @param context the execution context instance to store
      * @throws IllegalArgumentException if {@code context} is {@code null}
      */
-    <T> void setExecutionContext(Class<T> type, T context);
+    default void writeExecutionContext(Class<?> type, Object context) {
+        writeExecutionContext(type.getName(), context);
+    }
+
+    /**
+     * Retrieves non-serializable execution context from this scope by key.
+     * <p>
+     * Returns execution context previously stored via {@link #writeExecutionContext(String, Object)}.
+     *
+     * @param key the key used to store the execution context
+     * @return the execution context instance previously stored for this key, or {@code null} if none exists
+     */
+    Object executionContext(String key);
+
+    /**
+     * Retrieves non-serializable execution context from this scope by key with type-safe casting.
+     * <p>
+     * Returns execution context previously stored via {@link #writeExecutionContext(String, Object)}.
+     *
+     * @param key  the key used to store the execution context
+     * @param type the expected type of the execution context (for type-safe casting)
+     * @param <T>  the type of the execution context
+     * @return the execution context instance previously stored for this key, or {@code null} if none exists
+     * @throws ClassCastException if the stored context cannot be cast to the expected type
+     */
+    <T> T executionContextAs(String key, Class<T> type);
 
     /**
      * Retrieves non-serializable execution context from this scope by type.
      * <p>
-     * Returns execution context previously stored via {@link #setExecutionContext(Class, Object)}.
+     * This is a convenience method that delegates to {@link #executionContextAs(String, Class)}
+     * using the fully qualified class name as the key.
      *
      * @param type the class type used as the key for the execution context
      * @param <T>  the type of the execution context
      * @return the execution context instance previously stored for this type, or {@code null} if none exists
      */
-    <T> T getExecutionContext(Class<T> type);
+    default <T> T executionContextAs(Class<T> type) {
+        return executionContextAs(type.getName(), type);
+    }
 }
--- a/langchain4j-agentic/revapi.json
+++ b/langchain4j-agentic/revapi.json
@@ -5,6 +5,31 @@
       "ignore": true,
       "differences": [
         {
+          "code": "java.method.addedToInterface",
+          "new": "method void dev.langchain4j.agentic.scope.AgenticScope::writeExecutionContext(java.lang.String, java.lang.Object)",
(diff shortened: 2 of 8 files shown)


0xbrainkid:
Execution context on `AgenticScope` is the right addition for Planner API implementations — without it, each planner step operates without access to the broader agentic context, making it impossible to propagate trust, authorization, or provenance information through multi-step plans.

One dimension worth ensuring is captured in `executionContext`: **agent identity and trust metadata.** A Planner that spawns sub-steps or delegates to specialized agents needs to carry both the calling agent identity and its behavioral trust score through the execution context, so downstream steps can apply appropriate trust policies without re-querying.

A minimal addition to whatever `executionContext` structure is defined:

```java
public class ExecutionContext {
    // Planner-relevant fields
    private final String planId;
    private final int stepIndex;
    
    // Identity & trust provenance
    private final String callerAgentId;        // who initiated this plan
    private final double callerTrustScore;     // behavioral trust at plan start
    private final String attestationUrl;        // verifiable trust proof
    private final Instant contextCreatedAt;    // for decay calculation
}
```

With these fields in context, a downstream tool call can log `callerAgentId` alongside its output, creating an attributable execution trace — not just "this tool was called" but "this specific agent (with trust 0.88) called this tool as step 3 of plan X." This satisfies EU AI Act Article 12 traceability requirements without requiring a separate observability layer.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgenticScope.java
```diff
@@ -175,4 +175,29 @@ default boolean completePendingResponse(String responseId, Object value) {
...
+     * This allows custom {@link dev.langchain4j.agentic.planner.Planner} implementations to store
+     * runtime objects that are needed during execution but should not be persisted.
+     * The execution context is stored in a {@code transient} map and will not be serialized.
+     *
+     * @param type    the class type to use as the key for this context (must not be {@code null})
+     * @param context the execution context instance to store (must not be {@code null})
+     * @param <T>     the type of the execution context
+     * @throws IllegalArgumentException if {@code context} is {@code null}
+     */
+    <T> void setExecutionContext(Class<T> type, T context);
```
mariofusco: I'm doubtful about using the class of the context as key. Couldn't you have 2 different `String` s in the context?

mariofusco (changes requested):
We introduced the RevAPI plugin to check backward compatibility of our changes. Please add exclusions in the configuration file accordingly with this change.

Also check my comment about having the class as key in the context.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/AgenticScope.java
ricardozanini: Typically, an execution should be a unique complex type. I understand the concern. I'll make a minor change while keeping the reader contract as is.

mariofusco (approved):

PR #4935, opened 2026-04-15
Fix #4773: Gemini tools must serialize as array, not bare object

### LINKED ISSUE #4773: [BUG] `langchain4j-google-ai-gemini` serializes non-canonical Gemini `tools` shape (`object` instead of `array`)

anuragg-saxenaa:
### Problem
When serializing a Gemini request with a single tool, the output JSON was producing:
```json
"tools": { ... }
```
instead of the required:
```json
"tools": [{ ... }]
```

### Solution
Added regression tests to catch this issue:
- `GeminiToolsSerializationTest.java` — JUnit test verifying array serialization
- `ToolSerializationTest.java` — standalone demo/test for the serialization behavior

### Files
- `langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GeminiToolsSerializationTest.java`
- `langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/ToolSerializationTest.java`

### References
- Issue: langchain4j/langchain4j#4773

### DIFF
--- a/langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/ToolSerializationTest.java
+++ b/langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/ToolSerializationTest.java
@@ -13,7 +13,11 @@ class ToolSerializationTest {
                 .name("getWeather")
                 .description("Get the weather")
                 .parameters(JsonObjectSchema.builder()
-                        .addProperty("city", JsonStringSchema.builder().description("The city name").build())
+                        .addProperty(
+                                "city",
+                                JsonStringSchema.builder()
+                                        .description("The city name")
+                                        .build())
                         .required("city")
                         .build())
                 .build();
--- a/langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/ToolSerializationTest.java
+++ b/langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/ToolSerializationTest.java
@@ -1,0 +1,44 @@
+package dev.langchain4j.model.googleai;
+
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
+import dev.langchain4j.model.chat.request.json.JsonStringSchema;
+import dev.langchain4j.model.googleai.GeminiGenerateContentRequest.GeminiTool;
+import java.util.List;
+
+class ToolSerializationTest {
+    public static void main(String[] args) {
+        // Create a single tool spec
+        ToolSpecification spec = ToolSpecification.builder()
+                .name("getWeather")
+                .description("Get the weather")
+                .parameters(JsonObjectSchema.builder()
+                        .addProperty("city", JsonStringSchema.builder().description("The city name").build())
+                        .required("city")
+                        .build())
+                .build();
+
+        List<ToolSpecification> specs = List.of(spec);
+        List<GeminiTool> tools = FunctionMapper.fromToolSpecsToGTools(specs, false, false, false, false, false);
+
+        System.out.println("GeminiTool list size: " + tools.size());
+        System.out.println("GeminiTool: " + tools);
+
+        // Now build a request with tools
+        var request = GeminiGenerateContentRequest.builder()
+                .contents(List.of())
+                .tools(tools)
+                .build();
+
+        String json = Json.toJsonWithoutIndent(request);
+        System.out.println("Serialized request:");
+        System.out.println(json);
+
+        // Check if tools is array or object
+        if (json.contains("\"tools\":{")) {
+            System.out.println("\nBUG: tools is serialized as object, not array!");
+        } else if (json.contains("\"tools\":[")) {
+            System.out.println("\nOK: tools is correctly serialized as array");
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/ToolSerializationTest.java
```diff
@@ -0,0 +1,48 @@
+package dev.langchain4j.model.googleai;
+
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.model.chat.request.json.JsonObjectSchema;
+import dev.langchain4j.model.chat.request.json.JsonStringSchema;
+import dev.langchain4j.model.googleai.GeminiGenerateContentRequest.GeminiTool;
+import java.util.List;
+
+class ToolSerializationTest {
+    public static void main(String[] args) {
```
dliubarskyi, line 10: this is not a proper test

dliubarskyi:
@anuragg-saxenaa thank you, but this PR does not seem useful

PR #4940, opened 2026-04-16
Add guardrailName() to GuardrailExecutedEvent for observability

### LINKED ISSUE #4938: [FEATURE] Add guardrailName() to GuardrailExecutedEvent
## Feature: Add `guardrailName()` to `GuardrailExecutedEvent`

  Related discussion: https://github.com/langchain4j/langchain4j/discussions/4914

  ---

  ## Problem

  `GuardrailExecutedEvent.guardrailClass()` returns the adapter/wrapper class when guardrails are wrapped (e.g. decorator pattern).

  This makes it difficult to identify the actual guardrail in observability systems, since all executions appear as the adapter class instead of the logical guardrail.

  ```java
  // Current behavior — all guardrails report the same name
  event.guardrailClass().getSimpleName() // → "InputGuardrailAdapter"

  // Expected with guardrailName()
  event.guardrailName() // → "PromptInjectionGuardrail"
 ```
  ---
  Proposed solution

  Add a non-breaking method:
   ```java
  default String guardrailName() {
      return guardrailClass().getSimpleName();
  }
 ```
  Additionally, propagate the logical guardrail name at execution time so decorators can expose the underlying guardrail identity by overriding ```Guardrail.name()```.

  ---
  Notes

  - Backward compatible (default implementation)
  - Improves observability (e.g. OpenInference ```validator_name```)

  ---
  Happy to contribute a PR.

nhthinh-axonivy:
## Issue
Closes #4938

## Change

Adds `guardrailName()` to `GuardrailExecutedEvent` to improve observability, especially when guardrails are wrapped using decorators.

### Details

- Added `guardrailName()` to `GuardrailExecutedEvent` with a default implementation returning `guardrailClass().getSimpleName()`
- Propagated guardrail name from execution layer to observability events
- Added a default `name()` method to `Guardrail` interface to represent logical guardrail identity
- Updated event builder and default implementation to carry `guardrailName`
- Added tests covering wrapped guardrails for both success and failure scenarios

This allows listeners to distinguish logical guardrails (e.g. `PromptInjectionGuardrail`) instead of only seeing adapter classes (e.g. `InputGuardrailAdapter`).

Related discussion: https://github.com/langchain4j/langchain4j/discussions/4914

## General checklist
- [x] I have manually run all the unit and integration tests in the core and main modules, and they are all green
- [x] I have added/updated the documentation (Javadoc on Guardrail.name() andGuardrailExecutedEvent.guardrailName())

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/guardrail/Guardrail.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/guardrail/Guardrail.java
@@ -10,6 +10,18 @@ package dev.langchain4j.guardrail;
  *            The type of the {@link GuardrailResult}
  */
 public interface Guardrail<P extends GuardrailRequest, R extends GuardrailResult<R>> {
+
+    /**
+     * Returns the logical name of this guardrail.
+     *
+     * Wrappers/decorators can override this method to expose the wrapped guardrail's name.
+     *
+     * @return the logical guardrail name
+     */
+    default String name() {
+        return getClass().getSimpleName();
+    }
+
     /**
      * Validate the interaction between the model and the user in one of the two directions.
      *
--- a/langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/GuardrailExecutedEvent.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/GuardrailExecutedEvent.java
@@ -45,14 +45,12 @@ public interface GuardrailExecutedEvent<
     Class<G> guardrailClass();
 
     /**
-     * Retrieves the guardrail name associated with the validation process.
-     *
-     * @return the guardrail name. Defaults to {@code guardrailClass().getSimpleName()} when not explicitly set.
-     */
-    default String guardrailName() {
-        return guardrailClass().getSimpleName();
-    }
-
+    * Retrieves the guardrail name associated with the validation process.
+    *
+    * @return the guardrail name, as set by the executor (see {@link Guardrail#name()}).
+    */
+    String guardrailName();
+    
     /**
      * Retrieves the duration of the guardrail execution.
      *
--- a/langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/GuardrailExecutedEvent.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/GuardrailExecutedEvent.java
@@ -45,6 +45,15 @@ public interface GuardrailExecutedEvent<
     Class<G> guardrailClass();
 
     /**
+     * Retrieves the guardrail name associated with the validation process.
+     *
+     * @return the guardrail name. Defaults to {@code guardrailClass().getSimpleName()} when not explicitly set.
+     */
+    default String guardrailName() {
+        return guardrailClass().getSimpleName();
+    }
+
+    /**
      * Retrieves the duration of the guardrail execution.
      *
      * @return the duration of the guardrail validation process.
@@ -61,6 +70,7 @@ public interface GuardrailExecutedEvent<
         private P request;
         private R result;
         private Class<G> guardrailClass;
+        private String guardrailName;
         private Duration duration;
 
         protected GuardrailExecutedEventBuilder() {}
@@ -70,6 +80,7 @@ public interface GuardrailExecutedEvent<
             request(src.request());
             result(src.result());
             guardrailClass(src.guardrailClass());
+            guardrailName(src.guardrailName());
             duration(src.duration());
         }
 
@@ -79,6 +90,10 @@ public interface GuardrailExecutedEvent<
 
         public P request() {
             return request;
+        }
+
+        public String guardrailName() {
+            return guardrailName;
         }
 
         public R result() {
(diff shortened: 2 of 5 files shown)


langchain4j-core/src/main/java/dev/langchain4j/guardrail/Guardrail.java
```diff
@@ -10,6 +10,18 @@
...
+
+    /**
+     * Returns the logical name of this guardrail.
+     *
+     * Wrappers/decorators can override this method to expose the wrapped guardrail's name.
+     *
+     * @return the logical guardrail name
+     */
+    default String name() {
+        return getClass().getSimpleName();
```
edeandrea, line 22: I'm hesitant to use reflection to compute this value each and every time its needed.

Maybe leave it as an interface-level method here then set/override in the first class in the hierarchy?

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/GuardrailExecutedEvent.java
```diff
@@ -44,6 +44,15 @@ public interface GuardrailExecutedEvent<
      */
     Class<G> guardrailClass();
 
+    /**
+     * Retrieves the guardrail name associated with the validation process.
+     *
+     * @return the guardrail name. Defaults to {@code guardrailClass().getSimpleName()} when not explicitly set.
+     */
+    default String guardrailName() {
+        return guardrailClass().getSimpleName();
```
edeandrea, line 53: I'm hesitant to use reflection to compute this value each and every time its needed.

Maybe leave it as an interface-level method here then set/override in the first class in the hierarchy?

langchain4j-core/src/test/java/dev/langchain4j/guardrail/InputGuardrailExecutorTests.java
```diff
@@ -1,14 +1,19 @@
 package dev.langchain4j.guardrail;
 
 import static dev.langchain4j.test.guardrail.GuardrailAssertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThat;
```
edeandrea: How does this work?  You shouldn't need to import `org.assertj.core.api.Assertions.assertThat` since `GuardrailAssertions.assertThat` inherits from `org.assertj.core.api.Assertions.assertThat`

edeandrea (commented):
See my comments in-line.  I'd prefer not to use reflection each time the name is needed. Instead, can we compute it once in the base class and re-use it?

langchain4j-core/src/main/java/dev/langchain4j/guardrail/Guardrail.java
nhthinh-axonivy: Thanks for the review,

  I’ll keep `name()` as a lightweight interface-level method and rely on concrete guardrails or wrappers to override it when they need to expose a logical name (e.g. delegating to the inner guardrail).

  The default implementation is just a fallback, and the executor sets the value once and passes it to the event.

  Happy to adjust if you’d prefer removing the default entirely.

edeandrea: Got it. Thanks for the clarification. We need to default it _somewhere_ in the hierarchy - otherwise its not a backwards-compatible change.

I was just trying to avoid reflection if at all possible. But I suppose whether we do it this way or we compute it once upon construction and set it, its still only happening once, so I think its ok.

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/GuardrailExecutedEvent.java
nhthinh-axonivy: Got it, thanks for the clarification.

I’ve removed the default implementation from `guardrailName()` on the interface and kept the fallback in `DefaultGuardrailExecutedEvent` to preserve backward compatibility.

The executor now passes the logical name from `Guardrail.name()` as intended.

nhthinh-axonivy: I’ve restored the default implementation of `guardrailName()` on the interface to satisfy Revapi and avoid breaking existing implementations.

The executor still sets the logical name via `Guardrail.name()`, so when available, the correct (non-wrapper) guardrail name is used in observability. The default now effectively serves as a fallback for older or external implementations that don’t provide it.

This keeps the behavior backward compatible while still enabling accurate guardrail identification for wrapped/decorated guardrails.

nhthinh-axonivy:
Hi @edeandrea, just checking in. let me know if there's anything else you'd like me to adjust!

edeandrea:
Sorry @nhthinh-axonivy I was waiting for the PR to be marked as "ready" before I did another review. Generally when I see PR's that aren't marked "ready" I figure the author is still working on it.

nhthinh-axonivy:
Thank @edeandrea I marked it as ready. Please feel free to let me know if anything need to adjust. Happy to contribute.

nhthinh-axonivy:
Hi @edeandrea, any update on this PR?

edeandrea:
Sorry @nhthinh-axonivy I completely lost track of it. Thank you for pinging me. I will add it to my to-do list for today!

edeandrea (approved):

edeandrea:
Hi @dliubarskyi I've approved this but I'm not able to merge it (I don't have merge privs). Could you please merge it when you get a chance? Thank you!

dliubarskyi:
> I don't have merge privs

That is strange, you have a write role in this repo... I have just added you to some other priv lists (related to branch protection rules), could you please try again?

edeandrea:
> > I don't have merge privs
> 
> That is strange, you have a write role in this repo... I have just added you to some other priv lists (related to branch protection rules), could you please try again?

Now I have the privs!

PR #4947, opened 2026-04-16
fix: merge all choices in OpenAI response to avoid silent tool call loss

### LINKED ISSUE #4931: [BUG] OpenAiChatModel ignores tool_calls when an OpenAI-compatible provider returns them in a later choice
### Summary

`OpenAiChatModel` currently ignores tool calls if an OpenAI-compatible provider returns:

- assistant text in `choices[0]`
- `tool_calls` in `choices[1]`
use gpt-5.4

This happens because `OpenAiUtils.aiMessageFrom(...)` only parses `choices.get(0)`.

### Versions

Reproduced with:

- `langchain4j-open-ai 1.12.2`
- `langchain4j-open-ai 1.13.0`

### Actual behavior

If the provider returns a response like this:

```json
{
  "id": "chatcmpl-xxx",
  "object": "chat.completion",
  "model": "compatible-model",
  "choices": [
    {
      "index": 0,
      "finish_reason": "stop",
      "message": {
        "role": "assistant",
        "content": "I will first inspect the target file."
      }
    },
    {
      "index": 1,
      "finish_reason": "tool_calls",
      "message": {
        "role": "assistant",
        "content": null,
        "tool_calls": [
          {
            "id": "call_123",
            "type": "function",
            "function": {
              "name": "readFile",
              "arguments": "{\"path\":\"src/main/java/example/Foo.java\"}"
            }
          }
        ]
      }
    }
  ]
}
```
LangChain4j produces an AiMessage that contains only the text from choices[0].
So:
- aiMessage.text() is present
- aiMessage.toolExecutionRequests() is empty
This breaks tool execution flows.
Expected behavior
LangChain4j should not silently lose tool calls from later choices.
A more robust behavior could be one of these:
1. Merge multiple choices into one AiMessage
2. Prefer the choice containing tool_calls
3. Fail explicitly with a clear exception if the response shape is unsupported
Why this happens
From local inspection, OpenAiUtils.aiMessageFrom(...) still does:
response.choices().get(0)
and only parses content / tool calls from that first choice.
Notes
This can still happen even when n=1 is requested. Some OpenAI-compatible providers return multiple choices anyway.
OpenAiChatModel.doChat(...) still relies on OpenAiUtils.aiMessageFrom(...), so the issue remains in 1.13.0.

wade-microblueworld:
## Summary

When an OpenAI-compatible provider returns `tool_calls` in a later choice (e.g. `choices[1]`) while text is in `choices[0]`, the previous implementation only read `choices.get(0)`, silently discarding tool calls and breaking tool execution flows.

## Fix

Iterate over **all** choices in the response:
- Collect text from the first non-empty choice
- Aggregate tool execution requests from all choices
- Guard against empty choices list with a clear exception

Also adds: guard for empty choices list with descriptive exception.

Fixes #4931

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java
@@ -348,34 +348,49 @@ public class OpenAiUtils {
     }
 
     public static AiMessage aiMessageFrom(ChatCompletionResponse response, boolean returnThinking) {
-        AssistantMessage assistantMessage = response.choices().get(0).message();
+        List<ChatChoice> choices = response.choices();
 
-        String refusal = assistantMessage.refusal();
-        if (isNotNullOrBlank(refusal)) {
-            throw new ContentFilteredException(refusal);
+        if (isNullOrEmpty(choices)) {
+            throw new IllegalArgumentException("Chat completion returned no choices");
         }
 
-        String content = assistantMessage.content();
-
+        String content = null;
         String reasoningContent = null;
-        if (returnThinking) {
-            reasoningContent = assistantMessage.reasoningContent();
-        }
+        List<ToolExecutionRequest> toolExecutionRequests = new ArrayList<>();
 
-        List<ToolExecutionRequest> toolExecutionRequests =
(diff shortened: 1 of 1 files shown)


dliubarskyi:
Closing in favour of https://github.com/langchain4j/langchain4j/pull/5077

PR #4955, opened 2026-04-17
OpenAI Official Responses: add server tools support

### LINKED ISSUE #4838: [FEATURE] OpenAiOfficialResponsesStreamingChatModel - web_search tool
At the moment, `OpenAiOfficialResponsesStreamingChatModel` does not seem to support passing OpenAI native built-in tools such as `web_search`.

This is limiting when working with the OpenAI Responses API, because one of its main advantages is the ability to use built-in tools directly through the `tools` array. Without support for these native tools in `OpenAiOfficialResponsesStreamingChatModel`, developers cannot fully benefit from the official OpenAI capabilities while keeping streaming behavior in the same integration.

I would like `OpenAiOfficialResponsesStreamingChatModel` to support OpenAI native built-in tools, especially `web_search`, and ideally other built-in tools available in the Responses API as well.

This would make it possible to use official OpenAI tools like `web_search` directly from `OpenAiOfficialResponsesStreamingChatModel` without custom workarounds.

Relevant OpenAI documentation:
https://developers.openai.com/api/docs/guides/tools-web-search/

**Additional context**
According to the official OpenAI documentation, the Responses API supports built-in tools such as `web_search` through the `tools` field in the request. Supporting this natively in `OpenAiOfficialResponsesStreamingChatModel` would make the integration more complete and aligned with OpenAI's official API design.

leochame:
## Issue
Closes #4838

## Change
This PR covers OpenAI Official Responses support only.

- added `serverTools(...)` support to `OpenAiOfficialResponsesStreamingChatModel`
- added `rawResponse()` support to `OpenAiOfficialResponsesChatResponseMetadata`

### Validation
- ran `./mvnw -pl langchain4j-open-ai-official spotless:apply`
- ran `./mvnw -pl langchain4j-open-ai-official spotless:check`
- verified module unit and integration tests passed locally

## General checklist


DragonFSKY:
Hi, I’ve also been looking into hosted `tool_search` on the OpenAI Official Responses side, so this PR caught my attention.

I opened a related issue as well: #4965. From what I understand, you’ve already wired up the base flow for the generic `tool_search` type here, and I was wondering whether you’re planning to continue building on top of this.

The two things I’m mainly curious about are:
- whether `GPT-5.4` hosted `tool_search` will get more concrete support / integration coverage
- whether the continuation flow will also be handled when hosted MCP returns an `mcp_approval_request`

I’ve also been looking at a nearby flow locally, so I just wanted to better understand the direction here.

leochame:
> Hi, I’ve also been looking into hosted `tool_search` on the OpenAI Official Responses side, so this PR caught my attention.
> 
> I opened a related issue as well: #4965. From what I understand, you’ve already wired up the base flow for the generic `tool_search` type here, and I was wondering whether you’re planning to continue building on top of this.
> 
> The two things I’m mainly curious about are:
> 
> * whether `GPT-5.4` hosted `tool_search` will get more concrete support / integration coverage
> * whether the continuation flow will also be handled when hosted MCP returns an `mcp_approval_request`
> 
> I’ve also been looking at a nearby flow locally, so I just wanted to better understand the direction here.

yes for hosted tool_search, not yet for the broader continuation flow.

For hosted tool_search, this branch now has concrete support at the OpenAI Official Responses layer.

For the continuation flow, that is still only partially covered. We can already surface server-side outputs such as mcp_approval_request, but the full continuation path after that request is not implemented yet in this branch. So hosted MCP approval handling is still follow-up work.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -1178,6 +1283,22 @@ private void handleReasoningSummaryTextDelta(ResponseReasoningSummaryTextDeltaEv
             }
         }
 
+        private void handleWebSearchCallInProgress(Object event) {
```
dliubarskyi: what's the point of adding these empty handlers?

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -1245,11 +1366,18 @@ private void handleIncomplete(ResponseIncompleteEvent event) {
...
+            List<OpenAiOfficialServerToolResult> serverToolResults =
+                    returnServerToolResults ? extractServerToolResults(response.output()) : List.of();
             var aiMessage = buildAiMessage(
                     text,
                     extractReasoningSummary(response),
                     completedToolCalls,
                     extractEncryptedReasoning(response));
+            if (!serverToolResults.isEmpty()) {
+                Map<String, Object> attributes = new HashMap<>(aiMessage.attributes());
+                attributes.put(SERVER_TOOL_RESULTS_KEY, serverToolResults);
```
dliubarskyi: What about exposing server tool results in `OpenAiOfficialChatResponseMetadata` instead?
Then we could also remove `returnServerToolResults` and return results always without worries. WDYT?

I know that I have asked you to follow Anthropic's implementation, but after a second thought I have realized that Anthropic's implementation is not the best example. Sorry about that.

dliubarskyi: `AiMessage.attributes` are meant to store attributes that should be sent back to the LLM on the follow-up call. For example, encrypted reasoning signatures. But in case of server tool results, it should not be sent back and doesn't necessarily has to be persisted. It is just some metadata about how chat request was handled.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
dliubarskyi, line 1: BTW, we now have `OpenAiOfficialResponsesChatModel` as well, it would be nice to have a similar implementation there

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
```diff
@@ -114,13 +117,19 @@ public class OpenAiOfficialResponsesStreamingChatModel implements StreamingChatM
...
+            Set.of("compaction", "custom_tool_call", "function_call", "message", "reasoning");
+    private static final TypeReference<Map<String, Object>> STRING_OBJECT_MAP = new TypeReference<>() {};
     // do not change, will break backward compatibility!
     static final String ENCRYPTED_REASONING_KEY = "encrypted_reasoning";
 
     private final OpenAIClient client;
     private final ExecutorService executorService;
     private final OpenAiOfficialResponsesChatRequestParameters defaultRequestParameters;
     private final List<ChatModelListener> listeners;
+    private final List<OpenAiOfficialServerTool> serverTools;
```
dliubarskyi: it would be nice to add `List<OpenAiOfficialServerTool> serverTools` to `OpenAiOfficialResponsesChatRequestParameters` and follow the same merging strategy, see `this.defaultRequestParameters = OpenAiOfficialResponsesChatRequestParameters.builder()` below

dliubarskyi: At the end, there will be no need in a dedicated `List<OpenAiOfficialServerTool> serverTools` field in this class, as tools will be stored in default reqeust params

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialServerToolMapper.java
```diff
@@ -0,0 +1,697 @@
...
+
+    private static boolean isSupportedWebSearchPreviewToolType(String type) {
+        return type != null && type.startsWith("web_search_preview");
+    }
+
+    private static boolean isSupportedComputerUsePreviewToolType(String type) {
+        return type != null && type.startsWith("computer_use_");
+    }
+
+    private static WebSearchTool toWebSearchTool(OpenAiOfficialServerTool serverTool) {
```
dliubarskyi: This class is pretty long and complicated, and it seems that there is no benefit of using `OpenAiOfficialServerTool` in order to then map it to specific Official SDK classes. What about getting rid of `OpenAiOfficialServerTool` and ask users to specify Official SDK classes directly? Better type safety and less mapping back and forth. WDYT?

dliubarskyi (commented):
@leochame thank you!

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
leochame: Good point. I removed the empty handler methods and kept these event branches inline in `handleEvent`, since they only need to be explicitly ignored and do not require separate behavior.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
leochame: Addressed. `serverToolResults` now live in `OpenAiOfficialResponsesChatResponseMetadata` instead of `AiMessage.attributes`.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
leochame: Addressed. The non-streaming `OpenAiOfficialResponsesChatModel` now exposes server tool results in the same way.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesStreamingChatModel.java
leochame: Addressed. `serverTools` now live in `OpenAiOfficialResponsesChatRequestParameters` and follow the standard merge/defaulting flow.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialServerToolMapper.java
leochame: Addressed. The wrapper/mapper layer was removed and the API now accepts Official SDK `Tool` instances directly.

leochame:
@dliubarskyi  Thanks, these comments all point in the same direction and I have now aligned the implementation with that.

In the latest revision I:
- removed the `OpenAiOfficialServerTool` / mapper abstraction and now accept Official SDK `Tool` instances directly
- moved `serverTools` into `OpenAiOfficialResponsesChatRequestParameters` so they follow the normal merge/defaulting flow
- moved `serverToolResults` from `AiMessage.attributes` into `OpenAiOfficialResponsesChatResponseMetadata`
- removed `returnServerToolResults` and now expose server tool results in metadata consistently
- aligned both streaming and non-streaming Responses models
- removed the empty/no-op streaming handlers

I kept `encrypted_reasoning` in `AiMessage.attributes`, since that is continuation state, while server tool results are now treated as response metadata.

----

BTW,  running spotless:apply or fully adjusting everything to satisfy spotless:check would reformat a lot of files and code that I did not actually touch for this change. I felt that including those unrelated formatting changes would make the review harder, since the diff would become much larger and less focused on the actual OpenAI Official implementation.

For that reason, I intentionally did not apply broad formatting-only changes outside the files that were directly relevant to this work.

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialServerToolResult.java
```diff
@@ -0,0 +1,93 @@
...
+import dev.langchain4j.Experimental;
+import java.util.Objects;
+
+/**
+ * Result emitted by an OpenAI Responses API server-side tool.
+ *
+ * <p>Content is stored as raw {@link Object} to remain compatible with provider-side schema changes.
+ */
+@Experimental
+public class OpenAiOfficialServerToolResult {
```
dliubarskyi: I think we can get rid of this class and expose OpenAI Official SDK types instead

langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialResponsesChatResponseMetadata.java
```diff
@@ -3,20 +3,25 @@
...
 
+import static dev.langchain4j.internal.Utils.copy;
+
 @Experimental
 public class OpenAiOfficialResponsesChatResponseMetadata extends ChatResponseMetadata {
 
     private final Long createdAt;
     private final Long completedAt;
     private final String serviceTier;
+    private final List<OpenAiOfficialServerToolResult> serverToolResults;
```
dliubarskyi: I would expose `com.openai.models.responses.Response rawResponse` instead, WDYT? This will make sure user has access to all available information

leochame: Addressed. I removed OpenAiOfficialServerToolResult and now expose `com.openai.models.responses.Response rawResponse`  in OpenAiOfficialResponsesChatResponseMetadata.

dliubarskyi (approved):
@leochame thank you!

dliubarskyi:
Would you like to implement the same feature for the unofficial module (`OpenAiResponsesChatModel`/`OpenAiResponsesStreamingChatModel`)?

In case of an unofficial implementation, it will require adding a new type (e.g., `OpenAiServerTool`, similar to `AnthropicServerTool`), because we do not use any SDK in unofficial module.

leochame:
> Would you like to implement the same feature for the unofficial module (`OpenAiResponsesChatModel`/`OpenAiResponsesStreamingChatModel`)?
> 
> In case of an unofficial implementation, it will require adding a new type (e.g., `OpenAiServerTool`, similar to `AnthropicServerTool`), because we do not use any SDK in unofficial module.

@dliubarskyi Yes, I can take that on.

PR #4963, opened 2026-04-19
docs: add Declarative Agent Pattern documentation

Lumjiel:
## Summary


mariofusco:
The Declarative API is indeed an API alternative and complementary to the programmatic one and not a pattern at all. That said this pull request seems entirely AI generated and doesn't harmonize at all with the rest of the agents documentation.

PR #4972, opened 2026-04-20
Fix multi-choice tool_calls extraction in OpenAiUtils

### LINKED ISSUE #4931: [BUG] OpenAiChatModel ignores tool_calls when an OpenAI-compatible provider returns them in a later choice

anuragg-saxenaa:
## Fix multi-choice tool_calls extraction in OpenAiUtils

Fixes https://github.com/langchain4j/langchain4j/issues/4931

### Problem
`OpenAiUtils.aiMessageFrom(...)` only parsed `choices.get(0)`, silently dropping tool_calls from later choices. This breaks OpenAI-compatible providers (e.g. gpt-5.4) that split text and tool_calls across multiple choices.

### Solution
`aiMessageFrom` now iterates **all** choices and merges:
- `tool_calls` from every choice that has them
- `content` from the first choice that has it (preferring choices without tool_calls)
- `reasoningContent` from the first choice that has it
- legacy `functionCall` from every choice

### Changes
- `OpenAiUtils.java`: loop over all choices instead of just `choices.get(0)`
- `OpenAiUtilsTest.java`: two new regression tests:
  - `should_extract_tool_calls_from_all_choices_when_text_and_tools_are_split_across_choices`
  - `should_merge_tool_calls_when_multiple_choices_have_tool_calls`

### Testing
`mvn test -pl langchain4j-open-ai -Dtest=OpenAiUtilsTest` → 14 tests pass (all existing + 2 new)

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java
@@ -51,11 +51,11 @@ import dev.langchain4j.model.openai.OpenAiTokenUsage;
 import dev.langchain4j.model.openai.OpenAiTokenUsage.InputTokensDetails;
 import dev.langchain4j.model.openai.OpenAiTokenUsage.OutputTokensDetails;
 import dev.langchain4j.model.openai.internal.chat.AssistantMessage;
+import dev.langchain4j.model.openai.internal.chat.ChatCompletionChoice;
 import dev.langchain4j.model.openai.internal.chat.ChatCompletionRequest;
 import dev.langchain4j.model.openai.internal.chat.ChatCompletionResponse;
 import dev.langchain4j.model.openai.internal.chat.ContentType;
 import dev.langchain4j.model.openai.internal.chat.Function;
-import dev.langchain4j.model.openai.internal.chat.ChatCompletionChoice;
 import dev.langchain4j.model.openai.internal.chat.FunctionCall;
 import dev.langchain4j.model.openai.internal.chat.FunctionMessage;
 import dev.langchain4j.model.openai.internal.chat.ImageDetail;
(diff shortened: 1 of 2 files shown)


dliubarskyi:
@anuragg-saxenaa thank you, but current implementation is not what we want, we want to fail (throw an exception) in case response contains multiple choices, see https://github.com/langchain4j/langchain4j/issues/4931#issuecomment-4326887265

anuragg-saxenaa:
Got it — understood. The current implementation merges choices (fixes the crash), but you want it to fail loudly instead (throw an exception). I'm working on the revised implementation now and will push shortly.

dliubarskyi:
Feel free to reopen when it will be ready

PR #4975, opened 2026-04-21
feat: add ReturnBehavior.IMMEDIATE_IF_LAST for agentic halt-tool pattern

### LINKED ISSUE #4974: Add IMMEDIATE_IF_LAST ReturnBehavior to honor halt tools placed last in multi-tool batches
### Summary

Add a new `ReturnBehavior.IMMEDIATE_IF_LAST` enum value that halts the AI service execution loop when the annotated tool is the **last** tool called in a multi-tool batch response.

### Motivation

The existing `IMMEDIATE` behavior halts only when **all** tools in a batch are immediate. This is documented and intentional. However, there is a valid use case where an agent has a designated \"end of execution\" tool (e.g. `endExecutionAndGetFinalResult`) and the LLM intentionally places it **last** in a multi-tool response — meaning it completed all required side-effect actions first and then explicitly signaled completion.

With the current semantics, this results in one unnecessary extra LLM round-trip: the loop sends all results back to the model, the model re-calls the halt tool, and only then halts. Each spurious round trip costs one full LLM inference call (≥1 s, potentially much more if images/video are attached).

### Proposed Semantics

Add `ReturnBehavior.IMMEDIATE_IF_LAST` with the following behavior:

| Batch order | Result |
|---|---|
| `[regularTool, IMMEDIATE_IF_LAST_tool]` | **Halt** — last tool is immediate, model ordering is intentional ✓ |
| `[IMMEDIATE_IF_LAST_tool, regularTool]` | **Continue** — last tool is regular, treat as model error, preserve existing behavior ✓ |
| `[IMMEDIATE_IF_LAST_tool]` | **Halt** — single-tool batch, same as existing `IMMEDIATE` ✓ |

### Why not the other options

- `IMMEDIATE_OVERRIDE` — halts regardless of position; if the model accidentally places the halt tool first in a batch, execution halts before other tools in the batch run, producing a partial/incorrect result.
- `TO_LLM_WEAK` — solves the inverse problem (weakening `TO_LLM` when a sibling is `IMMEDIATE`), which is a different semantic concern.

### Proposed Implementation

1. Add `IMMEDIATE_IF_LAST` to the `ReturnBehavior` enum.
2. Populate a separate `immediateIfLastReturnTools` set in `ToolServiceContext` (mirroring the existing `immediateReturnTools` set).
3. In `executeInferenceAndToolsLoop`, after the existing tool-iteration loop, check if the **last** tool request name is present in `immediateIfLastReturnTools` — if yes, force `immediateToolReturn = true`.

This is purely additive — no existing behavior changes.

### Related

Originated from issue #4966 where this was misreported as a bug. The current `IMMEDIATE` behavior is working as documented; this feature request captures the new enum value needed to support the described use case.

suryateja-g13:
Fixes #4974

## What changed

Adds `ReturnBehavior.IMMEDIATE_IF_LAST`, a new enum value that halts the AI service execution loop when the annotated tool is the **last** tool called in a multi-tool batch and completed without error.

### Semantics

| Batch order | Behaviour |
|---|---|
| `[regularTool, IMMEDIATE_IF_LAST_tool]` | **Halt** — last tool signals completion ✓ |
| `[IMMEDIATE_IF_LAST_tool, regularTool]` | **Continue** — last is regular, treat as non-final ✓ |
| `[IMMEDIATE_IF_LAST_tool]` | **Halt** — single tool, same as existing `IMMEDIATE` ✓ |
| Last tool throws | **Continue** — error suppresses halt, same as `IMMEDIATE` ✓ |

### Files changed

- `ReturnBehavior.java` — new `IMMEDIATE_IF_LAST` enum value with Javadoc
- `AiServiceTool.java` — `immediateIfLastReturn` flag + builder method
- `ToolServiceContext.java` — `immediateIfLastReturnTools` set (field, accessor, builder, equals/hashCode/toString)
- `ToolProviderResult.java` — `immediateIfLastReturnToolNames` (field, accessor, builder method, `add(…, ReturnBehavior)` routing)
- `ToolService.java` — populates set from annotations and `ToolProvider`; post-loop check in `executeInferenceAndToolsLoop`; `isImmediateIfLastTool()` helper; `refreshDynamicProviders` propagation
- `AiServiceStreamingResponseHandler.java` — tracks `lastToolName`/`lastToolErrored` across both concurrent and sequential tool branches; post-loop check mirrors sync path
- `ImmediateIfLastReturnTest.java` — 6 unit tests covering all four semantic cases for both sync and streaming paths

## Before submitting this PR I confirmed the following

- [x] I have read the [contribution guidelines](https://github.com/langchain4j/langchain4j/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of my code
- [x] I have added tests that prove my fix/feature works (6 unit tests)
- [x] New and existing unit tests pass locally with my changes
- [x] I have updated the relevant documentation (Javadoc on enum value)
- [x] I confirm that the changes are backward compatible (purely additive, no existing behaviour changed)

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnBehavior.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReturnBehavior.java
@@ -20,5 +20,18 @@ public enum ReturnBehavior {
      * while a {@code RuntimeException} will be thrown attempting to use a tool with immediate return with an
      * AI service having a different return type.
      */
-    IMMEDIATE;
+    IMMEDIATE,
+
+    /**
+     * Returns immediately to the caller when the annotated tool is the <b>last</b> tool called
+     * in a multi-tool batch response, and it completed without error. When placed last, the model
+     * has intentionally signalled completion after all side-effect tools have run.
+     * <p>
+     * If the tool appears earlier in the batch (not last), execution continues normally as if
+     * {@link #TO_LLM} were used.
+     * <p>
+     * Like {@link #IMMEDIATE}, this is only allowed on AI services returning
+     * {@code dev.langchain4j.service.Result}.
+     */
+    IMMEDIATE_IF_LAST;
 }
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -5,8 +5,6 @@ import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 import static dev.langchain4j.service.AiServiceParamsUtil.chatRequestParameters;
 import static dev.langchain4j.service.tool.ToolService.refreshDynamicProviders;
 
-import dev.langchain4j.service.tool.search.ToolSearchService;
-
 import dev.langchain4j.Internal;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.data.message.AiMessage;
@@ -42,6 +40,7 @@ import dev.langchain4j.service.tool.ToolExecutionErrorHandler;
 import dev.langchain4j.service.tool.ToolExecutionResult;
 import dev.langchain4j.service.tool.ToolExecutor;
 import dev.langchain4j.service.tool.ToolServiceContext;
+import dev.langchain4j.service.tool.search.ToolSearchService;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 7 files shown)


dliubarskyi:
@suryateja-g13 thank you for your PR!
TBH I am not very happy with the current implementation of `AIServiceTool`, `ToolProviderResult`, etc.
These classes require refactoring becore `IMMEDIATE_IF_LAST` can be added properly without introducing more tech debt for the future.
I did this refactoring in https://github.com/langchain4j/langchain4j/pull/4998 before adding `IMMEDIATE_IF_LAST`.
I also added thorough testing of all combinations and refactored handling of immediate return logic in both sync and streaming AI Services. The documentation is also updated everywhere.
I hope you don't mind if I will close this PR in favor of https://github.com/langchain4j/langchain4j/pull/4998.

In any case, thank you for your PR!

PR #4956, opened 2026-04-17
Check that an untyped agent has all the required arguments before invoking it

mariofusco:
## Issue

Closes #4946

## General checklist

- [X] There are no breaking changes (API, behaviour)
- [X] I have added unit and/or integration tests for my change
- [X] The tests cover both positive and negative cases
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new maven module

- [ ] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #4976, opened 2026-04-21
fix: deserialize JSON strings to custom objects in agentic scope (#4897)

### LINKED ISSUE #4897: IllegalArgumentException: argument type mismatch when using SupervisorAgent with custom object parameters in Sub-Agents
When using SupervisorAgent to coordinate multiple sub-agents that have custom object parameters (e.g., ExerciseQuestion, ExerciseRequirement), the framework throws an IllegalArgumentException: argument type mismatch error during reflection-based method invocation.
(error cases like)
String generateExercise(@V("exerciseRequirement") ExerciseRequirement exerciseRequirement);

I wonder if there are any plans to support custom object parameters in sub-agents? Thank u very much for answering.

suryateja-g13:
## Summary

Fixes #4897 — `IllegalArgumentException: argument type mismatch` when a `SupervisorAgent` passes a custom domain object (serialized as a JSON string) into the agentic scope for a sub-agent whose method parameter expects that custom type.

**Root cause:** `AgentUtil.adaptValueToType()` handled only primitive wrapper types in its `String` switch. The `default` branch threw `IllegalArgumentException("Unsupported type: " + type)` instead of attempting deserialization.

**Fix:** Replace the `default` throw with `Json.fromJson(s, type)` so JSON-encoded objects are transparently deserialized to the expected parameter type using LangChain4j's existing Jackson-backed codec.

**Behavior matrix (unchanged cases still work as before):**

| Value type | Target type | Before | After |
|---|---|---|---|
| `String` | `String` / `Object` | pass-through | pass-through |
| `String` | `Integer`, `Long`, `Double`, `Float`, `Boolean` | coerced | coerced |
| `String` | Custom POJO / Record | `IllegalArgumentException` | JSON-deserialized |
| `String` | Custom POJO (invalid JSON) | `IllegalArgumentException` | `RuntimeException` from Jackson |
| Already correct type | any | returned as-is | returned as-is |

## Changes

- `AgentUtil.java` — one-line fix: `default -> Json.fromJson(s, type)` in `adaptValueToType()`
- `AgentUtilTest.java` — 6 unit tests covering string passthrough, JSON deserialization to record, direct object passthrough, String→Integer coercion, String→Boolean coercion, and invalid JSON error path

## Test plan

- [x] `AgentUtilTest` — 6 tests all pass
- [x] Spotless formatting applied
- [ ] Run full `langchain4j-agentic` test suite in CI


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentUtil.java
```diff
@@ -260,7 +261,7 @@ private static Object adaptValueToType(Object value, Class<?> type) {
                 case "double", "java.lang.Double" -> Double.parseDouble(s);
                 case "float", "java.lang.Float" -> Float.parseFloat(s);
                 case "boolean", "java.lang.Boolean" -> Boolean.parseBoolean(s);
-                default -> throw new IllegalArgumentException("Unsupported type: " + type);
+                default -> Json.fromJson(s, type);
```
mariofusco: This is a very reasonable and pragmatic improvement. The only thing that I would change here is catching the Exception eventually thrown by the `fromJson` method and rethrowing it as an `IllegalArgumentException` as it did before.

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/internal/AgentUtilTest.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Test
+    void should_throw_when_json_string_is_invalid_for_target_type() {
+        DefaultAgenticScope scope = DefaultAgenticScope.ephemeralAgenticScope();
+        scope.writeState("person", "not-valid-json");
+
+        assertThatThrownBy(() ->
+                        AgentUtil.agentInvocationArguments(scope, List.of(new AgentArgument(Person.class, "person"))))
+                .isInstanceOf(RuntimeException.class);
```
mariofusco: When you change the exception thrown into an `IllegalArgumentException`, also change this assertion accordingly.

mariofusco (changes requested):
Please change the exception thrown in case of unparsable JSON according to my comments. 

Also it would be nice to have an end-to-end integration test reproducing the original problem using the supervisor pattern, to be sure that this improvement actually fixes it.

suryateja-g13:
Thanks for the review @mariofusco!

**Exception wrapping (both inline comments):** Done — the `default` branch now wraps the `Json.fromJson` call in a `try/catch` and rethrows as `IllegalArgumentException` with the original exception as the cause, preserving the method's existing contract. The test assertion has been tightened from `RuntimeException` to `IllegalArgumentException` accordingly.

**End-to-end integration test:** I considered adding one, but the fix is a single-line change deep inside `AgentUtil.adaptValueToType()` — the exact method that `AgentUtilTest` exercises directly. A supervisor-level integration test would require either a live LLM or a substantial mock of the entire agentic orchestration stack, adding significant complexity for a code path that is already fully covered by the unit tests. That said, if you feel the integration test is important for the PR, I'm happy to add it — just let me know.

suryateja-g13:
The `integration_test (21)` CI failure is unrelated to this PR.

The failure is in the `langchain4j-anthropic` module (which we don't touch) and is caused by a `revapi` API compatibility check detecting breaking changes in `AnthropicChatModelName`:

- `CLAUDE_3_HAIKU_20240307` enum constant was removed
- Multiple enum constants had their ordinal positions shifted (e.g. `CLAUDE_OPUS_4_6` moved from position 0 → 1)

This appears to be a pre-existing breakage on `main` from a recent unrelated commit. Our changes are scoped entirely to `langchain4j-agentic`.

mariofusco:
> **End-to-end integration test:** I considered adding one, but the fix is a single-line change deep inside `AgentUtil.adaptValueToType()` — the exact method that `AgentUtilTest` exercises directly. A supervisor-level integration test would require either a live LLM or a substantial mock of the entire agentic orchestration stack, adding significant complexity for a code path that is already fully covered by the unit tests. That said, if you feel the integration test is important for the PR, I'm happy to add it — just let me know.

Up to you, as I said I believe that this is a reasonable improvement, so I can also merge this PR as it is, but if I do so I cannot close the related bug report, until I haven't reproduced the original problem and demonstrated that this fix also solve it. In this case I could create the integration test on my own. Please let me know what you prefer.

mariofusco:
> This appears to be a pre-existing breakage on `main` from a recent unrelated commit. Our changes are scoped entirely to `langchain4j-agentic`.

Don't mind about the broken CI, we're working on it and as you said it is unrelated with this change.

Just let me know if you want to try to create that integration test, or I should merge this PR as it is.

suryateja-g13:
Thanks @mariofusco! I noticed you've already added the `pojo_argument_test()` integration test in commit `0ce1d55` (PR #4956), which reproduces the exact scenario from issue #4897 — a `LoanApplication` POJO written to scope by one sub-agent and read as a typed parameter by another. That test exercises the precise code path our fix addresses.

Since the integration test is already in `main`, please go ahead and merge this PR as-is. No duplication needed on our side.

mariofusco (approved):

PR #5000, opened 2026-04-22
Fix #4918: Warn when Apache HttpClient 5 is used for streaming

### LINKED ISSUE #4918: [BUG] SpringRestClient silently uses Apache HttpClient 5, causing streaming tokens to arrive in bursts instead of real-time
**Describe the bug**                                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                      
  When using SpringRestClient for streaming, onPartialResponse callbacks arrive in bursts (15-25 second gaps) instead of token-by-token. The response text arrives all at once after a long delay, instead of streaming as tokens are generated by the API.           
                                         
  **Root cause**                                                                                                                                                                                                                                                          
                                                                                     
  SpringRestClient's constructor always calls ClientHttpRequestFactoryBuilder.detect().build(settings) and overwrites any requestFactory set on the provided RestClient.Builder. When Apache HttpClient 5 (httpclient5) is on the classpath — often transitively via ClickHouse JDBC, Google Cloud libraries, etc. — detect() picks it up automatically.
                                                                                                                                                                                                                                                                      
  Replacing Apache HttpClient 5 with the JDK built-in java.net.http.HttpClient completely resolves the issue — tokens stream in real-time with no gaps.                                                                                                               
                                         
  SpringRestClient also ignores a custom RestClient.Builder with a pre-configured requestFactory — it always overwrites it with the auto-detected factory, making it impossible to force a different HTTP client through the builder API.                             
                                                                                     
  **Expected behavior**                                                                                                                                                                                                                                                   
   
  onPartialResponse should be called as each token arrives from the API, with no artificial delays between tokens.                                                                                                                                                    
                                                                                     
  **Workaround**                             

  Implement a custom HttpClient using java.net.http.HttpClient (JDK built-in), bypassing SpringRestClient entirely.                                                                                                                                                   
   
  **Suggested fixes**                                                                                                                                                                                                                                                     
                                                                                     
  1. SpringRestClient should respect the requestFactory already set on the provided RestClient.Builder instead of always overwriting it with ClientHttpRequestFactoryBuilder.detect()                                                                                 
  2. Document that having Apache HttpClient 5 on the classpath can silently degrade streaming performance                                                                                                                                                             
                                                                                     
  **Environment**                                                                                                                                                                                                                                                         
  - LangChain4j version: 1.13                                                        
  - Java version: 25 (Virtual Threads)   
  - LLM: Anthropic Claude (likely affects all providers)
  - HTTP client on classpath: Apache HttpClient 5.5 (transitive dependency)
  - Spring Boot: 3.5.x                                                                                                                                                                                                                                                
   
  **Related issues**                                                                                                                                                                                                                                                      
  - #3298 (same symptoms reported for OpenAI)

anuragg-saxenaa:
## Summary

Fixes issue #4918: SpringRestClient silently uses Apache HttpClient 5, causing streaming tokens to arrive in bursts instead of real-time.

## Changes

- **SpringRestClientHttpClient.java**: Added warning log when Apache HttpClient 5 (httpclient5) is auto-detected, explaining the potential streaming performance issues
- Added `detectHttpClientFactory()` method that detects which HTTP client is being used
- Added `isApacheHttpClient5()` helper to check if the factory is Hc5-based
- **SpringRestClientHttpClientIT.java**: Added test to verify factory detection

## Why this helps

When Apache HttpClient 5 is on the classpath (often transitively via ClickHouse JDBC, Google Cloud libraries, etc.), Spring's `ClientHttpRequestFactories.get()` automatically selects it. This causes streaming responses to arrive in bursts rather than token-by-token in real-time.

The warning helps users diagnose this issue faster by logging a clear message when Hc5 is detected.

## Testing

- Unit test added to verify factory detection works

See: https://github.com/langchain4j/langchain4j/issues/4918


anuragg-saxenaa:
Closing — branch base was stale, causing unrelated upstream files to appear in the diff. Will resubmit as a clean single-file PR if the fix is still needed.

PR #5003, opened 2026-04-23
MCP: expose session id on StreamableHttpMcpTransport (#4757)

### LINKED ISSUE #4757: [FEATURE] MCP: expose session id
I have a conversational workflow (with a TTL) as part of a larger orchestrated workflow. 
The workflow requires me to contact 1..n MCP servers and invoke tools on them.

I need to maintain state across each of the conversations for each user talking to each server.
MCP - unlike (say) A2A doesn't maintain state. 
Instead I have to fall back to using the transports session id for session persistence.

1. I'd like an easy way to get to the session id from the McpClient/McpTransport somehow. 
I'm struggling to get this from the internals of the transport - I've poked around and don't think it'a an ID10T error, so would like it on a plate so I can use it as my session token.

2. I don't know if this one is possible without going back into the MCP protocol deeper than I'd like, so I'm going to throw it out here. 
We run inside kubernetes - but then who doesn't. 
I'd like to not have to have sticky sessions for this piece. I'd love a way to inject the context back into a client so I can create them on demand in any pod rather than have to save them for the duration of the conversation.

I don't really have an alternative other than to associate a generated session id with each client and save them for the duration of the transaction in a map in memory. That's not ideal at the concurrency we're looking at in production.

**Additional context**

zxuhan:
## Summary
- Add `mcpSessionId()` getter and `setMcpSessionId(String)` setter to `StreamableHttpMcpTransport`, allowing callers to read the server-assigned session ID and inject one obtained elsewhere (e.g. from another pod) for stateless deployments without sticky sessions.

Closes #4757.

## Test plan
- [x] New unit tests in `StreamableHttpMcpTransportTest` cover the default-null state and getter/setter round-trip
- [x] `mvn -pl langchain4j-mcp test -Dtest=StreamableHttpMcpTransportTest` passes (7/7)

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
@@ -174,7 +174,7 @@ public class StreamableHttpMcpTransport implements McpTransport {
      * captured from the {@code Mcp-Session-Id} response header during initialization
      * and reused on subsequent requests via the same header.
      */
-    public String mcpSessionId() {
+    public String getMcpSessionId() {
         return mcpSessionId.get();
     }
 
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
@@ -168,6 +168,26 @@ public class StreamableHttpMcpTransport implements McpTransport {
         this.onFailureCallback = actionOnFailure;
     }
 
+    /**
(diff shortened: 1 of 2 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/http/StreamableHttpMcpTransport.java
```diff
@@ -168,6 +168,26 @@ public void onFailure(Runnable actionOnFailure) {
         this.onFailureCallback = actionOnFailure;
     }
 
+    /**
+     * Returns the MCP session ID assigned by the server, or {@code null} if no session
+     * has been established yet (or the server does not use sessions). The session ID is
+     * captured from the {@code Mcp-Session-Id} response header during initialization
+     * and reused on subsequent requests via the same header.
+     */
+    public String mcpSessionId() {
```
jmartisk: Could we name this `getMcpSessionId` to follow the regular naming for getters?

jmartisk (approved):

PR #5004, opened 2026-04-23
PgVector: parenthesize isNotIn/isNotEqualTo to fix AND/OR precedence (#2513)

### LINKED ISSUE #2513: [BUG] Metadata Filter .isNotIn  leads to unintended .or clause that matches ALL content
We are heavily using the Metadata Key filter for our embedding search.
I just found that if you add a .isNotIn filter, with a list of Strings like so 

`.and(metadataKey("id").isNotIn(["A", "b", "C"]))`, the parsed query result inside of the `PgVectorEmbeddingStore class ` is:

`WHERE metadata->>'id' is null **or** metadata->>'id' not in ('A',`B', 'C') 

Which means, if you add this .isNotIn query at the end of any filter chain, chained with AND, it will lead to all content being matched which does not contain the mentioned id's in the list.

**To Reproduce**
Use any default implementation of langchain and use a .isNotIn filter with a random list of fields that should not be matched.

**Expected behavior**

- LangChain4j version: alp
- LLM(s) used:  1.0.0-alpha1
- Java version:  21
- Spring Boot version (if applicable):  3.2

**Additional context**

zxuhan:
## Summary
- Wrap the `is null or ...` clauses produced by `mapNotIn` and `mapNotEqual` in parentheses. Without them, `someFilter.and(metadataKey("id").isNotIn(...))` rendered as `something AND id is null OR id not in (...)`, which SQL parses as `(something AND id is null) OR (id not in (...))` — silently matching rows that should be filtered out.
- Matches the parenthesization already used in `MariaDbFilterMapper`.

Closes #2513.

## Test plan
- [x] New `PgVectorFilterMapperTest` asserts the parens for both `JSONFilterMapper` and `ColumnFilterMapper`, and that AND-combined queries keep correct precedence
- [x] `mvn -pl langchain4j-pgvector test` passes (10/10 unit tests)

### DIFF
--- a/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorFilterMapper.java
+++ b/langchain4j-pgvector/src/main/java/dev/langchain4j/store/embedding/pgvector/PgVectorFilterMapper.java
@@ -73,7 +73,7 @@ abstract class PgVectorFilterMapper {
     private String mapNotEqual(IsNotEqualTo isNotEqualTo) {
         String key =
                 formatKey(isNotEqualTo.key(), isNotEqualTo.comparisonValue().getClass());
-        return format("%s is null or %s != %s", key, key, formatValue(isNotEqualTo.comparisonValue()));
+        return format("(%s is null or %s != %s)", key, key, formatValue(isNotEqualTo.comparisonValue()));
     }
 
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@zxuhan thank you!

PR #5006, opened 2026-04-23
feat(azure-open-ai): expose stream and promptCacheKey parameters

### LINKED ISSUE #5005: [FEATURE] Hope Azure OpenAI SDK supports some important parameters
Currently, the Azure OpenAI SDK does not expose support for the **stream** and **prompt_cache_key** parameters as documented in the [Azure OpenAI Service API reference](https://learn.microsoft.com/en-us/azure/foundry/openai/latest#chat).

This makes it impossible to use key functionality when calling the chat completions API:
Without the stream parameter, we cannot enable streaming responses, which is essential for real-time chat and incremental UI updates.
Without the prompt_cache_key parameter, we cannot leverage prompt caching to reduce latency and cost for repeated or long system prompts.

I would like the Azure OpenAI SDK to:
Add support for the **stream** parameter in chat completions requests, allowing users to enable/disable streaming responses.
Add support for the **prompt_cache_key** parameter, so users can pass a cache key to enable prompt caching.
Both parameters should be exposed as optional, type-safe arguments in the SDK methods, matching the behavior described in the official Azure OpenAI documentation.

anuragg-saxenaa:
## Summary

This PR adds `stream(Boolean)` and `promptCacheKey(String)` parameters to both `AzureOpenAiChatModel` and `AzureOpenAiStreamingChatModel` builders, addressing issue #5005.

### Changes

**AzureOpenAiChatModel:**
- Added `stream(Boolean)` builder method - sets streaming via `ChatCompletionsOptionsAccessHelper.setStreamOptions()`
- Added `promptCacheKey(String)` builder method - sets `prompt_cache_key` in options metadata

**AzureOpenAiStreamingChatModel:**
- Added `stream(Boolean)` builder method - conditionally enables streaming
- Added `promptCacheKey(String)` builder method - sets prompt cache key in metadata

### Tests Added

- `should_support_stream_and_promptCacheKey_builder_params` in `AzureOpenAiChatModelIT`
- `should_support_promptCacheKey_without_stream` in `AzureOpenAiChatModelIT`
- `should_support_promptCacheKey_in_streaming` in `AzureOpenAiStreamingChatModelIT`

### Why This Matters

The Azure OpenAI API supports:
1. **stream** parameter: enables streaming responses for real-time UI updates
2. **prompt_cache_key** in metadata: enables prompt caching for reduced latency/cost

Both were documented in the Azure API reference but not exposed in langchain4j.

Fixes #5005

### DIFF
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
@@ -4,6 +4,7 @@ import static com.azure.ai.openai.models.CompletionsFinishReason.CONTENT_FILTERE
 import static dev.langchain4j.internal.Utils.copy;
 import static dev.langchain4j.internal.Utils.copyIfNotNull;
 import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.Utils.isNotNullOrBlank;
 import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 import static dev.langchain4j.model.ModelProvider.AZURE_OPEN_AI;
 import static dev.langchain4j.model.azure.InternalAzureOpenAiHelper.aiMessageFrom;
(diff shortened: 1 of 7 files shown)


dliubarskyi:
@anuragg-saxenaa thank you, but Azure OpenAI module is deprecated, we do not add new features there. Moreover, streaming is already supported via `AzureOpenAiStreamingChatModel`.

We recommend using Official module instead https://docs.langchain4j.dev/integrations/language-models/open-ai-official

PR #5007, opened 2026-04-23
feat(azure-open-ai): Add stream and prompt_cache_key parameters

### LINKED ISSUE #5005: [FEATURE] Hope Azure OpenAI SDK supports some important parameters

anuragg-saxenaa:
Fixes #5005. Adds stream boolean and promptCacheKey string parameters to AzureOpenAiChatModel matching Azure OpenAI Service API reference.

### DIFF
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/AzureOpenAiChatModel.java
@@ -25,6 +25,9 @@ import com.azure.ai.openai.models.ChatChoice;
 import com.azure.ai.openai.models.ChatCompletions;
 import com.azure.ai.openai.models.ChatCompletionsOptions;
 import com.azure.ai.openai.models.ReasoningEffortValue;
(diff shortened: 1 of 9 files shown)


anuragg-saxenaa:
Closing — branch included unrelated guardrail/MCP files due to stale base. The Azure stream/promptCacheKey feature is covered cleanly by #5006.

Issue #4873, opened 2026-04-08
[QUESTION] Why mcp messages are logged as error using stdio transport


cmoulliard:
Question

This is a question and potentially an issue to be fixed. Why are the messages of a MCP process launched logged as ERROR by the `ProcessStderrHandler`  using stdio transport?

```
2026-04-08 07:45:28,310 DEBUG [dev.lan.mcp.cli.tra.std.ProcessStderrHandler] (executor-thread-2) [ERROR] WARNING: sun.misc.Unsafe::objectFieldOffset will be removed in a future release
2026-04-08 07:45:28,502 DEBUG [dev.lan.mcp.cli.tra.std.ProcessStderrHandler] (executor-thread-2) [ERROR] __  ____  __  _____   ___  __ ____  ______ 
2026-04-08 07:45:28,503 DEBUG [dev.lan.mcp.cli.tra.std.ProcessStderrHandler] (executor-thread-2) [ERROR]  --/ __ \/ / / / _ | / _ \/ //_/ / / / __/ 
2026-04-08 07:45:28,503 DEBUG [dev.lan.mcp.cli.tra.std.ProcessStderrHandler] (executor-thread-2) [ERROR]  -/ /_/ / /_/ / __ |/ , _/ ,< / /_/ /\ \   
2026-04-08 07:45:28,503 DEBUG [dev.lan.mcp.cli.tra.std.ProcessStderrHandler] (executor-thread-2) [ERROR] --\___\_\____/_/ |_/_/|_/_/|_|\____/___/   
2026-04-08 07:45:28,503 DEBUG [dev.lan.mcp.cli.tra.std.ProcessStderrHandler] (executor-thread-2) [ERROR] 2026-04-08 07:45:28,500 INFO  [io.qua.mcp.ser.fil.MCPServerFS] (main) Starting file server with paths: [/Users/cmoullia/code/_temp/filesystem-assistant/playground]
2026-04-08 07:45:28,505 DEBUG [dev.lan.mcp.cli.tra.std.ProcessStderrHandler] (executor-thread-2) [ERROR] 2026-04-08 07:45:28,505 INFO  [io.quarkus] (main) mcp-server-filesystem 1.0.0.CR4 on JVM (powered by Quarkus 3.20.1) started in 0.419s. Listening on: 
[...]

PR #5010, opened 2026-04-24
Fix: change MCP stderr log prefix from [ERROR] to [MCP stderr] (#4873)

anuragg-saxenaa:
## Summary
This PR fixes issue #4873 by changing the  log prefix from  to .

## Changes
1. Changed  to  in 
2. Added a one-time INFO log when the stderr handler starts: 
3. Added regression test  that:
   - Verifies stderr output is logged with  prefix
   - Ensures no ERROR-level logs appear for normal stderr output
   - Confirms the INFO log message is emitted at startup

## Testing
A new unit test verifies the fix by:
- Running a mock process that outputs to stderr
- Capturing log events from 
- Asserting the correct prefix is used and no ERROR-level logs are generated for normal output


dliubarskyi:
this PR has a lot of unrelated changes. please fix and reopen

PR #5012, opened 2026-04-24
Support multi-value metadata in embedding stores (#1905)

### LINKED ISSUE #1905: [FEATURE] Support multi-value document metadata

anuragg-saxenaa:
## Summary
Adds support for Collection values in the Metadata class to enable multi-value metadata in embedding stores.

## Changes
- **Metadata.java**: Added collectionMetadata field and putCollection/getCollection methods
- **IsEqualTo.java**: Updated to check collection values when comparing metadata
- **Tests**: Added comprehensive tests for collection metadata operations

## Usage Example
Metadata metadata = new Metadata();
metadata.putCollection("tags", List.of("a", "b", "c"));

// IsEqualTo filter checks if value exists in the collection
Filter filter = MetadataFilterBuilder.metadataKey("tags").isEqualTo("b");

Fixes #1905

### DIFF
--- a/langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/goap/GoalOrientedSearchGraph.java
+++ b/langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/goap/GoalOrientedSearchGraph.java
@@ -40,7 +40,11 @@ public class GoalOrientedSearchGraph {
     }
 
     public List<AgentInstance> search(Collection<String> preconditions, String goal) {
-        List<Node> nodesPath = DependencyGraphSearch.search(nodes.get(goal), preconditions.stream().map(nodes::get).toList());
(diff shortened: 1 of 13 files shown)


dliubarskyi:
this PR has many unrelated changes. please fix and reopen

PR #5015, opened 2026-04-24
Add built-in tools support to unofficial OpenAI Responses models

leochame:
## Issue
Relates to #4838

## Change
This PR adds request-side support for OpenAI built-in tools in the unofficial Responses models:
- `OpenAiResponsesChatModel`
- `OpenAiResponsesStreamingChatModel`

The unofficial API shape stays minimal and provider-specific by introducing `serverTools` as `List<Map<String, Object>>` on `OpenAiResponsesChatRequestParameters` and the two model builders.

`toolSpecifications` continues to represent LangChain4j function tools, while `serverTools` carries raw OpenAI-shaped built-in tools. Both are merged into the outgoing OpenAI `tools` array, with function tools serialized first and built-in tools appended after them.

This keeps the unofficial implementation close to the OpenAI wire format without introducing a new `OpenAiServerTool` abstraction in this patch.

Added coverage includes:
- storing and overriding `serverTools` in request parameters
- propagating `serverTools` through unofficial model builders
- payload construction for built-in tools only
- payload construction for function tools only
- payload construction for mixed function tools and built-in tools
- `toolChoice` serialization when mixed tools are present

Note: this change is intentionally limited to request support. It does not add built-in tool result extraction yet.

## General checklist


leochame:
@dliubarskyi Responding to the review suggestion:

> In case of an unofficial implementation, it will require adding a new type (e.g., OpenAiServerTool, similar to AnthropicServerTool), because we do not use any SDK in unofficial module.

My current view is that this may not be necessary for the unofficial implementation.

The main reason is that Anthropic and OpenAI expose built-in tools with different payload shapes.

For Anthropic, the outer shape is comparatively stable. A built-in tool entry has fixed top-level fields such as `type` and `name`, while the tool-specific payload is carried in a nested structure. For example, Anthropic's web search tool uses a shape like `type` / `name` / `input`. This gives Anthropic a stable outer envelope plus a nested extension point, which makes a wrapper such as `AnthropicServerTool` feel natural.
Docs:
- https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/web-search-tool
- https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/code-execution-tool

OpenAI Responses built-in tools are different. They are discriminator-style objects: `type` selects the tool kind, and the tool-specific fields live directly at the same level as `type`. For example, OpenAI tools use shapes such as `type=web_search` with sibling fields like `filters` / `user_location`, or `type=file_search` with sibling fields like `vector_store_ids` / `filters`. There is no equivalent stable nested field like Anthropic's `input`.
Docs:
- https://platform.openai.com/docs/guides/tools-web-search
- https://platform.openai.com/docs/guides/tools-file-search

Because of that, an `OpenAiServerTool` type in the unofficial module may end up being only a thin wrapper around a `Map`.

So for the unofficial implementation, I think there is a reasonable case for keeping the request-side API closer to the raw OpenAI shape, for example with `Map` / `List`

dliubarskyi (commented):
@leochame thanks! Please add a documentation on how to configure server tools under https://docs.langchain4j.dev/integrations/language-models/open-ai#openai-responses-api

leochame:
> @leochame thanks! Please add a documentation on how to configure server tools under https://docs.langchain4j.dev/integrations/language-models/open-ai#openai-responses-api

@dliubarskyi that‘s ok， have added

dliubarskyi (approved):
@leochame thank you!

PR #5018, opened 2026-04-25
feat(openai): Support reasoning field alongside reasoning_content (#4796)

### LINKED ISSUE #4796: [FEATURE] Support reasoning as well as reasoning_content fields in the Open AI chat model

anuragg-saxenaa:
## Summary

Fixes #4796 - VLLM deprecated the reasoning_content field in favor of reasoning for returning LLM thinking. This change updates the Delta class to support both fields, preferring reasoning when both are present.

### Delta.java
- Add reasoning field with @JsonProperty("reasoning") annotation to accept VLLM new field name
- Add reasoning method that returns reasoning if present, otherwise falls back to reasoning_content
- Update equals, hashCode, and toString to include the reasoning field
- Update Builder class with reasoning(String) method

### OpenAiStreamingResponseBuilder.java
- Changed delta.reasoningContent() call to delta.reasoning() to use the new unified method

### StreamingResponseBuilderReasoningFieldTest.java (new)
- Regression test covering extraction from both new reasoning and legacy reasoning_content fields
- Verifies preference behavior when both fields are present

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiChatModel.java
@@ -3,6 +3,7 @@ package dev.langchain4j.model.openai;
 import static dev.langchain4j.internal.RetryUtils.withRetryMappingExceptions;
 import static dev.langchain4j.internal.Utils.copy;
 import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.Utils.isNullOrEmpty;
 import static dev.langchain4j.model.ModelProvider.OPEN_AI;
 import static dev.langchain4j.model.chat.Capability.RESPONSE_FORMAT_JSON_SCHEMA;
 import static dev.langchain4j.model.openai.internal.OpenAiUtils.DEFAULT_OPENAI_URL;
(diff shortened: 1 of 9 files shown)


redinside-dev (commented):
Hi @anuragg-saxenaa - great feature! CI failures in spotless. Main issues: OpenAiResponsesClient.java has long lines split differently than spotless expects. Run: ./mvnw spotless:apply -pl langchain4j-open-ai. Also: this PR and #5029 both add empty-choices guard to OpenAiChatModel.java - please ensure only one carries this change to avoid merge conflicts.

anuragg-saxenaa:
👋 Hi maintainers — this adds reasoning_content support to match reasoning field for VLLM-compatible providers. Ready for review. Happy to iterate!

squallliu:
A simpler method
```
@JsonAlias("reasoning")
public Builder reasoningContent(String reasoningContent) {
  this.reasoningContent = reasoningContent;
  return this;
}
```

anuragg-saxenaa:
Closing to address the compile failures and apply the simpler approach suggested in the review.

PR #3271, opened 2025-06-28
Support for the OpenAI gpt-image-1 image generation model

binocla:
## Issue
Closes #2921

## Change
Added `gpt-image-1` support for OpenAI Image API

## General checklist
- [X] There are no breaking changes
- [X] I have added unit and/or integration tests for my change
- [ ] The tests cover both positive and negative cases
- [ ] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [X] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new maven module
- [ ] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

## Checklist for adding new embedding store integration
- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

## Checklist for changing existing embedding store integration
- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #5039, opened 2026-04-27
feat(openai): add /v1/images/edits and gpt-image-* multi-image support

### LINKED ISSUE #2921: [FEATURE] Support for the OpenAI `gpt-image-1` image generation model
**Feature Request**

Please add official support for the new `gpt-image-1` model in the `dev.langchain4j:langchain4j-open-ai` library.

I tested image creation using this model with version `1.0.0-beta3`, and while the request itself seems to work, the response does **not** include the detailed token usage fields — specifically, `input_tokens_details`, which differentiate between text and image tokens.

According to the [OpenAI Images API reference](https://platform.openai.com/docs/api-reference/images/create), and the [pricing documentation](https://platform.openai.com/docs/models/gpt-image-1), these fields are now important, as image and text tokens are priced differently.

There are other fields like `moderation` and `background` that were just added for this new model.

**Current behavior**:  
No usage information is returned + `input_tokens_details` is missing.

**Expected behavior**:  
With full support for the `gpt-image-1` model, responses should include detailed token usage data, as returned by the API.

---

- LangChain4j version: `1.0.0-beta3`
- LLM(s) used: `OpenAI`
- Java version: `21`

---

**P.S.** I initially opened this as a bug report thinking the token details were missing from the OpenAI response. However, I’ve since learned that this is expected behavior unless the library explicitly supports the new `gpt-image-1` model. Everything else appears to work correctly, so this is more of a feature request to fully integrate support for this new model — especially as its usage details are now relevant for billing purposes.

unsignedint:
OpenAI have a new gpt-image-2 model which we would like to use in our app. We discovered some gaps with langchain4j that this PR aims to address. Generally, it will land comprehensive support for image generation and image edit, and with multi-image support.

Disclosure: AI generated with human-in-the-loop.

## Issue

Closes #2921 — adds OpenAI image-edit support and surfaces the gpt-image token-usage breakdown that the original issue called out.

For prior context: #3271 attempted a partial gpt-image-1 generation fix but went stale with unaddressed reviewer feedback. This PR reimplements the relevant bits cleanly, addresses #3271's review feedback (missing `@JsonDeserialize` annotations, missing native-image $Builder entries, ungrouped reflect-config), and goes further with edit support.

## Change

**Public API**

- `ImageModel` gains a canonical `edit(List<Image>, Image, String, int)`. The five existing edit overloads remain for ergonomics but are now **delegating defaults** routing to the canonical with `mask=null`/`n=1` as appropriate — implementations only override the canonical.
- `OpenAiImageModelName`: `GPT_IMAGE_1`, `GPT_IMAGE_2`.
- `OpenAiImageModel.OpenAiImageModelBuilder`: `background`, `moderation`, `outputCompression`, `outputFormat` (gen + edit), `inputFidelity` (edit-only; silently dropped for `gpt-image-2`, which rejects it).
- `OpenAiImageTokenUsage extends TokenUsage` — exposes the per-direction `text_tokens`/`image_tokens` breakdown via `Response<Image>.tokenUsage()`. dall-e responses still produce `tokenUsage() == null`.

**OpenAI internals**

- `EditImagesRequest` + `EditImageFile` carry edit parameters into a multipart request. URL-only `Image` inputs are rejected with a clear message (URL fetching is intentionally caller-side).
- `DefaultOpenAiClient.imagesEdit(...)` posts `multipart/form-data` to `/v1/images/edits`. Single-image uses field name `image`; multi-image uses repeated **`image[]`** parts (verified live — repeated `image` is rejected with `duplicate_parameter`).
- `response_format` omitted from the body for any `gpt-image-*` model (those reject the field). One central helper drives both the generation and edit paths.
- New `Usage` and `InputTokenDetails` JSON DTOs include the `@JsonDeserialize(builder = ...)` annotation that #3271 was missing. Native-image reflect-config covers the classes and their `$Builder` inner types.

**http-client multipart change (the only `@Experimental`-API breaking bit)**

`HttpRequest.formDataFiles()` returns `Map<String, List<FormDataFile>>` instead of `Map<String, FormDataFile>` so one field name can carry multiple parts (required for `image[]`). Marked `@Experimental` since 1.10.0, so the signature change is permitted. Existing single-file callers (audio transcription) are unchanged in semantics. JDK, OkHttp, and Apache implementations updated to iterate the lists.

## Verification

**Unit tests** (147, all green) — `OpenAiImageEditTest` captures the constructed `HttpRequest` via a fake `HttpClient` and asserts wire-format expectations across single/multi-image, mask, n, `response_format` omission, `input_fidelity` drop for gpt-image-2, mimeType handling, URL-only rejection, token-usage shape, and `OpenAiImageTokenUsage.add()`.

**Live integration tests** (`OpenAiImageModelIT`, kept `@Disabled`) — run against the live OpenAI API during authoring, all returned HTTP 200 with valid base64 content. Source images generated in-memory via `java.awt`. Coverage:

- `dall_e_2_edit_works` — confirms `tokenUsage()` is null
- `gpt_image_2_edit_single_image_works` — also exercises `inputFidelity("high")`, confirming the silent-drop fix (without it the API returns 400)
- `gpt_image_2_edit_with_mask_works`, `gpt_image_2_edit_multi_image_works` — confirm mask part and `image[]` array-form end-to-end
- `gpt_image_2_edit_with_mask_and_n_returns_multiple_images` — canonical method with `mask + n=2` returns a 2-element `Response<List<Image>>` with `OpenAiImageTokenUsage`

**Downstream testing** in our application with this PR + updated quarkus-langchain4j.

## Out of scope / follow-ups

- **Streaming / partial images** (`stream=true`, `partial_images`): the `ImageModel` interface is request-response only today.

## Design notes

- **`image[]` not `image` for multi-image** — OpenAI rejects repeated `image` parts with `duplicate_parameter`. Single-image stays `image` (dall-e-2 requires that name).
- **Subclass `TokenUsage`** for image usage — mirrors `OpenAiTokenUsage` for chat; keeps `instanceof`-castable type safety while base `TokenUsage` accessors continue to work for callers who don't need the split.
- **`Collections.singletonList` not `List.of`** in the delegating defaults — preserves the prior "no validation, route through" semantics for null images that some tests/edge cases relied on.
- **More edit overloads than generate** — edit has three independent axes (mask, n, input cardinality); generate has one (n). The canonical method keeps per-implementation cost flat.

## General checklist

- [ ] There are no breaking changes ← *the http-client multipart change is on `@Experimental`-marked API only, which the contract permits*
- [x] I have manually run all the unit tests in the modules I have touched, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules ← *verified for `langchain4j-core`, `langchain4j-http-client`, three http-client implementations, and `langchain4j-open-ai`; not run on the `langchain4j` aggregator separately*
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples)


Pahkho:
@dliubarskyi

Pahkho:
Does the plan lack this support? @dliubarskyi

PR #5048, opened 2026-04-27
Observability Feature Issue#4207

### LINKED ISSUE #4207: [FEATURE] - Enhance observability to include the entire interaction history
> Sorry for the delay in responding - i was out of the office all last week.
> 
> The challenge is that `AiServiceResponseReceivedEvent` doesn't capture an ENTIRE interaction (nor will it fire in the case of an error scenario). An interaction can include multiple round-trips to more than one LLM with different messages (think of an llm call that has tools, one of which is another llm call).
> 
> - call to llm (invocation 1)
> | - response
> | -- tool call
> | -- tool call
> | --- another llm call (invocation 2)
> | ---- tool call
> 
> etc
> 
> Maybe what we need is is a new event type that captures/records all of the events (in sequence) for an entire operation, in somewhat of a tree structure?  

 _Originally posted by @edeandrea in [#3986](https://github.com/langchain4j/langchain4j/issues/3986#issuecomment-3539358879)_

rayanahmed1215:
## Issue
Closes #4207

## Change
Introduces a new observability feature to capture the full interaction cycle of an AI service invocation.

-Added AiServiceInteractionEvent, which gathers all AiServiceEvent s for a invocation in firing order
-Added DefaultAiServiceInteractionEvent implementation using the existing builder pattern
-Added AiServiceInteractionListener for listening to connected interaction events
-Extended DefaultAiServiceListenerRegistrar to: track events per invocationId, store events throughout the invocation lifecycle and send a single AiServiceInteractionEvent when an invocation completes or errors

-Interaction events are sent only on AiServiceCompletedEvent or AiServiceErrorEvent
-Events are stored in order and returned as an immutable list

## General checklist


rayanahmed1215:
@edeandrea 
this is my first draft so far

edeandrea:
Thanks @rayanahmed1215 ! I'm out most of this week so please bear with me!

rayanahmed1215:
That's completely fine, I just wanted to get just drafting it finished so that I can focus on the finals studying.

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java
```diff
@@ -142,7 +142,7 @@ public Builder interfaceName(String interfaceName) {
         public Builder methodName(String methodName) {
             this.methodName = methodName;
             return this;
-        }
+        }
```
edeandrea, line 171: I'm not sure why the change to this class?

langchain4j-core/src/main/java/dev/langchain4j/invocation/DefaultInvocationContext.java
```diff
@@ -7,7 +7,7 @@
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
-
+
```
edeandrea, line 12: I'm not sure why the change to this class?

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/AiServiceEvent.java
```diff
@@ -17,7 +17,7 @@ public interface AiServiceEvent {
     // So that downstream frameworks/applications could define their own event types and still use the
     // registration/firing mechanisms provided here in LC4j
     <T extends AiServiceEvent> Class<T> eventClass();
-
+
```
edeandrea: I'm not sure why the change to this class?

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/AiServiceInteractionEvent.java
```diff
@@ -0,0 +1,68 @@
...
+import dev.langchain4j.invocation.InvocationContext;
+import dev.langchain4j.observability.event.DefaultAiServiceInteractionEvent;
+
+/**
+ * Aggregates all AiServiceEvent instances for a single invocationId,
+ * preserving the order in which they were fired.
+ */
+public interface AiServiceInteractionEvent extends AiServiceEvent {
+
+    List<AiServiceEvent> events();
```
edeandrea, line 19: All public methods need Javadocs

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/AiServiceInteractionEvent.java
```diff
@@ -0,0 +1,68 @@
...
+    default Class<AiServiceInteractionEvent> eventClass() {
+        return AiServiceInteractionEvent.class;
+    }
+
+    @Override
+    default AiServiceInteractionEventBuilder toBuilder() {
+        return new AiServiceInteractionEventBuilder(this);
+    }
+
+    static AiServiceInteractionEventBuilder builder() {
```
edeandrea, line 36: All public methods need Javadocs

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/AiServiceInteractionEvent.java
```diff
@@ -0,0 +1,68 @@
...
+        return new AiServiceInteractionEventBuilder();
+    }
+
+    class AiServiceInteractionEventBuilder extends Builder<AiServiceInteractionEvent> {
```
edeandrea, line 43: All public classes need Javadocs

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/AiServiceInteractionEvent.java
```diff
@@ -0,0 +1,68 @@
...
+        private final List<AiServiceEvent> events = new ArrayList<>();
+
+        public AiServiceInteractionEventBuilder() {}
```
edeandrea: The constructor in a builder should be `protected`. This is so the only way to instantiate it is to use the factory `builder` or `toBuilder` methods declared in the enclosing class.

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/AiServiceInteractionEvent.java
```diff
@@ -0,0 +1,68 @@
...
+
+        public AiServiceInteractionEventBuilder(AiServiceInteractionEvent src) {
```
edeandrea: The constructor in a builder should be `protected`. This is so the only way to instantiate it is to use the factory `builder` or `toBuilder` methods declared in the enclosing class.

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/AiServiceInteractionEvent.java
```diff
@@ -0,0 +1,68 @@
...
+            super(src);
+            events(src.events());
+        }
+
+        public AiServiceInteractionEventBuilder invocationContext(InvocationContext invocationContext) {
```
edeandrea, line 59: Needs javadocs

langchain4j-core/src/main/java/dev/langchain4j/observability/api/event/AiServiceInteractionEvent.java
```diff
@@ -0,0 +1,68 @@
...
+            return (AiServiceInteractionEventBuilder) super.invocationContext(invocationContext);
+        }
+
+        public AiServiceInteractionEventBuilder events(List<AiServiceEvent> events) {
+            if (events != null) {
```
edeandrea, line 67: When passing in a list of events, is the intention to overwrite the set of events that could already be in `this.events`? Or append to it?

If the intention is to overwrite, you should do a `this.events.clear()` prior to the null-check.

Either way there should be javadocs on the method that declare the intent.

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
```diff
@@ -0,0 +1,43 @@
...
+
+import dev.langchain4j.invocation.InvocationContext;
+import dev.langchain4j.observability.api.event.AiServiceEvent;
+import dev.langchain4j.observability.api.event.AiServiceInteractionEvent;
+
+import java.util.List;
+
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+public class DefaultAiServiceInteractionEvent implements AiServiceInteractionEvent {
```
edeandrea: Missing Javadocs

edeandrea: `DefaultAiServiceInteractionEvent` can extend `AbstractAiServiceEvent` in addition to implementing `AiServiceInteractionEvent`. This will take care of common attributes (like `InvocationContext`).

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final InvocationContext invocationContext;
```
edeandrea: Not needed by extending `AbstractAiServiceEvent`

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
```diff
@@ -0,0 +1,43 @@
...
+
+    private final List<AiServiceEvent> events;
+
+    public DefaultAiServiceInteractionEvent(AiServiceInteractionEvent.AiServiceInteractionEventBuilder builder){
+        this.invocationContext = ensureNotNull(builder.invocationContext(), "invocationContext");
+        this.events = List.copyOf(ensureNotNull(builder.events(), "events"));
```
edeandrea: You don't need to do `List.copyOf` because `builder.events()` already creates a copy.

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
```diff
@@ -0,0 +1,43 @@
...
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
+public class DefaultAiServiceInteractionEvent implements AiServiceInteractionEvent {
+
+    private final InvocationContext invocationContext;
+
+    private final List<AiServiceEvent> events;
+
+    public DefaultAiServiceInteractionEvent(AiServiceInteractionEvent.AiServiceInteractionEventBuilder builder){
+        this.invocationContext = ensureNotNull(builder.invocationContext(), "invocationContext");
```
edeandrea: Not needed by extending `AbstractAiServiceEvent`

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
edeandrea: You don't need `ensureNotNull` because `builder.events()` guarantees never to return `null`

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
edeandrea: Once you change the class to extend `AbstractAiServiceEvent` you can call `super(builder);`

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
```diff
@@ -0,0 +1,43 @@
...
+        this.events = List.copyOf(ensureNotNull(builder.events(), "events"));
+    }
+
+    @Override 
+    public InvocationContext invocationContext(){
```
edeandrea: Not needed once you extend `AbstractAiServiceEvent`

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
```diff
@@ -0,0 +1,43 @@
...
+        return invocationContext;
+    }
+
+    @Override 
+    public List<AiServiceEvent> events(){
+        return events;
+    }
+
+    @Override 
+    public String toString(){
```
edeandrea: Not needed

langchain4j-core/src/main/java/dev/langchain4j/observability/event/DefaultAiServiceInteractionEvent.java
```diff
@@ -0,0 +1,43 @@
...
+        return "DefaultAiServiceInteractionEvent{" + "invocationContext=" + invocationContext + ", events=" + events + "}";
+    }
+
+    @Override
+    public AiServiceInteractionEventBuilder toBuilder(){
```
edeandrea: Not needed - its implemented in the `AiServiceInteractionEvent` interface

langchain4j-core/src/main/java/dev/langchain4j/observability/api/listener/AiServiceInteractionListener.java
```diff
@@ -0,0 +1,14 @@
+package dev.langchain4j.observability.api.listener;
+
+import dev.langchain4j.observability.api.event.AiServiceInteractionEvent;
+
+/**
+ * Listener for AiServiceInteractionEvent, fired once per completed or failed invocation.
+ */
+public interface AiServiceInteractionListener extends AiServiceListener<AiServiceInteractionEvent> {
```
edeandrea, line 9: Interface should have `@FunctionalInterface` annotation

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
```diff
@@ -52,20 +60,79 @@ public <T extends AiServiceEvent> void unregister(AiServiceListener<T> listener)
...
+        }
+
+        synchronized List<AiServiceEvent> snapshot(){
+            return List.copyOf(events);
+        }
+
+    }
+
+    // Holds the events for each invocationId
+    private final Map<UUID, InvocationState> invocationStates = new ConcurrentHashMap<>();
```
edeandrea, line 78: Prefer keeping all class-level attribute declarations together at the top of the class.

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
```diff
@@ -52,20 +60,79 @@ public <T extends AiServiceEvent> void unregister(AiServiceListener<T> listener)
...
                 .ifPresent(l -> l.fireEvent(event));
+          
+
+        // Tracks events per invocationId and emits a single AiServiceInteractionEvent
+        // when the invocation completes or errors. If invocationId is null, aggregation
+        // is skipped to preserve backward compatibility with existing events. 
+        UUID invocationId = event.invocationContext().invocationId();
+        if (invocationId == null) {
+            return; 
+        }
```
edeandrea, line 102: Instead of this, I prefer the inverse without returning.

```java
if (invocationId != null) {
  // Do the rest of the stuff in the method
}
```

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
```diff
@@ -52,20 +60,79 @@ public <T extends AiServiceEvent> void unregister(AiServiceListener<T> listener)
...
      * Fires the given event to all registered {@link AiServiceListener}s.
      *
      * @param <T>   The type of the event, which must be a subtype of {@link AiServiceEvent}.
      * @param event The event to be fired to the listeners. Must not be null.
+     * 
      */
     @Override
     public <T extends AiServiceEvent> void fireEvent(T event) {
         ensureNotNull(event, "event");
         Optional.ofNullable(this.listeners.get(event.eventClass()))
```
edeandrea, line 91: We should probably add a check here that the event type is **NOT** an `AiServiceInteractionEvent`. Other "things" shouldn't be firing them, since they are "composite" events.

Unless @dliubarskyi feels otherwise, I think throwing a `RuntimeException` in that case would be better here (its failing fast in a condition that shouldn't be happening unless some downstream consumer is doing it, and in that case, they shouldn't be).

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
```diff
@@ -52,20 +60,79 @@ public <T extends AiServiceEvent> void unregister(AiServiceListener<T> listener)
...
 
+    // Holds ordered events for a single invocationId (thread-safe)
+    private static final class InvocationState{
+        private final List<AiServiceEvent> events = new ArrayList<>();
+
+        synchronized void add(AiServiceEvent event){
+            events.add(event);
+        }
+
+        synchronized List<AiServiceEvent> snapshot(){
```
edeandrea, line 71: instead of `snapshot` maybe something like `events()`?

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
```diff
@@ -52,20 +60,79 @@ public <T extends AiServiceEvent> void unregister(AiServiceListener<T> listener)
...
+            return List.copyOf(events);
```
edeandrea, line 72: Doesn't need to be a copy

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
```diff
@@ -52,20 +60,79 @@ public <T extends AiServiceEvent> void unregister(AiServiceListener<T> listener)
                 .ifPresent(eventListeners -> eventListeners.remove(listener));
     }
 
+    // Holds ordered events for a single invocationId (thread-safe)
+    private static final class InvocationState{
```
edeandrea, line 64: This isn't the correct way to ensure thread safety. Putting synchronized on each of the methods doesn't ensure that one thread can't be mutating the state while another is reading from it. It only ensures that 1 thread can be writing and 1 thread can be reading, but those could be 2 different threads that are still happening at the same time.

Generally-speaking, using the `synchronized` keyword is a code smell. There are a few valid reasons to use it, but generally it isn't a good idea because it is too coarse-grained.

Using a Read/Write lock is much better. If you look in the nested `EventListeners` class you can see an example of how its used.

Inside the `add` method you'd obtain a lock on the write lock, and in the `events` method you'd obtain a lock on the read lock.

In that scenario, any number of simultaneous threads can read from the list, but only a single thread can write, and the write lock can not be acquired until all reads have completed.

langchain4j-core/src/main/java/dev/langchain4j/observability/api/DefaultAiServiceListenerRegistrar.java
```diff
@@ -52,20 +60,79 @@ public <T extends AiServiceEvent> void unregister(AiServiceListener<T> listener)
...
+                AiServiceInteractionEvent.builder()
+                    .invocationContext(event.invocationContext())
+                    .events(events)
+                    .build();
+
+            Optional.ofNullable(this.listeners.get(interactionEvent.eventClass()))
+                .map(l -> (EventListeners<AiServiceInteractionEvent>) l)
+                .ifPresent(l -> l.fireEvent(interactionEvent));
+        }
+
```
edeandrea, line 130: I think the logic here could be made simpler.

Essentially you are always going to need the `InvocationState` and track things, so you could do something like this to simplify things:

```java
InvocationState state = invocationStates.computeIfAbsent(invocationId, id -> new InvocationState());
state.

add(event);

if ((event instanceof AiServiceCompletedEvent) || (event instanceof AiServiceErrorEvent)) {
    // The interaction is over
    // Stop tracking
    // Create the event
    // clean up the tracking of the interaction
    // fire the event
    List<AiServiceEvent> events = state.events();
    this.invocationStates.remove(invocationid);

    AiServiceInteractionEvent interactionEvent =
                AiServiceInteractionEvent.builder()
                    .invocationContext(event.invocationContext())
                    .events(events)
                    .build();

    Optional.ofNullable(this.listeners.get(interactionEvent.eventClass()))
                .map(l -> (EventListeners<AiServiceInteractionEvent>) l)
                .ifPresent(l -> l.fireEvent(interactionEvent));
}
```

edeandrea (changes requested):
Hi @rayanahmed1215 nice work!

I've gone through this.

One thing I noticed is I don't see any changes to https://github.com/langchain4j/langchain4j/blob/main/langchain4j/src/test/java/dev/langchain4j/service/AiServicesObservabilityTests.java? I would have expected to see those tests updated to additionally assert for the new event type.

Also, could you please add documentation in https://github.com/langchain4j/langchain4j/blob/main/docs/docs/tutorials/observability.md about this new event?

Feel free to ask any clarifying questions!

rayanahmed1215:
sorry for the late reply @edeandrea and thanks for the insight.

I missed those tests for observability as I just put some tests in the tests location I found and I missed the other, and for the documentation and code changes I will address them later. I will work on adding those other tests too.

I cant get into it for now due to finals but I will work on them and follow up on it to you.

edeandrea:
No rush and good luck!

rayanahmed1215:
@edeandrea Sorry for the silence, I tried to reach out but Im not sure it worked. But I was wondering if I should worry about the other closed feature.  If not I can get started on documentation and addressing the other topics you said.

edeandrea:
Hi @rayanahmed1215 nope dont worry about the other closed pull request. This one is all yours to finish!

PR #5060, opened 2026-04-29
AI Services: support polymorphic return types and tool parameters

### LINKED ISSUE #2557: [FEATURE]  LangChain4j's AiService to Support Polymorphic JSON Deserialization

dliubarskyi:
## Issue
Closes https://github.com/langchain4j/langchain4j/issues/2557

## Change
- Polymorphic types (sealed interfaces/classes, or types annotated with Jackson
    `@JsonSubTypes`/`@JsonTypeInfo`) can now be used as AI Service return types and as
    tool parameters. The schema sent to the LLM contains an `anyOf` over concrete
    subtypes with a discriminator property, and the LLM's response (or tool call) is
    dispatched to the correct subtype automatically.
- Works for the polymorphic type itself, for `List<T>`/`Set<T>` of polymorphic, and
    for polymorphic types nested inside other POJOs. Fixes the previously-empty schema
    generated for interface/abstract return types.
- Sealed types work with **zero annotations**; Jackson-annotated types are also
    supported and respect the user-configured `property`, `defaultImpl`, `visible`,
    `@JsonSubTypes.Type(name=...)`, and `@JsonTypeName`.
- Unsupported Jackson configurations (`Id.CLASS`, `Id.MINIMAL_CLASS`, `Id.CUSTOM`,
    `Id.DEDUCTION`, `As.WRAPPER_OBJECT`, `As.WRAPPER_ARRAY`, `As.EXTERNAL_PROPERTY`,
    `As.EXISTING_PROPERTY`) fail fast with a clear `UnsupportedFeatureException` at
    schema-generation time, instead of silently producing a payload Jackson can't parse.
- Discriminator field collisions on subtypes are detected and reported with three
    remediation options (rename, change `property=...`, or set `visible=true`).

## General checklist

### DIFF
--- a/docs/docs/tutorials/structured-outputs.md
+++ b/docs/docs/tutorials/structured-outputs.md
@@ -683,7 +683,7 @@ class names.
 | Attribute | Supported values |
 |---|---|
 | `use` | `Id.NAME`, `Id.SIMPLE_NAME` |
-| `include` | `As.PROPERTY` (default) |
+| `include` | `As.PROPERTY` (default), `As.EXISTING_PROPERTY` |
 | `property` | Any explicit value; defaults to `"@type"` when blank |
 | `defaultImpl` | Any concrete subclass — used when the LLM's discriminator is missing or unknown |
 | `visible` | `true` keeps the discriminator field on the deserialized bean (and bypasses the field-collision check) |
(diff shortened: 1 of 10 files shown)


mariofusco (dismissed):
Great job!

PR #5064, opened 2026-04-30
fix: expose Bedrock Cohere embedding token usage

### LINKED ISSUE #5057: BedrockCohereEmbeddingModel does not expose real token counts from X-Amzn-Bedrock-Input-Token-Count response header
## Problem

`BedrockCohereEmbeddingModel` (and `AbstractBedrockEmbeddingModel`) only reads token counts from the response **body** via `getInputTextTokenCount()`. However, Cohere Embed Multilingual v3 does **not** include token counts in the response body — it only returns `id`, `texts`, `embeddings`, and `response_type`.

As a result, `response.tokenUsage()` always returns `0` for Cohere Embed v3 on Bedrock, making cost tracking impossible.

## Root cause

AWS Bedrock returns the real token count in the HTTP response header:
```
X-Amzn-Bedrock-Input-Token-Count: 42
```

This header is available on the `InvokeModelResponse` via:
```java
response.sdkHttpResponse().firstMatchingHeader("x-amzn-bedrock-input-token-count")
```

But `AbstractBedrockEmbeddingModel` discards the raw `InvokeModelResponse` after reading the body, so the header is never surfaced.

## Expected behaviour

`response.tokenUsage().inputTokenCount()` should return the real token count from the header.

## Suggested fix

In `AbstractBedrockEmbeddingModel.embedAll()`, read the header from the `InvokeModelResponse` before discarding it, and use it to populate `TokenUsage` — falling back to `getInputTextTokenCount()` from the body for models that do include it there.

## Environment

- LangChain4j version: 1.12.2
- Model: `cohere.embed-multilingual-v3` on AWS Bedrock (region `eu-central-1`)
- Verified manually with a raw `curl` request that the header is present and correct
- Related issue: #3508

DragonFSKY:
Fixes #5057

## Summary
- Read Bedrock Cohere input token counts from `x-amzn-bedrock-input-token-count`
- Fall back to the response body token count when the header is not present
- Sum token usage across batched embedding requests and keep `tokenUsage()` null when no count is provided

## Testing
- `mvn -pl langchain4j-bedrock spotless:check`
- `mvn -pl langchain4j-bedrock -am -Dtest=BedrockCohereEmbeddingModelTest,BedrockCohereEmbeddingModelIT -Dsurefire.failIfNoSpecifiedTests=false test`

### DIFF
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockCohereEmbeddingModel.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockCohereEmbeddingModel.java
@@ -13,11 +13,13 @@ import dev.langchain4j.data.embedding.Embedding;
 import dev.langchain4j.data.segment.TextSegment;
 import dev.langchain4j.model.embedding.DimensionAwareEmbeddingModel;
 import dev.langchain4j.model.output.Response;
+import dev.langchain4j.model.output.TokenUsage;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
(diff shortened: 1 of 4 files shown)


dliubarskyi (approved):
@DragonFSKY thank you!

PR #5066, opened 2026-04-30
Allow ClassPathSkillLoader to load skills packaged in JAR files

### LINKED ISSUE #5058: [BUG] ClassPathSkillLoader in Non-Nested Spring Boot Applications Throws FileSystemNotFoundException
**Describe the bug**

ClassPathSkillLoader in Non-Nested Spring Boot Applications Throws FileSystemNotFoundException. 

**Log and Stack trace**

```
...
Exception in thread "main" java.nio.file.FileSystemNotFoundException
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getFileSystem(ZipFileSystemProvider.java:153)
        at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.getPath(ZipFileSystemProvider.java:139)
        at java.base/java.nio.file.Path.of(Path.java:206)
        at dev.langchain4j.skills.ClassPathSkillLoader.resolveClasspathDirectory(ClassPathSkillLoader.java:165)
        at dev.langchain4j.skills.ClassPathSkillLoader.loadSkills(ClassPathSkillLoader.java:55)
        at org.example.demo.Langchain4jSpringBootPlaygroundApplication.main(Langchain4jSpringBootPlaygroundApplication.java:13)
```

**To Reproduce**

1. I use the following Maven configuration to copy all JAR dependencies to the lib folder, keeping the application itself clean (using a non-nested Spring Boot JAR format).

```xml
...
    <build>
        <plugins>
            <!--
            <plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
                <configuration>
                    <excludes>
                        <exclude>
                            <groupId>org.projectlombok</groupId>
                            <artifactId>lombok</artifactId>
                        </exclude>
                    </excludes>
                </configuration>
            </plugin>
            -->
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <executions>
                    ...
                </executions>
            </plugin>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-jar-plugin</artifactId>
                <configuration>
                    <archive>
                        <manifest>
                            <mainClass>org.example.demo.Langchain4jSpringBootPlaygroundApplication</mainClass>
                            <addClasspath>true</addClasspath>
                            <classpathPrefix>lib/</classpathPrefix>
                        </manifest>
                    </archive>
                </configuration>
            </plugin>
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-dependency-plugin</artifactId>
                <executions>
                    <execution>
                        <id>copy-dependencies-to-lib</id>
                        <phase>package</phase>
                        <goals>
                            <goal>copy-dependencies</goal>
                        </goals>
                        <configuration>
                            <outputDirectory>${project.build.directory}/lib/</outputDirectory>
                            <includeScope>runtime</includeScope>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
        </plugins>
    </build>
...
```

2. Then, the main method loads skills using ClassPathSkillLoader.

```java
...
        final var classLoader = Langchain4jSpringBootPlaygroundApplication.class.getClassLoader();
        final var skills = Skills.from(ClassPathSkillLoader.loadSkills("skills", classLoader));
        SpringApplication.run(Langchain4jSpringBootPlaygroundApplication.class, args);
    }
...
```

3. Run the packaging, then launch the application via the command line.

```shell
./mvnw clean package
cd target
java -jar xxx.jar
```

Full source code:

[langchain4j-spring-boot-playground.zip](https://github.com/user-attachments/files/27199599/langchain4j-spring-boot-playground.zip)

**Expected behavior**

No exception.

- LangChain4j version: 1.13.1 (langchain4j-skills 1.13.1-beta23)
- LLM(s) used: N/A
- Java version: Java 25
- Spring Boot version (if applicable): 4.0.6

**Additional context**

jmartisk:
- Fixes: https://github.com/langchain4j/langchain4j/issues/5058

The second commit (about closing) is kinda up for debate but I think this is the best way to prevent leaking open `FileSystem` instances. If we think it's not a big issue, I could remove that commit.

### DIFF
--- a/langchain4j-skills/revapi.json
+++ b/langchain4j-skills/revapi.json
@@ -1,0 +1,34 @@
+[
+  {
+    "extension": "revapi.differences",
+    "configuration": {
+      "ignore": true,
+      "differences": [
+        {
+          "code": "java.method.returnTypeChanged",
+          "old": "method dev.langchain4j.skills.FileSystemSkill dev.langchain4j.skills.ClassPathSkillLoader::loadSkill(java.lang.String)",
+          "new": "method dev.langchain4j.skills.Skill dev.langchain4j.skills.ClassPathSkillLoader::loadSkill(java.lang.String)",
+          "justification": "ClassPathSkillLoader no longer exposes FileSystemSkill because skills loaded from JARs do not have a meaningful filesystem base path"
+        },
+        {
+          "code": "java.method.returnTypeChanged",
+          "old": "method dev.langchain4j.skills.FileSystemSkill dev.langchain4j.skills.ClassPathSkillLoader::loadSkill(java.lang.String, java.lang.ClassLoader)",
+          "new": "method dev.langchain4j.skills.Skill dev.langchain4j.skills.ClassPathSkillLoader::loadSkill(java.lang.String, java.lang.ClassLoader)",
+          "justification": "ClassPathSkillLoader no longer exposes FileSystemSkill because skills loaded from JARs do not have a meaningful filesystem base path"
+        },
+        {
+          "code": "java.method.returnTypeTypeParametersChanged",
+          "old": "method java.util.List<dev.langchain4j.skills.FileSystemSkill> dev.langchain4j.skills.ClassPathSkillLoader::loadSkills(java.lang.String)",
+          "new": "method java.util.List<dev.langchain4j.skills.Skill> dev.langchain4j.skills.ClassPathSkillLoader::loadSkills(java.lang.String)",
+          "justification": "ClassPathSkillLoader no longer exposes FileSystemSkill because skills loaded from JARs do not have a meaningful filesystem base path"
+        },
+        {
+          "code": "java.method.returnTypeTypeParametersChanged",
+          "old": "method java.util.List<dev.langchain4j.skills.FileSystemSkill> dev.langchain4j.skills.ClassPathSkillLoader::loadSkills(java.lang.String, java.lang.ClassLoader)",
+          "new": "method java.util.List<dev.langchain4j.skills.Skill> dev.langchain4j.skills.ClassPathSkillLoader::loadSkills(java.lang.String, java.lang.ClassLoader)",
+          "justification": "ClassPathSkillLoader no longer exposes FileSystemSkill because skills loaded from JARs do not have a meaningful filesystem base path"
+        }
+      ]
+    }
+  }
+]
--- a/langchain4j-skills/src/main/java/dev/langchain4j/skills/ClassPathSkillLoader.java
+++ b/langchain4j-skills/src/main/java/dev/langchain4j/skills/ClassPathSkillLoader.java
@@ -10,6 +10,9 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.nio.charset.MalformedInputException;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystemAlreadyExistsException;
+import java.nio.file.FileSystems;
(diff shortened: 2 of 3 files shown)


langchain4j-skills/src/main/java/dev/langchain4j/skills/DefaultFileSystemSkill.java
```diff
@@ -1,20 +1,42 @@
...
-public class DefaultFileSystemSkill extends AbstractSkill implements FileSystemSkill {
+public class DefaultFileSystemSkill extends AbstractSkill implements FileSystemSkill, Closeable {
 
     private final Path basePath;
+    private final FileSystem jarFileSystem;
 
     public DefaultFileSystemSkill(Builder builder) {
         super(builder);
         this.basePath = ensureNotNull(builder.basePath, "basePath");
+        this.jarFileSystem = builder.jarFileSystem;
```
dliubarskyi: Why do we need to keep the reference to the file system? Can't we close it automatically after the skill is instantiated?

jmartisk: The `basePath` field (right above this line) will stop working right after you close the filesystem because it is scoped to that filesystem. For example, `ShellSkills` would not be able to use this because of https://github.com/langchain4j/langchain4j/blob/1.13.1/experimental/langchain4j-experimental-skills-shell/src/main/java/dev/langchain4j/skills/shell/ShellSkills.java#L151... or if the application, for whatever reason, needs to further look at the file structure

jmartisk: It would also cause a change in behavior depending on whether the application executed from a JAR or not (for example, Quarkus Dev mode) - in JAR mode, the path wouldn't work, and in Dev mode, it would work. I would like to avoid such differences if possible..

dliubarskyi: > The basePath field (right above this line) will stop working right after you close the filesystem because it is scoped to that filesystem. For example, ShellSkills would not be able to use this because of https://github.com/langchain4j/langchain4j/blob/1.13.1/experimental/langchain4j-experimental-skills-shell/src/main/java/dev/langchain4j/skills/shell/ShellSkills.java#L151... or if the application, for whatever reason, needs to further look at the file structure

Hmm, so this is only for shell skills, right? I am not sure it is very useful. LLM will see the path (e.g., "/skills/using-process-tool/SKILL.md"), but how will it be able to read that skill inside the jar via shell command? There is even no path to the jar file so that LLM could unzip it and read.

dliubarskyi: In my understnading JAR-packaged skills are useful only in tool mode. For shell mode, a complete filesystem is required with potential write access (which I guess jar filesystem does not support?) and script execution

jmartisk:
Changed `ClassPathSkillLoader` to return just `Skill` instead of `FileSystemSkill`. Skills loaded specifically from JARs are now `DefaultSkill` instead of `DefaultFileSystemSkill`

The `FileSystem` of the JAR is now closed immediately after loading the skill.

dliubarskyi (approved):
@jmartisk thank you!

PR #5078, opened 2026-05-02
fix: read ReturnBehavior from ToolServiceContext in streaming handler

### LINKED ISSUE #5079: [BUG] ToolProvider's ReturnBehavior (IMMEDIATE / IMMEDIATE_IF_LAST) is ignored in streaming AI services
**Describe the bug**

When a tool is supplied through a `ToolProvider` (rather than registered statically via `@Tool` / `.tools(...)`) and is configured with `ReturnBehavior.IMMEDIATE` or `IMMEDIATE_IF_LAST`, the streaming AI service path (`AiServiceStreamingResponseHandler`) silently downgrades it to `TO_LLM` and proceeds to ask the model again instead of short-circuiting.

The non-streaming path (`ToolService.executeInferenceAndToolsLoop`, line 393) reads `ReturnBehavior` from the per-invocation `ToolServiceContext` — which correctly contains `ToolProvider`-supplied tools — so it works as expected there.

The streaming path reads from `context.toolService.returnBehavior(...)` instead. `ToolService` instance fields only ever contain **statically registered** tools (see `ToolService#tools(...)` and `ToolService#findTools(...)`). Tools coming from a `ToolProvider` are added to a local copy that becomes the `ToolServiceContext` (`ToolService.java:286-313`) and never make it into the long-lived `ToolService` state — by design, since `ToolProvider`s (especially dynamic ones) are evaluated per invocation. Consequently the streaming handler's lookup always returns the default `TO_LLM`.

Affected lines on current `main`:
- `langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java:313`
- `langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java:335`

**Log and Stack trace**

No stack trace at runtime — the bug is silent: the LLM is just called one extra time. In the unit test added in the linked PR the bug is observable as `IllegalArgumentException: aiMessage cannot be null` because the mock streaming model has no second response queued.

**To Reproduce**

```java
ToolSpecification tool = ToolSpecification.builder()
        .name("calculate")
        .description("Performs calculation")
        .build();

ToolExecutor executor = (req, memId) -> "4";

// Mock that returns a tool call exactly once.
AiMessage toolCall = AiMessage.from(ToolExecutionRequest.builder()
        .id("calc-1").name("calculate").arguments("{}").build());
StreamingChatModelMock mock = StreamingChatModelMock.thatAlwaysStreams(toolCall);

ToolProvider provider = req -> ToolProviderResult.builder()
        .add(tool, executor, ReturnBehavior.IMMEDIATE)
        .build();

interface Assistant { TokenStream chat(String msg); }

        .streamingChatModel(mock)
        .toolProvider(provider)
        .build();

assistant.chat("anything")
        .onCompleteResponse(r -> System.out.println("done: " + r))
        .onError(t -> t.printStackTrace())
        .start();
```

Expected: `onCompleteResponse` fires after the tool executes once.
Actual: handler treats the tool as `TO_LLM`, calls the streaming model a second time, `StreamingChatModelMock` has nothing left and throws `aiMessage cannot be null` (real LLMs would just be re-invoked unnecessarily).

The synchronous counterpart of this scenario (`ProgrammaticCreatedImmediateReturnToolTest#should_execute_dynamic_tool_with_immediate_return`) already exists in the test suite and works correctly — only streaming is broken.

**Expected behavior**

`ToolProvider`-supplied tools' `ReturnBehavior` should be honored in the streaming path, identical to the non-streaming path. After executing a tool with `IMMEDIATE` (or after the last tool when it has `IMMEDIATE_IF_LAST` and no other tool errored), the streaming handler should call `onCompleteResponse` and stop, not issue another LLM request.

- LangChain4j version: reproduced on `main` (1.15.0-SNAPSHOT) and present in the 1.14.0 release as well
- LLM(s) used: any streaming model — bug is in the AI service handler, not in any provider
- Java version: 17+

**Additional context**

Fix proposed in #5078 (https://github.com/langchain4j/langchain4j/pull/5078) — two-line change: read from the handler's `toolServiceContext` (which `ToolService.createContext(...)` already populated from both static tools and `ToolProvider` results) instead of from `context.toolService`. Includes a regression test mirroring the existing synchronous test. Would appreciate a maintainer review.

Ivan97:
## Issue
Closes #5079

`AiServiceStreamingResponseHandler` was reading `ReturnBehavior` from `ToolService` instance fields, which only contain statically registered tools. Tools supplied via `ToolProvider` are recorded in the per-invocation `ToolServiceContext`, so their configured `ReturnBehavior.IMMEDIATE` / `IMMEDIATE_IF_LAST` was silently downgraded to `TO_LLM` in the streaming path. The non-streaming path (`ToolService.executeInferenceAndToolsLoop`, line 393) already reads from `toolServiceContext` correctly — this PR aligns the streaming path with that.

## Change

**`AiServiceStreamingResponseHandler.java` — two-line fix at the call sites where the per-tool `ReturnBehavior` is collected:**

```diff
-returnBehaviors.add(context.toolService.returnBehavior(toolRequest.name()));
+returnBehaviors.add(toolServiceContext.returnBehavior(toolRequest.name()));
```

Applied to both the concurrent (`toolExecutor != null`) and sequential branches. The handler already holds the right `ToolServiceContext` as an instance field (`this.toolServiceContext`).

**Regression test:** `StreamingProgrammaticCreatedImmediateReturnToolTest` — streaming counterpart of the existing `ProgrammaticCreatedImmediateReturnToolTest`. Mocks the streaming model with a single tool-call message; if `IMMEDIATE` is honored, the handler completes after executing the tool. If `IMMEDIATE` is dropped (the bug), the handler attempts a second LLM call, the mock queue is empty, and `StreamingChatModelMock` throws `aiMessage cannot be null`, surfacing through `onError`.

Verified: test fails on `main` with `IllegalArgumentException: aiMessage cannot be null` and passes with the fix applied.

## General checklist
- [x] There are no breaking changes (API, behaviour) — restores intended behavior of an existing API
- [x] The tests cover both positive and negative cases — verified the test fails without the fix and passes with it
- [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green (`./mvnw -pl langchain4j -am test -DskipITs` → 1307 tests, 0 failures)
- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs) — N/A (bug fix, no doc-visible API change)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) — N/A
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) — N/A

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -5,9 +5,6 @@ import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 import static dev.langchain4j.service.AiServiceParamsUtil.chatRequestParameters;
 import static dev.langchain4j.service.tool.ToolService.refreshDynamicProviders;
 
-import dev.langchain4j.service.tool.ToolService;
-import dev.langchain4j.service.tool.search.ToolSearchService;
-
 import dev.langchain4j.Internal;
 import dev.langchain4j.agent.tool.ReturnBehavior;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
@@ -43,7 +40,9 @@ import dev.langchain4j.service.tool.ToolExecution;
 import dev.langchain4j.service.tool.ToolExecutionErrorHandler;
 import dev.langchain4j.service.tool.ToolExecutionResult;
 import dev.langchain4j.service.tool.ToolExecutor;
+import dev.langchain4j.service.tool.ToolService;
 import dev.langchain4j.service.tool.ToolServiceContext;
+import dev.langchain4j.service.tool.search.ToolSearchService;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
@@ -310,7 +309,7 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
                                 toResultMessage(toolRequest, toolResult);
                         addToMemory(toolExecutionResultMessage);
                         anyToolErrored = anyToolErrored || toolResult.isError();
-                        returnBehaviors.add(context.toolService.returnBehavior(toolRequest.name()));
+                        returnBehaviors.add(toolServiceContext.returnBehavior(toolRequest.name()));
                     } catch (ExecutionException e) {
                         if (e.getCause() instanceof RuntimeException re) {
                             throw re;
@@ -328,11 +327,10 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
                     toolResults.add(toolResult);
                     ToolRequestResult toolRequestResult = new ToolRequestResult(toolRequest, toolResult);
                     fireToolExecutedEvent(toolRequestResult);
-                    ToolExecutionResultMessage toolExecutionResultMessage =
-                            toResultMessage(toolRequest, toolResult);
+                    ToolExecutionResultMessage toolExecutionResultMessage = toResultMessage(toolRequest, toolResult);
                     addToMemory(toolExecutionResultMessage);
                     anyToolErrored = anyToolErrored || toolResult.isError();
-                    returnBehaviors.add(context.toolService.returnBehavior(toolRequest.name()));
+                    returnBehaviors.add(toolServiceContext.returnBehavior(toolRequest.name()));
                 }
             }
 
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@Ivan97 thank you!

PR #5082, opened 2026-05-02
fix: run tool loop before re-validating output guardrails on reprompt

### LINKED ISSUE #5074: [BUG] OutputGuardrail reprompt skips tool loop — tool-only response surfaces as successful AiMessage with null text
**Describe the bug**

When an `OutputGuardrail` returns `reprompt(...)`, `OutputGuardrailExecutor` calls `chatExecutor().execute(chatMessages)` once and then re-validates whatever comes back. The problem: if the reprompted model responds with `tool_calls` only (no text), the tool calls never get executed. The executor just hands the tool-only `AiMessage` to the next guardrail with empty content.

This works bad with AiServices whose system prompt nudges the model toward tools on retry (e.g. "verify the link with this tool"). The reprompted model will often pick a tool over prose, and any guardrail using `reprompt(...)` against a tool-bearing AiService runs into this.

**Asymmetry with the main flow**

The main AiServices flow deliberately doesn't run output guardrails on tool-only AiMessages. From `AiServiceStreamingResponseHandler#onCompleteResponse` (and the equivalent in `DefaultAiServices`):

```java
if (aiMessage.hasToolExecutionRequests()) {
    // ... loop tool execution, recurse — no guardrail call
    context.streamingChatModel.chat(nextChatRequest, handler);
} else {
    // ... only here do we run executeGuardrails(...) on the final textual response
}
```

So the invariant is already there in the codebase: *guardrails should only see the final textual response, never an intermediate tool-only AiMessage*. The reprompt path in `OutputGuardrailExecutor` breaks that invariant — it can hand a tool-only response to guardrails.

**To Reproduce**

1. Build an AiService with at least one tool registered (`@Tool` or via `ToolProvider`).
2. Configure an `OutputGuardrail` that returns `reprompt(...)` on some validation condition.
3. Send a request that triggers the reprompt, where the reprompted model emits a tool call instead of text. I hit this consistently with `google/gemini-3-flash-preview` on an agent whose system prompt tells the model to verify citations via a search tool. It also reproduces with other tool-enabled models — pretty much any time the reprompt hint pushes the model toward a tool.
4. `TokenStream#onCompleteResponse` (or the `String` return in the synchronous flow) ends up with an `AiMessage` whose `text()` is null.

**Expected behavior**

The reprompt path should hold to the same invariant the main flow enforces: guardrails only see a final textual response, never a tool-only one. Run the same loop that `AiServiceStreamingResponseHandler` / `ToolService.executeInferenceAndToolsLoop` already use on the initial response — validate guardrails only after the loop produces text.

**Environment:**
- LangChain4j version: 1.13.0
- LLM(s) used: `google/gemini-3-flash-preview` (via OpenRouter); reproduces broadly with tool-bearing models
- Java version: 21
- Spring Boot version (if applicable): n/a — using core langchain4j AiServices directly

PSchmitz-Valckenberg:
## Issue
Closes #5074

## Change
When an `OutputGuardrail` returns `reprompt()`, the executor previously called
`chatExecutor.execute(chatMessages)` directly and handed whatever came back
to the next guardrail — including tool-only `AiMessage`s with `text() == null`.

This broke the invariant the main AiServices flow enforces: guardrails should
only see the final textual response, never an intermediate tool-only AiMessage.

Fix: wrap the `ChatExecutor` passed to output guardrails so that `execute(messages)`
runs `executeInferenceAndToolsLoop` before returning, ensuring tool calls are
fully resolved before re-validation.

## General checklist


langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -512,6 +521,55 @@ private UserMessage invokeInputGuardrails(
...
+            public ChatResponse execute(List<ChatMessage> chatMessages) {
+                ChatRequest repromptRequest = ChatRequest.builder()
+                        .messages(chatMessages)
+                        .parameters(parameters)
+                        .build();
+                ChatExecutor repromptExecutor = ChatExecutor.builder(context.chatModel)
+                        .chatRequest(repromptRequest)
+                        .invocationContext(invocationContext)
+                        .eventListenerRegistrar(context.eventListenerRegistrar)
+                        .build();
```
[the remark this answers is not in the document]
PSchmitz-Valckenberg: Fixed — removed the unused chatMemory parameter from buildToolAwareRepromptExecutor

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -512,6 +521,55 @@ private UserMessage invokeInputGuardrails(
...
     }
 
+    private ChatExecutor buildToolAwareRepromptExecutor(
+            ChatExecutor rawChatExecutor,
+            AiServiceContext context,
+            Object memoryId,
+            ChatRequestParameters parameters,
+            ChatMemory chatMemory,
+            InvocationContext invocationContext,
+            ToolServiceContext toolServiceContext,
```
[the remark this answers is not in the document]
PSchmitz-Valckenberg: Fixed — delegating to rawChatExecutor.execute(chatMessages) instead of building a new ChatExecutor, which preserves the transformer behaviour

edeandrea (dismissed):
Thanks for this @PSchmitz-Valckenberg !

ajax-levashov-n:
@PSchmitz-Valckenberg may this be merged, please?

PSchmitz-Valckenberg:
@ajax-levashov-n I don't have write access to merge — 
@dliubarskyi could you merge this when you get a chance? 
It has 1 approval and CI is green.

edeandrea:
I have merge privileges but I don't want to step on @dliubarskyi 's toes if he is holding off for some reason

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -346,11 +345,20 @@ public Object invoke(Method method, Object[] args, InvocationContext invocationC
 
                         ChatResponse aggregateResponse = toolServiceResult.aggregateResponse();
 
+                        ChatExecutor toolAwareRepromptExecutor = buildToolAwareRepromptExecutor(
```
dliubarskyi: What about streaming path, do we need to update `AiServiceStreamingResponseHandler` similarly?

langchain4j/src/test/java/dev/langchain4j/service/guardrail/AiServiceGuardrailTests.java
```diff
@@ -363,6 +363,58 @@ public OutputGuardrailResult validate(AiMessage responseFromLLM) {
         }
     }
 
+    @Test
+    void output_guardrail_reprompt_should_execute_tool_loop_before_revalidating() {
+        // Scenario: guardrail reprompts on "bad response"; reprompted LLM responds with a tool call
+        // instead of text; tool executes; final LLM response passes the guardrail.
+        // Before the fix, the tool-only AiMessage (text == null) was handed directly to the guardrail.
+        var toolWasCalled = new java.util.concurrent.atomic.AtomicBoolean(false);
```
dliubarskyi: please use proper imports (everywhere)

langchain4j/src/test/java/dev/langchain4j/service/guardrail/AiServiceGuardrailTests.java
```diff
@@ -363,6 +363,58 @@ public OutputGuardrailResult validate(AiMessage responseFromLLM) {
         }
     }
 
+    @Test
+    void output_guardrail_reprompt_should_execute_tool_loop_before_revalidating() {
```
dliubarskyi, line 381: This single test covers the golden path. Worth adding:
  - Reprompt response with tool call → guardrail still fails after tool execution → second reprompt occurs (validates the multi-attempt behavior noted in concern #1).
  - Reprompt response with a tool call whose execution errors out.

langchain4j/src/test/java/dev/langchain4j/service/guardrail/AiServiceGuardrailTests.java
```diff
@@ -363,6 +363,58 @@ public OutputGuardrailResult validate(AiMessage responseFromLLM) {
...
+        RepromptAssistant assistant = AiServices.builder(RepromptAssistant.class)
+                .chatModel(model)
+                .tools(tools)
+                .outputGuardrails(repromptOnBad)
+                .build();
+
+        String result = assistant.chat("Hello");
+
+        assertThat(toolWasCalled).isTrue();
+        assertThat(result).isEqualTo("good response");
```
dliubarskyi, line 424: shouldn't this test verify that tool was called and how many times?

dliubarskyi (commented):
@PSchmitz-Valckenberg thanks!

How does it work / supposed to work with IMMEDIATE / IMMEDIATE_IF_LAST tools?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
PSchmitz-Valckenberg: Good point yes, the streaming path has the same issue. AiServiceStreamingResponseHandler passes the raw chatExecutor to output guardrails (line 404), so a reprompt there also bypasses the tool loop. I added the fix for the streaming path in a follow-up commit

langchain4j/src/test/java/dev/langchain4j/service/guardrail/AiServiceGuardrailTests.java
PSchmitz-Valckenberg: Fixed moved all inline fully-qualified class references (AtomicBoolean, ToolExecutionRequest, @Tool) to proper static imports at the top of the file

langchain4j/src/test/java/dev/langchain4j/service/guardrail/AiServiceGuardrailTests.java
PSchmitz-Valckenberg: Agreed the golden path alone isn't sufficient here. I added:

A test where the guardrail still fails after tool execution, triggering a second reprompt — this validates that the multi-attempt reprompt cycle works end-to-end with the tool loop.
A test where the tool throws an exception during execution, to verify the error is propagated correctly rather than swallowed or causing a NullPointerException in the guardrail.

langchain4j/src/test/java/dev/langchain4j/service/guardrail/AiServiceGuardrailTests.java
PSchmitz-Valckenberg: You're right  I switched from AtomicBoolean to AtomicInteger and assert toolCallCount.get() == 1, so the test explicitly verifies the tool ran exactly once and not more

PSchmitz-Valckenberg:
This is a gap in the current fix. With IMMEDIATE/IMMEDIATE_IF_LAST tools, executeInferenceAndToolsLoop returns early without a final LLM call — so aggregateResponse() still contains a tool-only AiMessage (no text), which would be handed directly to the guardrail.

In practice, IMMEDIATE/IMMEDIATE_IF_LAST tools require the method to return Result<T>, so they're incompatible with the typical output-guardrail setup (which expects a plain text response). But that's implicit, not enforced at the guardrail level.

I think the safe approach here is: if aggregateResponse() still has no text after the tool loop (i.e., shouldReturnImmediately fired), treat it the same way as the pre-fix bug — this scenario isn't supported with reprompting and we should either document it clearly or throw an explicit IllegalStateException.

How would you like me to handle this — fail-fast with a clear error, or is there a preferred behavior?

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -434,6 +434,39 @@ private ChatResponse finalResponse(ChatResponse completeResponse, AiMessage aiMe
...
+                // Run the tool loop without writing to memory (reprompt intermediates must not persist).
+                List<ChatMessage> current = new ArrayList<>(chatMessages);
+                int iterationsLeft = maxIterations;
+                while (response.aiMessage().hasToolExecutionRequests()) {
+                    if (iterationsLeft-- == 0) {
+                        throw runtime("Exceeded sequential tool invocations limit during output guardrail reprompt");
+                    }
+                    current.add(response.aiMessage());
+                    List<ToolExecutionResultMessage> toolResults = response.aiMessage().toolExecutionRequests().stream()
+                            .map(req -> toResultMessage(req, AiServiceStreamingResponseHandler.this.execute(req)))
```
dliubarskyi: Any reason to re-implement tool execution loop instead of calling `executeInferenceAndToolsLoop`?

PSchmitz-Valckenberg: Good point, you're right that re-implementing the loop was unnecessary. I've refactored to delegate to `context.toolService.executeInferenceAndToolsLoop(...)` the same way the non-streaming path does in `DefaultAiServices.buildToolAwareRepromptExecutor`. Both paths now use the exact same helper, no duplication.
One nuance worth flagging: in the streaming context, the initial reprompt call still goes through `chatExecutor` (which wraps `streamingChatModel`), but the subsequent tool-loop iterations inside `executeInferenceAndToolsLoop` use `context.chatModel`. In practice this is fine because the reprompt code path is already synchronous (we buffer partials until guardrails finish), but it does mean a streaming-only AI service without a `chatModel` configured would NPE on a tool-bearing reprompt response. Happy to add a fallback or clearer error if you think that's worth handling

PSchmitz-Valckenberg:
Rebased on latest main, merge conflict in DefaultAiServices.java is resolved. CI should be green again. @edeandrea feel free to merge if you're happy with the state, no need to wait on anyone else

edeandrea:
> Rebased on latest main, merge conflict in DefaultAiServices.java is resolved. CI should be green again. @edeandrea feel free to merge if you're happy with the state, no need to wait on anyone else

It looks good to me, but @dliubarskyi was the last one with comments, so I don't want to step on his toes.

edeandrea (dismissed):

joshiste:
@dliubarskyi any update on this fix? desperately waiting for this...

dliubarskyi (changes requested):
@PSchmitz-Valckenberg thank you, but it cannot be merged as it is right now.
Streaming path is not tested, and it seems there is a bug: `executeInferenceAndToolsLoop` is using `chatModel`, while streaming path has only `streamingChatModel` configured

dliubarskyi:
Please ping me once again by @dliubarskyi once the streaming issue is fixed, and I will merge it. 

BTW, the release is planned for 26 June.

PSchmitz-Valckenberg:
@dliubarskyi the streaming issue is fixed. buildToolAwareRepromptExecutor() in the streaming handler now implements its own tool loop using chatExecutor.execute() (which wraps streamingChatModel) instead of delegating to executeInferenceAndToolsLoop(), which internally calls context.chatModel and would throw a NPE when only a streamingChatModel is configured.

Streaming tests for all three scenarios (basic reprompt+tool, multi-reprompt, failing tool) have been added as well. All CI checks are passing now

dliubarskyi (approved):
@PSchmitz-Valckenberg thank you!

PR #5084, opened 2026-05-03
Fix #5079: ToolProvider ReturnBehavior ignored in streaming AI services

### LINKED ISSUE #5079: [BUG] ToolProvider's ReturnBehavior (IMMEDIATE / IMMEDIATE_IF_LAST) is ignored in streaming AI services

anuragg-saxenaa:
AiServiceStreamingResponseHandler was reading ReturnBehavior from
context.toolService.returnBehavior() which only reflects statically
registered tools. ToolProvider-supplied tools are in the per-invocation
toolServiceContext, so ReturnBehavior.IMMEDIATE was silently downgraded
to TO_LLM, causing extra LLM calls.

Fix: read ReturnBehavior from toolServiceContext.returnBehavior()
instead, matching the non-streaming path in ToolService.executeInferenceAndToolsLoop.

Also add regression test that verifies ToolProvider tools with
ReturnBehavior.IMMEDIATE in streaming mode return immediately without
an extra LLM call.

Signed-off-by: Anurag Saxena <anurag.saxena.aug@gmail.com>

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -310,7 +310,7 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
                                 toResultMessage(toolRequest, toolResult);
                         addToMemory(toolExecutionResultMessage);
                         anyToolErrored = anyToolErrored || toolResult.isError();
-                        returnBehaviors.add(context.toolService.returnBehavior(toolRequest.name()));
+                        returnBehaviors.add(toolServiceContext.returnBehavior(toolRequest.name()));
                     } catch (ExecutionException e) {
                         if (e.getCause() instanceof RuntimeException re) {
                             throw re;
@@ -332,7 +332,7 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
                             toResultMessage(toolRequest, toolResult);
                     addToMemory(toolExecutionResultMessage);
                     anyToolErrored = anyToolErrored || toolResult.isError();
-                    returnBehaviors.add(context.toolService.returnBehavior(toolRequest.name()));
+                    returnBehaviors.add(toolServiceContext.returnBehavior(toolRequest.name()));
                 }
             }
 
--- a/langchain4j/src/test/java/dev/langchain4j/service/StreamingAiServicesWithToolsIT.java
+++ b/langchain4j/src/test/java/dev/langchain4j/service/StreamingAiServicesWithToolsIT.java
@@ -1,5 +1,6 @@
 package dev.langchain4j.service;
 
+import static org.mockito.Mockito.when;
 import static dev.langchain4j.MockitoUtils.ignoreInteractions;
 import static dev.langchain4j.model.openai.OpenAiChatModelName.GPT_4_O_MINI;
 import static dev.langchain4j.service.AiServicesWithToolSearchToolIT.containsTool;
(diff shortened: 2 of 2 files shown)


dliubarskyi:
Thanks, but this was fixed in https://github.com/langchain4j/langchain4j/pull/5078

PR #5086, opened 2026-05-03
fix: default primitive tool parameters to zero/false when LLM omits them

### LINKED ISSUE #5085: [BUG] NPE for tool calls containing primitive types - here int - if LLM misses the parameter
The parameter binding has a null pointer as soon as the LLM is missing an argument if primitive types are in use.

**Describe the bug**

If you give me a hint where you have these test, I can try to add a test and fix in langchain4j.

### Tool:
```java
    @Tool(name = READ_ECLIPSE_FILE_TOOL, value = "Eclipse: Read workspace file (e.g. '/Project/src/Foo.java').")
    public String readWorkspaceFile(
            @P("workspace-relative path") String filePath,
            @P("Optional: first line to read (1-based). 0 = start of file.") int startLine,
            @P("Optional: last line to read (1-based). 0 = end of file.") int endLine) {

```

### Test
```java
        ToolService service = new ToolService();
        service.addTool(new EclipseWorkspaceReadFileTool());

        var tr = ToolExecutionRequest.builder().arguments("")
            .name("readWorkspaceFile")
            .arguments("""
                    {"arg0": "EclipseWorkspaceReadFilesToolTest.java"}""")
            .build();
        
        // WHEN
        var content = service.execute(tr, null, null, new ArrayList<ChatMessage>());
```

**Log and Stack trace**

```
ToolExecutionResultMessage{id='null', toolName='readWorkspaceFile', contents=[TextContent { text = "java.lang.NullPointerException: Cannot invoke "java.lang.Number.intValue()" because the return value of "sun.invoke.util.ValueConversions.primitiveConversion(sun.invoke.util.Wrapper, Object, boolean)" is null" }], isError=null, attributes={}}
```

**To Reproduce**
See test above.

**Expected behavior**
It should set 0 as the default value e.g. for `int` types -- as we can never ensure the LLM provides all values.

- LangChain4j version: 1.3.0 and 1.4.0
- LLM(s) used: unknown
- Java version: 21
- Spring Boot version (if applicable):  none

**Additional context**
https://github.com/sterlp/eclipse-peon-ai/issues/18

PSchmitz-Valckenberg:
## Issue
Closes #5085

## Change
When an LLM omits an argument for a primitive-typed tool parameter (\`int\`, \`long\`, \`boolean\`, etc.), the argument slot in \`DefaultToolExecutor.prepareArguments\` stayed \`null\`. Java's unboxing then threw an NPE during method invocation.

Fix: detect missing primitive parameters and supply the standard JVM default value (0 for numeric types, \`false\` for \`boolean\`, \`'\0'\` for \`char\`) instead of leaving the slot \`null\`. This matches the behaviour documented in the issue's expected behavior — primitive parameters should silently default rather than crashing.

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
@@ -175,7 +175,8 @@ public class DefaultToolExecutor implements ToolExecutor {
             return List.of(ImageContent.from(image));
         } else if (result instanceof Content content) {
             return List.of(content);
-        } else if (result instanceof Collection<?> collection && !collection.isEmpty()
+        } else if (result instanceof Collection<?> collection
+                && !collection.isEmpty()
                 && collection.iterator().next() instanceof Content) {
             return collection.stream().map(Content.class::cast).toList();
         } else if (result instanceof Content[] array) {
@@ -232,6 +233,8 @@ public class DefaultToolExecutor implements ToolExecutor {
                 arguments[i] = createOptional(argument, parameterName, parameterType);
             } else if (argument != null) {
                 arguments[i] = coerceArgument(argument, parameterName, parameterClass, parameterType);
+            } else if (parameterClass.isPrimitive()) {
+                arguments[i] = defaultPrimitiveValue(parameterClass);
             }
         }
 
@@ -265,6 +268,18 @@ public class DefaultToolExecutor implements ToolExecutor {
         Class<?> actualClass = extractActualClass(actualType);
         Object coercedValue = coerceArgument(argument, parameterName, actualClass, actualType);
         return Optional.of(coercedValue);
+    }
+
+    private static Object defaultPrimitiveValue(Class<?> primitiveClass) {
+        if (primitiveClass == boolean.class) return false;
+        if (primitiveClass == char.class) return '\0';
+        if (primitiveClass == byte.class) return (byte) 0;
+        if (primitiveClass == short.class) return (short) 0;
+        if (primitiveClass == int.class) return 0;
+        if (primitiveClass == long.class) return 0L;
+        if (primitiveClass == float.class) return 0.0f;
+        if (primitiveClass == double.class) return 0.0;
+        return null;
     }
 
     static Object coerceArgument(Object argument, String parameterName, Class<?> parameterClass, Type parameterType) {
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
@@ -234,6 +234,13 @@ public class DefaultToolExecutor implements ToolExecutor {
             } else if (argument != null) {
                 arguments[i] = coerceArgument(argument, parameterName, parameterClass, parameterType);
             } else if (parameterClass.isPrimitive()) {
+                P pAnnotation = parameter.getAnnotation(P.class);
+                boolean isRequired = pAnnotation == null || pAnnotation.required();
+                if (isRequired) {
+                    throw new ToolArgumentsException(new IllegalArgumentException(String.format(
+                            "Argument \"%s\" of type %s is required but was not provided",
+                            parameterName, parameterClass.getName())));
+                }
                 arguments[i] = defaultPrimitiveValue(parameterClass);
             }
         }
--- a/langchain4j/src/test/java/dev/langchain4j/service/tool/DefaultToolExecutorTest.java
+++ b/langchain4j/src/test/java/dev/langchain4j/service/tool/DefaultToolExecutorTest.java
@@ -364,6 +364,66 @@ class DefaultToolExecutorTest implements WithAssertions {
         assertThat(result).isEqualTo(expectedResult);
     }
 
+    private static class PrimitiveTool {
+
+        @Tool
+        public int sum(int a, int b) {
+            return a + b;
+        }
+
+        @Tool
+        public boolean and(boolean x, boolean y) {
+            return x && y;
+        }
+
+        @Tool
+        public long multiply(long a, long b) {
+            return a * b;
+        }
+    }
+
+    @Test
+    void missing_primitive_int_argument_defaults_to_zero() throws NoSuchMethodException {
+        ToolExecutionRequest request = ToolExecutionRequest.builder()
+                .id("1")
+                .name("sum")
+                .arguments("{ \"arg0\": 3 }")
+                .build();
+
+        DefaultToolExecutor executor = new DefaultToolExecutor(
+                new PrimitiveTool(), PrimitiveTool.class.getDeclaredMethod("sum", int.class, int.class));
+
+        assertThat(executor.execute(request, "DEFAULT")).isEqualTo("3");
+    }
+
+    @Test
+    void missing_primitive_boolean_argument_defaults_to_false() throws NoSuchMethodException {
+        ToolExecutionRequest request = ToolExecutionRequest.builder()
+                .id("1")
+                .name("and")
+                .arguments("{ \"arg0\": true }")
+                .build();
+
+        DefaultToolExecutor executor = new DefaultToolExecutor(
+                new PrimitiveTool(), PrimitiveTool.class.getDeclaredMethod("and", boolean.class, boolean.class));
+
+        assertThat(executor.execute(request, "DEFAULT")).isEqualTo("false");
+    }
+
+    @Test
+    void missing_primitive_long_argument_defaults_to_zero() throws NoSuchMethodException {
+        ToolExecutionRequest request = ToolExecutionRequest.builder()
+                .id("1")
+                .name("multiply")
+                .arguments("{}")
+                .build();
+
+        DefaultToolExecutor executor = new DefaultToolExecutor(
+                new PrimitiveTool(), PrimitiveTool.class.getDeclaredMethod("multiply", long.class, long.class));
+
+        assertThat(executor.execute(request, "DEFAULT")).isEqualTo("0");
+    }
+
     private static class PersonTool {
 
         @Tool
@@ -448,9 +508,7 @@ class DefaultToolExecutorTest implements WithAssertions {
                 .build();
         DefaultToolExecutor toolExecutor2 = new DefaultToolExecutor(new PersonTool(), request2);
         String result2 = toolExecutor2.execute(request2, "DEFAULT");
-        assertThat(result2)
-                .isEqualToIgnoringWhitespace(
-                        """
+        assertThat(result2).isEqualToIgnoringWhitespace("""
                 [
                   {
                     "name": "Klaus",
@@ -469,9 +527,7 @@ class DefaultToolExecutorTest implements WithAssertions {
                 .build();
         DefaultToolExecutor toolExecutor3 = new DefaultToolExecutor(new PersonTool(), request3);
         String result3 = toolExecutor3.execute(request3, "DEFAULT");
-        assertThat(result3)
-                .isEqualToIgnoringWhitespace(
-                        """
+        assertThat(result3).isEqualToIgnoringWhitespace("""
                 [
                   {
                     "name": "Peter",
@@ -491,9 +547,7 @@ class DefaultToolExecutorTest implements WithAssertions {
                 .build();
         DefaultToolExecutor toolExecutor4 = new DefaultToolExecutor(new PersonTool(), request4);
         String result4 = toolExecutor4.execute(request4, "DEFAULT");
-        assertThat(result4)
-                .isEqualToIgnoringWhitespace(
-                        """
+        assertThat(result4).isEqualToIgnoringWhitespace("""
                 {
                   "p1": {
                     "name": "Klaus",
@@ -512,9 +566,7 @@ class DefaultToolExecutorTest implements WithAssertions {
                 .build();
         DefaultToolExecutor toolExecutor5 = new DefaultToolExecutor(new PersonTool(), request5);
         String result5 = toolExecutor5.execute(request5, "DEFAULT");
-        assertThat(result5)
-                .isEqualToIgnoringWhitespace(
-                        """
+        assertThat(result5).isEqualToIgnoringWhitespace("""
                 [
                   {
                     "name": "Klaus",
--- a/langchain4j/src/test/java/dev/langchain4j/service/tool/DefaultToolExecutorTest.java
+++ b/langchain4j/src/test/java/dev/langchain4j/service/tool/DefaultToolExecutorTest.java
@@ -6,9 +6,11 @@ import static java.util.Collections.singletonMap;
 
 import com.fasterxml.jackson.core.JsonParseException;
 import com.fasterxml.jackson.core.type.TypeReference;
+import dev.langchain4j.agent.tool.P;
 import dev.langchain4j.agent.tool.Tool;
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.agent.tool.ToolMemoryId;
+import dev.langchain4j.exception.ToolArgumentsException;
 import dev.langchain4j.invocation.InvocationContext;
 import java.lang.reflect.Method;
 import java.math.BigDecimal;
@@ -372,7 +374,12 @@ class DefaultToolExecutorTest implements WithAssertions {
         }
 
         @Tool
-        public boolean and(boolean x, boolean y) {
+        public int sumWithOptionalB(int a, @P(required = false) int b) {
+            return a + b;
+        }
+
+        @Tool
+        public boolean and(@P(required = false) boolean x, @P(required = false) boolean y) {
             return x && y;
         }
 
@@ -383,7 +390,7 @@ class DefaultToolExecutorTest implements WithAssertions {
     }
 
     @Test
-    void missing_primitive_int_argument_defaults_to_zero() throws NoSuchMethodException {
+    void missing_required_primitive_int_argument_throws() throws NoSuchMethodException {
         ToolExecutionRequest request = ToolExecutionRequest.builder()
                 .id("1")
                 .name("sum")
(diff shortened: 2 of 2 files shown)


sterlp (commented):
One thing I observed is, that the current code is not throwing an exception. But returning a tool result which doesnt contain the whole stack.

Not sure if this is intentional.

The Fix Looks good to me and should fix the described issue.

I would somehow expect a clean exit with an exception -- as a tool result containing the npe exception Message is odd.

PSchmitz-Valckenberg:
Thanks for the review @sterlp ! You're right that returning the error as a tool result message (rather than propagating the exception) is the existing behavior — this PR only fixes the secondary crash when the exception has no message. 

Whether to propagate tool exceptions instead is a separate design question, already covered by the propagateToolExecutionExceptions flag in DefaultToolExecutor. Happy to note this in the issue if useful

sterlp (approved):
fix looks good to me. thanks pointing out DefaultToolExecutor propagateToolExecutionExceptions  - I will check on this - how I may set it.

mariofusco (dismissed):

dliubarskyi:
Please see https://docs.langchain4j.dev/tutorials/tools#required-and-optional

Since we expect all tool parameters to be required by default (unless explicitly marked as optional), when LLM does not provide one, it seems like an issue with LLM generation, so I am not sure it is a good idea to automatically fallback to _**some**_ value. I would expect in this case to call LLM with "parameter X is missing" error so that LLM can generate a missing value.

In case of optional primitive values, I guess we can indeed add an option to fall back to `0`/`false`, but it would probably be more useful to have some parameter in `@P` like `defaultValue` where user could explicitly specify the default value in case LLM did not provide one and `required = false`. When LLM did not provide value and `required = true`, it would make more sense to fail with `ToolArgumentsException` so that we could re-prompt LLM.

WDYT?

PSchmitz-Valckenberg:
Thanks @dliubarskyi — that's a valid distinction I hadn't considered. The current PR silently defaults for all missing primitives, which does hide LLM errors for required parameters.

I'd like to update the PR to reflect your suggestion:

If the parameter is required (default) and the LLM omits it → throw ToolArgumentsException with a clear message like "Parameter 'X' of type int is required but was not provided by the LLM" so re-prompting can occur
If the parameter is not required (required = false in @P or wrapped in Optional) → fall back to the JVM default (0, false, etc.)
On defaultValue in @P — should that be part of this PR or a separate follow-up? I'd prefer to keep the scope minimal for now unless you want both together.

Happy to push the update if this direction sounds right

dliubarskyi:
@PSchmitz-Valckenberg I would like to hear from @sterlp regarding [my suggestion](https://github.com/langchain4j/langchain4j/pull/5086#issuecomment-4369331842) before proceeding. I will also experiment a bit with it today

sterlp:
> Please see https://docs.langchain4j.dev/tutorials/tools#required-and-optional

Well I think it depends here a bit on the perspetive:

## Pauls assumption

1. LLMs will make errors
2. Keep the API simple
3. Avoid a suprise

## Pauls suggestion

- Optional primitive parameters are populated with default values (KIS)
- Optional object parameters are populated with `null` (KIS)
- In case of a problem
   - if propagate error is on: Throw a binding exception - with the tool name and parameter which is missing
   - If propagate error is off: Return a proper message to the LLM including tool name and parameter description and the info that the parameter is missing -- this should also be the fallback for unexpected problems

### Problems I currently see

Returning something to the developer code / API user is here the complex part. I would require a kind of validation exception (JSR validation?) - to handle the error. As right now I have a parameter name `arg0` or `arg1` without setting the name. Means I need the exception with the info: `arg1` is missing where `arg1` is the property name so I may react on it. I hope this makes any sense to you.

This is somehow **not simple** (KIS) anymore and harder to understand.

#### KIS but wrong

This is my current solution - which works just fine with `1.14.0`- it is wrong:

```java
    @Tool(name = READ_ECLIPSE_FILE_TOOL, value = "Eclipse: Read workspace file (e.g. '/Project/src/Foo.java').")
    public String readWorkspaceFile(
            @P("workspace-relative path") String filePath,
            @P("Optional: first line to read (1-based). 0 = start of file.") Integer startLine,
            @P("Optional: last line to read (1-based). 0 = end of file.") Integer endLine) {

        if (filePath == null || filePath.isBlank()) {
            throw new IllegalArgumentException("arg0 must not be empty");
        }
        if (startLine == null) startLine = 0;
```

#### Correct would be

This would be correct **BUT**:

```java
    public String readWorkspaceFile(
            @P(description = "workspace-relative path") String filePath,
            @P(description = "first line to read (1-based). 0 = start of file.", required = false) int startLine,
            @P(description = "last line to read (1-based). 0 = end of file.", required = false) int endLine) {
/* now langchain4j - currently missing
        if (filePath == null || filePath.isBlank()) {
            throw new IllegalArgumentException("filePath must not be empty");
        }
*/
```

Now my expectation would be:
1. langchain4j adds a info for the LLM that `arg1`and `arg2` are optional - how is the question.
2. langchain4j generates a proper error to the LLM if `arg0` is not provided - missing feature/bug - see above

**But:**
If we propagate the errors a kind of new "binding exception" is required, with the exact information. Which in the end has to wrap also JSR and Jackson exceptions? Not sure on that but we would have to then, right?

As otherwise I get an exception from Jackson or JSR validation (feature request) but still need the information from which parameter it comes `arg1` or `arg0`?

dliubarskyi:
@sterlp 

> As right now I have a parameter name arg0 or arg1 without setting the name.

Please check https://docs.langchain4j.dev/tutorials/tools#parameter-name, you don't have to settle with argX, there are ways to use proper names.

> langchain4j adds a info for the LLM that arg1and arg2 are optional - how is the question.

LC4j already marks parameters optional in the generated JSON schema if you set `required = false`

Regarding the rest of comments: let me experiment a bit, I will get back to you soon

dliubarskyi:
So, after a bit of experimenting and analyzing, here are my thoughts.

There are currently 2 issues with `@Tool` method parameters:
1. bug: incorrect error handling when LLM does not provide a value for primitive parameter
2. missing feature: default value for a tool parameter

Let's start from the bug. Currently, when primitive parameter is not supplied by the LLM, we fail with `NullPointerException` which is propagated into `ToolExecutionErrorHandler`. Instead, we should fail with a clear error message and exception of type `ToolArgumentsException`, so that exception is correctly routed to the `ToolArgumentsErrorHandler`, where user will be able to decide if LLM should retry or just fail. I would not default to the JVM default (e.g., `0`/`false`), because for most numeric parameters `0` is rarely the value the
  user actually intends (think IDs, amounts, counts, timestamps)~~, and the framework cannot distinguish "absent" from "explicitly zero". Silently substituting `0` would mask LLM
  mistakes and produce subtle bugs - this is exactly the JAX-RS `@QueryParam` footgun. Failing loudly gives the LLM a chance to retry.~~

I have also realized that user can configure primitive as optional (e.g., `@P(value="age", required=false) int age`) which is clearly incorrect and cannot work, so in this case we need to fail ASAP with `IllegalConfigurationException`.

Now to the missing feature. I can see how setting a default vlaue in your use case would be beneficial, so we will add this feature in the upcoming releases. Here is a proposed solution:
- add new `@P(defaultValue = "...")` attribute
- if this attribute is set, field is implicitly becoming optional (setting `defaultValue = "...", required = true` will fail with `IllegalConfigurationException`)
- default value will also be present in the generated JSON schema
- if LLM does not supply a value, we parse the `defaultValue` attribute and inject it instead of a missing value

WDYT?

---

As to your cusrrent config, I guess this is what will work fine before we'll fix the bug and introduce the missing feature:

```java
   @Tool(name = READ_ECLIPSE_FILE_TOOL, value = "Eclipse: Read workspace file (e.g. '/Project/src/Foo.java').")
    public String readWorkspaceFile(
            @P("workspace-relative path") String filePath,
            @P(description = "first line to read (1-based). 0 = start of file.", required = false) Integer startLine,
            @P(description = "last line to read (1-based). 0 = end of file.", required = false) Integer endLine) {

        if (filePath == null || filePath.isBlank()) {
            throw new IllegalArgumentException("arg0 must not be empty");
        }
        if (startLine == null) startLine = 0;
        ...
   }
```
You can also add information about default values to the description of `startLine`/`endLine` params, so that LLM is aware of  them.


Once missing feature is implemented, you would be able to change it so this:
```java
   @Tool(name = READ_ECLIPSE_FILE_TOOL, value = "Eclipse: Read workspace file (e.g. '/Project/src/Foo.java').")
    public String readWorkspaceFile(
            @P("workspace-relative path") String filePath,
            @P(description = "first line to read (1-based). 0 = start of file.", defaultValue = "0") int startLine,
            @P(description = "last line to read (1-based). 0 = end of file.", defaultValue = "0") int endLine) {

        if (filePath == null || filePath.isBlank()) {
            throw new IllegalArgumentException("arg0 must not be empty");
        }
        
        // startLine and endLine are always set
        ...
   }
```

dliubarskyi:
> langchain4j generates a proper error to the LLM if arg0 is not provided - missing feature/bug - see above

Oh, this is actually a third issue. Currently, if LLM does not provide a value for object type, we pass `null` regardless of optionality. I agree, this is not ideal, but this behavior cannot be changed right now, as suppose many users are relying on it already. But this is a good candidate for a change in 2.0 release.

> But:
> If we propagate the errors a kind of new "binding exception" is required, with the exact information. Which in the end has to wrap also JSR and Jackson exceptions? Not sure on that but we would have to then, right?
> 
> As otherwise I get an exception from Jackson or JSR validation (feature request) but still need the information from which parameter it comes arg1 or arg0?

Not sure I understand what you mean, could you please elaborate?

sterlp:
> mistakes and produce subtle bugs - this is exactly the JAX-RS `@QueryParam` footgun. Failing loudly gives the LLM a chance to retry.

I see your point and this issue becomes more complex in the fix as I expected the longer I think about it. Again I try to be brief, means some thoughts are missing. We should had a coffee about that.

Please also keep in mind that I am not an expert for langchain4j - just a user of the framework. As so my view is subjective and based on a "user" perspective. I try to get some distance :-). Means I don't like a surprise if I use it.

From my side the main issue is, that the framework does not behave as documented: 

# I like:

1. defaultValue option -> _as a Spring user_
2. the framework cannot distinguish "absent" from "explicitly zero". Silently substituting 0 would mask LLM -> _yes hmmm_
3. `IllegalConfigurationException` -> _makes sense but is breaking too_


# Statement

Any object parameter is optional, regardless of the setting of `@P` right now. 
Primitive parameters are all required, regardless of the setting of `@P` right now.
Doc != Code.

# Possible solutions 

None of this is good - as any change to it would break/surprise.

## I. Hotfix
1. Primitive parameters are populated with default values -> bug fix
2. Object parameters are populated with `null` -> unchanged to now

State the problem in the doc - fix will be applied in a major release.

## II. Mostly backward compatible fix

1. Primitive parameters are populated with default values -> bug fix
2.  Object parameters are populated with `null` -> unchanged to now
3. `@P` required is set to `false` <-- slightly breaking
4. If required is set to `true` - by the users
   a)  `IllegalConfigurationException` for primitive optional params
   b) Proper error to the LLM - propagateToolExecutionExceptions=false
   c) Proper exception to the user - propagateToolExecutionExceptions=true 

---

4a) may sound hard but is not surprising. Which is good but I think.
4b) would be the expectation of the most users now anyway - more a fix - I think we are good on that.
4c) breaking change maybe? Should be as it doesn't throw right now, right?
3) may sound bad, but that's the way how it works right now anyway. Only doc fix is required. But of course some code may miss `null` checks - but they will run into the issues anyway - only less likely.

So 3, 4a and 4c can be considered breaking changes.

## III. Proper fix

As above **II.** with all changes and keeping `required` on `true` for `@P` - but that's something for 2.0.0.

----

# Background

## JSR Validation

Somehow I had expected to be able - in the future - something like this:

```java
record Foo(@NotNull String bar, @Sice(min = 4) int anything) {}

public String myTool(
  @P("fill xxx") @NotNull @Valid Foo a,
  @P("fill yyy") @Valid Foo b,
```

So I would expect jsr/jakarta validation as a user for more complex validation tasks. I was a bit surprised to see `required` - but it makes sense too.

### Why: 
Well required is not always required. If I remember right `@JsonProperty(required = true)` allows to set the value in the JSON to `null` - maybe even to remove it. I think it is just for the doc generation. Yes verified this works:

```java
    @Data
    static class Foo {
        @JsonProperty(isRequired = OptBoolean.TRUE)
        String hallo;
    }
    @Test
    void test_serializtion_foo() throws Exception {
        // WHEN
        var read = mapper.readValue("{}", Foo.class);
        
        // THEN
        System.err.println(read);
    }

> Foo(hallo=null)
```

### Complex validation
Meas for structured objects and validation in the framework we would need jsr validation. Which comes with an "BindingException" of some kind like in Spring to say:

`arg0.bar` has not to be null -> to the LLM

## Code

```java
        @Override
        public ChatModel buildChatModel(LlmConfig c) {
            var http1 = JdkHttpClient.builder()
                    .httpClientBuilder(HttpClient.newBuilder()
                            .version(HttpClient.Version.HTTP_1_1));
            var result = OpenAiChatModel.builder()
                    .timeout(TIMEOUT)
                    .baseUrl(c.getUrl())
                    .modelName(c.getModel())
                    .strictJsonSchema(true)
                    .apiKey(StringUtil.hasValue(c.getApiKey()) ? c.getApiKey() : "lm-studio")
                    .httpClientBuilder(http1);
            // no thinking "off" supported
            result.returnThinking(Boolean.TRUE)
                  .sendThinking(Boolean.FALSE);
            return result.build();
        }

    @Tool(name = READ_ECLIPSE_FILE_TOOL, value = "Eclipse: Read workspace file (e.g. '/Project/src/Foo.java').")
    public String readWorkspaceFile(
            @P("workspace-relative path") String filePath,
            @P("first line to read (1-based). 0 = start of file.") Integer startLine,
            @P(description = "last line to read (1-based). 0 = end of file.", required = false) Integer endLine) {
        
        if (startLine == null) startLine = 0;
        if (endLine == null) endLine = 0;

        if (filePath == null || filePath.isBlank()) {
            throw new IllegalArgumentException("filePath must not be empty");
        }

```

## What LM Studio Qwen 3.6 thinks

image

Well close but wrong ...

## What we send
```json
    {
      "type": "function",
      "function": {
        "name": "readWorkspaceFile",
        "description": "Eclipse: Read workspace file (e.g. '/Project/src/Foo.java').",
        "parameters": {
          "type": "object",
          "properties": {
            "arg0": {
              "type": "string",
              "description": "workspace-relative path"
            },
            "arg1": {
              "type": "integer",
              "description": "first line to read (1-based). 0 = start of file."
            },
            "arg2": {
              "type": "integer",
              "description": "last line to read (1-based). 0 = end of file."
            }
          },
          "required": [
            "arg0",
            "arg1"
          ]
        }
      }
    },
```

## What "strict" really means

Just for reference - based on the discussion I think this is valuable for some readers - there is no such thing as "strict tools" - well at least not as long as LLMs are build as they are - but this would take too long to elaborate here. We would need more than one coffee :-)

image

Yes as I expected:
```java
        if (filePath == null || filePath.isBlank()) {
            throw new IllegalArgumentException("filePath must not be empty");
        }
```

Not `arg0 must not be empty` - the tool was called.

dliubarskyi:
> Any object parameter is optional, regardless of the setting of @P right now.
> Primitive parameters are all required, regardless of the setting of @P right now.
> Doc != Code.

There are 2 layers here:
1. Generated JSON schema. `@P(required)` actually affects whether parameter in JSON schema is required or not.
2. Parsing/preparing parameters before calling `@Tool` method. Here we indeed do not fail if required parameter is missing (because LLM did not adhere to the specified JSON schema) - I agree that this is a bug, and we will most probably fix it in 2.0 version, but right now I cannot change it as it will most probably break a lot of user code

dliubarskyi:
> @P required is set to false <-- slightly breaking

It is not **_slightly_** breaking, it will change all JSON schemas for tools, so that all paramerters will become optional by default, and many LLMs will definitely behave differently because of this

dliubarskyi:
> So I would expect jsr/jakarta validation as a user for more complex validation tasks.

This would be a nice feature, but I think this requires separate discussion and does not depend/influence our decisions for this issue

dliubarskyi:
> Primitive parameters are populated with default values

I'm not sure what exactly you mean by "default values", but I want to stress once again that I don't think it's a good idea to set primitive parameters to their default JVM values (e.g., `int` to `0`, `boolean` to `false`) because the LLM didn't provide a value. A few reasons:

1. `0`/`false` is rarely a sensible default.
2. This would be a breaking change. Currently, the LLM receives an error message and can either retry with updated parameters or notify the user about the issue, the `@Tool` method is not called. If we default to `0`/`false`, LLM hallucinations will go unnoticed, and the `@Tool` method will be invoked with questionable values.

What I propose instead:

1. When a primitive parameter is required and the LLM does not provide it, throw a `ToolArgumentsException` from `DefaultToolExecutor`. In the current 1.X version, this will mean that the AI Service invocation fails (the default behavior) or that `ToolArgumentsErrorHandler` handles it (if the user has explicitly configured one). In 2.0, the default `ToolArgumentsErrorHandler` will send a "missing parameter X" message back to the LLM, unless the user has configured a custom `ToolArgumentsErrorHandler`.
2. Add a warning to the [documentation](https://docs.langchain4j.dev/tutorials/tools) noting that although all parameters are required by default (as reflected in the generated JSON schema), we do not currently enforce this for non-primitive parameters. So if the LLM does not adhere to the schema, `null` will be passed for a required parameter. This is a known issue and will be fixed in 2.0: we will explicitly fail via `ToolArgumentsErrorHandler("missing parameter X")`, similar to how primitive parameters are handled.
3. Add a new `defaultValue` attribute to `@P`, so that users can explicitly specify a fallback value for parameters when the LLM does not provide one.

WDYT?

dliubarskyi:
Please check https://github.com/langchain4j/langchain4j/pull/5143

dliubarskyi:
Please check https://github.com/langchain4j/langchain4j/pull/5146 as well

sterlp:
> I'm not sure what exactly you mean by "default values", but I want to stress once again that I don't think it's a good idea to set primitive parameters to their default JVM values (e.g., `int` to `0`, `boolean` to `false`) because the LLM didn't provide a value. A few reasons:

I agree as soon I would use a primitive value I would expect it to be mandatory. Where 0 might make sense - for a boolean false is usually less usefull.

Nevertheless this is subjective. People may disagree here with us.

sterlp:
> 1. When a primitive parameter is required and the LLM does not provide it, throw a `ToolArgumentsException` from `DefaultToolExecutor`. In the current 1.X version, this will mean that the AI Service invocation fails (the default behavior) or that `ToolArgumentsErrorHandler` handles it (if the user has explicitly configured one). In 2.0, the default `ToolArgumentsErrorHandler` will send a "missing parameter X" message back to the LLM, unless the user has configured a custom `ToolArgumentsErrorHandler`.

Yes - this would be my expectation and it would be in sync with the doc. We get a proper error returned and handled by the langchain4j.

sterlp:
> 2\. Add a warning to the [documentation](https://docs.langchain4j.dev/tutorials/tools) noting that although all parameters are required by default (as reflected in the generated JSON schema), we do not currently enforce this for non-primitive parameters. So if the LLM does not adhere to the schema, `null` will be passed for a required parameter.

@dliubarskyi I would not do this. As if we add the check for primitive values and the error back to the LLM that a value is required - I would do the same for the required object types. As the current behavior, will cause NPE in the user code and forces us to define check like:

```java
    @Tool("Disk: Create/overwrite file. Creates parent dirs.")
    public String createDiskFile(@P(name = "filePath") String filePath, @P(name = "content") String content) {
        ArgsUtil.requireNonBlank(filePath, "filePath");
        ArgsUtil.requireNonBlank(content, "newContent");
```

In my opinion this would be inconsistent and a bug in itself

sterlp:
> 3\. Add a new `defaultValue` attribute to `@P`, so that users can explicitly specify a fallback value for parameters when the LLM does not provide one.

I like this one - and it should also work for primitive values than, right?

sterlp:
I would push all other changes than to the 2.0.0 adding more exception etc. to e.g. enforce required or default for primitive values etc.

PSchmitz-Valckenberg:
Since #5143 has been merged with a different approach (throwing a clear error 
instead of defaulting to zero/false), I'll close this PR to avoid duplication. 
Thanks for the detailed feedback @dliubarskyi — looking forward to the 
@P(defaultValue) feature in 2.0!

dliubarskyi:
> I like this one - and it should also work for primitive values than, right?

Yes, primitives will be supported as well

> looking forward to the
> @p(defaultValue) feature in 2.0!

[Support for default values](https://github.com/langchain4j/langchain4j/pull/5146) will be released in 1.15.0 at the end of this week

PR #5088, opened 2026-05-03
fix: fall back to exception class name when tool error has no message

### LINKED ISSUE #5087: [BUG] DefaultToolExecutor throws IllegalStateException instead of reporting tool error when exception has no message
## Describe the bug

When a tool method throws an exception with a `null` message (e.g. `throw new NullPointerException()` or `throw new IllegalStateException()`), `DefaultToolExecutor.executeWithContext` tries to build a `ToolExecutionResult` with a `null` `resultText`. The builder then throws an `IllegalStateException: One of resultText, resultTextSupplier, or resultContents must be provided` — masking the original tool failure entirely.

## Root cause

In [`DefaultToolExecutor.java`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java), both catch blocks for `InvocationTargetException` do:

```java
return ToolExecutionResult.builder()
        .isError(true)
        .resultText(e.getCause().getMessage())  // getMessage() can return null
        .build();
```

`Throwable.getMessage()` is specified to return `null` when no detail message was provided. `ToolExecutionResult.Builder` treats `null` as "not set" and throws when none of `resultText`, `resultTextSupplier`, or `resultContents` are provided.

## To Reproduce

```java
class BrokenTool {
    @Tool
    public String doSomething() {
        throw new NullPointerException(); // no message
    }
}
```

When this tool is invoked via `AiServices`, instead of receiving a tool error response, you get:

```
java.lang.IllegalStateException: One of resultText, resultTextSupplier, or resultContents must be provided
```

## Expected behavior

The tool failure should be reported back to the LLM as an error message (e.g. using the exception class name as fallback), not crash the executor with a secondary exception.

## Fix

Replace `e.getCause().getMessage()` with a null-safe fallback:

```java
String errorMessage = e.getCause().getMessage();
if (errorMessage == null) {
    errorMessage = e.getCause().getClass().getName();
}
return ToolExecutionResult.builder()
        .isError(true)
        .resultText(errorMessage)
        .build();
```

This affects both catch blocks at lines 115–119 and 125–129.

---

Happy to submit a PR for this if the approach looks good. @edeandrea @dliubarskyi — does the fallback to `getClass().getName()` seem reasonable, or would you prefer a different message format?

**LangChain4j version:** main (confirmed in current source)  
**Java version:** 21

PSchmitz-Valckenberg:
## Issue
Closes #5087

## Change
When a tool method throws an exception with a \`null\` message (e.g. \`throw new NullPointerException()\`), \`DefaultToolExecutor.executeWithContext\` passed \`null\` to \`ToolExecutionResult.builder().resultText()\`. The builder treated \`null\` as "not set" and threw:

\`\`\`
IllegalStateException: One of resultText, resultTextSupplier, or resultContents must be provided
\`\`\`

This secondary exception masked the original tool failure entirely.

Fix: extract error message resolution into a small \`errorMessage(Throwable)\` helper that falls back to \`cause.getClass().getName()\` when \`getMessage()\` returns \`null\`.

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
@@ -114,7 +114,7 @@ public class DefaultToolExecutor implements ToolExecutor {
                 } else {
                     return ToolExecutionResult.builder()
                             .isError(true)
-                            .resultText(e2.getCause().getMessage())
+                            .resultText(errorMessage(e2.getCause()))
                             .build();
                 }
             }
@@ -124,7 +124,7 @@ public class DefaultToolExecutor implements ToolExecutor {
             } else {
                 return ToolExecutionResult.builder()
                         .isError(true)
-                        .resultText(e.getCause().getMessage())
+                        .resultText(errorMessage(e.getCause()))
                         .build();
             }
         }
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@PSchmitz-Valckenberg thank you!

Issue #4909, opened 2026-04-13
[FEATURE] systemMessageProvider should give visbility into current in-flight user messge.


bitsofinfo:
The current systemMessageProvider contract just passes the memoryId and the current in-flight UserMessage is not available to it.

gaurav-code098:
Hi @bitsofinfo and @dliubarskyi! I'd love to take this on.

Since this touches the core AiServices builder, I want to ensure we align on the API design before I open a PR. To preserve backward compatibility, I propose we add an overloaded method to the AiServices builder that accepts a BiFunction providing both the memoryId and the UserMessage:

public AiServices<T> systemMessageProvider(BiFunction<Object, UserMessage, String> systemMessageProvider)

Alternatively, we could introduce a SystemMessageContext object if we anticipate needing to pass even more variables in the future.

@dliubarskyi - do you have a preference between the BiFunction approach or a Context object? I can wire this up as soon as we have a green light on the signature.

dliubarskyi:
@bitsofinfo currently, `systemMessageProvider` is called **_before_** we assembled `UserMessage` object from method arguments. But you could access method arguments (I guess one of them is a user message you are interested in), would that help?

@gaurav-code098 I would expose `InvocationContext` that contains `methodArguments`

gaurav-code098:
Thanks @dliubarskyi! That makes perfect sense regarding the order of operations. Getting access to the raw method arguments via an InvocationContext perfectly solves the routing use-case without disrupting the assembly order.

To implement this cleanly, I propose adding this overloaded method:
public AiServices<T> systemMessageProvider(Function<SystemMessageContext, String> systemMessageProvider)

Where SystemMessageContext (or InvocationContext if we want to reuse an existing generic one) is a simple record/class containing:

-     Object memoryId
-     Object[] methodArguments (or List<Object>)
-     (potentially the Method object itself if useful)

If this signature looks good to you, let me know  I'll start working to raise a PR.

dliubarskyi:
@gaurav-code098 please wait before @bitsofinfo confirms that this will work for their use case. I think this signature will be better: `AiServices systemMessageProvider(Function<InvocationContext, String> systemMessageProvider)`, `InvocationContext` already has all necessary info

DragonFSKY:
Hi, I’d like to work on this.

[...]

PR #5089, opened 2026-05-03
AI Services: add context-aware system message provider

DragonFSKY:
## Summary
- Add `systemMessageProviderWithContext(Function<InvocationContext, String>)` to expose the current AI Service invocation to system message providers.
- Keep the existing memoryId-based `systemMessageProvider(Function<Object, String>)` unchanged.
- Preserve the current invocation order: the provider runs before `UserMessage` is prepared, so callers can use `methodArguments()` without assembling `UserMessage` early.

Refs #4909

## Tests
- `./mvnw -pl langchain4j -am -Dtest=AiServicesContextAwareSystemMessageProviderTest#*,AiServicesTest#*,AiServicesSystemAndUserMessageConfigsTest#* -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -pl langchain4j spotless:check`

### DIFF
--- a/docs/docs/tutorials/ai-services.md
+++ b/docs/docs/tutorials/ai-services.md
@@ -138,6 +138,20 @@ Friend friend = AiServices.builder(Friend.class)
 ```
 As you can see, you can provide different system messages based on a chat memory ID (user or conversation).
 
+When you need access to the current AI Service invocation, use `systemMessageProviderWithContext`.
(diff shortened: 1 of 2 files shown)


DragonFSKY:
Closing this as it has been superseded by #5361, which introduced the same API on main. Thank you.

PR #5094, opened 2026-05-05
fix: return output printed by the GraalVM code execution engines

### LINKED ISSUE #4334: [BUG] GraalVM execution engines lose stdout/stderr output
**Describe the bug**

`GraalVmJavaScriptExecutionEngine` and `GraalVmPythonExecutionEngine` capture stdout/stderr to `ByteArrayOutputStream` but never read or return the captured output. Output from `console.log()`, `print()`, or runtime errors is lost.

**Log and Stack trace**

N/A

**To Reproduce**

```java
GraalVmJavaScriptExecutionEngine engine = new GraalVmJavaScriptExecutionEngine();
String result = engine.execute("console.log('hello'); 42");
// result is "42", but "hello" is lost
```

**Expected behavior**

The captured stdout/stderr should be included in the execution result, matching other code execution engines.

- LangChain4j version: (latest main)
- LLM(s) used: N/A
- Java version: 21

DragonFSKY:
## Issue

Fixes #4334

## Change

`GraalVmPythonExecutionEngine` and `GraalVmJavaScriptExecutionEngine` captured stdout and stderr into a `ByteArrayOutputStream` that was never read, so everything the executed code printed was thrown away. Both engines now return that output together with the value the code evaluated to.

Two related defects surfaced while fixing this:

- **Python code that does not end with an expression returned `<module '__main__'>`.** GraalPy evaluates such code (an assignment, an import, a bare `print(...)`) to the `__main__` module itself, and the engine turned that module into a string. So `print('hello')` returned the literal text `<module '__main__'>`. Used through `GraalVmPythonExecutionTool`, an LLM reads this as a failed tool call and keeps retrying.
- **Converting that module to a Java object trips an assertion inside GraalVM.** With assertions enabled — the default for Maven Surefire — `Value.as(Object.class)` on a module throws `PolyglotException: java.lang.AssertionError` instead of returning anything. Modules are now never converted.

Truffle writes its own diagnostics into the same streams, such as the multi-line "the polyglot engine uses a fallback runtime" performance warning. They are routed to a null log handler so they cannot leak into the returned result.

The `@P` descriptions of both tools now tell the model that the result may be returned **or** printed, which is what the engines accept. Changing that annotation is an API difference for revapi, so it is whitelisted in a module-level `revapi.json`.

## Breaking Changes

`execute(String)` returns a different string than before for code that prints something or that evaluates to nothing. Code that only evaluates to a value is unaffected.

| Executed code | Before | Now |
|---|---|---|
| `40 + 2` | `42` | `42` (unchanged) |
| `print('hello')` | `<module '__main__'>` | `Output:` <br> `hello` |
| `print('hello')` <br> `42` | `42` | `Output:` <br> `hello` <br> `Result:` <br> `42` |
| `x = 1` | `<module '__main__'>` | (empty string) |
| `console.log('hello');` | `null` | `Output:` <br> `hello` |
| `console.log('hello'); 42` | `42` | `Output:` <br> `hello` <br> `Result:` <br> `42` |
| `var x = 1;` | `null` | (empty string) |

If you read the returned string, expect the printed output in front of the value:

```java
String result = engine.execute("print('hello')\n42");
// Output:
// hello
// Result:
// 42
```

To keep getting only the value, let the code evaluate to it and print nothing — that case returns exactly what it returned before:

```java
String result = engine.execute("40 + 2"); // "42", unchanged
```

## Testing

- `GraalVmJavaScriptExecutionEngineTest` and `GraalVmPythonExecutionEngineTest`: 15 tests, all green
- `GraalVmPythonExecutionToolIT`: run 4 times against the OpenAI API, green every time, including the `verify(tool)` exactly-once assertion that the lost output used to break
- `mvn verify` on the module: revapi green

## General checklist

There **are** behaviour changes, described in the "Breaking Changes" section above. Everything the engines returned before was either the printed output being silently dropped, or the text `<module '__main__'>` / `null` in place of a result.

### DIFF
--- a/code-execution-engines/langchain4j-code-execution-engine-graalvm-polyglot/revapi.json
+++ b/code-execution-engines/langchain4j-code-execution-engine-graalvm-polyglot/revapi.json
@@ -1,0 +1,20 @@
+[
+  {
+    "extension": "revapi.differences",
+    "configuration": {
+      "ignore": true,
+      "differences": [
+        {
+          "code": "java.annotation.attributeValueChanged",
+          "new": "parameter java.lang.String dev.langchain4j.agent.tool.graalvm.GraalVmJavaScriptExecutionTool::executeJavaScriptCode(===java.lang.String===)",
+          "justification": "@P carries the natural-language description sent to the LLM, not a Java API contract; it now also tells the model that printing the result is accepted"
+        },
+        {
+          "code": "java.annotation.attributeValueChanged",
+          "new": "parameter java.lang.String dev.langchain4j.agent.tool.graalvm.GraalVmPythonExecutionTool::executePythonCode(===java.lang.String===)",
+          "justification": "@P carries the natural-language description sent to the LLM, not a Java API contract; it now also tells the model that printing the result is accepted"
+        }
+      ]
+    }
+  }
+]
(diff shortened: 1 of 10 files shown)


dliubarskyi (approved):
@DragonFSKY thanks!

PR #5120, opened 2026-05-06
fix: propagate Ollama streaming errors to handler onError

### LINKED ISSUE #5107: [BUG] Ollama streaming errors are ignored
Ollama returns a simple JSON when an error occurs during streaming (cf. https://docs.ollama.com/api/errors#errors-that-occur-while-streaming):
`{"error":"an error was encountered while running the model"}`

**Describe the bug**

Langchain4J tries to parse all server sent events as OllamaChatResponse, which doesn't correspond to the simple error format. See OllamaClient (line 185). As a result, the streaming chat handler never is informed that the chat response is done/failed.

**To Reproduce**

Hard to reproduce, as those errors occur seldomly. For example from ministral-3:3b I sometimes get:
`[ForkJoinPool.commonPool-worker-2] DEBUG dev.langchain4j.http.client.log.LoggingHttpClient - ServerSentEvent { event = null, data = "{"error":"invalid character '\\n' in string literal"}" }`

**Expected behavior**

The onError() method of the StreamingChatResponseHandler should be called with an appropriate Error/Exception.

- LangChain4j version: 1.14.0
- LLM(s) used: ministral-3:3b
- Java version: 21

**Additional context**

HariTechLab:
## Summary

Fixes #5107

When Ollama encounters an error during streaming, it returns a simple JSON error response:
```json
{"error":"an error was encountered while running the model"}
```
([Ollama docs: Errors that occur while streaming](https://docs.ollama.com/api/errors#errors-that-occur-while-streaming))

Currently, `OllamaChatResponse` and `CompletionResponse` use `@JsonIgnoreProperties(ignoreUnknown = true)`, which silently drops the `error` field. The parsed object ends up with all null fields, and the null message check in `OllamaClient` returns without ever calling `handler.onError()`. The streaming handler never learns the request failed.

## Changes

- Added `error` field to `OllamaChatResponse` and `CompletionResponse`
- Added error detection in `OllamaClient.streamingChat()` and `streamingCompletion()` — checks for error before processing the response
- Created `OllamaStreamingException` (follows the same pattern as `AnthropicStreamingException`)
- Added 2 unit tests using MockWebServer:
  - Error-only response — verifies `onError()` is called
  - Partial response then error — verifies partial content is received AND error is propagated

## Test plan

- [x] `mvn test -pl langchain4j-ollama -Dtest="OllamaStreamingErrorTest"` — 2/2 pass
- [x] `mvn test -pl langchain4j-ollama -Dtest="!*IT"` — 22/22 pass (all unit tests)
- [x] `mvn spotless:check -pl langchain4j-ollama` — formatting clean


langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
```diff
@@ -134,6 +134,13 @@ void streamingCompletion(CompletionRequest request, StreamingResponseHandler<Str
             public void onEvent(ServerSentEvent event) {
 
                 CompletionResponse completionResponse = fromJson(event.data(), CompletionResponse.class);
+
+                String error = completionResponse.getError();
+                if (error != null) {
+                    handler.onError(ExceptionMapper.DEFAULT.mapException(new OllamaStreamingException(error)));
```
dliubarskyi: shouldn't we call `onError` wrapped into `withLoggingExceptions`, like below?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
```diff
@@ -183,6 +190,13 @@ public void onEvent(ServerSentEvent event, ServerSentEventContext context) {
                 }
 
                 OllamaChatResponse ollamaChatResponse = fromJson(event.data(), OllamaChatResponse.class);
+
+                String error = ollamaChatResponse.getError();
+                if (error != null) {
+                    withLoggingExceptions(() -> handler.onError(new OllamaStreamingException(error)));
```
dliubarskyi: shouldn't we map the exception with `ExceptionMapper`, like above?

langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
HariTechLab: Good catch — both methods should use the same pattern as the existing transport-error handler (`withLoggingExceptions` + `ExceptionMapper`). Will update both.

HariTechLab:
@dliubarskyi — addressed the review feedback in the latest commits:
- Wrapped both `onError` calls with `withLoggingExceptions` + `ExceptionMapper` for consistency with the existing transport-error handler
- Made `OllamaStreamingException` public (matches `AnthropicStreamingException`)
- Renamed test to `OllamaStreamingErrorIT.java` (uses MockWebServer, follows the module's IT convention)

Also resolved a merge conflict in `CompletionResponse.java` introduced by #5157. PR is now mergeable.

Could you take another look when you get a chance? Would be great to land this before another conflict shows up on main.

dliubarskyi (approved):
@HariTechLab thank you!

PR #5123, opened 2026-05-06
fix(mistralai): null-guard MistralAiChatMessage.getContent() in aiMes…

### LINKED ISSUE #5121: [BUG] MistralAiMapper.aiMessageFrom NPE when content:null with tool_calls (vLLM / OpenAI-compatible servers)
**Describe the bug**

`MistralAiMapper.aiMessageFrom` throws `NullPointerException` when an OpenAI-compatible inference server (vLLM, llama.cpp, …) returns an assistant message with `content: null` and a non-empty `tool_calls` array. The mapper at [`MistralAiMapper.java#L171`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java#L171) calls `.stream()` on `aiMistralMessage.getContent()` without null-guarding, so any caller using `MistralAiChatModel` against such a server crashes the moment a tool is called.

**Log and Stack trace**

```
java.lang.NullPointerException: Cannot invoke "java.util.List.stream()" because the return value of
"dev.langchain4j.model.mistralai.internal.api.MistralAiChatMessage.getContent()" is null
    at dev.langchain4j.model.mistralai.internal.mapper.MistralAiMapper.aiMessageFrom(MistralAiMapper.java:157)
    at dev.langchain4j.model.mistralai.MistralAiChatModel.doChat(MistralAiChatModel.java:115)
    ...
```
 (Stack from langchain4j 1.12.2; the same code path is at line 171 on main.)

**To Reproduce**

```json
{
  "id": "chatcmpl-9ccb8be3370654bc",
  "model": "mistralai/Ministral-3-8B-Instruct-2512",
  "choices": [{
    "index": 0,
    "message": {
      "role": "assistant",
      "content": null,
      "tool_calls": [{
        "id": "chatcmpl-tool-b51c242ff1eacb66",
        "type": "function",
        "function": {
          "name": "verifier_date_incident",
          "arguments": "{\"date\": \"2026-05-06\"}"
        }
      }]
    },
    "finish_reason": "tool_calls"
  }],
  "usage": {"prompt_tokens": 4175, "completion_tokens": 23, "total_tokens": 4198}
}
```

Minimal Java reproducer mirroring the existing [`MistralAiChatModelReturnThinkingTest`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiChatModelReturnThinkingTest.java) pattern:

```java
MockHttpClient mock = MockHttpClient.thatAlwaysResponds(SuccessfulHttpResponse.builder()
        .statusCode(200).body(JSON_ABOVE).build());

ChatModel model = MistralAiChatModel.builder()
        .httpClientBuilder(new MockHttpClientBuilder(mock))
        .apiKey("x")
        .modelName("mistralai/Ministral-3-8B-Instruct-2512")
        .build();

model.chat(UserMessage.from("…"));   // → NullPointerException at MistralAiMapper.java:171
```

**Expected behavior**

The mapper should treat `content: null` as an empty list and return an `AiMessage` with the tool execution requests parsed and an empty `text`. That matches what `langchain4j-open-ai` does for the equivalent OpenAI-spec payload — and makes `MistralAiChatModel` interoperable with all OpenAI-compatible servers (vLLM, llama.cpp, Ollama in OpenAI mode, …).

- LangChain4j version: `1.12.2` (regression also confirmed on `main` as of 2026-05-06)
- LLM(s) used: `mistralai/Ministral-3-8B-Instruct-2512` served via vLLM
- Java version: `17`

**Additional context**

Regression history :
  - `1.0.0` – `1.10.0`: `aiMessageFrom` had an early-return on tool calls, and `getContent()` returned `String`. The null path was unreachable.
  - `1.11.0` (released 2026-02-04, multimodal / thinking-content refactor for Magistral): `getContent()` became `List<MistralAiMessageContent>`, the early-return was
  removed, and `.stream()` is now called unconditionally on a possibly-null list.
  - `1.11.0` – `1.12.2` and `main`: bug still present (verified 2026-05-06).

Encountered in production after upgrading from `1.0.0` to `1.12.2`. Mitigated at the application layer (rerouting traffic through `OpenAiChatModel`, which handles the same payload correctly) while preparing this report.

A draft PR with null-guard fix and a `MockHttpClient`-based unit test (matching the existing `MistralAiChatModelReturnThinkingTest` style) will be opened referencing this issue.

madaram:
## Issue

Closes #5121.

## Change

Minimal, behavior-preserving null-guard in `MistralAiMapper.aiMessageFrom`. OpenAI-compatible servers (vLLM, llama.cpp, …) return `content: null` on assistant messages that carry only `tool_calls`. The current mapper unconditionally calls `.stream()` on `aiMistralMessage.getContent()`, which NPEs on this spec-compliant payload.

The fix treats null content as an empty list — both the `text` and `thinking` extraction streams now operate on the local reference:

```java
List<MistralAiMessageContent> contents = aiMistralMessage.getContent();
if (contents == null) {
    contents = List.of();
}
```

No public API or behaviour change for callers: `text` is still `""` when there is no text content, `thinking` stays `null` when no thinking content. No new dependencies. Spotless applied.

Adds `MistralAiChatModelToolCallsTest` mirroring `MistralAiChatModelReturnThinkingTest` (uses `MockHttpClient`), with three scenarios:
1. `should_handle_response_with_only_tool_calls_and_null_content` — the regression case (vLLM-style payload, captured byte-for-byte from a real production response).
2. `should_handle_response_with_both_text_content_and_tool_calls` — sanity, both fields present.
3. `should_handle_response_with_text_only_and_no_tool_calls` — ensures the null-guard doesn't disturb the existing text-only path.

## General checklist

 ### Note on the test runs

Ran `mvn -fae clean test` from the repo root and a follow-up `mvn -fae clean test -pl '!embeddings/langchain4j-embeddings'`. Combined: **59 modules pass cleanly, 2 modules fail, 46 modules auto-skip** as transitive dependents of the failures.

**Both failures share a single root cause unrelated to this change**: `ai.djl.engine.EngineException: Failed to load Huggingface native library` (`Unexpected flavor:
  cpu` — DJL HuggingFace native binaries do not load on aarch64 / Apple Silicon). Affects:
  - `langchain4j-embeddings` directly
  - `langchain4j` main module (15 test classes touching `AllMiniLmL6V2QuantizedEmbeddingModel` static init)
  - `langchain4j-azure-ai-search` (`AzureAiSearchContentRetrieverTest` setup uses the same embedding model)

The module touched by this PR — `langchain4j-mistral-ai` — was tested independently: **21/21 unit tests green** (3 new + 18 pre-existing). CI will validate the rest.

### DIFF
--- a/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java
+++ b/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java
@@ -110,9 +110,8 @@ public class MistralAiMapper {
 
         if (message instanceof ToolExecutionResultMessage toolExecutionResultMessage) {
             if (!toolExecutionResultMessage.hasSingleText()) {
-                throw new UnsupportedFeatureException(
-                        "Mistral AI does not support non-text content in tool results. "
-                                + "Only text content is supported.");
+                throw new UnsupportedFeatureException("Mistral AI does not support non-text content in tool results. "
+                        + "Only text content is supported.");
             }
             return MistralAiChatMessage.builder()
                     .role(MistralAiRole.TOOL)
@@ -168,7 +167,14 @@ public class MistralAiMapper {
             toolExecutionRequests = toToolExecutionRequests(toolCalls);
         }
 
-        String text = aiMistralMessage.getContent().stream()
+        // OpenAI-compatible servers (e.g. vLLM) return content=null when the assistant
+        // emits only tool calls. Treat null content as empty to avoid an NPE.
+        List<MistralAiMessageContent> contents = aiMistralMessage.getContent();
+        if (contents == null) {
+            contents = List.of();
+        }
+
+        String text = contents.stream()
                 .filter(content -> "text".equals(content.getType()))
                 .map(MistralAiTextContent.class::cast)
                 .map(MistralAiTextContent::getText)
@@ -176,7 +182,7 @@ public class MistralAiMapper {
 
         String thinking = null;
         if (returnThinking) {
-            List<String> thinkingTexts = aiMistralMessage.getContent().stream()
+            List<String> thinkingTexts = contents.stream()
                     .filter(content -> "thinking".equals(content.getType()))
                     .map(MistralAiThinkingContent.class::cast)
                     .map(MistralAiThinkingContent::getThinking)
--- a/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java
+++ b/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java
@@ -167,8 +167,6 @@ public class MistralAiMapper {
             toolExecutionRequests = toToolExecutionRequests(toolCalls);
         }
 
-        // OpenAI-compatible servers (e.g. vLLM) return content=null when the assistant
-        // emits only tool calls. Treat null content as empty to avoid an NPE.
         List<MistralAiMessageContent> contents = aiMistralMessage.getContent();
         if (contents == null) {
             contents = List.of();
--- a/langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiChatModelToolCallsTest.java
+++ b/langchain4j-mistral-ai/src/test/java/dev/langchain4j/model/mistralai/MistralAiChatModelToolCallsTest.java
@@ -1,0 +1,194 @@
+package dev.langchain4j.model.mistralai;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import dev.langchain4j.agent.tool.ToolSpecification;
+import dev.langchain4j.data.message.UserMessage;
+import dev.langchain4j.http.client.MockHttpClient;
+import dev.langchain4j.http.client.MockHttpClientBuilder;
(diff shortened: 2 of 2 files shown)


langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java
```diff
@@ -168,15 +167,22 @@ public static AiMessage aiMessageFrom(MistralAiChatCompletionResponse response,
             toolExecutionRequests = toToolExecutionRequests(toolCalls);
         }
 
-        String text = aiMistralMessage.getContent().stream()
+        // OpenAI-compatible servers (e.g. vLLM) return content=null when the assistant
```
dliubarskyi:
```suggestion
```

langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/internal/mapper/MistralAiMapper.java
```diff
@@ -168,15 +167,22 @@ public static AiMessage aiMessageFrom(MistralAiChatCompletionResponse response,
...
+        // emits only tool calls. Treat null content as empty to avoid an NPE.
```
dliubarskyi:
```suggestion
```

dliubarskyi (approved):
@madaram thanks!

PR #5124, opened 2026-05-06
feat(streaming): add server tool execution callbacks

### LINKED ISSUE #5116: [FEATURE] Expose OpenAI Responses server tool streaming events
`OpenAiOfficialResponsesStreamingChatModel` supports OpenAI Responses server tools like `web_search`, but it does not seem to expose the live server-tool lifecycle events.

OpenAI emits streaming events such as:

- `response.web_search_call.in_progress`
- `response.web_search_call.searching`
- `response.web_search_call.completed`

Docs:
- https://platform.openai.com/docs/api-reference/responses-streaming
- https://platform.openai.com/docs/guides/tools-web-search?api-mode=responses

The OpenAI Java SDK also models these directly on `ResponseStreamEvent`:

```java
event.isWebSearchCallInProgress()
event.asWebSearchCallInProgress()

event.isWebSearchCallSearching()
event.asWebSearchCallSearching()

event.isWebSearchCallCompleted()
event.asWebSearchCallCompleted()

heyarny:
## Issue

Closes #5116

## General checklist


dliubarskyi (commented):
@heyarny thank you for your PR, but I am not sure this feature warrants adding 3 new callbacks to both streaming handler and token stream.

WDYT about adding a single `onRawEvent` callback to both? This would allow users to consume all events, even those that are not mapped to existing callback methods.

For model implementations that use `dev.langchain4j.http.client.HttpClient` abstraction, we could expose `ServerSentEvent` very easily, for other implementations we could expose an event `Object` (user will have to cast it and/or parse it manually).

heyarny:
@dliubarskyi  thanks for the review!

Yes, I see your point. Adding three lifecycle callbacks to both `StreamingChatResponseHandler` and `TokenStream` might be too specific for this use case.

The main thing I need is that provider/server-side streaming events are not swallowed, especially OpenAI Responses events like `response.web_search_call.searching` and `response.web_search_call.completed`, so applications can update the UI while a hosted tool is running.

Should I try to change the PR to something like `onRawEvent`, or do you already have a better solution in mind?

dliubarskyi:
Please change this PR to use the `onRawEvent` solution

heyarny:
@dliubarskyi what do you think?

dliubarskyi:
Hi @heyarny, thank you so much for this contribution and for tackling #5116 — the idea of exposing raw provider streaming events is a great one, and your PR is what got the ball rolling. 🙏

  While reviewing it I ended up reworking the approach fairly substantially, to the point where it no longer maps cleanly onto your branch, so I've opened a separate PR to carry it forward: #5589. The
  main differences:

  - **No duplication of already-exposed events.** The callback now fires *only* for events that aren't already delivered through a typed callback (`onPartialResponse`, `onPartialThinking`, `onPartialToolCall`,
  `onCompleteToolCall`, `onCompleteResponse`), so consumers don't receive the same event twice.
  - **Renamed `onRawEvent` → `onUnmappedRawEvent`.** This leaves room for a planned follow-up that also exposes the raw event behind *mapped* callbacks (e.g. `PartialToolCall.rawEvent()`), giving a clean
  mapped/unmapped split.
  - **Extended to all streaming providers** — OpenAI (Chat Completions + Responses), OpenAI-official (Chat Completions + Responses), Anthropic, Google AI Gemini, Google GenAI, Amazon Bedrock, Mistral, and Ollama
  — for consistent behavior across the board.
  - Added documentation and tests covering both the positive case and the no-duplication guarantee.

  Because of the scope of these changes I'm going to close this PR in favor of #5589, which closes #5116. This is absolutely not a reflection on your work — your PR was the foundation for it.

  Thanks again for the contribution and for your patience — I hope to see more PRs from you! 🙌

PR #5129, opened 2026-05-07
feat: add per-tool strict schema enforcement on ToolSpecification

### LINKED ISSUE #5128: [FEATURE] Per-tool strict schema enforcement on ToolSpecification

Both Anthropic and OpenAI support per-tool `strict` schema enforcement in their APIs — each tool object in the request payload has its own `strict` field. However, LangChain4j only exposes this as a model-level `strictTools` boolean (on `AnthropicChatModel`/`AnthropicStreamingChatModel` and `OpenAiChatModel`), which applies the same value to **every** tool in the request.

This becomes a hard blocker when using Anthropic's strict tools feature, which has a **limit of 20 strict tools per request**. If an agent sends more than 20 tools (which is common in agentic architectures with many tools and tasks), enabling `strictTools(true)` on the model causes Anthropic to reject the request.

There is no way to selectively mark only high-value tools as strict while leaving others non-strict in the same request.

Add a `Boolean strict` field to `ToolSpecification`:

```java
ToolSpecification.builder()
    .name("runSql")
    .description("Execute a SQL query")
    .parameters(schema)
    .strict(true)   // enforce strict for this tool
    .build();
```

Semantics:
- `null` (default) — defer to the model-level `strictTools` setting (backward compatible, no behavior change)
- `true` — enable strict enforcement for this tool, regardless of model-level setting
- `false` — disable strict enforcement for this tool, regardless of model-level setting

The per-tool value takes precedence over the model-level default. This aligns with how both the Anthropic and OpenAI APIs actually work — `strict` is a per-tool attribute in the wire format.

1. **Use `ToolSpecification.metadata()` to pass a `"strict"` key** — Works without core changes, but is stringly-typed with no compile-time safety, and requires each provider module to know about the convention independently.

2. **Put `strict` on `JsonObjectSchema` (the tool's parameters schema)** — Rejected because `strict` is not a JSON Schema concept. It's an API-level attribute on the tool object itself, not on the `input_schema`. The schema already has `additionalProperties` which is a *consequence* of strict mode, not a substitute for it. Mixing API-level concerns into the provider-agnostic schema class would be the wrong abstraction.

3. **Add a `strictToolNames` set to the model builder** — Would work but puts the responsibility in the wrong place. The tool author knows whether their tool's schema is strict-compatible; the model configuration shouldn't need to enumerate tool names.

**Additional context**

Anthropic strict tools limits (from [docs](https://docs.anthropic.com/en/docs/build-with-claude/structured-outputs)):
- Max **20 strict tools** per request
- Max **24 optional parameters** across all strict tool schemas combined
- Max **16 union-type parameters** across all strict schemas combined

pedrovieira:
Closes #5128

## Summary

- Adds a `Boolean strict` field to `ToolSpecification` enabling per-tool strict schema enforcement
- When `null` (default), defers to the model-level `strictTools` setting — fully backward compatible
- When `true`/`false`, overrides the model-level default for that specific tool
- Updates `AnthropicMapper` and `OpenAiUtils` to resolve per-tool strict before falling back to the model default
- Includes JSON serialization/deserialization support and tests for all three modules

## Motivation

Anthropic limits strict tools to **20 per request**. Agentic architectures that send 20+ tools in a single request cannot use `strictTools(true)` on the model without hitting this limit. Per-tool strict allows selectively enforcing strict on high-value tools while leaving others non-strict.

## Usage

```java
ToolSpecification.builder()
    .name("runSql")
    .description("Execute a SQL query")
    .parameters(schema)
    .strict(true)   // enforce strict for this tool
    .build();
```

## Test plan

- [x] `ToolSpecificationJsonTest` — round-trip for `strict` true, false, null; rejection of non-boolean
- [x] `AnthropicMapperTest` — per-tool true overrides model null, per-tool false overrides model true, per-tool null falls back to model
- [x] `OpenAiUtilsTest` — same three override/fallback scenarios

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
@@ -24,6 +24,7 @@ public class ToolSpecification {
     private final String description;
     private final JsonObjectSchema parameters;
     private final Map<String, Object> metadata;
+    private final Boolean strict;
 
     /**
      * Creates a {@link ToolSpecification} from a {@link Builder}.
@@ -35,6 +36,7 @@ public class ToolSpecification {
         this.description = builder.description;
         this.parameters = builder.parameters;
         this.metadata = copy(builder.metadata);
+        this.strict = builder.strict;
     }
 
     /**
@@ -74,6 +76,22 @@ public class ToolSpecification {
         return metadata;
     }
 
+    /**
+     * Returns whether this tool should use strict schema enforcement.
+     * <p>
+     * When {@code true}, the LLM provider will validate tool calls against this tool's schema server-side.
+     * When {@code false}, strict enforcement is explicitly disabled for this tool.
+     * When {@code null} (default), the model-level strict setting is used.
+     * <p>
+     * NOTE: Currently, per-tool strict is supported by the {@code langchain4j-anthropic}
+     * and {@code langchain4j-open-ai} modules.
+     *
+     * @return {@code true} to enable strict enforcement, {@code false} to disable, or {@code null} to use the model default.
+     */
+    public Boolean strict() {
+        return strict;
+    }
+
     @Override
     public boolean equals(Object another) {
         if (this == another) return true;
@@ -84,7 +102,8 @@ public class ToolSpecification {
         return Objects.equals(name, another.name)
                 && Objects.equals(description, another.description)
                 && Objects.equals(parameters, another.parameters)
-                && Objects.equals(metadata, another.metadata);
+                && Objects.equals(metadata, another.metadata)
+                && Objects.equals(strict, another.strict);
     }
 
     @Override
@@ -94,6 +113,7 @@ public class ToolSpecification {
         h += (h << 5) + Objects.hashCode(description);
         h += (h << 5) + Objects.hashCode(parameters);
         h += (h << 5) + Objects.hashCode(metadata);
+        h += (h << 5) + Objects.hashCode(strict);
         return h;
     }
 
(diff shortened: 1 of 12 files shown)


langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
```diff
@@ -24,6 +24,7 @@ public class ToolSpecification {
     private final String description;
     private final JsonObjectSchema parameters;
     private final Map<String, Object> metadata;
+    private final Boolean strict;
```
dliubarskyi, line 27:
```suggestion
    private final boolean strict;
```

Do we really need `Boolean` and model 3 states?
What if we use `boolean` with `false` by default? All tools in all providers are non-strict by default, so it should work

dliubarskyi (commented):
@pedrovieira thank you!

Please check that all modules and models that support `strictTools` are covered, it seems that OpenAI Official module (both chat and responses APIs) and OpenAI unofficial (responses APIs) are missing

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecification.java
pedrovieira: Thanks for the review @dliubarskyi !

You’re right that `boolean` with `false` default is simpler. The reason I went with `Boolean` was to support a 3-state semantic:
- `null` -> “defer to the model-level strictTools setting”
- `true` -> “this tool is strict”
-  `false` -> “this tool is not strict”. 

This way, if someone has strictTools(true) on the model, tools without an explicit per-tool value would still inherit that. With primitive boolean, there’s no way to distinguish “user explicitly set false” from “user didn’t set anything” 

What’s your preference?

dliubarskyi: You are right, makes sense to keep 3 options

dliubarskyi (approved):
@pedrovieira thank you!

PR #5149, opened 2026-05-08
fix: isolate per-agent event listeners when SPI factory returns singl…

### LINKED ISSUE #5137: [BUG] All agents see events from all other agents when using AiServiceListenerRegistrar through SPI
**Describe the bug**
In agentic mode, all sub agents are being notified of the response event of all other agents when using an AiServiceListenerRegistrarFactory through SPI.
The issue is that AgentInstance is now an AiServiceResponseReceivedListener, and the AiServiceContext always initializes its own local variable by calling AiServiceListenerRegistrar.newInstance() which eventually calls:
```java
    static AiServiceListenerRegistrar newInstance(boolean shouldThrowExceptionOnEventError) {
        var registrar = ServiceLoader.load(AiServiceListenerRegistrarFactory.class)
                .findFirst()
                .map(AiServiceListenerRegistrarFactory::get)
                .orElseGet(DefaultAiServiceListenerRegistrar::new);

        registrar.shouldThrowExceptionOnEventError(shouldThrowExceptionOnEventError);
        return registrar;
    }
```
When not using the SPI, a new instance of DefaultAiServiceListenerRegistrar is returned (every sub agent gets its own registrar), but when using SPI the get() method is called, which comes from the Supplier functional interface, and the javadoc states that "There is no requirement that a new or distinct result be returned each time the supplier is invoked".

**To Reproduce**
Here is a simple implementation of an AiServiceListenerRegistrarFactory. The Lc4jListenerRegistrarFactory is registered through the normal java SPI configuration file.

```java
public class Lc4jListenerRegistrarFactory implements AiServiceListenerRegistrarFactory {

	private static final Lc4jListenerRegistrar REGISTRAR = new Lc4jListenerRegistrar();
	
	@Override
	public AiServiceListenerRegistrar get() {
		return REGISTRAR;
	}
}

public class Lc4jListenerRegistrar extends DefaultAiServiceListenerRegistrar {

	private static final Lc4jThinkingListener LISTENER = new Lc4jResponseListener();
	
	public Lc4jListenerRegistrar() {
		register(LISTENER);
	}	
}

public class Lc4jResponseListener implements AiServiceResponseReceivedListener {
	
    @Override
    public void onEvent(AiServiceResponseReceivedEvent event) {
        // Do something here
    }
} 
```

**Expected behavior**
Subagents should only see their own events.
I am fine changing my implementation to return a new instance of the registrar, as the function seems to be called only in that one place. But I would prefer something more clearly defined to avoid surprises in the future.
Or maybe just a documentation change for now.

- LangChain4j version: 1.13.0, 1.13.1, 1.14
- LLM(s) used: Anthropic Claude
- Java version: 17

**Additional context**
In version 1.12.1 and older, there is no issue as AgentInstance is not an AiServiceResponseReceivedListener, a different mechanism (UserMessageRecorder) was used to track the messages.

PSchmitz-Valckenberg:
## Issue
Closes #5137

## Change
When `AiServiceListenerRegistrarFactory.get()` returns a singleton (a common and
reasonable implementation pattern), all agents shared the same
`AiServiceListenerRegistrar`. Every agent registered its listeners on that shared
instance, so each agent received events from every other agent.

**Fix:** `AiServiceListenerRegistrar.newInstance()` now always creates a fresh
`DefaultAiServiceListenerRegistrar` per agent context. If an SPI factory is
present, the factory-provided registrar is used as a **delegate** — events are
forwarded to it after notifying the per-agent local listeners. This means:

- Global listeners registered on the SPI registrar still receive all events from all agents (unchanged behaviour for this use case)
- Per-agent listener registrations are fully isolated — Agent A's listeners never receive Agent B's events
- SPI factory implementations may safely return a singleton

**Files changed:**
- `DefaultAiServiceListenerRegistrar` — new `delegate` constructor; `fireEvent` and `shouldThrowExceptionOnEventError` forward to delegate if present
- `AiServiceListenerRegistrar.newInstance()` — always wraps SPI result in a fresh instance
- `AiServiceListenerRegistrarFactory` — Javadoc updated to document the new delegate semantics

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/observability/api/AiServiceListenerRegistrar.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/observability/api/AiServiceListenerRegistrar.java
@@ -1,10 +1,10 @@
 package dev.langchain4j.observability.api;
 
-import java.util.Arrays;
-import java.util.ServiceLoader;
 import dev.langchain4j.observability.api.event.AiServiceEvent;
 import dev.langchain4j.observability.api.listener.AiServiceListener;
 import dev.langchain4j.spi.observability.AiServiceListenerRegistrarFactory;
+import java.util.Arrays;
+import java.util.ServiceLoader;
 
 /**
  * A registrar for registering {@link AiServiceListener}s.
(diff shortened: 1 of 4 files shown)


mariofusco:
See https://github.com/langchain4j/langchain4j/issues/5137#issuecomment-4420653767

PR #5151, opened 2026-05-09
feat(vertex-ai-gemini): support request labels for billing/cost attribution

### LINKED ISSUE #5136: [FEATURE] Vertex AI Gemini: support request labels for billing/cost attribution
## Summary                                                

VertexAiGeminiChatModel and VertexAiGeminiStreamingChatModel should expose the labels map that Vertex AI's generateContent request body accepts. This is the supported Google Cloud mechanism for tagging individual model invocations so they can be filtered/grouped on the GCP billing and cost reports.

Right now there is no public path through the langchain4j builder (or through the underlying GenerativeModel) to set this field, even though every Vertex AI request supports it.

##  Background                                                                                                                                                                                                                               
                                                         
  Per Google's official [Vertex AI REST reference](https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.endpoints/generateContent#request-body), the generateContent request body has a top-level `labels` field:

```                                                                                                                                                                                                         
 labels — map (key: string, value: string). Optional. The labels with user-defined metadata for the request. It is used for billing and reporting only.

Label keys and values can be no longer than 63 characters (Unicode codepoints) and can only contain lowercase letters, numeric characters, underscores and dashes. International characters are allowed. Label values are optional.    
Label keys must start with a letter.
```
Labels are read by GCP's billing pipeline and surface in the Cloud Console billing/cost reports and in BigQuery billing exports - which is exactly the integration point we need.
                                                                                                                                                                                                                                           The companion langchain4j-google-ai-gemini integration does not need this - the AI Studio API (generativelanguage.googleapis.com) does not have a labels field at all. This is specific to Vertex AI.                                    
   
##  Use case                                                                                                                                                                                                                                 
                                                         
  Multi-tenant SaaS that exposes Gemini-powered features to end customers. Every model call must be attributable to:                                                                                                                       
   
  - the tenant/customer it was made on behalf of (e.g. company_id)                                                                                                                                                                         
  - the end user that triggered it (e.g. user_id)        
  - the feature/agent that produced the call (e.g. agent=chatbot, agent=workflow)
                                                                                                                                                                                                                                           
  Without labels, all customers' usage collapses into a single bill against the GCP project. Operators can't answer "how much is customer X costing us?" without instrumenting an external accounting layer that duplicates what GCP  already provides natively.                                                                                                                                                                                                               
                                                                                                                                                                                                                                           
##  Proposed API                                           

  Add a labels(Map<String, String>) builder method on both Vertex AI Gemini chat model builders. When non-null, propagate the map into every `GenerateContentRequest` issued by the model via                                                
  `GenerateContentRequest.Builder.putAllLabels(...)`.
    
```                                                                                                                                                                                                                                       
  VertexAiGeminiChatModel model = VertexAiGeminiChatModel.builder()
      .project("my-gcp-project")
      .location("us-central1")
      .labels(Map.of(
          "company_id", "20096",                                                                                                                                                                                                           
          "user_id",    "12345",                         
          "feature",    "chatbot"))
      .build();                                                                                                                                                                                                                            
   ```

Same shape on `VertexAiGeminiStreamingChatModel.builder()`. 

**Please, verify that, I'll be happy to be wrong** if there is another approach to apply the labels!

Thanks,
Iliyan

ipeychev:
## Issue
Closes #5136

## Change

Adds a `labels(Map<String, String>)` builder method on both `VertexAiGeminiChatModel` and `VertexAiGeminiStreamingChatModel`. When set, the labels are placed on every `GenerateContentRequest` issued by the model via `GenerateContentRequest.Builder.putAllLabels(...)`.

Vertex AI's `generateContent` request body has a top-level `labels` map intended for billing and reporting, per the [REST spec](https://cloud.google.com/vertex-ai/generative-ai/docs/reference/rest/v1/projects.locations.endpoints/generateContent#request-body). Until now there was no path through the langchain4j Vertex AI Gemini chat model to set this field — the underlying Google `GenerativeModel` is `final` and builds the request privately.

### Implementation outline

When `labels` is empty (the default), the existing `GenerativeModel.generateContent(...)` / `generateContentStream(...)` SDK calls are used unchanged. When `labels` is non-empty, the model builds a `GenerateContentRequest` manually — mirroring the SDK's private `buildGenerateContentRequest` envelope and reading `getGenerationConfig`/`getSafetySettings`/`getTools`/`getToolConfig`/`getSystemInstruction` from the (already configured) `GenerativeModel` — and dispatches it through the SDK's public `PredictionServiceClient.generateContentCallable()` / `streamGenerateContentCallable()`. The fully-qualified resource name is reconstructed from the configured project/location/modelName, matching the SDK's private `getResourceName` rules.

This is the same wire shape that Google's official [`@google/genai` TypeScript SDK](https://github.com/googleapis/js-genai/blob/4383badd5b0057d00266d597f2e5dfbcdd69dccc/src/converters/_models_converters.ts#L1882-L1885) produces — TS lifts `config.labels` to the top of the request envelope; this PR does the equivalent in Java.

### Verified end-to-end

- 88 unit tests pass in `langchain4j-vertex-ai-gemini` (12 in `VertexAiGeminiChatModelBuilderTest`, 6 in `VertexAiGeminiStreamingChatModelBuilderTest`, including new tests covering: builder storage, default-empty behaviour, request-payload label injection, no-labels-when-empty, and the four `buildResourceName` branches).
- 1309 tests pass in `langchain4j-core` and `langchain4j` modules (no regressions).
- Real-world integration confirmed: a downstream consumer using this snapshot issued labelled `GenerateContent` and `StreamGenerateContent` requests against `aiplatform.googleapis.com`; the labels surfaced in **Cloud Billing → Reports → Group by → Labels** for the calling project.

### Backward compatibility

The new builder method is opt-in. All existing constructors and the deprecated constructor initialise `labels` to `Collections.emptyMap()`, so the new code path is never entered for callers that don't opt in.

## General checklist

### DIFF
--- a/langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiChatModel.java
+++ b/langchain4j-vertex-ai-gemini/src/main/java/dev/langchain4j/model/vertexai/gemini/VertexAiGeminiChatModel.java
@@ -16,6 +16,7 @@ import com.google.cloud.vertexai.VertexAI;
 import com.google.cloud.vertexai.api.Content;
 import com.google.cloud.vertexai.api.FunctionCall;
 import com.google.cloud.vertexai.api.FunctionCallingConfig;
+import com.google.cloud.vertexai.api.GenerateContentRequest;
 import com.google.cloud.vertexai.api.GenerateContentResponse;
 import com.google.cloud.vertexai.api.GenerationConfig;
 import com.google.cloud.vertexai.api.Part;
@@ -98,6 +99,8 @@ public class VertexAiGeminiChatModel implements ChatModel, Closeable {
 
     private final List<ChatModelListener> listeners;
     private final Set<Capability> supportedCapabilities;
+
+    private final Map<String, String> labels;
 
     public VertexAiGeminiChatModel(VertexAiGeminiChatModelBuilder builder) {
         ensureNotBlank(builder.modelName, "modelName");
@@ -213,6 +216,7 @@ public class VertexAiGeminiChatModel implements ChatModel, Closeable {
         this.logResponses = getOrDefault(builder.logResponses, false);
         this.listeners = copy(builder.listeners);
         this.supportedCapabilities = copy(builder.supportedCapabilities);
+        this.labels = copy(builder.labels);
     }
 
     /**
@@ -340,6 +344,7 @@ public class VertexAiGeminiChatModel implements ChatModel, Closeable {
 
         this.listeners = listeners == null ? emptyList() : new ArrayList<>(listeners);
         this.supportedCapabilities = copy(supportedCapabilities);
+        this.labels = Collections.emptyMap();
     }
 
     public VertexAiGeminiChatModel(GenerativeModel generativeModel, GenerationConfig generationConfig) {
@@ -365,6 +370,7 @@ public class VertexAiGeminiChatModel implements ChatModel, Closeable {
         this.logResponses = false;
         this.listeners = Collections.emptyList();
         this.supportedCapabilities = Set.of();
+        this.labels = Collections.emptyMap();
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


dliubarskyi (approved):
@ipeychev thank you!

ipeychev:
@dliubarskyi Thank you too! Do you know when the next release will be? I'm asking because we've drafted a workaround on our end in the meantime. If a release is coming soon, we can just wait.

dliubarskyi:
Should be released this week, most probably on Friday

glaforge:
@ipeychev do you think you could add similar support to the new Google GenAI Gemini module?
https://github.com/langchain4j/langchain4j/tree/main/langchain4j-google-genai
This is the new module using the Google GenAI Java SDK.

ipeychev:
@glaforge yes, I can, but I just checked and it seems you added them yourself meanwhile on this [PR](https://github.com/langchain4j/langchain4j/pull/5255). Isn't that the case?

glaforge:
Yes, I'm trying to cover as much ground as possible in this PR, to complement what we released in the last release of Friday.
Please tell me if you get to test it, or if you see missing things.

ipeychev:
@glaforge - here it is: https://github.com/langchain4j/langchain4j/pull/5255#issuecomment-4479793583. Thank you for the work!

PR #5158, opened 2026-05-11
Allow an agent to select among different ChatModels + Introduce voting agentic pattern

mariofusco:
## Issue

Closes #

## Change

This pull requests introduces two new features:

- It allows an agent to switch among different models based on the result of a selector function of the `AgenticScope`
- A new `Voting` agentic pattern that fans out all sub-agents in parallel and aggregates their results via a pluggable `VotingStrategy`. 

Regarding the new voting pattern, I understand that something quite similar could have been achieved already with the parallel workflow, plus an output function, but I received multiple feedback on the fact that it will be nice to have more explicit and use-case driven agentic patterns implementation and use them as building blocks for more complex agentic system, so I collected a few proposals and will add a few other patterns in the near future.

## General checklist


docs/docs/tutorials/agents.md
```diff
@@ -674,6 +674,28 @@ TokenStream tokenStream = novelCreator.writeStory("dragons and wizards", "young
...
+                scope -> {
+                    CritiqueResult critique = (CritiqueResult) scope.readState("critique");
+                    return critique != null && critique.score() > 7.8 ? 1 : 0;
+                },
+                baseModel(), enhancedModel())
+        .outputKey("story")
+        .build();
+```
+
+The selector function receives the current `AgenticScope` and returns an integer representing the index into the models array of the model to be used for the current invocation. In this example, a `StoryEditor` agent uses a `baseModel()` (index 0) by default, when the story to be edited has a critique score below 7.8, and then there is still a lot to improve and a cheaper model can be enough to do this job, switching to an `enhancedModel()` (index 1) when the score exceeds that threshold and then a better model could be necessary for the final refinements. The function is evaluated before every invocation, so the model can change across different invocations of the same agent.
```
dliubarskyi: nit: Do we have to restrict users to select only one of the pre-configured models? Wouldn't it be more flexible to have a method that just returns a `ChatModel` instance, so that users will be able to customize models on the fly?

dliubarskyi (dismissed):
@mariofusco thank you, nice feature!

docs/docs/tutorials/agents.md
mariofusco: Good point, let me think a bit about this. I will flag this pull request as a draft in the meanwhile.

mariofusco:
> nit: Do we have to restrict users to select only one of the pre-configured models? Wouldn't it be more flexible to have a method that just returns a ChatModel instance, so that users will be able to customize models on the fly?

This was not a nit at all, but a great suggestion to simplify the API and also allow to do same for streaming chat models in a straightforward way. Thanks a lot! 

Please give it a second look.

dliubarskyi (approved):
@mariofusco thank you!

PR #5160, opened 2026-05-11
feat(service): primitives for opt-in parallel tool dispatch and framework integration

unsignedint:
Adds primitives that quarkus-langchain4j needs to delegate tool dispatch back to langchain4j: so Quarkus can drop its own tool loops and turn on opt-in parallel tool execution (virtual threads or vert.x worker pool) without forking core behavior.

This PR attempts to address [@dliubarskyi's comment that execution should not be duplicated](https://github.com/quarkiverse/quarkus-langchain4j/pull/2418#issuecomment-4395252183) and ideally be concentrated upstream (langchain4j).

Disclosure: AI generated (GPT-5.5/Opus 4.7) with human-in-the-loop.

## Change

Adds core primitives and experimental integration hooks that allow framework integrations, especially for **quarkus-langchain4j**, to delegate more of the tool execution loop back to LangChain4j instead of duplicating that logic downstream. The goal is to support opt-in parallel tool execution while still allowing frameworks to manage threading, context propagation, per-call model selection, and framework-specific error handling.

Currently, on `main`, the non-streaming path executes tool calls serially by default and can fan out a batch when `executeToolsConcurrently(Executor)` is configured. The streaming path already supports per-tool executor scheduling, but the continuation after tool calls still runs on the model callback thread. That continuation includes result gathering, memory updates, dynamic-provider refresh, and the follow-up streaming request. Also, `maxSequentialToolsInvocations` limits the number of tool-calling responses, but there is no cap on how many tool calls a single response can contain.

This PR introduces `ToolBatchDispatcher` as a shared internal primitive for serial and concurrent tool batch execution. The non-streaming path now routes batch execution through it. Streaming uses the same dispatcher for inline or deferred dispatch, such as when no per-tool executor is configured or when `maxToolCallsPerResponse` is set. In the default streaming mode with a configured per-tool executor and no cap, tools are still scheduled eagerly from `onCompleteToolCall`; the hook then wraps result gathering, memory updates, dynamic-provider refresh, and follow-up streaming inference.

For streaming, this PR adds `StreamingToolDispatchHook`. The default hook runs inline on the callback thread, preserving the existing behavior. Frameworks can provide their own hook to move the continuation onto a managed worker thread and propagate framework context. In eager streaming mode, the hook does not move tool bodies that have already been scheduled; it wraps the downstream continuation.

This PR also adds `maxToolCallsPerResponse(int)` to cap the number of tool calls in a single model response. Non-streaming enforcement is strict and atomic. Streaming preserves eager scheduling by default, but when a positive cap is configured, dispatch is deferred until the full response is available so an over-limit response runs no tools.

The remaining hooks are aimed at framework integrations, specifically Quarkus compatibility:
- `forceToolChoiceAutoAfterFirstIteration(boolean)` preserves Quarkus behavior of forcing the first tool call, then allowing follow-up requests to either call another tool or produce the final answer
- `errorHandlerBypass(Predicate<Throwable>)` lets selected marker exceptions propagate instead of being converted into tool-result text
- `toolProviderRequestFactory(...)` lets integrations pass a custom `ToolProviderRequest` subtype to providers

_All public behavior is unchanged unless these new options are explicitly configured_. The default streaming hook is inline, the per-response cap is disabled by default, the tool-choice rewrite is off by default, the error-handler bypass predicate matches nothing by default, and the provider request factory builds the standard `ToolProviderRequest`.

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -3,7 +3,6 @@ package dev.langchain4j.service;
 import static dev.langchain4j.internal.Exceptions.runtime;
 import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 import static dev.langchain4j.service.AiServiceParamsUtil.chatRequestParameters;
-import static dev.langchain4j.service.tool.ToolService.refreshDynamicProviders;
 
 import dev.langchain4j.Internal;
 import dev.langchain4j.agent.tool.ReturnBehavior;
@@ -364,8 +363,8 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
 
             List<ChatMessage> messages = messagesToSend(invocationContext.chatMemoryId());
 
-            ToolServiceContext updatedToolContext =
-                    refreshDynamicProviders(toolServiceContext, messages, invocationContext);
+            ToolServiceContext updatedToolContext = context.toolService.refreshDynamicProvidersWithFactory(
+                    toolServiceContext, messages, invocationContext);
             updatedToolContext = ToolSearchService.addFoundTools(updatedToolContext, toolResults);
 
             ChatRequestParameters parameters =
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -16,8 +16,10 @@ import dev.langchain4j.guardrail.GuardrailRequestParams;
 import dev.langchain4j.guardrail.OutputGuardrailRequest;
 import dev.langchain4j.invocation.InvocationContext;
 import dev.langchain4j.memory.ChatMemory;
+import dev.langchain4j.model.chat.StreamingChatModel;
 import dev.langchain4j.model.chat.request.ChatRequest;
 import dev.langchain4j.model.chat.request.ChatRequestParameters;
+import dev.langchain4j.model.chat.request.ToolChoice;
 import dev.langchain4j.model.chat.response.ChatResponse;
 import dev.langchain4j.model.chat.response.CompleteToolCall;
 import dev.langchain4j.model.chat.response.PartialResponse;
@@ -34,8 +36,9 @@ import dev.langchain4j.observability.api.event.AiServiceRequestIssuedEvent;
 import dev.langchain4j.observability.api.event.AiServiceResponseReceivedEvent;
 import dev.langchain4j.observability.api.event.ToolExecutedEvent;
 import dev.langchain4j.service.tool.BeforeToolExecution;
+import dev.langchain4j.service.tool.StreamingToolDispatchHook;
 import dev.langchain4j.service.tool.ToolArgumentsErrorHandler;
-import dev.langchain4j.service.tool.ToolCallsLimitExceededException;
+import dev.langchain4j.service.tool.ToolBatchDispatcher;
 import dev.langchain4j.service.tool.ToolExecution;
 import dev.langchain4j.service.tool.ToolExecutionErrorHandler;
 import dev.langchain4j.service.tool.ToolExecutionResult;
@@ -46,13 +49,7 @@ import dev.langchain4j.service.tool.search.ToolSearchService;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import java.util.Queue;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
-import java.util.concurrent.Future;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import org.slf4j.Logger;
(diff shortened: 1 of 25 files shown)


dliubarskyi:
> Streaming path -> AiServiceStreamingResponseHandler.onCompleteResponse(...) runs tools inline on whatever thread the streaming model called the handler back on (typically an HTTP client event-loop thread). There is no way to switch threads, propagate framework context, or share the concurrent executor with the sync path.

This is not true, streming path supports concurrent tool execution, it uses `Executor toolExecutor`

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -226,16 +285,9 @@ public void onPartialToolCall(PartialToolCall partialToolCall, PartialToolCallCo
...
-                    () -> {
-                        ToolExecutionResult toolResult = execute(toolRequest);
-                        return new ToolRequestResult(toolRequest, toolResult);
-                    },
-                    toolExecutor);
-            toolExecutionFutures.add(future);
-        }
+        // No-op: tool calls are buffered on the AiMessage delivered via onCompleteResponse.
+        // This makes maxToolCallsPerResponse atomic — we cannot know the full count until the
+        // streamed tool-call section finishes, so dispatch is deferred to onCompleteResponse.
```
dliubarskyi: I don't think this is a good idea. I doubt many users set `maxToolCallsPerResponse`.
Firing tools as soon as possible with streming models should stay, at least when `maxToolCallsPerResponse` is not set specifically.

dliubarskyi (commented):
@unsignedint thanks, I have just started reviewing this, not sure it will be fast due to a complexity and number of changes.

unsignedint:
> > Streaming path -> AiServiceStreamingResponseHandler.onCompleteResponse(...) runs tools inline on whatever thread the streaming model called the handler back on (typically an HTTP client event-loop thread). There is no way to switch threads, propagate framework context, or share the concurrent executor with the sync path.
> 
> This is not true, streming path supports concurrent tool execution, it uses `Executor toolExecutor`

Yes, my original wording was too broad! The gap is much narrower and subtle: the streaming continuation around the tool batch still runs on the streaming callback thread (collecting tool results, updating memory, etc). The executor controls individual tool calls, but not that broader continuation or framework context propagation around it. I’ve updated the PR wording.

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
unsignedint: That makes sense. I thought it would be wise to defer in the name of purity - so both would go through `ToolBatchDispatcher`

I have reworked this completely and now I think it aligns with what you want:
- default uncapped streaming with executor uses eager scheduling (what you want)
- capped/no-executor streaming uses deferred via `ToolBatchDispatcher`

To be honest, I DO have a branch with a hybrid "best effort" for steaming w/ cap. (i.e. In `onCompleteToolCall(...)`, start tools immediately until the cap is reached. When tool call N + 1 arrives, stop scheduling more tools, fail the invocation, and optionally cancel unfinished futures. However, it adds code and complexity so I dropped it. If you disagree, let me know.

geoand:
Thanks for this! 

I find it hard to follow the description as it's all in bullet points. 
Could you please convert that to regular prose? 
Furthermore, could you add some info about how this aids advanced integrations, like the one in Quarkus LangChai4j?

unsignedint:
> Could you please convert that to regular prose?

will do (tomorrow)

> could you add some info about how this aids advanced integrations, like the one in Quarkus LangChai4j?

Yes! I can answer this now, but will merge into PR description.

quarkus-langchain4j is an advanced integration that adds a lot of framework-specific behavior around tool execution. This is currently implemented through custom execution-loop machinery in the Quarkus integration, with a lot of sync and streaming dispatch logic that overlaps with LangChain4j. Moreover, the Quarkus implementation has a significant limitation: it serializes tool batches. Complex AI agents often make many independent tool calls at once, so my primary motivation is to add concurrency to the tool execution process. (For example: make 5 web searches in parallel, today, they run sequentially in q-lc4j)

Based on earlier PR feedback, given the nature and size of the change, it makes sense to shift more of the core loop ownership upstream to LangChain4j, which already has the groundwork for concurrent tool execution. I believe that makes a lot of sense. But there are some gaps in LangChain4j that make it difficult to remove the Quarkus-owned execution loop while preserving the existing Quarkus LangChain4j feature set and runtime guarantees that users rely on.

That is where this PR comes in. It adds targeted hooks and reworks some larger pieces, especially around streaming dispatch, so advanced integrations can delegate the common inference-and-tools loop to LangChain4j while still plugging in their runtime-specific behavior. This lets Quarkus keep its existing guarantees and add concurrent/parallel tool calling without continuing to maintain a separate copy of the core dispatch loop (my other PR).

unsignedint:
@geoand to be clear, the quarkus-langchain4j special sauce we're supporting: vert.x context propagation (we need to manage context for memory, otel, etc), general event loop safety, supporting hooks for quarkus managed executors, error propagation, MCP client scoping, streaming callback/continuation handling etc.

unsignedint:
@geoand @dliubarskyi I have updated the PR description. Happy to answer any questions. Can rebase to resolve conflicts when ready. We're running this branch + quarkus-langchain4j changes right now, it's a nice improvement for our use case (also with [PR#5039](https://github.com/langchain4j/langchain4j/pull/5039). Would love to get it merged in :)

unsignedint:
Discovered an ordering bug on the new parallel-dispatch path: tools start eagerly from `onCompleteToolCall`, but the `IntermediateResponseEvent` carrying their inputs isn't emitted until `onCompleteResponse` on the stream thread, so a fast tool can surface its output before the corresponding input has been announced.

I've added a fix that keeps eager scheduling but holds stream-visible `onToolExecuted` delivery in an ordering buffer that drains once the intermediate response has gone out. One caveat: when a stream-level `beforeToolExecution` handler is registered, scheduling is deferred rather than eager. The handler can veto execution, so it has to run before the tool starts, which is incompatible with eager scheduling.

A new regression test covers these scenarios.

geoand:
Thanks!

> Can rebase to resolve conflicts when ready

What do you mean by this?

unsignedint:
> > Can rebase to resolve conflicts when ready
> 
> What do you mean by this?

Currently the branch has conflicts preventing a merge. I can rebase onto main now, if you'd like, but would prefer doing that post-review to reduce the window for new conflicts.

geoand:
You mean in this PR? If so, I don't see them

unsignedint:
Hey folks! @dliubarskyi @geoand just added a couple of commits from my colleague @Wal8800 that fixes cancellation behaviour. We have been running this branch on our production environment for a couple of weeks now. We would love to get this merged in :-)

geoand:
@dliubarskyi WDYT about this?

dliubarskyi:
Hey, sorry about the delay. I've been working on introducing async/reactive support in LC4j lately, and it touches the same areas (in particular, for tool calling). I am not 100% sure, but there is a big chance that after it lands, some changes suggested in this PR will not be needed at all, and some will be outdated. So I would not rush with this change right now and wait till async/reactive support in LC4j lands.

dliubarskyi:
But in general: sure, it is a good idea to reduce the duplication between LC4j and QLC4j, I am all for it

geoand:
Thanks for the context @dliubarskyi!

PR #5162, opened 2026-05-11
Propagate the AgenticScope when calling the planner agent of the supervisor agentic pattern

### LINKED ISSUE #5103: [BUG] NPE in AgentInvocationHandler when implementing a DefaultAiServiceListenerRegistrar
**Describe the bug**
In agentic mode, the first sub agent invoked by the planner will trigger an event that will produces as many NPEs as they are sub agents configured.
Every agent now implements AiServiceResponseReceivedListener and registers itself with the DefaultAiServiceListenerRegistrar. On the first invocation, the invocation context that is being built uses LangChain4jManaged.current() to set the managed parameter but that is null at the time. So when the AgentInvocationHandler.invoke method is called for each subagent (since they are now AiServiceResponseReceivedListener  and handle the onEvent message), the code will throw an NPE when trying to access the agentic scope from the managed parameters.

The changes from #4843 introduced the issue, but I don't believe they are the root cause.

**Log and Stack trace**
```
[http-nio-19980-exec-4] WARN dev.langchain4j.observability.api.DefaultAiServiceListenerRegistrar - An error occurred while firing event (dev.langchain4j.observability.event.DefaultAiServiceResponseReceivedEvent) to listener (jdk.proxy3.$Proxy110): Cannot invoke "java.util.Map.get(Object)" because the return value of "dev.langchain4j.invocation.InvocationContext.managedParameters()" is null
java.lang.NullPointerException: Cannot invoke "java.util.Map.get(Object)" because the return value of "dev.langchain4j.invocation.InvocationContext.managedParameters()" is null
	at dev.langchain4j.agentic.agent.AgentInvocationHandler.invoke(AgentInvocationHandler.java:82)
	at jdk.proxy3/jdk.proxy3.$Proxy110.onEvent(Unknown Source)
	at dev.langchain4j.observability.api.DefaultAiServiceListenerRegistrar$EventListeners.lambda$fireEvent$0(DefaultAiServiceListenerRegistrar.java:123)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at dev.langchain4j.observability.api.DefaultAiServiceListenerRegistrar$EventListeners.fireEvent(DefaultAiServiceListenerRegistrar.java:121)
	at dev.langchain4j.observability.api.DefaultAiServiceListenerRegistrar.lambda$fireEvent$1(DefaultAiServiceListenerRegistrar.java:66)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at dev.langchain4j.observability.api.DefaultAiServiceListenerRegistrar.fireEvent(DefaultAiServiceListenerRegistrar.java:66)
	at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:312)
	at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:155)
	at jdk.proxy3/jdk.proxy3.$Proxy116.plan(Unknown Source)
	at dev.langchain4j.agentic.supervisor.SupervisorPlanner.nextSubagent(SupervisorPlanner.java:134)
	at dev.langchain4j.agentic.supervisor.SupervisorPlanner.nextAction(SupervisorPlanner.java:93)
	at dev.langchain4j.agentic.planner.Planner.firstAction(Planner.java:36)
	at dev.langchain4j.agentic.internal.PlannerBasedInvocationHandler$PlannerLoop.loop(PlannerBasedInvocationHandler.java:364)
	at dev.langchain4j.agentic.internal.PlannerBasedInvocationHandler.executeAgentMethod(PlannerBasedInvocationHandler.java:203)
	at dev.langchain4j.agentic.internal.PlannerBasedInvocationHandler.invoke(PlannerBasedInvocationHandler.java:180)
	at jdk.proxy3/jdk.proxy3.$Proxy113.call(Unknown Source)
```

**To Reproduce**
Since I can only reproduce the issue when implementing and using an AiServiceListenerRegistrarFactory, here is the code for that part, the agents used in the agentic mode do not matter. The Lc4jListenerRegistrarFactory is registered through the normal java SPI configuration file.

```java
public class Lc4jListenerRegistrarFactory implements AiServiceListenerRegistrarFactory {

	private static final Lc4jListenerRegistrar REGISTRAR = new Lc4jListenerRegistrar();
	
	@Override
	public AiServiceListenerRegistrar get() {
		return REGISTRAR;
	}
}

public class Lc4jListenerRegistrar extends DefaultAiServiceListenerRegistrar {

	private static final Lc4jResponseListener LISTENER = new Lc4jResponseListener();
	
	public Lc4jListenerRegistrar() {
		register(LISTENER);
	}	
}

public class Lc4jResponseListener implements AiServiceResponseReceivedListener {
	
    @Override
    public void onEvent(AiServiceResponseReceivedEvent event) {
        // Do something here
    }
}
```

**Expected behavior**
I think there are 2 issues here: 
* the first is the NPE that should be fixed
* the second is every time a response event is fired, it is communicated to all sub agents and not only the one that is responsible for producing it. and that changes from the behavior of the UserMessageRecorder in 1.12 that was specific to each agent.

- LangChain4j version: 1.13.0, 1.13.1, 1.14.0
- LLM(s) used: Anthropic Claude
- Java version: 17

**Additional context**
The particular code that fails was added in 1.13.0, and it works fine in 1.12.2 and earlier versions

mariofusco:
## Issue

Closes #5103

## Change

The NPE only happens when using the supervisor pattern, because it internally uses a chat model on its own. I reproduced the bug in an isolated way and fixed the supervisor to always propagate the agentic scope even when calling its own internal AiService.

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/supervisor/SupervisorPlanner.java
@@ -3,10 +3,12 @@ package dev.langchain4j.agentic.supervisor;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import dev.langchain4j.agentic.internal.Context;
 import dev.langchain4j.agentic.planner.Action;
+import dev.langchain4j.invocation.LangChain4jManaged;
 import dev.langchain4j.agentic.planner.AgentArgument;
 import dev.langchain4j.agentic.planner.AgentInstance;
 import dev.langchain4j.agentic.planner.ChatMemoryAccessProvider;
@@ -131,7 +133,8 @@ public class SupervisorPlanner implements Planner, ChatMemoryAccessProvider {
                 ? SUPERVISOR_CONTEXT_PREFIX + "'" + agenticScope.readState(SUPERVISOR_CONTEXT_KEY, "") + "'."
                 : "";
 
-        AgentInvocation agentInvocation = planner(agenticScope).plan(agenticScope.memoryId(), agentsList, request, lastResponse, supervisorContext);
+        AgentInvocation agentInvocation = withAgenticScope(agenticScope,
+                () -> planner(agenticScope).plan(agenticScope.memoryId(), agentsList, request, lastResponse, supervisorContext));
         LOG.info("Agent Invocation: {}", agentInvocation);
 
         if (agentInvocation.getAgentName().equalsIgnoreCase("done")) {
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@mariofusco thank you!

PR #5165, opened 2026-05-12
fix(ollama): set content type for streaming requests

### LINKED ISSUE #5163: [BUG] langchain4j-ollama 1.14.0-beta24 sends HTTP body as ISO-8859-1, mangling non-ASCII characters into '?'
## Summary

  `langchain4j-ollama-spring-boot-starter:1.14.0-beta24` (transitively also affecting
  `langchain4j-ollama:1.14.0`) sets the outgoing HTTP `Content-Type` header to
  `text/plain;charset=ISO-8859-1` when sending the JSON request body to Ollama's
  `/api/chat`. As a result, every non-ASCII byte in the JSON payload (Chinese,
  Japanese, emoji, etc.) is replaced by `0x3f` ('?') on the wire, and the model
  sees garbage instead of the user's prompt.

  ## Environment

  - `langchain4j-ollama-spring-boot-starter`: `1.14.0-beta24`
  - `langchain4j-ollama`: `1.14.0`
  - `langchain4j-http-client-spring-restclient`: `1.14.0-beta24`
  - `langchain4j-reactor`: `1.14.0-beta24` (also reproduces without it)
  - Spring Boot: `3.5.13` (WebFlux)
  - JDK: `17.0.19` on Linux (WSL2), `file.encoding=UTF-8`, `defaultCharset=UTF-8`
  - Ollama: latest, model `qwen2.5:7b-instruct`

  ## Reproduction

  Minimal `application.yml`:
  ```yaml
  langchain4j:
    ollama:
      streaming-chat-model:
        base-url: http://localhost:11434
        model-name: qwen2.5:7b-instruct
        log-requests: true
        log-responses: true

```

  Send a Chinese prompt through StreamingChatModel:
  SystemMessage system = SystemMessage.from("你是中文 AI 助手，使用简体中文回答用户问题。");
  UserMessage user = UserMessage.from("你介绍一下自己");
  model.chat(List.of(system, user), handler);

  Expected

  The model replies with a normal self-introduction in Chinese.

  Actual

  The model replies with something like "您输入了一些表情符号 / 不清楚的内容..."
  because it never receives the original Chinese — all multi-byte UTF-8 sequences
  in the request body are mangled into ? on the wire.

  Evidence

  Putting a TCP sniffer between the Spring app and Ollama shows the actual bytes
  on the wire:

  POST /api/chat HTTP/1.1
  accept-encoding: gzip
  user-agent: ReactorNetty/1.2.16
  host: 127.0.0.1:11999
  accept: */*
  Content-Type: text/plain;charset=ISO-8859-1     <-- wrong content-type and charset
  Content-Length: 274

  {
    "model" : "qwen2.5:7b-instruct",
    "messages" : [ {
      "role" : "user",
      "content" : "???? AI ?????????????????\n\n???????"     <-- Chinese replaced by '?'
    } ],
    ...
  }

  The LoggingHttpClient debug output shows the intended (pre-serialization)
  body with Chinese intact — which is misleading, since the wire bytes are
  different.

  Sending the exact same JSON (with Content-Type: application/json) to Ollama
  directly via curl works perfectly and returns "我是来自阿里云的通义千问...".

  Likely cause

  The Ollama HTTP client code path appears to be constructing the request body
  using a default charset of ISO-8859-1 (Spring's default for text/plain when no
  explicit charset is given), instead of forcing UTF-8 (which the Ollama JSON API
  requires) and setting Content-Type: application/json.

  Workaround

  None at the application level — switching between langchain4j-reactor and
  langchain4j-http-client-spring-restclient does not help because both go
  through the same ISO-8859-1 body encoder configured by the ollama module.
  Downgrading or building a custom client around WebClient/RestClient are
  the only ways out at the moment.

By the way,question was found by Claude code.

DragonFSKY:
## Issue
Closes #5163

## Change
Adds `Content-Type: application/json` to Ollama streaming `/api/generate` and `/api/chat` requests.

The non-streaming Ollama request paths already set the JSON content type, but the streaming request paths did not. When those requests are sent through Spring RestClient, the JSON body is written as a `String`; without an explicit JSON content type, Spring can encode it as `text/plain;charset=ISO-8859-1`, which can replace non-ASCII characters with `?`.

This keeps the existing custom header behavior intact: custom headers are still applied after the default `Content-Type`, so users can override it if needed.

## Test Plan
- `./mvnw -pl langchain4j-ollama -DskipITs verify`

## General checklist

### DIFF
--- a/langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
+++ b/langchain4j-ollama/src/main/java/dev/langchain4j/model/ollama/OllamaClient.java
@@ -122,6 +122,7 @@ class OllamaClient {
         HttpRequest httpRequest = HttpRequest.builder()
                 .method(POST)
                 .url(baseUrl, "api/generate")
+                .addHeader("Content-Type", "application/json")
                 .addHeaders(buildRequestHeaders())
                 .body(toJson(request))
                 .build();
@@ -160,6 +161,7 @@ class OllamaClient {
         HttpRequest httpRequest = HttpRequest.builder()
                 .method(POST)
                 .url(baseUrl, "api/chat")
+                .addHeader("Content-Type", "application/json")
                 .addHeaders(buildRequestHeaders())
                 .body(toJson(ollamaChatRequest))
                 .build();
--- a/langchain4j-ollama/src/test/java/dev/langchain4j/model/ollama/OllamaStreamingContentTypeTest.java
+++ b/langchain4j-ollama/src/test/java/dev/langchain4j/model/ollama/OllamaStreamingContentTypeTest.java
@@ -1,0 +1,117 @@
+package dev.langchain4j.model.ollama;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import dev.langchain4j.http.client.MockHttpClient;
+import dev.langchain4j.http.client.MockHttpClientBuilder;
+import dev.langchain4j.model.StreamingResponseHandler;
+import dev.langchain4j.model.chat.response.ChatResponse;
+import dev.langchain4j.model.chat.response.StreamingChatResponseHandler;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.Test;
+
+class OllamaStreamingContentTypeTest {
+
+    @Test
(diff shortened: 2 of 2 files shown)


dliubarskyi (approved):
@DragonFSKY thank you!

PR #5167, opened 2026-05-12
fix: Improve schema generation of polymorphic types and nested collections

### LINKED ISSUE #5166: [BUG] ClassCastException in JsonSchemaElementUtils on fields with polymorphic subtypes
**Describe the bug**

`JsonSchemaElementUtils.getActualType` casts the single type argument of a `Collection<E>`-shaped field to `Class` unconditionally. 
The cast was introduced in c6f4277916 (Oct 2024) and stayed dormant until #5060 added `polymorphicSchemaFrom`. Before #5060 the walker treated polymorphic types as opaque, so subtype fields were never visited. After #5060 the walker descends into every concrete subtype and walks its declared fields, which surfaces the latent cast.

**Log and Stack trace**

```
Caused by: java.lang.ClassCastException: class sun.reflect.generics.reflectiveObjects.ParameterizedTypeImpl
cannot be cast to class java.lang.Class
at dev.langchain4j.internal.JsonSchemaElementUtils.getActualType(JsonSchemaElementUtils.java:327)
at dev.langchain4j.internal.JsonSchemaElementUtils.jsonSchemaElementFrom(JsonSchemaElementUtils.java:89)   
at dev.langchain4j.internal.JsonSchemaElementUtils.jsonObjectOrReferenceSchemaFrom(JsonSchemaElementUtils.java:269)
at dev.langchain4j.internal.JsonSchemaElementUtils.polymorphicSchemaFrom(JsonSchemaElementUtils.java:122)  
at dev.langchain4j.internal.JsonSchemaElementUtils.jsonSchemaElementFrom(JsonSchemaElementUtils.java:95)
at dev.langchain4j.internal.JsonSchemaElementUtils.jsonObjectOrReferenceSchemaFrom(JsonSchemaElementUtils.java:269)
at dev.langchain4j.internal.JsonSchemaElementUtils.jsonSchemaElementFrom(JsonSchemaElementUtils.java:98)
at dev.langchain4j.agent.tool.ToolSpecifications.toolSpecificationFrom(ToolSpecifications.java:112)
at dev.langchain4j.service.tool.ToolService.findTools(ToolService.java:170)
at dev.langchain4j.service.AiServices.tools(AiServices.java:477)
```

**To Reproduce**

Any `@Tool` parameter where a polymorphic field `@JsonSubTypes` has a concrete subtype declaring a `Collection<E>` field whose element type is itself parameterized. Minimal example:

```java
@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "kind")
@JsonSubTypes(@JsonSubTypes.Type(value = MySub.class, name = "sub"))
abstract class Base { }

class MySub extends Base {
    List<Box<String>> items;   // element type is ParameterizedType, not Class
}

class Box<T> { T value; }

class MyTools {
    @Tool void doStuff(Base input) { }
}

AiServices.builder(MyService.class).chatModel(model).tools(new MyTools()).build();
//  ClassCastException at construction time
```

**Expected behavior**

Schema generation should not crash on these field shapes. The element type can be reduced to a usable raw Class (Box, List, ...) so the walker can produce a sensible schema; it does not need to model the inner type argument.

- LangChain4j version: 1.14.x and the current 1.15.0-SNAPSHOT main
- LLM(s) used: any (fails before the model is ever called)
- Java version: 17+

**Additional context**

PR with fix and regression tests will follow.

rakow:
## Issue
Closes #5166 

## Change

- **`getActualType` no longer raw-casts the inner type argument.** It has been replaced with a `rawClassOf(Type)` helper that peels the type to its raw `Class`:
  - `Class<?>`-> returned as-is.
  - `ParameterizedType` -> its raw type.
  - `WildcardType` -> `rawClassOf(upperBounds[0])`.
  - `TypeVariable` -> `rawClassOf(bounds[0])`.
  - `GenericArrayType` -> `Object[].class`.
  - Anything else (or `null`) -> `Object.class`, so the walker degrades to an empty object schema instead of NPE-ing in downstream branches.
- **The `Collection` branch propagates the inner `Type` into the recursive call** (via a new `collectionElementType` helper), so nested generics like `List<List<X>>` resolve all the way down. Previously the call passed `null`, so anything below the first level lost its type information and crashed once the outer cast started succeeding.
- Behaviour for previously-working cases (`List<String>`, `List<MyClass>`, ...) is unchanged.
- The element type of `List<Foo<Bar>>` resolves to `Foo`; the inner generic argument is intentionally not modelled in the schema (the LLM contract for these never carried `Bar` either).

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonSchemaElementUtils.java
@@ -117,15 +117,15 @@ public class JsonSchemaElementUtils {
         }
 
         String reference = generateUUIDFrom(baseType.getName());
-        VisitedClassMetadata metadata =
-                new VisitedClassMetadata(JsonReferenceSchema.builder().reference(reference).build(), reference, false);
+        VisitedClassMetadata metadata = new VisitedClassMetadata(
+                JsonReferenceSchema.builder().reference(reference).build(), reference, false);
         visited.put(baseType, metadata);
 
         String discriminatorProperty = discriminatorPropertyName(baseType);
         List<JsonSchemaElement> options = new ArrayList<>();
         for (Class<?> subtype : findConcreteSubtypes(baseType)) {
-            JsonSchemaElement subtypeSchema = jsonObjectOrReferenceSchemaFrom(
-                    subtype, null, areSubFieldsRequiredByDefault, visited, false);
+            JsonSchemaElement subtypeSchema =
+                    jsonObjectOrReferenceSchemaFrom(subtype, null, areSubFieldsRequiredByDefault, visited, false);
             JsonSchemaElement withDiscriminator =
                     addDiscriminator(subtypeSchema, baseType, subtype, discriminatorProperty);
             options.add(withDiscriminator);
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@rakow thank you!

PR #5171, opened 2026-05-12
Allow a tool to declare its compensating action and invoke it when it fails

mariofusco:
## Issue

Closes #

## Change

Add a new `@ReverseTool` annotation allowing to declare the method reversing the action performed by a tool. When an `AiService` is configured as `transactional` (the name is very questionable let me know if you have any better idea about it) and a tool invocation fails, the reverse tool are invoked to roll back all the actions taken during the invocation of that `AiService`.

@dliubarskyi My plan is to use this feature also in the agentic module, but I will eventually do this with a follow up pull request, once we will find an agreement on this.

@jmartisk I don't know if this has some impact on MCP tools and if this feature could be extended also for them. Please let me know if you have any advice on this regard.

## General checklist


docs/docs/tutorials/tools.md
```diff
@@ -1536,6 +1536,98 @@ Assistant assistant = AiServices.builder(Assistant.class)
 As with the `ToolArgumentsErrorHandler`, there are two ways to handle errors in `ToolExecutionErrorHandler`:
 throw an exception or return a text message.
 
+### Transactional Tool Execution
+
+When an AI Service uses multiple tools to accomplish a task, a failure in one tool
+can leave the system in an inconsistent state — some tools have already executed
+successfully while others have not. For example, in a bank transfer the LLM might
+credit the recipient's account first and then fail to withdraw from the sender's
```
dliubarskyi, line 1643: Perhaps there are better examples? Bank transver is too naive

docs/docs/tutorials/tools.md
```diff
@@ -1536,6 +1536,98 @@ Assistant assistant = AiServices.builder(Assistant.class)
...
+```java
+class BankAccountService {
+
+    @Tool("credits money to a bank account")
+    void credit(String name, double amount) {
+        accounts.merge(name, amount, Double::sum);
+    }
+
+    @ReverseTool("credit")
+    void uncredit(String name, double amount) {
```
dliubarskyi, line 1671: In this example it is enough to have the name and amount, but in real world undoing an action might require some information about an actual tool execution. For example, `credit` tool could create an account transfer record and return transaction ID, in this case `uncredit` tool should have access to the `credit` return value to rollback the transaction. Perhaps this method should support accepting a `ToolExecution`?

langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
```diff
@@ -65,6 +65,10 @@ public DefaultToolExecutor(Object object, ToolExecutionRequest toolExecutionRequ
         this.propagateToolExecutionExceptions = false;
     }
 
+    public Method getOriginalMethod() {
```
dliubarskyi:
```suggestion
    public Method originalMethod() {
```

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -521,6 +569,18 @@ public ToolServiceResult executeInferenceAndToolsLoop(
...
     }
 
+    private void rollback(List<ToolExecutionRequest> successfulReversibleExecutions, InvocationContext invocationContext) {
+        for (int i = successfulReversibleExecutions.size() - 1; i >= 0; i--) {
+            ToolExecutionRequest request = successfulReversibleExecutions.get(i);
+            ToolExecutor reverseExecutor = reverseToolExecutors.get(request.name());
+            try {
+                reverseExecutor.executeWithContext(request, invocationContext);
+            } catch (Exception e) {
+                // best-effort: continue rolling back remaining actions
```
dliubarskyi: perhaps let's at least log the exception with WARN level?

langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
mariofusco: Fixed.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -464,10 +502,20 @@ public ToolServiceResult executeInferenceAndToolsLoop(
...
+                    successfulReversibleExecutions.add(request);
+                }
+
                 anyToolErrored = anyToolErrored || result.isError();
                 returnBehaviors.add(toolServiceContext.returnBehavior(request.name()));
             }
 
+            if (transactional && anyToolErrored) {
+                rollback(successfulReversibleExecutions, invocationContext);
+                throw runtime("Tool execution failed, rolled back %s previous tool action(s)",
```
dliubarskyi: If I read the code correctly, the `ChatMemory` will be in inconsistent state after the rollback. All tool calls that happened before and including the failing tool call will have corresponding `ToolExecutionResultMessage` in the `ChatMemory`. Remaining tool calls that did not happen (due to rollback) will not have any corresponding `ToolExecutionResultMessage`. This will basically block this AI service from being used again before manual `ChatMemory` fix (many LLM providers require that all tool calls have corresponding tool results). 

This behavior also does not sound "transactional".

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
```diff
@@ -2127,4 +2128,151 @@ void should_call_tool_with_jackson_annotated_polymorphic_argument(ChatModel chat
...
+
+        final Map<String, Double> accounts = new HashMap<>();
+
+        BankAccountService() {
+            accounts.put("Mario", 50.0);
+            accounts.put("Dmytro", 100.0);
+        }
+
+        @Tool("credits money to a bank account")
+        void credit(@P(name = "name", value = "account holder name") String name,
```
dliubarskyi:
```suggestion
        void credit(@P(name = "name", description = "account holder name") String name,
```
Would be nice to use `description` instead (everywhere)

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: I agree, do you have any suggestion to fix this? Could I simply remove all rolled back tools invocations from the chat memory?

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
```diff
@@ -2127,4 +2128,151 @@ void should_call_tool_with_jackson_annotated_polymorphic_argument(ChatModel chat
 
         verify(registry).registerShape(argThat(shape -> shape instanceof Circle circle && circle.radius == 4.0));
     }
+
+    static class BankAccountService {
```
dliubarskyi, line 2192: Could you please add more tests?
- verifying that rollback is in reverse
- multiple tool calls and the last one is failing
- multiple tool calls and the middle one is failing
- multiple tool calls and the first one is failing

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
dliubarskyi: I am not sure throwing an exception is a good "result" of a rollback. Yes, compensating action was triggered, but the agent/chat is unusable after this. Perhaps instead of an error we should call the LLM with all tool results? Tools that were not executed will have ToolExecutionResultMessage with something like "this tool was not called because of X". Tools that were successful but rolled back will also have this info in the tool result ("this result was rolled back") etc. This way ChatMemory is in a correct state, LLM has all the track record of what happened. LLM can either retry or just tell the user that something is wrong

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: Done.

docs/docs/tutorials/tools.md
mariofusco: Done.

docs/docs/tutorials/tools.md
mariofusco: Bank transfer is very first think that I think about when I hear the word "transaction" :) (still I'm not sure it is an appropriate term in this case, any better suggestion?), but I will check if I can find any better example.

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
mariofusco: I added tests to cover all those scenarios. This time I used a travel booking domain to implement them, do you prefer to use this domain, instead of the bank transfer one, also for documentation?

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: Claude seems to agree with your analysis, I will try to implement it.

```
  Option 1: Current behavior — throw exception after rollback
  - Pros: Simple, caller gets a clear programmatic signal
  - Cons: ChatMemory is corrupted, AI service is dead after this. Not truly "transactional" since it doesn't restore conversational state

  Option 2: Clean up ChatMemory, then throw
  - Remove the AiMessage (with tool requests) and any ToolExecutionResultMessages added during the failed round from ChatMemory before throwing
  - Pros: Memory is clean for subsequent calls, caller still gets the exception
  - Cons: LLM loses all context about what happened, may retry the exact same failing sequence in the next call. Removing messages from memory is fragile (especially with concurrent access or custom ChatMemory implementations)
  
  Option 3: Send informative results back to the LLM instead of throwing
  - Don't throw. Instead, produce a ToolExecutionResultMessage for every tool request:
    - Failed tool: normal error result (already exists)
    - Succeeded but rolled back: "Tool 'credit' was executed and then rolled back due to failure of tool 'withdraw'"
    - Never executed (skipped due to early stop): "Tool 'rentCar' was not executed because tool 'bookHotel' failed"
  - Pros: ChatMemory stays fully consistent, LLM has complete context and can decide what to do (retry differently, inform user), AI service remains usable. This is the most "agentic" approach
  - Cons: More complex. The caller doesn't get a programmatic signal that rollback happened (though the LLM's text response will likely mention it). The LLM could also misinterpret the rollback messages
  
  Option 4: Hybrid — send results to LLM + expose rollback info to the caller
  - Same as Option 3 but also record rollback details in the ToolServiceResult / Result<T> so the caller can detect that a rollback occurred without an exception
  - Pros: Best of both worlds — memory is clean, LLM has context, caller has programmatic access to rollback info
  - Cons: Most complex, touches more surface area

  ---
  My recommendation is Option 3. It aligns with what Dmytro suggested, keeps the conversation flowing naturally, and matches how non-transactional mode already works (errors go back to the LLM rather than blowing up). The key insight is that the LLM is better positioned than the caller to decide what to do
   after a rollback — it has the full conversation context.

```

mariofusco: Done (I think, please double check).

mariofusco:
@dliubarskyi I think I addressed all your suggestion, please give it a second look when you will have time.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -246,6 +256,59 @@ public static List<AiServiceTool> findTools(Object objectWithTools) {
...
+            Object objectWithTools, Map<String, ToolExecutor> toolExecutors) {
+        Map<String, BiConsumer<ToolExecution, InvocationContext>> reverseExecutors = new HashMap<>();
+        for (Method method : objectWithTools.getClass().getDeclaredMethods()) {
+            ReverseTool reverseTool = method.getAnnotation(ReverseTool.class);
+            if (reverseTool != null) {
+                String toolName = reverseTool.value();
+                ToolExecutor toolExecutor = toolExecutors.get(toolName);
+                if (!(toolExecutor instanceof DefaultToolExecutor)) {
+                    throw illegalConfiguration(
+                            "@ReverseTool(\"%s\") on method '%s.%s' references a non-existent tool",
```
dliubarskyi: Is it really a non-existent tool? Maybe it has a different type of executor (e.g., MCP). Worth refining this error message and noting in a javadco and docs that only `@Tool`-annotated tools are supported, programmatically and dynamically defined tools are not supported

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
```diff
@@ -521,6 +610,19 @@ public ToolServiceResult executeInferenceAndToolsLoop(
                 .build();
     }
 
+    private void rollback(List<ToolExecution> successfulReversibleExecutions, InvocationContext invocationContext) {
+        for (int i = successfulReversibleExecutions.size() - 1; i >= 0; i--) {
```
dliubarskyi: worth mentioning in the doc that rollbacks are running sequentially, even when tool execution is in parallel

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReverseTool.java
```diff
@@ -0,0 +1,33 @@
...
+ * all previously executed tools' reverse actions are called in reverse order
+ * to undo their effects.
+ * The annotated method must have the same parameter types as the tool it reverses.
+ *
+ * @since 1.15.0
+ */
+@Experimental
+@Retention(RUNTIME)
+@Target({METHOD})
+public @interface ReverseTool {
```
dliubarskyi:
```suggestion
public @interface CompensatingTool {
```

This seems to be more correct term

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -1218,6 +1218,19 @@ public AiServices<T> storeRetrievedContentInChatMemory(boolean storeRetrievedCon
...
 
+    /**
+     * Configures the AI Service as transactional.
+     * When enabled, if any tool execution fails, all previously executed tools'
+     * reverse actions (declared via {@link dev.langchain4j.agent.tool.ReverseTool}) are called
+     * in reverse order to compensate/rollback the effects of the successful tool calls.
+     *
+     * @return builder
+     */
+    public AiServices<T> transactional() {
```
```suggestion
    public AiServices<T> .compensateOnToolErrors(boolean compensateOnToolErrors) {
```
Transactional is indeed misleading, I would use "compensation" term here. And maybe worth having a boolean parameter so that it can be enabled/disabled on condition more easily

jmartisk:
I'm not sure about this TBH, without an actual integration with a transaction manager, this feels like re-inventing the world of transactions, but only a small subset and it doesn't feel very safe. I can imagine that writing the compensation tools to work properly in all cases will be quite difficult in the real world. Did you consider proper integration with JTA?

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReverseTool.java
dliubarskyi: I was also wondering if compensating tool should reference actual tool or vice versa, and here is what CLaude suggests:

```
Direction matters more than it looks. Forward → Reverse is the better choice in most dimensions, with one real caveat. Here's the comparison:

  Current: Reverse → Forward (compensator references tool)

  @Tool("credits money")
  void credit(String name, double amount) { ... }

  @ReverseTool("credit")          // points back to the tool
  void uncredit(String name, double amount) { ... }

  Alternative: Forward → Reverse (tool references compensator)

  @Tool(value = "credits money", compensatedBy = "uncredit")
  void credit(String name, double amount) { ... }

  void uncredit(String name, double amount) { ... }   // plain method

  Comparison

  Dimension: Reading a tool
  Reverse → Forward (current): No hint it's reversible — you have to scan the class to find a paired @ReverseTool
  Forward → Reverse: Compensability is visible right at the @Tool declaration — answers "is this rollback-safe?" instantly
  ────────────────────────────────────────
  Dimension: Mental model
  Reverse → Forward (current): Compensator is a first-class thing that "knows what it reverses"
  Forward → Reverse: Compensation is metadata about the forward action — matches saga literature, where the forward step declares its compensation
  ────────────────────────────────────────
  Dimension: Common question
  Reverse → Forward (current): Optimizes for "which tool does this compensation belong to?" — rarely asked
  Forward → Reverse: Optimizes for "what's the rollback for this tool?" — the question you actually have when auditing
  ────────────────────────────────────────
  Dimension: Deleting compensation
  Reverse → Forward (current): Delete the whole reverse method and remember it had @ReverseTool
  Forward → Reverse: Delete the compensatedBy = "..." attribute. Reverse method can stay or be deleted independently
  ────────────────────────────────────────
  Dimension: Typo / refactor safety
  Reverse → Forward (current): Caught at build time either way (with validation)
  Forward → Reverse: Same
  ────────────────────────────────────────
  Dimension: IDE "unused method" warning
  Reverse → Forward (current): ✅ Compensator has an annotation — IDE knows it's reflectively invoked
  Forward → Reverse: ❌ Plain uncredit() looks unused to IntelliJ/SpotBugs. Real friction.
  ────────────────────────────────────────
  Dimension: @Tool annotation surface
  Reverse → Forward (current): Stays minimal
  Forward → Reverse: Gains a compensatedBy attribute (clutter — but it's optional)

  My take

  Forward → Reverse wins on the dimensions that matter day-to-day — readability when scanning a tools class, alignment with saga semantics ("the forward action owns its
  compensation"), and matching how reviewers reason about reversibility.

  The IDE-unused-method warning is the one real downside. Two ways to neutralize it:

  1. Document that compensator methods should be package-private or public and add @SuppressWarnings("unused") if your IDE is strict.
  2. Keep a @Compensate annotation as a pure marker with no value — its job is just to tell IDEs/linters "called reflectively." The actual pairing still lives on @Tool:

  @Tool(value = "credits money", compensatedBy = "uncredit")
  void credit(...) { ... }

  @Compensate                           // marker only, no value
  void uncredit(...) { ... }

  2. Slight redundancy, but cleaner than @SuppressWarnings and lets static analysis recognize the role of the method. Many Java frameworks do this (e.g. JUnit's @ParameterizedTest
   is partly marker, partly config).

  Bonus: support both, prefer forward

  If you want to be friendly to users coming from the current PR design, you could accept both directions and resolve conflicts deterministically:

  - If @Tool(compensatedBy = "x") is set → use it.
  - Else if a method has @ReverseTool("foo") matching this tool → use it.
  - If both point at the same compensator → fine. If they disagree → fail at build time with a clear error.

  But for a v1 ship, I'd just pick one. Forward → Reverse with a marker @Compensate annotation is the design I'd land on — it puts compensability metadata where readers look for
  it, matches saga semantics, and keeps IDE tooling happy.
```

jmartisk:
> @jmartisk I don't know if this has some impact on MCP tools and if this feature could be extended also for them. Please let me know if you have any advice on this regard.

That will be hard unless MCP adds some native support for this. Other than that, the only way forward with MCP probably is requiring that the MCP server has the compensation tools and that the agent is instructed (via system message or something) to call them if one tool fails. MCP doesn't allow portably marking which tools are the reverse of which tool (and how exactly the reverse should be called)...

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReverseTool.java
mariofusco: I considered this as well, but at the end I decided to make the compensating method (probably also calling this thing `tool` is misleading?) referencing the tool, because the tool has a name (that by default is the method name, but anyway it is a logical name that the user can eventually configure), while to reference the compensating method from the tool would be possible only via reflection through the method name.

mariofusco:
> I'm not sure about this TBH, without an actual integration with a transaction manager, this feels like re-inventing the world of transactions, but only a small subset and it doesn't feel very safe. I can imagine that writing the compensation tools to work properly in all cases will be quite difficult in the real world. Did you consider proper integration with JTA?

I believe that the name "transaction" was quite misleading and indeed @dliubarskyi suggested to change it in `CompensatingTool` which describes much better what actually it is. Probably even calling this a tool is misleading at this point because it is not a tool at all. I'd rather call this `CompensatingAction`.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ReverseTool.java
mariofusco: As I wrote somewhere else this is not even a tool, so even that name is misleading. I renamed it `CompensateFor` so now the annotation reads more nicely imo, something like `@CompensateFor("credit")`.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
mariofusco: Done.

langchain4j/src/main/java/dev/langchain4j/service/tool/ToolService.java
mariofusco: Fixed.

mariofusco:
@dliubarskyi Thanks a lot for your suggestions, I should have addressed, or at least replied, to all of them. Please give it another look when you have time.

dliubarskyi:
@mariofusco sorry for the delay!
Do you still want to go ahead with this? Do you have a real use case for this feature?

mariofusco:
> @mariofusco sorry for the delay! Do you still want to go ahead with this? Do you have a real use case for this feature?

Yes, I still would like to go ahead with this feature. My use case is that I would like to give the possibility to an agent, or a whole agentic system, to execute atomically and rollback its side-effects in case of failure. Koog has a similar feature from where I took inspiration https://api.koog.ai/agents/agents-features/agents-features-snapshot/ai.koog.agents.snapshot.feature/-rollback-tool-registry/index.html

Do you see any blocker for this?

dliubarskyi (commented):
🔴 Correctness — double compensation across round trips (untested)

compensableExecutions is allocated once before the while loop (line ~507) and never cleared after a rollback(). Since the loop continues after a failure (the LLM is meant to retry), the same already-undone
  executions stay in the list and get compensated again on a later failure:

- Round 1: credit succeeds → compensableExecutions = [credit]
- Round 2: withdraw fails → rollback() runs uncredit. List still [credit].
- LLM retries → Round 3: credit succeeds again → [credit, credit']
- Round 4: withdraw fails again → rollback undoes credit' and credit a second time.

The existing tests don't catch this because no test triggers a second failure after a retry (should_rollback...when_transactional spans two round trips but fails only once). Suggested fix:
  compensableExecutions.clear() at the end of the rollback block (the actions it holds have all been undone).

🟠 Design — cross-round-trip rollbacks are invisible to the LLM

The informative "...was rolled back due to failure of tool 'X'" message is only written for requests in the current batch (loop at lines 566-575). But compensableExecutions accumulates across round trips, so a
  tool that succeeded in an earlier round trip and is undone now leaves its original success message in ChatMemory unchanged. The LLM's memory then says "credit succeeded" even though credit was reversed —
  directly at odds with the doc's stated goal that this "keeps the ChatMemory consistent." All the informative-message tests use a single batch, so this gap is untested.

Worth deciding explicitly: is compensation scoped to a single round trip, or the whole AiService invocation? The PR description says the latter ("all the actions taken during the invocation"), which is what
  makes both this and the double-compensation issue surface. If single-round-trip is acceptable, resetting the list each iteration resolves both at once and is much simpler.

🟡 Minor / discussion

- Validation runs unconditionally. findCompensatingActions is called from tools() regardless of compensateOnToolErrors, so a mismatched @CompensateFor throws IllegalConfigurationException even for users who
  never enable the feature. 
- getDeclaredMethods() (line 292) won't discover @CompensateFor methods inherited from a superclass. Check whether this matches how findTools resolves @Tool methods, for consistency.
- setAccessible(true) is invoked on every compensation call inside the ToolExecution-parameter lambda; hoist it to setup time.

mariofusco:
@dliubarskyi Thanks a lot for the very insightful review. I should have addressed all your comments, with a different commit for each of them. Please give it a second look.

dliubarskyi (approved):
@mariofusco thank you!

PR #5209, opened 2026-05-13
fix: skip empty tool_calls sentinel chunk in OpenAI streaming accumulator

Ivan97:
## Summary
Some OpenAI-compatible providers (observed with `deepseek-v4-flash`) emit a trailing sentinel chunk after all real `tool_calls` have streamed:

```json
{"index": 0, "id": "", "type": "function", "function": {"arguments": null}}
```

It carries an empty id, no `function.name`, and null `function.arguments`. When its index does not match a previously-seen tool call (e.g., arrives at a fresh `index`), the accumulator's `computeIfAbsent` creates an empty builder that surfaces as a ghost `ToolExecutionRequest { id = "", name = "", arguments = "" }` in the final result. Downstream code in `OpenAiUtils#toOpenAiMessage` then substitutes `"{}"` for the blank arguments when echoing the assistant message back, producing the well-known `ToolExecutionRequest { id = null, name = null, arguments = "{}" }` symptom in user-facing tool-execution chains.

The fix adds an `isSentinel(toolCall)` guard at the top of the `tool_calls` accumulation loop in `OpenAiStreamingResponseBuilder` that treats end-of-stream marker frames as no-ops (skipped via `continue`) before `computeIfAbsent` is invoked. Legitimate header chunks (id+name set) and argument-fragment chunks (non-null arguments) are unaffected because they have at least one of `id`, `function.name`, or non-null `function.arguments` populated.

## Test plan
- [x] New `should_ignore_trailing_sentinel_chunk_from_deepseek_v4_flash` exercises the exact stream pattern: header → argument fragments → sentinel — asserts exactly 1 `ToolExecutionRequest` with the full reconstructed arguments.
- [x] New `should_not_produce_ghost_for_orphan_sentinel_chunk` exercises a sentinel arriving at a fresh `index` with no prior data — asserts no ghost is produced. Without the fix this test produces `ToolExecutionRequest { id = "", name = "", arguments = "" }`.
- [x] All 138 existing unit tests in `langchain4j-open-ai` continue to pass.

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingResponseBuilder.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingResponseBuilder.java
@@ -164,6 +164,14 @@ public class OpenAiStreamingResponseBuilder {
 
         if (delta.toolCalls() != null) {
             for (ToolCall toolCall : delta.toolCalls()) {
+                // Skip end-of-tool-calls sentinel frames that carry no meaningful data
+                // (observed with deepseek-v4-flash on the OpenAI-compatible streaming API:
+                // a trailing chunk with empty id, no function.name and null function.arguments
+                // is emitted after all real tool_calls have streamed).
+                if (isSentinel(toolCall)) {
+                    continue;
+                }
+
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@Ivan97 thank you!

PR #5243, opened 2026-05-15
Introduce Blackboard agentic pattern

mariofusco:
## Issue

Closes #

## Change

This pull request introduces the new Blackboard agentic pattern working as explained in the new section added to the `agents.md` documentation. 

It also implements a minor non-breaking improvement to the `HumanInTheLoop` pattern allowing to optionally declare its input keys, so that they can be used by agentic patterns using the agent's inputs as precondition for its activation like GOAP, P2P and this new Blackboard one.

## General checklist

### DIFF
--- a/langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/blackboard/BlackboardPlanner.java
+++ b/langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/blackboard/BlackboardPlanner.java
@@ -1,0 +1,186 @@
+package dev.langchain4j.agentic.patterns.blackboard;
+
+import dev.langchain4j.agentic.planner.Action;
+import dev.langchain4j.agentic.planner.AgentArgument;
+import dev.langchain4j.agentic.planner.AgentInstance;
+import dev.langchain4j.agentic.planner.AgenticSystemTopology;
+import dev.langchain4j.agentic.planner.InitPlanningContext;
+import dev.langchain4j.agentic.planner.PlanningContext;
+import dev.langchain4j.agentic.planner.Planner;
+import dev.langchain4j.agentic.scope.AgenticScope;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+
+import static java.util.stream.Collectors.toMap;
+
+/**
+ * A blackboard planner that activates agents based on data availability in the shared scope.
+ * <p>
+ * Agents are knowledge sources that post partial results to the {@link AgenticScope} (the blackboard).
+ * After each agent completes, the planner inspects the blackboard and activates whichever single agent
+ * can contribute next. When multiple agents are ready, a {@link ConflictResolutionStrategy} determines
+ * which one fires; if no strategy is provided, declaration order is used.
+ * <p>
+ * The planner terminates when:
+ * <ul>
+ *   <li>The goal predicate is satisfied (by default, when the planner's outputKey is present in scope)</li>
+ *   <li>No agent can fire (quiescence)</li>
+ *   <li>The maximum number of invocations is reached</li>
+ * </ul>
+ */
+public class BlackboardPlanner implements Planner {
+
+    private static final Logger LOG = LoggerFactory.getLogger(BlackboardPlanner.class);
+
+    private static final int DEFAULT_MAX_INVOCATIONS = 50;
+
+    private Predicate<AgenticScope> goalPredicate;
+    private final ConflictResolutionStrategy conflictResolutionStrategy;
+    private final int maxInvocations;
+
+    private Map<String, AgentActivator> agentActivators;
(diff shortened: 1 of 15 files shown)


langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/blackboard/BlackboardPlanner.java
```diff
@@ -0,0 +1,186 @@
...
+        }
+
+        if (goalPredicate.test(scope)) {
+            LOG.info("Goal predicate satisfied after {} invocations", invocationCounter);
+            return done();
+        }
+
+        if (invocationCounter >= maxInvocations) {
+            LOG.warn("Maximum invocations ({}) reached without satisfying goal", maxInvocations);
+            return done();
```
dliubarskyi, line 110: shouldn't it fail in this case?

langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/blackboard/ConflictResolutionStrategy.java
```diff
@@ -0,0 +1,15 @@
+package dev.langchain4j.agentic.patterns.blackboard;
+
+import dev.langchain4j.agentic.planner.AgentInstance;
+import dev.langchain4j.agentic.scope.AgenticScope;
+
+/**
+ * Strategy for resolving conflicts when multiple agents can fire simultaneously on the blackboard.
+ * The strategy receives the current scope state and two candidate agents, and returns which one
+ * should be activated first. It is applied as a pairwise reduction over all ready agents.
```
dliubarskyi: >  It is applied as a pairwise reduction over all ready agents.

So if there are 3 agents, it will be called 2 times (e.g., `(a1, a2) -> a2, (a2, a3)`)?
Wouldn't it be more flexible to accept a list of agent instances and call it only once?

dliubarskyi (dismissed):
@mariofusco thank you!

langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/blackboard/ConflictResolutionStrategy.java
glaforge: I was wondering the same... hmmm...
It's probably easier to compare 2 elements together (programmaticaly-wise), but I'm wondering if the strategy could be different if there's awareness that another agent is part of the group. I mean: if we compare a1 and a2, maybe there are some situations where the result of the comparison could be different if the strategy also knew there was a3 in the subagent list?
Maybe it's a theoretical situation, I don't know. Just wondering.

dliubarskyi: When user has access to the entire list of agents, they could do pairwise comparison as well (if required). But if user sees only 2 agents at a time, it might be too restrictive in some cases

mariofusco: Yes, at the moment I do [this](https://github.com/mariofusco/langchain4j/blob/571449f3b79f38ce44aaa8e48b8594ec65ff413c/langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/blackboard/BlackboardPlanner.java#L139), but I agree that it will be better to make the `ConflictResolutionStrategy` aware of all the candidate agents at the same time. I will change this accordingly, thanks for suggestion.

langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/blackboard/BlackboardPlanner.java
mariofusco: I'm not sure, but I don't think so. This pattern is intended to be composable with any other patterns, and under my point of view this is a situation similar to a loop reaching the maximum number of allowed iterations without satisfying the exit condition.

dliubarskyi: I see, I thought it is similar to `maxSequentialToolsInvocations`/`maxToolCallingRoundTrips` where it is usually to protect from endless loop when LLM is lost/hallucinated

langchain4j-agentic-patterns/src/main/java/dev/langchain4j/agentic/patterns/blackboard/ConflictResolutionStrategy.java
mariofusco: @dliubarskyi @glaforge I followed your suggestion also adding to the `ConflictResolutionStrategy` a few convenient factory methods allowing to define that strategy in a bit less cumbersome way. Please give it a second look.

dliubarskyi (approved):
@mariofusco thank you!

Issue #5134, opened 2026-05-07
[BUG] VertexAiGeminiStreamingChatModel — Gemini 3 tool calls fail with `INVALID_ARGUMENT: missing thought_signature` on the second turn


anpe-efficy:
## Summary

Gemini 3 (e.g. `gemini-3.1-pro-preview`) returns a mandatory `thoughtSignature` on every `functionCall` part. The next request must echo that signature back, in the same content-part position, or the API immediately returns `400 INVALID_ARGUMENT`. The Vertex AI Gemini adapter (`langchain4j-vertex-ai-gemini`) does not extract or round-trip this field, so any conversation that calls a tool fails on the very next turn after the tool result is appended.

This is the same root problem described in the closed issue #4097 — but the work that landed for #4097 (PR #4200) targeted the OpenAI / DeepSeek `reasoning_content` path and does not cover Vertex AI Gemini. The sibling integration `langchain4j-google-ai-gemini` exposes `sendThinking(true)` / `returnThinking(true)` and (per its docs) handles the signature; the Vertex integration has no equivalent.

## Affected module

- `langchain4j-vertex-ai-gemini` 1.14.0-beta24 (alongside `langchain4j` 1.14.0)
- Both `VertexAiGeminiChatModel` and `VertexAiGeminiStreamingChatModel` are presumed affected; reproduced below on the streaming variant.

## Environment

- `langchain4j` 1.14.0
- `langchain4j-vertex-ai-gemini` 1.14.0-beta24
- `langchain4j-mcp` 1.14.0-beta24
- `com.google.cloud:google-cloud-vertexai` 1.27.0
- Java 17 (Eclipse Temurin)
- Model: `gemini-3.1-pro-preview`
- Location: `global` (with `apiEndpoint("aiplatform.googleapis.com")` workaround for the SDK's `global-aiplatform.googleapis.com` host bug)
- Tools: an `McpToolProvider` exposing 4 read-only MCP tools (`get_metadata`, `get_entity`, `query_entities`, `get_current_employee`)

## Reproduction

```java
StreamingChatModel model = VertexAiGeminiStreamingChatModel.builder()
    .project(projectId)
    .location("global")
    .apiEndpoint("aiplatform.googleapis.com")
    .modelName("gemini-3.1-pro-preview")
    .credentials(credentials)
    .build();

interface Agent
{
    @SystemMessage("{{systemPrompt}}")
[...]

PR #5255, opened 2026-05-17
feat(google-genai): introduce batch processing, grounding metadata, and image labels

glaforge:
- Implemented comprehensive Batch API support covering all major modalities: GoogleGenAiBatchChatModel, GoogleGenAiBatchEmbeddingModel, and GoogleGenAiBatchImageModel. These new classes support both inline requests and file-based batch job creation, along with retrieval, cancellation, and deletion operations.
- Exposed GroundingMetadata in GoogleGenAiChatResponseMetadata by mapping it directly from the SDK's Candidate response in GoogleGenAiContentMapper.
- Added support for custom labels mapping in GoogleGenAiImageModel and GoogleGenAiBatchImageModel using GenerateContentConfig.
- Updated test suites to use newer Gemini model versions, replacing gemini-2.0-flash with gemini-3.1-flash-lite.


ipeychev:
## Test report - PR #5255 against `gemini-3-flash-preview` (Vertex AI mode)

Hi @glaforge - following up on your request in [#5151](https://github.com/langchain4j/langchain4j/pull/5151#issuecomment-4478007545), I built PR #5255 from your `feature/google-genai-updates` branch and ran it end-to-end against `gemini-3-flash-preview` in Vertex AI mode. I also reproduced the same scenarios with the raw `@google/genai` (TS, 2.4.0) and `google-genai` (Python, 2.4.0) SDKs to separate wrapper-specific behavior from Gemini-side behavior.

**Setup:** `langchain4j-google-genai:1.16.0-beta26-SNAPSHOT` (your branch HEAD), `projectId="your-project-here"`, `location="global"`, `modelName="gemini-3-flash-preview"`, `thinkingBudget=1024`, `.labels(Map.of("company_id", ..., "user_id", ...))`.

## What works

- **Labels.** `.labels(Map<String, String>)` on `GoogleGenAiChatModel.Builder` builds and the request is accepted. (I have no Cloud Billing confirmation yet - that's a ~24h followup.)
- **Single-call `thought_signature` round-trip.** Test: `"What is the weather in Paris?"` with one `get_weather(city)` tool, `toolChoice=REQUIRED`. Turn 1 returns one `ToolExecutionRequest` and `AiMessage.attributes()` contains the corresponding `thought_signature_<callId>` (base64). Turn 2 replays `[user, ai, toolResult]` and Gemini accepts it - final text *"The weather in Paris is currently 20 degrees and sunny."* No `INVALID_ARGUMENT`. So #5134 is fixed for the single-call path here.

## What fails

**Parallel tool calls in one turn fail at Turn 2.** Same test but prompted *"What is the weather in Paris? Then also tell me the weather in Tokyo."* Turn 1 returns two `ToolExecutionRequest`s and I attach both `ToolExecutionResultMessage`s. Turn 2 fails deterministically with:

> `400 . Please ensure that the number of function response parts is equal to the number of function call parts of the function call turn.`

## Cross-SDK comparison (same prompt / model / project / location)

| | Single call | Parallel calls |
|---|---|---|
| `langchain4j-google-genai` (PR #5255) | ✅ | ❌ 400 |
| `@google/genai` (TS, 2.4.0) | ✅ | ✅ |
| `google-genai` (Python, 2.4.0) | ✅ | ✅ |

In the raw SDKs my Turn 2 looks like this (Python; TS is identical in shape):

```python
types.Content(
    role="user",
    parts=[
        types.Part.from_function_response(name="get_weather", response={"result": "...Paris..."}),
        types.Part.from_function_response(name="get_weather", response={"result": "...Tokyo..."}),
    ],
)
```

`GoogleGenAiContentMapper.toContents(List<ChatMessage>)` (on your branch) maps `1 ChatMessage → 1 Content` via `messages.stream().map(::toContent)`, so two `ToolExecutionResultMessage`s become two separate `Content` objects, each with `role="function"` and one `functionResponse` part. That's the structural difference vs the raw SDKs in my parallel-call run.

I haven't patched the mapper and re-tested, so I can't tell you whether grouping the parts alone is enough or whether `role="function"` also needs to flip to `"user"` for the parallel case. One data point I do have: the Java single-call path succeeds today with `role="function"` + one part, so `role="function"` is at least valid at that arity.

## Observation worth flagging: parallel `thoughtSignature` emission

In Turn 1 of the parallel-call test the model returns two `functionCall` parts but only the first carries a `thoughtSignature`. Consistent across all three SDKs in my runs:

- Java: `AiMessage.attributes()` has exactly one `thought_signature_<callId>` (for the first call).
- TS: `parts[0].thoughtSignature` present (68 bytes), `parts[1]` has none.
- Python: `parts[0].thought_signature` present (50 bytes), `parts[1]` has none.

Reporting this as observation, not a bug - the raw SDKs round-trip just that one signature and Gemini accepts it. Flagging in case it surprises you.

## Minor ergonomics in `GoogleGenAiClientFactory`

`createClient(...)` sets `vertexAI(true)` only inside `if (googleCredentials != null)`. Setting `.projectId(...).location(...)` alone with no credentials and no API key falls through to the `apiKey` branch and the underlying SDK throws:

> *API key must either be provided or set in the environment variable `GOOGLE_API_KEY` or `GEMINI_API_KEY`. If both are set, `GOOGLE_API_KEY` will be used.*

Workaround on my side: explicit `.googleCredentials(GoogleCredentials.getApplicationDefault())`. The raw SDKs resolve credentials automatically when `vertexai=true` + project + location are set. Auto-resolving ADC in this branch, or throwing a Vertex-mode-aware error, would save the next person a round.

## Reproduction

Three minimal programs (Java / TS / Python) with the exact scenarios above. Attached as a zip - the bundled `README.md` walks through ADC setup, replacing the `your-project-here` placeholder with your own GCP project, and the run command for each.

[pr-5255-repro.zip](https://github.com/user-attachments/files/27968094/pr-5255-repro.zip)

## Summary

Labels and the single-call round-trip work. The only blocker on our side is the parallel-call replay, since multi-tool turns are common in our agent workflows.

glaforge:
@ipeychev Great detailed feedback, thank you!
I've pushed a change, could you please have a look? Hopefully this will be fixed.

ipeychev:
@glaforge Just tested - both fixes work on 2562cb7.
Thank you for the work!

Do you have an ETA when the PR will be merged and langchain with google-genai will be released?

glaforge:
@ipeychev I don't know exactly, but probably in 2 or 3 weeks at the earliest, as I think Dmytro is away for a couple weeks.

jfigueiras-denodo:
Hi, I was hitting the same issue today, I am testing with the new **Gemini 3.5 Flash** model, and got stuck with the `thought_signature` issue. Glad that you are already working on this!

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiStreamingChatModel.java
jfigueiras-denodo, line 223: I have found a problem here. The `finalAiMessage` is set here without the attributes. Without this, the `aiMessage` **will not have the required thought signature**.

This makes that later, in the `GoogleGenAiContentMapper#toContent` method, when the attribute is requested via `aiMsg.attribute("thought_signature_" + req.id(), String.class)`, null is always returned.

To solve this, the attributes must be saved to a variable when processing each of the chunks and then they must be set to the `finalAiMessage`

jfigueiras-denodo (changes requested):

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiStreamingChatModel.java
jfigueiras-denodo: I have found the issue testing the patch. I applied the solution that I suggested here and now it works.

glaforge: Thanks @jfigueiras-denodo. I've pushed a new change on that PR branch. Could you please test it to see if I fixed it as you explained and it works for your use case? Thanks in advance.

jfigueiras-denodo: I have tested the new changes and yes, everything works now! Thank you very much!

jfigueiras-denodo:
I will try to test if everything works fine with the new Embedding classes

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiStreamingChatModel.java
glaforge: Ah, wonderful! 
Thanks a lot for testing it again!

glaforge:
@jfigueiras-denodo For embeddings, I'm looking forward to your feedback!
In the Google AI Gemini module, the logic I had written for batching a bunch of requests was a bit more involved, and was splitting requests in smaller batches. I'm curious if you'll test such cases, and if I should put back the same kind of logic. I'm still pondering.

glaforge:
@jfigueiras-denodo This is the micro-batching logic I was mentionning:
https://github.com/langchain4j/langchain4j/blob/main/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiEmbeddingModel.java#L72
I should probably add that back again, so that users can configure the window size of those micro batches, the max number of tokens per batch, etc.

jfigueiras-denodo:
> @jfigueiras-denodo This is the micro-batching logic I was mentionning: https://github.com/langchain4j/langchain4j/blob/main/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/GoogleAiEmbeddingModel.java#L72 I should probably add that back again, so that users can configure the window size of those micro batches, the max number of tokens per batch, etc.

Yes, that would be needed in our case. We currently allow to configure the max number of tokens per batch, so that would be a missing feature.

glaforge:
@jfigueiras-denodo OK, let me add that back in then.

glaforge:
@jfigueiras-denodo I have implemented it now. Please check if this works for you.
There's a special case for embedding text segments which have a title metadata key (as you can't batch with different titles, so I group by title). But maybe you're not concerned about this.
I've added retry logic.
And the default micro-batch size is 100 segments.

jfigueiras-denodo:
> @jfigueiras-denodo I have implemented it now. Please check if this works for you. There's a special case for embedding text segments which have a title metadata key (as you can't batch with different titles, so I group by title). But maybe you're not concerned about this. I've added retry logic. And the default micro-batch size is 100 segments.

Thank you very much for these contributions. I will try these changes and tell you about it as soon as posible!

jfigueiras-denodo:
@glaforge I have tested the `GoogleGenAiEmbeddingModel` class. I tried with the following models:

1. `text-multilingual-embedding-002`
2. `text-embedding-005`
3. `gemini-embedding-001`
4. `gemini-embedding-2`

The only one that gave me problems with our current use case is `gemini-embedding-2`.

With this model, I am getting this error:

```
java.lang.IllegalArgumentException: The embedContent API for this model only supports one content at a time.
	at com.google.genai.Models.embedContent(Models.java:7643)
	at dev.langchain4j.model.google.genai.GoogleGenAiEmbeddingModel.lambda$embedAll$2(GoogleGenAiEmbeddingModel.java:146)
	at dev.langchain4j.internal.ExceptionMapper.withExceptionMapper(ExceptionMapper.java:29)
	at dev.langchain4j.internal.RetryUtils.lambda$withRetryMappingExceptions$2(RetryUtils.java:326)
	at dev.langchain4j.internal.RetryUtils$RetryPolicy.withRetry(RetryUtils.java:204)
	at dev.langchain4j.internal.RetryUtils.withRetry(RetryUtils.java:263)
	at dev.langchain4j.internal.RetryUtils.withRetryMappingExceptions(RetryUtils.java:326)
	at dev.langchain4j.internal.RetryUtils.withRetryMappingExceptions(RetryUtils.java:309)
```

The error is coming from this check under the `com.google.genai.Models#embedContent` method:

```java
    boolean isVertexEmbedContentModel =
        this.apiClient.vertexAI() && Transformers.tIsVertexEmbedContentModel(model);
    if (isVertexEmbedContentModel && contents.size() > 1) {
      throw new IllegalArgumentException(
          "The embedContent API for this model only supports one content at a time.");
    }
```

jfigueiras-denodo:
> @glaforge I have tested the `GoogleGenAiEmbeddingModel` class. I tried with the following models:
> 
> 1. `text-multilingual-embedding-002`
> 2. `text-embedding-005`
> 3. `gemini-embedding-001`
> 4. `gemini-embedding-2`
> 
> The only one that gave me problems with our current use case is `gemini-embedding-2`.
> 
> With this model, I am getting this error:
> 
> ```
> java.lang.IllegalArgumentException: The embedContent API for this model only supports one content at a time.
> 	at com.google.genai.Models.embedContent(Models.java:7643)
> 	at dev.langchain4j.model.google.genai.GoogleGenAiEmbeddingModel.lambda$embedAll$2(GoogleGenAiEmbeddingModel.java:146)
> 	at dev.langchain4j.internal.ExceptionMapper.withExceptionMapper(ExceptionMapper.java:29)
> 	at dev.langchain4j.internal.RetryUtils.lambda$withRetryMappingExceptions$2(RetryUtils.java:326)
> 	at dev.langchain4j.internal.RetryUtils$RetryPolicy.withRetry(RetryUtils.java:204)
> 	at dev.langchain4j.internal.RetryUtils.withRetry(RetryUtils.java:263)
> 	at dev.langchain4j.internal.RetryUtils.withRetryMappingExceptions(RetryUtils.java:326)
> 	at dev.langchain4j.internal.RetryUtils.withRetryMappingExceptions(RetryUtils.java:309)
> ```
> 
> The error is coming from this check under the `com.google.genai.Models#embedContent` method:
> 
> ```java
>     boolean isVertexEmbedContentModel =
>         this.apiClient.vertexAI() && Transformers.tIsVertexEmbedContentModel(model);
>     if (isVertexEmbedContentModel && contents.size() > 1) {
>       throw new IllegalArgumentException(
>           "The embedContent API for this model only supports one content at a time.");
>     }
> ```
@glaforge Currently, a workaround would be to set `maxSegmentsPerBatch`=1 to make the `gemini-embedding-2` model to work. But in the documentation it is stated as an example to set it to 100. I think maybe there is a problem regarding this.

glaforge:
@jfigueiras-denodo Gemini Embedding 2 is a multimodal one, maybe it's more expensive to run, and they prevent more than one item at a time? (as you can pass videos, audio, images, in addition to text)

Which flavor did you try here, the one from Vertex AI? Or the one from the Gemini developer API (from DeepMind)? (or both?)

jfigueiras-denodo:
> @jfigueiras-denodo Gemini Embedding 2 is a multimodal one, maybe it's more expensive to run, and they prevent more than one item at a time? (as you can pass videos, audio, images, in addition to text)
> 
> Which flavor did you try here, the one from Vertex AI? Or the one from the Gemini developer API (from DeepMind)? (or both?)

@glaforge I am currently using Gemini Enterprise Agent Platform (formerly Vertex AI). 

It looks that this model (and I suppose that the next ones that will come) will trigger this function: `com.google.genai.Transformers#tIsVertexEmbedContentModel`

```java
  /**
   * Checks if a given model name is a Vertex AI embed content model.
   *
   * @param model The model name to check.
   * @return True if the model is a Vertex AI embed content model, false otherwise.
   */
  public static boolean tIsVertexEmbedContentModel(String model) {
    // Gemini Embeddings except gemini-embedding-001.
    return (model.contains("gemini") && !model.contains("gemini-embedding-001"))
        // Open-source MaaS embedding models.
        || model.contains("maas");
  }
```
Then, as I am doing, if you use Gemini Enterprise Agent Platform (formerly Vertex AI), the two conditions are true to check if there are more than 1 content:

```java
    boolean isVertexEmbedContentModel =
        this.apiClient.vertexAI() && Transformers.tIsVertexEmbedContentModel(model);
    if (isVertexEmbedContentModel && contents.size() > 1) {
      throw new IllegalArgumentException(
          "The embedContent API for this model only supports one content at a time.");
    }
```


Making this to throw the `IllegalArgumentException`.

jfigueiras-denodo:
@glaforge ¿Maybe they want these kind of models to use the batch endpoint?

glaforge:
I'm gonna ask around. 
But in the meantime, yes, either the batch API, or use max batch size == 1.

jfigueiras-denodo:
@glaforge 

Google GenAI version 1.54.0 has been released: https://mvnrepository.com/artifact/com.google.genai/google-genai/1.54.0

Version 1.53.0 (currently used here) can be updated.

glaforge:
@jfigueiras-denodo well spotted! upgrading!

dskarzh:
Hi @glaforge. Thanks for the work on this new integration.

We use Gemini extensively in our platform and would love to migrate from the old Gemini integration module to this new one. One thing is missing for feature parity (at least at the level we expose to our clients): frequency and presence penalties. I prepared a [tiny PR](https://github.com/glaforge/langchain4j/pull/1) adding support for them, targeting your branch since you made quite a lot of changes in the related code. Gemini Developer API rejects those settings for newer models, but they do work with Gemini Enterprise Agent Platform. I can retarget the PR to main if you would prefer to keep my changes separate from your PR.

glaforge:
Nice, thank you @dskarzh, I've merged your PR in my branch.

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiChatResponseMetadata.java
```diff
@@ -58,12 +70,18 @@ public static Builder builder() {
     public static class Builder extends ChatResponseMetadata.Builder<Builder> {
 
         private GenerateContentResponse rawResponse;
+        private GroundingMetadata groundingMetadata;
```
dliubarskyi: Isn't `GroundingMetadata` already included in the `GenerateContentResponse`?

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiTokenCountEstimator.java
```diff
@@ -0,0 +1,192 @@
...
+                ? builder.client
+                : GoogleGenAiClientFactory.createClient(
+                        builder.apiKey,
+                        builder.googleCredentials,
+                        builder.projectId,
+                        builder.location,
+                        builder.timeout,
+                        builder.customHeaders,
+                        builder.apiEndpoint);
+        this.modelName = ensureNotBlank(getOrDefault(builder.modelName, "gemini-2.5-flash-lite"), "modelName");
```
dliubarskyi: we stopped hardcoding default model names a while ago, as they become outdated very fast

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchChatModel.java
```diff
@@ -0,0 +1,422 @@
...
+    private final List<String> allowedFunctionNames;
+    private final String vertexSearchDatastore;
+    private final Map<String, String> labels;
+    private final String cachedContent;
+    private final ChatRequestParameters defaultRequestParameters;
+
+    private GoogleGenAiBatchChatModel(Builder builder) {
+        this.modelName = builder.modelName;
+        this.maxRetries = getOrDefault(builder.maxRetries, 3);
+        this.safetySettings = builder.safetySettings != null ? new ArrayList<>(builder.safetySettings) : null;
```
dliubarskyi:
```suggestion
        this.safetySettings = copy(builder.safetySettings);
```

please use `copy` where possible

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchChatModel.java
```diff
@@ -0,0 +1,422 @@
...
+        this.thinkingBudget = builder.thinkingBudget;
+        this.thinkingLevel = builder.thinkingLevel;
+        this.seed = builder.seed;
+        this.googleSearchEnabled = builder.googleSearch != null ? builder.googleSearch : false;
```
dliubarskyi:
```suggestion
        this.googleSearchEnabled = getOrDefault(builder.googleSearch, false);
```

please use `getOrDefault` where possible

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchChatModel.java
```diff
@@ -0,0 +1,422 @@
...
+                        builder.apiEndpoint);
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @Override
+    public BatchResponse<ChatResponse> submit(BatchRequest<ChatRequest> request) {
+        return submit("batch-chat-job", request.requests());
```
dliubarskyi: perhaps default display name should include some YYYY-MM-YY-HH-MM-SS timestamp or something, so that it is more user friendly ?

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchEmbeddingModel.java
```diff
@@ -0,0 +1,328 @@
...
+    private final Client client;
+    private final String modelName;
+    private final Integer maxRetries;
+
+    private final Integer outputDimensionality;
+    private final TaskTypeEnum taskType;
+    private final String titleMetadataKey;
+
+    private GoogleGenAiBatchEmbeddingModel(Builder builder) {
+        this.modelName = getOrDefault(builder.modelName, "gemini-embedding-2");
```
dliubarskyi: same here

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchEmbeddingModel.java
```diff
@@ -0,0 +1,328 @@
...
+                        builder.apiEndpoint);
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @Override
+    public BatchResponse<Response<Embedding>> submit(BatchRequest<TextSegment> request) {
+        return submit("batch-embedding-job", request.requests());
```
dliubarskyi: same here

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchEmbeddingModel.java
```diff
@@ -0,0 +1,328 @@
...
+        client.batches.delete(batchId, DeleteBatchJobConfig.builder().build());
+    }
+
+    private BatchResponse<Response<Embedding>> processResponse(BatchJob batchJob) {
+        String jobName = batchJob.name().orElse("unknown");
+        Known state = batchJob.state().map(JobState::knownEnum).orElse(Known.JOB_STATE_UNSPECIFIED);
+
+        BatchState translatedState;
+        switch (state) {
+            case JOB_STATE_PENDING:
```
dliubarskyi: there seems to be a lot of duplicated code, can some methods be extracted and re-used?

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchImageModel.java
```diff
@@ -0,0 +1,383 @@
...
+                        builder.apiEndpoint);
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    @Override
+    public BatchResponse<Response<Image>> submit(BatchRequest<String> request) {
+        return submit("batch-image-job", request.requests());
```
dliubarskyi: same here

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchImageModel.java
```diff
@@ -0,0 +1,383 @@
...
+                .build();
+    }
+
+    private BatchResponse<Response<Image>> processResponse(BatchJob batchJob) {
+        String jobName = batchJob.name().orElse("unknown");
+        Known state = batchJob.state().map(JobState::knownEnum).orElse(Known.JOB_STATE_UNSPECIFIED);
+
+        BatchState translatedState;
+        switch (state) {
+            case JOB_STATE_PENDING:
```
dliubarskyi: same here

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiEmbeddingModel.java
```diff
@@ -0,0 +1,299 @@
...
+        return outputDimensionality;
+    }
+
+    public static Builder builder() {
+        return new Builder();
+    }
+
+    public static class Builder {
+        private Client client;
+        private String modelName = "gemini-embedding-2";
```
dliubarskyi: same here

langchain4j-google-genai/src/main/java/dev/langchain4j/model/google/genai/GoogleGenAiImageModel.java
```diff
@@ -0,0 +1,355 @@
...
+                : GoogleGenAiClientFactory.createClient(
+                        builder.apiKey,
+                        builder.googleCredentials,
+                        builder.projectId,
+                        builder.location,
+                        builder.timeout,
+                        builder.customHeaders,
+                        builder.apiEndpoint);
+
+        this.modelName = getOrDefault(builder.modelName, "gemini-3.1-flash-image-preview");
```
dliubarskyi: same here

langchain4j-google-genai/src/test/java/dev/langchain4j/model/google/genai/GoogleGenAiBatchImageModelIT.java
```diff
@@ -0,0 +1,45 @@
...
+@EnabledIfEnvironmentVariable(named = "GOOGLE_AI_GEMINI_API_KEY", matches = ".+")
+class GoogleGenAiBatchImageModelIT {
+
+    private static final String GOOGLE_AI_GEMINI_API_KEY = System.getenv("GOOGLE_AI_GEMINI_API_KEY");
+
+    @Test
+    void test_create_and_cancel_batch() throws InterruptedException {
+        GoogleGenAiBatchImageModel batchModel = GoogleGenAiBatchImageModel.builder()
+                .apiKey(GOOGLE_AI_GEMINI_API_KEY)
+                .modelName("gemini-3.1-flash-image-preview") // Or another appropriate model
```
dliubarskyi:
```suggestion
                .modelName("gemini-3.1-flash-image-preview")
```
what is the point of this comment?

dliubarskyi (commented):
@glaforge thanks!

dliubarskyi (approved):
@glaforge thanks a lot!

PR #5256, opened 2026-05-18
fix(gemini): add IMAGE_RECITATION to GeminiFinishReason enum

Pahkho:
## Summary
- Add `IMAGE_RECITATION` to `GeminiFinishReason` enum in `GeminiGenerateContentResponse`
- Map `IMAGE_RECITATION` to `CONTENT_FILTER` in `FinishReasonMapper`

## Problem
Gemini API returns `IMAGE_RECITATION` as a finish reason when generated image content is too similar to copyrighted training data. Without this enum value, Jackson deserialization fails with:

```
InvalidFormatException: Cannot deserialize value of type `GeminiFinishReason` from String "IMAGE_RECITATION": not one of the values accepted for Enum class
```

## Test plan
- [x] Verify `IMAGE_RECITATION` deserializes correctly from Gemini API response
- [x] Verify it maps to `FinishReason.CONTENT_FILTER` as expected
- [x] Existing tests continue to pass


dliubarskyi (approved):
@Pahkho thanks!

PR #5264, opened 2026-05-18
Move parameter validation to AI service build time

serelli:
## Summary

- Moves `validateParameters` from invocation time to build time in `AiServiceValidation`, fulfilling the `TODO` comment in `DefaultAiServices.java`.
- Removes the now-unnecessary `private static final Set<Method> VALID_METHODS` cache that was used to avoid redundant per-invocation validation.
- Adds a call to `validateParameters(serviceClass, method)` at the end of `validateMethod`, so all parameter validation runs during `AiServices.build()`.
- Removes the `validateParameters` call (and its import) from the `invoke` handler in `DefaultAiServices`.
- Updates `AiServicesUserMessageConfigTest`: illegal-parameter tests now assert that `IllegalConfigurationException` is thrown at `.build()` time rather than at invocation time. Each test case that exercises a specific invalid parameter configuration is given its own dedicated single-method interface so the error is deterministic.

## Why

Fail-fast is better UX: developers discover misconfigured services immediately when the service is constructed, not on the first call in production. This change eliminates the static mutable cache and makes the validation lifecycle explicit and predictable.

## Test plan

- [ ] All existing tests in `AiServicesUserMessageConfigTest` continue to pass.
- [ ] Tests `illegal_user_message_configuration_3`, `_4`, `_7`, `_8`, `_9` now verify the exception is raised at `.build()` time.
- [ ] No regression in other `AiServices*Test` / `AiServices*IT` classes (validation only added, not removed).

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceValidation.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceValidation.java
@@ -6,7 +6,6 @@ import dev.langchain4j.model.chat.request.ChatRequestParameters;
 import dev.langchain4j.service.memory.ChatMemoryAccess;
 import java.lang.reflect.Method;
 import java.lang.reflect.Parameter;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
@@ -15,8 +14,6 @@ import static dev.langchain4j.service.IllegalConfigurationException.illegalConfi
 import static java.lang.reflect.Modifier.isStatic;
 
 class AiServiceValidation {
-
-    private static final Set<Method> VALID_METHODS = new HashSet<>();
 
     private AiServiceValidation() { }
 
@@ -70,12 +67,11 @@ class AiServiceValidation {
                 }
             }
         }
+
+        validateParameters(serviceClass, method);
     }
 
     static void validateParameters(Class<?> serviceClass, Method method) {
-        if (!VALID_METHODS.add(method)) {
-            return;
-        }
 
         Parameter[] parameters = method.getParameters();
         if (parameters == null || parameters.length < 2) {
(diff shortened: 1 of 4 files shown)


mariofusco (changes requested):
It seems that this pull request doesn't compile at all, are you still working on it? If so can you please flag this as a draft?

langchain4j/src/main/java/dev/langchain4j/service/AiServiceValidation.java
```diff
@@ -70,12 +66,11 @@ private static void validateMethod(Class<?> serviceClass, Method method, boolean
...
             }
         }
+
+        validateParameters(serviceClass, method);
     }
 
     static void validateParameters(Class<?> serviceClass, Method method) {
-        if (!VALID_METHODS.add(method)) {
-            return;
-        }
```
mariofusco, line 74: Why did you remove this? If the method is invalid or it has been already validated, what's the point of doing this check again?

langchain4j/src/main/java/dev/langchain4j/service/AiServiceValidation.java
```diff
@@ -70,12 +66,11 @@ private static void validateMethod(Class<?> serviceClass, Method method, boolean
                 }
             }
         }
+
+        validateParameters(serviceClass, method);
```
mariofusco, line 70: @dliubarskyi Is it enough and always correct to simply move the parameter validation here? I think we discussed this in the past, but there was at least one situation (that I don't remember now) where this couldn't work?

dliubarskyi: TBH I don't remember either...

mariofusco: If we don't have (or don't remember which is the same thing at this point) a valid reason to check the method parameters at runtime, and all tests pass with this change, I believe that doing the validation at build-time could be a nice improvement, wdyt?

PR #5266, opened 2026-05-18
fix(vespa): rethrow exception in VespaEmbeddingStoreIT#deployVespaApp

serelli:
## Summary

- Replaces `e.printStackTrace()` with `throw new RuntimeException("Failed to deploy Vespa application", e)` in `VespaEmbeddingStoreIT#deployVespaApp`.

## Why

Silently swallowing the exception means a Vespa deployment failure shows up as a confusing "connection refused" error in the actual tests rather than a clear setup failure. Rethrowing immediately surfaces the root cause.

## Test plan

- [x] `VespaEmbeddingStoreIT` continues to pass when Vespa starts successfully.
- [x] A deployment failure now surfaces a `RuntimeException` with a descriptive message and the original cause.


dliubarskyi (approved):
@serelli thanks!

PR #5271, opened 2026-05-19
fix(milvus): convert raw search score by MetricType instead of always cosine (#5251)

### LINKED ISSUE #5251: [Bug] Milvus: RelevanceScore always uses fromCosineSimilarity regardless of MetricType
## Bug Description

 In `Mapper.java`, the `toEmbeddingMatches` method always calls `RelevanceScore.fromCosineSimilarity(score)`
 regardless of the configured `MetricType`, which produces incorrect relevance scores when using **L2** or **IP** metric.

 https://github.com/langchain4j/langchain4j/blob/a23ef82caf23c142dfdada90573c3add8ed47513/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/Mapper.java#L90

 ## Root Cause

 | MetricType | Milvus raw score meaning | Expected conversion | Actual behavior |
 |---|---|---|---|
 | COSINE | similarity in `[-1, 1]` | `(1 + score) / 2` | ✅ Correct |
 | L2 | distance in `[0, +∞)` | `1.0 / (1.0 + score)` | ❌ Wrong |
 | IP | inner product | normalize accordingly | ❌ Wrong |

 For **L2**, the raw score is a *distance* (lower = more similar), but `fromCosineSimilarity`
 treats it as a *similarity* (higher = more similar), completely inverting the semantic meaning.

 ## Expected Fix

 ```java
 double relevanceScore;
 if (metricType == MetricType.L2) {
     relevanceScore = 1.0 / (1.0 + score); // L2 distance to relevance
 } else {
     relevanceScore = RelevanceScore.fromCosineSimilarity(score);
 }

Impact

Users relying on L2 metric will get incorrect relevance scores and wrong result rankings.

spor3006:
Fixes #5251.

## Problem

`Mapper.toEmbeddingMatches` in `langchain4j-milvus` unconditionally called `RelevanceScore.fromCosineSimilarity` on the raw score returned by Milvus, regardless of the `MetricType` configured on the collection's index. The Milvus score field has different semantics per metric:

| MetricType | Raw score meaning | Required conversion |
|---|---|---|
| `COSINE` | similarity in `[-1, 1]` (higher = more similar) | `(score + 1) / 2` |
| `IP` (inner product) | similarity, `[-1, 1]` for normalized vectors; unbounded otherwise | normalized: cosine mapping; unnormalized: clamp |
| `L2` | Euclidean **distance** in `[0, +∞)` (**lower** = more similar) | `1 / (1 + distance)` |

For `L2`, applying the cosine mapping inverted the semantic meaning of "relevance":

- A perfect match (distance `0`) was scored `0.5`.
- A distant point (distance `1`) was scored `1.0`.

Users running with `L2` (or unnormalized `IP`) saw incorrect relevance scores and result rankings.

## Fix

Introduce a `Mapper.toRelevanceScore(MetricType, double)` helper, thread the configured `metricType` from `MilvusEmbeddingStore` through `toEmbeddingMatches`, and branch on the metric:

```java
static double toRelevanceScore(MetricType metricType, double score) {
    if (metricType == MetricType.L2) {
        return 1.0 / (1.0 + score);                                // (0, 1]
    }
    if (metricType == MetricType.IP) {
        double relevance = RelevanceScore.fromCosineSimilarity(score);
        return Math.max(0.0, Math.min(1.0, relevance));            // clamp unnormalized
    }
    return RelevanceScore.fromCosineSimilarity(score);             // legacy default
}
```

- **`COSINE`**: behavior unchanged.
- **`L2`**: `1 / (1 + distance)`; identity vectors now score `1.0`, monotonically decreasing.
- **`IP`**: same as cosine for normalized vectors (the recommended Milvus pattern), clamped to `[0, 1]` so unnormalized scores can't escape the `EmbeddingMatch` contract.
- **Other metric types** (e.g. `HAMMING`, `JACCARD`): cosine conversion retained as the legacy default to preserve backwards compatibility — users on those metrics should validate scores match their expectations.

## Tests

Adds `MapperTest` (new file) with four unit tests against the helper:

| Test | What it pins |
|---|---|
| `cosineSimilarity_is_mapped_to_zero_to_one_range` | Backwards-compat: `COSINE` branch still produces `(score + 1) / 2`. |
| `l2_distance_is_mapped_so_that_closer_is_higher_relevance` | Regression for #5251: identity → `1.0`, monotonically decreasing, result stays in `(0, 1]`. |
| `inner_product_uses_cosine_mapping_for_normalized_vectors` | `IP` and `COSINE` agree on `[-1, 1]` inputs (normalized-vector case). |
| `inner_product_clamps_unnormalized_scores_to_unit_interval` | Unnormalized `IP` scores outside `[-1, 1]` are clamped to `[0, 1]`. |

All four pass; the module's existing unit tests still pass.

```
$ ./mvnw -pl langchain4j-milvus -DskipITs test
[INFO] Tests run: 4, Failures: 0, Errors: 0, Skipped: 0 — MapperTest
[INFO] BUILD SUCCESS
```

`./mvnw -pl langchain4j-milvus -Pspotless spotless:check` is clean.

## Scope notes

- Only `Mapper` + the single call site in `MilvusEmbeddingStore` change; existing IT tests (`MilvusEmbeddingStoreIT`, etc.) are unchanged and continue to pass against the unchanged `COSINE` default.
- No public API change.
- This addresses the score-conversion bug specifically; users on `L2` or unnormalized `IP` will see their relevance scores change shape (which is the point — they were wrong before). The PR body documents this so existing tests/dashboards that depend on the old (incorrect) score range can be updated.

Opened as draft per CONTRIBUTING.md; happy to mark ready for review once a maintainer takes a first look.

### DIFF
--- a/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/Mapper.java
+++ b/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/Mapper.java
@@ -1,5 +1,13 @@
 package dev.langchain4j.store.embedding.milvus;
 
+import static com.google.gson.ToNumberPolicy.LONG_OR_DOUBLE;
+import static dev.langchain4j.internal.Utils.isNullOrBlank;
+import static dev.langchain4j.internal.Utils.isNullOrEmpty;
+import static dev.langchain4j.store.embedding.milvus.CollectionOperationsExecutor.queryForVectors;
+import static dev.langchain4j.store.embedding.milvus.Generator.generateEmptyJsons;
+import static dev.langchain4j.store.embedding.milvus.Generator.generateEmptyScalars;
+import static java.util.stream.Collectors.toList;
+
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.JsonObject;
@@ -12,10 +20,10 @@ import dev.langchain4j.store.embedding.RelevanceScore;
 import io.milvus.client.MilvusServiceClient;
 import io.milvus.common.clientenum.ConsistencyLevelEnum;
 import io.milvus.exception.ParamException;
+import io.milvus.param.MetricType;
 import io.milvus.response.QueryResultsWrapper;
 import io.milvus.response.QueryResultsWrapper.RowRecord;
 import io.milvus.response.SearchResultsWrapper;
-
 import java.lang.reflect.Type;
 import java.math.BigDecimal;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


dliubarskyi:
Already fixed by other PR

Issue #5227, opened 2026-05-14
[BUG]  com.fasterxml.jackson.core.JsonParseException: Unexpected character ('"' (code 34)): was expecting comma to separate Object entries


gogym:
**Describe the bug**

com.fasterxml.jackson.core.JsonParseException: Unexpected character ('"' (code 34)): was expecting comma to separate Object entries
**Log and Stack trace**

Caused by: com.fasterxml.jackson.core.JsonParseException: Unexpected character ('"' (code 34)): was expecting comma to separate Object entries
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 64]
	at com.fasterxml.jackson.core.JsonParser._constructReadException(JsonParser.java:2660) ~[jackson-core-2.18.3.jar:2.18.3]
	at com.fasterxml.jackson.core.base.ParserMinimalBase._reportUnexpectedChar(ParserMinimalBase.java:741) ~[jackson-core-2.18.3.jar:2.18.3]
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._skipComma(ReaderBasedJsonParser.java:2429) ~[jackson-core-2.18.3.jar:2.18.3]
	at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextFieldName(ReaderBasedJsonParser.java:924) ~[jackson-core-2.18.3.jar:2.18.3]
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer._readAndBindStringKeyMap(MapDeserializer.java:608) ~[jackson-databind-2.18.3.jar:2.18.3]
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer.deserialize(MapDeserializer.java:449) ~[jackson-databind-2.18.3.jar:2.18.3]
	at com.fasterxml.jackson.databind.deser.std.MapDeserializer.deserialize(MapDeserializer.java:32) ~[jackson-databind-2.18.3.jar:2.18.3]
	at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:342) ~[jackson-databind-2.18.3.jar:2.18.3]
	at com.fasterxml.jackson.databind.ObjectMapper._readMapAndClose(ObjectMapper.java:4931) ~[jackson-databind-2.18.3.jar:2.18.3]
[...]

PR #5274, opened 2026-05-20
fix(anthropic): preserve raw replayed tool arguments

DragonFSKY:
## Issue
Addresses the Anthropic replay case reported in #5227.

## Change
This PR updates how the Anthropic mapper replays historical tool calls with `ToolExecutionRequest.arguments`.

`ToolExecutionRequest.arguments()` is already a JSON string. The previous flow parsed it into a `Map` and then let Jackson serialize it back to JSON, which added an unnecessary string -> map -> string roundtrip and could fail while replaying historical malformed arguments.

`AnthropicToolUseContent.input` is now a raw JSON `String` annotated with `@JsonRawValue`. `toAnthropicInput()` now only falls back to `{}` for blank arguments; otherwise it passes through `ToolExecutionRequest.arguments()` unchanged.

This preserves the original arguments payload when replaying historical `tool_use` messages, including malformed JSON, and lets Jackson write valid JSON object inputs as objects instead of quoted strings.

## Verification
- [X] `mvn -pl langchain4j-anthropic -Dtest=AnthropicMapperTest test`
- [X] `mvn -pl langchain4j-anthropic revapi:check`
- [X] `git diff --check`
- [X] GitHub compile/unit, spotless, compliance, and split-package checks are passing

Note: one `integration_test (21)` job is currently red in `langchain4j-mcp-docker`; it failed while reaching Docker Hub and appears unrelated to this Anthropic change.

## General checklist

### DIFF
--- a/langchain4j-anthropic/revapi.json
+++ b/langchain4j-anthropic/revapi.json
@@ -8,6 +8,35 @@
           "code": "java.field.removed",
           "old": "field dev.langchain4j.model.anthropic.AnthropicChatModelName.CLAUDE_3_HAIKU_20240307",
           "justification": "Model no longer exists and was previously deprecated; removed intentionally"
+        },
+        {
+          "ignore": true,
+          "code": "java.method.parameterTypeChanged",
+          "old": "parameter dev.langchain4j.model.anthropic.internal.api.AnthropicToolUseContent.Builder dev.langchain4j.model.anthropic.internal.api.AnthropicToolUseContent.Builder::input(===java.util.Map<java.lang.String, java.lang.Object>===)",
+          "new": "parameter dev.langchain4j.model.anthropic.internal.api.AnthropicToolUseContent.Builder dev.langchain4j.model.anthropic.internal.api.AnthropicToolUseContent.Builder::input(===java.lang.String===)",
+          "parameterIndex": "0",
+          "justification": "Internal serialize-only request DTO now stores raw JSON to avoid parsing and re-serializing tool arguments"
+        },
+        {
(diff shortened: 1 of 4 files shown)


DragonFSKY:
I used the existing behavior in `langchain4j-vertex-ai-anthropic` as a reference here. It already falls back to an empty input object when replaying historical tool arguments that cannot be parsed, so this PR aligns the standalone Anthropic mapper with that conservative behavior.

A more fine-grained approach might also be possible in the future, for example including a sanitized parse error or sanitized raw arguments in the tool error text so the model has more context to self-correct. I kept this PR narrow because the existing `ToolExecutionResultMessage` already carries the error result, and exposing raw malformed arguments may need a separate discussion around privacy/safety and provider semantics.

The latest CI run is green as well.

DragonFSKY:
Sorry for the delayed response, and thanks for the suggestion.

I updated the PR to follow this approach: `AnthropicToolUseContent.input` is now a raw JSON `String` annotated with `@JsonRawValue`, and `toAnthropicInput()` only falls back to `{}` when the arguments are blank. Otherwise, it passes through `ToolExecutionRequest.arguments()` unchanged.

This avoids the string -> map -> string roundtrip and preserves the original `tool_use` payload when replaying messages, including invalid JSON.

I also updated the related tests and added the Revapi ignore for the internal DTO shape change.

Please let me know if anything else needs adjustment, and I’ll address it promptly.

DragonFSKY:
Closing this draft since #5335 has been merged and #5227 is now resolved. Thanks for the discussion and guidance here.

PR #5286, opened 2026-05-21
Allow custom resolution of agentic parameters suppliers

geoand:


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/AgenticServices.java
```diff
@@ -68,6 +70,16 @@ public class AgenticServices {
 
     private AgenticServices() {}
 
+    private static final List<ChatSupplierParameterResolver> chatSupplierParameterResolvers = new CopyOnWriteArrayList<>();
```
mariofusco: I expect that this feature will be used only for integration reasons with the quarkus extension and not directly by end users. If you agree, maybe it would be better to move this list directly inside `DeclarativeUtil` thus avoid this to be part of the public API exposed by `AgenticServices`.

geoand: Fine with me, but how would the integration then enable it? Currently it's done like so: https://github.com/quarkiverse/quarkus-langchain4j/pull/2486/changes#diff-151c0669b4150b84ca30330f8b663692e5e4ebade280b02662318826c52568b5R43

mariofusco: I'm not sure I understand the question, `DeclarativeUtil` is a public class anyway, just flagged with the `@Internal` annotation.

geoand: Okay, PR updated

mariofusco:
I have seen what you have done here and it's ok for me. I have a couple of possible alternative proposals and I will sketch them here, just to weight the pros and cons of each.

1. Have an initializer of the `AgenticScope`, a `Consumer<AgenticScope>` that is called every time a new `AgenticScope` is created. This could be used for inject any state in the `AgenticScope` and then it could be automatically picked up to populate the variable passed to any of the *Supplier method. For instance, in the case of the ChatModelSupplier we can use this to inject in the `AgenticScope` all the configured `ChatModel`s and by doing so you could implement the ChatModelSupplier method something like: 

```
@ChatModelSupplier
static ChatModel chatModel(CritiqueResult critique, ChatModel baseModel, ChatModel enhancedModel) {
    return critique != null && critique.score() > 7.8 ? enhancedModel : baseModel;
}
```

2. Allow a method annotated with `@ChatModelSupplier` to optionally return a String representing the name of the model to be used. In this case the method allowing to select the different `ChatModel`s could be something like:

```
@ChatModelSupplier
static String chatModel(CritiqueResult critique) {
    return critique != null && critique.score() > 7.8 ? "enhancedModel" : "baseModel";
}
```

Note that the first solution is more general and could open more possibilities while the second is specifically tailored for the `ChatModel` use case. That's why I'm eventually in favor option 1.

Regarding the implementation of option 2. I also generated a plan to give you an idea of what this would imply.

 Implementation plan

  1. LangChain4j core change (upstream)

  In DeclarativeUtil.lambda$configureAgent$7 — the lambda that processes @ChatModelSupplier:

  - Check the method's return type (method.getReturnType())
  - If ChatModel.class: current behavior (no change)
  - If String.class: wrap the result through a model name resolver — a Function<String, ChatModel> stored on the
  AgentBuilder

  This means adding to AgentBuilder:
  - A field Function<String, ChatModel> chatModelNameResolver
  - A method chatModelNameResolver(Function<String, ChatModel>) to set it
  
  For the no-parameter case: invoke the method, get the String, pass it through the resolver, call
  agentBuilder.chatModel(resolved).

  For the parameterized case: call agentBuilder.chatModel(scope -> { invoke method with scope params → String → 
  resolver → ChatModel }).

  2. Quarkus extension changes

  AgenticProcessor.java (4 changes):
  - validateChatModelSupplier() (line 207): allow return type String in addition to ChatModel
  - detectAgents() (line 92-95): record whether the supplier returns String
  - cdiSupport() (lines 431-433): for String-returning suppliers, create a new ChatModelInfo variant and ensure all
  named ChatModel CDI beans are injectable
  - Remove the @ModelName + @ChatModelSupplier conflict check (lines 99-104) — no longer conflicting since String
  return naturally bridges the two
  
  AiAgentCreateInfo.java:
  - Add a new ChatModelInfo variant: FromAnnotationReturningModelName — permits the sealed interface to know the supplier returns a model name
  
  DetectedAiAgentBuildItem.java:
  - Add a boolean chatModelSupplierReturnsString field
  
  AgenticRecorder.java — in QuarkusAgenticContextConsumer.accept():
  - When the info indicates a String-returning supplier, set the chatModelNameResolver on the agentBuilder:

```
  agenticContext.agentBuilder().chatModelNameResolver(modelName -> {
      if (NamedConfigUtil.isDefault(modelName)) {
          return cdiContext.getInjectedReference(ChatModel.class);
      }   
      return cdiContext.getInjectedReference(ChatModel.class, ModelName.Literal.of(modelName));
  }); 
```

  3. Tests

  - Update NonChatModelReturnTypeChatModelSupplierTest: allow String return, still reject others (Object, Integer,
  etc.)
  - Update ModelNameWithChatModelSupplierConflictTest: remove or relax the conflict check
  - Add a new test: @ChatModelSupplier returning String (no params) — static model name selection
  - Add a new test: @ChatModelSupplier returning String (with params) — dynamic model name selection based on agent input

geoand:
TBH, what I did is much more in line with what other libs / frameworks do when they have to call a user supplied method and need to resolve parameters

mariofusco:
> TBH, what I did is much more in line with what other libs / frameworks do when they have to call a user supplied method and need to resolve parameters

Ok for me, I was just brainstorming other alternatives.

I think that I will provide the possibility to initialize the `AgenticScope` because I believe that this could be useful in many other different scenarios, but at this point that's unrelated with this issue.

mariofusco (approved):

geoand:
Sounds good!

geoand:
> I think that I will provide the possibility to initialize the AgenticScope because I believe that this could be useful in many other different scenarios, but at this point that's unrelated with this issue.

Right, that makes sense!

geoand:
@mariofusco do you know when LangChain4j will have a release that includes this?

Thanks

mariofusco:
> @mariofusco do you know when LangChain4j will have a release that includes this?

I'm in Portugal attending JNation and Dmytro is on PTO. I think he will be back on Thu and I already asked him to drop a patch release as soon as he will be back.

PR #5287, opened 2026-05-21
fix(agentic): avoid persisting streaming outputs in scope state

### LINKED ISSUE #5285: [BUG] DefaultAgenticScope.state contains runtime proxy objects, causing AgenticScopeSerializer.toJson() to fail under JPMS
## Issue Title

**`DefaultAgenticScope.state` contains runtime proxy objects, causing `AgenticScopeSerializer.toJson()` to fail under JPMS**

---

## Environment

- **langchain4j-agentic**: `1.15.0-beta25`
- **JDK**: 21 (JPMS module system enabled)
- **Jackson**: 2.21.2

---

## Problem Description

When implementing a custom `AgenticScopeStore` (e.g., persisting to Redis) via `AgenticScopePersister.setStore()`, the framework calls `AgenticScopeSerializer.toJson(agenticScope)` to serialize `DefaultAgenticScope` to JSON, which fails.

**Root Cause**: `DefaultAgenticScope.state` (a `ConcurrentHashMap`) gets automatically populated with runtime JDK dynamic proxy objects by the framework. These objects cannot be serialized by Jackson.

---

## Full Stacktrace

```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid type definition for type `jdk.proxy2.$Proxy102`: 
Failed to construct BeanSerializer for [simple type, class jdk.proxy2.$Proxy102]: 
(java.lang.IllegalArgumentException) Failed to call `setAccess()` on Field 'h' (of class `java.lang.reflect.Proxy`) 
due to `java.lang.reflect.InaccessibleObjectException`, 
problem: Unable to make field protected java.lang.reflect.InvocationHandler java.lang.reflect.Proxy.h accessible: 
module java.base does not "opens java.lang.reflect" to unnamed module

(through reference chain: 
  dev.langchain4j.agentic.scope.DefaultAgenticScope["state"]
  -> java.util.concurrent.ConcurrentHashMap["result"]
  -> dev.langchain4j.service.AiServiceTokenStream["context"]
  -> dev.langchain4j.service.AiServiceContext["eventListenerRegistrar"]
  -> dev.langchain4j.observability.api.DefaultAiServiceListenerRegistrar["listeners"]
  -> ... -> java.util.LinkedHashSet[0])
```

---

## Root Cause Analysis

When serializing `DefaultAgenticScope.state`, Jackson traverses the following object chain and eventually hits a JDK dynamic proxy:

```
DefaultAgenticScope
  └─ state: ConcurrentHashMap
       └─ key "result" → AiServiceTokenStream (JDK dynamic proxy $Proxy102)
            └─ context → AiServiceContext
                 └─ eventListenerRegistrar → DefaultAiServiceListenerRegistrar
                      └─ listeners → ConcurrentHashMap
                           └─ EventListeners.listeners → LinkedHashSet
                                └─ [0] → listener proxy object → BOOM
```

Core contradictions:

1. **`state` does not distinguish between serializable data and runtime context.** Objects like `AiServiceTokenStream`, `ToolServiceContext`, and `Method` are runtime dependencies and should not appear in `state` that needs to be persisted.
2. **JDK dynamic proxy objects are inherently non-serializable.** When Jackson attempts to introspect `$Proxy102`, it must access the `java.lang.reflect.Proxy.h` field, which JPMS prohibits.
3. **Users have no control over what gets put into `state`.** These runtime objects are automatically injected by the framework internally — users can neither remove nor replace them.

---

## Workarounds Attempted (none fully satisfactory)

| Approach | Result |
|----------|--------|
| Add `@JsonIgnoreType` Mixin for the proxy interface (`AiServiceTokenStream`) | ❌ Jackson matches Mixins by runtime actual type. `$Proxy102 extends Proxy`, not the interface, so the Mixin does not apply. |
| Add `@JsonIgnoreType` Mixin for `java.lang.reflect.Proxy` | ❌ `@JsonIgnoreType` lacks `@Inherited` meta-annotation. `$Proxy102` does not inherit this annotation from its parent class, rendering the Mixin ineffective. |
| Custom `AnnotationIntrospector` that forces `Proxy` subclasses to be marked as ignorable | ⚠️ Works as a user-side hack, but ignored runtime data is lost upon deserialization. |

---

## Suggested Fixes

Ideally addressed at the framework level:

1. **`DefaultAgenticScope.state` should only contain serializable data.** Runtime proxy objects (`AiServiceTokenStream`, etc.) should be separated from `state` into dedicated runtime context fields (e.g., `transient` fields).
2. Alternatively, the default `AgenticScopeJsonCodec` configuration should **pre-exclude known non-serializable runtime types**, so users don't have to figure this out when implementing custom stores.
3. Or, the documentation should explicitly state which non-serializable runtime objects a custom `AgenticScopeStore` implementation must handle.

DragonFSKY:
## Issue
Fixes #5285

## Change
This PR prevents live streaming outputs from being stored in persistent agentic scope data.

When an agent or workflow has an `outputKey` and produces a `TokenStream`, the persistent state entry for that key is cleared while the live stream is kept only in transient execution context for the current root call. This keeps final streaming workflows returning the original stream without serializing `AiServiceTokenStream` internals.

The same output policy is applied to `AgentInvocation` records, so persisted invocation history exposes a serialization-safe value for streaming outputs.

Async streaming is handled according to how the stream is used:
- if an async streaming output is the final propagated workflow output, it remains runtime-only and is not stored in any persisted checkpoint snapshot
- if an async streaming subagent is not the final output, the stream is consumed into text as before, so downstream agents can still read the `outputKey` value

Non-streaming outputs continue to be written to scope state as before. This change does not alter the public serializer/API surface.

## Verification
- [x] `mvn -pl langchain4j-agentic -Dtest=dev.langchain4j.agentic.internal.AgenticScopeOutputTest,dev.langchain4j.agentic.scope.AgenticScopeStreamingStateTest,dev.langchain4j.agentic.RecoverabilityTest test`
- [x] `mvn -pl langchain4j-agentic test`
- [x] `git diff --check`
- [x] GitHub Actions `spotless` passed on this head commit

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentExecutor.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/AgentExecutor.java
@@ -11,11 +11,11 @@ import dev.langchain4j.agentic.planner.Planner;
 import dev.langchain4j.agentic.scope.AgentInvocation;
 import dev.langchain4j.agentic.scope.DefaultAgenticScope;
 import dev.langchain4j.service.TokenStream;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public record AgentExecutor(AgentInvoker agentInvoker, Object agent) implements AgentInstance, InternalAgent {
 
@@ -38,7 +38,10 @@ public record AgentExecutor(AgentInvoker agentInvoker, Object agent) implements 
     }
 
     private Object handleAgentFailure(
-            AgentInvocationException e, DefaultAgenticScope agenticScope, Object invokedAgent, PlannerExecutor planner) {
+            AgentInvocationException e,
+            DefaultAgenticScope agenticScope,
+            Object invokedAgent,
+            PlannerExecutor planner) {
         ErrorRecoveryResult recoveryResult = agenticScope.handleError(agentInvoker.name(), e);
         return switch (recoveryResult.type()) {
             case THROW_EXCEPTION -> throw e;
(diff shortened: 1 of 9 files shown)


mariofusco:
This solution seems to be much more complex and invasive than the one that I have here https://github.com/langchain4j/langchain4j/pull/5290

They seem to address the same problem though, am I missing something?

DragonFSKY:
Thanks for taking a look.

I agree that my PR is broader and more invasive than a minimal serializer-side fix. My intent was to explore a slightly stricter boundary for streaming outputs: keeping the live TokenStream available at runtime, while avoiding putting runtime-only streaming objects into the persisted AgenticScope snapshot.

That is why I also tried to cover related persistence paths such as AgentInvocation output/input, not only the immediate state serialization failure.

That said, I understand this may be broader than what is needed for the immediate issue, and there may be a more elegant way to address the underlying boundary without changing this much. I’ll close this PR for now.

PR #5294, opened 2026-05-22
MCP: handle inbound notifications/cancelled from server

### LINKED ISSUE #5292: [BUG] MCP: inbound notifications/cancelled from server is silently ignored
## Summary

When the MCP server sends a `notifications/cancelled` to abort a long-running request it previously accepted, `langchain4j-mcp` ignores the notification. The client's pending `CompletableFuture` keeps blocking until the configured timeout fires, instead of failing fast with the server-supplied reason.

## Reproduction (conceptual)

1. Client calls `tools/call` with a long-running tool.
2. Server later decides to abort the request and sends `{"jsonrpc":"2.0","method":"notifications/cancelled","params":{"requestId":<id>,"reason":"…"}}`.
3. Client waits the full `toolExecutionTimeout` (default 60s) before completing the future.

## Root cause

`McpClientMethod.NOTIFICATION_CANCELLED` exists ([`McpClientMethod.java:12-13`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/protocol/McpClientMethod.java#L12-L13)) and the client *sends* these outbound (see tool-timeout path in [`DefaultMcpClient.java:319-321`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java#L319-L321)). But [`McpServerMethod`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/protocol/McpServerMethod.java) has no entry for it, so the switch in [`McpOperationHandler.handleNotification`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/McpOperationHandler.java#L111-L143) falls into the default branch and only logs a warning.

The MCP base protocol (since 2024-11-05) defines this as a *bidirectional* notification — both sides may send it. See spec: https://modelcontextprotocol.io/specification/2025-06-18/basic#cancellation

## Proposal

1. Add `NOTIFICATION_CANCELLED("notifications/cancelled")` to [`McpServerMethod.java`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/protocol/McpServerMethod.java).
2. In `handleNotification`, add a case that parses `params.requestId` + `params.reason`, calls `pendingOperations.remove(id)`, and `completeExceptionally(...)` with an `McpException` carrying the server-supplied reason. The pattern mirrors `cancelAllPendingOperations` at [`McpOperationHandler.java:176-182`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/McpOperationHandler.java#L176-L182).
3. Add a default no-op listener method `onNotificationCancelled(long requestId, String reason)` to [`McpClientListener`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClientListener.java), and invoke it from the handler.

This is non-breaking: recent commit [`c908ea4c`](https://github.com/langchain4j/langchain4j/commit/c908ea4c) ("support additional event types in McpClientListener") shows the same extension pattern is welcome.

## Tests

Extend [`DefaultMcpClientTest`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/test/java/dev/langchain4j/mcp/client/DefaultMcpClientTest.java) (or add a focused `McpOperationHandlerTest`) with a test that:

- registers a pending operation,
- feeds a synthesised `notifications/cancelled` `JsonNode` through `McpOperationHandler.handle(...)`,
- asserts the pending future completes exceptionally with the server's reason,
- asserts the listener hook is invoked.

## Collision check

None of the active PRs touch `McpOperationHandler`, `McpServerMethod`, or `McpClientListener`.

## Next step

Happy to open a draft PR once acked.

farzad-sedaghatbin:
## Summary

Closes #5292.

The MCP base protocol defines `notifications/cancelled` as **bidirectional**: both the client and the server can cancel a previously accepted request (see [spec, cancellation](https://modelcontextprotocol.io/specification/2025-06-18/basic#cancellation)). The client already sends these outbound on tool timeout ([`DefaultMcpClient.java:319-321`](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java#L319-L321)) — but when the **server** sends one to abort a long-running request, `McpOperationHandler.handleNotification` fell into the default branch and only logged a warning. The pending `CompletableFuture` then blocked until the configured timeout fired instead of failing fast with the server-supplied reason.

## Changes

1. **`McpServerMethod`** — add `NOTIFICATION_CANCELLED("notifications/cancelled")`.
2. **`McpOperationHandler`** — add a new `BiConsumer<Long, String> onServerCancelled` callback parameter and a `handleCancelledNotification(...)` branch in the switch. When the notification arrives, the matching pending operation (if any) is removed from `pendingOperations` and completed with a `CancellationException` carrying the server's reason. If the request id is unknown or already completed, the listener is still notified (debug-logged).
3. **`DefaultMcpClient`** — wire the new callback to fan out to listeners via `McpClientListener.onNotificationCancelled(...)`.
4. **`McpClientListener`** — new default no-op method `onNotificationCancelled(long requestId, String reason)`. Non-breaking; follows the same extension pattern as commit [`c908ea4c`](https://github.com/langchain4j/langchain4j/commit/c908ea4c) ("support additional event types in McpClientListener").

## Why `CancellationException`

`java.util.concurrent.CancellationException` is the JDK-standard signal for "this future was cancelled," propagates cleanly through `CompletableFuture#get()` (no `ExecutionException` wrapping), and matches user expectations. The existing `cancelAllPendingOperations` uses `IllegalStateException` for transport teardown — but that's a different signal (the connection died), whereas this is a clean protocol-level cancellation.

## Test plan

- [x] New `McpOperationHandlerTest` (3 tests):
  - completes the pending operation exceptionally with the server-supplied reason
  - invokes the listener even when the request id is unknown
  - ignores a malformed notification without `requestId`
- [x] All 65 MCP unit tests pass (`./mvnw -pl langchain4j-mcp -am test`).
- [x] `./mvnw -pl langchain4j-mcp spotless:check` passes.

## API impact

Additive only:

- one new enum entry on `McpServerMethod`
- one new optional method on `McpClientListener` (with a default no-op)
- one new constructor parameter on `McpOperationHandler` — internal; only `DefaultMcpClient` constructs it (verified via repo-wide grep)

No SPI break for downstream users.

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
@@ -184,7 +184,8 @@ public class DefaultMcpClient implements McpClient {
                         notifyListeners(l -> l.onNotificationProgress(notification));
                     },
                     () -> notifyListeners(l -> l.onServerPing()),
-                    () -> notifyListeners(l -> l.onServerRootsList()));
+                    () -> notifyListeners(l -> l.onServerRootsList()),
+                    (requestId, reason) -> notifyListeners(l -> l.onNotificationCancelled(requestId, reason)));
             ((ObjectNode) RESULT_TIMEOUT)
                     .putObject("result")
                     .putArray("content")
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClientListener.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpClientListener.java
@@ -242,6 +242,16 @@ public interface McpClientListener {
      */
     default void onNotificationProgress(McpProgressNotification notification) {}
 
+    /**
+     * Called when the server sends a {@code notifications/cancelled} notification to abort
+     * a previously accepted request. The corresponding pending operation, if any, is
+     * completed exceptionally before this listener is invoked.
+     *
+     * @param requestId the id of the cancelled request, as supplied by the server
+     * @param reason    an optional server-supplied reason for the cancellation; may be {@code null}
+     */
+    default void onNotificationCancelled(long requestId, String reason) {}
+
     // ========== Server-initiated requests ==========
 
     /**
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/McpOperationHandler.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/transport/McpOperationHandler.java
@@ -10,7 +10,9 @@ import dev.langchain4j.mcp.protocol.McpRootsListResponse;
 import dev.langchain4j.mcp.protocol.McpServerMethod;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
+import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 import org.slf4j.Logger;
@@ -36,6 +38,7 @@ public class McpOperationHandler {
     private final McpProgressHandler progressHandler;
     private final Runnable onServerPing;
     private final Runnable onServerRootsList;
+    private final BiConsumer<Long, String> onServerCancelled;
 
     public McpOperationHandler(
             Map<Long, CompletableFuture<JsonNode>> pendingOperations,
@@ -48,7 +51,8 @@ public class McpOperationHandler {
             Consumer<String> onResourceUpdate,
             McpProgressHandler progressHandler,
             Runnable onServerPing,
-            Runnable onServerRootsList) {
+            Runnable onServerRootsList,
+            BiConsumer<Long, String> onServerCancelled) {
         this.pendingOperations = pendingOperations;
         this.transport = transport;
         this.logMessageConsumer = logMessageConsumer;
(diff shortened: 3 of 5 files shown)


jmartisk:
Nice, I think this is a good improvement. The tests are failing though, can you have a look?

farzad-sedaghatbin:
Thanks for the review @jmartisk! Pushed a fix in 95a079f.

Root cause: the failing assertion was the `hasMessageContaining("42")` / `hasMessageContaining("user aborted")` checks on the JDK 25 matrix job. JDK 25 changed `CompletableFuture.reportGet()` to throw a *fresh* `CancellationException` with a generic message (`"get"`) rather than re-throwing the one originally supplied to `completeExceptionally`. JDK 17 and 21 keep the old behavior, so the test passed there. The other matrix jobs were cancelled because of `fail-fast: true` on the matrix.

Fix: read the stored throwable directly via `future.handle((v, e) -> e).join()` instead of going through `future.get()`. That bypasses `reportGet` and works the same on every JDK. Comment in the test explains the why so it isn't lost to future readers.

jmartisk (approved):

PR #5297, opened 2026-05-23
fix: make AgentMonitor thread-safe under concurrent invocations

PSchmitz-Valckenberg:
## Summary

`AgentMonitor` is auto-registered on every agent built via `AgentBuilder` ([AgentBuilder.java#L210-L212](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/agent/AgentBuilder.java#L210-L212)) and is therefore reachable from multiple threads whenever an agentic system runs a parallel topology (`ParallelAgentService`, `ParallelMapperService`, …) or when several invocations share the same `memoryId`. Two thread-safety defects could lose tracked executions or corrupt the result lists.

### 1. Non-atomic check-then-put on `ongoingExecutions`

`beforeAgentInvocation` did a plain check-then-put on a `ConcurrentHashMap`:

```java
MonitoredExecution currentExecution = ongoingExecutions.get(memoryId);
if (currentExecution == null) {
    currentExecution = new MonitoredExecution(agentRequest);
    ongoingExecutions.put(memoryId, currentExecution); // race
}
```

Two threads entering with the same `memoryId` could both observe `null`, both construct a new `MonitoredExecution` and both `put` — the losing `put` silently overwrites the winner, so one of the two invocations is never tracked and the corresponding `afterAgentInvocation` later throws `IllegalStateException` from `MonitoredExecution.afterAgentInvocation`. Structurally the same defect that was fixed for `ChatMemoryService.getOrCreateChatMemory` in #5253.

Fixed by replacing the check-then-put with `computeIfAbsent`, using an `AtomicBoolean` to distinguish the just-created case (no inner `beforeAgentInvocation` needed) from the existing case (delegate to the existing `MonitoredExecution`).

### 2. Non-thread-safe `ArrayList` inside `ConcurrentHashMap`

`afterAgentInvocation` and `onAgentInvocationError` stored plain `ArrayList`s as values:

```java
successfulExecutions.computeIfAbsent(memoryId, k -> new ArrayList<>()).add(execution);
failedExecutions.computeIfAbsent(memoryId, k -> new ArrayList<>()).add(execution);
```

`computeIfAbsent` is atomic, but the `add` that follows is not — concurrent completions on the same `memoryId` can corrupt the underlying array (lost entries, `ArrayIndexOutOfBoundsException`, `null` slots). Switched both maps to `Collections.synchronizedList(new ArrayList<>())`, which is the convention already used elsewhere in the module ([DefaultAgenticScope.java#L46-L47](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/scope/DefaultAgenticScope.java#L46-L47), [AgentInvocation.java#L19-L20](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentInvocation.java#L19-L20)).

## Changes

- `AgentMonitor.java` — `computeIfAbsent` + `AtomicBoolean` for `beforeAgentInvocation`; `synchronizedList` for `successfulExecutions` / `failedExecutions`.
- `AgentMonitorConcurrencyTest.java` — new tests:
  - `concurrent_invocations_with_distinct_memory_ids_are_all_tracked` — 200 top-level invocations across 32 threads with distinct memory IDs; verifies all are recorded and no state leaks into `ongoingExecutions`.
  - `concurrent_parallel_sub_agent_invocations_are_all_recorded` — 200 sub-agents running under a single root invocation, fanned out across 32 threads; verifies all nested invocations land in the root's `nestedInvocations` list and that the root completes cleanly.

## Test plan

- [x] `mvn -pl langchain4j-agentic test -DskipITs` — 45 tests pass, including the two new concurrency tests.
- [x] No public API changes (internal observability bookkeeping only).

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
@@ -30,10 +30,6 @@ public class AgentMonitor implements AgentListener {
     private final Map<Object, List<MonitoredExecution>> failedExecutions = new ConcurrentHashMap<>();
     private final Map<Object, MonitoredExecution> ongoingExecutions = new ConcurrentHashMap<>();
 
-    private static List<MonitoredExecution> newExecutionList(Object ignored) {
-        return Collections.synchronizedList(new ArrayList<>());
-    }
-
     @Internal
     public void setRootAgent(AgentInstance rootAgent) {
         this.rootAgent = rootAgent;
@@ -61,7 +57,7 @@ public class AgentMonitor implements AgentListener {
         if (execution.done()) {
             ongoingExecutions.remove(memoryId, execution);
             successfulExecutions
-                    .computeIfAbsent(memoryId, AgentMonitor::newExecutionList)
+                    .computeIfAbsent(memoryId, k -> Collections.synchronizedList(new ArrayList<>()))
                     .add(execution);
         }
     }
@@ -69,12 +65,11 @@ public class AgentMonitor implements AgentListener {
     @Override
     public void onAgentInvocationError(AgentInvocationError agentInvocationError) {
         Object memoryId = agentInvocationError.agenticScope().memoryId();
-        MonitoredExecution execution = ongoingExecutions.get(memoryId);
+        MonitoredExecution execution = ongoingExecutions.remove(memoryId);
         if (execution != null) {
             execution.onAgentInvocationError(agentInvocationError);
-            ongoingExecutions.remove(memoryId, execution);
             failedExecutions
-                    .computeIfAbsent(memoryId, AgentMonitor::newExecutionList)
+                    .computeIfAbsent(memoryId, k -> Collections.synchronizedList(new ArrayList<>()))
                     .add(execution);
         }
     }
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
@@ -10,7 +10,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Monitors agent executions and provides observability for the LangChain4j Agentic system.
@@ -47,13 +46,10 @@ public class AgentMonitor implements AgentListener {
     @Override
     public void beforeAgentInvocation(AgentRequest agentRequest) {
         Object memoryId = agentRequest.agenticScope().memoryId();
-        AtomicBoolean created = new AtomicBoolean(false);
-        MonitoredExecution currentExecution = ongoingExecutions.computeIfAbsent(memoryId, k -> {
-            created.set(true);
-            return new MonitoredExecution(agentRequest);
-        });
-        if (!created.get()) {
-            currentExecution.beforeAgentInvocation(agentRequest);
+        MonitoredExecution candidate = new MonitoredExecution(agentRequest);
+        MonitoredExecution existing = ongoingExecutions.putIfAbsent(memoryId, candidate);
+        if (existing != null) {
+            existing.beforeAgentInvocation(agentRequest);
         }
     }
 
@@ -63,7 +59,7 @@ public class AgentMonitor implements AgentListener {
         MonitoredExecution execution = ongoingExecutions.get(memoryId);
         execution.afterAgentInvocation(agentResponse);
         if (execution.done()) {
-            ongoingExecutions.remove(memoryId);
+            ongoingExecutions.remove(memoryId, execution);
             successfulExecutions.computeIfAbsent(memoryId, AgentMonitor::newExecutionList).add(execution);
         }
     }
@@ -71,9 +67,10 @@ public class AgentMonitor implements AgentListener {
     @Override
     public void onAgentInvocationError(AgentInvocationError agentInvocationError) {
         Object memoryId = agentInvocationError.agenticScope().memoryId();
-        MonitoredExecution execution = ongoingExecutions.remove(memoryId);
+        MonitoredExecution execution = ongoingExecutions.get(memoryId);
         if (execution != null) {
             execution.onAgentInvocationError(agentInvocationError);
+            ongoingExecutions.remove(memoryId, execution);
             failedExecutions.computeIfAbsent(memoryId, AgentMonitor::newExecutionList).add(execution);
         }
     }
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
@@ -10,6 +10,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Monitors agent executions and provides observability for the LangChain4j Agentic system.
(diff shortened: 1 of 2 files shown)


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
```diff
@@ -42,11 +47,12 @@ AgentInstance rootAgent() {
...
-            ongoingExecutions.put(memoryId, currentExecution);
-        } else {
+        AtomicBoolean created = new AtomicBoolean(false);
+        MonitoredExecution currentExecution = ongoingExecutions.computeIfAbsent(memoryId, k -> {
+            created.set(true);
+            return new MonitoredExecution(agentRequest);
+        });
+        if (!created.get()) {
             currentExecution.beforeAgentInvocation(agentRequest);
         }
```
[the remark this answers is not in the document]
PSchmitz-Valckenberg: Good call — replaced with the `putIfAbsent` candidate pattern in d1040fe. Cleaner control flow and no side-effect-in-mapping concern.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
```diff
@@ -58,7 +64,7 @@ public void afterAgentInvocation(AgentResponse agentResponse) {
         execution.afterAgentInvocation(agentResponse);
         if (execution.done()) {
             ongoingExecutions.remove(memoryId);
```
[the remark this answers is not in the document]
PSchmitz-Valckenberg: Good catch — switched to the conditional `ongoingExecutions.remove(memoryId, execution)` in d1040fe so a racing new execution for the same memoryId is preserved.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
```diff
@@ -68,7 +74,7 @@ public void onAgentInvocationError(AgentInvocationError agentInvocationError) {
         MonitoredExecution execution = ongoingExecutions.remove(memoryId);
         if (execution != null) {
             execution.onAgentInvocationError(agentInvocationError);
-            failedExecutions.computeIfAbsent(memoryId, k -> new ArrayList<>()).add(execution);
+            failedExecutions.computeIfAbsent(memoryId, AgentMonitor::newExecutionList).add(execution);
         }
```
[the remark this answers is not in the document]
PSchmitz-Valckenberg: Same pattern applied here in d1040fe: observe the execution via `get()`, then conditional `remove(memoryId, execution)` so a racing new entry is not dropped.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
```diff
@@ -30,6 +30,10 @@ public class AgentMonitor implements AgentListener {
     private final Map<Object, List<MonitoredExecution>> failedExecutions = new ConcurrentHashMap<>();
     private final Map<Object, MonitoredExecution> ongoingExecutions = new ConcurrentHashMap<>();
 
+    private static List<MonitoredExecution> newExecutionList(Object ignored) {
```
mariofusco: This method doesn't make any sense.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
```diff
@@ -57,18 +59,23 @@ public void afterAgentInvocation(AgentResponse agentResponse) {
         MonitoredExecution execution = ongoingExecutions.get(memoryId);
         execution.afterAgentInvocation(agentResponse);
         if (execution.done()) {
-            ongoingExecutions.remove(memoryId);
-            successfulExecutions.computeIfAbsent(memoryId, k -> new ArrayList<>()).add(execution);
```
mariofusco, line 61: You could simply create a synchronized list here.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
```diff
@@ -57,18 +59,23 @@ public void afterAgentInvocation(AgentResponse agentResponse) {
...
+                    .computeIfAbsent(memoryId, AgentMonitor::newExecutionList)
+                    .add(execution);
         }
     }
 
     @Override
     public void onAgentInvocationError(AgentInvocationError agentInvocationError) {
         Object memoryId = agentInvocationError.agenticScope().memoryId();
-        MonitoredExecution execution = ongoingExecutions.remove(memoryId);
+        MonitoredExecution execution = ongoingExecutions.get(memoryId);
```
mariofusco: I don't understand what's the point of doing a `get` here and a `remove` immediately after, can you please explain?

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/observability/AgentMonitorConcurrencyTest.java
```diff
@@ -0,0 +1,137 @@
...
+/**
+ * Concurrency tests for {@link AgentMonitor} verifying that it can be safely shared across
+ * threads. The patterns exercised here are produced by the agentic module's parallel and
+ * concurrent topologies, where the same monitor instance is invoked from multiple worker
+ * threads.
+ */
+class AgentMonitorConcurrencyTest {
+
+    private static final int THREADS = 32;
+    private static final int RUNS = 200;
```
mariofusco: I understand why you did this, but I'm afraid that these numbers can put our CI in trouble, I think that 4 threads and 10 runs will be more reasonable.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
PSchmitz-Valckenberg: Fair point, the helper was only there to satisfy the Function<K,V> signature of computeIfAbsent and didn't earn its keep as a named method. Dropped it and inlined Collections.synchronizedList(new ArrayList<>()) at both call sites

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
PSchmitz-Valckenberg: Done

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/observability/AgentMonitor.java
PSchmitz-Valckenberg: Two different reasons in the two methods:
afterAgentInvocation, the get and the conditional remove(memoryId, execution) are not back-to-back: between them we call execution.afterAgentInvocation(agentResponse) (which mutates the execution) and then check execution.done(). We only want to remove it from ongoingExecutions once the top-level invocation is actually finished, so we can't do an atomic remove(memoryId) up front, a nested sub-agent completion is also routed through this method and must leave the entry in place.

onAgentInvocationError, you're right, here the split into get + conditional remove(memoryId, execution) had no real justification. It was guarding against a same-memoryId race that doesn't realistically occur (a session won't start a new top-level execution while the previous one is still being errored out). Reverted to the original atomic ongoingExecutions.remove(memoryId)

langchain4j-agentic/src/test/java/dev/langchain4j/agentic/observability/AgentMonitorConcurrencyTest.java
PSchmitz-Valckenberg: Good point, I reduced THREADS from 32 to 4 and RUNS from 200 to 10. Both tests still exercise the same code paths and pass locally; they should be much friendlier to CI now

mariofusco (approved):

Issue #5275, opened 2026-05-20
cancel flux streaming


xunianchong:
The AI Service can stream response，  use flux

```
@AiService(
        wiringMode = AiServiceWiringMode.EXPLICIT,
        chatModel = "openAiChatModel",
        streamingChatModel = "openAiStreamingChatModel",
        chatMemoryProvider = "ragChatMemoryProvider7"

)
public interface MedicineAgent {
    @SystemMessage(MEDICINE_PROMPT)
    Flux<String > consultMedicine(@MemoryId String memoryId, @UserMessage String question);
}

  @Autowired
    private MedicineAgent medicineAgent;

@RequestMapping(value = "/chatRag7", produces = "text/event-stream;charset=UTF-8")
    public Flux<String> chat(String memoryId, String message){
      
                Flux<String> stream = medicineAgent.consultMedicine(memoryId,message);
                return stream
                    .doOnSubscribe(sub -> {
                        Subscription prev = activeStreams.put(memoryId, sub);
                        if (prev != null) {
                            prev.cancel();
                        }
                    })
                    .takeWhile(t -> {
                        boolean shouldStop = stopFlag.get();
                        log.info("***** takeWhile check, memoryId={}, shouldStop={}", memoryId, shouldStop);
                        return !shouldStop;
                    })
                    .doOnComplete(() -> {
                        log.info("***** stream completed, memoryId={}", memoryId);
                        sessionStopFlags.remove(memoryId);
                        activeStreams.remove(memoryId);
                    })
                    .doOnCancel(() -> {
                        log.info("***** stream cancelled, memoryId={}", memoryId);
                        sessionStopFlags.remove(memoryId);
                        activeStreams.remove(memoryId);
                    })
}

  @RequestMapping("/stopStream")
    public ResponseEntity<String> stopStream(String memoryId) {
        AtomicBoolean stopFlag = sessionStopFlags.get(memoryId);
        if (stopFlag != null) {
            stopFlag.set(true);
            log.info("***** stopFlag set to true, memoryId={}", memoryId);
        }
        Subscription subscription = activeStreams.remove(memoryId);
        if (subscription != null) {
            subscription.cancel();
            log.info("***** subscription cancelled, memoryId={}", memoryId);
            return ResponseEntity.ok("已停止");
        }
        return ResponseEntity.ok("无活跃会话");
    }
```

After aborting the stream using takeWhile, the log shows that there are event output, 
[...]

PR #5303, opened 2026-05-25
Expose StreamingHandle on TokenStream

DragonFSKY:
## Issue
Part of #5275

## Change
This PR adds an opt-in `TokenStream.onStreamingHandle(...)` callback so adapters can observe the existing `StreamingHandle` without changing the current plain token callback path.

The callback is a default no-op, so existing `TokenStream` implementations keep their current behavior unless they choose to expose a handle.

Changes included:
- add `TokenStream.onStreamingHandle(...)`;
- support the callback in `AiServiceTokenStream`;
- observe the handle in `AiServiceStreamingResponseHandler` from context-bearing partial response, thinking, and tool-call callbacks;
- preserve the existing `onPartialResponse(...)` token path;
- add tests for handle observation, duplicate callback registration, deduplication, and thinking/tool-call context propagation.

This is intended to support a companion `langchain4j-spring` PR that propagates Reactor `Flux` cancellation to `StreamingHandle.cancel()`.

## Verification
- [x] `./mvnw -pl langchain4j -Dtest=AiServiceTokenStreamStreamingHandleTest,AiServiceTokenStreamTest#start_onStreamingHandleInvokedMultipleTimes_shouldThrowException test`
- [x] `./mvnw -pl langchain4j -DskipTests install`
- [x] `mvn -Pspotless validate`
- [x] `git diff --check origin/main...HEAD`

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
@@ -53,6 +53,7 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import org.slf4j.Logger;
@@ -102,7 +103,7 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler 
     private final List<String> responseBuffer = new ArrayList<>();
     private final boolean hasOutputGuardrails;
 
-    private StreamingHandle observedStreamingHandle;
+    private final AtomicReference<StreamingHandle> observedStreamingHandle = new AtomicReference<>();
 
     private int toolCallingRoundTripsLeft;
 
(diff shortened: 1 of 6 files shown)


langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -413,6 +423,11 @@ public void onCompleteResponse(ChatResponse chatResponse) {
                     // completing
                     if (partialResponseHandler != null) {
                         responseBuffer.forEach(partialResponseHandler::accept);
+                    } else if (partialResponseWithContextHandler != null) {
+                        PartialResponseContext replayContext =
+                                new PartialResponseContext(new CancellationUnsupportedStreamingHandle());
```
dliubarskyi: can't we try using `observedStreamingHandle` instead of `CancellationUnsupportedStreamingHandle` (if available)?

langchain4j/src/main/java/dev/langchain4j/service/AiServiceStreamingResponseHandler.java
```diff
@@ -99,6 +101,8 @@ class AiServiceStreamingResponseHandler implements StreamingChatResponseHandler
     private final List<String> responseBuffer = new ArrayList<>();
     private final boolean hasOutputGuardrails;
 
+    private StreamingHandle observedStreamingHandle;
```
dliubarskyi: Shouldn't it be `volatile` or `AtomicReference`? `onPartialResponse` and other callbacks could be called from different threads

DragonFSKY:
Thanks for the review. I rebased this on the latest main and addressed both points:

- `observedStreamingHandle` now uses `AtomicReference` so callbacks from different threads have safe visibility.
- Context replay and legacy callbacks now reuse the observed streaming handle when available, falling back to `CancellationUnsupportedStreamingHandle` only when no handle has been observed.

I also added a regression test covering the mixed context/no-context partial response path.

DragonFSKY:
Before I continue with the companion Spring PR, could you confirm whether this `TokenStreamAdapter`-based approach is still the preferred direction given the `PublisherAdapter` work in #5527? If #5527 supersedes this path, I’m happy to close both PRs. I’ll keep the companion PR in draft in the meantime.

PR #5314, opened 2026-05-27
fix(qdrant): use Query RPC instead of legacy Search RPC to retrieve vectors

sam-hey:
## Issue

Closes #

### Problem

`QdrantEmbeddingStore.search()` was calling `client.searchAsync(SearchPoints)`, which dispatches to the legacy gRPC `Search` RPC (`PointsFutureStub.search`). In **Qdrant 1.13** the server-side behaviour of this legacy endpoint changed: it **no longer populates the `vectors` field on returned `ScoredPoint` messages**, even when `with_vectors = true` is explicitly requested.

The store then unconditionally called:
```java
Embedding embedding = toEmbedding(scoredPoint.getVectors().getVector());
```
With an empty protobuf default returned from `getVectors()`, `getDataList()` was empty, producing a **zero-dimension `Embedding`** and a downstream dimension error.

### Root cause

| Client method | gRPC stub call | Returns `vectors` in `ScoredPoint` (Qdrant ≥ 1.13) |
|---|---|---|
| `client.searchAsync(SearchPoints)` | `PointsFutureStub.search` — legacy Search RPC | **No** |
| `client.queryAsync(QueryPoints)` | `PointsFutureStub.query` — modern Query RPC | **Yes** |

- [Query API reference](https://api.qdrant.tech/api-reference/search/query-points) — full REST/gRPC spec

### Fix

Replace `SearchPoints` + `client.searchAsync()` with `QueryPoints` + `client.queryAsync()`, using `QueryFactory.nearest(List<Float>)` to supply the dense query vector ([`QueryFactory` source](https://github.com/qdrant/java-client/blob/master/src/main/java/io/qdrant/client/QueryFactory.java)). Both RPCs return `List<ScoredPoint>`, so `toEmbeddingMatch()` and the rest of the class are unchanged.

### Testing

Run `QdrantEmbeddingStoreIT` against a Qdrant ≥ 1.13 instance — this test exercises `search()` and would previously fail with a dimension error on any collection with stored vectors.

### Affected versions

Any deployment of langchain4j-qdrant against a Qdrant server ≥ 1.13. The Query gRPC RPC has been available since Qdrant 1.10, so there is no minimum server version concern for the fix.

## General checklist

- [x] I have manually verified that the `{Qdrant}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

### DIFF
--- a/langchain4j-qdrant/src/main/java/dev/langchain4j/store/embedding/qdrant/QdrantEmbeddingStore.java
+++ b/langchain4j-qdrant/src/main/java/dev/langchain4j/store/embedding/qdrant/QdrantEmbeddingStore.java
@@ -22,6 +22,7 @@ import dev.langchain4j.store.embedding.*;
 import dev.langchain4j.store.embedding.EmbeddingSearchRequest;
 import io.qdrant.client.QdrantClient;
 import io.qdrant.client.QdrantGrpcClient;
+import io.qdrant.client.QueryFactory;
 import io.qdrant.client.WithVectorsSelectorFactory;
 import io.qdrant.client.grpc.Common.Filter;
 import io.qdrant.client.grpc.Common.PointId;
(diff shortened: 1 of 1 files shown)


dliubarskyi (approved):
@sam-hey thank you!

PR #5316, opened 2026-05-28
Support A2A message context metadata

### LINKED ISSUE #5311: Agentic A2AClientAgent does not seem to support passing the contextId and taskId as part of the the message
i have setup a multi-turn conversation A2A quarkus app using A2A Java SDK

import org.a2aproject.sdk.spec.Message;

private Message userMessage = null;

userMessage = A2A.createUserTextMessage(mapper.writeValueAsString(mapper.readValue(question, DeepChatMessages.class)), null, null);
aiMessage = binanceA2aClientAgent.callAgent(userMessage, agentCard);

**- First turn**

contextId and taskId passed as part of the outgoing message are null so the A2A server agent will create a new tasks and added it to a new conversation which is return to the client and stored in app session.

**- Second turn** 

userMessage = A2A.createUserTextMessage(mapper.writeValueAsString(mapper.readValue(question, DeepChatMessages.class)), session.getContextId(), session.getTaskId());
aiMessage = binanceA2aClientAgent.callAgent(userMessage, agentCard);

as we send an existing contextId and taskId, the A2A Server agent will process the request without raising a new task. any artifacts will be attached the the existing task with taskId sent through in the message.

all works perfect. However I would like to replace the A2A Client Agent based on the A2A Java SDK with a new one based on Langchain4j Agentic using the @A2AClientAgent annotation, however i can't seem to be able to figure out how to specify the cntextId and taskId

this is the client agent i am testing

```
package org.acme;

import dev.langchain4j.agentic.declarative.A2AClientAgent;
import dev.langchain4j.service.UserMessage;

public interface BinanceAgenticA2aClientAgent {

    @A2AClientAgent(a2aServerUrl = "http://localhost:11001",
        outputKey = "aiMessage",
        description = "Binance Crypto Market Analizer")
    String analizeMarket(@UserMessage DeepChatMessages question);
}
```

how can I pass along the contextId and taskId?

DragonFSKY:
## Issue
Closes #5311

## Change
Adds opt-in support for passing A2A `contextId` and `taskId` through `@A2AClientAgent`.

- adds `contextIdKey` and `taskIdKey` to `@A2AClientAgent`
- adds `@A2AContextId` and `@A2ATaskId` for typed A2A client methods
- writes configured or annotated values to the A2A message envelope
- keeps legacy behavior unchanged unless the new keys or annotations are used
- adds tests for typed, untyped, negative, and declarative wiring paths

## General checklist

### DIFF
--- a/langchain4j-agentic-a2a/src/test/java/dev/langchain4j/agentic/a2a/A2AClientAgentMessageContextTest.java
+++ b/langchain4j-agentic-a2a/src/test/java/dev/langchain4j/agentic/a2a/A2AClientAgentMessageContextTest.java
@@ -1,0 +1,302 @@
+package dev.langchain4j.agentic.a2a;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyList;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+
+import dev.langchain4j.agentic.AgenticServices;
+import dev.langchain4j.agentic.UntypedAgent;
+import dev.langchain4j.agentic.declarative.A2AClientAgent;
+import dev.langchain4j.agentic.declarative.A2AContextId;
+import dev.langchain4j.agentic.declarative.A2ATaskId;
+import dev.langchain4j.agentic.declarative.SequenceAgent;
+import dev.langchain4j.agentic.internal.A2AClientBuilder;
+import dev.langchain4j.agentic.internal.A2AService;
+import dev.langchain4j.agentic.internal.AgentExecutor;
+import dev.langchain4j.agentic.internal.InternalAgent;
+import dev.langchain4j.agentic.planner.AgentArgument;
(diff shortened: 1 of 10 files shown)


langchain4j-agentic-a2a/src/test/java/dev/langchain4j/agentic/a2a/A2AClientAgentMessageContextTest.java
```diff
@@ -0,0 +1,302 @@
...
+                @A2AContextId @V("contextId") String contextId,
+                @A2ATaskId @V("taskId") String taskId);
+    }
+
+    public interface DeclarativeA2AAgent {
+
+        @A2AClientAgent(
+                a2aServerUrl = "http://test-a2a",
+                outputKey = "answer",
+                contextIdKey = "contextId",
```
mariofusco, line 72: I'm not fully understanding this solution. If you have those new `@A2AContextId` and `@A2ATaskId` annotation, why do you need also this `contextIdKey` and `taskIdKey` parameter on the `A2AClientAgent` one? They seem redundant or am I missing something?

DragonFSKY:
Closing this in favor of #5363, which provides a simpler annotation-only API for the same A2A context/task ID use case.

Issue #5280, opened 2026-05-20
[FEATURE] Support ResponseFormat on AiServices.builder()


zbendhiba:
Currently, ResponseFormat (structured JSON output) can only be configured at the ChatModel level, which ties the output schema to the model instance. This prevents reusing the same ChatModel across multiple AiServices with different output schemas.

Proposed API:
```
  AiServices.builder(MyService.class)
      .chatModel(chatModel)
      .responseFormat(responseFormat)
      .build();
```

DragonFSKY:
I looked at the existing options and the related discussion in #3871 / #4226 / #4264. My understanding is that there are already a few ways to handle `ResponseFormat` today:

- Model-level default request parameters: useful, but it ties the `ResponseFormat` to the shared `ChatModel` instance.
- Method-level `ChatRequestParameters`: good for per-call overrides, and it should probably remain the highest-priority option.
- `chatRequestTransformer`: flexible enough to inject a service-specific default, but a bit verbose for this use case.

Although these alternatives already cover the functionality to some extent, I think there may still be value in adding a small convenience API if maintainers think this use case is common enough.

In that case, I would see this issue as adding a service-level default rather than replacing the existing per-call mechanism: the same `ChatModel` could be reused across multiple `AiServices`, while each service defines its own default `ResponseFormat`.

If this direction sounds reasonable, I think the design should define the precedence explicitly. One possible ordering could be:

1. method-level `ChatRequestParameters.responseFormat()`
2. AI Service builder-level `responseFormat` with an explicit schema
3. the existing auto-generated JSON schema response format
4. AI Service builder-level simple `ResponseFormat.JSON` / `TEXT`
5. model-level default request parameters

This would keep per-call overrides strongest, preserve the existing auto-schema behavior for POJO return types, and still allow an AI Service to provide its own default when the model is shared.

The implementation would also need to preserve the selected `ResponseFormat` through streaming and tool-call follow-up requests, similar to the coverage added in #4264.

Would this be a direction worth pursuing? If so, I would be happy to help with a focused implementation.

zbendhiba:
Hey @DragonFSKY , thanks for jumping in! 
FYI, I've started exploring an implementation for this. 

[...]

PR #5325, opened 2026-05-29
ChatRequest: allow individual setters to override parameters

dliubarskyi:
## Issue
Triggered by https://github.com/langchain4j/langchain4j/issues/5280#issuecomment-4572579268

## Change
Previously, calling `ChatRequest.builder().parameters(p).responseFormat(rf).build()` (or any other individual setter combined with `parameters(...)`) threw
  `IllegalArgumentException: Cannot set both 'parameters' and 'responseFormat' on ChatRequest`. This made `chatRequest.toBuilder().responseFormat(rf).build()` unusable — a natural
  pattern when modifying a single field of an existing `ChatRequest` (e.g. inside a `chatRequestTransformer` on `AiServices`):

  ```java
  AiServices.builder(MyService.class)
      .chatModel(chatModel)
      .chatRequestTransformer(req -> req.toBuilder().responseFormat(responseFormat).build())
      .build();
  ```
  Individual setters now override the corresponding fields of parameters via ChatRequestParameters.overrideWith(...). All other fields of parameters — including provider-specific
  fields on subclasses (e.g. OpenAiChatRequestParameters) — are preserved.

  Notes:
  - When only .parameters(p) is set without any individual setters, p is passed through unchanged (identity preserved). This ensures mocked ChatRequestParameters keep working, and
  subclasses that don't override overrideWith keep their concrete type.
  - When individual setters ARE used together with .parameters(p), merging is performed via overrideWith — provider-specific subclasses that properly override overrideWith keep
  their type and their extra fields.
  - Null / empty-collection overrides are skipped (existing values kept) — same semantics as overrideWith. Documented in the Builder javadoc.
  - The per-field validation throws are gone; the constructor branches once on whether any individual setter was used.

## General checklist

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/ChatRequest.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/model/chat/request/ChatRequest.java
@@ -1,7 +1,6 @@
 package dev.langchain4j.model.chat.request;
 
 import static dev.langchain4j.internal.Utils.copy;
-import static dev.langchain4j.internal.Utils.isNullOrEmpty;
 import static dev.langchain4j.internal.ValidationUtils.ensureNotEmpty;
 import static java.util.Arrays.asList;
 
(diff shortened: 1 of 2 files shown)


zbendhiba:
Thanks @dliubarskyi !

PR #5328, opened 2026-05-29
Allow @Tool methods to also be inherited from superclasses and interfaces

### LINKED ISSUE #5302: [BUG] @Tool methods not inherited from superclas
`@Tool` methods not inherited from superclass.  I think this is a bug as this is something that should work.

I didn't 100% verify this, but my inherited tools weren't being called and the LLM was stating that it didn't know what tool to call.  Once I put the tool methods back in the concrete class, things worked, so I'm assuming that this bug report is true.

This should be done in LC4J too if it already doesn't work there.

mariofusco:
## Issue
Closes #5302

## Change
Allow `@Tool` methods to also be inherited from superclasses and interfaces.
See [updated documentation](https://github.com/langchain4j/langchain4j/pull/5328/changes#diff-9ca302b5805896cdeaec9345077d7d34ca9d203a9a62e017eb70c6445e869ae0) for more details

## General checklist

### DIFF
--- a/docs/docs/tutorials/tools.md
+++ b/docs/docs/tutorials/tools.md
@@ -769,7 +769,7 @@ This way, the LLM has more information to decide whether or not to call the give
 
 ### Inheritance and tool discovery
 
-`@Tool`-annotated methods are inherited from superclasses. When you pass a tool object to an AI Service, LangChain4j discovers `@Tool` methods from the object's class and all of its superclasses (up to, but excluding, `Object`).
+Concrete `@Tool`-annotated methods are inherited from superclasses and interfaces. When you pass a tool object to an AI Service, LangChain4j discovers `@Tool` methods from the object's class, all of its superclasses (up to, but excluding, `Object`), and `default` and `static` methods from implemented interfaces.
 
 ```java
 class BaseMathTools {
--- a/docs/docs/tutorials/tools.md
+++ b/docs/docs/tutorials/tools.md
@@ -767,6 +767,78 @@ Depending on the tool, the LLM might understand it well even without any descrip
 but it is usually better to provide clear and meaningful names and descriptions.
 This way, the LLM has more information to decide whether or not to call the given tool, and how to do so.
 
+### Inheritance and tool discovery
+
+`@Tool`-annotated methods are inherited from superclasses. When you pass a tool object to an AI Service, LangChain4j discovers `@Tool` methods from the object's class and all of its superclasses (up to, but excluding, `Object`).
+
+```java
+class BaseMathTools {
+
+    @Tool("Calculates the sum of two numbers")
+    int sum(int a, int b) {
+        return a + b;
+    }
+}
+
+class AdvancedMathTools extends BaseMathTools {
+
+    @Tool("Calculates the product of two numbers")
+    int multiply(int a, int b) {
+        return a * b;
+    }
+}
+
+Assistant assistant = AiServices.builder(Assistant.class)
+    .chatModel(model)
+    .tools(new AdvancedMathTools()) // both "sum" and "multiply" are available
+    .build();
+```
+
+A subclass can override a `@Tool` method from its parent. In that case, only the subclass version is used — including its `@Tool` annotation:
+
+```java
+class ParentTools {
+
+    @Tool("Returns the greeting")
+    String greet(String name) {
+        return "Hello, " + name;
+    }
+}
+
+class ChildTools extends ParentTools {
+
+    @Override
+    @Tool(name = "greet_formal", value = "Returns a formal greeting")
+    String greet(String name) {
+        return "Good day, " + name;
+    }
+}
+```
+
+Here the LLM will see a single tool named `greet_formal` with description "Returns a formal greeting".
+
+If a subclass declares a method with the same name as a parent method but with different parameters (an overload, not an override), both methods are discovered. Since tool names must be unique and default to the method name, you must give at least one of them an explicit name:
+
+```java
+class ParentTools {
+
+    @Tool(name = "process_text", value = "Processes a text input")
+    String process(String input) {
+        return input.toUpperCase();
+    }
+}
+
+class ChildTools extends ParentTools {
+
+    @Tool(name = "process_number", value = "Processes a numeric input")
+    int process(int input) {
+        return input * 2;
+    }
+}
+```
+
+If both methods resolve to the same tool name, an `IllegalArgumentException` is thrown at AI Service creation time.
+
 ### `@P`
 Method parameters can optionally be annotated with `@P`.
 
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
@@ -1,5 +1,6 @@
 package dev.langchain4j.agent.tool;
 
+import static dev.langchain4j.internal.Utils.allDeclaredMethods;
 import static dev.langchain4j.internal.Utils.isNotNullOrBlank;
 import static dev.langchain4j.internal.Utils.isNullOrBlank;
 import static java.util.Arrays.stream;
@@ -60,7 +61,7 @@ public class ToolSpecifications {
      * @return the {@link ToolSpecification}s.
      */
     public static List<ToolSpecification> toolSpecificationsFrom(Class<?> classWithTools) {
-        List<ToolSpecification> toolSpecifications = stream(classWithTools.getDeclaredMethods())
+        List<ToolSpecification> toolSpecifications = allDeclaredMethods(classWithTools).stream()
                 .filter(method -> method.isAnnotationPresent(Tool.class))
                 .map(ToolSpecifications::toolSpecificationFrom)
                 .collect(toList());
--- a/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
@@ -1,6 +1,6 @@
 package dev.langchain4j.agent.tool;
 
-import static dev.langchain4j.internal.Utils.allDeclaredMethods;
+import static dev.langchain4j.internal.Utils.allConcreteMethods;
 import static dev.langchain4j.internal.Utils.isNotNullOrBlank;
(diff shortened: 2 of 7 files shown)


langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
```diff
@@ -1587,25 +1588,37 @@ void should_rewrite_chat_request_using_tool(ChatModel chatModel) {
...
         }
 
         @Tool("calculates the multiplication of two integers")
         public int multiply(int first, int second) {
             int prod = first * second;
-            result.set(prod);
+            mulResult.set(prod);
             return prod;
         }
     }
```
dliubarskyi, line 1624: Could you please add a few more tests?
- tool with the same name is loaded from the superclass: we should probably fail?
- `@Tool`-annotated method from superclass is `@Override`n in the child class: which method is/should be invoked?
- `@Tool`-annotated method from superclass is `@Override`n in the child class with updated `@Tool` annotation: we need to make sure updated `@Tool` annotation is effective
- other possible corner cases

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
```diff
@@ -60,7 +61,7 @@ private ToolSpecifications() {}
      * @return the {@link ToolSpecification}s.
      */
     public static List<ToolSpecification> toolSpecificationsFrom(Class<?> classWithTools) {
-        List<ToolSpecification> toolSpecifications = stream(classWithTools.getDeclaredMethods())
+        List<ToolSpecification> toolSpecifications = allDeclaredMethods(classWithTools).stream()
```
dliubarskyi: Please update the documentation, we need to specify how inheritance works with tools

dliubarskyi: BTW, should all inherited tools be exposed, or only `public`?
What about `static` methods?

langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
```diff
@@ -553,6 +554,42 @@ public static Optional<Method> getAnnotatedMethod(Method method, Class<? extends
...
+
+    /**
+     * Returns all methods declared in the given class and its superclasses, excluding {@link Object}.
+     * If a subclass overrides a superclass method, only the subclass version is included.
+     */
+    public static List<Method> allDeclaredMethods(Class<?> clazz) {
+        return DECLARED_METHODS_CACHE.computeIfAbsent(clazz, Utils::findAllDeclaredMethods);
+    }
+
+    private static List<Method> findAllDeclaredMethods(Class<?> clazz) {
```
dliubarskyi: Claude has found a few issues with this implementation, could you please check?

```
The most serious is that override detection keys on method name only. Two methods with the same name but different signatures (overloads) are treated as overrides of each other. If a superclass has process(String) and the subclass has process(int), the subclass version added to overridden will suppress the superclass process(String), even though it's a completely different method that nobody overrode. To detect a real override you need name + parameter types (and ignore return type, modulo covariant returns). Something like a key of name + Arrays.toString(parameterTypes).
The cache reuse path is subtly broken for exactly the same reason, plus a second one. When you hit a cached superclass and replay its methods, you filter them against overridden — but overridden only contains names seen so far in the current walk, and the cached list is itself already deduped relative to its own hierarchy. The bigger problem: the cached entry for the superclass was computed as a standalone root (it excludes nothing from below it, correctly), but you're filtering it by subclass names using name-only matching, so an overload collision between the subclass and a cached ancestor method silently drops the ancestor method. The cache merge needs the same signature-based key as the main loop, not name-only.
The double iteration over getDeclaredMethods() is wasteful but also creates an ordering subtlety: within a single class, if that class declares two methods with the same name (overloads — common), the first loop adds both because neither is in overridden yet, which is correct. Good. But it means you can't collapse the two loops naively — you'd have to be careful not to let a method suppress its own sibling overload. Right now the separation accidentally handles intra-class overloads correctly while breaking cross-class overloads. Worth a comment, or better, switch fully to signature keys where the problem disappears and you can use one loop.
Bridge and synthetic methods aren't filtered. With generics, an overriding method generates a bridge method with the same name and erased signature. If you move to signature-based keys, you may end up with both the bridge and the real method, or the bridge suppressing the real one depending on order. Filter method.isBridge() and probably isSynthetic() before adding.
Thread-safety / caching of incomplete results. computeIfAbsent on a ConcurrentHashMap doesn't protect findAllDeclaredMethods, which itself reads DECLARED_METHODS_CACHE.get(clazz) for ancestors. If two threads compute sibling subclasses sharing an uncached ancestor concurrently, that's fine functionally, but note that the ancestor only gets cached if someone calls allDeclaredMethods(ancestor) directly — the internal walk reads the cache but never populates it for intermediate classes. So the cache benefit is shallower than it looks: deep hierarchies recompute ancestors on every distinct leaf unless those ancestors are independently requested. Also, if findAllDeclaredMethods is re-entrant under computeIfAbsent for the same class across threads, ConcurrentHashMap can recompute; not a correctness bug here since the function is pure, just wasted work.
Minor: returning the cached List directly (via computeIfAbsent) hands callers a mutable ArrayList that's shared and cached. If any caller mutates it, the cache is corrupted. Wrap in List.copyOf or Collections.unmodifiableList at store time.
The cleanest fix is to key dedup on a signature record rather than name:
javarecord MethodSignature(String name, List<Class<?>> params) {}

private static List<Method> findAllDeclaredMethods(Class<?> clazz) {
    List<Method> allMethods = new ArrayList<>();
    Set<MethodSignature> seen = new HashSet<>();
    while (clazz != null && clazz != Object.class) {
        for (Method m : clazz.getDeclaredMethods()) {
            if (m.isBridge() || m.isSynthetic()) {
                continue;
            }
            MethodSignature sig = new MethodSignature(m.getName(), List.of(m.getParameterTypes()));
            if (seen.add(sig)) {
                allMethods.add(m);
            }
        }
        clazz = clazz.getSuperclass();
    }
    return List.copyOf(allMethods);
}
This handles overloads correctly (different params → different signature → both kept), real overrides correctly (same signature → child wins because it's seen first), and drops bridges. I removed the ancestor-cache merge since with the name-only bug gone it was adding complexity for limited gain; if you want it back, replay the cached list through the same seen set using MethodSignature, not names.
One thing to confirm against your actual use: do you care about default methods on interfaces? This walks getSuperclass() only, so @Tool methods declared as default on an implemented interface won't be picked up. If tools-on-interfaces is a supported pattern, you'd need to walk getInterfaces() too, which adds its own diamond-dedup considerations.
```

mariofusco: Fixed. (I did this because I thought that since the tool name has to be unique it was ok to only index on the method name, but indeed at this stage it has to find all not overridden methods in order to allow the subsequent validation to do it's work).

langchain4j/src/test/java/dev/langchain4j/service/AiServicesWithToolsIT.java
mariofusco: Done.

langchain4j-core/src/main/java/dev/langchain4j/agent/tool/ToolSpecifications.java
mariofusco: I updated documentation. 

Regarding methods discovery, including `static` ones, I didn't change nothing compared with what it did before, it only traverses parent classes now. Do you have any doubt or remark about this?

mariofusco:
@dliubarskyi I think I addressed your comments, please give it a second look.

langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
```diff
@@ -552,6 +554,37 @@ public static Optional<Method> getAnnotatedMethod(Method method, Class<? extends
         return Optional.empty();
     }
 
+    private static final Map<Class<?>, List<Method>> DECLARED_METHODS_CACHE = new ConcurrentHashMap<>();
```
dliubarskyi: Is cache really required here? There are a couple of issue with it:

- Classloader leak. It's a static map holding strong references to Class<?> keys and Method values (a Method references its declaring Class). It is never evicted. In any
  environment that loads/unloads application classloaders — servlet container redeploys, OSGi, hot reload — this pins those classes and their entire classloader in memory
  permanently. For a widely-embedded library this is a classic leak anti-pattern.
- Marginal benefit. Two of the three call sites (toolSpecificationsFrom, ToolService) run once at AiServices.build(). The only per-invocation caller is
  DefaultToolExecutor.findMethod, and walking a short class hierarchy + one getDeclaredMethods() copy is negligible next to an LLM round-trip and the actual method.invoke. The JDK
  already caches the raw method array internally; only the defensive array copy is saved.

docs/docs/tutorials/tools.md
```diff
@@ -767,6 +767,78 @@ Depending on the tool, the LLM might understand it well even without any descrip
 but it is usually better to provide clear and meaningful names and descriptions.
 This way, the LLM has more information to decide whether or not to call the given tool, and how to do so.
 
+### Inheritance and tool discovery
+
+`@Tool`-annotated methods are inherited from superclasses. When you pass a tool object to an AI Service, LangChain4j discovers `@Tool` methods from the object's class and all of its superclasses (up to, but excluding, `Object`).
```
dliubarskyi: Perhaps we should also inherit `default` `@Tool`-annotated methods from **_interfaces_** as well?

langchain4j-core/src/main/java/dev/langchain4j/internal/Utils.java
mariofusco: I'm not strongly opinionated on this, and probably you're right: there more drawback than advantages, I will remove it.

mariofusco: Done.

docs/docs/tutorials/tools.md
mariofusco: Good point, done.

mariofusco:
@dliubarskyi please check again.

dliubarskyi (approved):
@mariofusco thank you!

Issue #5313, opened 2026-05-27
[BUG] LoopPlanner runs maxIterations+1 rounds instead of maxIterations


eye-gu:
**Describe the bug**

When setting maxIterations=N on a LoopPlanner-based loop agent, the loop actually executes N+1 full rounds instead of the expected N rounds.

**Log and Stack trace**

**To Reproduce**

```java
AtomicInteger callCount = new AtomicInteger(0);
ChatModel countingModel = new ChatModel() {
    @Override
    public ChatResponse chat(ChatRequest chatRequest) {
        callCount.incrementAndGet();
        return ChatResponse.builder()
                .aiMessage(AiMessage.from("0.5"))
                .build();
    }
};

Agents.StyleScorer scorer = AgenticServices.agentBuilder(Agents.StyleScorer.class)
        .chatModel(countingModel)
        .outputKey("score")
        .build();

UntypedAgent loop = AgenticServices.loopBuilder()
        .name("testLoop")
        .subAgents(scorer)
        .maxIterations(3)
        .exitCondition("score >= 0.8", scope -> scope.readState("score", 0.0) >= 0.8)
        .build();

loop.invokeWithAgenticScope(Map.of("story", "A test story", "style", "comedy"));

// Expected: 3, Actual: 4
assertThat(callCount.get()).isEqualTo(4);
```

**Expected behavior**

**Please complete the following information:**
- LangChain4j version:
- LLM(s) used: 
- Java version: 
- Spring Boot version (if applicable): 

**Additional context**

If this is the intended behavior, please feel free to close this issue. Otherwise, I'd be happy to submit a PR to fix this. The fix would likely involve adjusting the initial value of iterationsCounter or changing the comparison operator from > to >=.

https://github.com/langchain4j/langchain4j/blob/ed861e6c2d534a1c9e4e529d395b4af91c787e27/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/workflow/impl/LoopPlanner.java#L45-L56

PR #5330, opened 2026-05-30
fix(agentic): fix off-by-one in LoopPlanner iterations

Baqirrizvidev:
Fixes an off-by-one error in LoopPlanner that caused loops to execute maxIterations + 1 times.

This is a clean, focused fix that directly addresses the bug described in #5313 and adds a dedicated unit test (LoopPlannerTest) to explicitly verify that the loop ends after exactly maxIterations.


mariofusco:
Fixed by https://github.com/langchain4j/langchain4j/pull/5323

PR #5337, opened 2026-06-01
fix: add null check for chatResponse.usage() in Watsonx models

### LINKED ISSUE #5317: [BUG] NullPointerException in the Watsonx integration when chatResponse.usage() returns null
I found a NullPointerException in the Watsonx integration when chatResponse.usage() returns null:
Cannot invoke \"com.ibm.watsonx.ai.chat.model.ChatUsage.promptTokens()\" because the return value of \"com.ibm.watsonx.ai.chat.ChatResponse.usage()\" is null

This happens in both WatsonxChatModel.java:110 and WatsonxStreamingChatModel.java:97-100 where we call usage.promptTokens() without null-checking first.

Quick fix:

> ChatUsage usage = chatResponse.usage();
> TokenUsage tokenUsage = usage != null 
>     ? new TokenUsage(usage.promptTokens(), usage.completionTokens(), usage.totalTokens())
>     : new TokenUsage(0, 0, 0);

The Watsonx AI SDK can return null usage in certain scenarios (some model configs, specific endpoints).

EvanYao826:
Fixes #5317

## Changes

Added null check for  in both  and  to prevent  when the usage field is null.

### DIFF
--- a/langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
+++ b/langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxChatModel.java
@@ -107,7 +107,9 @@ public class WatsonxChatModel extends WatsonxChat implements ChatModel {
         }
 
         FinishReason finishReason = Converter.toFinishReason(choice.finishReason());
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@EvanYao826 thanks!

PR #5341, opened 2026-06-01
feat: de-duplicate overlap in sequential TextSegments (#1323)

### LINKED ISSUE #1323: [FEATURE] De-duplicate overlap in sequential TextSegments
When documents are split with an overlap, retrieved sequential `TextSegment`s should be de-duplicated

sameer-sde:
## Summary

Closes #1323

When documents are split with overlap, adjacent retrieved `TextSegment`s from the same document share a repeated suffix/prefix. This causes the LLM to receive duplicate content, wasting context and reducing answer quality.

## Changes

- Added `removeDuplicateOverlap` option (default: `true`) to `EmbeddingStoreContentRetriever` and its builder
- Implemented `deduplicateOverlap()` — groups segments by document identity (all metadata except `"index"`), detects adjacent pairs (index N and N+1), strips overlap from the start of the later segment
- Implemented `longestSuffixPrefixOverlap()` static helper using `String.regionMatches` for efficient comparison

## Example

```java
// Segment 0: "the quick brown fox"
// Segment 1: "brown fox jumps over"  <- "brown fox" is overlap

EmbeddingStoreContentRetriever retriever = EmbeddingStoreContentRetriever.builder()
    .removeDuplicateOverlap(true)  // default, can disable with false
    .build();

// Result segment 1: " jumps over"  <- overlap removed
```

## Tests

9 new unit tests covering:
- Overlap removed from adjacent segments
- Non-adjacent segments left unchanged
- Segments from different documents left unchanged
- Segments with no index metadata left unchanged
- Single segment returned unchanged
- No overlap — segments unchanged
- Metadata preserved after deduplication
- `longestSuffixPrefixOverlap` with overlap
- `longestSuffixPrefixOverlap` with no overlap
- `longestSuffixPrefixOverlap` with full overlap

## Checklist
- [x] No breaking changes — opt-in default, existing behaviour preserved
- [x] Unit tests added covering positive and negative cases
- [x] All existing tests pass

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/EmbeddingStoreContentRetriever.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/EmbeddingStoreContentRetriever.java
@@ -367,7 +367,8 @@ public class EmbeddingStoreContentRetriever implements ContentRetriever {
                     }
                     TextSegment newSegment = TextSegment.from(
                             deduplicated, next.getValue().textSegment().metadata());
-                    Content newContent = Content.from(newSegment, next.getValue().metadata());
+                    Content newContent =
+                            Content.from(newSegment, next.getValue().metadata());
                     replacements.put(next.getValue(), newContent);
                 }
             }
@@ -377,9 +378,7 @@ public class EmbeddingStoreContentRetriever implements ContentRetriever {
             return contents;
         }
 
-        return contents.stream()
-                .map(c -> replacements.getOrDefault(c, c))
-                .collect(Collectors.toList());
+        return contents.stream().map(c -> replacements.getOrDefault(c, c)).collect(Collectors.toList());
     }
 
     /**
--- a/langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/EmbeddingStoreContentRetriever.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/EmbeddingStoreContentRetriever.java
@@ -17,9 +17,12 @@ import dev.langchain4j.store.embedding.EmbeddingSearchRequest;
 import dev.langchain4j.store.embedding.EmbeddingSearchResult;
 import dev.langchain4j.store.embedding.EmbeddingStore;
 import dev.langchain4j.store.embedding.filter.Filter;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -53,12 +56,20 @@ import java.util.stream.Collectors;
  * - {@code dynamicFilter}: It is a {@link Function} that accepts a {@link Query} and returns a {@code filter} value.
  * It can be used to dynamically define {@code filter} value, depending on factors such as the query,
  * the user (using Metadata#chatMemoryId()} from {@link Query#metadata()}), etc.
+ * <br>
+ * - {@code removeDuplicateOverlap}: When documents are split with overlap, adjacent retrieved {@link TextSegment}s
+ * from the same document share a repeated suffix/prefix. When {@code true} (default), this duplicate overlap
+ * text is detected and removed from the start of the later segment, preventing the LLM from receiving redundant
+ * content. Set to {@code false} to disable this behaviour and return segments as-is.
  */
 public class EmbeddingStoreContentRetriever implements ContentRetriever {
+
+    static final String INDEX_METADATA_KEY = "index";
 
     public static final Function<Query, Integer> DEFAULT_MAX_RESULTS = (query) -> 3;
     public static final Function<Query, Double> DEFAULT_MIN_SCORE = (query) -> 0.0;
     public static final Function<Query, Filter> DEFAULT_FILTER = (query) -> null;
+    public static final boolean DEFAULT_REMOVE_DUPLICATE_OVERLAP = true;
 
     public static final String DEFAULT_DISPLAY_NAME = "Default";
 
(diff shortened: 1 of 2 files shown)


fw-coder-tl:
Hi, thanks for working on this PR!

I am new to LangChain4j and currently learning the RAG part of the project. I noticed this PR is addressing #1323, which I am interested in.

Would it be helpful if I tried this locally and checked whether there are any missing edge cases or tests around sequential `TextSegment` overlap de-duplication?

I would be happy to help with small follow-up changes if needed.

sameer-sde:
@fw-coder-tl Thanks for your interest! Feel free to test it locally  any feedback on edge cases or missing tests would be appreciated. Happy to make changes based on your findings.

fw-coder-tl:
I found one design concern around retrieval order vs document order.

EmbeddingStore.search(...) usually returns matches in relevance-score order, but deduplicateOverlap() groups contents by document metadata and processes adjacent indexes in document order. It then replaces the later segment while preserving the original retrieval order.

This means a higher-scored segment can be stripped even when its predecessor appears later in the returned list.

For example:


```java
Content c0 = Content.from(TextSegment.from("the quick brown fox", Metadata.from("index", "0")));
Content c1 = Content.from(TextSegment.from("brown fox jumps over", Metadata.from("index", "1")));

List<Content> result = EmbeddingStoreContentRetriever.deduplicateOverlap(asList(c1, c0));
```

Currently `result.get(0)` becomes `" jumps over"`, even though it is first in retrieval order. That can send a context fragment to the LLM before its preceding chunk.

Should deduplication only happen when the retrieved order is also document order, or should same-document chunks be reordered before stripping overlap?

I also added a few local edge-case tests for invalid index metadata, Chinese overlap, identical segments, empty/short text, and repeated text that is not suffix/prefix overlap; those pass locally.

sameer-sde:
@fw-coder-tl thanks for digging into this so thoroughly  that's a really good catch.

You're right that `deduplicateOverlap()` currently assumes retrieval order roughly aligns with document order for adjacent segments, and your example shows that's not guaranteed. Stripping the overlap from a higher-scored segment because a lower-scored predecessor appears later in the list could definitely send a confusing fragment to the LLM.

I think the safest fix for this PR is to only perform deduplication when the retrieval order for that pair actually matches document order (i.e., index N appears before index N+1 in the result list) and leave segments unchanged otherwise. Reordering same-document chunks before stripping is an interesting idea too, but that changes relevance ranking and feels like it deserves its own discussion/PR.

Could you share the edge-case tests you added locally (Chinese overlap, invalid index metadata, identical segments, etc.)? I'd like to incorporate those here along with a fix for the ordering issue you found.

Thanks again for the careful review!

fw-coder-tl:
Sure. I pushed the tests here:

https://github.com/fw-coder-tl/langchain4j/commits/overlap-dedup-edge-tests

There are two commits:
- `1e6a68db9`: passing edge-case tests for invalid index metadata, repeated non-suffix/prefix text, Chinese overlap, identical segments, and empty/short text.
- `75e94ce96`: a currently failing regression test for the retrieval-order/document-order mismatch.

The first commit passes locally. The second commit fails as expected with:
`expected: "brown fox jumps over" but was: " jumps over"`.

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/EmbeddingStoreContentRetriever.java
```diff
@@ -236,13 +286,123 @@ public List<Content> retrieve(Query query) {
...
+     *   result = 9  ("brown fox")
+     * </pre>
+     *
+     * @param a the earlier segment text
+     * @param b the later segment text
+     * @return the length of the longest overlapping suffix/prefix, or 0 if none
+     */
+    static int longestSuffixPrefixOverlap(String a, String b) {
+        int maxOverlap = Math.min(a.length(), b.length());
+        for (int len = maxOverlap; len > 0; len--) {
```
LD-RW, line 400: **thought (non-blocking):** Consider using the KMP algorithm to optimize worst-case time complexity.

Hi! I've been reading through the PR and had an idea. I'm not entirely sure if it's the most efficient route for our specific use case, but I wanted to share it anyway for discussion.

Currently, finding the longest suffix of `a` that equals a prefix of `b` tries every candidate length from $L = \min(|a|, |b|)$ down to $1$. Since each `regionMatches(..., len)` compares up to `len` characters, the worst-case time complexity is $O(L^2)$ (i.e., $L + (L-1) + \dots + 1$).

While that's the worst-case scenario and won't always happen, we could use the **KMP Algorithm** here to guarantee linear time complexity—exactly $O(N + M)$.

**The Trade-off:**

* **Current approach:** $O(L^2)$ worst-case time, but highly efficient $O(1)$ space complexity.
* **KMP approach:** $O(N + M)$ guaranteed time, but requires $O(N + M)$ linear space complexity.

I don't know if it's better for our overall design to accept the KMP memory hit, or if we should just keep the current approach and accept the potential $O(L^2)$ worst-case time to save space. What do you think?

sameer-sde:
@fw-coder-tl Thank you for pushing those tests  really appreciate the effort! The first commit's edge cases are great additions.
The failing test confirms a real issue with the current implementation. When retrieval returns segments out of document order (e.g. [c1, c0]), deduplication should not strip overlap from c1 just because it has a higher document index that would corrupt a higher-scored, more relevant segment.
The fix should be: only strip overlap between two segments when they also appear in document order in the retrieved list (i.e. the segment with the lower index comes before the segment with the higher index in the result). If they're retrieved in reverse document order, leave them untouched.
I'll update the implementation and add your failing test as a regression test. Thanks again for catching this!

langchain4j-core/src/main/java/dev/langchain4j/rag/content/retriever/EmbeddingStoreContentRetriever.java
sameer-sde: That's a great point! You're absolutely right about the theoretical worst-case complexity difference.
For our use case though, TextSegments in RAG pipelines are typically short (usually a few hundred characters at most), so the O(L²) worst case is unlikely to be hit in practice. Given that, I think the current approach's simplicity and O(1) space complexity is a better fit here than introducing KMP's additional complexity and O(N+M) memory overhead.
That said, if profiling ever shows this to be a bottleneck, KMP would be a great optimization to revisit. Thanks for raising it!

PR #5345, opened 2026-06-02
MCP tool guardrails

jmartisk:
Guardrails for tool calls over MCP. See the new documentation piece for the explanation.

Intentionally independent of other chat/tool guardrails (at first I thought they should be unified, but MCP is just too different)

On the Quarkus side, I imagine that the user will annotate a class that implements `McpToolInputGuardrail` or `McpToolOutputGuardrail` with `@McpClientName("client1")` to apply it to a specific client

CC @cescoffier @geoand @mariofusco


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/guardrail/McpToolGuardrailException.java
```diff
@@ -0,0 +1,16 @@
+package dev.langchain4j.mcp.client.guardrail;
+
+/**
+ * Thrown by an MCP tool guardrail to reject a tool invocation.
+ * The exception message is returned to the LLM as the tool error.
+ */
+public class McpToolGuardrailException extends RuntimeException {
```
dliubarskyi:
```suggestion
public class McpToolGuardrailException extends LangChain4jException {
```

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/guardrail/McpToolGuardrailException.java
```diff
@@ -0,0 +1,16 @@
+package dev.langchain4j.mcp.client.guardrail;
+
+/**
+ * Thrown by an MCP tool guardrail to reject a tool invocation.
+ * The exception message is returned to the LLM as the tool error.
```
dliubarskyi: Wouldn't it be more flexible for the input guardrail to return `ToolExecutionResult` instead of throwing an exception? User would be also able to to control `isError` flag and pass `attributes`

dliubarskyi: Perhaps `McpToolInputGuardrailResult.proceeed()/success()` to proceed with calling the tool and `McpToolInputGuardrailResult.reject(ToolExecutionResult )` to block/reject tool call?

dliubarskyi:
> Intentionally independent of other chat/tool guardrails (at first I thought they should be unified, but MCP is just too different)

Could you please elaborate on how MCP tool call is too different from the regular tool call? From the code it seems that the only difference is presence of `McpClient` in the context, but the rest of the logic could have easily be applied to the regular tool call.

docs/docs/tutorials/mcp.md
```diff
@@ -581,6 +581,72 @@ ToolExecutionRequest request = ToolExecutionRequest.builder()
...
+  reject the call, for example to block requests with forbidden content.
+- **Output guardrails** run after the MCP server returns a result. They can reject
+  the result or transform it, for example to redact sensitive data.
+
+Guardrails are configured on the `DefaultMcpClient.Builder`:
+
+```java
+McpClient mcpClient = new DefaultMcpClient.Builder()
+    .transport(transport)
+    .inputGuardrails(List.of(forbiddenWordGuardrail))
```
dliubarskyi, line 604: Should it be `inputToolGuardrails` to be more specific? I guess there could be other types of input guardrails?

dliubarskyi: But in general I am wondering if `McpClient` is the right place to be concerned with guardrails. It seems that `McpToolExecutor` and `McpToolProvider` are more suitable as they are specific to tools

dliubarskyi: But ofc in case tool guardrails will become MCP-agnostic, this comment should be ignored

jmartisk:
> > Intentionally independent of other chat/tool guardrails (at first I thought they should be unified, but MCP is just too different)
> 
> Could you please elaborate on how MCP tool call is too different from the regular tool call? From the code it seems that the only difference is presence of `McpClient` in the context, but the rest of the logic could have easily be applied to the regular tool call.

A `ToolInputGuardrailRequest` also contains `ToolMetadata` and `ToolInvocationContext` (https://github.com/quarkiverse/quarkus-langchain4j/blob/1.11.0.CR1/core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/ToolInputGuardrailRequest.java#L47) - these don't exist for MCP tools.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/guardrail/McpToolGuardrailException.java
jmartisk: Hm you're right, I will align it more with how the Quarkus tool guardrail implementation does this.

docs/docs/tutorials/mcp.md
jmartisk: > But in general I am wondering if `McpClient` is the right place to be concerned with guardrails. It seems that `McpToolExecutor` and `McpToolProvider` are more suitable as they are specific to tools

A MCP client can also be used directly without a `McpToolExecutor` so the guardrails wouldn't work in this case. And a `McpToolProvider` only provides metadata, it doesn't come into play when executing tools

dliubarskyi: > A MCP client can also be used directly without a McpToolExecutor so the guardrails wouldn't work in this case.

If user works with MCP client directly, then they have a quite specific use case, I would not bet on this being the major use case. IMHO `McpClient` is already too big and has too many responsibilities, but it's your call.

> And a McpToolProvider only provides metadata, it doesn't come into play when executing tools

`McpToolProvider` creates `McpToolExecutor`, which could be responsible for executing guardrails around mcp client call

dliubarskyi:
> A ToolInputGuardrailRequest also contains ToolMetadata and ToolInvocationContext (https://github.com/quarkiverse/quarkus-langchain4j/blob/1.11.0.CR1/core/runtime/src/main/java/io/quarkiverse/langchain4j/guardrails/ToolInputGuardrailRequest.java#L47) - these don't exist for MCP tools.

`ToolMetadata` contains `ToolSpecification` which is relevan for MCP as well. `Object toolInstance` could be `null` for MCP or contain `McpClient` instance. `ToolInvocationContext` contains `InvocationContext` which is relevant for MCP as well.

---

IMHO MCP tool calling is not conceptually different from regular tool calling, so I would try to implement a generic tool guardrails mechanism that can be customized for MCP case. Otherwise users will have to implement 2 sets of guardrails: one for regular tools, another for MCP tools. And we will have to maintain 2 independent mechanisms

jmartisk:
That's a matter of opinion then, all the differences that we've listed here persuade me that it's conceptually quite different :)
I would also say guardrails for tools that are known at build time will generally be quite different from guardrails for something that isn't known at build time at all and can be added dynamically at runtime.

`ToolInvocationContext` does contain `InvocationContext`, but `ToolInvocationContext` itself doesn't exist for MCP, and it contains some other stuff that isn't relevant (arguments as Java key-value pairs instead of JSON).

Well I'm not strictly opposed to somehow merging it with regular tool guardrails if you insist, but that lives in Quarkus at the moment, so this would have to wait, and be coordinated with moving those to langchain4j... (in the past I looked at implementing this in Quarkus and merging it with its implementation of tool guardrails, but it was entangled even more)

Maybe @cescoffier could share his view :)

dliubarskyi:
> all the differences that we've listed here persuade me that it's conceptually quite different

IIUC, the differences are that some MCP-specific context objects are present and some non-MCP-specific context objects are not present. Those are details IMHO. The execution flow and the shape of the API is the same though, this is what I mean by "conceptually".

> I would also say guardrails for tools that are known at build time will generally be quite different from guardrails for something that isn't known at build time at all and can be added dynamically at runtime.

Why so? From user point of view (I understand that it is different from impl point of view)

> ToolInvocationContext does contain InvocationContext, but ToolInvocationContext itself doesn't exist for MCP, and it contains some other stuff that isn't relevant (arguments as Java key-value pairs instead of JSON).

If I read it correctly, `ToolInvocationContext` contains only `InvocationContext`

jmartisk:
> > I would also say guardrails for tools that are known at build time will generally be quite different from guardrails for something that isn't known at build time at all and can be added dynamically at runtime.
> 
> Why so? From user point of view (I understand that it is different from impl point of view)

That was the response to "users will have to implement 2 sets of guardrails" and why I think that's not a real problem because these will quite possibly be different. 

> 
> > ToolInvocationContext does contain InvocationContext, but ToolInvocationContext itself doesn't exist for MCP, and it contains some other stuff that isn't relevant (arguments as Java key-value pairs instead of JSON).
> 
> If I read it correctly, `ToolInvocationContext` contains only `InvocationContext`

Ah right, I confused it a bit, it's actually info about the AI service invocation, not the tool invocation. And if it only contains the `InvocationContext`, so thinking about it again, I  don't quite understand what the point of that record is. The `Tool` in the name only makes it even more confusing. I think we may want to get rid of it as part of moving the guardrails to langchain4j. That would make the APIs for local tool guardrails and MCP guardrails more similar. 

Hm, I see one more slight problem is that `ToolMetadata` contains `ToolSpecification` and to ensure that we have an up-to-date specification, we would have to retrieve the tool list again. And worse, if tool list caching is disabled, then we would always have to retrieve a fresh tool list to get the specification. Perhaps with caching enabled, we could have a caveat that it's not 100% guaranteed the specification would be up-to-date, and with caching disabled, we simply would pass `null` as the specification.

Anyway if we do want to use the unified API, then this depends on moving tool guardrails to langchain4j.

And, uhm, one more thing coming to my mind. Should we extend this to resources and prompts too? Because then it would really be different from local tool guardrails. But I don't know if that makes sense at all.

dliubarskyi:
@jmartisk sorry, I totally forgot about this!

> Anyway if we do want to use the unified API, then this depends on moving tool guardrails to langchain4j.

It depends on the urgency, if you think these guardrails are needed ASAP, let's merge it  as-is and fix the potential duplication in 2.0.
Also, does MCP 2.0 affect it in any way?

> Should we extend this to resources and prompts too? Because then it would really be different from local tool guardrails. But I don't know if that makes sense at all.

What do you mean? Guardrails for resources and prompts? Is it a thing?

jmartisk:
> @jmartisk sorry, I totally forgot about this!
> 
> > Anyway if we do want to use the unified API, then this depends on moving tool guardrails to langchain4j.
> 
> It depends on the urgency, if you think these guardrails are needed ASAP, let's merge it as-is and fix the potential duplication in 2.0. Also, does MCP 2.0 affect it in any way?

I don't think it's super urgent, I don't remember having seen any users explicitly complaining that they don't have guardrails for MCP tools (yet).

The next MCP spec version shouldn't have much effect on this.

> 
> > Should we extend this to resources and prompts too? Because then it would really be different from local tool guardrails. But I don't know if that makes sense at all.
> 
> What do you mean? Guardrails for resources and prompts? Is it a thing?

Well it could be a thing if we decide that it should be :) that's why I'm asking. Maybe it's a nonsense.

dliubarskyi:
> Well it could be a thing if we decide that it should be :) that's why I'm asking. Maybe it's a nonsense.

As far as I understand, LLMs can only call MCP tools at the moment. Resources and prompts can be accessed either directly via `McpClient` (the need for guardrails is questionable here) or decorated as tools (could be covered by tool guardrails if they would be on the `ToolExecutor` level)

PR #5346, opened 2026-06-02
Fix JSON extraction with braces in string values

### LINKED ISSUE #5343: [BUG] JsonParsingUtils fails to extract JSON when string values contain } or ] with surrounding text
**Describe the bug**

`JsonParsingUtils.extractAndParseJson(...)` fails to extract valid JSON from LLM output when the response contains surrounding non-JSON text and a JSON string value contains a closing brace `}`.

For example, this is a realistic LLM response when summarizing or extracting information from logs:

```text
Sure, here is the JSON:
{"log":"Unexpected character '}' at position 42"}
```

The JSON object itself is valid, but the extraction logic can treat the `}` inside the `"log"` string as a structural JSON delimiter.

**To Reproduce**

A minimal unit test can reproduce the issue:

```java
@Test
void extract_object_with_closing_brace_in_string_and_prefix() throws Exception {
    String text = """
            Sure, here is the JSON:
            {"log":"Unexpected character '}' at position 42"}
            """;

    JsonParsingUtils.ParsedJson<LogEntry> result =
            JsonParsingUtils.extractAndParseJson(text, LogEntry.class);

    assertThat(result.value().log).isEqualTo("Unexpected character '}' at position 42");
}

static class LogEntry {
    public String log;
}
```

**Expected behavior**

`JsonParsingUtils.extractAndParseJson(...)` should ignore braces inside quoted JSON strings and correctly extract the valid JSON object from the surrounding LLM text.

* LLM(s) used: MiMo
* Java version: 17
* Spring Boot version (if applicable): N/A

**Additional context**

This can affect POJO output parsing and JSON guardrails when an LLM returns explanatory text before JSON. This is common in real LLM responses, especially when extracting structured data from logs, code snippets, or error messages.

The issue appears to be caused by JSON boundary detection that scans for braces without accounting for quoted strings and escape sequences.

I would like to work on this issue and submit a PR. Could you please assign it to me if this sounds like a valid bug?

SPUERSAIYAN:
## Issue

Closes #5343

## Change

Fixes JSON extraction from LLM-style outputs that contain text before or after the JSON block when quoted JSON string values include structural-looking characters such as `}` or `]`.

Previously, `JsonParsingUtils.extractAndParseJson(...)` searched for closing braces/brackets and matched braces without tracking whether the scanner was inside a JSON string. For example, this valid JSON could fail to be extracted when surrounded by text:

```text
Sure, here is the JSON:
{"log":"Unexpected character '}' at position 42"}
```

The extraction logic now scans candidate JSON blocks with a stack-based matcher that respects quoted strings and escaped characters, so `}` and `]` inside string values are ignored as JSON structure.

Added unit tests covering:

* object extraction when a string value contains `}` and the JSON has prefix/suffix text
* array extraction when a string value contains `]` and the JSON has prefix/suffix text

## General checklist

* [x] There are no breaking changes (API, behaviour)
* [x] I have added unit and/or integration tests for my change
* [x] The tests cover both positive and negative cases
* [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
* [x] I have manually run all the unit and integration tests in the core and main modules, and they are all green

* [ ] I have added/updated the documentation
* [ ] I have added an example in the examples repo (only for "big" features)
* [ ] I have added/updated Spring Boot starter(s) (if applicable)

* [ ] I have added my new module in the root pom.xml and langchain4j-bom/pom.xml

* [ ] I have added a {NameOfIntegration}EmbeddingStoreIT that extends from either EmbeddingStoreIT or EmbeddingStoreWithFilteringIT
* [ ] I have added a {NameOfIntegration}EmbeddingStoreRemovalIT that extends from EmbeddingStoreWithRemovalIT

* [ ] I have manually verified that the {NameOfIntegration}EmbeddingStore works correctly with the data persisted using the latest released version of LangChain4j

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/JsonParsingUtils.java
@@ -1,6 +1,10 @@
 package dev.langchain4j.internal;
 
 import dev.langchain4j.Internal;
+import java.util.ArrayDeque;
+import java.util.ArrayList;
+import java.util.Deque;
+import java.util.List;
 
 @Internal
 public class JsonParsingUtils {
@@ -16,7 +20,8 @@ public class JsonParsingUtils {
         return extractAndParseJson(text, s -> Json.fromJson(s, type));
     }
 
-    public static <T> ParsedJson<T> extractAndParseJson(String text, ThrowingFunction<String, T> parser) throws Exception {
+    public static <T> ParsedJson<T> extractAndParseJson(String text, ThrowingFunction<String, T> parser)
+            throws Exception {
         Exception parseException = null;
         try {
             return new ParsedJson<>(parser.apply(text), text);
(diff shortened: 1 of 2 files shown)


dliubarskyi:
THis has been fixed in #5354

PR #5350, opened 2026-06-02
feat: support lenient mode for unresolved template variables

### LINKED ISSUE #3197: [FEATURE] Support disabling template injection
When using the AiService, all user messages will be used as a template.
This cause issues if the message is not a template but still contains template variables.

Having a way to specify that a message is not a template, or having the template mechanism not failing if it cannot find a parameter.

None.

**Additional context**
A way to choose the template separator would also work even if more fragile.

eye-gu:
## Issue

Closes #3197

## Change

Support lenient mode for prompt template resolution. When enabled, unresolved template variables (e.g. {{unknown}}) are kept as-is instead of throwing an exception.

This is useful when a message is not intended as a template but happens to contain text that uses the {{...}} syntax (e.g. code examples, Mustache/Handlebars snippets).

## General checklist


eye-gu:
@dliubarskyi PTAL🙏

dliubarskyi (commented):
@eye-gu thank you, almost there!

1. Gap: the most common case isn't covered by the builder flag. In getUserMessageTemplateAndLenient, the "only argument" path (e.g. String chat(String message) — the single most common AiService shape, and
  exactly what #3197 describes) hardcodes lenient = false:
  if (templateFromTheOnlyArgument.isPresent())
      return new TemplateAndLenient(templateFromTheOnlyArgument.get(), false);
So AiServices.builder()...userMessageLenient(true) has no effect on chat(String). The only way to make that path lenient is @UserMessage(lenient = true) on the parameter. That's a real inconsistency worth
  raising with the author — either the builder flag should apply here too, or the limitation should be documented.
  2. Minor semantic: lenient also swallows the provided-but-explicitly-null case (previously "Value for the variable 'x' is null"), keeping the placeholder. Defensible, but it's slightly broader than "unresolved
  variable." Worth a one-line confirmation that this is intended.

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -303,6 +303,46 @@ public AiServices<T> systemMessageProviderWithContext(Function<InvocationContext
...
+     * <br>
+     * This only applies when the system message is provided via {@link #systemMessage(String)} or
+     * {@link #systemMessageProvider(Function)}. When {@code @SystemMessage} annotation is present,
+     * the annotation's own {@code lenient} attribute is used instead, and this setting has no effect.
+     * <p>
+     * Defaults to {@code false}.
+     *
+     * @param systemMessageLenient whether to use lenient mode for system message templates
+     * @return builder
+     * @since 1.12.0
```
```suggestion
     * @since 1.17.0
```

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -303,6 +303,46 @@ public AiServices<T> systemMessageProviderWithContext(Function<InvocationContext
...
+     * <br>
+     * This only applies when the user message is provided via {@link #userMessage(String)} or
+     * {@link #userMessageProvider(Function)}. When {@code @UserMessage} annotation is present,
+     * the annotation's own {@code lenient} attribute is used instead, and this setting has no effect.
+     * <p>
+     * Defaults to {@code false}.
+     *
+     * @param userMessageLenient whether to use lenient mode for user message templates
+     * @return builder
+     * @since 1.12.0
```
```suggestion
     * @since 1.17.0
```

dliubarskyi:
Please document this feature

eye-gu:
1. I switched it to use context.userMessageLenient.Verified that in this path the variables map only ever contains the parameter name (no value for an arbitrary {{x}} the user typed), so with lenient on the placeholder is preserved verbatim — which is the intended behavior here.
2. Yes. In lenient mode a null value is treated the same as any other unresolved variable — the placeholder is left as-is, which is consistent with the lenient contract ("don't fail, just keep the literal text").
3. Documentation Added a "Lenient Template Resolution" section to ai-services.md covering both the annotation attribute and the builder methods, with the annotation taking precedence.

PR #5355, opened 2026-06-03
Remove default fallback to Netty when a custom HttpClientProvider is provided for Azure OpenAI integration.

### LINKED ISSUE #5352: [FEATURE] azure-core-http-netty still required on classpath even when custom httpClientProvider is set
PR #2822 (which closed #2799) added `.httpClientProvider(...)` so users can replace the default Netty transport. The builder wiring works correctly — our custom provider is called. But `azure-core-http-netty` still **cannot be excluded from the classpath**, which means the dependency is not actually optional.

  ## Root cause

  In `InternalAzureOpenAiHelper.setupOpenAIClientBuilder`, the fallback Supplier is created **unconditionally** — before `getOrDefault` can check whether a custom provider was supplied:

  ```java
  // current — Supplier created unconditionally, forces class resolution
  HttpClientProvider resolved = Utils.getOrDefault(
      httpClientProvider,
      NettyAsyncHttpClientProvider::new  // ← evaluated before the null-check
  );

  Because NettyAsyncHttpClientProvider::new is a method reference, the JVM must resolve NettyAsyncHttpClientProvider the first time this line is executed — regardless of whether httpClientProvider is already set. If azure-core-http-netty is excluded,
  the result is:

  java.lang.NoClassDefFoundError: com/azure/core/http/netty/NettyAsyncHttpClientProvider
      at dev.langchain4j.model.azure.InternalAzureOpenAiHelper.setupOpenAIClientBuilder
      at dev.langchain4j.model.azure.InternalAzureOpenAiHelper.setupSyncClient
      at dev.langchain4j.model.azure.AzureOpenAiChatModel.<init>

  Proposed fix

  Only reference NettyAsyncHttpClientProvider inside the fallback branch, so it's never resolved when a custom provider is supplied:

  HttpClientProvider resolved = httpClientProvider != null
      ? httpClientProvider
      : loadDefaultProvider();  // e.g. ServiceLoader.load(HttpClientProvider.class).findFirst()
                                 // or Class.forName("...NettyAsyncHttpClientProvider") lazily

  ServiceLoader is already how Azure SDK itself discovers HTTP client providers — using it here would make azure-core-http-netty a genuine optional dependency.

  Version: langchain4j-azure-open-ai 1.15.1
  ```

nhthinh-axonivy:
## Issue
  Closes #5352

  ## Change

  Makes `azure-core-http-netty` optional when a custom `HttpClientProvider` is supplied via the builder.

  ### Details

  - Removed hard constructor reference `NettyAsyncHttpClientProvider::new` from `InternalAzureOpenAiHelper.setupOpenAIClientBuilder` — previously baked into the constant pool as a `MethodHandle`, causing `NoClassDefFoundError` at method invocation time even when a custom provider
  was set
  - Replaced with a lazy `ServiceLoader` lookup that only runs when no custom provider is supplied
  - Added package-private overload `loadDefaultHttpClientProvider(ClassLoader)` for clean testability without reflection or thread manipulation
  - Added tests covering: ServiceLoader discovery when Netty is on classpath, custom provider is used and default is skipped, and `IllegalStateException` when no provider is found

  `azure-core-http-netty` already registers `NettyAsyncHttpClientProvider` via `META-INF/services/com.azure.core.http.HttpClientProvider`, so default behaviour is unchanged.

  ## General checklist


dliubarskyi (approved):
@nhthinh-axonivy thank you!

PR #5358, opened 2026-06-04
feat(bedrock): add enableAdaptiveReasoning for Claude Opus 4.7+

### LINKED ISSUE #5138: [FEATURE] Support adaptive thinking for new Amazon Bedrock Claude models (e.g., Claude Opus 4.7)
I am encountering a hard blocker when trying to use the new **Claude Opus 4.7** model via the Bedrock module, and facing an impending deprecation for **Opus 4.6** and **Sonnet 4.6**.

Starting with Opus 4.7, manual thinking configuration (`type: "enabled"` with `budget_tokens`) is no longer supported; adaptive thinking is the *only* supported mode. Because the current `langchain4j-bedrock` implementation relies on the manual approach, trying to use Opus 4.7 results in a `400 Bad Request`:

> `Unable to establish a connection to Thinking LLM: The model returned the following errors: "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior. (Service: BedrockRuntime, Status Code: 400, Request ID: 27f93dbb-3113-48a4-a82c-4161d868f5dd)`

While Opus 4.6 and Sonnet 4.6 currently still tolerate the existing `budget_tokens` configuration, this approach is officially deprecated for those models as well, and Anthropic/AWS advises planning a migration before it is removed in future releases.

I would like the `langchain4j-bedrock` module to be updated to support the new **Adaptive Thinking** configuration, ensuring compatibility with Opus 4.7 and future-proofing the 4.6 models. Specifically:

1.  **Support for `adaptive` thinking type:** Allow setting `thinking.type` to `adaptive` instead of defaulting to `enabled`.
2.  **Add `effort` parameter:** Introduce support for the `output_config.effort` parameter (or top-level depending on the Bedrock payload structure) to control reasoning intensity.
3.  **Graceful handling:** Allow users to define the thinking strategy (manual with budget vs. adaptive with effort) so that the integration works seamlessly across older and newer model versions.

I think that is possible to have a workaround with the `additionalModelRequestFields`, but clients would need to avoid using the `enableThinking`  method, that is coupled with the other (non adaptive) implementation.

I have considered manually intercepting and modifying the JSON request payload before it hits the Bedrock client. However, because the `thinking` block is tightly coupled to the internal request builders in `langchain4j-bedrock`, there is no clean, out-of-the-box way to inject `thinking.type: "adaptive"` and the `effort` parameter without forking or bypassing the provided integration entirely. 

**Additional context**
According to the latest Anthropic/Bedrock documentation regarding thinking parameters:
* **Opus 4.7:** Manual thinking is entirely unsupported. It requires `thinking.type: "adaptive"`.
* **Opus 4.6 & Sonnet 4.6:** `thinking.type: "enabled"` and `budget_tokens` are deprecated and functional for now, but not recommended. The migration path requires shifting to `thinking.type: "adaptive"` with an associated `effort` parameter.

https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking

Aditya-Pimpalkar:
## Issue
Closes #5138

## Change
Adds `BedrockChatRequestParameters.Builder.enableAdaptiveReasoning(String effort)`,
which serializes to `reasoning_config: {type: "adaptive"}` and, when effort is
non-null, `output_config: {effort: "..."}`. This is the payload @jfigueiras-denodo
confirmed working against Claude Opus 4.7 in #5138, where the legacy `budget_tokens`
reasoning config now returns a 400.

The existing `enableReasoning(Integer budgetTokens)` is left unchanged, so
Opus 4.6 / Sonnet 4.6 users keep working — matching the deprecation/migration path
described in the issue. Added two unit tests covering adaptive-with-effort and
adaptive-without-effort (null) serialization.

## General checklist

### DIFF
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatRequestParameters.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/BedrockChatRequestParameters.java
@@ -131,6 +131,31 @@ public class BedrockChatRequestParameters extends DefaultChatRequestParameters {
         }
 
         /**
+         * Enables <a href="https://docs.aws.amazon.com/bedrock/latest/userguide/inference-reasoning.html">adaptive reasoning</a>,
+         * required for Claude Opus 4.7+ where the legacy {@code budget_tokens} reasoning configuration is no longer accepted.
+         * Older models (e.g. Claude Opus 4.6, Sonnet 4.6) continue to work with {@link #enableReasoning(Integer)}.
+         *
+         * @param effort controls reasoning intensity, serialized to Bedrock's {@code output_config.effort} field.
+         *               Accepted values: {@code "low"}, {@code "medium"}, {@code "high"}.
+         *               If {@code null}, only {@code reasoning_config.type = "adaptive"} is set and Bedrock applies its default.
+         * @see BedrockChatModel.Builder#returnThinking(Boolean)
+         * @see BedrockChatModel.Builder#sendThinking(Boolean)
+         */
+        public Builder enableAdaptiveReasoning(String effort) {
+            if (additionalModelRequestFields == null) {
+                additionalModelRequestFields = new HashMap<>();
+            }
+            Map<?, ?> reasoningConfig = Map.ofEntries(Map.entry("type", "adaptive"));
+            additionalModelRequestFields.put("reasoning_config", reasoningConfig);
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):

dliubarskyi:
@Aditya-Pimpalkar thank you!

PR #5361, opened 2026-06-04
Introduce `SystemMessageProvider`

geoand:
This is essentially an evolution of what
`AiServices#systemMessageProvider` provides
with additional contextual information.

Originally mentioned in https://github.com/quarkiverse/quarkus-langchain4j/issues/2466


langchain4j/src/main/java/dev/langchain4j/service/SystemMessageProvider.java
```diff
@@ -0,0 +1,50 @@
...
+     * @param invocationContext the invocation context
+     * @return a system message or a template containing unresolved variables (e.g. "{{name}}"),
+     *         which will be resolved using the values of method parameters annotated with @{@link V}
+     */
+    String get(InvocationContext invocationContext);
+
+    /**
+     * Context available to the {@link SystemMessageProvider} at invocation time.
+     */
+    interface InvocationContext {
```
dliubarskyi: We already have `dev.langchain4j.invocation.InvocationContext` type, I was expecting it to be used for this use case. Why is new type (with the same name) is required?

langchain4j/src/main/java/dev/langchain4j/service/AiServices.java
```diff
@@ -278,6 +278,25 @@ public AiServices<T> systemMessageProvider(Function<Object, String> systemMessag
...
+     * {@link SystemMessageProvider.InvocationContext} that also exposes the {@link dev.langchain4j.model.chat.ChatModel}
+     * configured on this AI service, allowing the system message to vary based on model capabilities.
+     * <p>
+     * When both {@code @SystemMessage} and this provider are configured,
+     * {@code @SystemMessage} takes precedence.
+     *
+     * @param systemMessageProvider the provider to use
+     * @return builder
+     */
+    public AiServices<T> contextualSystemMessageProvider(SystemMessageProvider systemMessageProvider) {
```
```suggestion
    public AiServices<T> systemMessageProviderWithContext(SystemMessageProvider systemMessageProvider) {
```

WDYT? It will be easier to find with the IDE autocompletion

langchain4j/src/main/java/dev/langchain4j/service/SystemMessageProvider.java
```diff
@@ -0,0 +1,50 @@
...
+ * <p>
+ * Unlike {@link AiServices#systemMessageProvider(java.util.function.Function)},
+ * this provider receives a {@link InvocationContext} that exposes the {@link ChatModel}
+ * configured on the AI service, allowing the system message to vary based on model capabilities.
+ * <p>
+ * When both {@code @SystemMessage} and this provider are configured,
+ * {@code @SystemMessage} takes precedence.
+ */
+@FunctionalInterface
+public interface SystemMessageProvider {
```
dliubarskyi: I am not hard against a new type, but it could have been `Function<InvocationContext, String>`. Do we really need a new type?

dliubarskyi (commented):
@geoand thank you!

langchain4j/src/main/java/dev/langchain4j/service/SystemMessageProvider.java
geoand: `InvocationContext` doesn't have to the default model, right? Are you proposing we add it?

geoand:
Thanks for the review! Before moving to address your comments, I added https://github.com/langchain4j/langchain4j/pull/5361#discussion_r3361684838

langchain4j/src/main/java/dev/langchain4j/service/SystemMessageProvider.java
dliubarskyi: Yes, I propose to add `ChatModel chatModel()` to `InvocationContext`.
This is what I meant in https://github.com/quarkiverse/quarkus-langchain4j/issues/2466#issuecomment-4612449056

Not sure if "default" should be in the name, since we currently have no way to override the model....

geoand: Got it, thanks

geoand: PR updated

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java
```diff
@@ -57,6 +58,16 @@ default UserMessage userMessage() {
      */
     Object chatMemoryId();
 
+    /**
+     * Returns the {@link ChatModel} configured on the AI service,
+     * or {@code null} if only a {@link StreamingChatModel} was configured.
+     *
+     * @since 1.14.0
```
dliubarskyi:
```suggestion
     * @since 1.16.0
```

geoand: Oops, fixed

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -140,12 +140,13 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                                         InvocationParameters.class, args, method.getParameters())
                                 .orElseGet(InvocationParameters::new);
 
-                        InvocationContext invocationContext = InvocationContext.builder()
+                        InvocationContext invocationContext = dev.langchain4j.invocation.InvocationContext.builder()
```
```suggestion
                        InvocationContext invocationContext = InvocationContext.builder()
```

langchain4j-core/src/main/java/dev/langchain4j/invocation/InvocationContext.java
```diff
@@ -57,6 +58,16 @@ default UserMessage userMessage() {
      */
     Object chatMemoryId();
 
+    /**
+     * Returns the {@link ChatModel} configured on the AI service,
+     * or {@code null} if only a {@link StreamingChatModel} was configured.
```
dliubarskyi: I totally forgot about `StreamingChatModel`.... We need to add it as well. But this way user will have to manually check which type of model is actually used by the AI Service. Or should we just add `ChatRequestParameters defaultRequestParameters()` and `ModelProvider provider()` to the `InvocationContext` instead of models? That would be easier for the user I guess

geoand: Yeah, I think the latter would be the easiest

geoand: I made the change, but I am pretty sure the way I select the model is wrong. What is the proper way to determine which model is going to be used?

langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
```diff
@@ -162,6 +165,20 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                         }
                     }
 
+                    private static ChatRequestParameters determineChatRequestParameters(AiServiceContext context) {
+                        if (context.chatModel != null) {
+                            return context.chatModel.defaultRequestParameters();
+                        }
+                        return context.streamingChatModel != null ? context.streamingChatModel.defaultRequestParameters() : null;
+                    }
```
dliubarskyi, line 172: It is an acceptable proxy for "is it streaming or not?" since 99% of users would configure only the model they intend to use. The proper one is `boolean streaming = returnType == TokenStream.class || canAdaptTokenStreamTo(returnType);`, but it might be more complicated to check that at this point

geoand: Up to you :). Should I leave it as is?

dliubarskyi: Let's leave it as is for now, thanks!

geoand: Thank you!

dliubarskyi (approved):
@geoand thank you!

geoand:
Thanks for your help!

PR #5364, opened 2026-06-04
Improve parallel agents execution

### LINKED ISSUE #5360: perf: parallel execution waits on submission order rather than completion order
## Summary

In `PlannerBasedInvocationHandler.parallelExecution()`, parallel sub-agent results are collected using a sequential `future.get()` loop:

```java
for (CompletableFuture<Object> future : futures) {
    results.add(future.get());
}
```

This blocks on futures in submission order. If the first future submitted is the slowest (e.g., a slow sub-agent submitted before fast ones), the loop waits for it before collecting any of the already-completed fast ones — adding latency equal to the difference between the slowest and fastest task, multiplied by the number of tasks ahead of it.

## Fix

Replace the sequential loop with `CompletableFuture.allOf()` then `join()`:

```java
CompletableFuture.allOf(futures.toArray(new CompletableFuture[0])).get();
for (CompletableFuture<Object> future : futures) {
    results.add(future.join()); // safe after allOf completes
}
```

This collects results only after all futures complete, in a single wait — wall-clock time equals the slowest single task rather than submission-order blocking.

## Impact

Any `@ParallelAgent` or similar construct that uses `parallelExecution()` with heterogeneous sub-agent latencies pays unnecessary latency overhead under the current implementation.

## Platform independence

This is a pure performance fix with no dependency on any DI framework. It is equally applicable to Spring, Micronaut, Guice, or plain Java users of `langchain4j-agentic`.

mariofusco:
## Issue

Closes #5360

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/internal/PlannerBasedInvocationHandler.java
@@ -21,7 +21,6 @@ import java.util.Map;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
-import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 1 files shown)


dliubarskyi (approved):
@mariofusco thank you"

Issue #5366, opened 2026-06-04
[FEATURE] support non Result return types for AiService when tool execution is IMMEDIATE


patriot1burke:
**Is your feature request related to a problem? Please describe.**
If using ReturnBehavior.IMMEDIATE, your AiService method has to return Result even if
all tools called return `void` or all return `null` or there's one tool method that returns same type as AiService return type.

**Describe the solution you'd like**
I'd like to support non-Result return types for AiService methods when tool execution has IMMEDIATE results based on these rules:

* If ai service method return type is void, just return null
* If any tool executions are TO_LLM throw exception (IMMEDIATE_IF_LAST case)
* If all tool executions have a null result and return type is nullable (not a primitive), then return null
* If there is one non-null tool execution result, and the result can be resolved to the return type, then return that object

I have a PR for this open in Quarkus too:

https://github.com/quarkiverse/quarkus-langchain4j/pull/2529

PR #5372, opened 2026-06-05
Document non-Result IMMEDIATE behavior

patriot1burke:
The documents issue #5366


dliubarskyi (approved):
@patriot1burke thank you!

PR #5356, opened 2026-06-04
Remove `style` field and related logic from `OpenAiImageModel`.

edeandrea:
OpenAI no longer supports this, and actually breaks if you send it no matter which image model you use.

See https://developers.openai.com/api/reference/resources/images

```
- method: POST
- url: https://api.openai.com/v1/images/generations
- headers: [Accept: application/json], [Authorization: Be...oA], [Content-Type: application/json], [User-Agent: Quarkus REST Client], [content-length: 2285]
- body: {                                                                                                                                                                                                                                                                                                                                  
  "model" : "dall-e-3",                                                                                                                                                                                                                                                                                                                 
  "prompt" : "Under the shadowy skyline of Gotham City, the air crackled with tension as two formidable figures faced off in a deserted alley. The dim glow of streetlights illuminated the scene, casting long shadows that danced across the brick walls. One figure, clad in a dark cloak, wielded a crimson blade that hummed ominously, while the other, with a determined gaze, brandished a brilliant blue lightsaber that shimmered with hope. The atmosphere was thick with anticipation, as the echoes of distant sirens and the faint rustle of the night enveloped them.\n\nWith a swift motion, the dark-cloaked figure lunged forward, the red blade slicing through the air with deadly precision. But the determined warrior was ready, channeling the energy of the Force to anticipate the attack. In a fluid motion, he parried the strike, sending sparks flying as the two weapons clashed. The sound reverberated through the alley, a testament to their opposing ideals—one fueled by fear and anger, the other by courage and resolve. Each swing of the lightsaber was not just a physical battle, but a clash of wills, echoing the age-old struggle between light and darkness.\n\nAs the fight intensified, the warrior tapped deeper into the Force, his movements becoming a graceful dance of agility and strength. He dodged and weaved, countering with precise strikes that showcased his mastery. With each successful maneuver, he gained confidence, and the dark figure began to falter, his attacks growing more desperate. The alley, once a silent witness to corruption, now bore witness to a transformative moment as the tides of battle shifted, illuminating the path of righteousness.\n\nIn a final, breathtaking clash, the warrior unleashed a surge of energy, channeling the very essence of hope. With a decisive strike, he disarmed his opponent, sending the crimson blade skittering across the pavement. The dark figure, now vulnerable, fell to his knees, the weight of his choices heavy upon him. The victorious champion stood tall, a beacon of light amidst the shadows of Gotham, reminding all that even in the darkest of places, hope could prevail.",                                                                                                                                                                               
[...]

PR #5373, opened 2026-06-05
Migrate OpenAI image models from DALL-E to new GPT image API

### LINKED ISSUE #5371: [FEATURE] Move OpenAiImageModel away from dall-e to the new images api
OpenAI recently deprecated & remove the dall-e models and APIs and replaced with the [images api](https://developers.openai.com/api/reference/resources/images).

The existing `ImageModel` stuff no longer works due to this. We should make this change as well.

edeandrea:
## Summary

- Adds GPT image model names (`gpt-image-1`, `gpt-image-1-mini`, `gpt-image-1.5`, `gpt-image-2`, `chatgpt-image-latest`) and deprecates `DALL_E_2`/`DALL_E_3`
- Removes `style` and `response_format` parameters that are rejected by the API (see #5356 for context on the 400 error)
- Adds new GPT image model parameters: `background`, `output_format`, `output_compression`, `moderation`
- Captures full API response fields including token `usage` info
- Sets `mimeType` on `Image` from the response's `output_format`
- Updates both `langchain4j-open-ai` (legacy) and `langchain4j-open-ai-official` modules

## Test plan

- [x] Both modules compile cleanly (`mvn test-compile`)
- [ ] Run `OpenAiImageModelIT` manually with `OPENAI_API_KEY` set
- [ ] Run `OpenAiOfficialImageModelIT` manually with `OPENAI_API_KEY` set

Closes #5371

### DIFF
--- a/langchain4j-open-ai-official/revapi.json
+++ b/langchain4j-open-ai-official/revapi.json
@@ -8,6 +8,26 @@
           "ignore": true,
           "code": "java.class.externalClassExposedInAPI",
           "justification": "OpenAI SDK types (e.g., ReasoningEffort, Reasoning.Summary, OpenAIClient) are intentionally exposed on the public API of this module, since wrapping the official OpenAI Java SDK is its purpose."
+        },
+        {
+          "code": "java.method.removed",
+          "old": "method dev.langchain4j.model.openaiofficial.OpenAiOfficialImageModel.Builder dev.langchain4j.model.openaiofficial.OpenAiOfficialImageModel.Builder::style(java.lang.String)",
+          "justification": "The 'style' parameter is not supported by the new OpenAI GPT image models"
+        },
+        {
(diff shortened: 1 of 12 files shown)


geoand:
So IIUC, without this, the image support doesn't work at all right? If true, we should get this in now and not wait for the upstream fix

dliubarskyi (approved):
@edeandrea thanks a lot!

I am going to release this in 1.16.1 now

geoand:
Thanks @dliubarskyi!

dliubarskyi:
Released: https://github.com/langchain4j/langchain4j/releases/tag/1.16.1

edeandrea:
> So IIUC, without this, the image support doesn't work at all right? If true, we should get this in now and not wait for the upstream fix

Yes unless a user updates the jackson mappings themselves image generation returns 400 errors

PR #5381, opened 2026-06-08
Surface MCP icons in MCP client models

### LINKED ISSUE #5258: langchain4j-mcp: surface MCP 2025-11-25 icons[] on Tool / Resource / Prompt
## Summary

The MCP 2025-11-25 specification adds an `icons[]` array (peer of `name` / `description` / `_meta`) on tools, resources, and prompts. Each icon is `{src, mimeType?, sizes?}`. `langchain4j-mcp` v1.15.0 does not parse this field — it lands in JSON-RPC responses but is dropped before reaching `ToolSpecification` / `McpResource` / `McpPrompt`.

Reference: https://modelcontextprotocol.io/specification/2025-11-25/server/tools#icons (and the parallel sections under `/server/resources` and `/server/prompts`).

## What's missing

- `ToolSpecificationHelper.toolSpecificationListFromMcpResponse(...)` parses `annotations` (via `processMcpToolAnnotations`) and `_meta` but not `icons[]`.
- `McpResource` exposes `uri / name / description / mimeType` — no `icons`.
- `McpPrompt` exposes `name / description / arguments` — no `icons`.

## Proposed shape

Either:

**A. Typed accessors** (preferred):

```java
public record Icon(String src, String mimeType, List<String> sizes) {}

ToolSpecification.icons()  // List<Icon>
McpResource.icons()
McpPrompt.icons()
```

**B. Metadata-map pass-through** (cheaper):

Add `McpToolMetadataKeys.ICONS` and stash the raw `JsonNode` array under that key in `ToolSpecification.metadata()`. Mirror on Resource/Prompt by adding a `metadata()` map (currently absent).

## Workaround we're using

We decorate `McpTransport` and intercept the `tools/list` / `resources/list` / `prompts/list` JSON-RPC envelopes via `executeOperationWithResponse(McpCallContext)`, then parse icons[] ourselves. This works but couples our code to internal types (`McpClientMessage.method`, etc.) and we'd love to delete it once icons are first-class.

## Why this matters

Icons are part of the standard tool/resource/prompt UX surface in MCP-aware clients. Without library support, every consumer either drops icons or rolls their own probe — which is exactly what we did.

Happy to send a PR if there's interest. Just wanted to file the issue first to check appetite + preferred shape (A vs B).

## Related

- #4282 — reports the same MCP 2025-11-25 spec gap from a different angle (deserialization failure on `McpResourceTemplate.icons`, asks for customizable `ObjectMapper`). Surfacing `icons[]` as typed accessors here would also close that failure mode for resource templates if extended to `McpResourceTemplate`.

DragonFSKY:
## Issue
Closes #5258

## Change
Adds typed MCP icon support for MCP list results:
- stores top-level `icons` from MCP tools in `ToolSpecification.metadata()` under `McpToolMetadataKeys.ICONS` as `List<McpIcon>`;
- adds `McpIcon` and `McpIconTheme` for the MCP icon schema;
- exposes `_meta` via `metadata()` and top-level `icons` via `icons()` on `McpResource`, `McpResourceTemplate`, and `McpPrompt`;
- keeps existing public constructors for resource/prompt/template models for source compatibility;
- removes the metadata helper;
- adds focused parsing tests for present and absent metadata/icons.

## General checklist

## Verification
- `mvn -pl langchain4j-mcp test -Dmaven.javadoc.skip=true -DskipITs -Ddokka.skip=true`
- `mvn -pl langchain4j-mcp -Dtest=McpMetadataParsingTest,ToolSpecificationHelperTest test -Dmaven.javadoc.skip=true -DskipITs -Ddokka.skip=true`
- `mvn -pl langchain4j-mcp spotless:check -Dmaven.javadoc.skip=true -DskipTests -DskipITs -Ddokka.skip=true`
- `git diff --check`

### DIFF
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpPrompt.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpPrompt.java
@@ -1,9 +1,12 @@
 package dev.langchain4j.mcp.client;
 
+import static dev.langchain4j.internal.Utils.copy;
+import static dev.langchain4j.internal.Utils.mutableCopy;
+
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
-
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 
 /**
@@ -16,16 +19,28 @@ public class McpPrompt {
     private final String name;
     private final String description;
     private final List<McpPromptArgument> arguments;
+    private final Map<String, Object> metadata;
+    private final List<McpIcon> icons;
+
+    public McpPrompt(
+            @JsonProperty("name") String name,
+            @JsonProperty("description") String description,
+            @JsonProperty("arguments") List<McpPromptArgument> arguments) {
+        this(name, description, arguments, null, null);
+    }
 
     @JsonCreator
     public McpPrompt(
             @JsonProperty("name") String name,
             @JsonProperty("description") String description,
-            @JsonProperty("arguments") List<McpPromptArgument> arguments
-    ) {
+            @JsonProperty("arguments") List<McpPromptArgument> arguments,
+            @JsonProperty("_meta") Map<String, Object> metadata,
+            @JsonProperty("icons") List<McpIcon> icons) {
         this.name = name;
         this.description = description;
         this.arguments = arguments;
+        this.metadata = copy(mutableCopy(metadata));
+        this.icons = icons == null ? List.of() : List.copyOf(icons);
     }
 
     public String name() {
@@ -40,26 +55,38 @@ public class McpPrompt {
         return arguments;
     }
 
+    public Map<String, Object> metadata() {
+        return metadata;
+    }
+
+    public List<McpIcon> icons() {
+        return icons;
+    }
+
     @Override
     public boolean equals(Object obj) {
         if (obj == this) return true;
         if (obj == null || obj.getClass() != this.getClass()) return false;
         var that = (McpPrompt) obj;
-        return Objects.equals(this.name, that.name) &&
-                Objects.equals(this.description, that.description) &&
-                Objects.equals(this.arguments, that.arguments);
+        return Objects.equals(this.name, that.name)
+                && Objects.equals(this.description, that.description)
+                && Objects.equals(this.arguments, that.arguments)
+                && Objects.equals(this.metadata, that.metadata)
+                && Objects.equals(this.icons, that.icons);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(name, description, arguments);
+        return Objects.hash(name, description, arguments, metadata, icons);
     }
 
     @Override
     public String toString() {
-        return "McpPrompt[" +
-                "name=" + name + ", " +
-                "description=" + description + ", " +
-                "arguments=" + arguments + ']';
+        return "McpPrompt[" + "name="
+                + name + ", " + "description="
+                + description + ", " + "arguments="
+                + arguments + ", " + "metadata="
+                + metadata + ", " + "icons="
+                + icons + ']';
     }
 }
--- a/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpPrompt.java
+++ b/langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpPrompt.java
@@ -1,7 +1,6 @@
 package dev.langchain4j.mcp.client;
 
 import static dev.langchain4j.internal.Utils.copy;
-import static dev.langchain4j.internal.Utils.mutableCopy;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
@@ -39,7 +38,7 @@ public class McpPrompt {
         this.name = name;
         this.description = description;
         this.arguments = arguments;
-        this.metadata = copy(mutableCopy(metadata));
+        this.metadata = copy(metadata);
         this.icons = icons == null ? List.of() : List.copyOf(icons);
     }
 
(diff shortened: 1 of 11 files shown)


langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/ToolSpecificationHelper.java
```diff
@@ -54,6 +54,9 @@ static List<ToolSpecification> toolSpecificationListFromMcpResponse(ArrayNode ar
             if (tool.has("outputSchema")) {
                 builder.addMetadata(OUTPUT_SCHEMA, OBJECT_MAPPER.convertValue(tool.get("outputSchema"), Object.class));
             }
+            if (tool.has("icons")) {
+                builder.addMetadata(ICONS, OBJECT_MAPPER.convertValue(tool.get("icons"), Object.class));
```
jmartisk: Can't we introduce a class for this, why `Object`? Or rather a `record`. The schema is well-defined: https://github.com/modelcontextprotocol/modelcontextprotocol/blob/2025-11-25/schema/2025-11-25/schema.json#L1303

So it would be a record containing `String mimeType`, `List<String> sizes`, `String src`, `IconTheme theme`, where `IconTheme` is an enum with values `dark` and `light`

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpResourceTemplate.java
```diff
@@ -14,6 +15,7 @@ public class McpResourceTemplate {
     private final String name;
     private final String description;
     private final String mimeType;
+    private Map<String, Object> metadata;
```
jmartisk: I would say we can have a separate field with icons here, no? A `private final List<Icon> icons`.
The same in `McpResource` and `McpPrompt`.
The only case where you need to put it into the metadata is in the `ToolSpecification` because that is not a MCP-specific class. But `McpResource`, `McpResourceTemplate` and `McpTools` are MCP-specific so we can have these fields here.
When you do that, you will also be able to make the `metadata` field `final`, as it should be.

jmartisk: ... and also get rid of the `McpMetadataHelper` class, which is a bit confusing IMO

DragonFSKY:
Thanks for the review.

I updated the PR to model icons as typed `McpIcon` values with `McpIconTheme`. `McpResource`, `McpResourceTemplate`, and `McpPrompt` now expose root-level `icons` directly, while `_meta` stays in `metadata()`.

For tools, icons still go through `ToolSpecification.metadata()` since `ToolSpecification` is not MCP-specific. I also removed the metadata helper.

Please let me know if anything else needs adjustment, and I’ll follow up promptly.

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpPrompt.java
```diff
@@ -16,16 +19,28 @@ public class McpPrompt {
...
             @JsonProperty("description") String description,
-            @JsonProperty("arguments") List<McpPromptArgument> arguments
-    ) {
+            @JsonProperty("arguments") List<McpPromptArgument> arguments,
+            @JsonProperty("_meta") Map<String, Object> metadata,
+            @JsonProperty("icons") List<McpIcon> icons) {
         this.name = name;
         this.description = description;
         this.arguments = arguments;
+        this.metadata = copy(mutableCopy(metadata));
```
jmartisk: I guess we just need an immutable copy, so the `mutableCopy` call can go away :)

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpResource.java
```diff
@@ -13,17 +18,31 @@ public class McpResource {
...
+            @JsonProperty("name") String name,
+            @JsonProperty("description") String description,
+            @JsonProperty("mimeType") String mimeType,
+            @JsonProperty("_meta") Map<String, Object> metadata,
+            @JsonProperty("icons") List<McpIcon> icons) {
         this.uri = Utils.warnIfNullOrBlank(uri, "uri", McpResource.class);
         this.name = Utils.warnIfNullOrBlank(name, "name", McpResource.class);
         this.description = description;
         this.mimeType = mimeType;
+        this.metadata = copy(mutableCopy(metadata));
```
jmartisk: Same here, no need for `mutableCopy`

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/McpResourceTemplate.java
```diff
@@ -14,17 +19,31 @@ public class McpResourceTemplate {
...
+            @JsonProperty("name") String name,
+            @JsonProperty("description") String description,
+            @JsonProperty("mimeType") String mimeType,
+            @JsonProperty("_meta") Map<String, Object> metadata,
+            @JsonProperty("icons") List<McpIcon> icons) {
         this.uriTemplate = Utils.warnIfNullOrBlank(uriTemplate, "uriTemplate", McpResourceTemplate.class);
         this.name = Utils.warnIfNullOrBlank(name, "name", McpResourceTemplate.class);
         this.description = description;
         this.mimeType = mimeType;
+        this.metadata = copy(mutableCopy(metadata));
```
jmartisk: again, unnecessary call to `mutableCopy`

jmartisk (commented):
Almost perfect now! Just one little note (repeating itself 3 times)

Also, could you also add a short note about this to the documentation? The section about tool metadata is: https://docs.langchain4j.dev/tutorials/mcp/#mcp-tool-metadata
Perhaps just add a single-sentence paragraph like this to the end of the section:

> If the tool has icons, they are exposed under the key `McpToolMetadataKeys.ICONS`

jmartisk (approved):
Nice! Thanks

DragonFSKY:
Thanks for the review!

I addressed the feedback by removing the unnecessary `mutableCopy(metadata)` calls and adding the docs note. Please let me know if anything else comes up, and I’ll be happy to follow up.

PR #5386, opened 2026-06-08
fix: remove double python prefix in Gemini code execution output

acefun29:
Change

  Fix double "python" prefix in Gemini code execution output.

  Before:
  Code executed:
  ```pythonpythonprint("hello")```

  After:
  Code executed:
  ```python
  print("hello")

  **Root cause:** `PartsAndContentsMapper.java` hardcodes `"```python"` as a literal string, then appends
  `executableCode.programmingLanguage().toString()` which also returns `"python"` (since `GeminiLanguage.PYTHON.toString()` returns
  `name().toLowerCase()`). The compact constructor of `GeminiExecutableCode` guarantees `programmingLanguage` is never null (defaults
  to `PYTHON`), making the null check dead code.

  **Fix:** Use only the programming language from the API response and add proper newlines for correct markdown rendering. Removed
  dead null check and associated TODO comments.

  ## General checklist
  - [ ] I have added unit and/or integration tests for my change — *existing integration tests pass; formatting was not previously
  tested*
  - [ ] I have added an example in the examples repo
  - [ ] I have added/updated Spring Boot starter(s)


dliubarskyi:
THis was fixed in #5520

PR #5387, opened 2026-06-08
fix(elasticsearch): apply hybrid filters to full-text retriever

### LINKED ISSUE #5383: [BUG] ElasticsearchConfigurationHybrid does not apply Filter to the full-text search component
## Description

  When using ElasticsearchConfigurationHybrid with a Filter, the filter is only applied to the KNN (vector) retriever, but not to the full-text retriever. After RRF (Reciprocal Rank Fusion) merges the results from both
  retrievers, unfiltered documents from the full-text path appear in the final results, effectively bypassing the filter.

 ## Environment

  - LangChain4j version: 1.15.1-beta25
  - Module: langchain4j-elasticsearch

  ## Reproduction
  ```java
  Filter filter = new IsEqualTo("year", 2023);

  ElasticsearchContentRetriever contentRetrieverHybrid = ElasticsearchContentRetriever.builder()
      .client(client)
      .configuration(ElasticsearchConfigurationHybrid.builder().build())
      .indexName(indexName)
      .maxResults(3)
      .filter(filter)
      .build();

  List<Content> results = contentRetrieverHybrid.retrieve(Query.from("movies about space battles"));
  ```
  Actual results:
  ```
  2023:tt13539646:The Wandering Earth 2
  2014:tt0816692:Interstellar     <-- should be filtered out
  2023:tt6791350:Guardians of the Galaxy Vol. 3
  ```
  Expected results: Only documents with `year == 2023` should appear.

  As a control, running the same query with ElasticsearchConfigurationKnn + the same filter correctly returns only 2023 documents, confirming the filter itself works.

  ## Root Cause

  In ElasticsearchConfigurationHybrid.java, the hybridSearch() method applies the filter to the KNN retriever but not to the full-text MatchQuery:
```java
  // KNN part — filter IS applied ✅
  KnnRetriever.Builder krb = new KnnRetriever.Builder()
          .field(VECTOR_FIELD)
          .queryVector(embeddingSearchRequest.queryEmbedding().vectorAsList());

  if (embeddingSearchRequest.filter() != null) {
      krb.filter(ElasticsearchMetadataFilterMapper.map(embeddingSearchRequest.filter()));
  }

  // Full-text part — filter is NOT applied ❌
  MatchQuery matchQuery =
          new MatchQuery.Builder().field(TEXT_FIELD).query(textQuery).build();

  // RRF combines both — unfiltered full-text results leak through
  return client.search(s -> s
          .retriever(r -> r.rrf(rf -> rf.retrievers(List.of(
                  RRFRetrieverEntry.of(rre -> rre.retriever(rt -> rt.standard(st -> st.query(matchQuery)))),
                  RRFRetrieverEntry.of(rre -> rre.retriever(rt -> rt.knn(knn))))
          ))),
          Document.class);
  ```
 ## Suggested Fix

  When a filter is present, wrap the full-text MatchQuery in a BoolQuery that includes the filter, so both retrievers are constrained equally:
  ```java
  Query fullTextQuery;
  if (embeddingSearchRequest.filter() != null) {
      Query filterQuery = ElasticsearchMetadataFilterMapper.map(embeddingSearchRequest.filter());
      fullTextQuery = Query.of(q -> q
          .bool(b -> b
              .must(matchQuery._toQuery())
              .filter(filterQuery)
          )
      );
  } else {
      fullTextQuery = matchQuery._toQuery();
  }

  // Then use fullTextQuery in the standard retriever instead of matchQuery
  ```
  This ensures the filter is applied consistently to both the full-text and KNN components before RRF fusion.

SJvaca30:
## Summary

- apply the metadata filter to the Elasticsearch hybrid full-text retriever
- keep the same filter on the KNN retriever so both RRF inputs are constrained
- add a unit test that captures the generated SearchRequest and verifies both retrievers carry the filter

Fixes #5383

## Test Plan

- ./mvnw -pl langchain4j-elasticsearch -Dtest=ElasticsearchConfigurationHybridTest test
- ./mvnw -pl langchain4j-elasticsearch verify
- ./mvnw -pl langchain4j-elasticsearch spotless:check
- git diff --check

### DIFF
--- a/langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchConfigurationHybrid.java
+++ b/langchain4j-elasticsearch/src/main/java/dev/langchain4j/store/embedding/elasticsearch/ElasticsearchConfigurationHybrid.java
@@ -4,7 +4,8 @@ import co.elastic.clients.elasticsearch.ElasticsearchClient;
 import co.elastic.clients.elasticsearch._types.ElasticsearchException;
 import co.elastic.clients.elasticsearch._types.KnnRetriever;
 import co.elastic.clients.elasticsearch._types.RRFRetrieverEntry;
-import co.elastic.clients.elasticsearch._types.query_dsl.MatchQuery;
+import co.elastic.clients.elasticsearch._types.StandardRetriever;
+import co.elastic.clients.elasticsearch._types.query_dsl.Query;
 import co.elastic.clients.elasticsearch.core.SearchResponse;
 import co.elastic.clients.elasticsearch.core.search.SourceConfig;
 import dev.langchain4j.store.embedding.EmbeddingSearchRequest;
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@SJvaca30 thank you!

PR #5405, opened 2026-06-10
Fix filtering for non-TextSegment entries in InMemoryEmbeddingStore

### LINKED ISSUE #5407: InMemoryEmbeddingStore should handle filters consistently for non-TextSegment entries
## Description
`InMemoryEmbeddingStore` handles metadata filters inconsistently when the store contains entries whose embedded object is not a `TextSegment`.

When a filter is present, non-`TextSegment` entries cannot be evaluated against metadata. They should not match the filter, but current behavior can be inconsistent between `search()` and `removeAll(Filter)`.

## Expected behavior
When a metadata filter is present, non-`TextSegment` entries should not match it. `search()` and `removeAll(Filter)` should apply the same behavior.

## Actual behavior
Non-`TextSegment` entries can either bypass the filter during `search()` or cause `removeAll(Filter)` to throw `UnsupportedOperationException`.

## Additional context
Regression tests and a fix are proposed in #5405.

Q1Xuan:
## Issue
Fixes #5407

## Change
Fixes an inconsistency in `InMemoryEmbeddingStore` when metadata filters are used with entries whose embedded object is not a `TextSegment`.

Previously, non-`TextSegment` entries could either bypass the filter during `search()` or cause `removeAll(Filter)` to throw `UnsupportedOperationException`.

This PR:
- Extracts a shared `matchesFilter()` helper
- Returns `false` for non-`TextSegment` entries when a filter is present
- Applies the same filtering behavior to both `search()` and `removeAll(Filter)`
- Adds `InMemoryEmbeddingStoreFilterRegressionTest` covering mixed stores

## ⚠️ Potential issues for users (behavioral change)

This PR changes the behavior of **`search(...)`** for non-`TextSegment` entries when a filter is applied.

### What changed

`InMemoryEmbeddingStore<Embedded>` is generic and can hold entries whose embedded object is `null` (e.g. added via `add(id, embedding)` / `add(embedding)`) or some non-`TextSegment` type. Filters operate on `TextSegment` metadata, so they have nothing to test against for these entries.

| Operation | Entry type | Before | After |
|---|---|---|---|
| `search` (with filter) | `TextSegment` | filtered on metadata | filtered on metadata (unchanged) |
| `search` (with filter) | `null` / non-`TextSegment` | **included** in results | **excluded** from results |
| `search` (no filter) | any | included | included (unchanged) |
| `removeAll(filter)` | non-`TextSegment` (non-null) | **threw** `UnsupportedOperationException` | not removed (no throw) |
| `removeAll(filter)` | `null` | not removed | not removed (unchanged) |

### Who is affected

- The common case — `InMemoryEmbeddingStore<TextSegment>` where every entry carries metadata — is **not affected**.
- Affected only if you run a **filtered search** over a store that contains `null`-embedded entries or non-`TextSegment` embedded objects. Such entries that previously slipped through a filtered search will now be omitted, so a filtered query may return fewer (or zero) results than before.

### Why this is the intended behavior

The new semantics are consistent across both methods: an entry with no filterable metadata is treated as **not matching the filter** — so it is excluded from `search` results and left in place by `removeAll`. The previous `search` behavior (returning metadata-less entries from a metadata-filtered query) was effectively a bug, and `removeAll` previously threw for the same entries. This change unifies them.

### Migration

If you relied on filtered `search` returning non-`TextSegment`/`null`-embedded entries, either:
- run the search **without a filter** and apply your own post-filtering, or
- ensure the entries you want returned are stored as `TextSegment`s with appropriate metadata.

The behavior is now documented in the javadoc of `search`, `removeAll(Filter)`, and the `matchesFilter` helper.

## General checklist
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j) modules, and they are all green

N/A

N/A

N/A

## Verification
GitHub Actions passed:
- `compile_and_unit_test (17)`
- `compile_and_unit_test (21)`
- `compile_and_unit_test (25)`
- `integration_test (21)`
- `spotless`
- `compliance`
- `check-split-packages`

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
+++ b/langchain4j/src/main/java/dev/langchain4j/store/embedding/inmemory/InMemoryEmbeddingStore.java
@@ -12,6 +12,7 @@ import static java.util.Comparator.comparingDouble;
 import static java.util.stream.Collectors.toList;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+import dev.langchain4j.data.document.Metadata;
 import dev.langchain4j.data.embedding.Embedding;
 import dev.langchain4j.data.segment.TextSegment;
 import dev.langchain4j.spi.store.embedding.inmemory.InMemoryEmbeddingStoreJsonCodecFactory;
@@ -127,6 +128,14 @@ public class InMemoryEmbeddingStore<Embedded> implements EmbeddingStore<Embedded
         entries.removeIf(entry -> idSet.contains(entry.id));
     }
 
+    /**
+     * Removes all entries whose embedded object matches the given {@link Filter}.
+     * <p>
+     * Filtering is applied only to entries whose embedded object is a {@link TextSegment},
+     * by testing the filter against the segment's {@link Metadata}.
+     * Entries whose embedded object is {@code null} or is not a {@link TextSegment} are considered
+     * non-matching and are therefore never removed.
+     */
     @Override
     public void removeAll(Filter filter) {
         ensureNotNull(filter, "filter");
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@Q1Xuan thank you!

PR #5413, opened 2026-06-11
Allow to optionally propagate the exception raised by a tool execution

mariofusco:
## Issue

Closes #

## Change
This pull request allows to optionally propagate the exception raised by a tool execution. The main goal is to remove tons of code duplication in the Quarkus extension that in some cases (like for tools guardrails) relies on this behavior, but in general I believe this is a nice addition also for LangChain4j itself.

## General checklist

### DIFF
--- a/docs/docs/tutorials/tools.md
+++ b/docs/docs/tutorials/tools.md
@@ -1588,7 +1588,7 @@ Assistant assistant = AiServices.builder(Assistant.class)
         .chatModel(chatModel)
         .tools(tools)
         .toolArgumentsErrorHandler((error, errorContext) -> {
-            if (errorContext.originalException() instanceof MyCriticalException) {
+            if (errorContext.rawError() instanceof MyCriticalException) {
                 return ToolErrorHandlerResult.propagateException();
             }
             return ToolErrorHandlerResult.text(error.getMessage());
@@ -1596,8 +1596,8 @@ Assistant assistant = AiServices.builder(Assistant.class)
         .build();
 ```
 
-The `error` argument is the unwrapped cause of the exception. Use `errorContext.originalException()`
-to access the original exception as thrown, before cause-unwrapping — useful when the wrapper type
+The `error` argument is the unwrapped cause. Use `errorContext.rawError()`
+to access the raw error as thrown, before cause-unwrapping — useful when the wrapper type
 (not the cause) determines how the error should be handled.
 
 #### Handling Tool Execution Errors
@@ -1634,8 +1634,8 @@ Assistant assistant = AiServices.builder(Assistant.class)
 
 As with the `ToolArgumentsErrorHandler`, there are three ways to handle errors in `ToolExecutionErrorHandler`:
 return a text message, throw an exception, or return `ToolErrorHandlerResult.propagateException()`
-to re-throw the original exception. You can use `errorContext.originalException()` to inspect
-the original exception before cause-unwrapping.
+to re-throw the raw error. You can use `errorContext.rawError()` to inspect
+the raw error before cause-unwrapping.
 
 ## Model Context Protocol (MCP)
 
--- a/docs/docs/tutorials/tools.md
+++ b/docs/docs/tutorials/tools.md
@@ -1596,9 +1596,10 @@ Assistant assistant = AiServices.builder(Assistant.class)
         .build();
 ```
 
-The `error` argument is the unwrapped cause. Use `errorContext.rawError()`
-to access the raw error as thrown, before cause-unwrapping — useful when the wrapper type
-(not the cause) determines how the error should be handled.
+When a tool throws an exception that wraps another (e.g. `ToolGuardrailException` wrapping `SecurityException`),
+LangChain4j extracts the inner cause via `getCause()` and passes it as the `error` argument.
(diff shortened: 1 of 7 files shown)


docs/docs/tutorials/tools.md
```diff
@@ -1547,11 +1547,13 @@ Assistant assistant = AiServices.builder(Assistant.class)
...
 ```
 
-Currently, there are two ways to handle errors inside the `ToolArgumentsErrorHandler`:
+There are three ways to handle errors inside the `ToolArgumentsErrorHandler`:
 
-- Throw an exception: this will stop the AI service flow.
 - Return a text message (e.g., an error description) that will be sent back to the LLM,
   allowing it to respond appropriately (for example, by correcting the error and retrying).
+- Throw an exception: this will stop the AI service flow.
+- Return `ToolErrorHandlerResult.propagateException()`: re-throws the original exception
```
dliubarskyi: Please update ToolExecutionErrorHandler/ToolArgumentsErrorHandler javadoc as well

docs/docs/tutorials/tools.md
```diff
@@ -1579,6 +1581,25 @@ try {
...
 ```
 
+**Selective propagation (re-throw only certain exceptions):**
+
+```java
+Assistant assistant = AiServices.builder(Assistant.class)
+        .chatModel(chatModel)
+        .tools(tools)
+        .toolArgumentsErrorHandler((error, errorContext) -> {
+            if (errorContext.originalException() instanceof MyCriticalException) {
```
dliubarskyi:
```suggestion
            if (errorContext.rawError() instanceof MyCriticalException) {
```
Given that we already have an `error`, it would make sense to have `xyzError` than `xyzException` to highlight the relation, WDYT?

Not sure about "original": other alternatives: unprocessed, unwrapped, raw

docs/docs/tutorials/tools.md
```diff
@@ -1579,6 +1581,25 @@ try {
...
+                return ToolErrorHandlerResult.propagateException();
+            }
+            return ToolErrorHandlerResult.text(error.getMessage());
+        })
+        .build();
+```
+
+The `error` argument is the unwrapped cause of the exception. Use `errorContext.originalException()`
```
dliubarskyi: Could you please briefly explain in the docs what does "unwrapped" mean

docs/docs/tutorials/tools.md
```diff
@@ -1579,6 +1581,25 @@ try {
...
 
+**Selective propagation (re-throw only certain exceptions):**
+
+```java
+Assistant assistant = AiServices.builder(Assistant.class)
+        .chatModel(chatModel)
+        .tools(tools)
+        .toolArgumentsErrorHandler((error, errorContext) -> {
+            if (errorContext.originalException() instanceof MyCriticalException) {
+                return ToolErrorHandlerResult.propagateException();
```
dliubarskyi: I am wondering if third option for the error handler is really needed. Since user now has access to the original/unprocessed exception, they can throw it directly from the error handler, right?

dliubarskyi (commented):
@mariofusco thank you!

docs/docs/tutorials/tools.md
mariofusco: Done.

docs/docs/tutorials/tools.md
mariofusco: Renamed to `rawError`.

docs/docs/tutorials/tools.md
mariofusco: Done.

docs/docs/tutorials/tools.md
mariofusco: You're right and this is indeed a good point. Personally I'd keep it as it is because it is more consistent with the API that is based on the return of the `ToolErrorHandlerResult`. Having all three options expressible as return values (`text(...)`, `propagateException()`, or `throw`) keeps the API uniform. Without it, "propagate the original" is the only option that requires a throw, which is asymmetric.

That said if you prefer to slim down the API and drop the `propagateException()` that's ok for me, please let me know.

dliubarskyi: > Without it, "propagate the original" is the only option that requires a throw, which is asymmetric

That is not true, currently throwing an exception is done not via returning `ToolErrorHandlerResult`, but via throwing:

```java
.toolExecutionErrorHandler((error, errorContext) -> {
    if (someCondition()) {
        throw error;
        // OR
        throw new MyCustomException(...);
        // OR
        throw errorContext.rawError(); // this is what you are trying to do
    }
})
```

So adding a third option is "do option 1, but more specific and with a different API", which does not feel right?

mariofusco: Agreed, I implemented your suggestion.

mariofusco:
@dliubarskyi please review this again.

docs/docs/tutorials/tools.md
```diff
@@ -1579,6 +1579,26 @@ try {
 }
 ```
 
+**Selective propagation (re-throw only certain exceptions):**
```
dliubarskyi:
```suggestion
##### Accessing raw exception
```

dliubarskyi (dismissed):
@mariofusco thank you!

PR #5420, opened 2026-06-12
chore: update google-genai dependency to 1.58.0

glaforge:
Bumps `com.google.genai:google-genai` dependency version to the latest `1.58.0`.


glaforge:
Integration test failure is unrelated to this PR

PR #5425, opened 2026-06-12
Expose MCP server instructions

### LINKED ISSUE #5421: [FEATURE]  MCP client: capture and expose the server instructions field from the initialize result
The MCP `initialize` result has an optional `instructions` field. The spec says it
is meant to help the client improve the model's understanding of the server, and
"may be added to the system prompt":

> `instructions?: string` — Instructions describing how to use the server and its
> features. This can be used by clients to improve the LLM's understanding of
> available capabilities. It can be thought of as a "hint" to the model and may be
> added to the system prompt.
>
> — MCP specification, `InitializeResult` (schema 2025-11-25)

The langchain4j MCP client receives this field at handshake, but does not model or
expose it. There is no way for application code to read a connected server's
instructions, so applications cannot forward them into the system prompt, which is
the field's stated purpose.

### Current behavior

Checked against `dev.langchain4j:langchain4j-mcp:1.14.1-beta24`:

- `dev.langchain4j.mcp.client.DefaultMcpClient.initialize()` already receives the
  full handshake response as a raw `JsonNode`, uses it for one debug log line, and
  then discards it:

  ```java
  private void initialize() {
      ...
      McpCallContext context = new McpCallContext(null, request);
      notifyListeners(l -> l.beforeInitialize(context));
      ...
      JsonNode capabilities =
              transport.initialize(request).get(initializationTimeout.toMillis(), MILLISECONDS);
      log.debug("MCP server capabilities: {}", capabilities.get("result"));
      notifyListeners(l -> l.afterInitialize(context));   // <-- response NOT passed
      ...
  }
  ```

  `capabilities.get("result")` contains `instructions`, but `capabilities` is a
  local variable — nothing is stored, nothing is returned.
- In latest version added `beforeInitialize` / `afterInitialize` / `onInitializeError`
  to `McpClientListener`, so there is now a hook at the right moment. But it does
  not help recover instructions: `afterInitialize(McpCallContext)` receives a
  `McpCallContext`, which is `record(InvocationContext, McpClientMessage message)`
  — and `message` is the initialize **request**, not the response. The listener is
  notified *that* initialize finished, but is never handed the response JSON, so it
  still cannot read `instructions`.
- `dev.langchain4j.mcp.protocol.McpInitializeResult.Result` has three fields —
  `protocolVersion`, `capabilities`, `serverInfo`. There is **no** `instructions`
  field, so the typed parse cannot hold it either.
- `dev.langchain4j.mcp.client.McpClient` has no `getInstructions()` and no
  `getServerInfo()` accessor.
- `dev.langchain4j.mcp.client.DefaultMcpClient` keeps `key`, `clientName`,
  `clientVersion`, `protocolVersion` as fields — not the server instructions.

### Expected behavior

1. `McpInitializeResult.Result` deserializes the optional `instructions` field
   (nullable when the server omits it).
2. `DefaultMcpClient` retains the value from the initialize response.
3. `McpClient` exposes it, e.g.:

   ```java
   /** Server instructions from the initialize result, or null if the server sent none. */
   String instructions();
   ```

This lets an application read `mcpClient.instructions()` and inject the text into
the system prompt — the use the spec describes.

### Proposed change

- Add a nullable `instructions` field to `McpInitializeResult.Result` with the
  matching Jackson mapping.
- Store it in `DefaultMcpClient` during `initialize()`. The insertion point already
  exists — `initialize()` holds the `result` node as a local `JsonNode`. Reading
  `capabilities.get("result").get("instructions")` (or the typed field once added)
  into an instance field is a few lines, no new request.
- Add `instructions()` (returning `String`, nullable) to the `McpClient` interface
  with a default of `null` to keep the change backward compatible for other
  implementations.

The fix is small precisely because the data is already received — this is wiring an
existing value through to an accessor, not adding a protocol round-trip.

An alternative (or complement) that fits the listener machinery added in
`1.14.1-beta24`: pass the initialize response into `afterInitialize`. Today
`McpCallContext.message` is the request only; if the context (or a new
`afterInitialize` overload) also carried the response `result`, a listener could
read `instructions` without any new accessor on `McpClient`. Either approach
solves it; the `instructions()` accessor is the simpler one for consumers.

### Why it matters

Forwarding server instructions into the system prompt is the field's intended use
and a common need for agents that talk to several MCP servers. Without an accessor,
each application has to wrap the transport or re-parse the handshake JSON to recover
a value the library already received.

fw-coder-tl:
## Summary

- add `McpClient.instructions()` to expose server instructions from the MCP initialize result
- store the optional initialize `result.instructions` value in `DefaultMcpClient`
- model `instructions` on `McpInitializeResult.Result` and cover serialization/client behavior

Closes #5421

## Testing

- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -pl langchain4j-mcp -DskipITs -DskipIntegrationTests test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 17) ./mvnw -pl langchain4j-mcp -Pspotless spotless:check`


fw-coder-tl:
CI is green now. Could someone please take a look when you have time?

langchain4j-mcp/src/main/java/dev/langchain4j/mcp/client/DefaultMcpClient.java
```diff
@@ -114,6 +115,7 @@ public class DefaultMcpClient implements McpClient {
     private final List<McpClientListener> listeners;
     private final McpMetaSupplier metaSupplier;
     private final McpToolResultExtractor toolResultExtractor;
+    private volatile @Nullable String instructions;
```
aliaksei-dudko: What do you think about replacing the string with McpInitializeResult?

fw-coder-tl: Changed this to store the full McpInitializeResult and have instructions() read from it. I also added JsonCreator/JsonProperty annotations so the initialize response can be deserialized into the protocol model, plus a serialization/deserialization assertion in McpProtocolSerializationTest.

Verified locally:
- ./mvnw -pl langchain4j-mcp -DskipITs -DskipIntegrationTests test
- ./mvnw -pl langchain4j-mcp -Pspotless spotless:check

fw-coder-tl: Updated this to keep the full McpInitializeResult while avoiding Jackson annotations on the public constructors. The Revapi issue is fixed and CI is fully green again.

fw-coder-tl:
Hi @jmartisk, the review suggestion has been addressed and CI is green again. Could you please take another look when you have time?

aliaksei-dudko (approved):

jmartisk (approved):

jmartisk:
Thanks!

PR #5427, opened 2026-06-13
Fix: Log exceptions when tool execution fails

### LINKED ISSUE #5402: [BUG] Default tool exception handling eats exceptions without even a log
The default tool exception handling eats exceptions thrown from within `@Tool` methods and does not even log the exception.  Unforeseen exceptions are thus hard to debug.

Can the default behavior at least log an exception?

Q1Xuan:
## Issue
Fixes #5402

## Change
When a `@Tool` method throws an exception and the default tool-execution error handling is in effect (the default configuration), the exception was silently swallowed — only its message was forwarded to the LLM, with no logging — making debugging extremely difficult.

This PR adds error logging with full stack traces when tool execution fails in the default configuration. The behavior for the LLM remains unchanged — it still receives the error message as before. Only developers now get proper logging to debug issues.

Crucially, logging happens only in the default configuration. When the user customizes error handling by configuring a `ToolExecutionErrorHandler` (via `AiServices.toolExecutionErrorHandler(...)`), the framework does not log — the custom handler is fully responsible for how errors are reported. This is enforced structurally: the logging lives inside the default handler itself, so a custom handler cannot trigger it.

Changes:
- Log tool execution failures with full stack traces at `WARN` level in the default `ToolExecutionErrorHandler` (`ToolService`)
- No framework logging when a custom `ToolExecutionErrorHandler` is configured
- Added tests verifying that the default configuration logs the exception (with stack trace) and that a custom handler does not

## General checklist

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
@@ -118,7 +118,7 @@ public class DefaultToolExecutor implements ToolExecutor {
                     throw new ToolExecutionException(e2.getCause());
                 } else {
                     Throwable cause = e2.getCause();
-                    log.error("Error executing tool '{}': {}", methodToInvoke.getName(), cause.getMessage(), cause);
+                    log.warn("Error executing tool '{}': {}", request.name(), cause.getMessage(), cause);
                     return ToolExecutionResult.builder()
                             .isError(true)
                             .resultText(errorMessage(cause))
@@ -130,7 +130,7 @@ public class DefaultToolExecutor implements ToolExecutor {
                 throw new ToolExecutionException(e.getCause());
             } else {
                 Throwable cause = e.getCause();
-                log.error("Error executing tool '{}': {}", methodToInvoke.getName(), cause.getMessage(), cause);
+                log.warn("Error executing tool '{}': {}", request.name(), cause.getMessage(), cause);
                 return ToolExecutionResult.builder()
                         .isError(true)
                         .resultText(errorMessage(cause))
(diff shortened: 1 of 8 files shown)


langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
```diff
@@ -113,19 +117,23 @@ public ToolExecutionResult executeWithContext(ToolExecutionRequest request, Invo
                 if (propagateToolExecutionExceptions) {
                     throw new ToolExecutionException(e2.getCause());
                 } else {
+                    Throwable cause = e2.getCause();
+                    log.error("Error executing tool '{}': {}", methodToInvoke.getName(), cause.getMessage(), cause);
```
dliubarskyi:
```suggestion
                    log.warn("Error executing tool '{}': {}", request.name(), cause.getMessage(), cause);
```

langchain4j/src/main/java/dev/langchain4j/service/tool/DefaultToolExecutor.java
```diff
@@ -113,19 +117,23 @@ public ToolExecutionResult executeWithContext(ToolExecutionRequest request, Invo
...
+                            .resultText(errorMessage(cause))
                             .build();
                 }
             }
         } catch (InvocationTargetException e) {
             if (propagateToolExecutionExceptions) {
                 throw new ToolExecutionException(e.getCause());
             } else {
+                Throwable cause = e.getCause();
+                log.error("Error executing tool '{}': {}", methodToInvoke.getName(), cause.getMessage(), cause);
```
dliubarskyi:
```suggestion
                log.warn("Error executing tool '{}': {}", request.name(), cause.getMessage(), cause);
```

dliubarskyi (approved):
@Q1Xuan thank you!

PR #5428, opened 2026-06-13
Fix Milvus L2 relevance score mapping

Q1Xuan:
## Summary
- Convert Milvus L2 distances to relevance scores with lower distance mapped higher.
- Pass the configured Milvus metric type into result mapping.
- Add mapper unit coverage for L2 and non-L2 score conversion.

## Verification
- MapperTest passed via Maven with embedding model download skipped.


dliubarskyi (approved):
@Q1Xuan thank you!

PR #5437, opened 2026-06-14
feat(agentic): support SupplierParameterResolver for @ParallelExecutor

### LINKED ISSUE #5378: feat(agentic): allow @ParallelExecutor methods to accept DI-injected parameters
## Problem

`@ParallelExecutor` methods are currently enforced as `static` + no-args. This means DI frameworks cannot inject beans into them — e.g. a Quarkus `ManagedExecutor` or a Spring `TaskExecutor`.

## What doesn't work today

```java
@Agent(description = "Research agent")
public interface ResearchAgent {

    @UserMessage("{{q}}")
    @ParallelAgent(outputKey = "result", subAgents = {SubAgentA.class, SubAgentB.class})
    String research(@V("q") String query);

    // This is the ONLY way to provide a custom executor today:
    @ParallelExecutor
    static Executor executor() {
        // Must be no-args — cannot accept DI-injected parameters.
        // So you have to instantiate or look up the executor manually:
        return Executors.newFixedThreadPool(4); // no context propagation
    }
}
```

On Quarkus, the user wants to inject a `ManagedExecutor` (which propagates CDI context, OTel spans, SecurityIdentity to worker threads). But this fails at build time:

```java
// FAILS — @ParallelExecutor enforces no-args
@ParallelExecutor
static Executor executor(@CdiBean ManagedExecutor managed) {
    return managed;
}
```

The validation in `DeclarativeUtil` / `ParallelAgentServiceImpl.configureParallel()` rejects any parameters on `@ParallelExecutor` methods, even though `SupplierParameterResolver` can resolve them.

## Desired behaviour

Now that #5377 is resolved (#5394 merged — `SupplierParameterResolver` generalised to all supplier types), `@ParallelExecutor` should accept resolver-provided parameters, just like `@ChatModelSupplier`, `@ContentRetrieverSupplier`, and all other supplier annotations:

```java
// SHOULD WORK — same pattern as @ChatModelSupplier with parameters
@ParallelExecutor
static Executor executor(@CdiBean ManagedExecutor managed) {
    return managed;
}
```

This eliminates the need for manual executor lookup and lets DI frameworks provide context-propagating executors through the standard supplier parameter mechanism.

## Scope

The generalised `SupplierParameterResolver` SPI is already in place. This requires:
1. Relaxing the no-args validation for `@ParallelExecutor` methods when a parameter resolver is registered
2. Applying `invokeSupplierWithResolvers` to the `@ParallelExecutor` resolution path in `DeclarativeUtil` / `ParallelAgentServiceImpl.configureParallel()`

No new API — this is applying the existing resolver mechanism to one more annotation type.

DragonFSKY:
## Issue
Closes #5378

## Change
- Resolve `@ParallelExecutor` method parameters through the existing `SupplierParameterResolver` mechanism.
- Reuse the same executor resolution path for declarative `@ParallelAgent` and `@ParallelMapperAgent` services.
- Fail fast when a parameterized `@ParallelExecutor` has no registered resolver, instead of passing unresolved parameters through the default path.
- Add tests for resolved parallel executors, resolved parallel mapper executors, and the missing-resolver failure case.

## Tests
- `./mvnw -pl langchain4j-agentic -Dtest=SupplierParameterResolverTest test`
- `./mvnw -pl langchain4j-agentic spotless:check`
- `./mvnw -pl langchain4j-agentic test`

## General checklist

### DIFF
--- a/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/DeclarativeUtil.java
+++ b/langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/DeclarativeUtil.java
@@ -2,6 +2,7 @@ package dev.langchain4j.agentic.declarative;
 
 import static dev.langchain4j.agentic.internal.AgentUtil.AGENTIC_SCOPE_ARG_NAME;
 import static dev.langchain4j.agentic.internal.AgentUtil.agentInvocationArguments;
+import static dev.langchain4j.agentic.internal.AgentUtil.argumentFromParameter;
 import static dev.langchain4j.agentic.internal.AgentUtil.argumentsFromMethod;
 import static dev.langchain4j.agentic.internal.AgentUtil.getAnnotatedMethodOnClass;
 
@@ -15,6 +16,7 @@ import dev.langchain4j.agentic.agent.ErrorRecoveryResult;
 import dev.langchain4j.agentic.observability.AgentListener;
 import dev.langchain4j.agentic.planner.AgentArgument;
 import dev.langchain4j.agentic.planner.AgenticService;
+import dev.langchain4j.agentic.planner.AgenticSystemConfigurationException;
 import dev.langchain4j.agentic.scope.AgenticScope;
 import dev.langchain4j.memory.ChatMemory;
 import dev.langchain4j.model.chat.ChatModel;
@@ -27,12 +29,14 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.lang.reflect.Parameter;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.Executor;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
@@ -215,6 +219,14 @@ public class DeclarativeUtil {
         buildListener(agentServiceClass, builder);
     }
 
+    public static Optional<Executor> parallelExecutor(Class<?> agentServiceClass) {
+        return selectMethod(
+                        agentServiceClass,
+                        method -> method.isAnnotationPresent(ParallelExecutor.class)
+                                && Executor.class.isAssignableFrom(method.getReturnType()))
+                .map(method -> invokeParallelExecutor(agentServiceClass, method));
+    }
+
     private static <T> Optional<Function<ErrorContext, ErrorRecoveryResult>> buildErrorHandler(
             Class<T> agentServiceClass) {
         return selectMethod(agentServiceClass, method -> method.isAnnotationPresent(ErrorHandler.class))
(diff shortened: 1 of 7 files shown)


langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/DeclarativeUtil.java
```diff
@@ -256,27 +268,41 @@ public static Predicate<AgenticScope> agenticScopePredicate(Method predicateMeth
     }
 
     private static <T> T invokeSupplierWithResolvers(Class<?> agentType, Method method, Class<T> targetClass) {
+        return invokeSupplierWithResolvers(agentType, method, targetClass, false);
+    }
+
+    private static <T> T invokeSupplierWithResolvers(
+            Class<?> agentType, Method method, Class<T> targetClass, boolean requireAllParametersResolved) {
```
mariofusco: I don't understand this new `requireAllParametersResolved` argument, I guess we want to fail fast every time a parameter cannot be resolved, or am I missing something?

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/DeclarativeUtil.java
```diff
@@ -256,27 +268,41 @@ public static Predicate<AgenticScope> agenticScopePredicate(Method predicateMeth
...
         return fn.apply(null);
     }
 
     private static <T> Function<AgenticScope, T> agenticScopeFunctionWithSupplierParameterResolver(
             Class<?> agentType, Method functionMethod, Class<T> targetClass) {
+        return agenticScopeFunctionWithSupplierParameterResolver(agentType, functionMethod, targetClass, false);
+    }
+
+    private static <T> Function<AgenticScope, T> agenticScopeFunctionWithSupplierParameterResolver(
+            Class<?> agentType, Method functionMethod, Class<T> targetClass, boolean requireAllParametersResolved) {
```
mariofusco: Same here.

langchain4j-agentic/src/main/java/dev/langchain4j/agentic/declarative/DeclarativeUtil.java
```diff
@@ -256,27 +268,41 @@ public static Predicate<AgenticScope> agenticScopePredicate(Method predicateMeth
...
+
+    private static <T> T invokeSupplierWithResolvers(
+            Class<?> agentType, Method method, Class<T> targetClass, boolean requireAllParametersResolved) {
         if (method.getParameterCount() == 0) {
             return invokeStatic(method);
         }
         List<SupplierParameterResolver> resolvers = getSupplierParameterResolvers();
         if (resolvers.isEmpty()) {
+            if (requireAllParametersResolved) {
+                throw missingSupplierParameterResolver(method, method.getParameters()[0]);
```
mariofusco: Can you please add a test where this exception is thrown?

mariofusco (changes requested):
Nice job, thank you, please just check if that `requireAllParametersResolved` boolean is really necessary and if so clarify why.

DragonFSKY:
Thanks, I removed the `requireAllParametersResolved` switch and split `@ParallelExecutor` into a dedicated resolver path. Regular supplier methods keep the existing mixed resolver/agent-argument behavior, while `@ParallelExecutor` now requires every parameter to be resolved by a `SupplierParameterResolver` and fails fast otherwise.

I also added a test for the non-matching resolver case, so the exception path is covered when resolvers are registered but none supports the `@ParallelExecutor` parameter.

mariofusco (approved):

PR #5456, opened 2026-06-16
fix: Throw typed exception for null-valued required shell tool argument

### LINKED ISSUE #5455: [BUG] NPE on null-valued required shell tool argument
**Describe the bug**
`RunShellCommandToolExecutor.getRequiredArgument` guards a missing argument key but not a present-but-null value.
When an LLM emits `{"command": null}`, parsed map `{command=null}` passes `containsKey` and `arguments.get(name).toString()` throws a raw `NullPointerException`.
That call runs before the `try` in `executeWithContext`, so the NPE escapes un-wrapped, bypassing the typed exception selected by `throwToolArgumentsExceptions`.
The missing-key path already throws the correct typed exception; only the null-value path is inconsistent.

**Log and Stack trace**
`java.lang.NullPointerException` at `RunShellCommandToolExecutor.getRequiredArgument` (`arguments.get(argumentName).toString()`).

**To Reproduce**
```java
executor.executeWithContext(
    ToolExecutionRequest.builder().name("run_shell_command").arguments("{\"command\": null}").build(),
    null);
```

**Expected behavior**
Treat a present-but-null required argument like a missing one: a typed `ToolExecutionException` (or `ToolArgumentsException` when `throwToolArgumentsExceptions` is true), not a raw NPE.

- LangChain4j version: current/main
- LLM(s) used: N/A
- Java version: 17

**Additional context**
Module: experimental/langchain4j-experimental-skills-shell.

thswlsqls:
## Issue
Closes #5455

## Change
`RunShellCommandToolExecutor.getRequiredArgument` guarded a missing argument key, not a present-but-null value.
An LLM emitting `{"command": null}` produced `{command=null}`, so `containsKey` passed the guard and `arguments.get(name).toString()` threw a raw `NullPointerException`. That call runs before the `try` in `executeWithContext`, so it escaped un-wrapped, bypassing the typed `ToolExecutionException`/`ToolArgumentsException` selected by `throwToolArgumentsExceptions`.
The fix resolves the value once and null-checks it, treating a null value like a missing key with the same message. Backward compatible: missing-key behavior and message unchanged; no signature change.

```java
private String getRequiredArgument(String argumentName, Map<String, Object> arguments) {
    Object value = isNullOrEmpty(arguments) ? null : arguments.get(argumentName);
    if (value == null) {
        throwException("Missing required tool argument '%s'".formatted(argumentName));
    }
    return value.toString();
}
```

Added two negative-case unit tests to `RunShellCommandToolExecutorTest`: `{"command": null}` yields `ToolExecutionException` (default) and `ToolArgumentsException` (when `throwToolArgumentsExceptions(true)`).

## General checklist

### DIFF
--- a/experimental/langchain4j-experimental-skills-shell/src/main/java/dev/langchain4j/skills/shell/RunShellCommandToolExecutor.java
+++ b/experimental/langchain4j-experimental-skills-shell/src/main/java/dev/langchain4j/skills/shell/RunShellCommandToolExecutor.java
@@ -1,5 +1,8 @@
 package dev.langchain4j.skills.shell;
 
+import static dev.langchain4j.internal.Utils.isNullOrEmpty;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
+
 import dev.langchain4j.agent.tool.ToolExecutionRequest;
 import dev.langchain4j.exception.ToolArgumentsException;
 import dev.langchain4j.exception.ToolExecutionException;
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thank you!

PR #5467, opened 2026-06-17
fix: Omit blank/null Content-Type part header in Apache multipart body

thswlsqls:
## Issue
Closes #5466

## Change
`MultipartBodyPublisher.addFile` in `langchain4j-http-client-apache` concatenated `"Content-Type: " + file.contentType()` into the part header unguarded, emitting a literal `Content-Type: null` line for null and an empty `Content-Type:` for blank — both malformed multipart. This is reachable on a normal path: `Audio.mimeType()` is null when unset and `DefaultOpenAiClient.audioTranscription()` forwards `file.mimeType()` unguarded.

Fix: omit the `Content-Type:` line when null or blank via `dev.langchain4j.internal.Utils.isNullOrBlank`, mirroring OkHttp. The valid-content-type path is unchanged (backward compatible).

Tests: three `MultipartBodyPublisherTest` cases — null and blank (no `Content-Type:` line) and a valid regression (`Content-Type: audio/wav` still emitted).

Only the Apache client changes; the JDK client (`langchain4j-http-client-jdk`) has the identical bug, held for a follow-up after this merges (probe-first).

Note: Spotless ratchet reindented a few text blocks (CI-required), not an unrelated change.

## General checklist
- [X] There are no breaking changes (API, behaviour)
- [X] I have added unit and/or integration tests for my change
- [X] The tests cover both positive and negative cases
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green 
- [ ] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs) 
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features) 
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

Issue #5466, opened 2026-06-17
[BUG] Apache HTTP client emits malformed Content-Type for null/blank multipart parts


thswlsqls:
**Describe the bug**
`MultipartBodyPublisher.addFile` in `langchain4j-http-client-apache` concatenates `"Content-Type: " + file.contentType()` into the part header unguarded. A null content type emits a literal `Content-Type: null` line; a blank one emits an empty `Content-Type:`. Both are malformed multipart.

This is reachable on a normal path: `Audio.mimeType()` is null when unset and `DefaultOpenAiClient.audioTranscription()` forwards `file.mimeType()` unguarded. The OkHttp sibling omits the per-part Content-Type for blank/null, so the clients are inconsistent.

**Log and Stack trace**
N/A (wrong output, not an exception).

**To Reproduce**
Build a multipart file part with a null or blank content type.

**Expected behavior**
Omit the `Content-Type:` line when the content type is null or blank, matching OkHttp.

**Please complete the following information:**
- LangChain4j version: 1.17.0-SNAPSHOT (main)
- LLM(s) used: N/A
- Java version: N/A
- Spring Boot version (if applicable): N/A

**Additional context**
The JDK client (`langchain4j-http-client-jdk`) has the identical bug; a follow-up PR will address it after this merges.

PR #5480, opened 2026-06-18
fix: omit blank/null Content-Type part header in JDK multipart body

Q1Xuan:
## Issue
Relates to #5466. That issue is fixed for the **Apache** client in #5467 and explicitly notes that *"The JDK client (`langchain4j-http-client-jdk`) has the identical bug"*. This PR fixes the **JDK** client half. No dedicated JDK issue exists, so this references #5466 rather than closing it — happy to track it under a separate issue if maintainers prefer.

## Change
`MultipartBodyPublisher.addFile` in `langchain4j-http-client-jdk` concatenated `"Content-Type: " + file.contentType()` into the part header unconditionally, so:

- a **null** content type emitted a literal `Content-Type: null` line, and
- a **blank** content type emitted an empty `Content-Type:` line,

both of which are malformed multipart. This is reachable on a normal path: `Audio.mimeType()` is null when unset and `DefaultOpenAiClient.audioTranscription()` forwards `file.mimeType()` unguarded — and the JDK client is the default HTTP client.

The fix guards the header with `Utils.isNullOrBlank(...)` so the per-part `Content-Type` line is omitted when the content type is null or blank, matching the OkHttp client and the accepted Apache fix (#5467). Behaviour is unchanged when a content type is present.

Added unit tests for the null and blank cases; the existing `should_build_body_with_file` test already covers the positive (content-type-present) case.

## General checklist
- [ ] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j) modules, and they are all green

> **Note on testing:** I ran the changed module's unit tests (`langchain4j-http-client-jdk`, `MultipartBodyPublisherTest`) — 5/5 green — and `spotless:check` passes. The module's `*IT` integration tests require live HTTP endpoints and were not run. The change is isolated to `langchain4j-http-client-jdk`, which the core/main modules do not depend on.

### DIFF
--- a/http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/MultipartBodyPublisher.java
+++ b/http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/MultipartBodyPublisher.java
@@ -1,13 +1,13 @@
 package dev.langchain4j.http.client.jdk;
 
+import static dev.langchain4j.internal.Utils.isNullOrBlank;
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import dev.langchain4j.Experimental;
 import dev.langchain4j.http.client.FormDataFile;
-
 import java.net.http.HttpRequest;
 import java.util.ArrayList;
 import java.util.List;
-
-import static java.nio.charset.StandardCharsets.UTF_8;
 
 @Experimental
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@Q1Xuan thank you!

PR #5486, opened 2026-06-19
Update Azure SDK BOM to 1.3.7 and migrate to azure-search-documents 12.0.0

jdubois:
## Issue

N/A - routine dependency maintenance prompted by the new Azure Java SDK release.

## Change

Bumps `com.azure:azure-sdk-bom` from `1.3.5` to `1.3.7` (latest). Most managed libraries are minor/patch bumps, but the new BOM pulls two changes that require work:

- **`azure-search-documents` 11.8.1 -> 12.0.0** (major release) used by `langchain4j-azure-ai-search`.
- **`azure-core` 1.57.1 -> 1.58.0**, which shifts the required Netty/Reactor stack.

**azure-search-documents 12.0.0 migration** (`langchain4j-azure-ai-search`). The public langchain4j API of the module is unchanged; only the internal calls into the Azure SDK were rewritten:

- `search(text, options, Context)` -> `search(options.setSearchText(text))`
- `SearchDocument` / `getDocument(...)` removed -> read fields via `SearchResult.getAdditionalProperties()` (a `Map`)
- `uploadDocuments` / `deleteDocuments` removed -> `indexDocuments(IndexDocumentsBatch)` built from `IndexAction`s (new private `toUploadBatch` / `toSearchDocument` helpers)
- `VectorSearchOptions` / `SemanticSearchOptions` removed -> `setVectorQueries(...)` and `setSemanticConfigurationName(...)`
- `getSemanticSearch().getRerankerScore()` -> `getRerankerScore()`; `setKNearestNeighborsCount` -> `setKNearestNeighbors`; `new SearchIndex(name, fields)` constructor

**Dependency re-alignment.** The bump to `azure-core 1.58.0` raises the minimum Netty/Reactor versions, so the local `dependencyManagement` workaround pins in `github-models`, `azure-cosmos-nosql`, `document-loader-azure-storage-blob`, and `code-execution-engine-azure-acads` were updated (`reactor-core 3.7.17`, `reactor-netty-http 1.2.16`, `azure-core-http-netty 1.16.4` and siblings, `netty-bom 4.1.132.Final`) so the `maven-enforcer` `RequireUpperBoundDeps` rule keeps passing.

Note for reviewers: this is a major version jump of `azure-search-documents`, so please give the rewritten index/read/write paths in `AbstractAzureAiSearchEmbeddingStore` a careful look. End-user behaviour of the module is intended to be unchanged. Integration tests require live Azure credentials and were not run; all unit tests pass locally across every affected Azure module.

## General checklist

N/A - no new module.

N/A - no new embedding store integration.

- [ ] I have manually verified that the `AzureAiSearchEmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

### DIFF
--- a/langchain4j-parent/pom.xml
+++ b/langchain4j-parent/pom.xml
@@ -53,7 +53,7 @@
         <assertk.version>0.28.1</assertk.version>
         <awaitility.version>4.3.0</awaitility.version>
         <aws.java.sdk.version>2.41.34</aws.java.sdk.version>
-        <azure-sdk.version>1.3.5</azure-sdk.version>
+        <azure-sdk.version>1.3.7</azure-sdk.version>
         <jackson.version>2.21.3</jackson.version>
         <jspecify.version>1.0.0</jspecify.version>
         <jtokkit.version>1.1.0</jtokkit.version>
@@ -531,13 +531,6 @@
                     <groupId>org.revapi</groupId>
                     <artifactId>revapi-maven-plugin</artifactId>
                     <version>0.15.1</version>
-                    <dependencies>
-                        <dependency>
-                            <groupId>org.revapi</groupId>
-                            <artifactId>revapi-java</artifactId>
-                            <version>0.28.4</version>
-                        </dependency>
-                    </dependencies>
                     <configuration>
                         <oldVersion>RELEASE</oldVersion>
                         <analysisConfigurationFiles>
@@ -571,6 +564,13 @@
                             </revapi.differences>
                         </analysisConfiguration>
                     </configuration>
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.revapi</groupId>
+                            <artifactId>revapi-java</artifactId>
+                            <version>0.28.4</version>
+                        </dependency>
+                    </dependencies>
                 </plugin>
             </plugins>
         </pluginManagement>
--- a/code-execution-engines/langchain4j-code-execution-engine-azure-acads/pom.xml
+++ b/code-execution-engines/langchain4j-code-execution-engine-azure-acads/pom.xml
@@ -19,7 +19,7 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-bom</artifactId>
-                <version>4.1.130.Final</version>
+                <version>4.1.132.Final</version>
                 <type>pom</type>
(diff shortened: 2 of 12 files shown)


langchain4j-parent/pom.xml
```diff
@@ -571,6 +564,13 @@
                             </revapi.differences>
                         </analysisConfiguration>
                     </configuration>
+                    <dependencies>
+                        <dependency>
+                            <groupId>org.revapi</groupId>
+                            <artifactId>revapi-java</artifactId>
```
dliubarskyi, line 570: Is this change required? If yes, why?

dliubarskyi (commented):
@jdubois thank you! Did you test these changes against the real API? I do not have setups for Cosmos

langchain4j-parent/pom.xml
jdubois: Yes, but only as a side-effect of spotless, not a deliberate change.

CI runs `mvn -Pspotless validate` with `ratchetFrom origin/main`, so the sortPom formatter only enforces on files that differ from `main`. This PR bumps `azure-sdk.version` (1.3.5 -> 1.3.7) in this file, which makes the ratchet apply full sortPom formatting to the whole file. sortPom's canonical plugin order puts `<configuration>` before `<dependencies>`, and the existing revapi block had them reversed (a pre-existing ordering that the ratchet hid while this file was untouched on `main`). So `spotless validate` fails until the `<dependencies>` block is moved after `<configuration>`.

I verified this by taking a clean `main`, applying only the version bump, and running `spotless:check` - it demanded this exact reorder. Happy to revert it if you'd prefer to keep the diff minimal, but then the spotless check fails.

jdubois:
@dliubarskyi no I didn't

dliubarskyi:
@jdubois could you please test it? Otherwise it is quite dangerous to merge and release this without testing.

jdubois:
@dliubarskyi you were right, there was a bug!!!! It might be a bit older than this commit, which is worrying....
Integration tests now pass in the langchain4j-azure-cosmos-nosql module, here's a screenshot
Screenshot 2026-06-23 at 16 14 56

dliubarskyi (approved):
@jdubois thank you!

PR #5491, opened 2026-06-19
Introduce AgentsRegistry SPI for dynamic agent discovery

carlesarnal:
Adds an `AgentsRegistry` SPI to `langchain4j-agentic` so that agentic systems can dynamically discover agents at runtime.

The supervisor pattern re-discovers agents on each `init()` call, making newly registered agents available between invocations without rebuilding the system.

Motivated by quarkiverse/quarkus-langchain4j#2501 — this enables implementations (e.g. Apicurio Registry-based) to provide agent discovery as a first-class agentic concept rather than using the agent-as-tool pattern.

I have three points where I would really appreciate feedback:

On discoverAgents() taking no context:
The current discoverAgents() is parameterless — the registry has no way to know what system is asking or filter by capabilities. For the Apicurio use case we might want to filter by tags, agent type, or capabilities. Should we add an overload like discoverAgents(AgentDiscoveryContext) now, or keep it simple and add filtering later if needed?

On agentsRegistry() on AgenticService:
agentsRegistry() is on AgenticService, which means all workflow types (sequence, parallel, loop, conditional) inherit it. But dynamic agent discovery only makes sense for planner-based patterns like supervisor — sequences and parallel workflows have statically wired agents. Should this method live only on SupervisorAgentService (and PlannerBasedService) instead of on the generic interface?

On toMap() mutability in SupervisorPlanner.init():
The putIfAbsent call after toMap() relies on Collectors.toMap() returning a mutable map, which is an implementation detail not guaranteed by the spec. Should use new HashMap<>(...) explicitly to be safe?


mariofusco:
First of all I strongly suggest to develop this feature in a TDD way. Having one or more practical use cases will greatly help us to better shape this feature and understand how to integrate it with the rest of the agentic framework. Please implement a few test cases using a mock registry and then we could also use them to test the Apicurio integration in practice.

A few random feedback and general advices:
- It's ok, for all `Planner` implementations to be provided with an  `AgentsRegistry`. It's up to the specific single implementation to decide if to make use of the registrty or not.
- The `discoverAgents()` method should be a function of the `AgenticScope` so it could be contextualized with the current state of the agentic system.
- For what regards the supervisor probably the discovery should run at each iteration and not only on the `init()` so that the supervisor itself could be dynamically provided with more agents during the execution, based on how the execution is proceeding. This is one of the thing that we could better figure out through practical use cases.
- What worries me the most at the moment is that the registry should provide `AgentInstance`s so they could be directly used and invoked by the agentic patterns. I'm not entirely sure how to do this, Maybe the possibility to create agents programmatically could help to create a sort of adapting layer. This is something else that will require some test cases to be properly designed. 
- I wonder if we could also have a A2A based implementation of the registry, but this is something that we could figure out later.

Please let me know if my suggestions make sense and if you're willing to keep working on this. I have time to work on this right now, so I could take over if you prefer.

carlesarnal:
> First of all I strongly suggest to develop this feature in a TDD way. Having one or more practical use cases will greatly help us to better shape this feature and understand how to integrate it with the rest of the agentic framework. Please implement a few test cases using a mock registry and then we could also use them to test the Apicurio integration in practice.
> 
> A few random feedback and general advices:
> 
> * It's ok, for all `Planner` implementations to be provided with an  `AgentsRegistry`. It's up to the specific single implementation to decide if to make use of the registrty or not.
> * The `discoverAgents()` method should be a function of the `AgenticScope` so it could be contextualized with the current state of the agentic system.
> * For what regards the supervisor probably the discovery should run at each iteration and not only on the `init()` so that the supervisor itself could be dynamically provided with more agents during the execution, based on how the execution is proceeding. This is one of the thing that we could better figure out through practical use cases.
> * What worries me the most at the moment is that the registry should provide `AgentInstance`s so they could be directly used and invoked by the agentic patterns. I'm not entirely sure how to do this, Maybe the possibility to create agents programmatically could help to create a sort of adapting layer. This is something else that will require some test cases to be properly designed.
> * I wonder if we could also have a A2A based implementation of the registry, but this is something that we could figure out later.
> 
> Please let me know if my suggestions make sense and if you're willing to keep working on this. I have time to work on this right now, so I could take over if you prefer.

I'm happy to contribute, but it probably makes more sense for you to go ahead and take over the SPI design. You know the agentic internals much better and the points about contextual discovery via AgenticScope and per-iteration re-discovery are things I wouldn't have gotten right without a lot of back and forth.

I'll pick up the Apicurio implementation once the SPI stabilizes.

mariofusco (changes requested):
Replaced by https://github.com/langchain4j/langchain4j/pull/5551

PR #5501, opened 2026-06-21
fix: Escape double quotes and backslashes in Cosmos DB NoSQL filter values

### LINKED ISSUE #5500: [BUG] Cosmos DB NoSQL filter mapper does not escape double quotes and backslashes in values
**Describe the bug**
`DefaultAzureCosmosDBNoSqlFilterMapper` wraps String filter values in double quotes without escaping embedded `"` or `\`, so a valid metadata value containing a double quote produces broken Cosmos DB NoSQL SQL: an unterminated string literal that corrupts the WHERE clause. All String paths are affected because `mapIsEqualTo`, `mapContainsString`, `mapInValues`, and the full-text mappings route through `formatValue(Object)`.

**Log and Stack trace**
N/A (silent wrong output, not an exception).

**To Reproduce**
```java
AzureCosmosDBNoSqlFilterMapper mapper = new DefaultAzureCosmosDBNoSqlFilterMapper();
String sql = mapper.map(new IsEqualTo("size", "12\"display"));
// actual:   c.size = "12"display"
// expected: c.size = "12\"display"
```

**Expected behavior**
Embedded double quotes and backslashes in String values are escaped using Cosmos NoSQL backslash escaping (`\"`, `\\`), producing valid SQL.

- LangChain4j version: 1.11.8
- LLM(s) used: N/A
- Java version: 17

**Additional context**
The fix is isolated to `formatValue(Object)`; values without quotes or backslashes are unchanged.

thswlsqls:
## Issue

Closes #5500

## Change
`DefaultAzureCosmosDBNoSqlFilterMapper.formatValue(Object)` wrapped String values in double quotes without escaping embedded `"` or `\`, producing broken Cosmos DB NoSQL SQL (e.g. `IsEqualTo("size","12\"display")` -> `c.size = "12"display"`).

`formatValue` is the single entry point for every String filter path (`mapIsEqualTo`, `mapContainsString`, `mapInValues`, `mapFullText*`), so the defect reaches all of them.

Fix: escape backslashes first, then double quotes:
```java
String escaped = ((String) value).replace("\\", "\\\\").replace("\"", "\\\"");
return "\"" + escaped + "\"";
```

Values without `"` or `\` are emitted unchanged, so existing tests stay green. Added 3 unit tests: a quote, a backslash, and a quote propagated through `ContainsString`.

The sibling `langchain4j-azure-ai-search` mapper shares the same gap; left for a follow-up.

## Pre-checks
- Unit tests: 21 run, 0 failures (18 existing + 3 new). The 3 new tests fail on the unpatched mapper and pass after the fix.
- Spotless: `spotless:check` cannot run in the worktree (JGit "Cannot find git repository"), so it ran in the main checkout — `apply` made no diff, `check` passed, and the checkout was restored to pristine.
- Integration tests (`*IT`, `@EnabledIfEnvironmentVariable`) not run (require Azure credentials); none added or modified.

## General checklist
- [ ] I have manually run all the unit and integration tests in the core and main modules 
- [ ] I have added an example in the examples repo 
- [ ] I have added/updated Spring Boot starter(s)

### DIFF
--- a/langchain4j-azure-cosmos-nosql/src/main/java/dev/langchain4j/rag/content/retriever/azure/cosmos/nosql/DefaultAzureCosmosDBNoSqlFilterMapper.java
+++ b/langchain4j-azure-cosmos-nosql/src/main/java/dev/langchain4j/rag/content/retriever/azure/cosmos/nosql/DefaultAzureCosmosDBNoSqlFilterMapper.java
@@ -152,7 +152,8 @@ public class DefaultAzureCosmosDBNoSqlFilterMapper implements AzureCosmosDBNoSql
 
     private String formatValue(Object value) {
         if (value instanceof String) {
-            return "\"" + value + "\"";
+            String escaped = ((String) value).replace("\\", "\\\\").replace("\"", "\\\"");
+            return "\"" + escaped + "\"";
         }
         return value.toString();
     }
--- a/langchain4j-azure-cosmos-nosql/src/test/java/dev/langchain4j/rag/content/retriever/azure/cosmos/nosql/DefaultAzureCosmosDBNoSqlFilterMapperTest.java
+++ b/langchain4j-azure-cosmos-nosql/src/test/java/dev/langchain4j/rag/content/retriever/azure/cosmos/nosql/DefaultAzureCosmosDBNoSqlFilterMapperTest.java
@@ -140,6 +140,27 @@ class DefaultAzureCosmosDBNoSqlFilterMapperTest {
     }
 
     @Test
+    void map_escapesDoubleQuotesInValue() {
+        IsEqualTo filter = new IsEqualTo("size", "12\"display");
+        String result = mapper.map(filter);
+        assertThat(result).isEqualTo("c.size = \"12\\\"display\"");
+    }
+
+    @Test
+    void map_escapesBackslashInValue() {
+        IsEqualTo filter = new IsEqualTo("path", "a\\b");
+        String result = mapper.map(filter);
+        assertThat(result).isEqualTo("c.path = \"a\\\\b\"");
+    }
+
+    @Test
+    void map_escapesDoubleQuotesInContainsString() {
+        ContainsString filter = new ContainsString("desc", "a\"b");
+        String result = mapper.map(filter);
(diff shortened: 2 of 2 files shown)


jdubois:
Verdict: ✅ Approve. Correct, minimal, well-targeted fix for a real bug. All 7 CI checks pass (unit tests on JDK 17/21/25, spotless, compliance).

## Why the fix is correct

formatValue is the single choke point for every String filter path (mapIsEqualTo, mapContainsString, mapInValues, mapFullText*), so escaping there covers all cases in one place. I confirmed the consuming code inlines the result directly into SQL text:

```Java
// AbstractAzureCosmosDBNoSqlEmbeddingStore.java:448
String query = "SELECT TOP @topK * FROM c" + " WHERE " + filterMapper.map(filter);
```

Values are concatenated raw (not parameterized), so unescaped quotes genuinely produced broken/unterminated SQL. The fix is both a correctness and a safety improvement.

Escaping order is right: backslash first (\ → \\), then quote (" → \"). Doing it the other way would double-escape the backslash introduced by quote-escaping. The three new tests decode correctly and cover quote, backslash, and quote-through-ContainsString. Happy-path values are emitted unchanged, so existing tests stay green.

## Non-blocking notes (suggest as follow-ups, not changes to this PR)

- Field keys are still uninjected. filter.key() is interpolated via "c.%s" with no escaping/validation anywhere in the mapper — a separate (and arguably broader) injection vector than values. Worth a follow-up issue. Keys are usually developer-controlled, so lower priority.
- Other control characters. Only " and \ are escaped. A value containing a raw newline/tab could still produce questionable SQL. Realistically rare; fine to defer.
- Long-term: the whole mapper builds SQL by string concatenation. Migrating to Cosmos SqlParameter bindings would eliminate this class of bug entirely — but that's a large refactor well beyond this PR's scope.
The author already flagged the sibling langchain4j-azure-ai-search mapper as having the same gap and deferred it — reasonable.

jdubois (approved):

thswlsqls:
Thanks @jdubois! Agreed the key-escaping and `SqlParameter` migration are separate, broader items — I'll keep this PR to value escaping and open follow-up issues for those if useful.

PR #5503, opened 2026-06-21
fix: prevent integer overflow in AwsDocumentConverter for long and BigInteger tool arguments

### LINKED ISSUE #5502: [BUG] AwsDocumentConverter silently overflows long/BigInteger tool arguments
**Describe the bug**

`AwsDocumentConverter.getDocument(JsonNode)` silently overflows large integers. The integer branch matches `isInt() || isLong() || isShort() || isBigInteger()` but converts with `Document.fromNumber(value.asInt())`. Jackson `asInt()` narrows `long`/`BigInteger` to `(int)`, so any value outside the int range is corrupted.

This is on a normal call path: `AbstractBedrockChatModel.convertToolRequests()` converts LLM tool-call argument JSON to a `Document` via `documentFromJson(...)`. A tool argument like a large id or timestamp is therefore mangled before being sent to Bedrock.

The same class is internally inconsistent: the reverse `documentToObject()` uses `doc.asNumber()` and the `double` branch uses `asDouble()`, both width-preserving. Only the integer branch narrows.

**To Reproduce**

```java
Document d = AwsDocumentConverter.documentFromJson("{\"id\":9223372036854775807}");
d.asMap().get("id").asNumber().longValue(); // returns -1 instead of 9223372036854775807
```

**Expected behavior**

The full integer value is preserved for `long` and `BigInteger` inputs; in-range `int` values are unchanged.

- LangChain4j version: 1.8.0-SNAPSHOT (main)
- LLM(s) used: N/A
- Java version: N/A

**Additional context**

Module: `langchain4j-bedrock`. Fix: use `Document.fromNumber(value.bigIntegerValue())` (a `BigInteger` overload exists in sdk-core).

thswlsqls:
## Issue

Closes #5502

## Change

`AwsDocumentConverter.getDocument(JsonNode)` matched the integer branch on `isInt() || isLong() || isShort() || isBigInteger()` but converted with `Document.fromNumber(value.asInt())`. Jackson `asInt()` narrows `long`/`BigInteger` to `(int)`, silently corrupting values outside the int range (`9223372036854775807` becomes `-1`).

This is reached on a normal path: `AbstractBedrockChatModel.convertToolRequests()` converts LLM tool-call argument JSON to a `Document` via `documentFromJson(...)`, so large integer tool arguments were mangled before being sent to Bedrock.

Fix: convert with `value.bigIntegerValue()`. A `Document.fromNumber(BigInteger)` overload exists in sdk-core, so no new import or dependency is needed. This is width-preserving and symmetric with the reverse `documentToObject()` (`asNumber()`) and the `double` branch (`asDouble()`). In-range `int` values are unchanged (backward compatible).

Tests: added `documentFromJson` round-trip tests for `Long.MAX_VALUE`, a value above `Long.MAX_VALUE`, and an in-range int regression. The first two fail on the old code and pass after the fix.

Note: spotless ratchet reformatted three pre-existing text blocks in the touched test file (required to pass CI `spotless:check`).

## General checklist

### DIFF
--- a/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/AwsDocumentConverter.java
+++ b/langchain4j-bedrock/src/main/java/dev/langchain4j/model/bedrock/AwsDocumentConverter.java
@@ -90,7 +90,7 @@ class AwsDocumentConverter {
         } else if (value.isDouble() || value.isFloat() || value.isBigDecimal()) {
             doc = Document.fromNumber(value.asDouble());
         } else if (value.isInt() || value.isLong() || value.isShort() || value.isBigInteger()) {
-            doc = Document.fromNumber(value.asInt());
+            doc = Document.fromNumber(value.bigIntegerValue());
         } else if (value.isArray()) {
             List<Document> list = new ArrayList<>();
             for (JsonNode node : value) {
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thanks!

PR #5388, opened 2026-06-08
feat: implement image token estimation for HIGH/AUTO detail levels

acefun29:
### PR Description

Image token estimation in AzureOpenAiTokenCountEstimator was hardcoded to 85 tokens regardless of detail level. This significantly underestimates costs for HIGH detail images (which actually cost 765+ tokens for a typical 1024x1024 image).

Implement detail-level-aware token estimation:
- LOW: 85 tokens (unchanged, per OpenAI documentation)
- MEDIUM: 400 tokens (conservative estimate)
- HIGH/ULTRA_HIGH/AUTO: 765 tokens (typical for 1024x1024 image)

The exact token count for HIGH detail depends on image dimensions (85 + 170 per 512x512 tile), but since we don't have access to dimensions in the estimator, we use a conservative default.

References:
- OpenAI Vision documentation: https://platform.openai.com/docs/guides/vision
- ImageContent.DetailLevel enum in langchain4j-core

---

## Issue
Closes # (discovered via code audit)

## Change

Implement detail-level-aware image token estimation in `AzureOpenAiTokenCountEstimator`.

**Before:** All images estimated at 85 tokens regardless of detail level.

**After:**
- LOW: 85 tokens (unchanged, per OpenAI documentation)
- MEDIUM: 400 tokens (conservative estimate)
- HIGH/ULTRA_HIGH/AUTO: 765 tokens (typical for 1024x1024 image)

The exact token count for HIGH detail depends on image dimensions (85 + 170 per 512x512 tile), but since we don't have access to dimensions in the estimator, we use a conservative default.

The `ImageContent.detailLevel()` method and `DetailLevel` enum already exist in langchain4j-core but were being ignored.

## General checklist
- [X] There are no breaking changes (API, behaviour)
- [ ] I have added unit and/or integration tests for my change
- [X] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [ ] I have added/updated the documentation

PR #5509, opened 2026-06-21
fix: forward image detail level to Azure OpenAI chat requests

### LINKED ISSUE #5508: [BUG] Azure OpenAI: image detailLevel is ignored and not sent to the API
**Describe the bug**

`InternalAzureOpenAiHelper.toOpenAiMessage()` ignores `ImageContent.detailLevel()` when building the Azure OpenAI request. It constructs `ChatMessageImageUrl` from the URL only and never calls `setDetail(...)`, so the user-specified image detail (LOW/HIGH/AUTO) is dropped and Azure always applies its server default.

This is inconsistent with:
- The sibling `langchain4j-open-ai` module already forwards detail via `OpenAiUtils.toDetail(content.detailLevel())`.
- `AzureOpenAiTokenCountEstimator.estimateImageTokens()` (PR #5388) is detail-aware, so token estimation and the actual request disagree.

**To Reproduce**

Send a `UserMessage` containing `ImageContent.from(url, DetailLevel.HIGH)` through the Azure OpenAI chat model. The serialized request contains no `detail` field; HIGH is silently lost.

**Expected behavior**

The detail level is forwarded to Azure (`"detail":"high"` / `"low"` / `"auto"`), matching the OpenAI module and the token estimator.

- LangChain4j version: 1.x (main)
- LLM(s) used: Azure OpenAI
- Java version: N/A

**Additional context**

Azure SDK `ChatMessageImageDetailLevel` only supports AUTO/LOW/HIGH; MEDIUM/ULTRA_HIGH have no mapping.

thswlsqls:
## Issue

Closes #5508

## Change
`InternalAzureOpenAiHelper.toOpenAiMessage()` did not forward `ImageContent.detailLevel()` to Azure, so the user-specified image detail was dropped and Azure used its server default.

This PR maps the detail level onto `ChatMessageImageUrl.setDetail(...)`, mirroring `OpenAiUtils.toDetail()` in the sibling `langchain4j-open-ai` module. A private `toDetail(ImageContent.DetailLevel)` maps LOW/HIGH/AUTO to `ChatMessageImageDetailLevel`; MEDIUM/ULTRA_HIGH throw `UnsupportedFeatureException` because the Azure SDK `ChatMessageImageDetailLevel` only supports AUTO/LOW/HIGH.

This aligns the request with the detail-aware `AzureOpenAiTokenCountEstimator` (PR #5388), removing the estimate/send mismatch.

Behaviour change: `ImageContent`'s default detail level is `LOW` (non-null), so callers that never set a detail now send `low` instead of relying on Azure's default. This matches the OpenAI module and the estimator above.

Note: the spotless ratchet (`ratchetFrom=origin/main`) reformatted a few pre-existing lines in the two touched files; no unrelated logic was changed.

Tests: positive LOW/HIGH assertions on the serialized JSON (`"detail":"low"` / `"high"`), and an `UnsupportedFeatureException` assertion for ULTRA_HIGH. Verified fail-then-pass against the unfixed code.

## General checklist

### DIFF
--- a/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
+++ b/langchain4j-azure-open-ai/src/main/java/dev/langchain4j/model/azure/InternalAzureOpenAiHelper.java
@@ -30,6 +30,7 @@ import com.azure.ai.openai.models.ChatCompletionsToolDefinition;
 import com.azure.ai.openai.models.ChatCompletionsToolSelection;
 import com.azure.ai.openai.models.ChatCompletionsToolSelectionPreset;
 import com.azure.ai.openai.models.ChatMessageImageContentItem;
+import com.azure.ai.openai.models.ChatMessageImageDetailLevel;
 import com.azure.ai.openai.models.ChatMessageImageUrl;
 import com.azure.ai.openai.models.ChatMessageTextContentItem;
 import com.azure.ai.openai.models.ChatRequestAssistantMessage;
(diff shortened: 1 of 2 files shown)


dliubarskyi:
@thswlsqls thanks! But I would avoid introducing breaking changes. Also, azure module is outdated, so users should migrate to open-ai-official one

PR #5511, opened 2026-06-21
fix: wmv MIME type mapping key in CustomMimeTypesFileTypeDetector

### LINKED ISSUE #5510: [BUG] Wrong key for wmv MIME mapping in CustomMimeTypesFileTypeDetector
**Describe the bug**
In `CustomMimeTypesFileTypeDetector`, the default video mapping for WMV uses a typo as its key: `defaultMappings.put("mmv", "video/wmv")`. Every other video entry keys on the file extension (`mp4`→`video/mp4`, `mov`→`video/mov`, `webm`→`video/webm`, …), and the value `video/wmv` shows the key was meant to be `"wmv"`. Because of this, `.wmv` files never match the intended mapping, and the `.mmv` key is unreachable (no such extension exists). For a real `.wmv` file the detector falls through to the JDK, returning `video/x-ms-wmv` instead of the intended `video/wmv` used by the Vertex AI Gemini multimodal video requirements.

**Log and Stack trace**
N/A (no exception; wrong/missing mapping).

**To Reproduce**
```java
new CustomMimeTypesFileTypeDetector().probeContentType("video.wmv");
// returns "video/x-ms-wmv" (JDK fallback), expected "video/wmv"
```

**Expected behavior**
`.wmv` resolves to `video/wmv` from the default mappings.

- LLM(s) used: N/A
- Java version: 17

**Additional context**
One-line key fix; value unchanged. Affects multimodal video MIME detection for Gemini/Vertex providers.

thswlsqls:
## Issue
Closes #5510

## Change
`CustomMimeTypesFileTypeDetector`'s default video mappings had a typo in the WMV entry key: `defaultMappings.put("mmv", "video/wmv")`. All sibling video entries key on the file extension (`mp4`, `mov`, `avi`, `webm`, …), and the value `video/wmv` makes the intended key `"wmv"` obvious. With the typo, real `.wmv` files never matched the default mapping and fell through to the JDK (`Files.probeContentType` → `video/x-ms-wmv`), while the `.mmv` key was dead (no such extension). This affects multimodal video MIME detection for the Gemini/Vertex AI providers.

This PR changes the key `"mmv"` → `"wmv"`, keeping the value `"video/wmv"` unchanged. One production line; no behaviour change for any other extension, no API change, no new dependency. Spotless (palantir) collapses the manual column alignment only on the changed line, as expected for `ratchetFrom origin/main`.

## Tests
Added a deterministic unit test mirroring the existing default-mapping String assertions:
`assertThat(new CustomMimeTypesFileTypeDetector().probeContentType("video.wmv")).isEqualTo("video/wmv");`
This is a positive fail-then-pass assertion: it fails on current `main` (returns the JDK fallback `video/x-ms-wmv`) and passes after the fix. A negative case is not meaningful for a mapping-key correction, so none was added.

## General checklist
- [ ] The tests cover both positive and negative cases — positive-only; negative case not meaningful for a mapping-key fix
- [X] I have manually run all the unit and integration tests in the core and main modules, and they are all green
- [ ] I have added/updated the documentation — not applicable (internal mapping fix)
- [ ] I have added an example in the examples repo — not applicable
- [ ] I have added/updated Spring Boot starter(s) — not applicable

### DIFF
--- a/langchain4j-core/src/main/java/dev/langchain4j/internal/CustomMimeTypesFileTypeDetector.java
+++ b/langchain4j-core/src/main/java/dev/langchain4j/internal/CustomMimeTypesFileTypeDetector.java
@@ -89,7 +89,7 @@ public class CustomMimeTypesFileTypeDetector extends FileTypeDetector {
         defaultMappings.put("avi",    "video/avi");
         defaultMappings.put("flv",    "video/x-flv");
         defaultMappings.put("webm",   "video/webm");
-        defaultMappings.put("mmv",    "video/wmv");
+        defaultMappings.put("wmv", "video/wmv");
         defaultMappings.put("3gpp",   "video/3gpp");
     }
 
--- a/langchain4j-core/src/test/java/dev/langchain4j/internal/CustomMimeTypesFileTypeDetectorTest.java
+++ b/langchain4j-core/src/test/java/dev/langchain4j/internal/CustomMimeTypesFileTypeDetectorTest.java
@@ -42,6 +42,18 @@ class CustomMimeTypesFileTypeDetectorTest {
     }
 
     @Test
+    void should_return_a_mime_type_for_wmv_from_default_mapping() {
+        // given
+        CustomMimeTypesFileTypeDetector detector = new CustomMimeTypesFileTypeDetector();
+
+        // when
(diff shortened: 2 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thank you!

PR #5517, opened 2026-06-22
Preserve response whitespace when parsing GPULlama3 thinking content

### LINKED ISSUE #5515: GPULlama3ResponseParser collapses response whitespace when a thinking block is present
**Describe the bug**

`GPULlama3ResponseParser.parseResponse(String)` collapses the actual response body whenever the raw response contains a `<think>...</think>` block. After stripping the thinking block, it runs `actualResponse.replaceAll("\\s+", " ").trim()`, which replaces every run of newlines, indentation, and spaces with a single space. Line breaks, code blocks, and list formatting in the answer are destroyed.

Responses without a `<think>` block are returned unchanged, and the streaming parser forwards characters verbatim. So an identical answer keeps its formatting in those paths but is flattened here, purely because thinking content is present.

**To Reproduce**

```java
ParsedResponse r = GPULlama3ResponseParser.parseResponse(
        "<think>reasoning</think>line1\nline2\n\n  indented");
// r.getActualResponse() == "line1 line2 indented"  (formatting lost)
```

**Expected behavior**

The response body should keep its internal whitespace, like the no-think and streaming paths. Expected: `"line1\nline2\n\n  indented"`.

- LangChain4j version: 1.17.0-beta27-SNAPSHOT
- LLM(s) used: GPULlama3.java
- Java version: 25

**Additional context**

The preceding `.trim()` already removes leading/trailing whitespace, so the `replaceAll` line is unnecessary.

thswlsqls:
## Issue
Closes #5515

## Change

`GPULlama3ResponseParser.parseResponse` flattened the response body whenever a `<think>...</think>` block was present. After removing the thinking block it ran `actualResponse.replaceAll("\\s+", " ").trim()`, collapsing newlines, indentation, and spaces into single spaces. The no-think path and the streaming path both keep formatting, so the same answer came out differently depending on whether thinking content was present.

This removes that `replaceAll` line. The preceding `(beforeThink + afterThink).trim()` already strips leading and trailing whitespace, and the thinking block sits at the start of the response (per the class Javadoc format), so no double-space is introduced at the join point.

Behaviour change: responses containing a `<think>` block now preserve internal whitespace (the bug fix).

Added `GPULlama3ResponseParserTest` (pure unit test, no GPU required) covering: whitespace preservation after a thinking block, leading/trailing trim, the unchanged no-think path, thinking-tag extraction, and the null-input guard. The module's existing `*IT` tests need a GPU/TornadoVM runtime and were not run locally.

## General checklist

### DIFF
--- a/langchain4j-gpu-llama3/src/main/java/dev/langchain4j/model/gpullama3/GPULlama3ResponseParser.java
+++ b/langchain4j-gpu-llama3/src/main/java/dev/langchain4j/model/gpullama3/GPULlama3ResponseParser.java
@@ -95,9 +95,6 @@ public class GPULlama3ResponseParser {
             String beforeThink = rawResponse.substring(0, thinkStart);
             String afterThink = rawResponse.substring(thinkEnd + 8); // Skip </think>
             actualResponse = (beforeThink + afterThink).trim();
-
-            // Clean up any extra whitespace
-            actualResponse = actualResponse.replaceAll("\\s+", " ").trim();
         }
 
         return new ParsedResponse(thinking, actualResponse);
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thanks!

PR #5522, opened 2026-06-22
fix: Escape SQL LIKE single-character wildcard in Hibernate metadata contains filter

### LINKED ISSUE #5521: [BUG] HibernateEmbeddingStore ContainsString filter mistranslates SQL LIKE wildcards
**Describe the bug**

`HibernateEmbeddingStore` translates a `ContainsString` metadata filter into a SQL `LIKE` predicate with `\` as the escape character. The escaping is wrong: it escapes `?`, which is not a `LIKE` wildcard, and it fails to escape `_`, the single-character `LIKE` wildcard.

So a filter whose value holds an underscore (e.g. `user_id`) treats `_` as a wildcard, returning false positives. The core contract `ContainsString.test()` is `String.contains()`, a plain substring match the translation must preserve but breaks.

**To Reproduce**

Apply a `ContainsString` filter whose value contains `_`, e.g. `metadataKey("k").containsString("a_b")`. Rows whose value is `axb` also match, although `axb` does not contain the literal `a_b`.

**Expected behavior**

The `LIKE` pattern escapes `_` and `%` (the only `LIKE` wildcards) and leaves `?` as a literal, matching the same rows as `String.contains()`.

- LLM(s) used: N/A
- Java version: 17

**Additional context**

One-line fix in `HibernateEmbeddingStore.mapContains`: replace `.replace("?", "\\?")` with `.replace("_", "\\_")`.

thswlsqls:
## Issue

Closes #5521

## Change

`HibernateEmbeddingStore.mapContains` builds a SQL `LIKE` pattern from a `ContainsString` filter value, using `\` as the escape character.

The escaping was wrong: it escaped `?`, which is not a `LIKE` wildcard, and did not escape `_`, the single-character `LIKE` wildcard. So a value like `a_b` matched `axb` — a false positive that violates the core `ContainsString.test()` contract (`String.contains()`, a plain substring match).

Fix: replace `.replace("?", "\\?")` with `.replace("_", "\\_")`. Dropping the `?` escape is safe because `?` is a `LIKE` literal; escaping of `\` and `%` is unchanged.

Test: `HibernateEmbeddingStoreContainsFilterTest` (new) asserts the generated `LIKE` pattern for inputs containing `_`, `%`, `?`, and `\`. It needs no database: it creates the store via Mockito to bypass the constructor, injects a mapped attribute and invokes the private `mapContains` by reflection so the real method runs against mocked Hibernate criteria collaborators, then captures the pattern with an `ArgumentCaptor`. Removing the fix makes the `_` and `?` cases fail, confirming the test is not tautological.

## General checklist

### DIFF
--- a/langchain4j-hibernate/src/main/java/dev/langchain4j/store/embedding/hibernate/HibernateEmbeddingStore.java
+++ b/langchain4j-hibernate/src/main/java/dev/langchain4j/store/embedding/hibernate/HibernateEmbeddingStore.java
@@ -834,7 +834,7 @@ public class HibernateEmbeddingStore<E> implements EmbeddingStore<TextSegment> {
                                         .comparisonValue()
                                         .replace("\\", "\\\\")
                                         .replace("%", "\\%")
-                                        .replace("?", "\\?")
+                                        .replace("_", "\\_")
                                 + "%",
                         '\\'));
     }
--- a/langchain4j-hibernate/src/test/java/dev/langchain4j/store/embedding/hibernate/HibernateEmbeddingStoreContainsFilterTest.java
+++ b/langchain4j-hibernate/src/test/java/dev/langchain4j/store/embedding/hibernate/HibernateEmbeddingStoreContainsFilterTest.java
@@ -1,0 +1,85 @@
+package dev.langchain4j.store.embedding.hibernate;
+
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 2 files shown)


dliubarskyi:
Hi @beikov could you please review?

beikov (approved):
You're absolutely right. I totally forgot that it was `_`, that represented a single wildcard character and not `?`, like in Regex.

PR #5527, opened 2026-06-22
Non-blocking / reactive support

### LINKED ISSUE #3391: Introduce Reactive/Non-Blocking APIs for Core LangChain4J
### **Is your feature request related to a problem? Please describe.**

Yes. Currently, the core LangChain4J library's APIs are predominantly **synchronous and blocking**. For instance, interacting with chat models, embeddings, or using features like chat memory (e.g., the ChatMemoryStore) all block the calling thread until the operation is complete.

This presents a significant challenge when integrating LangChain4J into modern **reactive Java applications** (e.g., using Spring WebFlux, Vert.x, or Project Reactor). To avoid blocking the event loop, developers are forced to wrap every LangChain4J call in boilerplate code, such as `Mono.fromCallable(...).subscribeOn(Schedulers.boundedElastic())`. This is not only cumbersome but also less efficient than a native non-blocking implementation.

Interestingly, the `quarkus-langchain4j` extension provides an excellent reactive, non-blocking API out of the box. This starkly contrasts with the core library, making the pure Java version feel less suited for modern, high-concurrency environments. The superb design of the Quarkus extension highlights what's missing in the core framework.

-----

### **Describe the solution you'd like 🚀**

We propose enhancing the core LangChain4J library to natively support **non-blocking, asynchronous operations**. This would involve introducing reactive counterparts to the existing blocking methods.

A great approach would be to provide methods that return a `CompletableFuture<T>`. This is a standard Java construct and provides a solid foundation for asynchronous programming that can be easily bridged to other reactive libraries.

For example, the existing `ChatMemoryStore` interface could be updated:

```java

public interface ChatMemoryStore {

    // Existing blocking method
    void updateMessages(Object memoryId, List<ChatMessage> messages);

    // Proposed non-blocking alternative
    CompletableFuture<Void> updateMessages(Object memoryId, List<ChatMessage> messages);
 
   //or even using reactive streams Publisher
   Publisher<Void> updateMessages(Object memoryId, List<ChatMessage> messages)

}
```

This would allow seamless integration into any reactive project without requiring developers to manage thread pools manually. The implementation for these async methods could leverage async HTTP clients (like the one already used by the Quarkus extension) to provide true non-blocking I/O.

-----

### **Describe alternatives you've considered 🤔**

1.  **Wrapping Blocking Calls:** The current workaround is to manually wrap every blocking call in a reactive publisher and schedule it on a dedicated thread pool for blocking tasks (e.g., `Schedulers.boundedElastic()`). This adds significant boilerplate and is prone to errors if not done correctly for every single call. It's a patch, not a solution.

2.  **Migrating to Quarkus:** While the `quarkus-langchain4j` extension is excellent, moving an entire project to a new framework like Quarkus is a major undertaking and often not a feasible option for existing applications built on other stacks like Spring Boot.

-----

### **Additional context**

Adopting a reactive-first or reactive-supported approach in the core library would massively broaden the appeal and utility of LangChain4J. It would position it as a first-class citizen in the modern Java ecosystem, which is increasingly moving towards non-blocking paradigms for better resource utilization and scalability. The Quarkus extension has already proven this is a viable and highly effective design pattern. Bringing that same level of quality to the core library would be a game-changer.

dliubarskyi:
Non-blocking / reactive support

## Issue

Closes https://github.com/langchain4j/langchain4j/issues/3391

## Goal

Today an AI Service method blocks the calling thread for the entire interaction — the model HTTP call, tool
execution, memory I/O and guardrails. This PR adds genuinely non-blocking ways to call an AI Service, so a single
thread can drive many concurrent interactions: what reactive stacks (Quarkus/Mutiny, Vert.x, WebFlux) need, and
what makes high-concurrency services cheap.

The work is **additive**. Existing synchronous and `TokenStream` APIs are unchanged, and non-blocking is threaded
through *every* layer — a blocking gap at any level re-blocks the whole call.

Everything added here is annotated `@Experimental` and `@since 1.20.0`.

## What's delivered

You pick the mode by the method's **return type**; the interface definition is otherwise unchanged:

| Mode | Return type | Nature |
|---|---|---|
| Synchronous (existing) | `String`, POJO, `Result<T>`, … | blocking |
| Streaming callback (existing) | `TokenStream` | handler-based |
| **Async single-response (new)** | `CompletableFuture<T>`, `CompletionStage<T>` | non-blocking |
| **Reactive streaming (new)** | `Flow.Publisher<AiServiceStreamingEvent>`, `Flow.Publisher<String>` | non-blocking, cold |

Building blocks, each non-blocking end to end:

- **Reactive event vocabulary** — `AiServiceStreamingEvent`, an open (extensible) event set for streaming AI
  Services, each event carrying an `InvocationContext`.
- **Reactive tool loop** — a cold publisher consuming the model's reactive stream round by round.
- **Non-blocking tool calling** — sync and `CompletableFuture` tools, sequential and concurrent.
- **Non-blocking chat memory, guardrails and RAG** — async SPI across `ChatMemory`/`ChatMemoryStore`,
  `Guardrail`, and the whole retrieval graph down to `EmbeddingModel`, `EmbeddingStore`, `ScoringModel` and
  `WebSearchEngine`.
- **New module `langchain4j-reactive-streaming`** — the shared publisher plumbing, sitting below the providers and
  above core/http-client, so `langchain4j-core` keeps speaking only JDK `Flow`.

User documentation: **[Non-blocking and Reactive](https://docs.langchain4j.dev/tutorials/non-blocking)**, plus
updates to the AI Services, response-streaming, chat models, chat memory, guardrails, RAG, tools, embedding
stores, MCP, observability, Kotlin and HTTP-client pages.

### 1. Kotlin `chatAsync` extension

`ChatModel` now declares a Java member `chatAsync(ChatRequest): CompletableFuture<ChatResponse>`. In Kotlin a
same-signature member wins over an extension, so a bare single-argument call that previously resolved to the
`suspend` extension (returning `ChatResponse`) now resolves to the member (returning `CompletableFuture`):

```kotlin
// Before — resolved to the suspend extension, returned ChatResponse:
val response: ChatResponse = model.chatAsync(request)

// After — the bare single-arg call resolves to the Java member:
val future: CompletableFuture<ChatResponse> = model.chatAsync(request)

// Migration — await the future…
val response: ChatResponse = model.chatAsync(request).await()
// …or supply a coroutineContext, which still selects the suspend extension:
val response: ChatResponse = model.chatAsync(request, Dispatchers.IO)
```

### 2. `UnsupportedFeatureException` is now a `NonRetriableException`

```java
UnsupportedFeatureException  extends NonRetriableException        // was LangChain4jException
AsyncNotSupportedException   extends UnsupportedFeatureException  // new type
```

`UnsupportedFeatureException` is still a `LangChain4jException`, so every existing `catch` clause keeps compiling
and catching — only an intermediate supertype is inserted. **Behaviour change:** `RetryUtils` no longer retries it.
That is the intended fix (a provider does not begin to support a feature on the third attempt), but if you relied
on those retries, catch and retry explicitly. This also lets one `catch (UnsupportedFeatureException)` cover the
new "this component is not asynchronous" failures.

## Defaults that differ from the synchronous modes

Deliberate, and documented in the tutorial — worth a careful look in review:

| | Synchronous / `TokenStream` | `CompletableFuture` / `Flow.Publisher` |
|---|---|---|
| Multiple tool calls | executed sequentially | executed **concurrently** |
| Tool **execution** error | sent back to the LLM | **fails the invocation** |
| Tool **argument-parse** error | fails the invocation | **sent back to the LLM** |
| `@Moderate` | supported | rejected at AI Service creation |

Rationale: an execution failure sent to the LLM hides a bug in your tool and invites the model to invent an answer
around it; a malformed argument string is something the model produced and can usually fix when told. Both remain
configurable, and an explicitly configured handler is used by every mode.

On the reactive path a tool starts as soon as its `CompleteToolCallEvent` is emitted, so a round's tool-execution
events may precede that round's `IntermediateResponseEvent` — the one ordering difference from `TokenStream`.

## Threading and failure model

- Genuinely async model I/O; async tools are composed, never waited on.
- Blocking tools are offloaded — to virtual threads on Java 21+, and to an unbounded platform-thread pool on
  Java 17–20, since LangChain4j targets 17. Override via the `ExecutorProvider` SPI.
- **Thread-local context does not follow the async hops** (MDC, tracing spans, security context). Return a
  context-propagating executor from `ExecutorProvider` to make it follow the work. `InvocationContext` is
  unaffected — it is passed explicitly.
- Model listeners are invoked on the transport's I/O threads and **must not block**.
- A component that has not implemented its async counterpart **fails loudly** through the return type with
  `AsyncNotSupportedException`, rather than silently blocking a thread.
- Cancelling the future or the subscription stops further rounds and best-effort aborts the in-flight I/O. A tool
  that has already started is not interrupted; its result is discarded and compensation is attempted.

## Testing

- **7 Reactive Streams TCK classes** (38 tests each) over the HTTP-client publisher, the model publishers
  (OpenAI Chat Completions and Responses, Anthropic, Bedrock) and both AI Service publishers.
- **3 BlockHound integration tests** that *prove* nothing blocks the policed I/O threads, rather than asserting it.
- Deterministic mock-HTTP coverage of retry classification, cancellation and wire-format parsing, so the async
  paths are testable without API keys.

## Provider support

| | `chatAsync` | reactive `chat` |
|---|---|---|
| OpenAI (Chat Completions) | ✅ | ✅ |
| OpenAI (Responses) | ✅ | ✅ |
| Anthropic | ✅ | ✅ |
| Bedrock | ✅ | ✅ |

Also: `embedAsync` on OpenAI, `scoreAsync` on Cohere, `searchAsync` on Tavily. Every other provider keeps working
and fails loudly on the non-blocking path until it opts in.

## Not included yet

- **No Mutiny/Reactor adapter ships.** Returning `Uni`/`Multi` or `Mono`/`Flux` from a non-blocking AI Service
  method does not work yet: the `CompletableFutureAdapter` and `PublisherAdapter` seams exist and are exercised by
  tests, but they are `@Internal` and no implementation or `ServiceLoader` registration is shipped. Use
  `CompletableFuture`/`Flow.Publisher` and adapt at the call site. (`Flux<String>` via `langchain4j-reactor` is a
  separate, pre-existing `TokenStream` adapter and is unrelated to this path.)

## General checklist

- [ ] There are no breaking changes (API, behaviour) — see **Breaking Changes** above
- [x] I have manually run all the unit and integration tests in the core and main modules, and they are all green
- [x] I have added/updated the documentation


jponge:
Hi @dliubarskyi, I've finally had a chance to review this draft 😄 

First off I am a stranger to this code base and the PR is big, so apologies if I am making any mistake.

Overall the patterns I've seen seem correct. It makes sense that streams are unbounded due to the nature of the source that emits tokens. It also looks like the reactive streams protocol is well handled / both due to `mutiny-zero` enforcing most of the rules, but also due to the TCK tests you added for custom publishers.

Some observations.

- In `StreamingChatModelHelper` the `complete()` method transforms a `null` `response` to an empty `ChatResponse`. Perhaps you want to fail the future in such cases (e.g., `completeExceptionally`) to forward an error rather than mask it.
- There are opportunities for using `mutiny-zero` operators. We don't offer many like a full reactive programming library would do, but the `mutiny.zero.operators` package has a few: `Transform`, `Select`, `Recover`, `Retry`, `Concatenate`. They could save you a few nested `Tube` constructs.

Examples where operators could help (I haven't checked in depth, but you get the idea).

`AiServiceStreamingEventPublisher#toTextPublisher()`:

```java
public static Flow.Publisher<String> toTextPublisher(
        Flow.Publisher<AiServiceStreamingEvent> events, boolean hasOutputGuardrails, int bufferSize) {
    if (hasOutputGuardrails) {
        return new Transform<>(
                new Select<>(events, e -> e instanceof FinalResponseEvent fe
                        && fe.chatResponse().aiMessage().text() != null),
                e -> ((FinalResponseEvent) e).chatResponse().aiMessage().text());
    } else {
        return new Transform<>(
                new Select<>(events, e -> e instanceof PartialResponseEvent),
                e -> ((PartialResponseEvent) e).partialResponse().text());
    }
}
```

`StreamingChatModel#chat(str)`:

```java
default Publisher<String> chat(String userMessage) {
    ChatRequest chatRequest =
            ChatRequest.builder().messages(UserMessage.from(userMessage)).build();
    return new Transform<>(
            new Select<>(chat(chatRequest), e -> e instanceof PartialResponse),
            e -> ((PartialResponse) e).text());
}
```

Again what you did seems correct, but perhaps you could simplify in such places since the logic you have is essentially that of a `map` / `filter`.

Hope it helps!

dliubarskyi:
Hi @jponge, thanks a lot for the review!

I have fixed points 1 (`complete()`) and 2 (`toTextPublisher()` -> `Select` + `Transform`), thanks for the heads up!

Regarding your `StreamingChatModel#chat(str)` suggestion, it looks good, but unfortunately this code lives in `langchain4j-core` which does not have a dependency on `mutiny-zero`, so I cannot really simplify it here...

Thanks again for the review!

PR #5542, opened 2026-06-22
fix: make metadata key collisions and skipped documents visible during document loading

### LINKED ISSUE #5541: [BUG] Kotlin parseAsync throws on source/document metadata key collision unlike DocumentLoader.load
**Describe the bug**

The Kotlin extension `DocumentParser.parseAsync(source, context)` diverges from the synchronous `DocumentLoader.load()` when forwarding source metadata to the parsed `Document`.

`DocumentLoader.load()` does `document.metadata().putAll(source.metadata().toMap())`: source overwrites document on a shared key, never throws.

`parseAsync` instead merges with `Metadata.merge()`, which throws `IllegalArgumentException` on ANY shared key. So for the same `(DocumentParser, DocumentSource)` input, the sync path succeeds but the async path throws. The public API accepts arbitrary `DocumentParser` implementations, so a parser emitting a key the source also sets (e.g. `file_name` from `FileSystemSource`) is a supported input. `FileSystemDocumentLoaderExtensions.loadDocuments()` calls `parseAsync`, so one such file breaks a whole directory load.

**Log and Stack trace**

`java.lang.IllegalArgumentException: Metadata keys are not unique. Common keys: [file_name]`

**To Reproduce**

Call `parseAsync` with a parser whose `Document` metadata contains `file_name` while the source also sets `file_name`.

**Expected behavior**

`parseAsync` forwards source metadata like `DocumentLoader.load()`: source overwrites on collision, never throws.

- LLM(s) used: N/A
- Java version: 17

**Additional context**

Module: `langchain4j-kotlin`. Extensions added in PR #2640.

thswlsqls:
## Issue

  Related to #5541. **This PR intentionally does not close #5541** — the Kotlin `parseAsync` divergence stays open and is addressed in 2.0 (see "Follow-up" below).

  ## Change

  `DocumentLoader.load()` forwards the `DocumentSource` metadata onto the parsed `Document` with `putAll()`. When the parser and the source define the same key (e.g. a custom `DocumentParser` that sets
  `file_name`, or `ApachePdfBoxDocumentParser` / `ApacheTikaDocumentParser` with `includeMetadata = true`, which copy keys straight out of the file), the document value is dropped and nothing is reported.

  This PR keeps the existing "source wins" resolution — no behaviour change, no breaking change — but stops it being silent, and stops batch loaders losing documents invisibly:

  **`langchain4j-core`**
  - `DocumentLoader.load()` logs a warning naming the conflicting key **and both values**, and points at the workaround:
    ```
    Metadata key "file_name" is set both by the document ("report.pdf") and by the source ("2024-report.pdf").
    Keeping the source value and discarding the document value. To control this, remove the key in your
    DocumentParser before returning the Document.
    ```
  - Javadoc on `DocumentLoader.load()` now states the collision behaviour explicitly instead of just "forwards the source Metadata".
  - `Metadata.merge()` reports the conflicting **values**, not only the key names:
    `Metadata keys are not unique. Common keys and their values: {key2=("value2", "value3")}`

  **Batch loaders** (`FileSystemDocumentLoader`, `ClassPathDocumentLoader`, `AmazonS3DocumentLoader`, `AzureBlobStorageDocumentLoader`, `GoogleCloudStorageDocumentLoader`, `TencentCosDocumentLoader`,
  `GitHubDocumentLoader`)                                                                   

  All seven swallow per-document exceptions and continue, so a caller can silently receive a shorter `List<Document>` than expected. Each now logs a summary when anything was lost:
  ```
  Loaded 9997 of 10000 documents from '/docs'. Skipped 3 that failed to load and 0 that were blank.
  ```
  `FileSystemDocumentLoader` and `ClassPathDocumentLoader` additionally pass the throwable to the logger instead of only its message, so failures get a stack trace like the cloud loaders already did.

  ## Why not throw on collisions

  An earlier revision of this PR made both `DocumentLoader.load()` and the Kotlin `parseAsync` throw on collisions via `Metadata.merge()`. That turns out to be the wrong fix for 1.x: every batch loader treats an
  exception as "this file is broken, skip it". A perfectly readable document would then be dropped from the result list purely because of a metadata key name — replacing a silently discarded *value* with a
  silently discarded *document*. The trigger is user data (keys authored by whoever produced the PDF), not user code, so it is not something the caller can reliably avoid.

  A workaround exists today and needs no new API — decorate the parser to drop the key:

  ```java
  DocumentParser stripped = inputStream -> {
      Document document = myParser.parse(inputStream);
      document.metadata().remove(Document.FILE_NAME);
      return document;
  };
  ```

  ## Follow-up (2.0)

  Collisions should not be resolved silently, but the fix needs two pieces that only fit in a major release, and they have to land together:

  - **Collisions fail by default**, with an explicit resolution strategy to opt out (`sourceWins()`, `documentWins()`, or a custom resolver), so the library never guesses which value was meant.
  - **Bulk loading gets a real failure policy**: fail-fast by default, explicit opt-in to tolerance, and failures returned as **data** rather than only logged — so "throw" can never mean "document silently
  vanishes".

  Until then, the Kotlin `parseAsync` extension keeps throwing where the Java path warns. That divergence is deliberate, and is why #5541 stays open — better one behaviour change in 2.0 than two in a row.

  ## Notes for reviewers

  - No API changes and no new dependencies; `revapi` is unaffected.
  - The only `langchain4j-kotlin` change is a test assertion updated for the new `Metadata.merge()` message. The Kotlin `parseAsync` main source is back to `main`.
  - `Metadata.merge()`'s exception **message** changed. The exception type and the conditions that trigger it are unchanged; only assertions on the exact message text are affected.
  - `GoogleCloudStorageDocumentLoader` shows a larger diff than its change warrants: its imports were already unformatted on `main` and only hit the spotless ratchet now that the file is touched.

  ## General checklist
  - [X] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and


dliubarskyi:
Hi @kpavlov, could you please review?

kpavlov:
Hey, @dliubarskyi , yes, in my evening I will have a look

kpavlov:
To be honest, I'm not sure silently merging metadata is the right approach. It's surprising behavior to me. Which metadata should win? Should we update [DocumentLoader.load()](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentLoader.java#L26) to throw an exception on collisions instead? @dliubarskyi what do you think?

kpavlov (commented):
Technically, LGTM
But I am not sure silently merging is the right approach, because document metadata gets silently overwritten.

thswlsqls:
This only aligns the Kotlin `parseAsync` with core `DocumentLoader.load()` (it already does `putAll(source.metadata())`), so source wins on collisions — that's existing behavior, not new here. Whether collisions should throw instead is a core-wide decision; happy to open a separate issue if you and @dliubarskyi want to pursue it.

dliubarskyi:
> I'm not sure silently merging metadata is the right approach. It's surprising behavior to me. Which metadata should win? Should we update [DocumentLoader.load()](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/main/java/dev/langchain4j/data/document/DocumentLoader.java#L26) to throw an exception on collisions instead?

Agreed, this is a bug, we need to throw in case of collisions. @thswlsqls could you please update both Java and Kotlin versions to throw an exception in case of collisions? Perhaps it is best to use `Metadata.merge()` method that already checks for collisions

thswlsqls:
Done — both `DocumentLoader.load()` and the Kotlin `parseAsync` now forward source metadata via `Metadata.merge()`, which throws on key collisions instead of silently overwriting. Tests updated. Note this changes `DocumentLoader.load()` for all loaders, not just Kotlin.

thswlsqls:
Hi @dliubarskyi, thanks again for the earlier review. A gentle follow-up — both `DocumentLoader.load()` and the Kotlin `parseAsync` now throw on metadata key collisions via `Metadata.merge()` as you suggested, with tests in both modules, and the branch is up to date with `main`. No rush at all — and if you'd prefer not to take this behavior change, I'm completely fine with closing the PR.

dliubarskyi:
Hi @thswlsqls — thank you for your patience here, and for keeping the branch current. You did exactly what @kpavlov and I asked for on 24 June, and the implementation was correct. Now I'm changing
  direction, and I owe you the reasoning rather than just a revert request.

  ## Why we're not throwing on collisions in 1.x

  Looking harder at the impact, throwing behaves badly in 1.x: **every batch loader treats an exception as "this file is broken, skip it and move on."**

  - `FileSystemDocumentLoader.loadDocuments()` — `catch (Exception)` → `log.warn`, document dropped
  - `ClassPathDocumentLoader`, `AmazonS3DocumentLoader`, `AzureBlobStorageDocumentLoader`, `TencentCosDocumentLoader`, `GoogleCloudStorageDocumentLoader` — same
  - `GitHubDocumentLoader` — `catch (RuntimeException)`, same

  That assumption was fine when the only exceptions were I/O and parse errors. It stops being fine once a **perfectly readable document** can throw purely because of a metadata key name. A change motivated by
  "metadata should not be dropped silently" would instead drop the **whole document** silently, leaving the user with a `List<Document>` that is quietly shorter than they expect.

  The trigger is also user *data*, not user *code*. `ApachePdfBoxDocumentParser(includeMetadata = true)` copies every key from the PDF Document Information Dictionary verbatim and unprefixed;
  `ApacheTikaDocumentParser(includeMetadata = true)` copies every Tika metadata name. Those key spaces belong to whoever authored the file. And the likeliest trigger is the one you originally reported — a custom
  `DocumentParser` that sets `file_name`, which would then fail on every file in the directory.

  So your diagnosis was right, and the silent overwrite really is a bug. What we got wrong is that 1.x has nowhere safe for the exception to land.

  ## What this PR does now

  Keep source-wins (no behaviour change, no breaking change), but stop it being silent — and stop the batch loaders losing documents invisibly:

  - `DocumentLoader.load()` logs a warning naming the key **and both values**, and points at the workaround
  - `Metadata.merge()` reports the conflicting values, not just the key names
  - `FileSystemDocumentLoader` / `ClassPathDocumentLoader` pass the throwable to the logger, so failures get stack traces like the cloud loaders already did
  - All seven batch loaders log a summary whenever documents were lost:
    `Loaded 9997 of 10000 documents from '/docs'. Skipped 3 that failed to load and 0 that were blank.`

  Worth noting there is already a fix available to users today needing no new API — decorate the parser to drop the key:

```java
  DocumentParser stripped = inputStream -> {
      Document document = myParser.parse(inputStream);
      document.metadata().remove(Document.FILE_NAME);
      return document;
  };
  ```

  ## The plan for 2.0

  Your instinct — that a collision shouldn't be resolved silently — is the part we're keeping:

  - **Collisions fail by default**, with an explicit resolution strategy to opt out (`sourceWins()`, `documentWins()`, or a custom resolver), so the library never guesses which value you meant
  - Bulk loading gets a real **failure policy**: fail-fast by default, explicit opt-in to tolerance, and failures returned as **data** instead of only logged — so "throw" can never again mean "document silently
  vanishes"

  Those two have to land together, which is why the throw waits for 2.0 instead of going in now.

  ## About the commits on your branch

  I pushed to your branch directly so you could see the concrete shape rather than review a description — sorry for not asking first. I also reverted the `langchain4j-kotlin` `parseAsync` changes back to `main`,
  so this PR is now purely `langchain4j-core` plus the loaders.

  That does mean **#5541 stays open**: `parseAsync` will still throw where the Java path warns. We're deliberately leaving that divergence until 2.0 fixes both paths together, rather than changing the behaviour
  twice in a row.

  ## Are you OK with this?

  Completely understand if this isn't what you signed up for. Three options, your call:

  1. Keep the PR as it now stands — we'll update the title and description, and it merges as a 1.x visibility improvement
  2. You'd rather rework it yourself — revert anything of mine you don't like
  3. You'd rather close it — I'll move the changes to a separate PR and credit you

  Either way, thank you for pushing on the original question. It surfaced a genuine bug in `DocumentLoader.load()` and turned into a much better 2.0 design than where we started.

thswlsqls:
@dliubarskyi Thanks for the write-up, and no apology needed for pushing to the branch. The batch-loader point convinced me — a throw that leaves `loadDocuments()` dropping whole documents is worse than the overwrite it fixes. Option 1 works for me.

Two things for when you update the description. It still says `Closes #5541`, so merging as-is would auto-close the issue you're keeping open for 2.0. And spotless was failing, so I pushed the `mvn spotless:apply` fix — `DocumentLoaderTest`, plus `GoogleCloudStorageDocumentLoader`, whose imports were already unformatted on `main` and only hit the ratchet now that the file is touched.

dliubarskyi (approved):

dliubarskyi:
@thswlsqls thanks!

PR #5547, opened 2026-06-23
Forward enableEnhancedCivicAnswers to the Gemini generation config

### LINKED ISSUE #5546: [BUG] enableEnhancedCivicAnswers is ignored in GoogleAiGeminiChatModel and GoogleAiGeminiStreamingChatModel
**Describe the bug**
`GoogleAiGeminiChatModel` and `GoogleAiGeminiStreamingChatModel` expose a builder option
`enableEnhancedCivicAnswers(Boolean)`, but the value is never sent to the Gemini API. `BaseGeminiChatModel`
stores it in a field, yet `createGenerateContentRequest(...)` does not pass it to the `GeminiGenerationConfig`
it builds, so the request omits `enableEnhancedCivicAnswers` and the setting has no effect.

`GeminiGenerationConfig` already supports the field (it has `enableEnhancedCivicAnswers` with the matching
`@JsonProperty`), so only the population in `createGenerateContentRequest` is missing. The builder method is
documented as "Enables enhanced civic answers", and `enableEnhancedCivicAnswers` is a real `generationConfig`
field (it also exists in the official `google-genai` Java SDK's `GenerateContentConfig`), so the option is
expected to take effect. The sibling `responseLogprobs` is forwarded the same way; only this one was missed.

**To Reproduce**
```java
        .enableEnhancedCivicAnswers(true)   // accepted, but never sent
        .build();
```
The serialized request body contains no enableEnhancedCivicAnswers field; the option is silently dropped.
Both the sync and streaming chat models are affected, since they share createGenerateContentRequest.

Expected behavior
When enableEnhancedCivicAnswers(true) is set, the request's generationConfig includes
"enableEnhancedCivicAnswers": true, consistent with the other generation-config options
(temperature, topK, responseLogprobs, etc.).

**LangChain4j version**
1.x (verified on current main, 1.17.0-SNAPSHOT)

subhashpolisetti:
`GoogleAiGeminiChatModel` and `GoogleAiGeminiStreamingChatModel` accept `enableEnhancedCivicAnswers(Boolean)`
on the builder, but the value was never sent to the API. `BaseGeminiChatModel#createGenerateContentRequest`
builds the `GeminiGenerationConfig` without it, so the option was silently ignored.

`GeminiGenerationConfig` already maps the field (`enableEnhancedCivicAnswers` with the matching
`@JsonProperty`); this just populates it from the existing field, next to `responseLogprobs`. The fix is shared
by the sync and streaming models, since both go through `createGenerateContentRequest`.

The builder option is documented as "Enables enhanced civic answers", and the field is a real Gemini
`generationConfig` parameter (it also exists in the official `google-genai` Java SDK's `GenerateContentConfig`),
so this is a missed wiring rather than an intentionally unsupported option. The sibling `responseLogprobs` is
forwarded the same way.

Closes #5546

### Changes
- `BaseGeminiChatModel`: pass `enableEnhancedCivicAnswers` into the `GeminiGenerationConfig` builder (shared by
  the sync and streaming models).
- `GoogleAiGeminiEnhancedCivicAnswersTest` (new): asserts the flag is forwarded when enabled and defaults to
  `false` otherwise, using the existing mocked-`GeminiService` request-capture pattern.
- `GoogleAiGeminiStreamingChatModelTest`: added enabled and default cases on the streaming path, following that
  test's existing `createGenerateContentRequest` style.
- `GoogleAiGeminiChatModelTest#shouldSendRequestWithAllParameters`: expected config updated to include the
  now-forwarded default, mirroring the existing `responseLogprobs(false)` assertion.

### Testing
`mvn -pl langchain4j-google-ai-gemini test` -> 334 passing, 0 failures. `spotless:check` passes.

### Checklist
- [x] No breaking changes (additive; default behavior unchanged, since `false` was already the effective default).
- [x] Unit tests added (positive: enabled is forwarded; default: false; both sync and streaming).
- [x] Ran the module build and tests locally (green) + spotless.

### DIFF
--- a/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/BaseGeminiChatModel.java
+++ b/langchain4j-google-ai-gemini/src/main/java/dev/langchain4j/model/googleai/BaseGeminiChatModel.java
@@ -157,6 +157,7 @@ class BaseGeminiChatModel {
                         .presencePenalty(parameters.presencePenalty())
                         .frequencyPenalty(parameters.frequencyPenalty())
                         .responseLogprobs(responseLogprobs)
+                        .enableEnhancedCivicAnswers(enableEnhancedCivicAnswers)
                         .logprobs(logprobs)
                         .thinkingConfig(this.thinkingConfig)
                         .mediaResolution(this.mediaResolution)
--- a/langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiChatModelTest.java
+++ b/langchain4j-google-ai-gemini/src/test/java/dev/langchain4j/model/googleai/GoogleAiGeminiChatModelTest.java
@@ -361,6 +361,7 @@ class GoogleAiGeminiChatModelTest {
                             .seed(42)
                             .candidateCount(1)
                             .responseLogprobs(false)
+                            .enableEnhancedCivicAnswers(false)
                             .build());
         }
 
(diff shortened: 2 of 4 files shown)


dliubarskyi (approved):
@subhashpolisetti thanks!

PR #5548, opened 2026-06-23
Fix NPE in OpenAiOfficialChatModel/StreamingChatModel with user-supplied OpenAIClient

### LINKED ISSUE #5545: [BUG] OpenAiOfficialBaseChatModel defaultRequestParameters NPE problem.
https://github.com/langchain4j/langchain4j/blob/release-1.5.0/langchain4j-open-ai-official/src/main/java/dev/langchain4j/model/openaiofficial/OpenAiOfficialBaseChatModel.java

This class when i init using client supported mode, the defaultRequestParameters has skiped. And then, program call model.chat, method of defaultRequestParameters() will throw the NPE.

More things, Line number 101 of OpenAiOfficialStreamingChatModel.class also throw NPE.

PSchmitz-Valckenberg:
## Summary
- Fixes #5545: `defaultRequestParameters()` and `doChat(...)` throw `NullPointerException` when an `OpenAIClient`/`OpenAIClientAsync` is supplied directly via `.openAIClient(...)` / `.openAIClientAsync(...)` ("bring your own client" mode).
- Root cause: `OpenAiOfficialChatModel`/`OpenAiOfficialStreamingChatModel` constructors only called `init(...)` when no client was supplied, so `defaultRequestParameters`, `modelProvider`, `strictTools`, `strictJsonSchema`, `listeners`, `supportedCapabilities` were left `null` in that mode.
- `OpenAiOfficialBaseChatModel.init(...)` now always runs the parameter/provider initialization; only the actual client construction (`setupSyncClient`/`setupAsyncClient`) is skipped when a pre-built client is passed in.

## Test plan
- [x] Added `OpenAiOfficialBringYourOwnClientTest` covering both `OpenAiOfficialChatModel` and `OpenAiOfficialStreamingChatModel` built with a mocked `OpenAIClient`/`OpenAIClientAsync`, asserting `defaultRequestParameters()` and `provider()` are populated instead of throwing.
- [x] Verified the new test fails (build breaks) without the fix and passes with it.
- [x] `mvn -pl langchain4j-open-ai-official test` — all unit tests pass.
- [x] `mvn -pl langchain4j-open-ai-official spotless:check` — clean.


dliubarskyi (approved):
@PSchmitz-Valckenberg thanks!

Issue #3284, opened 2025-07-01
[BUG]  ChatMemoryStore read invoked 3× → duplicate messages saved & sent to LLM, chat history balloons


krdk01:
**Describe the bug** During a single call to my `@AiService` method, `ChatMemoryStore` is invoked in the following pattern:

1. `getMessages()`   ← first read  
2. `getMessages()`   ← second read  
3. `getMessages()`   ← third read  
4. `updateMessages()`  ← **only once**

Because `getMessages()` is executed three times before the single `updateMessages()`, the same user message is included three times in the list that is ultimately written back.  

As a result:

* The chat message is **saved three times** in the database.
* The bloated history is then forwarded to the LLM
  so the model receives duplicate context.
* With ~300 conversations I already have **14 000+ rows**, and the table keeps
  growing rapidly whenever I reuse a session.

This duplication inflates storage, slows history retrieval, and wastes tokens in the LLM prompt.

**Additional context**
In `DefaultAiServices` the following block seems to trigger three store interactions:

```java
if (chatMemory != null) {
    systemMessage.ifPresent(chatMemory::add); // 1st read cycle
    chatMemory.add(userMessage);              // 2nd read cycle
    messages = chatMemory.messages();         // 3rd read
}
```

**Log and Stack trace**
```
getMessages called                         (1)
getMessages called                         (2)
getMessages called                         (3)

/* shortened stack – identical for each call */
at com.chattest.test.ai.llm.chat.memory.PostgreSqlChatMemoryStore.getMessages(PostgreSqlChatMemoryStore.java:41)
at dev.langchain4j.memory.chat.MessageWindowChatMemory.messages(MessageWindowChatMemory.java:80)
at dev.langchain4j.memory.chat.MessageWindowChatMemory.add(MessageWindowChatMemory.java:55)
at dev.langchain4j.service.DefaultAiServices$1.invoke(DefaultAiServices.java:174)
...
```

**To Reproduce**
```java
// 1 ChatMemoryProvider
ChatMemoryProvider provider = id -> MessageWindowChatMemory.builder()
        .id(id)
        .maxMessages(50)
        .chatMemoryStore(new PostgreSqlChatMemoryStore(chatMessageService))
        .build();
aiServiceBuilder.chatMemoryProvider(provider);

// 2 Minimal custom store
public class PostgreSqlChatMemoryStore implements ChatMemoryStore {
    @Override
    public List<ChatMessage> getMessages(Object memoryId) {
[...]

PR #5550, opened 2026-06-23
Reduce redundant ChatMemoryStore reads in AiServices

cfcromn:
Related to #3284.

  ## Summary
  This PR reduces redundant `ChatMemoryStore.getMessages()` calls in the basic non-streaming `DefaultAiServices` memory path.

  Request preparation now reads chat memory once, applies the optional system message in memory, builds the LLM request messages explicitly from the request
  snapshot, and persists the prepared memory via the existing `ChatMemory.set(...)` API.

  The LLM request message list does not rely on `ChatMemory.set(...)` mutating the list passed to it.

  ## Verification
  - `./mvnw -pl langchain4j -Dtest=AiServicesChatMemoryStoreReadsTest test`
  - `./mvnw -pl langchain4j -Dtest="MessageWindowChatMemoryTest,TokenWindowChatMemoryTest,AiServicesRagChatMemoryBehaviorTest" test`
  - `./mvnw -pl langchain4j -Dtest=AiServicesWithChatMemoryIT test`

  ## Notes
  - Read count without system message: `3 → 2`
  - Read count with system message: `4 → 2`
  - Duplicate persistence from append-style custom stores is out of scope.
  - RAG/tool loop, streaming path, provider-specific code, and public API changes are out of scope.
  - `AiServicesWithChatMemoryIT` completed with build success; 6 tests were skipped due to the `OPENAI_API_KEY` gate.

### DIFF
--- a/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
+++ b/langchain4j/src/main/java/dev/langchain4j/service/DefaultAiServices.java
@@ -262,12 +262,11 @@ class DefaultAiServices<T> extends AiServices<T> {
                         List<ChatMessage> messages = new ArrayList<>();
                         if (context.hasChatMemory()) {
                             systemMessage.ifPresent(chatMemory::add);
-                            messages.addAll(chatMemory.messages());
-                            if (context.storeRetrievedContentInChatMemory) {
-                                chatMemory.add(userMessage);
-                            } else {
-                                chatMemory.add(originalUserMessage);
-                            }
+                            List<ChatMessage> memoryMessages = new ArrayList<>(chatMemory.messages());
+                            messages.addAll(memoryMessages);
+                            memoryMessages.add(
+                                    context.storeRetrievedContentInChatMemory ? userMessage : originalUserMessage);
+                            chatMemory.set(memoryMessages);
                             messages.add(userMessage);
                         } else {
                             systemMessage.ifPresent(messages::add);
--- a/langchain4j/src/test/java/dev/langchain4j/memory/chat/AiServicesChatMemoryStoreReadsTest.java
+++ b/langchain4j/src/test/java/dev/langchain4j/memory/chat/AiServicesChatMemoryStoreReadsTest.java
@@ -1,0 +1,230 @@
+package dev.langchain4j.memory.chat;
+
+import static dev.langchain4j.data.message.AiMessage.aiMessage;
+import static dev.langchain4j.data.message.SystemMessage.systemMessage;
+import static dev.langchain4j.data.message.UserMessage.userMessage;
+import static org.assertj.core.api.Assertions.assertThat;
+
+import dev.langchain4j.data.message.ChatMessage;
+import dev.langchain4j.model.chat.mock.ChatModelMock;
+import dev.langchain4j.service.AiServices;
+import dev.langchain4j.service.MemoryId;
+import dev.langchain4j.service.SystemMessage;
+import dev.langchain4j.service.UserMessage;
+import java.util.List;
+import org.junit.jupiter.api.Test;
+
+class AiServicesChatMemoryStoreReadsTest {
+
+    interface Assistant {
+
+        String chat(@MemoryId int memoryId, @UserMessage String message);
+
+        @SystemMessage("You are helpful")
+        String chatWithSystemMessage(@MemoryId int memoryId, @UserMessage String message);
+
+        @SystemMessage("You are funny")
+        String chatWithAnotherSystemMessage(@MemoryId int memoryId, @UserMessage String message);
+    }
+
+    @Test
+    void should_not_read_chat_memory_store_multiple_times_per_ai_service_call() {
+        HitCountChatMemoryStore store = new HitCountChatMemoryStore();
+        ChatModelMock chatModel = ChatModelMock.thatAlwaysResponds("ok");
+        Assistant assistant = assistant(chatModel, store);
+
+        HitCountChatMemoryStore.HitCounts counts = store.measureHitCounts(() -> assistant.chat(1, "hello"));
+
+        assertThat(counts).isEqualTo(new HitCountChatMemoryStore.HitCounts(2, 2, 0));
+        assertThat(store.getMessages(1)).containsExactly(userMessage("hello"), aiMessage("ok"));
+        assertThat(chatModel.getRequests()).containsExactly(List.of(userMessage("hello")));
+    }
+
+    @Test
+    void should_reduce_chat_memory_store_reads_when_system_message_is_present() {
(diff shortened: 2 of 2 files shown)


dliubarskyi:
Thanks for working on this! The read-count reduction is a real improvement and the tests for the default path are solid. However, I think the current approach has a behavioral regression that we need to
  address before this can go in.

  Behavioral regression with alwaysKeepSystemMessageFirst(true)

  The new addSystemMessage(...) helper always appends the system message to the end of the list:

  private static void addSystemMessage(List<ChatMessage> messages, SystemMessage systemMessage) {
      ...
      messages.add(systemMessage); // always at the end
  }

  But both MessageWindowChatMemory.add() and TokenWindowChatMemory.add() honor the documented alwaysKeepSystemMessageFirst builder option, which inserts the system message at index 0. By bypassing
  ChatMemory.add() and reimplementing only the default placement, this PR silently breaks the true configuration.

  I confirmed it through AiServices with alwaysKeepSystemMessageFirst(true) and a changing system message (e.g. a templated @SystemMessage whose rendered content differs per call):

  ┌─────────┬───────────────────────────────────────────────────────┐
  │         │            Persisted memory & LLM request             │
  ├─────────┼───────────────────────────────────────────────────────┤
  │ main    │ [System, user, ai, user, ...] — system stays first ✅ │
  ├─────────┼───────────────────────────────────────────────────────┤
  │ this PR │ [user, ai, System, user, ...] — system displaced ❌   │
  └─────────┴───────────────────────────────────────────────────────┘

  This affects both the persisted memory and the request sent to the model. The new tests don't catch it because they all use the default (false).

  Root cause / suggested direction

  The deeper issue is that addSystemMessage re-implements memory-internal rules (system-message placement, replace-on-change) in the service layer. Each ChatMemory implementation owns this logic — and it's
  configurable — so duplicating a subset here is fragile and will drift out of sync (which is exactly what happened with alwaysKeepSystemMessageFirst).

  A cleaner direction that still reduces reads: keep delegating the system message to chatMemory.add(systemMessage) (it already handles same-content de-dup, replacement, and alwaysKeepSystemMessageFirst
  correctly), and only collapse the redundant user-message read/write into the single set(...). That keeps the optimization while letting each memory implementation remain the single source of truth for its own
  ordering/eviction rules. Could you explore that approach?

  Either way, please add explicit tests for both alwaysKeepSystemMessageFirst values so this placement is covered going forward.

  One more note: this reduces reads but doesn't address the original duplicate-row symptom in #3284 (a custom append-style store), which you've correctly scoped out — so I'd avoid auto-closing that issue with
  this PR.

cfcromn:
Thanks for the detailed review and for catching this.

You’re right — by reimplementing the system message handling in DefaultAiServices, this PR bypasses the ChatMemory implementation-specific behavior, including alwaysKeepSystemMessageFirst(true). That can change both the persisted memory and the request messages, so I agree this needs to be fixed before the PR can go in.

I’ll update the PR to keep system message handling delegated to chatMemory.add(systemMessage) and limit the optimization to the redundant user-message read/write path. I’ll also add explicit tests for both alwaysKeepSystemMessageFirst(false) and alwaysKeepSystemMessageFirst(true) so the placement behavior is covered.

I’ll also adjust the PR description to avoid auto-closing #3284, since this PR only reduces redundant reads and does not address custom append-style ChatMemoryStore implementations.

cfcromn:
Thanks again for the detailed review.

I updated the PR to delegate system message handling back to `chatMemory.add(systemMessage)` and kept the optimization limited to the user-message read/write path.

I also added coverage for both `alwaysKeepSystemMessageFirst(false)` and `alwaysKeepSystemMessageFirst(true)`, including a changing system message case. The tests now verify both persisted memory and LLM request message ordering.

cfcromn:
Hi @dliubarskyi, gentle ping on this PR.

I updated the implementation based on your feedback:
- system message handling is delegated back to `chatMemory.add(systemMessage)`
- the optimization is now limited to the user-message read/write path
- tests were added for both `alwaysKeepSystemMessageFirst(false)` and `alwaysKeepSystemMessageFirst(true)`

All checks are passing now.  
Could you please take another look when you have time?

Thanks!

dliubarskyi:
Hi @cfcromn, first of all thank you for the thorough work here — the iterations, the detailed verification notes, and especially the added test coverage for both alwaysKeepSystemMessageFirst(false) and (true)
  (including the changing-system-message case) are all genuinely valuable, and you were very responsive to the earlier feedback. 🙏

  After thinking about this more holistically, I've decided to close this PR, and I want to be transparent about why — it's not a reflection on the quality of your work.

  The core reason is direction, not correctness. The optimization here removes one ChatMemoryStore read per invocation on the simple path by switching the user-message persistence from chatMemory.add(...) to
  chatMemory.set(...). That has two downsides I'd rather not take on in 1.x:

  - It makes ChatMemory.set(...) the first thing LangChain4j calls automatically, which contradicts that method's own contract (its Javadoc states it is "never called automatically by LangChain4j").
  - For custom ChatMemory implementations that don't override set(...), it routes user-message persistence through the default clear() + add(...), which is non-atomic and does more IO — the opposite of the
  intended effect for those users.

  More importantly, the real inefficiency isn't this single read — it's that ChatMemory persists eagerly on every mutation (system, user, each assistant/tool message), rewriting the full list each time, which is
  O(n·k) store round-trips per invocation. I'm planning to address this properly in 2.0 by reworking memory around a transactional / unit-of-work model: hydrate once at the start of an AI Service round, keep
  all mutations in memory, and flush once at the end (commit). That collapses the IO to a single read + single write per invocation and, as a bonus, makes a conversation turn atomic. That redesign supersedes
  this optimization entirely, so I'd prefer not to change the set(...) contract in the meantime for a smaller, interim gain.

  Please don't let this discourage you — the analysis and tests you produced here were solid, and I'd genuinely welcome your involvement in the 2.0 memory discussion once it's opened. Thank you again for
  contributing! 🙌

cfcromn:
Thank you for the detailed explanation, and no worries at all.

I completely understand the direction here. After reading your reasoning, I agree that changing the behavior around ChatMemory.set(...) in 1.x could introduce an API/contract concern that is not worth the smaller interim optimization, especially for custom ChatMemory implementations that rely on the default clear() + add(...) behavior.

The transactional / unit-of-work direction for 2.0 also makes a lot of sense to me. Solving this at the invocation level with one hydrate + one commit seems much cleaner than optimizing one specific mutation path in the current eager-persistence model.

Thank you again for reviewing the PR and for the kind words. I learned a lot from working through this and from your feedback. I’d be happy to follow and participate in the 2.0 memory discussion when it opens, especially around the ChatMemory redesign and test coverage for persistence behavior.

PR #5551, opened 2026-06-23
Introduce AgentRegistry for dynamic agents discovery

mariofusco:
## Issue

Closes #

## General checklist


mariofusco:
@carlesarnal Please give a look at this. I'm still keeping this as a draft for now, because there is still something that I want to review and possibly improve, but this pull request implements the idea of `AgentRegistry` that I had in mind.

My suggestion is to give this a try and check in practice if this is a good fit for the Apicurio integration that you have in mind. Ideally I would like to have a working prototype of that integration before merging this pull request. Please let me know if you have any feedback or suggestion. Any feedback is very appreciated.

/cc @geoand

carlesarnal:
Hey, thank you for this.

Looks good overall. I just updated the Apicurio implementation locally against this branch and found a few friction points:

`@AgentRegistrySupplier` requires a static method. This is the main issue for the Apicurio (and probably any external service-backed) integration. The registry client is a CDI bean created at runtime, so you can't get it in a static context. On the Quarkus side we'd need the deployment processor to auto-wire the registry bean, bypassing the annotation  similar to how @McpToolBox works. Not a blocker for this PR, but worth knowing that CDI-based implementations will need special handling.

`discoverAgents(AgenticScope)` is called every iteration. Our implementation hits the Apicurio Registry HTTP API each time. A supervisor running 5-10 iterations means 5-10 HTTP calls per user request. We'll need caching with a TTL on our side, but it might be worth mentioning in the docs that implementations should be mindful of this.

Agent instances should be cached across iterations. `AgenticServices.a2aBuilder(url).build()` creates a new A2A client connection each time. Implementations need to cache AgentInstance objects and only recreate when registry contents change. Again, our problem to solve, but the SPI could note this expectation.

I think most of this if not all is kind of expected, but I think all those points are worth noting.

mariofusco:
@carlesarnal Thanks for feedback, can you please share the prototype where you did your experiments so I could give it a look?

carlesarnal:
> @carlesarnal Thanks for feedback, can you please share the prototype where you did your experiments so I could give it a look?

Sure, I have updated the original work on the quarkus side: https://github.com/quarkiverse/quarkus-langchain4j/pull/2501 — specifically the ApicurioAgentsRegistry class.

mariofusco:
> > @carlesarnal Thanks for feedback, can you please share the prototype where you did your experiments so I could give it a look?
> 
> Sure, I have updated the original work on the quarkus side: [quarkiverse/quarkus-langchain4j#2501](https://github.com/quarkiverse/quarkus-langchain4j/pull/2501) — specifically the ApicurioAgentsRegistry class.

Did you push what you did on that branch? I don't think so since I'm still seeing the old implementation with the  `discoverAgents` method not taking any argument. Am I missing something?

carlesarnal:
> > > @carlesarnal Thanks for feedback, can you please share the prototype where you did your experiments so I could give it a look?
> > 
> > 
> > Sure, I have updated the original work on the quarkus side: [quarkiverse/quarkus-langchain4j#2501](https://github.com/quarkiverse/quarkus-langchain4j/pull/2501) — specifically the ApicurioAgentsRegistry class.
> 
> Did you push what you did on that branch? I don't think so since I'm still seeing the old implementation with the `discoverAgents` method not taking any argument. Am I missing something?

I honestly thought so, done now, sorry for the confusion.

mariofusco:
Thanks for feedback @carlesarnal

I believe that all the issues that you mentioned need to be resolved at the Quarkus extension level, but I agree they all worth to be considered and I will check if I could mitigate at least a few of them on LangChain4j side.

On a different note. I'm rethinking this new API and now I find it quite limited and not satisfying. At the moment we're only using it for dynamic discovery and as it is it makes sense only for the supervisor pattern, but this is not the only use case that you may want to cover. For instance why couldn't I define a sequence of 3 agents where 2 of them are provided by the external registry?

Let me know if you agree with me that this is a valid and interesting use case. I will try to rethink the `AgentRegisty` in those term, but if I do so very likely it will be necessary to turn it into a proper SPI and load the Apicurio implementation via the standard Java service loading mechanism. Is this ok for you?

carlesarnal:
> Thanks for feedback @carlesarnal
> 
> I believe that all the issues that you mentioned need to be resolved at the Quarkus extension level, but I agree they all worth to be considered and I will check if I could mitigate at least a few of them on LangChain4j side.
> 
> On a different note. I'm rethinking this new API and now I find it quite limited and not satisfying. At the moment we're only using it for dynamic discovery and as it is it makes sense only for the supervisor pattern, but this is not the only use case that you may want to cover. For instance why couldn't I define a sequence of 3 agents where 2 of them are provided by the external registry?
> 
> Let me know if you agree with me that this is a valid and interesting use case. I will try to rethink the `AgentRegisty` in those term, but if I do so very likely it will be necessary to turn it into a proper SPI and load the Apicurio implementation via the standard Java service loading mechanism. Is this ok for you?

I do agree with that use case yes. Making it an actual SPI is absolutely fine, actually, it was written that way in my PR.

mariofusco:
> I do agree with that use case yes. Making it an actual SPI is absolutely fine, actually, it was written that way in my PR.

Ok, since this will imply a total rewriting, for now I will keep this pull request as it is and attempt that different implementation on another branch. I will keep you posted on my progresses.

glaforge:
Hi @mariofusco 

A couple of thoughts on this pull request.

## Synchronous / blocking agent discovery

The `Collection<AgentInstance> discoverAgents(AgenticScope scope)` is synchronous / blocking, and it's called at every `nextAction()` step, right? I'm wondering how this would play out with integration with some cloud based / cloud hosted agent registry beyond the in-memory one. I mean, there could be some significant latency, or hammering the remote service. 

I'm wondering if you have considered something asynchronous, or some policy that wouldn't call a potential remote registry too frequently? Maybe having some caching agent registry could help or would be needed to be developed by a third-party in their implementation? (or it could be provided by LangChain4j too)

What about adding some `init()` / `start()` / `close()` (Autocloseable maybe?) methods to the `AgentRegistry` interface? Maybe a registry would need to handle auth, or closing resources properly, etc. 

## Refreshing already known agents but updated

In `refreshRegistryAgents()` you check if an agent is present or not (`putIfAbsent()`) but what about the case where an existing agent has been updated (with a different config, different tools..)? It wouldn't be refreshed.

mariofusco:
Thanks for feedback @glaforge. As you probably have seen from my former comment I'm rethinking and about to rewrite this whole feature with a slightly different and broader scope. I will take count of your suggestion while implementing it.

mariofusco:
@carlesarnal See https://github.com/langchain4j/langchain4j/pull/5653#issuecomment-4844633612

mariofusco:
Replaced by https://github.com/langchain4j/langchain4j/pull/5653

PR #5553, opened 2026-06-23
fix: Map ContainsString filter to regex in MongoDB Atlas metadata filter mapper

### LINKED ISSUE #5552: [BUG] MongoDB Atlas embedding store throws UnsupportedOperationException for ContainsString filter
**Describe the bug**
`MongoDbMetadataFilterMapper.map()` does not handle the core filter type `dev.langchain4j.store.embedding.filter.comparison.ContainsString`. Passing a `ContainsString` filter to `MongoDbEmbeddingStore.search()` or `removeAll(Filter)` falls into the default branch and throws `UnsupportedOperationException`. Sibling stores (milvus, pgvector, azure-cosmos-nosql) already handle `ContainsString`; only mongodb-atlas is missing it.

**Log and Stack trace**
```
java.lang.UnsupportedOperationException: Unsupported filter type: dev.langchain4j.store.embedding.filter.comparison.ContainsString
	at dev.langchain4j.store.embedding.mongodb.MongoDbMetadataFilterMapper.map(MongoDbMetadataFilterMapper.java)
```

**To Reproduce**
```java
embeddingStore.search(EmbeddingSearchRequest.builder()
    .queryEmbedding(embedding)
    .filter(new ContainsString("name", "lang"))
    .build());
```

**Expected behavior**
The filter is applied. `ContainsString` checks a literal substring (`str.contains(value)`), so it should map to a `$regex` match on `metadata.<key>` with the value treated literally. This is applied in the post-search `$match` aggregation stage and in `removeAll`'s `deleteMany`, both of which support `$regex`.

- LangChain4j version: 1.17.0-beta27-SNAPSHOT
- LLM(s) used: N/A
- Java version: 17

thswlsqls:
## Issue
Closes #5552

## Change
`MongoDbMetadataFilterMapper.map()` had no branch for the core filter type `ContainsString`, so it fell through to the default branch and threw `UnsupportedOperationException`. Sibling stores (milvus, pgvector, azure-cosmos-nosql) already handle it; this closes that parity gap.

Added a `ContainsString` branch before the default `else`, plus a `mapContains` method that renders `Filters.regex(metadata.<key>, Pattern.quote(value))`.

`ContainsString.test()` uses `str.contains(value)`, a literal substring check. `Pattern.quote(...)` escapes the value so regex metacharacters (`.`, `*`, `(`) match literally, matching milvus's `LIKE %value%` semantics.

Where this runs: `ContainsString` is not a `$vectorSearch` pre-filter. It is applied in the post-search `$match` aggregation stage in `MongoDbEmbeddingStore.search()` and in `removeAll(Filter)`'s `deleteMany`. Both are ordinary query contexts that support `$regex` (per MongoDB docs), so the fix is valid at runtime.

Scope: the new branch, `mapContains`, and two imports, plus two spotless-ratchet reformats of pre-existing lines in the touched file (long `throw` wrap, a stray space) required for `spotless:check` to pass.

## General checklist

### DIFF
--- a/langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbMetadataFilterMapper.java
+++ b/langchain4j-mongodb-atlas/src/main/java/dev/langchain4j/store/embedding/mongodb/MongoDbMetadataFilterMapper.java
@@ -2,6 +2,7 @@ package dev.langchain4j.store.embedding.mongodb;
 
 import com.mongodb.client.model.Filters;
 import dev.langchain4j.store.embedding.filter.Filter;
+import dev.langchain4j.store.embedding.filter.comparison.ContainsString;
 import dev.langchain4j.store.embedding.filter.comparison.IsEqualTo;
 import dev.langchain4j.store.embedding.filter.comparison.IsGreaterThan;
 import dev.langchain4j.store.embedding.filter.comparison.IsGreaterThanOrEqualTo;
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thanks!

PR #5561, opened 2026-06-24
fix: Populate logProbs in OpenAI streaming response metadata

### LINKED ISSUE #5560: [BUG] OpenAI streaming response metadata never returns logProbs
**Describe the bug**
`OpenAiStreamingChatModel` never returns logprobs in its response metadata. With `logprobs(true)`, the streaming path returns `OpenAiChatResponseMetadata.logProbs() == null`, while the non-streaming `OpenAiChatModel` returns the populated list for the same request — a parity gap between two implementations of the same interface.

The cause is in `OpenAiStreamingResponseBuilder`: it reads `finishReason()` and `delta()` from each streamed choice but never `choice.logprobs()`, and `buildMetadata()` never calls `.logProbs(...)`. OpenAI streams the per-token `logprobs.content` array on each chunk, so the data arrives but is discarded.

PR #4306 (Closes #4298) added logprobs to the non-streaming `OpenAiChatModel` only; the streaming sibling was missed.

**Log and Stack trace**
N/A

**To Reproduce**
1. Build an `OpenAiStreamingChatModel` with `.logprobs(true)`.
2. Call `chat(...)` and read `ChatResponse.metadata()`.
3. Cast to `OpenAiChatResponseMetadata` and call `logProbs()`.
4. It returns `null`.

**Expected behavior**
The streaming path returns the same accumulated logprobs as the non-streaming path. When logprobs are not requested, it stays `null`.

- LLM(s) used: OpenAI
- Java version: 17

**Additional context**
N/A

thswlsqls:
## Issue
Closes #5560

## Change
`OpenAiStreamingResponseBuilder` now accumulates logprobs from streamed choices and exposes them in the response metadata.

- `append(ChatCompletionResponse)` reads `choice.logprobs().content()` from each chunk and appends it, in order, to a thread-safe `CopyOnWriteArrayList` (this builder is called from streaming threads).
- `buildMetadata()` converts the accumulated content with the existing `OpenAiUtils.logProbsFrom(...)` helper and sets it via `.logProbs(...)`. When nothing was accumulated (logprobs not requested), it stays `null`.

This closes the parity gap with the non-streaming `OpenAiChatModel.doChat()`, which already populates logprobs. PR #4306 added logprobs only to the non-streaming path; this completes the streaming sibling.

No public signature changes; behavior is unchanged when logprobs are not requested (backward compatible).

Added two unit tests to `OpenAiStreamingResponseBuilderTest` (no API key or mock needed): one asserts logprobs accumulate in order across chunks, one asserts the field stays `null` when no logprobs are present.

## General checklist

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingResponseBuilder.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/OpenAiStreamingResponseBuilder.java
@@ -3,6 +3,7 @@ package dev.langchain4j.model.openai;
 import static dev.langchain4j.internal.Utils.isNullOrBlank;
 import static dev.langchain4j.internal.Utils.isNullOrEmpty;
 import static dev.langchain4j.model.openai.internal.OpenAiUtils.finishReasonFrom;
+import static dev.langchain4j.model.openai.internal.OpenAiUtils.logProbsFrom;
 import static dev.langchain4j.model.openai.internal.OpenAiUtils.tokenUsageFrom;
 import static java.util.stream.Collectors.toList;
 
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thank you!

PR #5563, opened 2026-06-24
fix: Generate native IN clause for single-type IN/NOT IN filters in Oracle store

### LINKED ISSUE #5562: [BUG] Oracle store never generates native IN clause for single-type IN/NOT IN filters
**Describe the bug**
In `langchain4j-oracle`, `SQLFilters.SQLInFilter.create(...)` should emit a native `IN (?, ...)` / `NOT IN (?, ...)` clause when all `IsIn`/`IsNotIn` values map to a single non-CLOB Oracle type. That fast path is dead code: the method builds one iterator (`sqlTypes.iterator()`) but advances a *second*, fresh one (`sqlTypes.iterator().next()`), so the guard `!sqlTypeIterator.hasNext()` is always false. Uniform-type filters therefore always fall through to the OR-chain, and the documented JavaDoc output `name IN (?, ?, ?)` never occurs.

This is **result-equivalent**: the OR-chain and the native IN path return identical rows for all inputs, including absent metadata keys. The problem is dead code, a broken JavaDoc contract, and less efficient generated SQL, not wrong query results.

**To Reproduce**
`SQLFilters.create(metadataKey("x").isIn(1, 2), (k,t) -> k).toSQL()` returns `(NVL(x = ?, false) OR NVL(x = ?, false))` instead of `NVL(x IN (?, ?), false)`.

**Expected behavior**
Uniform single non-CLOB type IN/NOT IN filters generate the documented native `IN`/`NOT IN` clause.

- LangChain4j version: main
- LLM(s) used: N/A
- Java version: 17

**Additional context**
`langchain4j-oracle/.../store/embedding/oracle/SQLFilters.java`.

thswlsqls:
## Issue
Closes #5562

## Change
`SQLFilters.SQLInFilter.create(...)` was meant to emit a native `IN (?, ...)` / `NOT IN (?, ...)` clause when all comparison values map to a single non-CLOB Oracle type. That path was dead code: the method created one iterator but advanced a separate, fresh one (`sqlTypes.iterator().next()`), so the guard `!sqlTypeIterator.hasNext()` was always false for any non-empty set. Uniform-type `IsIn`/`IsNotIn` filters therefore always expanded to an OR-chain, the `SQLInFilter` constructor/`setParameters` were never reached, and the class JavaDoc contract (`name IN (?, ?, ?)`) was never met.

Fix: replace the buggy double-iterator guard with `sqlTypes.size() == 1`, then take the single type and use the native path when it is not CLOB. Non-uniform and CLOB inputs keep the existing OR-chain fallback (backward compatible).

This is **result-equivalent**: the OR-chain and native IN path return identical rows for all inputs (including absent metadata keys). Only the generated SQL changes — single-type filters now produce `NVL(x IN (?, ?), false)` instead of `(NVL(x = ?, false) OR NVL(x = ?, false))` — which is the documented, more efficient form. No API or behavioural change.

Added `SQLFilterTest` (new `*Test`, runs under surefire; the module previously had no unit tests) asserting native `IN`/`NOT IN` output for uniform NUMBER inputs. Verified fail-then-pass on JDK 17.

## General checklist

### DIFF
--- a/langchain4j-oracle/src/main/java/dev/langchain4j/store/embedding/oracle/SQLFilters.java
+++ b/langchain4j-oracle/src/main/java/dev/langchain4j/store/embedding/oracle/SQLFilters.java
@@ -409,13 +409,14 @@ final class SQLFilters {
 
             Set<OracleType> sqlTypes =
                     comparisonValues.stream().map(SQLFilters::toOracleType).collect(Collectors.toSet());
-            Iterator<OracleType> sqlTypeIterator = sqlTypes.iterator();
-            OracleType sqlType = sqlTypes.iterator().next();
 
             // IN and NOT IN conditions can operate on multiple data types, but the keyMapper needs to return the key as
             // a single data type. The IN and NOT IN conditions cannot operate on a CLOB, even if that's the only type.
-            if (!sqlTypeIterator.hasNext() && sqlType != OracleType.CLOB) {
-                return new SQLInFilter(key, keyMapper, isIn, comparisonValues, sqlType);
+            if (sqlTypes.size() == 1) {
+                OracleType sqlType = sqlTypes.iterator().next();
+                if (sqlType != OracleType.CLOB) {
+                    return new SQLInFilter(key, keyMapper, isIn, comparisonValues, sqlType);
+                }
             }
 
             // Replicate IN and NOT IN conditions as a sequence of OR conditions: "key = value0 OR key = value1 OR ..."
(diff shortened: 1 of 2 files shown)


dliubarskyi:
Hi @psilberk, could you please review?

psilberk:
> Hi @psilberk, could you please review?

Yes, looking into it now.

fmeheust (approved):
Looks good to me

psilberk:
Looks good, thanks!

PR #5564, opened 2026-06-24
Support input_image format for OpenAI-compatible images

### LINKED ISSUE #4551: [BUG] Protocol for submitting images to models doesn't match what some agents do
This is arguably not quite a bug, but here goes: langchain4j submits images in the following manner:

```json
{
  "model" : "...",
  "messages" : [ {
    "role" : "system",
    "content" : "Describe the image"
  }, {
    "role" : "user",
    "content" : [ {
      "type" : "image_url",
      "image_url" : {
        "url" : "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD.....",
        "detail": "low"
       }
    }
...
```

On some servers this yields a 500 Internal Server Error. Yeah, I know. Other agents _do_ work and from inspecting their code, it appears that this is because they send images slightly differently, like this:

```json
{"type":"message","role":"user","content":[{"type":"input_text","text":"What do you see in [test.jpg 798x800]"},{"type":"input_image","image_url":"data:image/jpeg;base64,/9j/4AAQSkZJRgA......"}]}
```

Note the differences:

- `type` is now `input_image` not `image_url`
- `image_url` is now a string and not an object
- there is no `detail` field

----
 
- LangChain4j version: 1.11.0
- LLM(s) used: gpt5 via a non-official openai serving stack (Oracle Code Assist)

I checked Codex CLI and Kilo Code. Both use this alternative style and both work. 

This is arguably a server bug. However, there is nothing I can do to change the server. Unfortunately, there's also nothing I can do to change langchain4j other than submit a patch and wait for a release. So. Could we add an API or system property to change the serialization style for images?

smf-h:
## Summary
  - Add an opt-in `useInputImageFormat(boolean)` builder option to `OpenAiChatModel`
    and `OpenAiStreamingChatModel` for OpenAI-compatible servers that expect the
    `input_image` image payload.
  - When enabled, image content is serialized as
    `{"type":"input_image","image_url":"data:..."}` (string `image_url`, no `detail`)
    instead of the default Chat Completions
    `{"type":"image_url","image_url":{"url":"data:...","detail":"..."}}`.
  - Disabled by default — existing `image_url` object payload is unchanged.
  - `Content` now supports serialization **and** deserialization of both the
    object-shaped `image_url` and the string-shaped `input_image` form.

  Fixes #4551

  ## Usage
  ```java
  OpenAiChatModel model = OpenAiChatModel.builder()
          .baseUrl(...)
          .apiKey(...)
          .modelName(...)
          .useInputImageFormat(true) // opt in to the input_image payload
          .build();
  ```
  Testing

  - Added OpenAiImageContentFormatTest covering:
    - default image_url object format (serialization) for chat and streaming chat
    - input_image string format when useInputImageFormat(true) is set
    - deserialization of both the object-shaped image_url and string-shaped
  input_image forms
  - <fill in: ran mvn -pl langchain4j-open-ai test locally / relying on CI>

  Checklist

  - [x] No breaking changes (default behavior unchanged; new option defaults to false)
  - [ ] No new dependencies added
  - [ ] New code follows existing OpenAI builder conventions (boolean opt-in, mirrors sendThinking/returnThinking)
  - [ ] Tests added and passing

### DIFF
--- a/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java
+++ b/langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java
@@ -89,21 +89,31 @@ public class OpenAiUtils {
     public static final String DEFAULT_USER_AGENT = "langchain4j-openai";
 
     public static List<Message> toOpenAiMessages(List<ChatMessage> messages) {
-        return toOpenAiMessages(messages, false, null);
+        return toOpenAiMessages(messages, false, null, false);
     }
 
     public static List<Message> toOpenAiMessages(
             List<ChatMessage> messages, boolean sendThinking, String thinkingFieldName) {
+        return toOpenAiMessages(messages, sendThinking, thinkingFieldName, false);
+    }
+
+    public static List<Message> toOpenAiMessages(
+            List<ChatMessage> messages, boolean sendThinking, String thinkingFieldName, boolean useInputImageFormat) {
         return messages.stream()
-                .map(message -> toOpenAiMessage(message, sendThinking, thinkingFieldName))
+                .map(message -> toOpenAiMessage(message, sendThinking, thinkingFieldName, useInputImageFormat))
                 .collect(toList());
     }
 
     public static Message toOpenAiMessage(ChatMessage message) {
-        return toOpenAiMessage(message, false, null);
+        return toOpenAiMessage(message, false, null, false);
     }
 
     public static Message toOpenAiMessage(ChatMessage message, boolean sendThinking, String thinkingFieldName) {
+        return toOpenAiMessage(message, sendThinking, thinkingFieldName, false);
+    }
+
+    public static Message toOpenAiMessage(
+            ChatMessage message, boolean sendThinking, String thinkingFieldName, boolean useInputImageFormat) {
         if (message instanceof SystemMessage) {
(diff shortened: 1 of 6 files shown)


langchain4j-open-ai/src/main/java/dev/langchain4j/model/openai/internal/OpenAiUtils.java
```diff
@@ -214,6 +216,13 @@ private static dev.langchain4j.model.openai.internal.chat.Content toOpenAiConten
     }
 
     private static dev.langchain4j.model.openai.internal.chat.Content toOpenAiContent(ImageContent content) {
+        if (INPUT_IMAGE_CONTENT_FORMAT.equals(System.getProperty(IMAGE_CONTENT_FORMAT_PROPERTY))) {
```
dliubarskyi: Using system property is a bad idea, let's use a normal property instead in the model builder

smf-h: Agreed. I replaced the system property with a normal builder option: `useInputImageFormat(true)`.   
  The default remains `image_url`, while `input_image` is now opt-in per model instance. Added tests  
  for both request shapes, including streaming.

smf-h:
@dliubarskyi Thanks! I addressed the review feedback, replaced the system property with the builder option, and added tests for both request formats including streaming. All checks are passing. Could you please take another look?

dliubarskyi (approved):
@smf-h thanks!

PR #5566, opened 2026-06-24
http-client-jdk: set Content-Type for multipart/form-data requests

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## Issue
N/A — small standalone fix.

## Change
This PR has been **trimmed down** to a single, self-contained core fix, following @glaforge's feedback that the TwelveLabs integration itself belongs in `langchain4j-community`. The integration has moved to langchain4j/langchain4j-community#706; this PR now contains only the core bug fix that I found underneath it.

**The fix:** the JDK `HttpClient` built `multipart/form-data` request bodies but never set the `Content-Type` header (with the boundary), so servers that parse multipart requests rejected them (e.g. `content_type_invalid`). The OkHttp and Apache clients already set this header — this brings the JDK client in line.

- `JdkHttpClient` now sets `Content-Type: multipart/form-data; boundary=...` whenever the request carries form-data fields or files.
- `MultipartBodyPublisher.contentType()` exposes the boundary-bearing header value so the body and header always agree.
- Added a regression test asserting the advertised boundary matches the one used to delimit the body.

This is **non-breaking and additive** — it only adds a header that was previously missing on multipart requests; no existing behaviour changes.

### How it was tested
- `MultipartBodyPublisherTest` — added a regression test for the `Content-Type`/boundary fix; all JDK http-client unit tests are green.

## General checklist

### DIFF
--- a/http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
+++ b/http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
@@ -6,9 +6,9 @@ import static java.util.stream.Collectors.joining;
 
 import dev.langchain4j.exception.HttpException;
 import dev.langchain4j.exception.TimeoutException;
+import dev.langchain4j.http.client.FormDataFile;
 import dev.langchain4j.http.client.HttpClient;
 import dev.langchain4j.http.client.HttpRequest;
-import dev.langchain4j.http.client.FormDataFile;
 import dev.langchain4j.http.client.SuccessfulHttpResponse;
 import dev.langchain4j.http.client.sse.ServerSentEventListener;
 import dev.langchain4j.http.client.sse.ServerSentEventParser;
@@ -58,7 +58,7 @@ public class JdkHttpClient implements HttpClient {
         } catch (HttpTimeoutException e) {
             throw new TimeoutException(e);
         } catch (InterruptedException e) {
-            Thread.currentThread().interrupt();  
+            Thread.currentThread().interrupt();
             throw new RuntimeException(e);
         } catch (IOException e) {
             throw new RuntimeException(e);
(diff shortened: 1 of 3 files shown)


glaforge:
I think the PR should be made against the langchain4j-community repository instead of the langchain4j one.

mohit-twelvelabs:
Thanks @glaforge, that makes total sense — appreciate the steer. I've moved the TwelveLabs (Marengo) embedding model integration to the community repo: **langchain4j/langchain4j-community#706**. It follows the existing Cohere model module's structure and conventions there (parent POM, BOM entry, `langchain4j-http-client` client), and the live IT passes green against the Marengo endpoint.

While building the integration I hit a genuine core bug that's independent of TwelveLabs: the JDK `HttpClient` builds `multipart/form-data` bodies but never sets the `Content-Type` header (with the boundary), so any multipart request through it gets rejected (the OkHttp and Apache clients already set it). Since that's a real core fix that benefits any multipart user, I've **trimmed this PR (#5566) down to just that http-client fix** — the integration module/docs/BOM changes are gone, and it's now a small standalone core PR with a regression test (`MultipartBodyPublisherTest`). New head commit: 6fd9ff999.

Happy to adjust scope further if you'd prefer I close this and fold the header into the community module instead (the community PR already sets the header itself the same way Cohere does, so it doesn't block on this). Thanks again!

http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
```diff
@@ -116,6 +116,9 @@ private java.net.http.HttpRequest toJdkRequest(HttpRequest request) {
             }
         } else {
             bodyPublisher = ofMultipartData(request.formDataFields(), request.formDataFiles());
+            // The multipart body is delimited by a boundary that the server needs in the Content-Type header to
```
dliubarskyi:
```suggestion
```

http-clients/langchain4j-http-client-jdk/src/main/java/dev/langchain4j/http/client/jdk/JdkHttpClient.java
```diff
@@ -116,6 +116,9 @@ private java.net.http.HttpRequest toJdkRequest(HttpRequest request) {
...
+            // parse it; the OkHttp and Apache clients do the same.
```
dliubarskyi:
```suggestion
```

dliubarskyi (approved):

PR #5574, opened 2026-06-24
fix: skip recording token usage metric when token count is null

### LINKED ISSUE #5571: [BUG] Micrometer metrics listener throws NPE on null input/output token count
**Describe the bug**

`MicrometerMetricsChatModelListener` throws a `NullPointerException` when a chat response carries a `TokenUsage` whose input or output count is `null`. `TokenUsage` documents each component as "null if unknown", but the listener guards only the `TokenUsage` container, not the individual counts, and the private `addTokenMetric` takes a primitive `int`, so the nullable `Integer` count auto-unboxes and throws. Input is recorded first, so the response loses both token metrics. The framework catches and logs the exception, so the call does not fail — the loss is silent.

**Log and Stack trace**

`java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because "TokenUsage.inputTokenCount()" is null`

**To Reproduce**

Attach the listener to a chat model whose response has a partial `TokenUsage`, e.g. `new TokenUsage(null, 20)` or `new TokenUsage()`. Ollama and Google GenAI streaming produce such partial usage on normal paths.

**Expected behavior**

Record the token type that is present and skip the one whose count is `null`, without throwing.

- LLM(s) used: N/A (provider-independent)
- Java version: 17

**Additional context**

OTel `gen_ai.client.token.usage` records a data point per known token type, so an unknown count should be skipped.

thswlsqls:
## Issue
Closes #5571

## Change

`MicrometerMetricsChatModelListener` threw a `NullPointerException` when a chat response had a `TokenUsage` with a `null` input or output count. `TokenUsage` documents each component as "null if unknown", but `addTokenMetric` took a primitive `int`, so the nullable `Integer` count auto-unboxed and threw. Because input was recorded first, the response lost both token metrics; `ChatModelListenerUtils.onResponse` caught and logged the exception, so token usage was silently dropped and a warning logged on every such response. Providers reach this on normal paths (e.g. Ollama `InternalOllamaHelper`, Google GenAI streaming `GoogleGenAiStreamingChatModel`).

Changed `addTokenMetric`'s count parameter to `Integer` and skip recording when it is `null`, so the known token type is recorded and the unknown one skipped. This matches OpenTelemetry `gen_ai.client.token.usage`, which records a data point per known token type. Tags, metric name and registration are unchanged, as is behaviour when both counts are present.

Added unit tests for null input, null output, and all-null counts (each reproduces the NPE before the fix). No new dependencies.

## General checklist

### DIFF
--- a/langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
+++ b/langchain4j-micrometer-metrics/src/main/java/dev/langchain4j/micrometer/metrics/listeners/MicrometerMetricsChatModelListener.java
@@ -75,7 +75,13 @@ public class MicrometerMetricsChatModelListener implements ChatModelListener {
     }
 
     private void addTokenMetric(
-            ChatModelResponseContext responseContext, OTelGenAiTokenType tokenType, int tokenCount) {
+            ChatModelResponseContext responseContext, OTelGenAiTokenType tokenType, Integer tokenCount) {
+
+        if (tokenCount == null) {
+            // Token counts are nullable (TokenUsage documents each component as "null if unknown").
+            // Skip recording a token type whose count is unknown instead of failing the whole response.
+            return;
+        }
 
         DistributionSummary.builder(OTelGenAiMetricName.TOKEN_USAGE.value())
                 .baseUnit("tokens")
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thanks!

PR #5577, opened 2026-06-25
fix: Escape backslashes in Milvus metadata filter values

### LINKED ISSUE #5576: [BUG] Milvus: metadata filter values containing a backslash produce a broken filter expression
**Describe the bug**
`MilvusMetadataFilterMapper.formatValue()` escapes double quotes (`"` → `\"`) but not backslashes. Milvus treats backslash as the escape character in string literals (`\\` represents a literal `\`), so a filter value containing a backslash (e.g. a Windows path) produces a malformed expression — either misinterpreted (`\b` → backspace) or rejected as an invalid escape sequence. This affects every comparison operator, `IsIn`/`IsNotIn`, and `removeAll(ids)`, since all route through `formatValue`/`formatValues`.

**Log and Stack trace**
N/A — produces wrong matches or a Milvus expression parse error rather than a Java exception.

**To Reproduce**
```java
Filter filter = metadataKey("path").isEqualTo("C:\\reports");
// generated expr: metadata["path"] == "C:\reports"   <-- \r misinterpreted
// expected expr:  metadata["path"] == "C:\\reports"
```

**Expected behavior**
Backslashes in string values are escaped (`\` → `\\`) before quotes, so the generated expression is valid and matches the intended documents.

- LangChain4j version: 1.7.0 (current main)
- LLM(s) used: N/A
- Java version: 17

**Additional context**
Sibling stores already escape (e.g. azure-cosmos-nosql, infinispan, pgvector). Reference: Milvus Scalar Filtering Rules (https://milvus.io/docs/boolean.md).

thswlsqls:
## Issue
Closes #5576

## Change
`MilvusMetadataFilterMapper.formatValue()` escaped double quotes but not backslashes. Milvus treats backslash as the escape character in string literals (`\\` represents a literal `\`), so a value containing a backslash (e.g. a Windows path) produced a malformed expression — either misinterpreted (`\b` → backspace) or rejected as an invalid escape sequence. This affected every operator, `IsIn`/`IsNotIn`, and `removeAll(ids)`, since all route through `formatValue`/`formatValues`.

Fix: escape backslashes first, then double quotes (`replace("\\", "\\\\").replace("\"", "\\\"")`), mirroring sibling stores (azure-cosmos-nosql, infinispan, pgvector). One production line. Added `MilvusMetadataFilterMapperTest` (pure unit test) asserting the exact generated expression for backslash, backslash+quote, collection values, and a no-special-character regression. Values without backslashes are unchanged — the only behaviour change is for backslash values, which previously produced an invalid expression.

## General checklist

- [ ] I have manually verified that the `MilvusEmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

### DIFF
--- a/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusMetadataFilterMapper.java
+++ b/langchain4j-milvus/src/main/java/dev/langchain4j/store/embedding/milvus/MilvusMetadataFilterMapper.java
@@ -116,8 +116,8 @@ class MilvusMetadataFilterMapper {
 
     private static String formatValue(Object value) {
         if (value instanceof String stringValue) {
-            // Escape double quotes by replacing them with \"
-            final String escapedValue = stringValue.replace("\"", "\\\"");
+            // Escape backslashes first, then double quotes (Milvus treats backslash as the escape character)
+            final String escapedValue = stringValue.replace("\\", "\\\\").replace("\"", "\\\"");
             return "\"" + escapedValue + "\"";
         } else if (value instanceof UUID) {
             return "\"" + value + "\"";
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thanks!

PR #4603, opened 2026-02-20
Add strict option for JSON schema response format to Mistral AI

xhaggi:
## Issue
Closes #4602

## Change
This adds a builder method to the `MistralAiChatModelBuilder` to configure strict JSON schema for response format, as is done in the OpenAI implementation.

## General checklist
- [x] There are no breaking changes (API, behaviour)
- [x] I have added unit and/or integration tests for my change
- [x] The tests cover both positive and negative cases
- [x] I have manually run all the unit and integration tests in the module I have added/changed, and they are all green
- [x] I have manually run all the unit and integration tests in the [core](https://github.com/langchain4j/langchain4j/tree/main/langchain4j-core) and [main](https://github.com/langchain4j/langchain4j/tree/main/langchain4j) modules, and they are all green

- [x] I have added/updated the [documentation](https://github.com/langchain4j/langchain4j/tree/main/docs/docs)
- [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

## Checklist for adding new maven module

- [ ] I have added my new module in the root `pom.xml` and `langchain4j-bom/pom.xml`

## Checklist for adding new embedding store integration

- [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [ ] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

## Checklist for changing existing embedding store integration

- [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

PR #5583, opened 2026-06-25
fix: Honor strictJsonSchema option in MistralAiStreamingChatModel

### LINKED ISSUE #5582: [BUG] strictJsonSchema option is ignored by MistralAiStreamingChatModel
**Title:** [BUG] strictJsonSchema option is ignored by MistralAiStreamingChatModel

**Describe the bug**
`MistralAiStreamingChatModel` cannot enable strict JSON schema. Its builder has no `strictJsonSchema` setter, and `doChat()` passes the literal `false` when building the request, so the `strict` flag in the JSON schema response format is always `false`.

The synchronous `MistralAiChatModel` supports this option, so the two models behave differently for the same request. This is an incomplete fix from PR #4603, which added strict JSON schema support to the synchronous model only.

**Log and Stack trace**
N/A

**To Reproduce**
```java
StreamingChatModel model = MistralAiStreamingChatModel.builder()
        .apiKey(System.getenv("MISTRAL_AI_API_KEY"))
        .modelName("mistral-small-latest")
        // no way to enable strict JSON schema
        .build();
```
The outgoing request body always contains `"strict":false` even when a JSON schema response format is set.

**Expected behavior**
`MistralAiStreamingChatModel` should expose a `strictJsonSchema` builder option and send `"strict":true` when it is enabled, matching `MistralAiChatModel`.

- LLM(s) used: MistralAI
- Java version: 17

**Additional context**
N/A

---

thswlsqls:
## Issue
Closes #5582

## Change
`MistralAiStreamingChatModel.doChat()` passed the literal `false` for the `strict` flag and the builder had no setter, so strict JSON schema was always disabled for the streaming model, unlike the synchronous `MistralAiChatModel`. This is an incomplete fix from PR #4603, which wired the option into the synchronous model only.

This mirrors the synchronous model: a `strictJsonSchema` field initialized via `getOrDefault(builder.strictJsonSchema, false)`, a builder field and setter, and the field passed to `createMistralAiRequest` instead of the hardcoded `false`.

```java
MistralAiChatCompletionRequest request =
        createMistralAiRequest(chatRequest, safePrompt, randomSeed, true, sendThinking, strictJsonSchema);
```

The default stays `false`, so existing callers are unaffected.

Added `MistralAiStreamingChatModelStrictJsonSchemaTest` (`MockHttpClient`, no API key): a positive test asserts the request body contains `"strict":true` when `strictJsonSchema(true)` is set, and a regression test asserts `"strict":false` when it is not.

## General checklist

Spotless: jgit ratchet fails inside the worktree ("Cannot find git repository"), so the two changed files were copied to the main checkout and `spotless:check -pl langchain4j-mistral-ai` was run there (0 files needed changes, both already clean); the temporary copies were then reverted.

### DIFF
--- a/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiStreamingChatModel.java
+++ b/langchain4j-mistral-ai/src/main/java/dev/langchain4j/model/mistralai/MistralAiStreamingChatModel.java
@@ -46,6 +46,7 @@ public class MistralAiStreamingChatModel implements StreamingChatModel {
     private final boolean sendThinking;
     private final List<ChatModelListener> listeners;
     private final Set<Capability> supportedCapabilities;
+    private final boolean strictJsonSchema;
     private final ChatRequestParameters defaultRequestParameters;
 
     @SuppressWarnings({"unchecked"})
@@ -66,6 +67,7 @@ public class MistralAiStreamingChatModel implements StreamingChatModel {
         this.sendThinking = getOrDefault(builder.sendThinking, false);
         this.listeners = copy(builder.listeners);
         this.supportedCapabilities = copy(builder.supportedCapabilities);
+        this.strictJsonSchema = getOrDefault(builder.strictJsonSchema, false);
         this.defaultRequestParameters = initDefaultRequestParameters(builder);
     }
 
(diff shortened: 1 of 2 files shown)


dliubarskyi (approved):
@thswlsqls thanks!

